Issue #117, opened 2018-08-20
Add typings for component props


levithomason:
### Problem description
We are missing or have removed props typings in our components.

### Proposed solution
See #33 fo the discussion and agreement on this proposal.  Let's do this one component and PR at a time to give room for addressing issues that will arise.

1. Add props typings only for `handledProps` as this is most helpful to the user.
1. Make all props interfaces extensible so users can pass any prop value they wish (also aligns with our practice of spreading rest props).
1. Provide a `*Strict` named interface which is not extensible.  This allows users to lock allowed props to strictly those supported by the component if they wish, see https://github.com/Semantic-Org/Semantic-UI-React/issues/2836.  Our props interfaces will union this interface making it extensible.

### TODO

- [x] Accordion #33
- [x] Avatar #129
- [x] Button  #130
- [x] Chat #131
- [x] Divider #132
- [x] Grid #133
- [x] Header #139
- [x] Icon #137
- [x] Image #147 
- [x] Input #140
- [x] ItemLayout #142
- [x] Label #148
- [x] Layout #143 
- [x] List #153 
- [x] Menu #155
- [x] Provider
- [x] Text #154
- [x] Radio #174 
- [x] Segment #175

levithomason:
It is worth nothing that we have a few global props such as `variables`, `styles`, and `as`.  This actually excludes some edge case components like Portal and Provider.  However, we should consider how we advertise these prop in docs and enforce them in tests.

kuzhelov:
lets close it as main concern expressed by this issue is addressed now

PR #137, opened 2018-08-27
feat(Icon): add props interface

mnajdova:
This PR introduces props interface for the Icon component, as part of #117


src/components/Icon/Icon.tsx
```diff
@@ -4,16 +4,44 @@ import { customPropTypes, UIComponent, createShorthandFactory } from '../../lib'
...
+  bordered?: boolean
+  circular?: boolean
+  className?: string
+  disabled?: boolean
+  font?: boolean | string
+  name?: string
+  size?: IconSize
+  svg?: boolean
+  xSpacing?: IconXSpacing
+  accessibility?: object | Function
```
kuzhelov, line 33: just as a side note - we should introduce it as a type consumed from `accessibility` module

kuzhelov (approved):

src/components/Icon/Icon.tsx
mnajdova: Sure, I was defining now the accessibility exactly as it is describe in the PropTypes, once we add interface for it, we can easily update this prop across all components.

PR #139, opened 2018-08-27
feat(Header): add props interface

mnajdova:
This PR introduces props interface for the Header component, as part of #117


src/components/Header/Header.tsx
```diff
@@ -3,6 +3,19 @@ import * as React from 'react'
 
 import { childrenExist, customPropTypes, UIComponent } from '../../lib'
 import HeaderDescription from './HeaderDescription'
+import { Extendable, ItemShorthand } from '../../../types/utils'
+import { ComponentVariablesInput, IComponentPartStylesInput } from '../../../types/theme'
+
+export interface IHeaderProps {
+  as?: any
+  children?: React.ReactNode
```
kuzhelov: it seems that we are lying here, as it could be the case that several nodes are provided as children. We shouldn't make any assumptions that only one child node will be provided - otherwise we could accidentally introduce subtle problems in processing logic that won;t be caught by the compiler

kuzhelov: `React.ReactNodeArray | React.ReactNode` seems to be a desired alternative here

kuzhelov (changes requested):
seems that we should reconsider types introduced for `children` prop - otherwise we are, essentially, making false promises to compiler

src/components/Header/Header.tsx
mnajdova: Good catch, thanks a lot! I will provide in the utils.d.ts `Children` type and reuse it for the children prop.

mnajdova:
I added `Children` type in our utils.d.ts and updated the other components containing children as well. Please review now.

kuzhelov:
have renamed `Children` to `ReactChildren` just for the sake of indicating that this is a type from React domain. No claims that this is the best name possible, but just to differentiate React-specific types from general-purpose ones

kuzhelov (approved):

PR #141, opened 2018-08-27
chore: Component performance tests

miroslavstastny:
# Measuring rendering performance

We would like to have some kind of performance tests (1) to answer questions like *"How will using stardust components influence my app's performance"* and (2) to avoid any performance regressions.

As a proof-of-concept, this PR introduces performance tests which measure render times for all doc site examples:
```sh
$ yarn test:perf --headless
...

========================== Performance tests results ===========================
components/Accordion 3 test(s), avg = 35.1 ms, max = 57.9 ms, min = 20.9 ms
components/Avatar 9 test(s), avg = 35.388777777777776 ms, max = 105.4 ms, min = 13.999 ms
components/Button 16 test(s), avg = 24.050125 ms, max = 56.9 ms, min = 16.2 ms
components/Chat 2 test(s), avg = 18.25 ms, max = 20.2 ms, min = 16.3 ms
components/Divider 11 test(s), avg = 17.345454545454544 ms, max = 30.6 ms, min = 13.2 ms
components/Grid 8 test(s), avg = 35.574875 ms, max = 47 ms, min = 19.3 ms
components/Header 8 test(s), avg = 15.949875 ms, max = 22.599 ms, min = 12.9 ms
components/Icon 6 test(s), avg = 24.48333333333333 ms, max = 36.2 ms, min = 18.101 ms
components/Image 4 test(s), avg = 14.949749999999998 ms, max = 23.7 ms, min = 10.9 ms
components/Input 5 test(s), avg = 24.48 ms, max = 30.5 ms, min = 18 ms
components/ItemLayout 10 test(s), 3 ERROR(S)!!!, avg = 11.209999999999999 ms, max = 18.7 ms, min = 14.599 ms
components/Label 9 test(s), avg = 18.155444444444445 ms, max = 21.3 ms, min = 14.299 ms
components/Layout 6 test(s), 1 ERROR(S)!!!, avg = 13.766666666666666 ms, max = 19.9 ms, min = 13.5 ms
components/List 20 test(s), 6 ERROR(S)!!!, avg = 19.544999999999995 ms, max = 34.7 ms, min = 21.3 ms
components/Menu 16 test(s), avg = 25.356375 ms, max = 32 ms, min = 22.001 ms
components/Radio 1 test(s), avg = 25.501 ms, max = 25.501 ms, min = 25.501 ms
components/Segment 2 test(s), avg = 11.45 ms, max = 11.8 ms, min = 11.1 ms
components/Text 11 test(s), avg = 13.145363636363635 ms, max = 17.2 ms, min = 11.4 ms
================================================================================

[14:51:54] Finished 'lighthouse' after 17 min
[14:51:54] Finished 'perf' after 17 min
```

## Current results

  | tests | avg [ms] | max [ms] | min [ms]
-- | -- | -- | -- | --
components/Accordion | 3 | 35.1 | 57.9 | 20.9
components/Avatar | 9 | 35.4 | 105.4 | 14.0
components/Button | 16 | 24.1 | 56.9 | 16.2
components/Chat | 2 | 18.3 | 20.2 | 16.3
components/Divider | 11 | 17.3 | 30.6 | 13.2
components/Grid | 8 | 35.6 | 47.0 | 19.3
components/Header | 8 | 15.9 | 22.6 | 12.9
components/Icon | 6 | 24.5 | 36.2 | 18.1
components/Image | 4 | 14.9 | 23.7 | 10.9
components/Input | 5 | 24.5 | 30.5 | 18.0
components/ItemLayout | 10 | 11.2 | 18.7 | 14.6
components/Label | 9 | 18.2 | 21.3 | 14.3
components/Layout | 6 | 13.8 | 19.9 | 13.5
components/List | 20 | 19.5 | 34.7 | 21.3
components/Menu | 16 | 25.4 | 32.0 | 22.0
components/Radio | 1 | 25.5 | 25.5 | 25.5
components/Segment | 2 | 11.5 | 11.8 | 11.1
components/Text | 11 | 13.1 | 17.2 | 11.4

## What to measure
React uses [User Timing API](https://developer.mozilla.org/en-US/docs/Web/API/User_Timing_API) to track component performance. Chrome Devtools displays these metrics in Performance:
image

This PR adds `stardust.perf.measureMount` into `ExternalExampleLayout`. This measures time from `componentWillMount()` to `componentDidMount()`.

## How to measure
[Lighthouse](https://developers.google.com/web/tools/lighthouse/) is an automated tool to run tests against a web page.
It can be used from command line. It runs inside chrome, and, among other things, measures it's performance. The final report includes our custom `stardust.perf.measureMount` which we can evaluate to measure performance of our components.

### What to test
The current implementation tests individual examples which mostly contain single instance of a component. Rendering times are ~20ms. It would be difficult to measure regressions on these - having an increase of 4ms is 20% in relative numbers which is huge, but in absolute numbers, just 4ms difference is almost no difference.
We can also create example page layouts which would contain hundreds of components, create/generate performance pages with 100 instances of a single component...

To minimise jitter, we can run the same tests several times, or use a baseline example (with html only) to normalise results.

### CI workflow
Currently the test only prints results to console. To be valuable it must be run as part of the CI. Lighthouse supports Travis, it [reportedly works in CircleCI](https://github.com/ebidel/lighthouse-ci/issues/16).

We can run it in Circle CI.

We can build or own solution running in Azure. This would give us full control of the pipeline and should give us lowest jitter. But obviously a lot of work - it is not only about running the tests, it's also about storing the results, comparing them from build to build, drawing fancy graphs... 
On the other hand - this would be just one work item away from our own screenshot tests.

We can also use 3rd party cloud solution.

### When to run tests in CI
With every commit? Before merge to master? Nightly on master only?
Running the tests for all examples takes more than 15 minutes on a powerful Mac. Running it with every commit could quickly become overwhelming.

### Development version of React
The test uses docsite which runs with development version of react. We should expect better performance results with release version, but different runs with development version should still be comparable.

## Resources
- https://github.com/paulirish/pwmetrics
- https://github.com/A-gambit/CSS-IN-JS-Benchmarks


sophieH29:
Lighthouse is a cool tool to use! However, as far as I remember, they set default throttling options for tests (like fast 3g). Curious, should we test more speed options?

layershifter:
I refactored this brunch to use Profiler component from https://github.com/bvaughn/rfcs/blob/profiler/text/0000-profiler.md. This API is used by React Dev Tools and provides user metrics. I also used Puppeteer instead of Lighthouse.

---

Example output:
image

layershifter:
Another improvement is a very fast run:
```
[18:13:38] Finished 'perf' after 35 s
Done in 43.87s.
```

package.json
```diff
@@ -21,7 +21,7 @@
     "deploy:docs": "gulp deploy:docs",
     "lint": "tslint -t stylish -p .",
     "lint:fix": "yarn lint --fix",
-    "perf": "cross-env PERF=true gulp perf",
+    "perf": "cross-env PERF=true gulp perf --times=50",
```
kuzhelov, line 24: 👍

perf/src/index.tsx
```diff
@@ -1,16 +1,58 @@
...
+const performanceExamplesContext = require.context('docs/src/examples/', true, /.perf.tsx$/)
+
+// Heads up!
+// We want to randomize examples to avoid any notable issues with always first example
+const performanceExampleNames = _.shuffle(performanceExamplesContext.keys())
+const performanceMeasures = {}
 
-const renderCycle = Component => {
-  ReactDOM.render(<Component />, mountNode)
+const renderCycle = (exampleName: string, Component: React.ComponentType, order: number) => {
```
kuzhelov: would suggest to rename `order` to something that would be a bit more descriptive (`exampleIndex` or `exampleKey`)

perf/src/index.tsx
```diff
@@ -1,16 +1,58 @@
...
+const Profiler = (React as any).unstable_Profiler
+
 const mountNode = document.querySelector('#root')
-const performanceExamples = require.context('docs/src/examples/', true, /.perf.tsx$/)
+const performanceExamplesContext = require.context('docs/src/examples/', true, /.perf.tsx$/)
+
+// Heads up!
+// We want to randomize examples to avoid any notable issues with always first example
+const performanceExampleNames = _.shuffle(performanceExamplesContext.keys())
+const performanceMeasures = {}
```
kuzhelov: lets avoid global objects, as there is nothing wrong (from client's point of view) to run `window.runMeasures()` twice (and even during performance tests this is the thing that happens when several rounds of tests run)

While I do understand that results will be overwritten, however, it will happen only during fully successful run cycle - but in case if some error will happen, it will be helpful to have `performanceMeasures` object with the state that would exactly correspond to the latest run's results

package.json
```diff
@@ -40,6 +40,7 @@
     "test:watch": "gulp test:watch",
     "test:vulns": "gulp test:vulns",
     "test:visual": "gulp screener",
+    "test:perf": "gulp perf",
```
kuzhelov: do we need two tasks for it? cannot immediately see _semantical_ difference between them

build/gulp/plugins/gulp-webpack.ts
```diff
@@ -0,0 +1,28 @@
+import * as webpack from 'webpack'
+import config from '../../../config'
+
+const { log, PluginError } = require('gulp-load-plugins')().util
+
+const webpackPlugin = (webpackConfig, cb) => {
```
kuzhelov, line 6: 👍

build/gulp/tasks/perf.ts
```diff
@@ -0,0 +1,117 @@
...
+  const values = _.sortBy(measures, key)
+
+  const lowMiddle = Math.floor((values.length - 1) / 2)
+  const highMiddle = Math.ceil((values.length - 1) / 2)
+
+  return (values[lowMiddle][key] + values[highMiddle][key]) / 2
+}
+
+const reduceMeasures = (measures: ProfilerMeasure[], key: string) => {
+  let min = measures[0][key]
```
kuzhelov, line 33: may suggest to process corner case of measures `count === 0` and return early in this case

build/gulp/tasks/perf.ts
```diff
@@ -0,0 +1,117 @@
...
+  fs.writeFileSync(
+    paths.perfDist('result.json'),
+    JSON.stringify(normalizeMeasures(measures), null, 2),
+  )
+})
+
+task('perf:serve', cb => {
+  server = express()
+    .use(express.static(paths.perfDist()))
+    .listen(config.server_port, config.server_host, () => {
```
kuzhelov: lets use port value that is different from the docs one

build/gulp/tasks/perf.ts
```diff
@@ -0,0 +1,117 @@
...
+    const measuresFromStep = await page.evaluate(() => window.runMeasures())
+    await page.close()
+
+    measures.push(measuresFromStep)
+  }
+
+  await browser.close()
+
+  fs.writeFileSync(
+    paths.perfDist('result.json'),
```
kuzhelov: lets create some console log entry about that or, much better, display results to the console somehow. Currently it is completely non obvious to the client where these results could be found :)

build/gulp/tasks/perf.ts
```diff
@@ -0,0 +1,117 @@
...
+  const times = argv.times || 10
+
+  const browser = await puppeteer.launch()
+
+  for (let i = 0; i < times; i++) {
+    const page = await browser.newPage()
+    await page.goto(`http://${config.server_host}:${config.server_port}`)
+
+    const measuresFromStep = await page.evaluate(() => window.runMeasures())
+    await page.close()
```
kuzhelov: would suggest to put it into `finally` block and, thus, guarantee that `page` process resources will be reclaimed no matter what

build/gulp/tasks/perf.ts
```diff
@@ -0,0 +1,117 @@
...
+  rimraf(paths.perfDist(), cb)
+})
+
+task('perf:build', cb => {
+  webpackPlugin(require('../../../build/webpack.config.perf').default, cb)
+})
+
+task('perf:run', async () => {
+  const measures: ProfilerMeasureCycle[] = []
+  const times = argv.times || 10
```
kuzhelov: may we introduce constant for `10`

kuzhelov (commented):
left couple of comments, those are most about some cleanup stuff. Do like to approach and the structure introduced to gulp tasks (special thanks for `gulp-webpack` refactoring;)

package.json
layershifter: Cruft, removed 👍

perf/src/index.tsx
layershifter: Done 👍

perf/src/index.tsx
layershifter: Refactored, `window.runMeasures()` now is a Promise that returns results 👍

build/gulp/tasks/perf.ts
layershifter: Add `perf_port` to config 👍

build/gulp/tasks/perf.ts
layershifter: Added log entry:
```
[16:31:18] Results are written to "...PATH...\react\perf\dist\result.json"
```

build/gulp/tasks/perf.ts
layershifter: We don't handling any errors there at all, not sure that we should do this now 💭

build/gulp/tasks/perf.ts
layershifter: Add a condition there, an error will be thrown in this case 👍

build/gulp/tasks/perf.ts
layershifter: Added `DEFAULT_RUN_TIMES` 👍

build/gulp/tasks/perf.ts
kuzhelov: would still suggest that, as those are a bit different types of problems
- if those that we are not handling are about some correctness problems, they won't have global effect to the platform process is running on
- here, in contrast, this global effect might take place in case if OS resources are consumed without being reclaimed on failure

Not saying that we should introduce error handling measures everywhere, but here see this as a very important case we need to address now

build/gulp/tasks/perf.ts
```diff
@@ -0,0 +1,119 @@
...
+    await page.close()
+
+    measures.push(measuresFromStep)
+  }
+
+  await browser.close()
+  const resultsFile = paths.perfDist('result.json')
+
+  fs.writeFileSync(resultsFile, JSON.stringify(normalizeMeasures(measures), null, 2))
+  log(colors.green('Results are written to "%s"'), resultsFile)
```
kuzhelov, line 109: 👍

config.ts
```diff
@@ -57,6 +57,7 @@ const config = {
   // ----------------------------------
   server_host: 'localhost',
   server_port: process.env.PORT || 8080,
+  perf_port: process.env.PERF_PORT || 8081,
```
kuzhelov, line 60: 👍 👍

perf/src/index.tsx
```diff
@@ -1,16 +1,80 @@
...
+
 const mountNode = document.querySelector('#root')
-const performanceExamples = require.context('docs/src/examples/', true, /.perf.tsx$/)
+const performanceExamplesContext = require.context('docs/src/examples/', true, /.perf.tsx$/)
+
+// Heads up!
+// We want to randomize examples to avoid any notable issues with always first example
+const performanceExampleNames: string[] = _.shuffle(performanceExamplesContext.keys())
+
+const asyncRender = (element: React.ReactElement<any>, container: Element) =>
```
kuzhelov: although `container` is absolutely correct, would suggest to rename it to `mountNode` - to better reflect the context it is used for. It make it possible to get semantics behind each of the args much easier

layershifter: Agree, I left it because `container` comes from typings

kuzhelov (approved):
looks very good to me 💯 one thing that I would like to be fixed, though, is a potential case that may lead to uncontrolled OS resources consumption: https://github.com/stardust-ui/react/pull/141/files#r249771557

build/gulp/tasks/perf.ts
layershifter: Added error handling, please check 👍

build/gulp/tasks/perf.ts
```diff
@@ -83,25 +83,28 @@ task('perf:build', cb => {
...
-
-    try {
+    for (let i = 0; i < times; i++) {
+      const page = await browser.newPage()
       await page.goto(`http://${config.server_host}:${config.perf_port}`)
 
       const measuresFromStep = await page.evaluate(() => window.runMeasures())
       measures.push(measuresFromStep)
-    } catch (e) {
-      throw e
```
kuzhelov: as we are just rethrowing here, `try - finally` should be enough. The same applies to the `browser` as well.

```js
try { .. } finally { await page.close() }
```
In future we might think to encapsulate it into abstraction like
```js
withResource(browser, async () => {
   ... logic that might throw ..
}) // resource is guaranteed to be released here
```

PR #142, opened 2018-08-27
feat(ItemLayout): add props interface and improve props typings

mnajdova:
This PR introduces props interface for the ItemLayout component, as part of #117


levithomason (approved):
👍

PR #144, opened 2018-08-27
feat(portal): simple base implementation

bmdalex:
# Portal

This PR will introduce **Portal** component, a component that allows you to render children outside their parent (for now all portals are rendered at the end of the `body` DOM element); a **Portal** comes with the following props: 
- `onMount?: (props: IPortalProps) => void`: Called when the portal is mounted on the DOM
- `onUnmount?: (props: IPortalProps) => void`: Called when the portal is unmounted from the DOM
- `open?: boolean`: Controls whether or not the portal is displayed
- `trigger?: JSX.Element`: Element to be rendered in-place where the portal is defined
- `triggerRef?: (node: HTMLElement) => void`: Called with a ref to the trigger node

and the following helper components used by **Portal**:
- **PortalInner** with props:
  - `onMount?: (props: IPortalProps) => void`: Called when the portal is mounted on the DOM
  - `onUnmount?: (props: IPortalProps) => void`: Called when the portal is unmounted from the DOM

- **Ref** with props:
  - `innerRef?: (ref: HTMLElement) => void`: Called with `componentDidMount`

### TODO

- [ ] Conformance test
- [ ] Stardust base theme
- [x] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [x] Confirm RTL usage
- [x] Update the CHANGELOG.md

# Proposal

## Basic portal

screen shot 2018-08-27 at 17 27 10

Renders a portal that is:
- opened when clicking on the `trigger`, a component sent to the portal as a prop;
- closed when clicking away or on the `trigger` (since it acts like a toggle)

```jsx
<Portal
  content={<div>This is a basic portal</div>}
  trigger={<Button content={'Toggle portal'} onClick={this.handleClick} />}
/>
```
generates
```html
<button>Toggle portal</button>
```
and at the end of `body` DOM element:
```html
<body>
  ...
  <div>This is a basic portal</div>
</body>
```

## Controlled portal

screen shot 2018-08-27 at 17 26 25

Renders a portal that is rendered or not depending on the value of `open` property

```jsx
<Portal
  content={<div>This is a controlled portal</div>}
  open={true}
/>
```
generates
```html
<body>
  ...
  <div>This is a controlled portal</div>
</body>
```
<button>Toggle portal</button>
at the end of `body` DOM element.

### DIFF
--- a/docs/src/examples/components/Portal/Types/PortalExample.shorthand.tsx
+++ b/docs/src/examples/components/Portal/Types/PortalExample.shorthand.tsx
@@ -1,62 +1,51 @@
 import React from 'react'
 import { Button, Divider, Header, Label, Portal } from '@stardust-ui/react'
 
-class PortalExamplePortal extends React.Component<any, any> {
-  public state = {
-    log: [],
-    logCount: 0,
-  }
+class PortalExamplePortal extends React.Component {
+  state = { log: [], logCount: 0 }
 
-  public render() {
-    const { log, logCount } = this.state
-
-    const controls = (
-      <div>
-        <Button size="small" onClick={this.clearLog} content="Clear" />
-        <span>
-          Event Log <Label circular>{logCount}</Label>
-        </span>
-        <pre>{log.map((e, i) => <div key={i}>{e}</div>)}</pre>
-      </div>
-    )
-
-    const portalContent = (
-      <div
-        style={{
-          background: '#ddd',
-          position: 'fixed',
-          left: '40%',
-          top: '45%',
-          zIndex: 1000,
-        }}
-      >
-        <Header>This is a basic portal</Header>
-        <p>Portals have tons of great callback functions to hook into.</p>
-        <p>To close, simply click the close button or click away</p>
-      </div>
-    )
-
-    return (
-      <div>
-        {
-          <Portal
-            content={portalContent}
-            trigger={<Button content={'Toggle portal'} onClick={this.handleClick} />}
-          />
-        }
-        <Divider />
-        {controls}
-      </div>
-    )
-  }
-
-  private handleClick = () =>
+  handleClick = () =>
     this.setState({
       log: [`${new Date().toLocaleTimeString()}: handleClick`, ...this.state.log].slice(0, 20),
       logCount: this.state.logCount + 1,
     })
 
-  private clearLog = () => this.setState({ log: [], logCount: 0 })
+  clearLog = () => this.setState({ log: [], logCount: 0 })
+
+  render() {
+    const { log, logCount } = this.state
+
+    return (
+      <div>
+        <Portal
+          content={
+            <div
+              style={{
+                background: '#ddd',
+                position: 'fixed',
+                left: '40%',
+                top: '45%',
+                zIndex: 1000,
+              }}
+            >
+              <Header>This is a basic portal</Header>
+              <p>Portals have tons of great callback functions to hook into.</p>
+              <p>To close, simply click the close button or click away</p>
+            </div>
+          }
+          trigger={<Button content={'Toggle portal'} onClick={this.handleClick} />}
+        />
+        <Divider />
+        <div>
+          <Button size="small" onClick={this.clearLog} content="Clear" />
+          <span>
+            Event Log <Label circular>{logCount}</Label>
+          </span>
+          <pre>{log.map((e, i) => <div key={i}>{e}</div>)}</pre>
+        </div>
+      </div>
+    )
+  }
 }
 
 export default PortalExamplePortal
--- a/docs/src/examples/components/Portal/Types/PortalExample.shorthand.tsx
+++ b/docs/src/examples/components/Portal/Types/PortalExample.shorthand.tsx
@@ -1,0 +1,62 @@
+import React from 'react'
+import { Button, Divider, Header, Label, Portal } from '@stardust-ui/react'
+
+class PortalExamplePortal extends React.Component<any, any> {
+  public state = {
+    log: [],
+    logCount: 0,
+  }
+
+  public render() {
+    const { log, logCount } = this.state
+
+    const controls = (
+      <div>
+        <Button size="small" onClick={this.clearLog} content="Clear" />
+        <span>
+          Event Log <Label circular>{logCount}</Label>
+        </span>
+        <pre>{log.map((e, i) => <div key={i}>{e}</div>)}</pre>
+      </div>
+    )
+
+    const portalContent = (
+      <div
+        style={{
+          background: '#ddd',
+          position: 'fixed',
+          left: '40%',
+          top: '45%',
+          zIndex: 1000,
+        }}
+      >
+        <Header>This is a basic portal</Header>
+        <p>Portals have tons of great callback functions to hook into.</p>
+        <p>To close, simply click the close button or click away</p>
+      </div>
+    )
+
+    return (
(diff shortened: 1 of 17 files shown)


docs/src/examples/components/Portal/Types/PortalExample.shorthand.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react'
+import { Button, Divider, Header, Label, Portal } from '@stardust-ui/react'
+
+class PortalExamplePortal extends React.Component<any, any> {
+  public state = {
+    log: [],
+    logCount: 0,
+  }
+
+  public render() {
```
levithomason: I don't think we should be typing public/private like this in React components.  Even though `method` and other custom methods are publicly accessible, it is not idiomatic React to actually access them.  You should never reach into a React component and call a method.

In other words, if we did type them, it would be more proper to consider all methods as `private` but I believe that would cause some other issues.

docs/src/examples/components/Portal/Types/PortalExampleControlled.tsx
```diff
@@ -0,0 +1,65 @@
+import * as React from 'react'
```
levithomason: No typescript in examples please.

src/components/Portal/PortalInner.tsx
```diff
@@ -0,0 +1,48 @@
+import * as PropTypes from 'prop-types'
+import { invoke } from 'lodash'
```
levithomason: Let's keep convention consistent in the repo, it makes tooling and programmatic updates easier to manage.

There is a `babel-plugin-lodash` which we use at SUIR to rewrite the import statements to cherry pick only used methods.  We should either implement this plugin, or enforce cherry picking lodash methods, but, we should not mix both patterns without a guarantee of either one.

src/components/Portal/PortalInner.tsx
```diff
@@ -0,0 +1,48 @@
...
+  public componentDidMount() {
+    invoke(this.props, 'onMount', { ...this.props })
+  }
+
+  public componentWillUnmount() {
+    invoke(this.props, 'onUnmount', { ...this.props })
+  }
+
+  public render() {
+    return createPortal(this.props.children, document.body)
```
levithomason: Accessing `document` will fail in server side rendering.  This should probably be a `defaultProp` instead that is safe guarded by a browser check first:

```jsx
import { isBrowser } from '../../lib'

static defaultProps = {
  context: isBrowser() ? document.body : null
}

render() {
  const { children, context } = this.props

  return createPortal(children, context)
}
```

src/components/Portal/PortalInner.tsx
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Called when the portal is unmounted from the DOM
+     *
+     * @param {object} data - All props.
+     */
+    onUnmount: PropTypes.func,
+  }
+
+  public componentDidMount() {
+    invoke(this.props, 'onMount', { ...this.props })
```
levithomason: Since we're not adding / changing any props here, not sure that a shallow clone is required.  We can save on the cycles by just passing `this.props`.

docs/src/examples/components/Portal/Types/PortalExampleControlled.shorthand.tsx
```diff
@@ -0,0 +1,65 @@
+import * as React from 'react'
+import { Button, Divider, Header, Label, Portal } from '@stardust-ui/react'
+
+class PortalExampleControlled extends React.Component<any, any> {
+  public state = {
+    log: [],
+    logCount: 0,
+    open: false,
+  }
```
levithomason: In effort to minimize lines, this object can be single line:

```jsx
state = { log: [], logCount: 0, open: false }
```

docs/src/examples/components/Portal/Types/PortalExampleControlled.shorthand.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  public render() {
+    const { log, logCount, open } = this.state
+    const content = open ? 'Close Portal' : 'Open Portal'
+
+    const controls = (
```
levithomason: `controls` and `portalContent` can also be inlined into the render method to save some lines.

levithomason (changes requested):
Thanks for porting this.  See individual comments for change request details.

docs/src/examples/components/Portal/Types/PortalExampleControlled.tsx
smykhailov: I'd rather allow both, the similar way as it is done on the Android Developer docs
image

docs/src/examples/components/Portal/Types/PortalExample.shorthand.tsx
bmdalex: I don't agree with this and here are my reasons:

1. The current pattern I see all over the project is to not specify any modifier for class members which means all members are public (class members are public by default in typescript).
This is wrong and is one of the main reasons I'd recommend enforcing modifiers; developers often forget to specify a method is private since no modifier is required and this makes it private.

2. `render` is a public method as it comes from `React.Component` or `UIComponent`, the classes we usually extend.

3. Specifying modifiers makes it clear for developers coming from other environments where `private` is the default class member modifier (such as `C#`).

**We had this discussion before and since it's occurring again I can create an RFC for enforcing modifiers if you think it's valuable.**

@levithomason @kuzhelov @miroslavstastny @mnajdova what do you guys think?

docs/src/examples/components/Portal/Types/PortalExampleControlled.tsx
bmdalex: then these should be `jsx` files, not `tsx`, right? :) I can do the change for now but it's confusing to see these `tslint` errors..

src/components/Portal/PortalInner.tsx
bmdalex: ok

src/components/Portal/PortalInner.tsx
bmdalex: I made the change but I don't really understand it. Server side rendering? Where do we do that?

src/components/Portal/PortalInner.tsx
bmdalex: made the change as suggested

docs/src/examples/components/Portal/Types/PortalExample.shorthand.tsx
bmdalex: since this is an example I'll remove the modifiers so that we make it `jsx`, but the points stand for actual typescript code

docs/src/examples/components/Portal/Types/PortalExampleControlled.shorthand.tsx
bmdalex: done

docs/src/examples/components/Portal/Types/PortalExampleControlled.shorthand.tsx
bmdalex: I think my approach makes it more clear how to actually use the components since the `controls` and `portalContent` are lengthy and users can get confused; that's why I did it like this... Anyway, I made the change as suggested

levithomason (approved):

PR #135, opened 2018-08-24
fix(Button): icon colors and layout

kuzhelov:
## TODO

- [x] Address issues screener has reported about (site variables were not applied for some reason?)
- [x] Update the CHANGELOG.md

--------------

Introduces changes that are aimed to solve the following problems with `Button`.

# Colors

## No color for default button
This customization parameter was absent for default button - although there was `backgroundColor` variable.

```jsx
<Button icon='chess rook' variables={{ color: 'red' }} />
```

**_was (irregardless of value provided)_**

image

**_now_**

image

## Icon's color of primary/secondary Button
Was not inherited - was hardcoded to be either black or white.

```jsx
<Button primary ... variables={{ typePrimaryColor: 'red' }} />
```

**_was_**

image

**_now_**

image

## Styles vanished in disabled state
All button in disabled state were displayed as grey - styling system hasn't honored any colors of enabled look of the button.

```jsx
<Button disabled primary ... variables={{ typePrimaryColor: 'red' }} />
```

**_was_**

image

**_now_**

image

# Layout
Layout styles were generally simplified, problems for the following cases were addressed:

## iconPosition='before'

```jsx
<Button content='...some long text...' icon='chess rook' iconPosition='before' />
```

**_was_**

image

**_now_**

image

## iconPosition='after'

```jsx
<Button content='...some long text...' icon='chess rook' iconPosition='after' />
```

**_was_**

image

**_now_**

image

PR #145, opened 2018-08-28
docs: add quick start page

levithomason:
This PR adds a little info to the Introduction page and links the Quick Start guide (many more guides on another branch will be posted tomorrow).

image

image

### DIFF
--- a/build/tsconfig.common.json
+++ b/build/tsconfig.common.json
@@ -6,6 +6,7 @@
     "baseUrl": "../",
     "paths": {
       "@stardust-ui/react": ["src"],
+      "@stardust-ui/react/themes/teams": ["src/themes/teams"],
       "*": ["*", "types/*"]
     },
     "types": [
--- a/docs/src/components/CodeSnippet.tsx
+++ b/docs/src/components/CodeSnippet.tsx
@@ -1,54 +1,50 @@
 import * as React from 'react'
 import Editor, { EDITOR_BACKGROUND_COLOR } from './Editor'
 
-export interface ICodeSnippetProps {
+export interface CodeSnippetProps {
   label?: string
   mode?: 'jsx' | 'html' | 'sh'
   value: string
   style?: React.CSSProperties
 }
 
-const CodeSnippet = ({ label = '', value, mode, style, ...rest }: ICodeSnippetProps) => {
-  const editorMode: any = mode || label.replace(/.*\./, '') || 'jsx'
-
-  return (
+const CodeSnippet = ({ label = '', value, mode, style, ...rest }: CodeSnippetProps) => (
+  <div
+    style={{
+      position: 'relative',
+      padding: '1rem',
+      marginBottom: '2rem',
+      background: EDITOR_BACKGROUND_COLOR,
+      ...style,
+    }}
+  >
     <div
       style={{
-        position: 'relative',
-        padding: '1rem',
-        marginBottom: '2rem',
-        background: EDITOR_BACKGROUND_COLOR,
-        ...style,
+        position: 'absolute',
+        padding: '0.2rem 0.35rem',
+        top: '1rem',
+        right: '1rem',
+        lineHeight: 1,
+        color: '#899',
+        fontFamily: 'monospace',
+        fontSize: '0.8rem',
+        border: '1px solid #566',
+        zIndex: 100,
       }}
     >
-      <div
-        style={{
-          position: 'absolute',
-          padding: '0.2rem 0.35rem',
-          top: '1rem',
-          right: '1rem',
-          lineHeight: 1,
-          color: '#899',
-          fontFamily: 'monospace',
-          fontSize: '0.8rem',
-          border: '1px solid #566',
-          zIndex: 100,
-        }}
-      >
-        {label || editorMode}
-      </div>
-      <Editor
-        id={btoa(value)}
-        highlightActiveLine={false}
-        highlightGutterLine={false}
-        mode={editorMode}
-        readOnly
-        showGutter={false}
-        showCursor={false}
-        value={editorMode === 'sh' ? value.replace(/^/g, '$  ') : value}
-        {...rest}
-      />
+      {label || mode}
     </div>
-  )
-}
+    <Editor
+      id={btoa(value)}
+      highlightActiveLine={false}
+      highlightGutterLine={false}
+      mode={mode}
+      readOnly
+      showGutter={false}
+      showCursor={false}
+      value={mode === 'sh' ? value.replace(/^/g, '$  ') : value}
+      {...rest}
+    />
+  </div>
+)
 export default CodeSnippet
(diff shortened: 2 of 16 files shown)


src/themes/teams/components/Icon/fontAwesomeIconStyles.ts
```diff
@@ -1,4 +1,5 @@
 const fontAwesomeIcons = new Map([
+  ['arrow right', 'f061'],
```
levithomason, line 2: Icon used in quick start example.

src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -86,6 +88,7 @@ const buttonStyles: IComponentPartStylesInput = {
               backgroundColor: typeSecondaryBackgroundColor,
               borderColor: typeSecondaryBorderColor,
               ':hover': {
+                color: typeSecondaryColor,
```
levithomason, line 91: Fixed button styles when rendering as an `a` tag.

src/index.ts
```diff
@@ -1,3 +1,6 @@
+import * as themes from './themes'
+export { themes }
```
levithomason, line 2: Export themes from our package.

docs/src/components/Editor/Editor.tsx
```diff
@@ -45,16 +46,19 @@ const semanticUIReactCompleter = {
 languageTools.addCompleter(semanticUIReactCompleter)
 
 export interface IEditorProps extends AceEditorProps {
-  id?: string
+  id: string
```
levithomason, line 49: This is a DOM id and is required by AceEditor to work correctly.

smykhailov:
@levithomason, can we also add an explanation, that Stardust UI consists of two parts:

- Component Library
- The framework which allows creating a new components 

This part is a little bit confusing for consumers of the library. I've seen that consumers create new components extended from `React.Component` instead of extending them from `UIComponent` which allows them to leverage styling, accessibility, and theming provided by Framework.

src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -45,6 +45,7 @@ const buttonStyles: IComponentPartStylesInput = {
         (iconPosition
           ? {
               display: 'inline-flex',
+              alignItems: 'center',
```
kuzhelov, line 48: as discussed, there is a corresponding PR #135 where bunch of related `Button` problems are addressed. 

As a next work item related to that - the PR with conformance tests that will ensure that shorthand prop names correspond to component's anatomy parts, so that it would be intuitive for client to decide which name should be used for styling specific component's part:

```jsx
<Button icon={...} styles={ icon: () => ... } />
```

build/tsconfig.common.json
```diff
@@ -6,6 +6,7 @@
     "baseUrl": "../",
     "paths": {
       "@stardust-ui/react": ["src"],
+      "@stardust-ui/react/themes/teams": ["src/themes/teams"],
```
kuzhelov, line 9: 👍

docs/src/components/CodeSnippet.tsx
```diff
@@ -0,0 +1,54 @@
+import * as React from 'react'
+import Editor, { EDITOR_BACKGROUND_COLOR } from './Editor'
+
+export interface ICodeSnippetProps {
```
kuzhelov: here is a point for further thought: to me it seems a bit unreasonable to use 'I' prefix for interface types - this leads to the fiolloiwng situation where `interface` types and alias `type` types - both with exactly the same semantics, - are mixed:

```jsx
```

Here we are importing two type interfaces (in general sense) for styles and variables - but for some reason one uses `I` prefix :) What I would suggest is to omit `I` prefix altogether for all types - this will make them consistent. Actually, the same approach React has taken for its types - quite a few of them are TS interfaces, but all the exposed names are consistent to the client as there are no `I` prefixes

docs/src/components/CodeSnippet.tsx
```diff
@@ -0,0 +1,54 @@
...
+  label?: string
+  mode?: 'jsx' | 'html' | 'sh'
+  value: string
+  style?: React.CSSProperties
+}
+
+const CodeSnippet = ({ label = '', value, mode, style, ...rest }: ICodeSnippetProps) => {
+  const editorMode: any = mode || label.replace(/.*\./, '') || 'jsx'
```
kuzhelov: it seems that we might obtain not supported editor mode's value here (like `tsx` from `label`). Is it fine to pass this value to AceEditor then, or we should opt to some default (maybe, 'text') mode, so that only explicitly supported modes will be rendered (and, thus, guaranteed to be rendered properly)?

kuzhelov (approved):

markpete:
The command should be 'yarn add @stardust-ui/react'  yarn install doesn't work.

docs/src/components/CodeSnippet.tsx
levithomason: I have no opinion here other than it should follow the best practice.  It is nice to know when you can use `extends` (interfaces but not types).  However, it appears the [official recommendation](https://github.com/Microsoft/TypeScript-Handbook/issues/121) is to _not_ use prefixes.  Will update.

levithomason: I went to update this and it appears we have 44 interfaces in 24 files prefixed with "I" and only 4 interfaces in 3 files that aren't.  I think we should make a separate PR that addresses this and updates the team about the new standard.

I will still update my usage here to not include the prefix.

docs/src/components/CodeSnippet.tsx
levithomason: True.  This is partial left over from a previous iteration where `label` was named `filename`.  However, the point still stands.  I'm going to revert to using the less magical `mode` prop just as the Editor does.  This will ensure we can explicitly set the mode/highlighting separate from the label in the code snippet.

markpete:
The library @stardust-ui/react/themes/teams isn't installed by installing the core @stardust-ui/react and trying to add that module directly doesn't work.  Where does one get that module from?  Without it, the base example doesn't work.

docs/src/components/CodeSnippet.tsx
levithomason: You also asked if it is OK to pass an unsupported value, no.  It will break the highlighting and also throw.  Also, `tsx` is not a supported mode.

levithomason:
>...command should be 'yarn add @stardust-ui/react'
Fixed!

>...@stardust-ui/react/themes/teams isn't installed by installing the core @stardust-ui/react 

It is installed, however, it is not exported at that path until this PR is released.  You can find it in `v0.3.0` at `@stardust-ui/dist/react/es/themes/teams` where `es` is your module preference.  There is also `commonjs`.  Have a look in `node_modules/@stardust-ui` to confirm the path.

levithomason:
>@levithomason, can we also add an explanation, that Stardust UI consists of two parts:
>
>- Component Library
>- The framework which allows creating a new components

@smykhailov Agreed but I think we need to flesh this out still.  We haven't yet put forward a proposal for how consumers will do this.  Let's save that for another PR.

markpete:
Non-SVG icons still don't seem to work with these instructions.  They come in as boxes.  Is there a need to include a font such as FontAwesome or something?

PR #146, opened 2018-08-28
fix(typings): introduce specific types instead of 'any'

kuzhelov:
Small set of changes that introduce specific types to props that initially were of `any` type.

### DIFF
--- a/src/components/Chat/Chat.tsx
+++ b/src/components/Chat/Chat.tsx
@@ -5,13 +5,13 @@ import * as React from 'react'
 import { childrenExist, customPropTypes, UIComponent } from '../../lib'
 import ChatMessage from './ChatMessage'
 import { ComponentVariablesInput, IComponentPartStylesInput } from '../../../types/theme'
-import { Extendable, ReactChildren } from '../../../types/utils'
+import { Extendable, ReactChildren, ItemShorthand } from '../../../types/utils'
 
 export interface IChatProps {
   as?: any
   className?: string
   children?: ReactChildren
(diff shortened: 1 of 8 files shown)


src/components/Chat/Chat.tsx
```diff
@@ -5,13 +5,13 @@ import * as React from 'react'
...
 import { ComponentVariablesInput, IComponentPartStylesInput } from '../../../types/theme'
-import { Extendable, ReactChildren } from '../../../types/utils'
+import { Extendable, ReactChildren, ItemShorthand } from '../../../types/utils'
 
 export interface IChatProps {
   as?: any
   className?: string
   children?: ReactChildren
-  messages?: any[]
+  messages?: ItemShorthand[]
```
layershifter, line 14: Can we transform `ItemShorthand` to a generic type?

`ItemShorthand => ItemShorthand<IChatMessageProps>`
`ItemShorthand[] => CollectionShorthand<IChatMessageProps>`

kuzhelov: absolutely agree that this is rather a good move - although it doesn't add much in terms of type checking (due to React.ReactNode is very permissive), it adds on the side of expressivity. However, would rather like to introduce this move later, as, for example, for Accordion component it would require some additional thought: whether generic one should be used for content (and in that case we will end up with two types instead of one - adds complexity), use `Shorthand<any>`, anything else. Let me, please, defer solution for this problem - as it seems that type checks are strict enough even now

levithomason (approved):

PR #147, opened 2018-08-28
feat(Image): add props interface

kuzhelov:
This PR introduces props interface for the Image component, as part of #117

### DIFF
--- a/src/components/Image/Image.tsx
+++ b/src/components/Image/Image.tsx
@@ -5,6 +5,21 @@ import { customPropTypes, UIComponent } from '../../lib'
 import { ImageBehavior } from '../../lib/accessibility'
 import { Accessibility } from '../../lib/accessibility/interfaces'
 
+import { ComponentVariablesInput, IComponentPartStylesInput } from '../../../types/theme'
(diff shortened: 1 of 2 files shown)


src/components/Image/Image.tsx
```diff
@@ -5,6 +5,21 @@ import { customPropTypes, UIComponent } from '../../lib'
 import { ImageBehavior } from '../../lib/accessibility'
 import { Accessibility } from '../../lib/accessibility/interfaces'
 
+import { ComponentVariablesInput, IComponentPartStylesInput } from '../../../types/theme'
+import { Extendable, ReactChildren } from '../../../types/utils'
+
+export interface IImageProps {
+  accessibility?: Accessibility
+  as?: any
```
mnajdova, line 13: Oops, I was adding the accessibility prop as object | Function, as I didn't know that there is prop for this. We may need to update this in the other components

mnajdova (approved):

miroslavstastny (approved):

src/components/Image/Image.tsx
kuzhelov: sure, will do

PR #148, opened 2018-08-28
feat(Label): add props interface

kuzhelov:
This PR introduces props interface for the Label component, as part of #117

### DIFF
--- a/types/utils.d.ts
+++ b/types/utils.d.ts
@@ -12,4 +12,4 @@ export type Extendable<T> = T & {
 
 export type ItemShorthand = React.ReactNode | object | (React.ReactNode | object)[]
 export type ReactChildren = React.ReactNodeArray | React.ReactNode
-export type SyntheticEventHandler<TProps> = (event: React.SyntheticEvent, data: TProps) => void
+export type ComponentEventHandler<TProps> = (event: React.SyntheticEvent, data: TProps) => void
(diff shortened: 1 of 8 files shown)


types/utils.d.ts
```diff
@@ -12,3 +12,4 @@ export type Extendable<T> = T & {
 
 export type ItemShorthand = React.ReactNode | object | (React.ReactNode | object)[]
 export type ReactChildren = React.ReactNodeArray | React.ReactNode
+export type SyntheticEventHandler<TProps> = (event: React.SyntheticEvent, data: TProps) => void
```
levithomason: Propose we call this something like `StardustEventHandler`.  The signature here is unique to Stardust and does not apply all synthetic event handlers.

levithomason (approved):
One suggestion, see comment.

types/utils.d.ts
kuzhelov: agree with the reasons provided, but would rather suggest to use `ComponentEventHandler` then. The problem I see with using `StardustEventHandler` is that with the approach of prefixing library-specific types with 'Stardust' we would potentially need to introduce this prefix to other types too (`ItemShorthand`, as an example?). To me it seems reasonable to assume that all types that are introduced as part of the library are `Stardust` ones, and there is no additional need to emphasize this :) 

Let me, please, change it to `ComponentEventHandler` for now and make a promise to rename it in case it won't work well for us

PR #150, opened 2018-08-28
feat(popup): base implementation

bmdalex:
# Popup

This PR introduces the **Popup**, a component that displays additional information on top of a page at the end of the `body` DOM element. It comes with the following props:

- `basic?: boolean`: when `true`, it introduces basic CSS styling for the popup (default `false`)
- `align?: 'top' | 'bottom' | 'start' | 'end' | 'center'`
- `position?: 'above' | 'below' | 'before' | 'after'`: the position of the popup
- `trigger: JSX.Element`: the `React` component that controls the popup; for now, clicking it toggles the visibility of the popup

### TODO

- [x] Conformance test
- [ ] Stardust base theme
- [x] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [ ] Confirm RTL usage
- [x] Update the CHANGELOG.md

# API Proposal

## Default

screen shot 2018-08-28 at 18 41 50

A default popup requires the `trigger` prop and renders a popup at the end of the `body` DOM element

```jsx
<Popup trigger={<Button content="Add" />} content="Add users to your feed" />
```
generates 
```html
<button>Add</button>
```
and at the end of `body` DOM element:
```html
<body>
  ...
  <div>Add users to your feed</div>
</body>
```
## Basic

screen shot 2018-08-28 at 18 36 13

A basic popup renders a popup with very basic CSS.

```jsx
<Popup basic trigger={<Button content="Add" />} content="Add users to your feed" />
```
generates 
```html
<button>Add</button>
```
and at the end of `body` DOM element:
```html
<body>
  ...
  <div>Add users to your feed</div>
</body>
```

## Alignment and Position

A popup can be rendered in 12 different combinations of `align` and `position` props around the trigger element: 

- `position="above" align="start"`:

| LTR  | RTL |
| ------------- | ------------- |
| screen shot 2018-09-11 at 01 55 02 | screen shot 2018-09-11 at 01 58 00 |

- `position="above" align="center"`:

| LTR  | RTL |
| ------------- | ------------- |
| screen shot 2018-09-11 at 01 55 19 | screen shot 2018-09-11 at 01 58 08 |

- `position="above" align="end"`:

| LTR  | RTL |
| ------------- | ------------- |
| screen shot 2018-09-11 at 01 55 31 | screen shot 2018-09-11 at 01 58 17 |

- `position="below" align="start"`:

| LTR  | RTL |
| ------------- | ------------- |
| screen shot 2018-09-11 at 01 55 44 | screen shot 2018-09-11 at 01 58 27 |

- `position="below" align="center"`:

| LTR  | RTL |
| ------------- | ------------- |
| screen shot 2018-09-11 at 01 55 54 | screen shot 2018-09-11 at 01 58 36 |

- `position="below" align="end"`:

| LTR  | RTL |
| ------------- | ------------- |
| screen shot 2018-09-11 at 01 56 01 | screen shot 2018-09-11 at 01 58 47 |

- `position="before" align="top"`:

| LTR  | RTL |
| ------------- | ------------- |
| screen shot 2018-09-11 at 01 56 09 | screen shot 2018-09-11 at 01 59 46 |

- `position="before" align="center"`:

| LTR  | RTL |
| ------------- | ------------- |
| screen shot 2018-09-11 at 01 56 16 | screen shot 2018-09-11 at 01 59 55 |

- `position="before" align="bottom"`:

| LTR  | RTL |
| ------------- | ------------- |
| screen shot 2018-09-11 at 01 56 23 | screen shot 2018-09-11 at 02 00 03 |

- `position="after" align="top"`:

| LTR  | RTL |
| ------------- | ------------- |
| screen shot 2018-09-11 at 01 56 31 | screen shot 2018-09-11 at 02 00 11 |

- `position="after" align="center"`:

| LTR  | RTL |
| ------------- | ------------- |
| screen shot 2018-09-11 at 01 56 38 | screen shot 2018-09-11 at 02 00 18 |

- `position="after" align="bottom"`:

| LTR  | RTL |
| ------------- | ------------- |
| screen shot 2018-09-11 at 01 56 51 | screen shot 2018-09-11 at 02 00 27 |

```jsx
<Popup
  align="start"
  position="above"
  trigger={<Button content="Add" />}
  content="Add users to your feed"
/>
```
generates 
```html
<button>Add</button>
```
and at the end of `body` DOM element:
```html
<body>
  ...
  <div>Add users to your feed</div>
</body>
```


layershifter:
The positioning part of the `Popup` component is one of complicated things. Can we omit it and use PopperJS there?

/cc @levithomason

bmdalex:
@layershifter 
agreed, RTL and handling viewport limits do not work in this implementation; I took a look at [react-popper package](https://www.npmjs.com/package/react-popper) and looks pretty cool.

@levithomason this is definitely something we want to try; do we want to address it in this PR or should we iterate?

levithomason:
Popper looks extremely promising.  It looks like the React wrapper is also gaining a lot of adoption:
https://www.npmjs.com/package/react-popper

Let's definitely do our due diligence here and compare what we'd gain and lose by switching.  I would love nothing more than to off load the problem components, such as Popup, to a shared community work load.
***
**Edit**

Just saw your latest comment @Bugaa92, yes, let's try Popper now before merging.

src/components/Popup/Popup.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+  public state = { style: {} }
+
+  public render() {
+    const { basic, children, content, trigger } = this.props
+    const { style } = this.state
+
+    return (
+      <Portal onMount={this.handlePortalMount} trigger={trigger} triggerRef={this.handleTriggerRef}>
+        <Popup.Content basic={basic} triggerRef={this.handlePopupRef} styles={{ root: style }}>
```
sophieH29: ```triggerRef``` will be actually set to popup content Ref, so a bit confused about the naming
@Bugaa92

sophieH29:
From the accessibility point of view, as a User, I would like to add Popup behavior which will have defined actions to open/close popup with specified keys 

```js
actionsDefinition: {
    trigger: {
      open: {
        keyCombinations: [{ keyCode: keyboardKey.Enter }, { keyCode: keyboardKey.Spacebar }, { keyCode: keyboardKey.ArrowDown }],
      },
    },
    portal: {
      close: {
      },
    }
  }
```

And since as open/close logic is enclosed into the portal component, can we rethink its API to give a possibility to trigger open/close function from a popup? What your thoughts about it?

@Bugaa92 @levithomason @jurokapsiar

src/components/Popup/Popup.tsx
```diff
@@ -0,0 +1,177 @@
...
+  style?: CSSProperties
+}
+
+/**
+ * A Popup displays additional information on top of a page.
+ * @accessibility This is example usage of the accessibility tag.
+ * This should be replaced with the actual description after the PR is merged
+ */
+export default class Popup extends React.Component<Extendable<IPopupProps>, IPopupState> {
+  private popupOffset = 8
```
sophieH29: Is there any plan to make ```offset``` a prop? So a user can set it as he wants

miroslavstastny:
Also we should think about screener tests for this. With current examples, there is no screener coverage for the examples, because these are the screenshots we compare:
image

So we either need to add examples with popup initially displayed (which will not work for some variants like fullscreen) or better introduce [Test integrations for Screener tests](https://screener.io/v2/docs/test-interactions).

src/components/Popup/Popup.tsx
bmdalex: changed to `innerRef`, thx

src/components/Popup/Popup.tsx
bmdalex: yup, we will do that in next iterations of popup

bmdalex:
@levithomason 
@kuzhelov 
@mnajdova 

what do you guys think about @miroslavstastny 's suggestions regarding visual tests? see his comment below, he has a very good point!

currently we don't seem to have the infrastructure to do this kind of scenario tests (that involve actions) and I'm afraid having the popups open by default are going to offer a bad UX

bmdalex:
achieved to get positioning of popup working with `react-popper` and fixed `RTL` mode as well

@levithomason @layershifter @miroslavstastny 

can you guys please take another look?

package.json
```diff
@@ -139,6 +139,7 @@
     "react-document-title": "^2.0.3",
     "react-dom": "^16.0.0",
     "react-hot-loader": "^4.1.3",
+    "react-popper": "^1.0.2",
```
miroslavstastny: This should go to `dependencies`, not `devDependencies`

docs/src/examples/components/Popup/Variations/PopupExampleBasic.shorthand.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+import { Button, Popup } from '@stardust-ui/react'
+
+const PopupExampleBasic = () => (
+  <Popup basic trigger={<Button icon="expand" />} content="Add users to your feed" />
```
miroslavstastny: I think it would make sense to add Children API example.

src/components/Ref/Ref.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export interface IRefProps {
+  innerRef?: (ref: HTMLElement) => void
+}
+
+/**
+ * This component exposes a callback prop that always returns the DOM node of both functional and class component
+ * children.
+ */
+export default class Ref extends Component<Extendable<IRefProps>, any> {
```
miroslavstastny: - You are not handling rest of props anywhere, remove `Extendable`,
- there is no state in the component, remove `any`:
```jsx
export default class Ref extends Component<IRefProps> {
```

src/components/Portal/Portal.tsx
```diff
@@ -0,0 +1,167 @@
...
+  doesNodeContainClick,
+} from '../../lib'
+import { ItemShorthand, Extendable } from '../../../types/utils'
+import Ref from '../Ref'
+import PortalInner from './PortalInner'
+import { IComponentPartStylesInput, ComponentVariablesInput } from 'theme'
+
+type ReactMouseEvent = React.MouseEvent<HTMLElement>
+
+export interface IPortalProps {
```
miroslavstastny: The interface does not match `propTypes` - `as`, `children`, `styles`, `variables`

src/components/Popup/Popup.tsx
```diff
@@ -0,0 +1,200 @@
...
+enum PositionParts {
+  top = 'top',
+  bottom = 'bottom',
+  left = 'left',
+  right = 'right',
+  before = 'before',
+  after = 'after',
+}
+
+type Position =
```
miroslavstastny: Maybe it's just me, but I am still confused with the naming here. Let's discuss what the two parts of the position name mean in both RTL and LTR and agree on the naming in broader group.
I guess the main source of confusion is `start`/`end` meaning something different in combination with `top`/`bottom` vs `before`/`after`. I can see the inspiration in flexbox'es main/cross axis, but still...

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
```diff
@@ -0,0 +1,61 @@
...
+    />
+    <Popup
+      position="before"
+      trigger={<Button icon="expand" />}
+      content="Popup in before position"
+    />
+    <Popup
+      position="before-end"
+      trigger={<Button icon="expand" />}
+      content="Popup in before-end position"
```
miroslavstastny: It would be great if the popup would be taller. It took me 10 minutes to see the difference among `before-*` positions. (Two lines of text do the trick for me)

src/components/Popup/PopupContent.tsx
```diff
@@ -0,0 +1,79 @@
+import * as _ from 'lodash'
+import * as PropTypes from 'prop-types'
+import * as React from 'react'
+
+import { customPropTypes, UIComponent, IRenderResultConfig } from '../../lib'
+import { ComponentVariablesInput, IComponentPartStylesInput } from '../../../types/theme'
+import { Extendable } from '../../../types/utils'
+
+export interface IPopupContentProps {
```
miroslavstastny: `children?` here?

miroslavstastny (changes requested):
`top-*` and `bottom-*` positions are not flipped in RTL:
image
image

src/components/Ref/Ref.tsx
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Called when componentDidMount.
+     *
+     * @param {HTMLElement} node - Referred node.
+     */
+    innerRef: PropTypes.func,
+  }
+
+  componentDidMount() {
+    _.invoke(this.props, 'innerRef', findDOMNode(this))
```
layershifter: There was a condition in SUIR code, let's use it there, too

package.json
bmdalex: oh damn, fixed, thanks!

docs/src/examples/components/Popup/Variations/PopupExampleBasic.shorthand.tsx
bmdalex: we don't support `Children API` for Popup

src/components/Ref/Ref.tsx
bmdalex: agreed, done

src/components/Portal/Portal.tsx
bmdalex: thanks, fixed

src/components/Popup/PopupContent.tsx
bmdalex: fixed, thanks

src/components/Ref/Ref.tsx
bmdalex: not sure it's necessary since `_.invoke` invokes `innerRef` function only if it's defined; isn't it the same as SUIR?

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
bmdalex: good point, fixed, I'll update screenshots in description

src/components/Popup/Popup.tsx
bmdalex: yea, I'll add this to your meeting notes to discuss it with everybody

bmdalex:
@miroslavstastny 
- I fixed RTL for all positions
- I added UTs to cover these properly
- I updated screenshots with new popup styles and RTL examples to make things more obvious

test/specs/components/Popup/Popup-test.tsx
```diff
@@ -0,0 +1,135 @@
...
+  }) =>
+    it(`Popup ${position} position sends ${expectedPlacement} to Popper props`, () => {
+      const props: IPopupProps = {
+        content: <p />,
+        position,
+        trigger: <button />,
+      }
+      wrapper = mount(
+        <Provider theme={{ rtl }}>
+          <Popup content={<p />} {...props} />
```
miroslavstastny: nit: You are always overwriting the `content` with props. So why is the content here?

test/specs/components/Popup/Popup-test.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+import Provider from 'src/components/Provider'
+import Popup, { IPopupProps, Position } from 'src/components/Popup/Popup'
+import Portal from 'src/components/Portal'
+
+describe('Popup', () => {
+  let wrapper: any
+
+  const mountPopup = (props: IPopupProps = {}, rtl?: boolean) =>
+    (wrapper = mount(<Popup content={<p />} {...props} />))
```
miroslavstastny, line 15: nit: You are always overwriting the `content` with props. So why is the content here?

docs/src/examples/components/Popup/Variations/PopupExampleBasic.shorthand.tsx
miroslavstastny: Hmm, in that case we should at least update props doc:
```jsx
    /** Primary content. */
    children: PropTypes.node,
   ...
    /** Simple text content for the popover. */
    content: customPropTypes.itemShorthand,
```

src/components/Popup/Popup.tsx
```diff
@@ -0,0 +1,203 @@
...
+      )
+    )
+  }
+
+  private renderPopperChildren = (rtl: boolean, { ref, style }: PopperChildrenProps) => {
+    const { basic, content } = this.props
+    const computedStyle = rtl ? rtlCSSJS(style) : style
+
+    return (
+      <Popup.Content innerRef={ref} basic={basic} styles={{ root: computedStyle }}>
```
miroslavstastny: RTL is broken in docsite as `dir="rtl"` is not set on `body` which is a parent of `Popup.Content`.
Do we want to add something like `{...rtl && {dir:"rtl"}}` here? @jurokapsiar

test/specs/components/Popup/Popup-test.tsx
bmdalex: copy paste from a helper function in the same file 🤦‍♂️ 
good catch

docs/src/examples/components/Popup/Variations/PopupExampleBasic.shorthand.tsx
bmdalex: used `customPropTypes.deprecate` for `children` and updated `content` description

src/components/Popup/Popup.tsx
bmdalex: fixed, very good catch; I missed it because of using simple text as content (with no punctuation)

miroslavstastny (approved):

src/components/Popup/positioningHelper.ts
```diff
@@ -0,0 +1,72 @@
...
+  ['bottom', () => PlacementParts.end],
+  ['center', () => PlacementParts.center],
+] as [Alignment, PlacementCb][])
+
+const shouldAlignToCenter = (p: Position, a: Alignment) => {
+  const isPositionedVertically = p === 'above' || p === 'below'
+  const alignedVertically = a === 'top' || a === 'bottom'
+
+  return (
+    (isPositionedVertically && alignedVertically) || (!isPositionedVertically && !alignedVertically)
```
miroslavstastny: nit:
```jsx
isPositionedVertically === alignedVertically
```

miroslavstastny (approved):

sophieH29:
@Bugaa92 I would add a description to each button in the example (with used props ```align="start" position="start"```) as you did in the PR's description. Because currently you're just guessing what position you'll get when clicking on a button :) 
popup _ stardust 2018-09-11 14-54-47

What do you think?

PR #154, opened 2018-08-29
feat(Text): add prop interfaces

kuzhelov:
This PR introduces props interface for Text component, as part of #117


src/components/Text/Text.tsx
```diff
@@ -3,10 +3,30 @@ import * as React from 'react'
...
+  disabled: boolean
+  error: boolean
+  important: boolean
+  size: 'xs' | 'sm' | 'md' | 'lg' | 'xl' | '2x' | '3x' | '4x'
+  weight: 'light' | 'semilight' | 'regular' | 'semibold' | 'bold'
+  success: boolean
+  timestamp: boolean
+  truncated: boolean
+  styles: IComponentPartStylesInput
+  variables: ComponentVariablesInput
```
bmdalex: @kuzhelov shouldn't all these be optional?

kuzhelov: yes, exactly - updating..

miroslavstastny (approved):

PR #156, opened 2018-08-29
feat(accessibility): add HTML accessibility rules validation tests

kolaps33:
# Add AXE validation for unit tests

- to the AXE validator is sent HTML element(as string) which is validated against set of accessibility rules

Failing example:
if icon button has no text which could be narrated by screen readers then axe return error like:
 "Buttons must have discernible text (button-name)"

image


miroslavstastny:
Can we add similar verifications to all docsite examples?

test/specs/components/Button/Button-test.tsx
```diff
@@ -50,6 +52,51 @@ describe('Button', () => {
...
+            </div>,
+          )
+          const results = await axe(html)
+          expectAxe(results).toHaveNoViolations()
+        })
+      })
+
+      describe('different buttons variants', () => {
+        test('button', async () => {
+          const html = ReactDOMServer.renderToString(<Button>Simple test button</Button>)
```
kuzhelov: we are using the same pattern over an over again - essentially
- rendering some portion of JSX to HTML string
- requesting aXe check
- validating result

I would rather suggest to make it all being a function that will be exported by 'expectAxe' module (we might consider to rename it to 'checkAxe' in this case)

kuzhelov (approved):

PR #165, opened 2018-08-30
fix: update button styles to match current implementation

notandrew:
Updating primary & secondary buttons to match the current implementation in Teams. Not updating disabled states or circled buttons.


kuzhelov (approved):

mnajdova (approved):
We should merge this soon, as there are multiple open PRs for the button component that will have conflicts with this.

mnajdova (changes requested):
@notandrew I was working together with @silviuavram for merging the master into this branch as he is currently working on PR dependent on this for applying focus styles on the button when the focus is done with keyboard. There are new variants for buttons added yesterday (the text button and the group of buttons), so we were wondering if the focus styles for these variants are also going to be addressed in this PR. In addition, the default buttons don't have any focus styles, is this intended?

PR #152, opened 2018-08-28
docs: add theming guide

levithomason:
This PR is another breakout of a larger documentation branch.  It adds only a theming guide.

PR #167, opened 2018-08-31
docs: add accessibility guide

hughreeling:
Adding accessibility on top of #152 - adding theming documentation.


docs/src/views/Accessibility.tsx
```diff
@@ -0,0 +1,452 @@
...
+      </li>
+      <li>Properties</li>
+      <li>Variations with examples</li>
+    </ul>
+    <p>
+      Examples show the recommended way of using the components in the different variations. For
+      each example, you are able to edit the JSX code, see the rendered HTML, change themes and
+      validate the rendering in RTL scenario.{' '}
+    </p>
+    <p>Integration of accessibility testing into the DocSite is under consideration.</p>
```
jurokapsiar: Stardust basics and Using DocSite probably do not need to be here since the user already is using doc site

docs/src/views/Accessibility.tsx
```diff
@@ -0,0 +1,452 @@
...
+    <p>
+      While Stardust goes a long way in making the application accessible by default, it does build
+      on having correct semantic HTML as the base.
+    </p>
+
+    <Header as="h3" content="Out of Scope" />
+    <p>
+      Internationalization, globalization, keyboard shortcuts and language detection are
+      deliberately not part of Stardust and should be handled by the hosting application.
+    </p>
```
jurokapsiar, line 56: probably not directly related to accessibility - should we move to some more generic part?

docs/src/views/Accessibility.tsx
```diff
@@ -0,0 +1,452 @@
...
+        'return (',
+        '  <Menu>',
+        '    <Menu.Item content="Editorials" />',
+        '  </Menu>',
+        ')',
+      ].join('\n')}
+    />
+    <p>
+      This way, Stardust will generate a child component based on the provided content string and
+      will apply the role.
```
kohlikohl: Would it be worth adding an example of the generated elements here as well?

docs/src/views/Accessibility.tsx
hughreeling: I think this is important to have here to set scope - so it's clear that these are aspects we're not handling - Can't immediately think of better place to put this.

docs/src/views/Accessibility.tsx
hughreeling: Added.

docs/src/views/Accessibility.tsx
hughreeling: Updated

kolaps33 (approved):

docs/src/views/Accessibility.tsx
```diff
@@ -0,0 +1,517 @@
...
+      to navigate - and although Stardust make this easier to change than pure html implementations
+      - it is still costly in terms of developer time to correctly make these types of changes
+      later.
+    </p>
+    <p>
+      Our strong recommendation here is to have clear accessibility designs and example user flows
+      as early in the design process as possible.{' '}
+    </p>
+
+    <Header as="h1" content="Accessibility" />
```
kuzhelov: frankly, this was a bit surprising to see this section just here after being reading for a while. Actually, broadly applicable thesisis are mentioned here, so I may suggest to introduce this right at the top of the page, as a first thing user will see.

docs/src/components/ExampleSnippet/ExampleSnippet.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react'
+import CodeSnippet, { CodeSnippetProps } from '../CodeSnippet'
+
+const ExampleSnippet = ({
```
kuzhelov: just to be sure - where this one is used?

docs/src/views/Accessibility.tsx
```diff
@@ -0,0 +1,525 @@
...
+      behaviour as they, for example, move between fields in form using tab. This is provided by the
+      browser and for simple applications may be sufficient.
+    </p>
+    <p>
+      However, for complex applications, particularly those which display large amounts of
+      actionable data on the screen the number of 'focusable' elements can become huge and the
+      process of moving between them with tab becomes un-usable. This can be solved by breaking the
+      application up into 'Zones', and the user navigates with the tab key between the zones, and
+      between the actionable elements with navigation keys. Stardust uses Focus Zones both within
+      it's own library components and as a component that can be added by the user.
```
kuzhelov: it is a bit hard to get the idea of this last sentense

kuzhelov (changes requested):
lets agree on whether we do need ExampleSnippet component that is introduced in this PR first - and merge afterwards

kuzhelov (approved):
agreed to remove `ExampleSnippet`

PR #168, opened 2018-08-31
feat: Add StatusIndicator as a separate component

musingh1:
# StatusIndicator

A status indicator is a graphical representation of user's status. Updated the Avatar component to consume this

### TODO

- [x] Conformance test
- [x] Stardust base theme
- [ ] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [x] Confirm RTL usage
- [x] Update the CHANGELOG.md

# API Proposal

## status
image 
image

The status prop is used to denote the current availability of the user.

```jsx
<StatusIndicator
      icon={{ name: 'check', variables: { color: 'white', backgroundColor: 'green' } }}
      title="Available"
    />
```

```html
<div title="Available" class="ui-statusindicator i z ab ac l">
    <span class="ui-icon ae af ag ah ai aj ak al h am an ao ap aq ar as at au av aw ax ay r az ba bb" aria-hidden="true">
    </span>
</div>
```


src/components/PresenceIndicator/PresenceIndicator.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+import { customPropTypes, UIComponent } from '../../lib'
+import { ComponentVariablesInput, IComponentPartStylesInput } from 'theme'
+import { Extendable } from 'utils'
+
+export interface IPresenceIndicatorProps {
+  as?: any
+  className?: string
+  size?: number
+  status?:
```
hughreeling: All very Teams specific values, colours.  I think we should pass object with associates states / colours / texts that represent the possible indicators.  Then this become useful generic component.  Perhaps even presentIndicator is wrong name, perhaps stateIndicator? #Resolved

mnajdova: Totally agree on this. The presence (or state) indicator, as a component should provide name, color, icon and border. We should not try to define generic things here, as I am sure that in the future there will be a library that would require some additional `status` that we haven't defined. This is drawback of the Avatar component as well, and it would be great if we can extract all this in the indicator component. The project that will use the stardust, will define all different statuses they have and provide the indicators available in that project. #Resolved

musingh1: Yes, we had discussed this with Levi and Brian. The plan was to first extract this in a separate component and then allow customizing the states and styles. Brian is already working on some of that. #Closed

musingh1: Renamed the component to StatusIndicator. Experimenting with customizations. #Closed

musingh1: One way to do this is through the theme variables, like I did in the PR, but I have not seen objects in the variables so far and customizing them through the website does not work. 
Another way of doing it (as suggested by @mnajdova ) is to have props for iconName, iconColor, indicatorColor, borderColor. The consuming application can define the objects like const Available = {indicatorColor: 'green', iconName: 'check'...} etc and pass them to the component. Additionally, we could define and expose some states in the component if consumers want to use them directly. I am going to try this approach now. @kuzhelov , @hughreeling, @levithomason  , please let me know your thoughts. #Closed

musingh1: Discussed offline with @kuzhelov , we will address the custom status issue separately. This PR will only separate out StatusIndicator in a separate component #WontFix

src/components/StatusIndicator/StatusIndicator.tsx
```diff
@@ -0,0 +1,119 @@
...
+ */
+class StatusIndicator extends UIComponent<Extendable<IStatusIndicatorProps>, any> {
+  static className = 'ui-statusindicator'
+
+  static displayName = 'StatusIndicator'
+
+  static handledProps = ['as', 'className', 'size', 'status', 'styles', 'variables']
+
+  static propTypes = {
+    /** An element type to render as (string or function). */
```
kuzhelov: could you, please, introduce empty lines between each of the 'prop type + comment' section - so that it would be easier to read this #Resolved

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Adding `behaviors` section to the menu, under the components @kolaps33 ([#119](https://github.com/stardust-ui/react/pull/119))
 - Add `avatar` prop to `Chat.Message` subcomponent @Bugaa92 ([#159](https://github.com/stardust-ui/react/pull/159))
 - add `iconOnly` prop to `Button` @mnajdova ([#182](https://github.com/stardust-ui/react/pull/182))
+- Create StatusIndicator as a separate component @musingh1 ([#168](https://github.com/stardust-ui/react/pull/168))
```
kuzhelov: for the sake of consistency with other CHANGELOG entries, please, brace `StardustIndicator` with back ticks (`) - as it is done for all types/properties/methods. Thanks! #Resolved

kuzhelov (approved):
agreed to move in smaller steps - specifically, by making this PR being responsible for segregating `StatusIndicator` only. 

Concerns around the fact that `StatusIndicator` uses Teams-specific status remain to be valid, but those are decided to be addressed by means of dedicated PR (due to this component is not the only one that introduces this issue - see #169).

mnajdova:
Can we, before merging at least consider using this component inside the Avatar component and be sure that there are no regressions in the look of the component? I am okay for it to be merged as is now, but the point of extracting the component was exactly that... @kuzhelov thoughts? #Resolved

kuzhelov:
agree with @mnajdova . Also, if we'll consider things that we've discussed today during sync up with @levithomason, it seems that there is no much need to segregate this component at the moment, as different parts of the Stardust component (`Avatar` in our example) could be customized - once again,  in our case it will be possible to render custom indicator, as well as to not render anything at all on its place. #Resolved

docs/src/examples/components/Avatar/Variations/AvatarExampleExcludedInitials.shorthand.tsx
```diff
@@ -1,15 +1,20 @@
 import React from 'react'
 import { Avatar } from '@stardust-ui/react'
 
+const Available = {
```
levithomason: We should probably call this `status` or `availableStatus` to follow code conventions.  Uppercase names are unique to classes. #Resolved

docs/src/examples/components/StatusIndicator/Types/StatusIndicatorExample.shorthand.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <StatusIndicator
+      icon={{ name: '', variables: { color: 'white', backgroundColor: 'red' } }}
+      title="Busy"
+    />
+    <StatusIndicator
+      icon={{ name: 'minus', variables: { color: 'white', backgroundColor: 'red' } }}
+      title="Do not disturb"
+    />
+    <StatusIndicator
+      icon={{ name: '', variables: { color: 'white', backgroundColor: 'grey' } }}
```
levithomason: Feels odd to pass an empty name string here.  Let's investigate ways to improve the API. #Resolved

docs/src/examples/components/Avatar/Variations/AvatarExampleStatus.shorthand.tsx
```diff
@@ -1,21 +1,83 @@
...
 const AvatarExampleStatusShorthand = () => (
   <div>
-    <Avatar src="public/images/avatar/small/matt.jpg" status="Available" />
+    <Avatar
+      src="public/images/avatar/small/matt.jpg"
+      status={StatusIndicatorWrapper('Available')}
+    />
     &emsp;
-    <Avatar src="public/images/avatar/small/matt.jpg" status="Busy" />
+    <Avatar src="public/images/avatar/small/matt.jpg" status={StatusIndicatorWrapper('Busy')} />
```
levithomason: Example code should be very concise as its primary purpose is to educate the user.  It is hard to grasp a concept quickly if it is too large or complex.  We also want to avoid making the user scroll up and down to see/edit the code when viewing the rendered example.

Issues
- This file is too many lines.
- Let's also remove the function call here for something more simple.  It looks like this could be done with an object instead, `statusIndicators.busy` should work.   Also, using props objects instead of `<StatusIndicator />` elements would help clean it up. #Resolved

docs/src/examples/components/Avatar/Variations/AvatarExampleName.shorthand.tsx
```diff
@@ -1,6 +1,13 @@
...
-import { Avatar } from '@stardust-ui/react'
+import { Avatar, StatusIndicator } from '@stardust-ui/react'
 
-const AvatarExampleNameShorthand = () => <Avatar status="Available" name="John Doe" />
+const Available = (
+  <StatusIndicator
+    icon={{ name: 'check', variables: { color: 'white', backgroundColor: 'green' } }}
+    title="Available"
+  />
+)
```
levithomason: This could also be a props object for simplicity, correct?

```jsx
const status = {
  icon: { name: 'check', variables: { color: 'white', backgroundColor: 'green' } },
  title: "Available",
}
```

This removes a couple lines and the need to import the status indicator. #Closed

src/components/Avatar/Avatar.tsx
```diff
@@ -177,18 +125,14 @@ class Avatar extends UIComponent<Extendable<IAvatarProps>, any> {
...
-              size={size < 4 ? 'micro' : size < 6 ? 'mini' : 'tiny'}
-              circular
-              name={icon}
-              variables={iconVariables}
-              title={status}
-            />
-          </div>
-        )}
+        {status &&
+          StatusIndicator.create(status, {
```
levithomason: There is no need to short circuit if `status` does not exist.  The factory will safely handle falsy values.  The `status && ` syntax can be removed. #Resolved

src/components/StatusIndicator/StatusIndicator.tsx
```diff
@@ -0,0 +1,82 @@
...
+  variables?: ComponentVariablesInput
+}
+
+/**
+ * A status indicator is a graphical representation of user's status
+ */
+class StatusIndicator extends UIComponent<Extendable<IStatusIndicatorProps>, any> {
+  static create: Function
+
+  static className = 'ui-statusindicator'
```
levithomason, line 24: We may have a shortcoming in our conformance tests that would fail this, but the `className` should be `status-indicator` since it is multiple words. #Pending

musingh1:
Updated the Avatar component to consume StatusIndicator

---
In reply to: [418745231](https://github.com/stardust-ui/react/pull/168#issuecomment-418745231) [](ancestors = 418745231)

musingh1:
Updated both Avatar and StatusIndicator. I think this should allow easy customizations for this component.

---
In reply to: [418807844](https://github.com/stardust-ui/react/pull/168#issuecomment-418807844) [](ancestors = 418807844)

src/components/PresenceIndicator/PresenceIndicator.tsx
musingh1: Updated

---
In reply to: [214710476](https://github.com/stardust-ui/react/pull/168#discussion_r214710476) [](ancestors = 214710476)

docs/src/examples/components/StatusIndicator/Types/StatusIndicatorExample.shorthand.tsx
musingh1: Removed the empty icon for now. We can revisit this later.

---
In reply to: [215418325](https://github.com/stardust-ui/react/pull/168#discussion_r215418325) [](ancestors = 215418325)

src/components/StatusIndicator/StatusIndicator.tsx
musingh1: yes, status-indicator results in test failure. Will keep it this way for now and we can fix it later.

---
In reply to: [215421511](https://github.com/stardust-ui/react/pull/168#discussion_r215421511) [](ancestors = 215421511)

PR #169, opened 2018-08-31
feat(Text): Adding timestamp hover treatment for the Teams theme

codepretty:
## Hover treatment for timestamp text in the Teams theme

**Before** (there was no hover treatment)
image

**After**
image


kuzhelov:
This PR showcases few important points that we should consider in the short-term perspective. 

Looking on the props (or, semantically, variants) of Text component, it seems that several of them are not generic enough to be declared as 'good for web' (i.e. be part of general Stardust) - as those are rather Teams-specific:
- `timestamp`
- `atmention`

Here are couple of ways we might consider to resolve the problem
- on the side of `teams` module, consider to introduce higher-order component for each variant (like `Timestamp`, `AtMention`) - so that those will reuse Text functionality inside, as well as expose the API (props) interface that is relevant for each specific variant only. In that case those variants will come as part of `teams` module, so that general `Text` API will contain only general props
- rethink how each theme could introduce its own custom properties for components: in that case Stardust's `Text` component will remain to have just general set of props, and `teams` theme will be responsible to introduce concept of `atMention` and `timestamp` variants. 
  - here the main problem is the following: now our components declare `handledProps` - those that won't be passed to their children. If we'll suggest to handle some custom props in theme, this would imply that consumer is welcomed to introduce these custom props to component - and **_those will always come up as `unhandled` ones, thus, will be passed to children tree_**. There should be a way to extend `handledProps` set of a component from theme then.

kuzhelov (approved):

mnajdova (approved):

Issue #161, opened 2018-08-30
[RFC] Provide merged props and state to Styles and Accessibility


silviuaavram:
## Feature Request
The feature is meant to provide a unified object of the component's `props` and `state` in the context of Styles and Accessibility.

### Problem description
So far, in both the cases of Accessibility and Styles, it was required that we have access to both state and props when applying custom behaviour or styles to the component. 

The following scenarios should be considered as motivation for this feature:
- accessibility: we need to provide an `aria-checked` to a checkbox, depending on its `check` state.
- styles: we have hit the issue where we need to differentiate focus source (e.g. keyboard or mouse) so we can apply a different style to the component. If the focus comes from the mouse, this is reflected in the component's `state`, and then we need to apply different style (removing the focus ring) based on that. 

### Proposed solution
- unified (merged) object of `props` and `state` should be passed as argument to component's style functions, as well as to the accessibility behaviours. 
- if there is a property that is common for both `state` and `props` objects, the one from `props` will override the one from `state`). 

### Which name should be used for merged object?
Currently we are using `props` as name for the arg that is passed to style function, and the same for the one that is passed to accessibility behavior. For the sake of not introducing misleading associations with React domain (this will quite loudly apply to style functions, where `props` arg will contain `props+state` data) it is better to consider another name for it. Proposed options so far:
- `properties` (or `componentProperties`)
- `data` (or `componentData`)

levithomason:
Problem 👍 
Proposed solutions 1 and 2 👍 

As for the style function argument, even though `props` was popularized by React it seems many CSS in JS libraries call this argument `props`, even if they aren't supporting React.  Some also call it `state`, as CSS in JS is also described as "styles as a function of state."

I think we should look around at the ecosystem and make a determination based on what is emerging and stay away from inventing anything new.  Here are a few popular libraries

https://github.com/Khan/aphrodite
https://github.com/paypal/glamorous
https://github.com/FormidableLabs/radium
https://emotion.sh/
https://github.com/cssinjs/jss
http://fela.js.org/
https://github.com/styled-components/styled-components
https://github.com/styled-components/polished
https://github.com/styletron/styletron

Even more are listed here here:

https://github.com/MicheleBertoli/css-in-js

kuzhelov:
required functionality is now introduced, let's spend some time to figure out if something needs to be changed for the name of `props` - based on the experience of other libs (totally agree that we'd better to stick to commonly taken approach).

[...]

PR #173, opened 2018-09-03
feat(renderComponent): merged state and props for styles and accessibility

bmdalex:
# Component style function

This PR adds the ability to receive a unified (merged) object of `props` and `state` as argument to component's style functions, as well as to the accessibility behaviors.

More details in the dedicated issue: #161 

### TODO

- [x] Conformance test
- [ ] Stardust base theme
- [ ] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [ ] Confirm RTL usage
- [x] Update the CHANGELOG.md

# API Proposal

## Before

```typescript
const buttonStyles: IComponentPartStylesInput = {
  root: ({
    props,
    variables
  }: { props: IButtonProps; variables: IButtonVars }): ICSSInJSStyle => {
    // props contains all props properties of the button
    return {}
  },
}
```

## After

```typescript
const buttonStyles: IComponentPartStylesInput = {
  root: ({
    props,
    variables
  }: { props: IButtonProps & IButtonState; variables: IButtonVars }): ICSSInJSStyle => {
    // props contains props and state properties merged, with props taking priority
    // in case of overlapping properties
    return {}
  },
}
```

### DIFF
--- a/src/lib/renderComponent.tsx
+++ b/src/lib/renderComponent.tsx
@@ -10,12 +10,12 @@ import getUnhandledProps from './getUnhandledProps'
 
 import {
   ComponentStyleFunctionParam,
-  ComponentVariablesInput,
   ComponentVariablesObject,
   IComponentPartClasses,
-  IComponentPartStylesInput,
   IComponentPartStylesPrepared,
   IProps,
+  IPropsWithVarsAndStyles,
+  IState,
   IThemeInput,
   IThemePrepared,
 } from '../../types/theme'
@@ -34,27 +34,18 @@ export interface IRenderResultConfig<P> {
 
 export type RenderComponentCallback<P> = (config: IRenderResultConfig<P>) => any
 
-type IRenderConfigProps = {
-  [key: string]: any
-  variables?: ComponentVariablesInput
-  styles?: IComponentPartStylesInput
-}
-
 export interface IRenderConfig {
   className?: string
   defaultProps?: { [key: string]: any }
   displayName: string
   handledProps: string[]
-  props: IRenderConfigProps
-  state: { [key: string]: any }
+  props: IPropsWithVarsAndStyles
+  state: IState
 }
 
-const getAccessibility = <P extends {}>(props, state) => {
+const getAccessibility = (props: IState & IPropsWithVarsAndStyles) => {
   const { accessibility: customAccessibility, defaultAccessibility } = props
-  return callable(customAccessibility || defaultAccessibility || DefaultBehavior)({
-    ...props,
-    ...state,
-  })
+  return callable(customAccessibility || defaultAccessibility || DefaultBehavior)(props)
 }
 
 const renderComponent = <P extends {}>(
@@ -83,7 +74,11 @@ const renderComponent = <P extends {}>(
 
         // Resolve styles using resolved variables, merge results, allow props.styles to override
         const mergedStyles = mergeComponentStyles(componentStyles[displayName], props.styles)
-        const styleParam: ComponentStyleFunctionParam = { props, variables: resolvedVariables }
+        const stateAndProps = { ...state, ...props }
+        const styleParam: ComponentStyleFunctionParam = {
+          props: stateAndProps,
+          variables: resolvedVariables,
+        }
         const resolvedStyles = Object.keys(mergedStyles).reduce(
           (acc, next) => ({ ...acc, [next]: callable(mergedStyles[next])(styleParam) }),
           {},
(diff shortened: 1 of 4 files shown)


test/specs/commonTests/stylesFunction-test.tsx
```diff
@@ -0,0 +1,115 @@
...
+      expected: { propsAttr: 'props' },
+    })
+
+    testStylesForComponent({
+      props: { propsAttr: 'props', commonAttr: 'props' },
+      expected: { propsAttr: 'props', commonAttr: 'props' },
+    })
+  })
+
+  it('simple merge of props and state if both are set but there are no overlapping properties', () => {
```
kuzhelov: it would be better to describe result in the test name, like `props and state properties are merged together` and `props properties take precedence over ones of state` for the next one

src/lib/renderComponent.tsx
```diff
@@ -83,7 +73,11 @@ const renderComponent = <P extends {}>(
 
         // Resolve styles using resolved variables, merge results, allow props.styles to override
         const mergedStyles = mergeComponentStyles(componentStyles[displayName], props.styles)
-        const styleParam: ComponentStyleFunctionParam = { props, variables: resolvedVariables }
+        const appliedProps = { ...state, ...props }
```
kuzhelov: 👍

src/lib/renderComponent.tsx
```diff
@@ -34,27 +33,18 @@ export interface IRenderResultConfig<P> {
...
 }
 
-const getAccessibility = <P extends {}>(props, state) => {
+const getAccessibility = props => {
   const { accessibility: customAccessibility, defaultAccessibility } = props
-  return callable(customAccessibility || defaultAccessibility || DefaultBehavior)({
-    ...props,
-    ...state,
-  })
+  return callable(customAccessibility || defaultAccessibility || DefaultBehavior)(props)
```
kuzhelov, line 48: accessibility functions end up with being provided with state only now - but previously `state` data was provided as well (this was a correct and expected behavior)

test/specs/commonTests/stylesFunction-test.tsx
bmdalex: I don't think I understand, let's discuss

src/lib/renderComponent.tsx
bmdalex: @jurokapsiar is this correct here?
@kuzhelov anything to change?

jurokapsiar: accessibility need a merge of props and state. props should override state. looking at line 83, I think this is correctly implemented now.

kuzhelov: I would rather change the name of the argument, because it is a bit misleading - to indicate that 'state' is merged to 'props'

src/lib/renderComponent.tsx
kuzhelov: one note - before we'll come up with a name for this 'merged' object, maybe we could use something that is a bit dumb but explicit and descriptive - like `stateAndProps`?

types/theme.d.ts
```diff
@@ -77,7 +86,7 @@ export interface ICSSInJSStyle extends React.CSSProperties {
 }
 
 export interface ComponentStyleFunctionParam {
-  props: IProps
+  props: IState & IProps
```
kuzhelov: 👍

types/theme.d.ts
```diff
@@ -20,7 +21,15 @@ type ObjectOf<T> = { [key: string]: T }
 // Props
 // ========================================================
 
-export type IProps = ObjectOf<any>
+export type IProps = Extendable<{
+  variables?: ComponentVariablesInput
```
kuzhelov: this won't work for all the components, unfortunately (`Provider` as an example) - lets keep to be generic for now

src/lib/renderComponent.tsx
```diff
@@ -34,27 +33,18 @@ export interface IRenderResultConfig<P> {
...
-}
-
 export interface IRenderConfig {
   className?: string
   defaultProps?: { [key: string]: any }
   displayName: string
   handledProps: string[]
-  props: IRenderConfigProps
-  state: { [key: string]: any }
+  props: IProps
```
kuzhelov: lets use the type that was used before - as `IProps` is not generic enough to suite all the components if `styles` and `variables` are introduced

bmdalex: done

src/lib/renderComponent.tsx
bmdalex: done, thanks guys

src/lib/renderComponent.tsx
bmdalex: done, thanks

test/specs/commonTests/stylesFunction-test.tsx
bmdalex: discussed with @kuzhelov and amended

types/theme.d.ts
bmdalex: ok, done

kuzhelov (approved):
👍

PR #177, opened 2018-09-03
feat(Button): add text variant

mnajdova:
# Button

Adding text variant to the button, that will display the button without any background and border.

### TODO

- [x] Conformance test
- [ ] Stardust base theme
- [ ] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [ ] Confirm RTL usage
- [x] Update the CHANGELOG.md

# API Proposal

## text: boolean

image

A button can be formatted to show only text in order to indicate some less-pronounced actions.

```jsx
<Button text icon="book" content="Default" iconPosition="before" />
```

```html
<button class="ui-button ..." aria-disabled="false">
  <span class="q r s">Default</span>
</button>
```

At this moment, the styles of the buttons containing the text prop, are just reducing the background and border, and on hover are altering the color of the text to be the `brand` color. We may need to change this, as the current text color of the primary button is white, which is maybe not the most appropriate one. Once the API is approved, we can agree on colors used for the default, primary and secondary variations of the text button.


miroslavstastny:
Shorthand vs Children API examples render differently

mnajdova:
Thanks for the catch @miroslavstastny, it is fixed now.

docs/src/examples/components/Button/Variations/ButtonExampleText.shorthand.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react'
+import { Button } from '@stardust-ui/react'
+
+const ButtonExampleTextShorthand = () => (
+  <div>
+    <Button text icon="book" content="Default" iconPosition="before" />
+    <Button text type="primary" content="Primary" />
+    <Button text type="secondary" content="Secondary" />
+    <Button text circular icon="coffee" />
```
miroslavstastny: Does `text circular` have any use?

miroslavstastny (commented):
I am fine with the API proposal. I would like to see a use case for the text button in order to be able to discuss it's styling aspects - I am not sure if you want to keep the same button dimensions (height and padding).

docs/src/examples/components/Button/Variations/ButtonExampleText.shorthand.tsx
mnajdova: This example just showcase how every button variation would work together with the text. I agree that maybe text circular doesn't make sense, but at this moment it actually have different result exactly because the dimension would be different then the other combination. I agree that we may reconsider changing the height and width when the text prop is applied. The only definition we have for the text button is 'a button that doesn't have any background and border'. The usage would be something like this:
image

docs/src/examples/components/Button/Variations/index.tsx
```diff
@@ -19,6 +19,11 @@ const Variations = () => (
       description="A button can be circular and formatted to show different levels of emphasis."
       examplePath="components/Button/Variations/ButtonExampleEmphasisCircular"
     />
+    <ComponentExample
+      title="Text"
+      description="A button can be shown in form of a text to indicate some less-pronounced actions"
```
kuzhelov: nit's nit: just for the sake of consistency, lets put dot at the end of description :)

miroslavstastny (approved):

docs/src/examples/components/Button/Variations/index.tsx
mnajdova: This is not a nit, is valid thing to be fixed as this is shown in our docs page :)

src/components/Button/Button.tsx
```diff
@@ -140,27 +145,19 @@ class Button extends UIComponent<Extendable<IButtonProps>, any> {
...
 
-  public renderIcon = variables => {
-    const { icon, iconPosition, content, type } = this.props
+  public renderIcon = () => {
+    const { icon, iconPosition, content } = this.props
 
     return Icon.create(icon, {
       defaultProps: {
         xSpacing: !content ? 'none' : iconPosition === 'after' ? 'before' : 'after',
-        variables: {
```
kuzhelov, line 161: 👍

kuzhelov (approved):

mnajdova:
Thanks for the updates @kuzhelov, will update the branch and merge it.

Issue #162, opened 2018-08-30
[RFC] Decouple variables passed to child components from parent component's logic


kuzhelov:
from @levithomason post:

## Motivation

Any variables or styles used directly in a component's logic implicitly become part of a global theming interface.  

All themes will have to implement `variables.typePrimaryColor`, for instance, in order for this to work.  Additionally, we are removing the styling from the theme and baking it into the component instead.

We would like all theming values and theming logic living in the theme's files.  This way, each theme is free to choose the values and logic behind their styles.  The question is how?

### Proposal 1 - Component part variables

Component variables files could define variables for each component parts as well.  Then, variable values and logic for component parts can be left to the theme and not dictated to all themes by the component.  Here's an example of the Button theme defining an `icon` part in its variables.

```jsx
// themes/teams/components/Button/buttonVariables.tsx

export default (siteVariables) => {
  const variables = {
    color: undefined,
    typePrimaryColor: siteVariables.white,
    typeSecondaryColor: siteVariables.black,
  }

  variables.icon = {
    color:
      type === "primary"
        ? variables.typePrimaryColor
        : type === "secondary"
          ? variables.typeSecondaryColor
          : variables.color
  }

  return variables
}
```

Now, the component simply implements a contract that says "we pass variables[part] to each component part":

```jsx
// Button.tsx

Icon.create(icon, {
  defaultProps: {
    variables: variables.icon
  }
})
```

This would also allow us to do this consistently for all components, including the styles for each component part.  That then would allow us to write a conformance test for it.  This would make the API easier to understand and work with as well since there is only one concept, "components pass variables/styles to each component part".

### Proposal 2 - A high-level theme interface

This proposal would introduce a new concept.  Something like a lightweight theme interface that is allowed to pass between components.  It would consist of `backgroundColor`, `foregroundColor`, and perhaps an `accentColor` initially (names and keys are inspired by present convergence talks with several teams at Microsoft).

The Button in this case would define these three values for its bounds.  Then, they would passed to the Icon.

/cc @levithomason, @miroslavstastny

kuzhelov:
Agree with all the points provided. We've had similar discussion with @miroslavstastny where exactly the same points were raised by him - at that point we've decided to pass variables directly in component's logic, primarily for the sake of not introducing additional complexity for theme mechanisms that were just introduced then (and, thus, were not exploited enough to be considered as robust solution).

[...]

PR #179, opened 2018-09-04
feat(Button): added ButtonGroup component

mnajdova:
# ButtonGroup

The ButtonGroup component, as part of the Button component, will be used to organize multiple buttons to apear together.

### TODO

- [x] Conformance test
- [ ] Stardust base theme
- [ ] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [ ] Confirm RTL usage
- [x] Update the CHANGELOG.md

# API Proposal

## circular?: boolean

This property is used for creating group of circular buttons

image

```jsx
<Button.Group
    circular
    buttons={[
      { icon: 'book', type: 'primary' },
      { icon: 'coffee' },
      { icon: 'play', type: 'primary' },
    ]}
  />
```

```html
<div class="ui-buttons">
  <button class="ui-button ..." aria-disabled="false">
  </button>
  <button class="ui-button ..." aria-disabled="false">
  </button>
  <button class="ui-button ..." aria-disabled="false">
  </button>
</div>
```

## buttons?: ItemShorthand[]

Shorthand for setting the buttons that should appear in the group.

image

```jsx
<Button.Group buttons={[{ icon: 'book', iconOnly: true }, { icon: 'coffee', iconOnly: true }, { icon: 'play', iconOnly: true }]} />
```

```html
<div class="ui-buttons">
  <button class="ui-button ..." aria-disabled="false">
  </button>
  <button class="ui-button ..." aria-disabled="false">
  </button>
  <button class="ui-button..." aria-disabled="false">
  </button>
</div>
```

## variables?: ComponentVariablesInput

# Question
Do we have mechanism for testing collectionShorthand (list of itemShorthand). If not, should I create one and use for testing the buttons inside the ButtonGroup component?


layershifter:
@mnajdova we don't have common tests for collections in SUIR, so I think that Stardust doesn't have them, too.

mnajdova:
Got it, thanks @layershifter, will add one common test for the collectionShorthand.

test/specs/commonTests/implementsCollectionShorthandProp.tsx
```diff
@@ -0,0 +1,68 @@
...
+    options: ShorthandTestOptions = DefaultShorthandTestOptions,
+  ) {
+    const { mapsValueToProp } = options
+
+    describe(`shorthand property for '${ShorthandComponent.displayName}'`, () => {
+      test(`is defined`, () => {
+        expect(Component.propTypes[shorthandPropertyName]).toBeTruthy()
+      })
+
+      test(`array of string values are handled as ${
```
kuzhelov: grammar nit: 'as' -> 'is' :)

test/specs/commonTests/implementsCollectionShorthandProp.tsx
```diff
@@ -0,0 +1,68 @@
...
+        ShorthandComponent.displayName
+      }s' ${mapsValueToProp}`, () => {
+        const shorthandValue = ['some value', 'some other value']
+        const props = { [shorthandPropertyName]: shorthandValue }
+        const wrapper = mount(<Component {...props} />)
+
+        const shorthandComponentProps = wrapper.find(ShorthandComponent.displayName)
```
kuzhelov: seems that name of the variable is not correct - due to `wrapper.find(ShorthandComponent.displayName)` returns a nodes wrapper, not props

test/specs/commonTests/implementsCollectionShorthandProp.tsx
kuzhelov: would also recommend to consider test renaming, to better indicate what is going on, something like 'array of string values **_is spread as_** ...'

test/specs/commonTests/implementsCollectionShorthandProp.tsx
mnajdova: Thanks for the catch! I changed it in the other test and forgot about this one.. :)

test/specs/commonTests/implementsCollectionShorthandProp.tsx
mnajdova: Agreed, it is better explained that way!

test/specs/commonTests/implementsCollectionShorthandProp.tsx
```diff
@@ -0,0 +1,68 @@
...
+        const allShorthandPropertiesArePassedToFirstShorthandComponent = Object.keys(
+          _.first(shorthandValue),
+        ).every(
+          propertyName =>
+            propertyName === 'key'
+              ? true
+              : _.first(shorthandValue)[propertyName] ===
+                shorthandComponents.first().prop(propertyName),
+        )
+        const allShorthandPropertiesArePassedToLastShorthandComponent = Object.keys(
```
kuzhelov: nit: it would be a bit easier to read if there will be an empty line between these statements

test/specs/commonTests/implementsCollectionShorthandProp.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+        const props = { [shorthandPropertyName]: shorthandValue }
+        const wrapper = mount(<Component {...props} />)
+
+        const shorthandComponents = wrapper.find(ShorthandComponent.displayName)
+        const allShorthandPropertiesArePassedToFirstShorthandComponent = Object.keys(
+          _.first(shorthandValue),
+        ).every(
+          propertyName =>
+            propertyName === 'key'
```
kuzhelov: can be simplified

```js
propertyName => propertyName === 'key' || _.first(shorthandValue)[propertyName] === shorthandComponents.first().prop(propertyName)
```

test/specs/components/Button/ButtonGroup-test.tsx
```diff
@@ -2,7 +2,12 @@ import * as React from 'react'
...
 import { isConformant } from 'test/specs/commonTests'
 import ButtonGroup from 'src/components/Button/ButtonGroup'
+import implementsCollectionShorthandProp from '../../commonTests/implementsCollectionShorthandProp'
+import Button from 'src/components/Button'
+
+const buttonGroupImplementsShorthandProp = implementsCollectionShorthandProp(ButtonGroup)
 
 describe('ButtonGroup', () => {
   isConformant(ButtonGroup)
+  buttonGroupImplementsShorthandProp('buttons', Button)
```
kuzhelov: 👍

src/index.ts
```diff
@@ -3,6 +3,7 @@ export { themes }
 
 export { default as Accordion } from './components/Accordion'
 export { default as Button } from './components/Button'
+export { ButtonGroup } from './components/Button'
 export { default as Chat } from './components/Chat'
 export { default as ChatMessage } from './components/Chat'
```
kuzhelov: seems that we have a problem here - as `Chat` is exported as `ChatMessage`. Could you, please, fix this? :)

src/themes/teams/components/Button/buttonVariables.ts
```diff
@@ -6,6 +6,7 @@ export interface IButtonVariables {
   height: string
   minWidth: string
   maxWidth: string
+  defaultBorderRadius: string
```
kuzhelov: are we using any different other than `default` one for `Button`? If not, we should probably consider to rename it to something like just `borderRadius` - what do you think?

src/components/Button/ButtonGroup.tsx
```diff
@@ -0,0 +1,116 @@
...
+  variables?: ComponentVariablesInput
+}
+
+/**
+ * A button group.
+ */
+class ButtonGroup extends UIComponent<Extendable<IButtonGroupProps>, any> {
+  public static displayName = 'ButtonGroup'
+
+  public static className = 'ui-button__buttons'
```
kuzhelov: seems that `BEM` conventions are violated here - this reads as `'buttons' element inside 'ui-button' - but I would rather expect something simpler like 'ui-buttons' or 'ui-button-group'`

src/components/Button/ButtonGroup.tsx
```diff
@@ -0,0 +1,116 @@
...
+import { Extendable, ItemShorthand, ReactChildren } from '../../../types/utils'
+import Button from './Button'
+
+export interface IButtonGroupProps {
+  as?: any
+  children?: ReactChildren
+  circular?: boolean
+  className?: string
+  content?: React.ReactNode
+  buttons?: ItemShorthand[]
```
kuzhelov, line 16: it seems that we have the same story for `Menu` and `List` - could we introduce shorthand collection tests for them as well, please? I know that it is not related to the PR, but something that will ensure consistency with these changes being introduced

src/components/Button/ButtonGroup.tsx
```diff
@@ -0,0 +1,116 @@
...
+              circular,
+              styles: { root: this.getStyleForButtonIndex(styles, buttons, idx) },
+            },
+          }),
+        )}
+      </ElementType>
+    )
+  }
+
+  getStyleForButtonIndex = (styles, buttons, idx) => {
```
kuzhelov: it seems that logic of this function relies (semantically) only on `isFirst` and `isLast` flags - for the sake of making its interface lighter and more explicit, could we introduce these two instead of `buttons` and `idx`?

src/components/Button/ButtonGroup.tsx
```diff
@@ -0,0 +1,116 @@
...
+      )
+    }
+
+    return (
+      <ElementType {...rest} className={classes.root}>
+        {_.map(buttons, (button, idx) =>
+          Button.create(button, {
+            defaultProps: {
+              circular,
+              styles: { root: this.getStyleForButtonIndex(styles, buttons, idx) },
```
kuzhelov: this showcases the problem we currently have for `variables` as well (#162) - with current implementation approach we will have the same negative consequences for styles as the ones described there for variables. Lets consider to address this moment later

kuzhelov (approved):
please, take a look on comments before merging - nothing critical, though. Awesome job with introducing `implementsCollectionShorthandProp` 👍

src/components/Button/ButtonGroup.tsx
mnajdova: We have utility function for generating the className inside getComponentInfo.ts: 

  info.componentClassName = (info.isChild
    ? \`ui-${info.parentDisplayName}__${info.subcomponentName.replace(
        /Group$/,
        \`${info.parentDisplayName}s\`,
      )}\`
    : \`ui-${info.displayName}\`
  ).toLowerCase()

I was just following it in order to make the component conformant. Do you think we should change this logic?

src/themes/teams/components/Button/buttonVariables.ts
mnajdova: Yes, we have circularRadius as well, that is why this is named like that.

src/components/Button/ButtonGroup.tsx
kuzhelov: the logic of this function seems correct to me - the problem is that for `info.parentDisplayName` 'button' is used for the case of `ButtonGroup`, and this is semantically incorrect

mnajdova: In the end, we decided that the components that contain Group inside the name should result into the main components class with appended 's' in the end (For the ButtonGroup it should be ui-buttons). We are doing this because we already had some logic for the Group components,  and we can still keep the group components together with the main components that they represent.

PR #180, opened 2018-09-04
[RFC] Extendable Portal functionality

sophieH29:
This PR offers an alternative way of creating a ```Portal``` component.

Based on @Bugaa92's PRs https://github.com/stardust-ui/react/pull/144/, https://github.com/stardust-ui/react/pull/150/, 
- ```Popup``` uses ```Portal``` inside during rendering
-  Open/Close logic is enclosed in a ```Portal```, so Popup has no access to it
- if to use ```trigger``` element outside of ```Portal```, some logic can be missed and will need to be repeated, like handling document click

### Alternative way

The alternative way is to have all common logic from Portal in some abstract component, let's name it ```PortalGeneric```. And then it can be inherited from ```Popup```, ```Portal``` or any other specific custom component.
```PortalGeneric``` has logic of ```renderPortal```, ```renderTrigger```, ```open```, ```close```. So then it can be simply called from child component.

generic-portal

#### Pros:
- Base functionality stays in one place, can be overridden by children if needed
- Inherited components have a cleaner look
- From the Accessibility perspective, keydown action handlers can have access to open/close methods and manipulate focus, by passing ```afterRenderClbk``` to those methods

#### Cons:
- for accessibility it will be needed to extend ```renderTrigger``` method to pass extra aria-* attributes and keydown handlers. Not sure, if it is related to cons, but probably breaks some consistency.

Please, guys, feel free to share your opinion on this proposal, so we can have a broader discussion to see if this worth doing it.

**P.S.:** during a review, better concentrate on ```PortalGeneric.tsx```, ```Portal.tsx``` and ```Popup.tsx``` files, as the main changes were made there. The rest is just the same as in Alex's PRs.

### CODE REVIEW

File: src/lib/PortalGeneric.tsx [hunk 1]
```diff
@@ -0,0 +1,132 @@
...
+
+  protected open = (afterRenderClbk?: Function) => {
+    this.setState({ open: true }, () => afterRenderClbk && afterRenderClbk())
+  }
+
+  protected close = (afterRenderClbk?: Function) => {
+    this.setState({ open: false }, () => afterRenderClbk && afterRenderClbk())
+  }
+
+  protected handleTriggerClick = (e: ReactMouseEvent, trigger) => {
```
bmdalex: type for `trigger`

src/lib/PortalGeneric.tsx
```diff
@@ -0,0 +1,132 @@
...
+    trigger && _.invoke(trigger, 'props.onClick', e, trigger.props) // Call original event handler
+
+    this.setState({ open: !this.state.open })
```
bmdalex: if you're introducing open and close, shouldn't this also call the `afterRenderCb`? that would be:
```typescript
if (this.state.open) {
  this.close(afterRenderCb)
} else {
  this.open(afterRenderCb)
}
```

src/lib/PortalGeneric.tsx
```diff
@@ -0,0 +1,132 @@
...
+    this.portalNode = portalNode
+  }
+
+  protected handleTriggerRef = (triggerNode: HTMLElement) => {
+    this.triggerNode = triggerNode
+
+    _.invoke(this.props, 'triggerRef', triggerNode)
+  }
+
+  protected open = (afterRenderClbk?: Function) => {
```
bmdalex: nit: rename to `afterRenderCb`

src/lib/PortalGeneric.tsx
```diff
@@ -0,0 +1,132 @@
...
+  protected handleDocumentClick = (e: ReactMouseEvent) => {
+    if (
+      !this.portalNode || // no portal
+      doesNodeContainClick(this.triggerNode, e) || // event happened in trigger (delegate to trigger handlers)
+      doesNodeContainClick(this.portalNode, e) // event happened in the portal
+    ) {
+      return // ignore the click
+    }
+
+    this.setState({ open: false })
```
bmdalex: let's call `this.close()` to have only one entry point for handling opening and closing the portal

src/lib/PortalGeneric.tsx
```diff
@@ -0,0 +1,132 @@
...
+     * Called when the portal is unmounted from the DOM.
+     *
+     * @param {object} data - All props.
+     */
+    onUnmount: PropTypes.func,
+  }
+
+  static autoControlledProps = ['open']
+
+  getInitialAutoControlledState() {
```
bmdalex: is this meant to be public?

src/components/Portal/Portal.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  public state = { open: false }
+
+  public renderComponent({  }: IRenderResultConfig<IPortalProps>): React.ReactNode {
+    const { content, children, trigger } = this.props
+    const portalContent = (childrenExist(children) ? children : content) as JSX.Element
+    return (
+      <React.Fragment>
+        {this.renderPortal(portalContent)}
+        {this.renderTrigger(trigger as JSX.Element)}
```
bmdalex: no need for `trigger as JSX.Element` casting, just `trigger` should be enough

src/components/Popup/Popup.tsx
```diff
@@ -0,0 +1,197 @@
...
+      </React.Fragment>
+    )
+  }
+
+  protected handleMount() {
+    super.handleMount()
+    this.setPopupStyle()
+  }
+
+  private computePopupStyle(): CSSProperties {
```
bmdalex, line 132: this will be removed after the approach `react-popper` will become stable

src/components/Popup/Popup.tsx
```diff
@@ -0,0 +1,197 @@
...
+    defaultOpen: PropTypes.bool,
+  }
+
+  public static defaultProps = {
+    position: 'top start',
+  }
+
+  public state = { style: {}, open: false }
+
+  public renderComponent({  }: IRenderResultConfig<IPopupProps>): React.ReactNode {
```
bmdalex: just remove the args of `renderComponent ` if you're not using them

src/components/Portal/Portal.tsx
```diff
@@ -0,0 +1,76 @@
...
+     * Called with a ref to the trigger node.
+     *
+     * @param {JSX.Element} node - Referred node.
+     */
+    triggerRef: PropTypes.func,
+  }
+
+  public state = { open: false }
+
+  public renderComponent({  }: IRenderResultConfig<IPortalProps>): React.ReactNode {
```
bmdalex: just remove the args of `renderComponent ` if you're not using them

bmdalex (changes requested):

bmdalex:
@sophieH29 pls don't merge this anyway before I come up with a final solution for opening the Popup using `react-popper`; also, I'm in the process of adding UTs for the Portal in #144 and Popup in #150 

after discussing briefly with @mnajdova we feel that it'd be better to merge #144 and #150 to have a working v1 for Portal and Popup and then adjust this proposal and add it on top as first iteration provided it'g going to be accepted;

thoughts, @kuzhelov , @levithomason ?

src/components/Popup/Popup.tsx
sophieH29: yep, remember that

sophieH29:
@Bugaa92 I haven't planned to merge it. Just want to hear more people's opinions on it and to see if this approach is worth doing.

src/lib/PortalGeneric.tsx
sophieH29: I was also thinking about that and didn't come up with a solid use case. Because it's more like toggling, and a user might want to keep a focus on a trigger. If user expects different behavior, it is possible to call open()/close() outside.
Moreover, afterRenderCb might be different for open() and close() methods.

src/lib/PortalGeneric.tsx
sophieH29: agree, missed that

src/lib/PortalGeneric.tsx
sophieH29: right, it can be private. It is called from AutoControlledComponent
```js
 const state = _.invoke(this, 'getInitialAutoControlledState', this.props) || {}
```

src/components/Portal/Portal.tsx
sophieH29: True! Will remove it

levithomason:
Recapping conversation from this morning for others:

### Composition vs Inheritance
We want to stick with React's recommendation to prefer composition over inheritance. 
 See: https://reactjs.org/docs/composition-vs-inheritance.html.  In addition to the reasoning given in React's docs, this will also ensure that the features of our components can be used by consumers whether they are building their own components or composing 3rd party components.

### Managing State
We also discussed how components should manage state, such as `open` in a Portal.  Currently, we've adopted AutoControlledComponent from Semantic UI React in some of our components.  Though this was done, it is a short term solution and not a pattern we want to propagate.

The solution we want is for Stardust to provide vanilla JS state managers for stateful components.  These state managers would be wired into React's lifecycle and state management mechanisms.  More real world work and experimenting is needed around this idea, but you can reference these places to get the idea:

- React Docs: https://reactjs.org/docs/lifting-state-up.html
- State machines with React https://css-tricks.com/robust-react-user-interfaces-with-finite-state-machines/
- XState https://github.com/davidkpiano/xstate
- App showing interop of state charts and working state machines https://sketch.systems/
- My Codesandbox prototype integrating some of these ideas: https://codesandbox.io/s/jzjpxwjp4y

These links and ideas give a basis for what we'd like to do with state.  We want to provide vanilla JS state managers for common UI that can be consumed and composed by any framework (or no framework).

***

Leaving this PR open for now as I'm not sure if there are other ideas or improvements, apart from the class extension, that we can incorporate.

levithomason (changes requested):
See above comment.  Let's not merge class extension as a pattern.  Any other good work here we can consider salvaging.

sophieH29:
@levithomason thank you for the recap and sharing your thoughts on it! Will check those references.

sophieH29:
Not applicable any more

PR #181, opened 2018-09-04
feat(docs): adding accessibility description for behaviors

kolaps33:
# Adding description for behaviors

- moving description form component to the particular behavior
- adding new description for behaviors (which been missing)
- adding link on the component page pointing into the default description 
- goal of this PR should have uniform description in some 'regex' way. Descr. should be reused in generation for unit tests, then:
https://github.com/stardust-ui/react/compare/feat/365417-behavior-unit-test

image

image


docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -1,27 +1,84 @@
...
+}
+
 class ComponentDocTag extends React.Component<any, any> {
   getTagDescription = (forTag, fromInfo) => {
     const tags = (fromInfo.docblock && fromInfo.docblock.tags) || []
     return _.result(_.find(tags, 'title', forTag), 'description')
   }
 
+  getBehaviorURLandName = fromInfo => {
+    const defaultBeharior = {
```
alinais: typo

jurokapsiar: where?

mnajdova: defaultBeharior -> defaultBehavior

src/lib/accessibility/Behaviors/Button/ButtonBehavior.ts
```diff
@@ -1,5 +1,11 @@
 import { Accessibility } from '../../interfaces'
 
+/**
+ * @description
+ *  Adds role='button' if element type is other than 'button'. This allows the screen readers handle component as button.
+ *  Adds attribute 'aria-disabled=true' based on the property 'disabled'.
```
mnajdova: I would change ' This allows the screen readers handle component as button.' to 'This allows the screen readers to handle the component as button.'

src/lib/accessibility/Behaviors/Button/ToggleButtonBehavior.ts
```diff
@@ -1,5 +1,12 @@
 import { Accessibility } from '../../interfaces'
 
+/**
+ * @description
+ * Adds role='button' if element type is other than 'button'. This allows the screen readers handle component as button.
+ * Adds attribute 'aria-pressed=true' based on the property 'active'. Based on this, screen readers can recognize state of button, if it is pressed or not.
```
mnajdova: Same as above.

src/lib/accessibility/Behaviors/Menu/MenuItemBehavior.ts
```diff
@@ -1,5 +1,13 @@
 import { Accessibility } from '../../interfaces'
 
+/**
+ * @description
+ *  Adds role 'presentation' on the root element
+ *  Adds role 'menuitem' on anchor element
+ *  Adds attribute 'aria-expanded=true' on anchor element based on "submenuOpened" property
+ *  The behavior is designed for particular structure of menu item. The item consist of root element and anchor inside the root element.
```
mnajdova: Typo: 'The item consists of root element and anchor inside the root element.'

docs/src/components/ComponentDoc/ComponentDocTag.tsx
miroslavstastny: Remove `defaultBehavior` completely. Instead of returning `{name: undefined, url: undefined}` return just `undefined`.

docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -1,27 +1,84 @@
...
+      const behaviorFile = behaviorFiltered[0].defaultFile.replace('.ts', '')
+      return {
+        name: `${behaviorFile}`,
+        url: `${behaviorsFolder}/${behaviorFiltered[0].displayName}#${_.kebabCase(behaviorFile)}`,
+      }
+    }
+    return defaultBeharior
+  }
+
+  getDefaultBehavior(info) {
```
miroslavstastny: Not all code paths return a value.

docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -1,27 +1,84 @@
...
+    const defaultBehavior = this.getBehaviorURLandName(info)
+    if (defaultBehavior && defaultBehavior.name && defaultBehavior.url) {
```
miroslavstastny: If you decide that `getBehaviorURLandName` returns `undefined`, the rest of the condition can be safely removed.

docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -1,27 +1,84 @@
...
+  defaultFile: string
+}
+
 class ComponentDocTag extends React.Component<any, any> {
   getTagDescription = (forTag, fromInfo) => {
     const tags = (fromInfo.docblock && fromInfo.docblock.tags) || []
     return _.result(_.find(tags, 'title', forTag), 'description')
   }
 
+  getBehaviorURLandName = fromInfo => {
```
miroslavstastny: What about adding typings at least for the return value?
```jsx
getBehaviorURLandName = (fromInfo): {name: string, url: string} | undefined => {
  ...
```

docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -1,27 +1,84 @@
...
+  }
+
   render() {
     const { info, tag, title, errorMessage } = this.props
-    const description = this.getTagDescription(tag, info) || (
-      <Message error content={errorMessage} compact={true} />
-    )
+    const accDescription = this.getTagDescription(tag, info)
+    const description =
+      accDescription || this.getBehaviorURLandName(info).name ? (
```
miroslavstastny: You are calling `getBehaviorURLandName()` twice in `render()` (The second call is in `getDefaultBehavior()`)

docs/src/components/ComponentDoc/ComponentDocTag.tsx
miroslavstastny: +nit: change capitalisation to `getBehaviorUrlAndName`?

docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -1,27 +1,92 @@
 import * as React from 'react'
 import * as _ from 'lodash'
 import { Header, Message } from 'semantic-ui-react'
+const behaviorMenuItems = require('docs/src/behaviorMenu')
 
 const headerStyle = {
   whiteSpace: 'pre-line',
 }
 
+type defaultBehavior = {
```
jurokapsiar: interface DefaultBehavior { ...

docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -7,59 +7,67 @@ const headerStyle = {
   whiteSpace: 'pre-line',
 }
 
-type behaviorFiltered = {
+type defaultBehavior = {
```
miroslavstastny: Type names start with capital letter

docs/src/components/ComponentDoc/ComponentDocTag.tsx
miroslavstastny: Not fixed yet

miroslavstastny (approved):

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Adding `behaviors` section to the menu, under the components @kolaps33 ([#119](https://github.com/stardust-ui/react/pull/119))
 - Add `avatar` prop to `Chat.Message` subcomponent @Bugaa92 ([#159](https://github.com/stardust-ui/react/pull/159))
 - add `iconOnly` prop to `Button` @mnajdova ([#182](https://github.com/stardust-ui/react/pull/182))
+- Adding accessibility description for behaviors @kolaps33 ([#181](https://github.com/stardust-ui/react/pull/181))
```
miroslavstastny: nit: This should go to Documentation section :-)

CHANGELOG.md
```diff
@@ -22,11 +22,14 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Features
 - Add `state` to `props` in component styling functions @Bugaa92 ([#173](https://github.com/stardust-ui/react/pull/173))
-- Adding `behaviors` section to the menu, under the components @kolaps33 ([#119](https://github.com/stardust-ui/react/pull/119))
 - Add `avatar` prop to `Chat.Message` subcomponent @Bugaa92 ([#159](https://github.com/stardust-ui/react/pull/159))
 - add `iconOnly` prop to `Button` @mnajdova ([#182](https://github.com/stardust-ui/react/pull/182))
 - Add Label `image` and `imagePosition`, removed `onIconClick` prop  @mnajdova ([#55](https://github.com/stardust-ui/react/pull/55/))
 
+### Documentation
+- Adding `behaviors` section to the menu, under the components @kolaps33 ([#119](https://github.com/stardust-ui/react/pull/119))
```
kuzhelov: for the sake of consistency, lets use present indefinite for these entries (e.g. 'Add `behaviors` section ...')

docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -1,27 +1,80 @@
...
+  url: string
+}
+
 class ComponentDocTag extends React.Component<any, any> {
   getTagDescription = (forTag, fromInfo) => {
     const tags = (fromInfo.docblock && fromInfo.docblock.tags) || []
     return _.result(_.find(tags, 'title', forTag), 'description')
   }
 
+  findDefaultBehaviorInfo = (fromInfo): DefaultBehaviorInfo => {
```
kuzhelov: nit: given `fromInfo` arg name, it seems that name like `fetchDefaultBehaviorInfo` will fit better here. Still, please, feel free to reason by yourself whether it should be changed :)

kuzhelov (changes requested):
when trying `yarn start`, it fails to me. Lets take a look on that tomorrow

kuzhelov:
have checked with the latest branch - there is no `yarn start` issue anymore on my machine.

docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -1,27 +1,82 @@
...
+  url: string
+}
+
 class ComponentDocTag extends React.Component<any, any> {
   getTagDescription = (forTag, fromInfo) => {
     const tags = (fromInfo.docblock && fromInfo.docblock.tags) || []
     return _.result(_.find(tags, 'title', forTag), 'description')
   }
 
+  findDefaultBehaviorInfo = (fetchDefaultBehaviorInfo): DefaultBehaviorInfo => {
```
kuzhelov: the argument should rather be called by noun, `componentInfo`:

rename `fetchDefaultBehaviorInfo` -> `componentInfo`

docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -1,27 +1,82 @@
...
+    const accessibilityProperty = fetchDefaultBehaviorInfo.props.find(
```
kuzhelov: probably, `accessibilityDescription` would be a bit more descriptive name for this `accessibilityProperty` variable

docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -1,27 +1,82 @@
...
+      property => property.name === 'accessibility',
+    )
+    if (!accessibilityProperty) {
+      return undefined
+    }
+
+    const accPropertyFormatted = accessibilityProperty.defaultValue.split('.').pop() + '.ts'
```
kuzhelov: this variable should be better renamed as well, as `accPropertyFormatted` doesn't provide any idea that it contains a file name. Maybe `defaultBehaviorFilename` or `defaultAccBehaviorFilename` would work better (as this is, essentially, what is stored in this variable)?

docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -1,27 +1,82 @@
 import * as React from 'react'
 import * as _ from 'lodash'
 import { Header, Message } from 'semantic-ui-react'
+const behaviorMenuItems = require('docs/src/behaviorMenu')
```
kuzhelov, line 4: what I am worrying about here is the following: currently we are using pretty much generic name for this accessibility concern, like just `behavior`. Next day we are planning to add state management functionality to our components, and there is a chance that this state management aspects will be called as `behaviors` as well, but those won't have anything to share with accessibility. In that case generic `behavior` name would be confusing, as it would have two different meanings.

Won't suggest to change anything now, but this is something that we should keep in mind

docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -1,27 +1,82 @@
...
+  createDefaultBehaviorInfo(behavior: any, variation: any): DefaultBehaviorInfo {
+    const behaviorsFolder = behavior.type + 's'
+    const defaultBehaviorFile = variation.name.replace('.ts', '')
+    return {
+      name: `${defaultBehaviorFile}`,
+      url: `${behaviorsFolder}/${behavior.displayName}#${_.kebabCase(defaultBehaviorFile)}`,
+    }
+  }
+
+  renderDefaultBehavior(info) {
```
kuzhelov: maybe it would be better to make the name of the arg more specific, `info` -> `componentInfo`

docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -1,27 +1,82 @@
...
+
+    return undefined
+  }
+
   render() {
     const { info, tag, title, errorMessage } = this.props
-    const description = this.getTagDescription(tag, info) || (
-      <Message error content={errorMessage} compact={true} />
-    )
+    const accDescription = this.getTagDescription(tag, info)
```
kuzhelov: this is a bit misleading - `tag` value could be anything, not just `accessibility` that we are interest in. Thus, we shouldn't name this variable as `accDescription` - it could be not the case. Also, not sure that we should change the logic of displaying `errorMessage`

docs/src/components/ComponentDoc/ComponentDocTag.tsx
```diff
@@ -1,27 +1,82 @@
...
+    const description =
+      accDescription || this.findDefaultBehaviorInfo(info) ? (
```
kuzhelov: so, we are not using the result of `findDefaultBehaviorInfo `? seems that we should call `renderDefaultBehavior()` here directly - otherwise we are calling the same method twice with no good reason for that

kuzhelov (changes requested):
lets try to start with rename refactoring, and after that try to make the logic of `ComponentDocTag`'s `render()` method a bit more clear and consistent with the previous behavior. Otherwise looks good to me :)

docs/src/components/ComponentDoc/ComponentDocTag.tsx
kolaps33: I haven't introduced the method: getTagDescription(tag, info)
So wouldn't like to change origin code. 
As I understand this method is responsible for get accessibility description from component itself.  The acc description in component(s) still exists and is used. 

I think we should change logic in displaying errorMessage because currently with our change there can appear following states:
-  both, acc description in component and default behavior acc description, exists 
- there is NO acc description in component but default behavior acc description exists
- there is NO default behavior acc description but acc description in component exists

and last state when I think we should display error message is only when:
- there is NO default behavior acc description and as well there is NO acc description in component

docs/src/components/ComponentDoc/ComponentDocTag.tsx
kolaps33: in this part of the code we just checking if there is any description or there should be error message. This part of structure I took from origin code. 

We are using the result in this condition, because if there is no default behavior, then the method findDefaultBehaviorInfo() returns undefined...

About calling method twice. It is correct. Maybe we could save result of the method in some global variable and then when we will call it second time and global variable will exists, then we do not need to call it again. What do you think about?

docs/src/components/ComponentDoc/ComponentDocTag.tsx
kuzhelov: yes, sure - there is no need to change this method. Let me try to assist you with that, will suggest changes to make here (without touching existing functionality). 

In terms of changing logic of displaying `errorMessage` - actually, I would suggest to display it **_anytime `errorMessage` is provided (not falsy)_**. If there are any tweaks needed to be made for that, those should be done for the upper level of abstraction (for the place where `ComponentDocTag` is composed)

kolaps33: when you take look on original code:
```
  render() {
    const { info, tag, title, errorMessage } = this.props
    const description = this.getTagDescription(tag, info) || (
      <Message error content={errorMessage} compact={true} />
    )

    return (
      <Header as="h2" style={headerStyle}>
        <Header.Content>{title}</Header.Content>
        <Header.Subheader>{description}</Header.Subheader>
      </Header>
    )
  }
```

not sure what you ment by 
" I would suggest to display it anytime errorMessage is provided (not falsy)"

because errorMessage is set to const "description" based of "TagDescription" availability...
In my opinion the errorMessage is not provided in the fact,  but "description" and there can be anything...

docs/src/components/DocsBehaviorRoot.tsx
```diff
@@ -25,46 +26,65 @@ class DocsBehaviorRoot extends React.Component<any, any> {
...
       background: '#fff',
       boxShadow: '0 1px 2px #ccc',
-      margin: '10px',
+      margin: '1em 1em 1em 2em',
     }
 
     const commentBox: React.CSSProperties = {
       padding: 5,
     }
+    const topRowStyle = {
```
kuzhelov: it would be better to inline this style as well for now, for the sake of consistency

kuzhelov (approved):

docs/src/components/ComponentDoc/ComponentDocTag.tsx
kolaps33: FYI: @jurokapsiar

jurokapsiar: yes, we can have that discussion and change it later. it might be accessibility behavior for example. I will put it on the agenda for one of the next meetings

PR #183, opened 2018-09-04
feat(icons): add support for styling SVG icons

kuzhelov:
### TODO
- [ ] Update CHANGELOG.md

## Problem

Introduce mechanisms in Stardust to provide/extend set of SVG icons by themes.

## Further problem analysis

Essentially, task of introducing custom SVG icons (by means of themes) could be reduced to solving the following tasks
- add mechanisms for rendering custom SVG node
- add support for styling different parts of rendered SVG

Functionality that aims to solve these tasks should be added as part of core Stardust functionality, and there should be no assumptions made on Stardust's side about specific structure of rendered SVGs or their styling aspects - suggested mechanisms should be flexible enough to meet different needs of consumers (theme authors).

## Proposed solution

- allow each theme to provide collection of named functions: where name corresponds to the name of the icon, and function's responsibility is to provide SVG node, with styles (classes) being provided as function's argument
- conventions about both styling aspects, as well as structure of SVG for each individual icon are entirely theme's concerns
- changes in core Stardust interface
    - extend theme properties to include `svgIcons` object - this maps icon's name to the icon rendering function: `(iconClasses) => SVG as ReactNode`. Please, note that although `iconClasses` is something that is related to `Icon` component, this is an aspect that is entirely controlled by theme - thus, all the styling aspects could be customized by theme

## Use case

Suppose that there is a task for theme developer to introduce custom set of SVG icons. Stardust requires them to be exported in the following form from theme:

```jsx
const icons = {
   'iconName': (classes) => (<svg>...</svg>)
   ....
}

export default theme = {
   ... (componentStyles, componentVariables, fonts, etc) ...
  svgIcons: icons
}
```

With that the question of 'what' should be rendered (for each particular icon) is solved - the only question remains is how it should be styled. For that `classes` object is provided as argument, so that those could be used to style individual parts of the icon's element tree - and theme developer has full control over introducing necessary set of parts and styles for `Icon`:

```js
// iconStyles.js file
export default {
  root: ...
  ...
  filledSvg: ({ props, variables }) => ({
    fill: variables.filledIconColor,
    ...
  }),
  ...
}

// svgIcons.js file
const icons = {
   'filled': (classes) => (<svg className={classes.filledSvg}>...</svg>)
   ....
}
```


src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -128,9 +130,27 @@ const iconStyles = {
     }
   },
 
-  svg: ({ variables: v }): ICSSInJSStyle => ({
-    fill: v.color,
-  }),
+  svg: getSvgStyle('svg'),
```
kuzhelov, line 133: yes-yes, let me guess what you might think about :) Actually, this is rather a 'have to' thing, because currently we don't have any possibility to introduce variable set of style parts, such that, for example, if `props.name` is 'endCall' (a name of icon), there would be only parts in styles that are relevant to that icon (`root`, `phone`, `cross-line`, etc). However, there are couple of things we should consider before, potentially, making this move:
- if set of styling parts will be decided by values from `props`, it will drastically complicate overall picture - one thing that immediately comes to my mind is increased debugging complexity
- problem of hardcoded style parts - the ones that should fit **_all component from provided set_**, actually, is not that serious: generally there are just handful different parts of the icon that should be styled differently at any given moment, and these parts are entirely theme's aspects: there are no restrictions introduced from core components' functionality on how these style parts should be named, organised or consumed

With that being said, I rather see suggested approach as being flexible enough to meet quite broad set of introducing styles for custom set of icons.

types/theme.d.ts
```diff
@@ -242,3 +244,12 @@ export interface IFontFace {
 }
 
 export type FontFaces = IFontFace[]
+
+// ========================================================
+// SVG Icons
+// ========================================================
+
+type Classes = { [iconPart: string]: string }
```
mnajdova, line 252: This seems to be like a general type that can be reused in some other places. Should we change the definition with something like `type Classes = { [componentPart: string]: string }` and place it in the utils section? If not, maybe we should change the name to IconClasses in order to be more specific.

src/themes/teams/index.ts
```diff
@@ -1,15 +1,29 @@
...
 import * as siteVariables from './siteVariables'
 import * as componentVariables from './componentVariables'
 import * as componentStyles from './componentStyles'
 import fontFaces from './fontFaces'
 import staticStyles from './staticStyles'
 
+import { default as svgIconsAndStyles } from './components/Icon/svg'
+import { IconSpec } from './components/Icon/svg/types'
+
+const svgIcons = Object.keys(svgIconsAndStyles as { [iconName: string]: IconSpec }).reduce<
```
mnajdova: Is this worth converting to a type: `{ [iconName: string]: IconSpec }`

src/themes/teams/components/Icon/svg/icons/call.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react'
+import { IconSpec } from '../types'
+
+export default {
+  icon: classes => (
+    <svg viewBox="0 0 32 32" role="presentation" focusable="false" className={classes.svg}>
```
mnajdova, line 6: Should we pass the classes to the svg element here?

src/themes/teams/components/Icon/svg/icons/callControlPresentNew.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react'
+import { IconSpec } from '../types'
+
+export default (classes => (
+  <svg viewBox="0 0 32 32" role="presentation">
+    <path
+      d="M22.5,9h-13C8.672852,9,8,9.672852,8,10.5v9C8,20.327148,8.672852,21,9.5,21H13c0.276367,0,0.5-0.223633,0.5-0.5
```
mnajdova, line 7: Should we pass the classes to the svg element and the paths here? Is this where the path1, path2 etc. objects from the iconStyles will be used?

src/themes/teams/components/Icon/iconStyles.ts
mnajdova: Should we add at least first add as many paths properties here as there is in our svg icons so far, and add increase the number of them accordingly if we are adding icons with multiple path parts? Maybe we can generate the paths properties in the begging of the file in order to avoid them being added like this?

kuzhelov: > Maybe we can generate the paths properties in the begging of the file in order to avoid them being added like this?

yes, but this leads to exact problem I've described above - where the solution is to somehow allow variable amount of component parts being declared for styles (ideally, this number should be dependent on props). While it is an option, there are some drawbacks associated with it - have mentioned them above.

> Should we add at least first as many paths properties here as there is in our svg icons so far, and add increase the number of them accordingly if we are adding icons with multiple path parts

we may consider to do this, yes

types/theme.d.ts
kuzhelov: yes, thanks - let me make this move if we'll agree on the general approach

src/themes/teams/components/Icon/svg/icons/call.tsx
kuzhelov: yes - but this is intentionally left to icons' maintainers, as current system (where both CSS class selectors and DOM selectors are used for styling) should be refactored

src/themes/teams/components/Icon/iconStyles.ts
mnajdova: If we have just as many paths properties as our svg icons actually have, I am okay with them being hardcoded, I see your point. With this, we will be aware about what each of those represent.

src/themes/teams/components/Icon/svg/icons/callControlPresentNew.tsx
kuzhelov: >  Is this where the path1, path2 etc. objects from the iconStyles will be used?

yes, exactly. I would leave this work to icons' maintainers, as well as making final decision on how styling parts should be named for icons. Once we'll agree that provided mechanisms are sufficient enough to meet their needs, the task will reduce only to styling aspects that I would rather prefer to be addressed by design folks

jurokapsiar:
Heads up for the future: For RTL we will need to be able to flip certain icons, for example pointing arrows. It should be possible to do this in a generic way (applying css transformations on the svg). It is important that approach proposed in this PR will allow us to do that in the future.

kuzhelov:
@jurokapsiar, this is a very good point. With provided functionality we have couple of ways to address that, essentially
- we could address it on the core level, by introducing general 'RTL transform' functionality to `Icon` component. This approach will mimic the one that is currently used in Teams, where class name is added to root SVG element to make this transform
- or we could pass `rtl` value to SVG function (so it will be `(classes, rtl) => React.ReactNode`). Although this approach ptovides more control over rendered SVG tree, I think that at early stages we should limit amount of data passed to icon's rendering function, so that overall picture will be simplier

src/lib/mergeThemes.ts
```diff
@@ -159,6 +160,9 @@ const mergeThemes = (...themes: IThemeInput[]): IThemePrepared => {
 
     acc.componentStyles = mergeThemeStyles(acc.componentStyles, next.componentStyles)
 
+    // Merge icons set, last one wins in case of collisions
+    acc.svgIcons = { ...acc.svgIcons, ...(next.svgIcons || {}) }
```
miroslavstastny: UTs?

src/themes/teams/components/Icon/svg/icons/index.ts
```diff
@@ -0,0 +1,17 @@
+export { default as call } from './call'
+export { default as callEnd } from './callEnd'
+
+export { default as callVideo } from './callIncomingVideo'
```
miroslavstastny: What's the reason for exporting with a name different from filename?

src/lib/mergeThemes.ts
kuzhelov: sure, will introduce them once we will agree on the approach

src/themes/teams/components/Icon/svg/icons/index.ts
kuzhelov: the reason was to save mapping from the icon that is provided (with simple name) to the one that was originally used by Teams (longer name) - with that I thought it will be easier to track which one corresponds to which at the initial stages of adoption. Absolutely ok to make any renaming that will suit Teams theme devs

types/theme.d.ts
levithomason: Note, this type exists in this file already:

https://github.com/stardust-ui/react/blob/e460060d318cbe3018a48f4625dfc9b263625b12/types/theme.d.ts#L107

kuzhelov:
closing as this set of changes will be introduced by means of #260

Issue #184, opened 2018-09-05
[RFC] Typescript type checking options


miroslavstastny:
Stardust uses Typescript, but all it's [type checking options](https://www.typescriptlang.org/docs/handbook/compiler-options.html) are disabled, which kills a lot of Typescript's power.
Enabling (some of) the checks would make the code better and it'd easier to enable them in the early stage of the project.

I went through the list of compiler options which could be worth considering and also checked tsconfig in other projects which use Typescript.

option | project A | project B | [ionic/react](https://github.com/ionic-team/ionic/blob/master/react/tsconfig.json) | [vscode/src](https://github.com/Microsoft/vscode/blob/master/src/tsconfig.json) | [rxjs](https://github.com/ReactiveX/rxjs/blob/master/tsconfig.base.json) | description
-- | -- | -- | -- | -- | -- | --
`noImplicitReturns` |   | true | true | true | true | Report error   when not all code paths in function return a value.
`noImplicitAny` (1) |   | true | true |   | true | Raise error on   expressions and declarations with an implied `any` type.
`noImplicitThis` (1) |   |   |   | true | true | Raise error   on `this` expressions   with an implied `any` type.
`noUnusedParameters` |   | true | true |   |   | Report errors   on unused parameters.
`noUnusedLocals` |   | true | true | true |   | Report errors   on unused locals.
`alwaysStrict` (1) | true |   |   | true |   | Parse in strict mode and emit `"use   strict"` for each source file
`strictNullChecks` (1) |   |   |   |   |   | In strict null   checking mode, the `null` and `undefined` values are not in   the domain of every type and are only assignable to themselves and `any` (the one exception being   that `undefined` is   also assignable to `void`).
`strictFunctionTypes` (1) |   |   |   |   | true | Disable   bivariant parameter checking for function types.
`strictPropertyInitialization`   (1) |   |   |   |   |   | Ensure   non-undefined class properties are initialized in the constructor. This   option requires `--strictNullChecks` be enabled in order to take effect.

(1) part of `--strict`
Default value for all options is `false`.

Can we think of what would make sense to be enabled?

kuzhelov:
There was a TS compilation problem caused by `--strict` flag which has enabled all these checks for our project (#158) - as a result, we've had to (temporarily) disable it.

[...]

PR #185, opened 2018-09-05
chore: Enable noImplicitReturns TS check

miroslavstastny:
Enables `noImplicitReturns` in tsconfig.json.
See #184.


docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -526,7 +526,7 @@ class ComponentExample extends React.PureComponent<IComponentExampleProps, IComp
 
   private renderVariables = () => {
     const { showVariables } = this.state
-    if (!showVariables) return
```
kuzhelov, line 529: great catch 👍 although nothing has changed semantically, for the cases where, say `boolean` is a return value, it will make crucial difference

kuzhelov (approved):

alinais (approved):

PR #187, opened 2018-09-05
Chore: Fix gulp watch on Windows

### LINKED ISSUE #172: Docsite: watch issues
## gulp watch does not pick changes on Windows

1. On Windows, run `yarn start`
2. Edit component source
3. `build:docs:docgen` should be triggered, but nothing happens

The reason is that the watch path contains backslashes as `path.resolve()`, which is used to construct it, uses backslashes on Windows.
According to [documentation](https://github.com/isaacs/node-glob/blob/master/README.md#windows) glob pattern must contain forward slashes only.

## Incremental build breaks examples menu
For each component single JSON file with a list of examples is built from multiple `index.tsx` files.
When one of the files is changed, watch triggers incremental build for the single changed file, and the resulting JSON is overwritten with data from that single file only.
(Have types and variants examples, update types' index, all variants are now missing in example menu).
This can be solved by adding [`gulp-remember`](https://github.com/ahaurw01/gulp-remember) to the watch pipe.

miroslavstastny:
This fixes #172:
- glob filter for `gulp watch` always uses forward slashes regardless the OS,
- `build:docs:example-menu` remembers files, so it is run on all files when a file changes.

While the [`gulp.watch`](https://github.com/gulpjs/gulp/blob/master/docs/API.md#gulpwatchglobs-opts-fn) documentation clearly states that [forward slashes must be used](https://github.com/isaacs/node-glob#windows), there is no such statement for [`gulp.src`](https://github.com/gulpjs/gulp/blob/master/docs/API.md#gulpsrcglobs-options). For now it seems that gulp.src works even with backslashes. I am not sure whether we want to change it there as well or not (make is consistent vs. don't fix anything that's not broken).


config.ts
```diff
@@ -36,8 +37,11 @@ const paths = {
   docsDist: base.bind(null, envConfig.dir_docs_dist),
   docsSrc: base.bind(null, envConfig.dir_docs_src),
   umdDist: base.bind(null, envConfig.dir_umd_dist),
+  posix: undefined, // all the sibling values, but with forward slashes regardless the OS
 }
 
+paths.posix = _.mapValues(paths, func => (...args) => func(...args).replace(/\\/g, '/'))
```
kuzhelov, line 43: seems that we should consider it to be a mandatory preprocessing method for provided values, so that for `paths` we would have POSIX ones by default:

```
const paths = posix({
   src: ...,
   dist: ...,
   ...
})

// and then
paths.src()    // returns POSIX variant by default

// and, if it will be necessary, we can do the following (this will be about extending functionality of 'posix()')
paths.original.src()
```

The reason I see it being a better approach is that we will be provided with a safe defaults as a result - in contrast to the approach that is currently suggested. What do you think?

build/gulp/tasks/docs.ts
```diff
@@ -87,6 +91,7 @@ task('build:docs:component-menu-behaviors', () =>
 
 task('build:docs:example-menu', () =>
   src(examplesSrc, { since: lastRun('build:docs:example-menu') })
+    .pipe(remember('example-menu')) // FIXME: with watch this unnecessarily processes index files for all examples
```
kuzhelov, line 94: another problem that I might see here is that if file will be removed, it will still remain to be part of the stream - and this will result in non-relevant produced result.

miroslavstastny: https://github.com/stardust-ui/react/pull/187/files#diff-d0223606702a9bcfa799433ee11ed0aeR26

kuzhelov: for some reason this 'unlink' event doesn't fire on my machine - seems to be an issue with 'chokidar' lib (while for 'change' events everything works). This was, essentially, the reason to ask - thanks!

kuzhelov (approved):

PR #204, opened 2018-09-06
fix: teams theme font sizes

levithomason:
The Teams theme uses an html font size of 10px and scales the body by `1.4rem` to achieve a 14px body size.

This PR updates the html and body font size values in Teams' theme.  It also scales the doc site Provider values to maintain Semantic UI's 14px html font size need.


mnajdova (approved):

kuzhelov (approved):

levithomason:
Hey guys, for safety let's let authors call for reviews and merge their own PRs.  I was still working on this PR and it was failing screener for valid reasons.

Merge themes is not respecting static styles nor font faces.  The docs are not using the correct font faces or static styles after this merge.  I was adding fixes for that before merge / review.

I will push the fixes in a separate PR to keep things moving forward.

Issue #136, opened 2018-08-24
[RFC]theming: Theme-provided or theme-restricted Enumeration?


bcalvery:
## Theme-provided or theme-restricted Enumeration?
How can a theme **restrict usage of possible variables**, or **provide a custom range of variables**. And how can other theme-specific information be provided in a more abstract/customizable/flexible way?

### Problem
Right now there are multiple _enums_ which are defined in _src_. Specifically there are (for the Text component) _Sizes_ and _Weights_. I can easily imagine there will soon be additional application-specific enums, such as _AvatarSizes_, _PresenceStates_ and other components as well.

**The basic questions are:** What happens when the design team providing a theme wants to limit the number of Text sizes available, or wants to "outlaw" certain font weights? How do they do that? What happens if a size/weight becomes deprecated? What about making various combinations of sizes/weights allowed or disallowed?

**Related question:** What about things that are somewhat more application-specific, like in the case of Avatar, _AvatarSizes_ and _PresenceStates_?  Would it be possible to make the theme (and business logic) provide the data for those? For _AvatarSizes_ Stardust might say "we need S, M, L and XL avatar sizes" and it would be nice if the theme definition could say "M is not allowed, don't let M be used." But what about PresenceStates? _Presence_ seems very specific, the idea of "Available" or "Away" is pretty heavily tied to a chat app... What if some other product that uses Stardust has representations of people and wants to use Avatar, but has _very_ different ideas of what an avatar's "state" could be? If we've hard-coded "presence" enum to be specific to our application (available, away, etc) that doesn't seem to be a very open/flexible system.

**Another related question:** Would it be possible to streamline/simplify the amount of redundancy between the siteVariables and the multiple component-specific files? Using Text as an example again: siteVariables feeds into textVariables, which feeds into textStyles files? Having a more systematic way to provide/consume the styles might streamline the existing approach.
![image](https://user-images.githubusercontent.com/15043460/44608181-bb62ee00-a7a7-11e8-9986-38e399b50413.png)

### Possible solution:
Could an interface or json object format be a solution for this? Stardust could have an ITextStyle interface, and the theme could provide text style objects that conform to the interface. One of the properties the interface requires could be a "valid" or "deprecated" Boolean. Perhaps another "object" that the theme could provide is a "map to" which would tell stardust what sizes/weights are valid and/or what sizes/weights to substitute?

[...]

PR #205, opened 2018-09-07
fix: text component docs

codepretty:
# Updating the text doc to have all examples & have accessibility info
- all text types, states, and variations now have shorthand & children examples
- fixed some cases where children/shorthand were reversed

**Before**
image

**After**
image

### DIFF
--- a/docs/src/examples/components/Text/Types/TextSizesExample.shorthand.tsx
+++ b/docs/src/examples/components/Text/Types/TextSizesExample.shorthand.tsx
@@ -3,21 +3,21 @@ import { Text } from '@stardust-ui/react'
 
 const TextSizesExampleShorthand = () => (
   <div>
-    <Text size="xs">Dicta voluptatum dolorem.</Text>
+    <Text size="xs" content="Dicta voluptatum dolorem." />
     <br />
-    <Text size="sm">Dicta voluptatum dolorem.</Text>
+    <Text size="sm" content="Dicta voluptatum dolorem." />
     <br />
-    <Text size="md">Dicta voluptatum dolorem.</Text>
+    <Text size="md" content="Dicta voluptatum dolorem." />
     <br />
-    <Text size="lg">Dicta voluptatum dolorem.</Text>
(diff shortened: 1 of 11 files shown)


miroslavstastny (approved):

docs/src/examples/components/Text/Types/TextSizesExample.shorthand.tsx
```diff
@@ -3,21 +3,21 @@ import { Text } from '@stardust-ui/react'
 
 const TextSizesExampleShorthand = () => (
   <div>
-    <Text size="xs">Dicta voluptatum dolorem.</Text>
+    <Text size="xs" content="Dicta voluptatum dolorem." />
```
kuzhelov, line 6: 👍

docs/src/examples/components/Text/Types/TextSizesExample.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { Text } from '@stardust-ui/react'
+
+const TextSizesExample = () => (
+  <div>
+    <Text size="xs">Dicta voluptatum dolorem.</Text>
```
kuzhelov, line 6: actually, this raises another point related to the values used for sizes. The problem is that it is hard to intuitively reason for the client about, say, which one is bigger: `xl` or `2x`. Would propose to use values that are more intuitive in this sense (and address the issue by means of separate PR). 

Quite similar sentiments are expressed in #136.

kuzhelov (changes requested):
`changelog.md` should be updated as this PR introduces publicly visible changes

kuzhelov (approved):

docs/src/examples/components/Text/Types/TextSizesExample.tsx
codepretty: Completely agree. My next PR is going to use a theme-based enum to show the appropriate sizes. 

What I'm still not sure about is what to do for naming and how that could vary among themes...

PR #206, opened 2018-09-07
feat(Radio): added value, disabled, name, checked, defaultChecked and onChange props

mnajdova:
# Radio

This PR introduces the following props to the Radio component: value, disabled, checked, and onChange. 

### TODO

- [x] Conformance test
- [ ] Stardust base theme
- [x] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [x] Confirm RTL usage
- [x] Update the CHANGELOG.md

# API Proposal

## value

The HTML input value.

## disabled

A radio can appear disabled and be unable to change states.

```jsx
<Radio label="Disabled" disabled />
```

```html
<div class="ui-radio a b">
  <label class="ui-label j e f k l m i n o p d g h">
    <input type="radio" disabled="" class="c"/>Disabled
  </label>
</div>
```

## onChange

Callback invoked when the radio is clicked.

## checked
Whether or not the radio is checked.

```jsx
 <Radio label="This radio comes pre-checked" checked />
```

```html
<div class="ui-radio a b">
  <label class="ui-label j e f k l m i n o p d g h">
    <input type="radio" checked="" class="c"/>This radio comes pre-checked
  </label>
</div>
```


docs/src/examples/components/Radio/States/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+const States = () => (
+  <ExampleSection title="States">
+    <ComponentExample
+      title="Checked"
+      description="A radio can come pre-checked."
+      examplePath="components/Radio/States/RadioExampleChecked"
+    >
+      <Segment style={{ marginTop: '10px' }}>
+        Use{' '}
```
mnajdova, line 14: I stole this from SUIR, I didn't know we can do this. Very cool! :D

kuzhelov:
!! now with `variables` being not a flat object (in general case) we should ensure that those will be properly merged by theme providers

jurokapsiar:
Radio group is necessary for accessibility purposes.
It can render just as a div. That will allow us to apply correct aria role and add keyboard handling.
See https://www.w3.org/TR/2017/WD-wai-aria-practices-1.1-20170628/examples/radio/radio-1/radio-1.html

mnajdova:
@jurokapsiar so according to the post, we should render the RadioGroup and the Radio as div elements with the proper roles. Should I actually replace the <input type='radio' /> implementation with <div role='radio' /> and style the ::before part to look like a radio button as part of this PR? If this is the final decision about how it should be defined, I would rather do this now then later.

levithomason:
We'll pause this work while we first style the radio.  This will require an icon or pseudo elements and some custom event handling.  Let's make sure that works before designing this next phase, groups.

mnajdova:
As per our call, the RadioGroup is stopped for now, but there is an example added of how the buttons can be used as a group. The styles are also updated: 

image

jurokapsiar:
looking into accessibility, but it will take more time to come up with a working solution if we do not want to use default input as is.

looks like the missing thing is handling of the focused state.
let's try to use focused state introduced in https://github.com/stardust-ui/react/pull/213

If not possible, you can take a look at https://www.w3.org/TR/wai-aria-practices/examples/radio/radio-1/radio-1.html which unfortunately does not work correctly with JAWS screen reader, but is a better starting point.

docs/src/examples/components/Radio/Types/RadioExample.shorthand.tsx
```diff
@@ -1,6 +1,10 @@
 import React from 'react'
 import { Radio } from '@stardust-ui/react'
 
-const RadioExample = () => <Radio label="Make your choice" />
+const handleChange = () => {
+  console.log('The radio checked value was changed!')
```
kuzhelov: probably, it is better to use `alert` calls here, if the intent is to make it clear to the client that some changes have happened

src/components/Button/ButtonGroup.tsx
```diff
@@ -30,7 +30,7 @@ class ButtonGroup extends UIComponent<Extendable<IButtonGroupProps>, any> {
     /** An element type to render as (string or function). */
     as: customPropTypes.as,
 
-    /** A button can take the width of its container. */
+    /** The buttons contained inside the ButtonGroup. */
```
kuzhelov, line 33: +1 👍

src/components/Radio/Radio.tsx
```diff
@@ -1,66 +1,157 @@
...
+    checked: PropTypes.bool,
+
     /** Child content * */
     children: PropTypes.node,
 
     /** Additional classes. */
     className: PropTypes.string,
 
+    /** Initial checked value. */
+    defaultChecked: PropTypes.bool,
```
kuzhelov, line 62: this one is not used anywhere - should we delete it?

kuzhelov: also, it seems that this one is rather pertains to `RadioGroup` concept, rather than to individual `Radio`

docs/src/examples/components/Radio/Types/RadioExample.shorthand.tsx
mnajdova: Makes sense, will change it!

src/components/Radio/Radio.tsx
mnajdova: Every auto-controlled prop must have a default prop associated to it, that is why we have it. I may need to provide example usage of it.

mnajdova:
The current implementation of the Radio and the example of the group regarding accessibility are behaving the same as the https://www.w3.org/TR/wai-aria-practices/examples/radio/radio-1/radio-1.html The next thing that should be implemented is having the focused border around the buttons. Should we immediately check the button if it is focused, or should be have the same implementation as in the W3 example? What's the proposal for further checks/fixes?

src/components/Radio/Radio.tsx
```diff
@@ -1,66 +1,157 @@
...
+    const { onChange, disabled } = this.props
+    const { checked } = this.state
+
+    if (disabled) {
+      e.preventDefault()
+      return
+    }
+
+    if (onChange) {
+      onChange(e, { ...this.props, checked: !checked })
```
kuzhelov: worrying about the fact that this callback method is invoked before state change will be applied to component - in case of potential error in the logic provided by the callback state change won't be applied

src/components/Radio/Radio.tsx
```diff
@@ -1,66 +1,157 @@
...
     type: 'radio',
   }
 
-  renderComponent({ ElementType, classes, rest, styles }) {
-    const { type, label } = this.props
+  static autoControlledProps = ['checked']
+
+  private handleChange = (e: React.SyntheticEvent) => {
+    const { onChange, disabled } = this.props
+    const { checked } = this.state
```
kuzhelov: it would be more robust to obtain this flag directly from `e.target.checked` - due to async state updates there is a room (although very small one) for the `state` being out of sync with the component's visual look

kuzhelov (approved):
couple of comments that we could discuss/address, other than that everything is 👍

src/components/Radio/Radio.tsx
mnajdova: Agreed, will move it before invoking the callback.

src/components/Radio/Radio.tsx
mnajdova: Makes sense. Will change it!

src/components/Radio/Radio.tsx
mnajdova: @kuzhelov , I added the usage of the default checked value. This value is intended to be used as a initial check value of the radio button. As the component is auto-controlled, the user should be able to initially set the value of the checked property. Please comment if you have some more doubts on this.

mnajdova:
As the accessibility aspect of the radio group are pretty complicated, this PR will address ONLY the properties that needs to be implemented in the Radio, and then a dedicated PR will be created for the RadioGroup.

src/components/Radio/Radio.tsx
```diff
@@ -1,66 +1,157 @@
...
+    const checked = (e.target as HTMLInputElement).checked
+
+    if (disabled) {
+      e.preventDefault()
+      return
+    }
+
+    this.trySetState({ checked })
+
+    if (onChange) {
```
miroslavstastny: `_.invoke()` does the check for you

src/components/Radio/Radio.tsx
```diff
@@ -1,66 +1,157 @@
...
     return (
       <ElementType {...rest} className={classes.root}>
-        <Label styles={{ root: styles.label }}>
+        <Label as="label" styles={{ root: styles.label }}>
+          {Icon.create(icon || '', {
+            defaultProps: {
+              circular: true,
+              size: 'mini',
+              variables: {
+                color: checked ? variables.checkedIconColor : variables.uncheckedIconColor,
```
miroslavstastny: I see you are already discussing this in #207. It would be great if, eventually, we would not need to touch variables in component code.

kuzhelov: unfortunately, this makes 'variables defined in theme' story a bit more complicated, as now we might need to get an access to component props there - as it is discussed in #207. This would be quite significant move, and we should consider whether there are any other not-so-radical alternatives. Still, totally agree that we should finally come up with solution that will eliminate this code from component's logic

src/components/Radio/Radio.tsx
kuzhelov: ok - but in that case let me share what my concerns are about. Now we have this `defaultChecked` to be defined in prop types, but there is no such a property in the component's TS interface - with that it, essentially, means that me as a client is not able to safely specify value for it in the TS code (I mean, this prop won't be advertised as one that is supported by the component, so I could provide some `string` value, for example).

src/components/Radio/Radio.tsx
mnajdova: Exactly, I was not satisfied with having to use the variable in the renderComponent, but as I mention in the other PR we should either be able to define the variables depending on the state and props, or rethink about the API provided so far for the components.

src/components/Radio/Radio.tsx
mnajdova: Sorry, that is my fault, the defaultChecked prop **SHOULD** be provide in the TS interface. Thanks for the catch!

src/components/Radio/Radio.tsx
mnajdova: I replaced this with _.invoke, but I prefer much this type of invoking because I know for sure when clicking the property that it is used and invoke somewhere in the code, and with using the invoke, the name of the callback is just a string.

miroslavstastny (approved):

PR #207, opened 2018-09-07
feat: ensure variables and styles are passed to component parts

### LINKED ISSUE #162: [RFC] Decouple variables passed to child components from parent component's logic

kuzhelov:
### TODO
- [x] ensure variable objects (not flat in general) are properly merged, introduce tests for that
- [x] update `CHANGELOG.md`

------

This PR aims to ensure that both `styles` and `variables` can be passed to component's parts (and, as a result, close #162 - as this relationship will become to be part of theming concerns, see https://github.com/stardust-ui/react/pull/207/files#diff-b9cf59e13107dcf58e9567148a6c8367L125)

As an example - if `Avatar` component defines `icon` part in its anatomy, the following things should be possible:

- for Avatar variables module (`avatarVariables.js`)
```js
const avatarColor = '...'

return {
  color: avatarColor,
  ...
  icon: {
    color: avatarColor // define the same color to be used by 'icon'
   ...
}
```

- for `styles` provided as props
```jsx
<Avatar styles={{ icon: { color: 'red' } }} ... />
```

- for `variables` provided as props
```jsx
// value for icon's 'color' is provided
<Avatar variables={{ icon: { color: 'red' } }} ... />
```

For both of these cases the name that corresponds to component's part name (i.e. `icon`) should be used.

**Unit tests are introduced to ensure this behavior.**


test/specs/commonTests/implementsShorthandProp.tsx
```diff
@@ -48,6 +48,34 @@ export default Component => {
 
         expect(allShorthandPropertiesArePassedToShorthandComponent).toBe(true)
       })
+
+      test(`shorthand's styles may be passed as '${shorthandPropertyName}' prop of ${
+        Component.displayName
+      }'s styles`, () => {
+        const props = { [shorthandPropertyName]: 'some value' }
```
mnajdova, line 55: These are great changes @kuzhelov. Should we introduce similar changes to the implementsCollectionShorthandProp test?

src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -1,13 +1,20 @@
...
-  vars.backgroundColor = 'rgb(232, 232, 232)'
-  vars.startPaddingLeft = '0px'
-  vars.endPaddingRight = '0px'
-  vars.height = pxToRem(20)
-  return vars
+  const color = 'rgba(0, 0, 0, 0.6)'
+
+  return {
+    circularRadius: pxToRem(9999),
+    padding: `0 ${pxToRem(4)} 0 ${pxToRem(4)}`,
```
mnajdova, line 8: It feels kind a awkward not having the root with with the variables for the component. I know it is complicating things, but with that we would be consistent with the styles.

kuzhelov: actually, this is the thing that we explicitly agreed upon - the fact that `variables` should be plain object. The alternative approach (of introducing `root`) would make things much more complex. Would rather vote to leave things as they are at the moment, as it seems that there is no inconsistency for `variables` and `styles` exposed to the consumer.

test/specs/commonTests/implementsShorthandProp.tsx
kuzhelov: oh, great suggestion - let me take a look on that..

kuzhelov:
!! now with `variables` being not a flat object (in general case) we should ensure that those will be properly merged by theme providers

test/specs/lib/mergeThemes/mergeComponentVariables-test.ts
```diff
@@ -17,17 +17,17 @@ describe('mergeComponentVariables', () => {
   })
 
   test('gracefully handles null and undefined', () => {
-    expect(() => mergeComponentVariables({ color: 'black' }, null)).not.toThrow()
-    expect(() => mergeComponentVariables({ color: 'black' }, { color: null })).not.toThrow()
+    expect(mergeComponentVariables({ color: 'black' }, null)).not.toThrow()
```
kuzhelov, line 20: previously function hasn't been called, thus no condition has been verified

mnajdova:
Couple of things we should think about before merging this PR.

1. Merging of the variables when they are used in the parent component and in the shorthand at the same time. For example if we have the following:
```jsx
<Button variables={icon: {color: 'red'}} icon={..., variables: {color: 'blue', backgroundColor: 'yellow'}} />
```
What would the outcome of this be? I would expect the variables inside the icon's shorthand will have precedence so we would end up with: `variables:  {color: 'blue', backgroundColor: 'yellow'}`

2. The second point is, how can we send different variables to the child component if they are dependent on some props/state. For example, in the Radio button we have an icon representing the radio, which should propagate different variables: color, backgroundColor, borderColor dependent on the state's checked variable. Here is the code:
```jsx
 renderComponent({ ElementType, classes, rest, styles, variables }) {
    const { type, label, ... } = this.props
    const { checked } = this.state
    return (
          ...
          {Icon.create(icon || '', {
            defaultProps: {
              circular: true,
              size: 'mini',
              variables: {
                color: checked ? variables.checkedIconColor : variables.uncheckedIconColor,
                backgroundColor: checked
                  ? variables.checkedIconBackgroundColor
                  : variables.uncheckedIconBackgroundColor,
                borderColor: checked
                  ? variables.checkedIconBorderColor
                  : variables.uncheckedIconBorderColor,
              },
              styles: { root: styles.icon },
            },
          })}
          ...
    )
}
       
```
Should the variables as a function receive props/state in similar variable as the styles are, or did we made mistake in the design of the API for the Icon regarding defining this as variables and not properties? Is this something that we can even avoid?

kuzhelov:
see the second point as the most appealing one to be solved before this PR will be merged - as, otherwise, we won't be able to solve the problem of using variables in component's logic, the problem this PR was originally designed for

kuzhelov:
One of the options we might have is to introduce props as an argument to component variables functions. 

Actually, if we would think about the problem, essentially, whatever move we would consider, in either case we would need to obtain information about component's state somehow - and if we would like this state to define rather abstract styling concepts for component (like general `color` that might not directly correspond to CSS property's value), we would inevitably need to introduce this knowledge to variables' functions.

```jsx
 renderComponent({ ElementType, classes, rest, styles, variables }) {
    const { type, label, ... } = this.props
    const { checked } = this.state
    return (
          ...
          {Icon.create(icon || '', {
            defaultProps: {
              circular: true,
              size: 'mini',
              variables: { variables.icon },
              styles: { root: styles.icon },
            },
          })}
          ...
    )
}

/// radioVariables.js
export default (siteVariables, props) => {
  const checkedIconColor = ...
  const uncheckedIconColor = ...

  const checkedIconBackgroundColor = ...
  const uncheckedIconBackgroundColor = ...
  
  return {
    ...
    icon: {
     color: props.checked ? checkedIconColor : uncheckedIconColor,
     backgroundColor: props.checked
                  ? checkedIconBackgroundColor
                  : uncheckedIconBackgroundColor,
     borderColor: props.checked
                  ? checkedIconBorderColor
                  : uncheckedIconBorderColor
   }
}
```

With minimal changes to current codebase this works - but the question is whether we are ok to make this move from a conceptual point of view. My opinion is that we should at least consider it, because otherwise the whole idea of specifying variables for the child (slot) components **_in theme_** won't be possible for aforementioned reasons (because we need to have an access to component's state in certain cases).

Please, let me know about your thoughts, folks.

/cc: @levithomason, @mnajdova, @miroslavstastny

kuzhelov:
done - `props+state` are now introduced as an arg of variables function. I've intentionally omitted to make a signature of variables function being consistent with the one of styling functions (so that it would be `({ siteVariables, props }) => ...` instead of `(siteVariables, props)`) due to it would require quite significant attention for the sake of not introducing subtle bugs - would rather like to address this in a separate PR devoted to this change exclusively. Also, at that moment we would be pretty sure if introducing `props` is a way to go for us, or it is better alternative.

mnajdova (approved):

kuzhelov:
agreed to address potential concerns related to `collectionShorthandProp` tests by means of separate PR, to speed up the process a bit

levithomason:
I'm not sure we should move forward with this pattern just yet.  It is a big change that has many implications.  I also spent some time discussing it with Fabric yesterday and I think we reached a few good conclusions:

It seems there are only two options here.  
1. **Many variables** - Code combinatorial variables for every component (primaryBackground, primaryDisabledBackground, etc) 
1. **Many Branches** - Code combinatorial function branches that return the proper variable values

Using many variables does have advantages over using functions:
1. Hover, active, focus states can be handled with CSS selectors
1. It is JSON serializable
1. We can show edit controls in the doc experience for variables
1. Designers are able to edit variable values and JSON
1. It is more simple

Using variable functions has a few downsides:
1. Handling hover, active, focus requires JS listeners and state updates
1. The logic is not JSON serializable
1. We cannot give a path to customizing the values, unless we give a full featured code editor
1. Only engineers skilled in JS will be capable of customizing it
1. It is more complex

### Example implementation

Here's how we could use the existing `props` passed to our style function to override the icon style when used in the Radio:

```jsx
const radioStyles = {
  root: ({ props, variables }) => ({
    '.ui-icon': {
      color: props.checked
        ? variables.checkedIconBackgroundColor
        : variables.uncheckedIconBackgroundColor
    }
  })
}
```

Although styles aren't as robust as updating the Icon's variables, it does save us from having to add JS listeners.  In the variable function example, we'd need to pass in the `hover` state of the Radio if we wanted to update the icon when the Radio was being hovered.  That means we'd need mouse enter/leave listeners on the Radio, set state on change, then pass those values into the variable function.  This will not only cause React renders but also re-evaluate the style function every time.  It also means consumers and 3rd parties would need to follow that same pattern when using our variables functions.

Instead, if we use the "many variables" approach, we can just use CSS selectors for hover, active, focus states:

```jsx
const radioStyles = {
  root: ({ props, variables }) => ({
    ':hover': {
      '.ui-icon': {
        color: 'red'
      }
    }
  })
}
```

There are obviously cons to this approach as well, but I'm thinking the main points are:
1. Listeners (variable functions) vs child selectors (many variables) for hover, active, focus
1. Friendly JSON serializable values (many variables) vs JS dev only theme updates (variable functions)

Thoughts?

kuzhelov:
## Objections

Let me, please, directly challenge the examples of styling `Radio`'s icon you've suggested:

```jsx
const radioStyles = {
  root: ({ props, variables }) => ({
    '.ui-icon': {
      color: props.checked
        ? variables.checkedIconBackgroundColor
        : variables.uncheckedIconBackgroundColor
    }
  })
}
```

Unfortunately, this approach is not robust enough - more than that, for some scenarios (including the one you've suggested) it simply won't work. Let's face it - here our intent is to define icon's color, and previously _variables were able to abstract us away from the implementation details around how color is applied to icon_. Now, in contrast, we have a problem: what if it will be an SVG icon? in that case it won't be styled properly, as SVG icons generally use `fill` property (not `color`) for coloring.

In general, in case of parent-child (component-slot) relationship **_variables serve like an interface that introduce robustness in expressing intents from parent to child component_**. Without this interface there are no ways for us to reliably introduce styles and guarantee that on the next day, with implementation approach being changed for child component, styles will remain to work properly.

------
Another question is about `.ui-icon` - what does it mean? Is it about naming property of `classes` object (why not use `icon` as we've done before than), or it is about introducing static CSS styles (this will be applied globally to the page, what we probably won't like to have).

## Proposals
My opinion on that is the following: we should either
- preserve the approach we have now, where `props+state` is passed as an argument to variables function. The reason for that is that this approach solves the problem of styling child components in a robust way - in contrast to any approach that will be about tweaking CSS of child component directly
- if we would like to get **all the pros of `many variables` approach** mentioned above, we should consider to introduce `prop+state` argument to slot's variables only. Although it will increase complexity, it is worth to notice that in general this pattern could be omitted by developer - and used only in cases where needed (and `Radio` is a good example for that)
```jsx
// radioVariables
export default (siteVariables) => ({
  ...
 icon: ({ props }) => ({
   ...
   color: props.checked
        ? checkedIconBackgroundColor
        : uncheckedIconBackgroundColor
  })
})
```

levithomason:
Agree, however, compare the alternative.  I'm afraid there isn't a reasonable alternative here.  If there is, please elaborate as I'd like to not have to go this path.

### Child Icon _styles_ is a _challenge_

The challenge of handling child style implementation details exists but is not a blocker by any means.  It is already required to be solved in the theme whether or not we use child selectors.

The knowledge of the icon style implementation is already owned by the Teams theme in the Icon styles.  The Icon style itself must handle all the correct cases of the icon and correctly apply variables in those cases (color vs fill for instance).  It seems reasonable to me then that the Radio style (in the same theme) would also have this implementation knowledge.  The implementation knowledge of the Icon's style is not leaking beyond the scope of the Teams theme package there.

### Child Icon _variable function_ is a _blocker_

The corresponding issue with variable functions is, how would a parent component change a slot variable on hover?  I can see no other way than for the parent to have mouse enter/leave listeners on the parent to update its state and pass it to the variable function.  I think this is a blocker opposed to a challenge.  We cannot reasonably put our seal of approval on a pattern that requires every UI component on the page to have JS listeners for hover, active, and focus just so variable functions can have access to those states.

kuzhelov:
lets follow on this discussion here, for the sake of it being more visible to other folks: #231

PR #209, opened 2018-09-10
fix(chore): add git tags verification logic

kuzhelov:
This change prevents wrong git tags (version labels) to be pushed from local repo. Provided logic is quite naive and extremely simple, but it should be quite robust in helping us to solve the git tags issues. 

One of the key principle of this changes is to not introduce any changes automatically on client machine, just inform about the fact that those should be considered (and recommend specific instructions). Once we will obtain evidence that these steps are safe and OS-agnostic, we could consider to make them automatic.

image

----------

Note that git tags should be cleaned for remote repo before merging these changes - will take care of that.

### DIFF
--- a/build/gulp/tasks/git/prepush.ts
+++ b/build/gulp/tasks/git/prepush.ts
@@ -1,0 +1,30 @@
+import { execSync } from 'child_process'
+
+// naive check is implemented for now - this checks if Semantic UI tags are present
+const hasWrongTags = (gitTagsList): boolean => {
+  // this one is guaranteed being relevant to Semantic UI only
+  const SemanticUiVersionTag = 'v0.1.2'
+
+  return gitTagsList.includes(SemanticUiVersionTag)
+}
+
+type Error = string
+
+export default (): Error => {
+  const gitTagsList = `${execSync('git tag -l')}`
+  if (hasWrongTags(gitTagsList)) {
+    const errorMessage = [
+      '!! YOUR LOCAL REPO CONTAINS WRONG GIT TAGS !!',
+      '-----------------------------------------------',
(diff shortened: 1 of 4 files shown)


build/gulp/tasks/git/prepush.ts
```diff
@@ -0,0 +1,30 @@
...
+type Error = string
+
+export default (): Error => {
+  const gitTagsList = `${execSync('git tag -l')}`
+  if (hasWrongTags(gitTagsList)) {
+    const errorMessage = [
+      '!! YOUR LOCAL REPO CONTAINS WRONG GIT TAGS !!',
+      '-----------------------------------------------',
+      'To prevent wrong git tags from being pushed, please, consider to do the following steps:',
+      ' - delete all local tags: git tag -l | xargs git tag -d',
```
miroslavstastny, line 20: Would you mind adding Windows instructions as well (there is no `xargs` on Windows)
```sh
FOR /f "tokens=*" %a in ('git tag -l') DO @git tag -d %a
```
Source: https://gist.github.com/RandomArray/fdaa427878952d9768b0

miroslavstastny (approved):

miroslavstastny:
Also we need to delete the tags in the repo before this gets merged :-)

build/gulp/tasks/git/prepush.ts
kuzhelov: actually, these ones should work for Windows - have tested on my Windows 10 machine (thanks to WSL). But, for sure, if there will be any problem with them, will add Windows-specific ones

kuzhelov:
> Also we need to delete the tags in the repo before this gets merged :-)

sure - this is something I've mentioned in the description :) currently am working on it

PR #211, opened 2018-09-10
feat: add all font awesome icon names

levithomason:
This PR adds all Font Awesome icons to the style Map, Icon typings, and Icon propTypes.  This unblocks some of our consumers right away.  It is also something we'll keep in the future, perhaps in the `default` theme so we're not adding throw away work here.


miroslavstastny (dismissed):
This is ~80kB (unzipped). In the feature we should think about moving it to a separate package?

miroslavstastny:
image

miroslavstastny:
btw, the warning in console, if you use incorrect icon name is also cool:
image

levithomason:
>This is ~80kB (unzipped). In the feature we should think about moving it to a separate package?

This size may have included typings and propTypes which shouldn't be part of the production bundle.  However, I've removed the typings and propTypes for now.  Themes will likely define their own icon names and icons.  Until we resolve this, I don't think we should validate names.

Yes, this should be included in a separate package in the future, maybe the `default` theme.

>(Some of) the icons are broken

Fixed.

>btw, the warning in console, if you use incorrect icon name is also cool:

Fixed by relaxing the check to `string`.

Just FYI, `customPropTypes.suggest()` is capable of doing this (from SUIR):

image

levithomason:
This is passing on Screener and works locally.  I'm not sure why icons weren't loading for you locally.  However, I also have `fix/merge-themes` which fixes an issue with merging static styles and fonts in the theme.

I'm going to merge this PR adding icon names as is since it is passing and only adds the missing names to the existing pattern.  I will then finish the merge themes fix before making a release.

miroslavstastny:
The reason is all the icons which do not work for me are from **Font Awesome 5 Brands**, not from **Font Awesome 5 Free**.
`Icon` component currently supports free and outline only, not brands:
https://github.com/stardust-ui/react/blob/7cde3ddf63bd406232b8f83d78c4f5b28fe77617/src/themes/teams/components/Icon/iconStyles.ts#L25

levithomason:
Heads up, I've fixed brand icons as well now.

PR #213, opened 2018-09-11
feat(accessibility): add logic to differentiate mouse focus from keyboard focus

silviuaavram:
Removing focus ring when the focus is applied by mouse. Only added it on Buttons for now.

Checking with the library what-input the source of the focus. If focus comes from keyboard, make the state property isLastFocusFromKeyboard true, which triggers the custom style from the buttonStyles.ts

This focus style should be added in a new PR, and I've added the boolean variable in styles file for reference. It should be used like that or in a similar way, so the focus styles apply only on keyboard focus.

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -1,6 +1,7 @@
 import * as PropTypes from 'prop-types'
 import * as React from 'react'
 import whatInput from 'what-input'
+import * as _ from 'lodash'
 
 import { UIComponent, childrenExist, customPropTypes, createShorthandFactory } from '../../lib'
 import Icon from '../Icon'
@@ -27,7 +28,7 @@ export interface IButtonProps {
   iconOnly?: boolean
   iconPosition?: 'before' | 'after'
   onClick?: ComponentEventHandler<IButtonProps>
-  onFocus?: ComponentEventHandler<IButtonProps>
+  onFocus?: ComponentEventHandler<Extendable<IButtonProps>>
   text?: boolean
   type?: 'primary' | 'secondary'
   accessibility?: Accessibility
@@ -142,7 +143,7 @@ class Button extends UIComponent<Extendable<IButtonProps>, any> {
   static Group = ButtonGroup
 
   public state = {
-    isLastFocusFromKeyboard: false,
+    isFromKeyboard: false,
   }
 
   public renderComponent({
@@ -196,11 +197,11 @@ class Button extends UIComponent<Extendable<IButtonProps>, any> {
   }
 
   private handleFocus = (e: React.SyntheticEvent) => {
-    const { onFocus } = this.props
-    this.setState({ isLastFocusFromKeyboard: whatInput.ask() === 'keyboard' })
-    if (onFocus) {
-      onFocus(e, this.props)
-    }
+    const isFromKeyboard = whatInput.ask() === 'keyboard'
+
+    this.setState({ isFromKeyboard })
+
+    _.invoke(this.props, 'onFocus', e, { ...this.props, isFromKeyboard })
   }
 }
 
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -28,7 +28,7 @@ export interface IButtonProps {
   iconOnly?: boolean
   iconPosition?: 'before' | 'after'
   onClick?: ComponentEventHandler<IButtonProps>
-  onFocus?: ComponentEventHandler<Extendable<IButtonProps>>
+  onFocus?: ComponentEventHandler<IButtonProps>
   text?: boolean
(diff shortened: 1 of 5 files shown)


src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -49,6 +49,13 @@ const buttonStyles: IComponentPartStylesInput = {
       verticalAlign: 'middle',
       cursor: 'pointer',
 
+      ':focus': {
+        ...(isLastFocusFromKeyboard &&
+          {
+            // focus styles should be added like this, since they should be applied only on keyboard.
+          }),
+      },
```
mnajdova: Should we merge this into this PR https://github.com/stardust-ui/react/pull/165 or we want to have this on the master so that it can be picked up for other styles as well?

src/components/Button/Button.tsx
```diff
@@ -179,6 +194,14 @@ class Button extends UIComponent<Extendable<IButtonProps>, any> {
       onClick(e, this.props)
     }
   }
+
+  private handleFocus = (e: React.SyntheticEvent) => {
+    const { onFocus } = this.props
+    this.setState({ isLastFocusFromKeyboard: whatInput.ask() === 'keyboard' })
+    if (onFocus) {
+      onFocus(e, this.props)
```
sophieH29: Shouldn't we use _.invoke, like this?
```_.invoke(this.props, 'onFocus', e, this.props)```
I see in some places we call it directly in others using ```_.invoke```, shouldn't we be consistent or is there any reason for using it here like this? Probably it's more question to @kuzhelov @mnajdova

File: src/themes/teams/components/Button/buttonStyles.ts (hunk 1 above)
Comment (sophieH29) [reply src/themes/teams/components/Button/buttonStyles.ts]: @mnajdova it was decided on the standup, that @silviuavram PR will go first, then Andrew will make his changes with styles based on it. @silviuavram correct me if I am wrong :)

File: src/components/Button/Button.tsx [hunk 3]
```diff
@@ -179,6 +194,14 @@ class Button extends UIComponent<Extendable<IButtonProps>, any> {
...
+    }
+  }
```
mnajdova: In my opinion it would be useful to send the isLastFocusFromKeyboard variable to the user as well, as it is something connected to the focus. I would propose this to be restructured in the following way:
```jsx
    const { onFocus } = this.props
    const isLastFocusFromKeyboard = whatInput.ask() === 'keyboard'
    
    if (onFocus) {
      onFocus(e, {...this.props, isLastFocusFromKeyboard})
    }

     this.setState({ isLastFocusFromKeyboard })
  }
```

Note: we need to have this additional variable as setState is asynchronous.

src/themes/teams/components/Button/buttonStyles.ts
mnajdova: Disregard this, I was just updated that that PR will be replaced with a new one.

src/components/Button/Button.tsx
kuzhelov: my vote would be to use just any safe approach that would work - in other words, as long as it is safe and declarative enough (so it is easy to read), I would be fine with any approach chosen. 

For this particular example, frankly, doesn't see much value in using `invoke` instead - but this is just my opinion.

The only thing I might suggest is to simplify this expression a bit:

```js
  ...
  onFocus && onFocus(e, this.props)
```

mnajdova (approved):

src/components/Button/Button.tsx
mnajdova: I prefer this invoking of the functions as well, as they look quite more simple and direct about what we want to do.

mnajdova:
I left one comment for consideration. Other than that it is good for merge. Good job!

kuzhelov (approved):

src/components/Button/Button.tsx
kuzhelov: would rather suggest to rename it then, as too much information is unnecessary in the state's name: something like `isFromKeyboard` should be sufficient

src/components/Button/Button.tsx
sophieH29: As for me, both of these ways seem to be quite safe and declarative. But I would prefer to select one option that will be used across stardust, to keep consistency.

src/themes/teams/components/Button/buttonStyles.ts
silviuaavram: @sophieH29 I will not correct you because you are not wrong.

src/components/Button/Button.tsx
miroslavstastny: I vote for consistency = `_.invoke()`

miroslavstastny (approved):

PR #214, opened 2018-09-11
BREAKING(Text): change font size scheme and Text size API

codepretty:
# Before

```jsx
<Text size="xs" />
<Text size="sm" />
<Text size="md" />
<Text size="lg" />
<Text size="xl" />
<Text size="2x" />
<Text size="3x" />
<Text size="4x" />
```
  
# After

```jsx
<Text size="smallest" />
<Text size="smaller" />
<Text size="small" />
<Text size="medium" />
<Text size="large" />
<Text size="larger" />
<Text size="largest" />
```

***

* Updated sizes to be smallest, smaller, small, medium, large, larger, largest
* Updated doc examples to show font sizes that are supported by the theme

**Was**
* showed hard-coded font sizes that were the same for all themes
image
 
**Now**
* shows only the font sizes that are defined in a theme's siteVariables
image


docs/src/examples/components/Text/Types/TextSizesExample.shorthand.tsx
```diff
@@ -1,23 +1,16 @@
...
-    <br />
-    <Text size="3x" content="Dicta voluptatum dolorem." />
-    <br />
-    <Text size="4x" content="Dicta voluptatum dolorem." />
-  </div>
+  <Provider.Consumer
+    render={({ siteVariables }) => {
+      return _.map(siteVariables.fontSizes, (value, key) => (
+        <div key={key}>
+          <Text size={key} content="Dicta voluptatum dolorem." />
```
miroslavstastny: I would also add the size to the `content` so I can immediately see the size names in docsite.

miroslavstastny (commented):
By this change, you are defining a contract for all themes that they must define their sizes in `fontSizes`.
How do you plan to enforce it? It should be at least added to `ISiteVariables`

levithomason (approved):
We will consider how to enforce this with typings later.  This topic also warrants an RFC on /specifications for how we deal with font sizes and component sizes at the framework API level.  The changes here are the result of many conversations and ideas with Fabric, teams around MS, and the open source.  We chose a middle ground path for sizes for now. We are open to future changes.

PR #215, opened 2018-09-11
fix: Teams zwj gender emoji in Chrome

markpete:
Chrome incorrectly handles zwj (zero-width joiner) gender emoji's and renders them as two emoji instead of a single one.  
For an example, please see this emoji in both Chrome and Edge within stardust.
🤦‍♂️

Adding fonts that correctly support ZWJ's for Mac and Windows fixes the issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## [Unreleased]
 
 ### Fixes
+- Fixed fonts to support zwj gender emojis ([#215](https://github.com/stardust-ui/react/pull/215))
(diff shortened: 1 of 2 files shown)


levithomason (changes requested):
Changes look good 👍 

Clean up tips:

### Semantic PR Title
PR is failing the title check.  It needs to conform to [semantic pull request](https://github.com/probot/semantic-pull-requests) titles.  These are essentially the same as Angular commit message guidelines.

This PR might be title `fix: Teams theme zwj gender emoji in Chrome`

### Changelog

Since this is a user facing change, the CHANGELOG.md in the root of the project needs an entry.  Instructions are included at the top of the  changelog.

levithomason (approved):

PR #216, opened 2018-09-11
feat: Updating buttons to be in line with new styles

notandrew:
Updating the colors and borders to match current Teams redlines


jurokapsiar:
Text and starndard buttons are not readable when focused by keyboard:

image
image

Also, border for circular emphasis does not look correct:
image

Will these be part of separate PR or should they be fixed here?

notandrew:
> Text and starndard buttons are not readable when focused by keyboard:
> 
> image
> image
> 
> Also, border for circular emphasis does not look correct:
> image
> 
> Will these be part of separate PR or should they be fixed here?

@jurokapsiar -- I wasn't looking at those buttons for some reason. I fixed the outlines for non-primary & secondary buttons. In a future checkin, I'll update those to the correct styles.

kuzhelov (approved):

PR #217, opened 2018-09-11
fix(mergeThemes): preserve fonts and static styles

levithomason:
Merge themes currently doesn't handle font faces and static styles.  When merging themes, this information is lost.  We started merging themes at the top of the doc site in #204.  Therefore, the doc site fonts and static styles are not correct.

This PR introduces font face and static style merging which resolves the regression introduced.

### DIFF
--- a/src/lib/mergeThemes.ts
+++ b/src/lib/mergeThemes.ts
@@ -4,6 +4,7 @@ import {
   ComponentVariablesInput,
   IComponentPartStylesInput,
   IComponentPartStylesPrepared,
+  IFontFace,
   ISiteVariables,
   IThemeComponentStylesInput,
   IThemeComponentStylesPrepared,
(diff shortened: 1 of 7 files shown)


src/lib/mergeThemes.ts
```diff
@@ -143,11 +147,21 @@ export const mergeRTL = (target, ...sources) => {
...
+}
+
 const mergeThemes = (...themes: IThemeInput[]): IThemePrepared => {
   const emptyTheme = {
     siteVariables: {},
     componentVariables: {},
     componentStyles: {},
+    fontFaces: [],
+    staticStyles: [],
   } as IThemePrepared
```
alinais, line 166: looks good to me

alinais: posted in wrong pr...

PR #218, opened 2018-09-12
feat(Accessibility): Add accessibility behavior to Popup

sophieH29:
This PR aims to make popup accessible, by adding ```PopupBehavior``` to the ```Popup``` component.

For accessibility purpose, we should be able to manage ```open```/```close``` state of the Portal from Popup component. So in th PR I've added ```popupState``` to the Popup which controlls ```open``` prop of the Portal.

```PopupBehavior``` details:
- ```aria-* attributes``` - popup's trigger will receive attributes like 'aria-haspopup'
- ```keyActions``` 
       - for the ```trigger``` component's part - ```performClick``` (on ```Enter``` & ```Space```), which currently will toggle open/close state) 
       - for ```popup``` component's part - ```closeAndFocusTrigger``` - when user presses ```Esc``` on a popup, it will be closed and the trigger node will be focused

Extended ```Portal```'s API
- ```triggerAccessibility``` - receives accessibility related attributes, like

```js
{
  aria-haspopup: true,
  aria-disabled: false,
  onKeyDown: f()
}
```
- ```onOutsideClick``` - callback whick can be invoked when document was clicked outside of trigger and popup nodes
- ```onTriggerClick``` - since it's needed to control ```open``` state from popup, we have to handle trigger click in a popup too

Suggestion, probably for future:
- extend Popup API, by adding props ```onOpen```, ```onClose```, so users can controll what should be done before opening and closing a popup, for e.g. managing focus


src/components/Popup/Popup.tsx
```diff
@@ -146,11 +183,32 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
+    this.setState({ popupOpened: true }, () => afterRenderCb && afterRenderCb())
+  }
+
+  private closePopup = (e: Event, afterRenderCb?: () => void) => {
+    e.preventDefault()
+    this.setState({ popupOpened: false }, () => afterRenderCb && afterRenderCb())
+  }
+
+  private focusTrigger = () => {
+    const triggerRef = this.state.triggerRef
```
bmdalex: can be one liner:
```typescript
_.invoke(this.state, 'triggerRef')
```

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
```diff
@@ -0,0 +1,34 @@
+import { Accessibility } from '../../interfaces'
+import * as keyboardKey from 'keyboard-key'
+
+const isButton = props => {
+  return props && props.trigger && props.trigger.props && props.trigger.props.as === 'button'
```
bmdalex: U can also do this to avoid all the ugly checks.

```typescript
try {
  return props.trigger.props.as === 'button'
catch (_) {
  return false
}
```

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
```diff
@@ -0,0 +1,34 @@
+import { Accessibility } from '../../interfaces'
+import * as keyboardKey from 'keyboard-key'
+
+const isButton = props => {
```
bmdalex: this is dangerous, hacky and definitely not robust enough; imagine a prop name is changed..this won't work anymore; pls. consider another way or at least add types to `props` here so that the compiler might help prevent bugs in case of prop changes

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
```diff
@@ -0,0 +1,34 @@
...
+  return props && props.trigger && props.trigger.props && props.trigger.props.as === 'button'
+}
+
+const PopupBehavior: Accessibility = (props: any) => ({
+  attributes: {
+    trigger: {
+      role: isButton(props) ? undefined : 'button',
```
bmdalex: pls cache `isButton` check

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
bmdalex: I just noticed that you use this to add some attributes to non-buttons; is that really necessary? I don't see any problem in overriding attributes here

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
bmdalex: also, what about links (`a` tags), other `Button` components rendered as something else, etc?

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
kolaps33: not sure how we should handle these (role and tabindex) 
If it will be for example anchor, then we whould not set tabindex and neither role.

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
```diff
@@ -0,0 +1,34 @@
+import { Accessibility } from '../../interfaces'
```
kolaps33, line 1: please add description :) and if you can use format of the description/sentences which is alrady in ohters beharviors files... Thank you :)

src/components/Popup/Popup.tsx
```diff
@@ -146,11 +183,32 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
     )
   }
 
+  private onTriggerClick = () => {
+    this.setState({ popupOpened: !this.state.popupOpened })
+  }
+
+  private openPopup = (e: Event, afterRenderCb?: () => void) => {
+    e.preventDefault()
+    this.setState({ popupOpened: true }, () => afterRenderCb && afterRenderCb())
```
bmdalex: no need for `afterRenderCb` logic, just pass the parameter along, it will be handled by `setState` function
```tsx
this.setState({ popupOpened: false }, afterRenderCb)
```
same below

also, `openPopup` and `closePopup` contain duplicated logic; rather have one method with `handlePopupState` that contains `popupOpened` as param;

```typescript
private handlePopupState = (e: Event, popupOpened: boolean, afterRenderCb?: () => void) => {
    e.preventDefault()
    this.setState({ popupOpened }, afterRenderCb)
}
```

src/components/Popup/Popup.tsx
```diff
@@ -96,47 +108,72 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
     align: 'start',
     position: 'above',
+    accessibility: PopupBehavior as Accessibility,
   }
 
-  public state = { triggerRef: undefined }
+  public state = { triggerRef: undefined, popupOpened: false }
+
+  actionHandlers: AccessibilityActionHandlers = {
+    open: e => this.openPopup(e),
```
bmdalex: no need to create other functions here, they already are:
```typescript
open: this.openPopup,
close: this.closePopup,
closeAndFocusTrigger: this.closePopup.bind(this, this.focusTrigger),
```

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
sophieH29: yes, you're right. I am not sure for now how better handle it, but it something that we can discuss.

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
sophieH29: you are right that this check looks not nice, haven't yet come up with a better idea.
And yes, it's important to set these attributes ```role=button``` and ```tabindex=0```, in case it's not a focusable element, for e.g. when user will use just ```div``` as a trigger.

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
```diff
@@ -0,0 +1,34 @@
...
+      'aria-haspopup': 'true',
+      'aria-disabled': !!props['disabled'],
+    },
+  },
+  keyActions: {
+    trigger: {
+      open: {
+        keyCombinations: [{ keyCode: keyboardKey.Enter }, { keyCode: keyboardKey.Spacebar }],
+      },
+      close: {
```
kolaps33: as I spoke with Sofia and understood that this scenario is closing the popup when focus is on the element (which triggered the popup) 
I am not sure how it should behave because usually user closing by ESC key the stuff which he is focusing... 
This is just point to the discussion...

sophieH29: definitely, it is something we need to discuss. If the focus is on trigger element, should we close on Esc, or toggle popup using Enter and Space keys. Actually, the latest option I would consider better
cc @jurokapsiar

src/components/Popup/Popup.tsx
sophieH29: ok, make sense. I thought it would bring more readability :)

src/components/Popup/Popup.tsx
sophieH29: ```.bind``` works in a not very straightforward way, so I would need to change function signature, by replacing order of params. Which confuses while reading the code. Also ```this.focusTrigger``` is an arrow function now, in a case to use it as you proposed, I would need either move it above the ```actionHandlers``` or make it a regular function and bind context in the constructor. 
It's kind of tradeoff. If you don't have serious concerns regarding it, I would leave as it is.
Also glad to hear other opinions on it.

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
bmdalex: I didn't mean not adding the rol and tabindex, but adding them all the time, regardless if we have a button or not; is it really necessary to check the component is button? we can just set the `role=button` and `tabindex=0` anyway since a button has them already set

src/components/Popup/Popup.tsx
bmdalex: hm, ok, no strong objections; is `actionHandlers` meant to be public btw?

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
sophieH29: We can set it, but then we may break user's implementation if they need anything else.
To be safer, we need to put ```role=button```  and ```tabIndex=0```, if these attributes weren't set already by a user or the element is not focusable (which are in fact 'input', 'button', 'a').
I've made changes to that part, let me know what do you think.

src/components/Popup/Popup.tsx
sophieH29: It is protected, inherited from UIComponent.

src/components/Popup/Popup.tsx
```diff
@@ -96,47 +109,71 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
+        onOutsideClick={e => this.handlePopupState(e, false)}
+        onTriggerClick={this.onTriggerClick}
       >
-        {childrenExist(children) ? children : this.renderContent(rtl)}
+        {childrenExist(children)
+          ? children
+          : this.renderContent(
+              computePopupPlacement({ align, position, rtl }),
+              this.renderPopperChildren.bind(this, rtl, accessibility),
+            )}
```
mnajdova, line 152: renderPopperChildren is an arrow function, do we still need the bind?

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
mnajdova: I am wondering if the best way is inside the Popup behavior to be handling the trigger's accessibility. This can be any component, and if the component itself already has some accessibility defined, isn't that enough?

src/components/Popup/Popup.tsx
sophieH29: As I understood from the previous implementation, ```.bind(this)``` passes ```{ ref, style }: PopperChildrenProps``` to the method through a callback

mnajdova: Aahh, got it, this is used in the Popper component.

src/components/Popup/Popup.tsx
```diff
@@ -28,6 +36,7 @@ export interface IPopupProps {
 
 export interface IPopupState {
   triggerRef: HTMLElement
+  popupOpened?: boolean
```
miroslavstastny, line 39: Why is `Popup` not an auto controlled component (with the same prop name as `Portal`)?

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
```diff
@@ -0,0 +1,64 @@
...
+      role: getAriaAttribute('role', props, 'button'),
+      tabIndex: getAriaAttribute('tabIndex', props, '0'),
+      'aria-haspopup': 'true',
+      'aria-disabled': !!props['disabled'],
+    },
+  },
+  keyActions: {
+    trigger: {
+      performClick: {
+        keyCombinations: [{ keyCode: keyboardKey.Enter }, { keyCode: keyboardKey.Spacebar }],
```
miroslavstastny, line 24: What's intended behavior here?

I understand I can open popup using Enter or Spacebar - where should the focus go after that? Now it remains on the trigger, there is no way how I could navigate inside the popup. What if there is a menu inside the popup?

And when the focus remains on the trigger, I would expect it would be closable by Esc (copying `closeAndFocusTrigger` from ~5 lines below to here works for me), and personally I am not sure that I expect Enter to close it.

sophieH29: Focus management would be introduced in future PR (+ FocusTrapZone which also handles focus after popup was rendered https://github.com/stardust-ui/react/pull/239). This PR just adds accessibility behavior + action handlers. 

Regarding closing the popup by Esc, initially it was implemented as you mentioned. But after discussion with Milan, we came up that it is not correct from an accessibility point of view. Esc should work for a popup to close it when the focus is on the popup. For trigger element, it will behave the same way as clicking with a mouse - which will toggle popup currently.

src/components/Popup/Popup.tsx
sophieH29: Actually, a good question :) Initial implementation of Popup doesn't expose this prop to users, and when I was making this change, I didn't want to change API of a component but leave as it is. 
So only behavior and key handlers would be added. Anyway, this can be a point for discussion as well as a separate PR.

kuzhelov (changes requested):
@sophieH29, please, consider to fix project's build

src/components/Popup/Popup.tsx
```diff
@@ -96,56 +109,97 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
     as: Portal,
     align: 'start',
     position: 'above',
+    accessibility: PopupBehavior as Accessibility,
```
kuzhelov: would rather suggest the following more robust approach to achieve the same thing:

```jsx
public static defaultProps: IPopupProps = {
  ...
  accessibility: PopupBehavior
}
```

This will ensure that all the changes in the IPopupProps interface (that is used in many other places) will take effect here as well.

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
```diff
@@ -0,0 +1,64 @@
...
+      },
+    },
+    popup: {
+      closeAndFocusTrigger: {
```
kuzhelov, line 28: nice naming 👍

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
```diff
@@ -0,0 +1,64 @@
...
+      (type === 'a' && href !== undefined) ||
+      as === 'button'
+    )
+  } catch {
+    return false
+  }
+}
+
+const getAriaAttribute = (attribute, props, defaultValue) => {
+  try {
```
kuzhelov: what could go wrong here? Couldn't we check some conditions instead?

src/lib/accessibility/interfaces.ts
```diff
@@ -170,9 +170,11 @@ export type AccessibilityActionHandlers = {
 }
 
 export type ActionsKeyHandler = {
-  [partName: string]: {
-    onKeyDown?: KeyboardHandler
-  }
+  [partName: string]: OnKeyDownHandler
```
kuzhelov, line 173: 👍

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
kuzhelov: according to what I see from the lines above (where, essentially, the same pattern is applied), it is about possibility of `props.trigger` being null/undefined. If so, it would be better to introduce checks for that

src/components/Popup/Popup.tsx
```diff
@@ -96,56 +109,97 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
     rtl,
   }: IRenderResultConfig<IPopupProps>): React.ReactNode {
-    const { children, trigger } = this.props
+    const { children, trigger, position, align } = this.props
 
     return (
       <ElementType
         className={classes.root}
         {...rest}
+        open={this.state.popupOpened}
```
kuzhelov, line 135: even when it was just about `trigger` and `triggerRef` props being defined for the `ElementType` component, we already had the problem related to restricted options for `as` values (because now we are, essentially, requesting all these newly added props to be supported). Now the problem is worse - we definitely need to address this in the nearest future, as there are some options for that

src/components/Popup/Popup.tsx
```diff
@@ -96,56 +109,97 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
-    const placement = computePopupPlacement({ align, position, rtl })
 
     return (
       triggerRef && (
         <Popper
-          placement={placement}
+          placement={popupPlacement}
           referenceElement={triggerRef}
-          children={this.renderPopperChildren.bind(this, rtl)}
+          children={renderChildrenCb}
```
kuzhelov: would suggest to rename it to something like `renderPopperChildrenFromProps`, this may prevent some sources of confusion (like the ones that @mnajdova has experienced while writing here comment, I felt the same ones)

src/components/Popup/Popup.tsx
```diff
@@ -96,56 +109,97 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
+        {...accessibility.attributes.popup}
+        {...accessibility.keyHandlers.popup}
+      >
         {content}
       </Popup.Content>
     )
   }
 
+  private onTriggerClick = (e: Event) => {
+    this.handlePopupState(e, !this.state.popupOpened)
```
kuzhelov: `state` could be updated asynchronously in React - for that reason I would suggest to use the following form of `setState` method in `handlePopupState`:

```jsx
 private handlePopupState = (..., getPopupOpened: (previousOpened) => boolean, ...) => {
    ..
    this.setState((previousState) => { popupOpened: getPopupOpened(previousState.popupOpened) }, ...)
  }
```

Key difference is that value of the previous state is safely provided here.

src/components/Popup/positioningHelper.ts
```diff
@@ -2,6 +2,7 @@ import { Placement } from 'popper.js'
 
 export type Position = 'above' | 'below' | 'before' | 'after'
 export type Alignment = 'top' | 'bottom' | 'start' | 'end' | 'center'
+export type Placement = Placement
```
kuzhelov: `export { Placement } from 'popper.js` would be more descriptive

src/lib/accessibility/Behaviors/Popup/PopupBehavior.ts
```diff
@@ -0,0 +1,64 @@
...
+      type === 'input' ||
+      (type === 'a' && href !== undefined) ||
+      as === 'button'
+    )
+  } catch {
+    return false
+  }
+}
+
+const getAriaAttribute = (attribute, props, defaultValue) => {
```
kuzhelov: nit: just for the sake of making it easier to read, especially at place where method is used:

```jsx
const getAriaAttributeFromProps = (attributeName: string, props: ..., defaultValue: string) => {
```

src/components/Popup/Popup.tsx
```diff
@@ -96,56 +109,97 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
 
     return (
       <ElementType
         className={classes.root}
         {...rest}
+        open={this.state.popupOpened}
         trigger={trigger}
         triggerRef={this.handleTriggerRef}
+        triggerAccessibility={{
+          ...accessibility.attributes.trigger,
```
kuzhelov: won't it be better to use the object of the following structure instead, so that attributes and keyHandlers will be explicitly splitted? Seems to be the safest option at the first stages - and it will resemble we currently have for accessibility behavior objects 

```jsx
triggerAccessibility={{
  attributes: accessibility.attributes.trigger,
  handlers: accessibility.keyHandlers.trigger
}}
```

src/components/Popup/Popup.tsx
sophieH29: Totally agree with that, same as passing accessibility attributes for the ```trigger``` element

src/components/Popup/Popup.tsx
sophieH29: agree, will rename it

kuzhelov (changes requested):
please, take a look on the comments - the most important ones are about state change handling where value of previous 'state' is taken into account

src/components/Popup/Popup.tsx
sophieH29: It always good to explicitly set values. But then in ```Portal``` will need to handle it as object
```
{
   attributes:...
   handlers:...
}
```

Which brings us to handling one more Type.

Currently, we are safely passing from ```Popup``` trigger related attributes
And in ```Portal``` we expect ```triggerAccessibility?: IAccessibilityAttributes & OnKeyDownHandler``` which also applies some constraints and make it safer

What do you think?

src/components/Popup/Popup.tsx
kuzhelov: yes, lets address it by means of the follow up PR, will take care of that

src/components/Popup/Popup.tsx
```diff
@@ -92,60 +105,108 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
-    const placement = computePopupPlacement({ align, position, rtl })
 
     return (
       triggerRef && (
         <Popper
-          placement={placement}
+          placement={popupPlacement}
           referenceElement={triggerRef}
-          children={this.renderPopperChildren.bind(this, rtl)}
+          children={renderPopperChildrenFromProps}
```
kuzhelov, line 168: 👍

src/components/Popup/Popup.tsx
kuzhelov: sorry, but still not sure that this is the reason to avoild explicitness, as the new style is introduced even now by using `IAccessibilityAttributes & OnKeyDownHandler` expression - and there is no enforcement for this type on the side where object is provided. The same could be achieved if type literal object will be used instead:

```js
triggerAccessibility?: { attributes: IAccessibilityAttributes, handlers: OnKeyDownHandler }
```

Lets keep this in mind and try to address this thing in future - especially given that we both agree on the benefits of having these two being provided as explicit props

sophieH29: actually, I understood you a bit wrong from the 1st message, sorry. Agree with this proposal and will make the change. Thank you!

src/components/Portal/Portal.tsx
```diff
@@ -113,17 +137,19 @@ class Portal extends AutoControlledComponent<IPortalProps, IPortalState> {
...
-    const { trigger } = this.props
+    const { trigger, triggerAccessibility } = this.props
 
     return (
       trigger && (
         <Ref innerRef={this.handleTriggerRef}>
-          {React.cloneElement(trigger, { onClick: this.handleTriggerClick })}
+          {React.cloneElement(trigger, {
+            onClick: this.handleTriggerClick,
+            ...triggerAccessibility,
```
kuzhelov: another related problem (to the fact that there is no enforcement over the `triggerAccessibility` type on the client-code side) is that currently it is very easy to accidentally do things wrong: it is sufficient to add any property to `triggerAccessibility` object that cannot be properly expanded to element's attributes to break the entire app. For the sake of argument, you could try the following for `Popup.tsx` (it won't be caught by the compiler now): 

```jsx
return (
      <ElementType
       ... 
        triggerAccessibility={{
          ...accessibility.attributes.trigger,
          ...accessibility.keyHandlers.trigger,
          someFoo: { bar: true }
        }}
       ...
```

I would really suggest to introduce some type to restrict the set of possible values, and use it on the client side as well - the least we can do for now:

```jsx
// in Popup
  return (
      <ElementType
       ... 
        triggerAccessibility={{
          ...accessibility.attributes.trigger,
          ...accessibility.keyHandlers.trigger,
        } as TriggerAccessibility}
```

```jsx
// in types declaration
type TriggerAccessibility =  IAccessibilityAttributes & OnKeyDownHandler // or better { attributes: ..., handlers: ...}
```

kuzhelov (changes requested):
great improvement - lets address (discuss) the problem related to triggerAccessibility object's type, and it seems that we would be good to go :)

src/components/Portal/Portal.tsx
sophieH29: I'll make this type

```
type TriggerAccessibility = {
  attributes?: IAccessibilityAttributes
  keyHandlers?: OnKeyDownHandler
}
```

sophieH29:
@kuzhelov Thank you for your review! Made changes regarding ```triggerAccessebility``` type

src/components/Popup/Popup.tsx
```diff
@@ -92,60 +105,108 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
       <ElementType
         className={classes.root}
         {...rest}
+        open={this.state.popupOpened}
         trigger={trigger}
         triggerRef={this.handleTriggerRef}
+        triggerAccessibility={{
+          attributes: accessibility.attributes.trigger,
+          keyHandlers: accessibility.keyHandlers.trigger,
+        }}
```
kuzhelov: lets, please, make explicit `as` cast here (and use `TriggerAccessibility` type) - this will ensure that it would be hard to accidentally break the app with new props being added to either attributes or handlers

kuzhelov (approved):
lets, please, ensure that it would be hard to break the app with new property being added to either `attributes` or `keyHandlers` part of the `triggerAccessibility` object - have left a comment. Other than that we are good to merge 👍 

Also, please, do not forget to mention the changes made in the CHANGELOG file.

PR #219, opened 2018-09-12
feat(test): changed logic for reusing the component in the isConformant

mnajdova:
# isConformant

This PR introduces a change in the way the component is reused in the isConformant test. Also the version of the enzyme and enzyme-adapter-react-16 are updated to newer version. The reason for these changes are to see if the performance of the tests will be improved.


kuzhelov:
@mnajdova, are there any numbers available that could verify these changes will improve UTs performance? it would be great to see the measurements first, have clear understanding of where is a bottleneck, and after that see that these numbers are improved with the changes suggested. Please, don't get me wrong - not saying that these ones shouldn't be applied, but would rather like to see how this problem is measured/tracked to be sure that with these changes we are really addressing the problem

test/specs/commonTests/isConformant.tsx
```diff
@@ -31,19 +31,25 @@ export default (Component, options: any = {}) => {
   const { throwError } = helpers('isConformant', Component)
 
   const componentType = typeof Component
+  const wrappedComponent = mount(<Component {...requiredProps} />)
+
+  const createComponent = props => {
```
kuzhelov, line 36: this should be better named - with `create` used in method's name there is a reasonable assumption that new instance will be initialised - this is not the case as already created instance will be reused.

test/specs/commonTests/isConformant.tsx
```diff
@@ -31,19 +31,25 @@ export default (Component, options: any = {}) => {
   const { throwError } = helpers('isConformant', Component)
 
   const componentType = typeof Component
+  const wrappedComponent = mount(<Component {...requiredProps} />)
```
kuzhelov, line 34: with this change we are breaking tests isolation to some degree - now it could be the case that some state of the component that was set as a matter of previous test will affect next one. Just something that we should be aware of before making this change - we should carefully weight props and cons of this move, and to get idea about pros we would need to measure numbers first (and understand the source of these numbers - in either case, obtain evidence that the thing we are trying to optimize is indeed a source of bottleneck)

mnajdova: Totally aware of this, when we really need for some test the component with the exact props provided rendered, we should use the mount, but if we are ok with reusing the existing with just adding some props, this should be enough. This was considered for the tests where it is changed, because there is no real need for the component to be mounted over and over again.

mnajdova:
@kuzhelov I open the PR in order for us to actually discuss how we want this to be measured. I was comparing it on my local machine, but it is hard to decide because after couple of execution of the tests, everything goes much faster if there are no changes made. Here are the results I get from first time running the tests yesterday on the master and on this branch (synchronized with the latest master).

**Before:**
screenshot_8


**After:**
screenshot_9

But as i said, after couple of execution of the tests, the process on the both branches is finishing in approximately the same time. So any suggestion about how to really test the performance are welcomed :)

test/specs/commonTests/isConformant.tsx
mnajdova: Will try to come up with better name for this one.

kuzhelov:
yes, lets try to think about how we could reliably measure tests performance and reliably identify the bottlenecks (to prove the `mount` hypothesis)

test/specs/commonTests/isConformant.tsx
kuzhelov: yes, but here is the problem that could happen: suppose that now we are totally aware that this particular test is safe to reuse component, because during its execution nothing in terms of component's state is changed/used. Then, on the next day (after the test is introduced) the logic of component changes, and it results in our test checking for wrong condition - because of this isolation invariant (that is, actually, a fundamental assumption) being not satisfied. This is the main factor for me to make it as a deliberate step and only if we are absolutely sure that it is critically important for us

mnajdova:
Closing this in favor of: #223

PR #220, opened 2018-09-13
feat: add Attachment component

levithomason:
# Attachment

An Attachment displays a file attachment.

### TODO

- [X] Conformance test
- [X] Minimal doc site example
- [X] Teams Light theme
- [X] Confirm RTL usage
- [x] [W3 accessibility](https://www.w3.org/standards/webdesign/accessibility) check
- [x] [Stardust accessibility](https://github.com/stardust-ui/accessibility) check
- [x] Update the CHANGELOG.md
- [ ] Handle focus style outline
- [ ] add colors for various file types

# API Proposal

This PR is implementing the Attachment component from https://github.com/stardust-ui/specifications/issues/1.  The latest Teams theme redlines were used.  Sizes, spacing, colors, and states are redline correct for this version of the component.

image

```jsx
<Attachment
  icon="file word outline"    // font awesome, for now
  header="Document.docx"  // Text shorthand
  description="800 Kb"     // Text shorthand
  action={{ icon: 'x' }}      // Button shorthand
  progress={33}               // percent complete
/>
```

The markup includes a div for each slot.  This makes flexbox and css grid layouts easier.

```html
<div class="ui-attachment">
  <!-- Icon -->
  <div class="...">
  </div>

  <!-- Header / Description -->
  <div class="...">
    <div class="...">Document.docx</div>
    <div class="...">800 Kb</div>
  </div>

  <!-- Action -->
  <div class="...">
    </button>
  </div>

  <!-- Progress -->
  <div class="..."></div>
</div>
```

### Actionable

This prop styles the component to appear actionable by the user.  A pointer cursor is used and a hover color is applied.

```jsx
<Attachment actionable ... />
```

### Things to know

- **Limited features**: This PR only introduces the base component and its slots.  Other PRs will handle the various types and colors of Attachments.
- **Icon**: These are currently font awesome icons, pending #183 and subsequent work.
- **Action**: The action (Button) likely needs more specific positioning and styling to match the Teams theme.  This will require some button updates and maybe some theme updates.  A separate PR should handle this.
- **Multiple Actions** - The spec called for multiple actions, however, this usage was limited to Slack.  Gmail, Teams, and Outlook all use a single action.  A single `action` slot was used instead.  The render prop callback can be used to provide a button group if the user wishes.


src/themes/teams/components/Attachment/attachmentStyles.ts
```diff
@@ -0,0 +1,69 @@
...
+  }),
+
+  header: ({ props, variables }: StyleParam): ICSSInJSStyle => ({
+    fontSize: variables.headerFontSize,
+    fontWeight: variables.headerFontWeight,
+    lineHeight: variables.headerLineHeight,
+  }),
+
+  description: ({ props, variables }: StyleParam): ICSSInJSStyle => ({
+    opacity: 0.5,
```
notandrew: Can you remove this? We don't want opacity set for the description...

src/themes/teams/components/Attachment/attachmentStyles.ts
```diff
@@ -0,0 +1,69 @@
...
+    flex: '0 0 auto',
+    marginRight: variables.iconSpace,
+  }),
+
+  action: ({ props }: StyleParam): ICSSInJSStyle => ({
+    flex: '0 0 auto',
+  }),
+
+  progress: ({ props, variables }: StyleParam): ICSSInJSStyle => ({
+    transition: 'width 0.2s',
```
notandrew: looks like the animation's working!

notandrew (approved):

src/themes/teams/components/Attachment/attachmentStyles.ts
```diff
@@ -0,0 +1,69 @@
...
+  description: ({ props, variables }: StyleParam): ICSSInJSStyle => ({
+    opacity: 0.5,
+    fontSize: variables.descriptionFontSize,
+    fontWeight: variables.descriptionFontWeight,
+    lineHeight: variables.descriptionLineHeight,
+  }),
+
+  icon: ({ props, variables }: StyleParam): ICSSInJSStyle => ({
+    flex: '0 0 auto',
+    marginRight: variables.iconSpace,
```
notandrew: I think we don't want this margin since the attachment's 8px padding is enough to get this icon in the correct place.

src/components/Attachment/Attachment.tsx
```diff
@@ -0,0 +1,126 @@
...
+    return (
+      <ElementType {...rest} className={classes.root}>
+        <div className={classes.icon}>
+          {Icon.create(icon, {
+            defaultProps: {
+              size: 'big',
+              fitted: true,
+            },
+          })}
+        </div>
```
mnajdova: I am wondering if this is the best approach, adding div's for handling our styling. I don't see a way how can the user alter these, unless using CSS selectors. I had similar propose https://github.com/stardust-ui/react/pull/199#issuecomment-419465588 but I don't think thins should be the general approach when defining shorthands (see the comments below). What are your thoughts on this @levithomason ?

src/themes/teams/components/Attachment/attachmentStyles.ts
levithomason: This is between the icon and the header/description.  I believe the redline called for a larger value here.

docs/src/examples/components/Attachment/Slots/index.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react'
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
+
+const Slots = () => (
```
kuzhelov, line 5: one thing that am thinking about - probably, it would be nice it we would come up with some general logic of presenting these slots in a way that those will be more visible to the client

docs/src/examples/components/Attachment/Slots/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+  <ExampleSection title="Slots">
+    <ComponentExample
+      title="Icon"
+      description="An attachment can have an icon."
+      examplePath="components/Attachment/Slots/AttachmentIconExample"
+    />
+    <ComponentExample
+      title="Header"
```
kuzhelov, line 13: header is advertised as _slot_, and maybe I've made a false conclusion from it, but the first thing I've tried was the following - essentially, one of those things that we might expect from the shorthands:

```jsx
<Attachment header={{ content: '...', styles: { color: 'red' } }} />
```

Apparently, it doesn't work. So, the question is the following: do we see this general pattern (that is applicable for shorthands) to be applicable to any slot content? Actually, more interested about our vision on that. 

To me it seems to be a very good thing to ensure this consistency in our customization mechanisms - especially given that the following approach for style tweaking seems to be deprecated:

```jsx
<Attachment styles={{ header: {..<header styles>..} }}
```

docs/src/examples/components/Attachment/Variations/index.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
+
+const Variations = () => (
+  <ExampleSection title="Variations">
+    <ComponentExample
+      title="Actionable"
+      description="An Attachment can be styled to indicate possible user interaction."
```
kuzhelov: not sure, but maybe we could better advertise to the user that this (and couple of other ones) example is clickable - frankly, was able to infer this fact only by looking on the code :)

src/components/Attachment/Attachment.tsx
kuzhelov: @mnajdova, sorry, but just to be sure - does your comment above, essentially, attribute to the problem of customizing slot's tree (the one that referenced issue is about)?

kuzhelov (approved):

src/components/Attachment/Attachment.tsx
mnajdova: Yes, exactly. Not strictly for this component, but just as e reference to the RFC about the customizing of the slot's tree.

docs/src/examples/components/Attachment/Slots/index.tsx
brianespinosa: @kuzhelov will this PR solve that? https://github.com/stardust-ui/react/pull/238

kuzhelov: @brianespinosa, sorry for the late response, my bad :( my promise to be better next time.

> @kuzhelov will this PR solve that? #238 

Unfortunately, now - and the reason is that `header` is currently introduced as a simple `string`, and not a Stardust component that encapsulates this knowledge on how this styling should be applied. 

However, this is quite easy to fix - we could use `Text` component for `header` and `content`, so that all functionality of slots (where main aspect is styling) will be preserved.

kuzhelov:
lets address pending comments that were left for already provided functionality, and merge this PR afterwards. 

After that there will be additional PR to solve the rest two requirements:

- [ ] Handle focus style outline
- [ ] add colors for various file types

Will take care of that.

src/components/Attachment/Attachment.tsx
```diff
@@ -0,0 +1,122 @@
...
+          {Icon.create(icon, {
+            defaultProps: {
+              size: 'big',
+            },
+          })}
+        </div>
+        {(header || description) && (
+          <div className={classes.content}>
+            {Text.create(header, {
+              defaultProps: {
```
mnajdova, line 95: +1

mnajdova (approved):

PR #221, opened 2018-09-13
fix(docs): font size issue

kuzhelov:
# Repro steps:
- open root docs page (https://stardust-ui.github.io/react/)
- switch to `Radio` component page (https://stardust-ui.github.io/react/components/radio)

### Expected result
- default font size remains to be 14px
image

### Actual result
- default font size is switched on 10px
image

# Cause:
The issue is caused by the following tree being rendered:
```jsx
<Provider theme={mergeThemes(teams,
  siteVariables: {
    ...
   htmlFontSize: '14px'
 })}
 ...
  <ComponentExample>
     <Provider theme={mergeThemes(teams, newTheme)}>
     ...
     </Provider>
   </ComponentExample>
} />
```

As we might see, when component example is rendered, `teams` theme as applied once again - with all its static styles that invalidate `htmlFontSize: 14px` override and set `htmlFontSize` to `10px`, the value defined by `teams` theme.

Provided fix eliminates specifying theme for individual example, as we would rather like to apply theme changes for all examples altogether.

# Broader issue to fix
It is a very dangerous phenomena related to the fact that theme applied to _subtree of components_ might change look of the overall page - with global CSS selectors being used in its static styles. More than that, this problem is very hard to diagnose, and for some situations I might imagine that it will be even hard to spot. I would propose to apply mechanisms of scoping static CSS selectors to the subtree theme as applied for (by default), with opened possibility to explicitly allow global CSS for Provider:

```jsx
{ /*global CSS selectors are allowed, so that theme could change overall page's look */ }
<Provider theme={teams} allowGlobalCss>  
  ...
  { /* static CSS styles provided will be scoped to Provider's subtree only */ }
  <Provider theme={someTheme}> 
     ...
  </Provider>
...
</Provider>
```


mnajdova (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixes
+- Fix `Popup` component imports @mnajdova ([#222](https://github.com/stardust-ui/react/pull/222))
+- Fix font size issue in docs @kuzhelov ([#221](https://github.com/stardust-ui/react/pull/221))
```
levithomason: I don't think we should advertise doc changes here, only library changes.  Thoughts on keeping the changelog specific to the library itself?

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixes
+- Fix `Popup` component imports @mnajdova ([#222](https://github.com/stardust-ui/react/pull/222))
```
levithomason: Cruft entry here.

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -354,7 +354,7 @@ class ComponentExample extends React.PureComponent<IComponentExampleProps, IComp
     }
 
     return (
-      <Provider theme={mergeThemes(themes.teams, newTheme)}>
+      <Provider theme={newTheme}>
```
levithomason, line 357: Good find!  I was wondering what this was from 😊 There is now a dead import to `mergeThemes` in this file that needs to be cleaned up.

levithomason (changes requested):
Some cleanup is due here.

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
levithomason: Resolved in #228

levithomason:
Agree on the broader issue.  @davezuko and I also discussed this similar issue.  There is also the converse case that something rendered down the tree might need to add statics or font faces in order to work.

levithomason (approved):

PR #97, opened 2018-08-16
chore: bundle size statistics

levithomason:
This PR uses webpack to compute bundle size for individual stardust components as well as for the whole package.

### TODO

- [x] Build bundle size stats for index (entire package)
- [x] Build bundle size stats for `/lib`
- [x] Build bundle size stats for each component
- [x] Build bundle size stats for each theme
- [ ] Ensure size stats are available in each PR for visibility
  - we are going to implement this in a separate PR
- [ ] Update release workflow to insert stats for the version number being released

***

Next, we'll make a stats page in the docs for visualizing and advertising this to users.  We will extend the stats to include API surface area and render performance later as well.

### Computing the statistics
`yarn build:stats` runs webpack with `webpack.config.stats.ts` configuration which builds bundles for:
- entire package,
- lib (core),
- all individual components,
- all individual themes.

All build are run in series to avoid memory issues.

Then `Unreleased` section of `docs/src/bundleStats.json` file is updated:
```json
{
  "Unreleased": {
    "bundles": [
      ...
      {
        "name": "component-Animation.js",
        "size": 262818
      },
      ...   
    ]
  },
  "0.12.0": {
    /* the same "bundles" structure */
   }
}
```
Order of all items in the JSON is deterministic so that only change in size (or a new version or package removed/added) results in this file being changed.

### Storing the results
New **Store bundle statistics** step was added to CircleCI config. This script, if run in PR build, checks that the file changed and if so, it commits it back to GIT ([see this commit](https://github.com/stardust-ui/react/pull/97/commits/447c365fcc57f181fd0ce79cfde3c1d1e5d1338c)).

### Every change which results in changed `bundleStats.json` triggers two build
**Store bundle statistics** commits a `bundleStats.json` back to Github if it has changed. This causes several problems in CI workflow:
- As we have **Auto-cancel redundant builds** enabled in the CircleCI, the build is cancelled,
- as a new commit appears on the branch, it starts a new build.
  - Commiting the stats with *[ci skip]* does not help as our merge policy says that build  for the last commit must be green to allow a merge.
  - If there was a change in the stats (again) the whole loop would happen again - that can happen due to a bug in our stats computation and would always happen for performance stats which have less stable results. 

### Developers have to deal with mid-air collisions
Every commit on a branch results in a commit from CI - Developers will have to `git pull --rebase` before pushing, will have to be more careful when force-pushing.

PR #224, opened 2018-09-13
perf: enable tree shaking for Webpack

levithomason:
This PR sets `sideEffects: false` in our `package.json` to inform Webpack that our library is safe for tree shaking.

https://webpack.js.org/guides/tree-shaking


dvdzkwsk:
giphy

dvdzkwsk (approved):

brianespinosa:
@levithomason how much did you benefit in package size from this change? I see you are using Webpack 3. I didn't realize `sideEffects: false` also worked for this version. The current Webpack docs are for Webpack 4.

levithomason:
The docs will always import 100% of the features so the size will be maxed at all times.  The change here was made for consumers to benefit from.

We are currently working on library level perf tracking #141 and library level bundle size tracking #97.  It would be nice to track partial use bundle size as well.

levithomason:
@brianespinosa ^

PR #227, opened 2018-09-14
fix(Provider): not rendering fonts

levithomason:
Currently, font faces are not being applied correctly in the docs.  This turned out to be due to both our doc build and a Provider bug.

### DIFF
--- a/build/gulp/tasks/generate/templates/docs-examples-components/$DisplayName/Types/$DisplayNameExample.shorthand.tsx
+++ b/build/gulp/tasks/generate/templates/docs-examples-components/$DisplayName/Types/$DisplayNameExample.shorthand.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import { $DisplayName } from '@stardust-ui/react'
 
-const $DisplayNameExampleShorthand = () => <$DisplayName />
+const $DisplayNameExampleShorthand = () => <$DisplayName content="Hello from $DisplayName!" />
 
 export default $DisplayNameExampleShorthand
--- a/package.json
+++ b/package.json
@@ -104,6 +104,7 @@
     "awesome-typescript-loader": "4.0.1",
     "connect-history-api-fallback": "^1.3.0",
     "copy-to-clipboard": "^3.0.8",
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -104,6 +104,7 @@
     "awesome-typescript-loader": "4.0.1",
     "connect-history-api-fallback": "^1.3.0",
     "copy-to-clipboard": "^3.0.8",
+    "copy-webpack-plugin": "^4.5.2",
```
levithomason, line 107: We need to copy `docs/src/public` to `docs/dist` on build.  Otherwise, the font files are not included.

src/components/Provider/Provider.tsx
```diff
@@ -92,15 +94,15 @@ class Provider extends React.Component<IProviderProps, any> {
 
     if (!fontFaces) return
 
-    const renderFontObject = font => {
+    const renderFontObject = (font: IFontFace) => {
       if (!_.isPlainObject(font)) {
         throw new Error(`fontFaces must be objects, got: ${typeof font}`)
       }
-      felaLtrRenderer.renderFont(font.name, font.path, font.style)
+      felaLtrRenderer.renderFont(font.name, font.paths, font.style)
```
levithomason, line 100: Fonts were also not being rendered because `path` was being used here instead of `paths`.  Added typings to ensure it doesn't happen again.

build/gulp/tasks/generate/templates/docs-examples-components/$DisplayName/Types/$DisplayNameExample.shorthand.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import { $DisplayName } from '@stardust-ui/react'
 
-const $DisplayNameExampleShorthand = () => <$DisplayName />
+const $DisplayNameExampleShorthand = () => <$DisplayName content="Hello from $DisplayName!" />
```
levithomason, line 4: Unrelated freebie.  The initial component example from a generated component had no content.  Added some here so something shows up in the docs.

levithomason:
This is blocking other work... merging

Issue #195, opened 2018-09-06
[RFC] Radio - create group options and add event handlers


alinais:
### Problem description
A clear and concise description of what the problem is. Ex. I have an issue when [...]

### Proposed solution
A clear and concise description of what you want to happen. Add any considered drawbacks.

### MVP
[Fork, update, and replace this pen if you can show the proposed solution]:
https://codesandbox.io/s/2l3n74j9y

jurokapsiar:
fixed in https://github.com/stardust-ui/react/pull/229

PR #229, opened 2018-09-14
feat(radio): Radio group compliant with ARIA patterns

jurokapsiar:
# RadioGroup

This is a ARIA-patterns compliant implementation of Radio and RadioGroup.
My initial testing shows that it works with all required screen reader/browser combinations.
The only exception is that the focused option is not highlighted with JAWS in virtual cursor mode.

This breaks the current Radio API as ``Radio`` is replaced with ``RadioGroup.Item``

### TODO

- [X] Conformance test
- [X] Minimal doc site example
- [X] Stardust base theme
- [X] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [X] Confirm RTL usage
- [X] [W3 accessibility](https://www.w3.org/standards/webdesign/accessibility) check
- [X] [Stardust accessibility](https://github.com/stardust-ui/accessibility) check
- [X] Update the CHANGELOG.md

# API Proposal

## [Group]

image

```jsx
<RadioGroup defaultCheckedIndex={0} items={items} onChange={this.handleChange} />
```

```html
  <div role="radiogroup" class="ui-radiogroup">
    <div role="radio" tabindex="0" aria-checked="true" index="0" class="ui-radio f g">
      <span class="ui-label ...">
        Capricciosa
      </span>
    </div>
  </div>
```

Comparison of approaches that I tried: (the last row is this implementation)
image


CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fixed missing colors in Teams' siteVariables @mnajdova ([#200](https://github.com/stardust-ui/react/pull/200))
 - Fixed Teams' siteVariables font sizes @levithomason ([#204](https://github.com/stardust-ui/react/pull/204))
 - Fixed docs examples of `Text` component @codepretty ([#205](https://github.com/stardust-ui/react/pull/205))
+- Add `value`, `disabled`, `checked`, `defaultChecked` and `onChange` props to `Radio` component @mnajdova ([#206](https://github.com/stardust-ui/react/pull/206))
 - Preserve fonts and static styles in mergeThemes @levithomason ([#217](https://github.com/stardust-ui/react/pull/217))
```
mnajdova: Please delete this from here, it was added in the unreleased section

src/components/Radio/RadioGroup.tsx
```diff
@@ -0,0 +1,151 @@
...
+  renderItems = (variables: ComponentVariablesObject) => {
+    const { items } = this.props
+    const { checkedIndex } = this.state
+
+    return _.map(items, (item, index) =>
+      Radio.create(item, {
+        defaultProps: {
+          index,
+          checked: parseInt(checkedIndex, 10) === index,
+        },
```
mnajdova: Maybe is better that we use the value for this instead of index? Is there reason why we are using index?

src/components/Radio/RadioGroup.tsx
```diff
@@ -0,0 +1,151 @@
...
+        overrideProps: this.handleItemOverrides,
+      }),
```
mnajdova: As we are explicitly handling the checked value, we are the one responsible for invoking the onChange handler in the Radio button when the checked value is changing.

src/components/Radio/Radio.tsx
```diff
@@ -31,13 +35,16 @@ export interface IRadioProps {
   onChange?: ComponentEventHandler<IRadioProps>
   type?: string
   styles?: IComponentPartStylesInput
+
+  // TODO: what is the difference between value and checked?
   value?: string | number
```
mnajdova: Multiple radio buttons in a group may have different values, we must have this in order to know which radio button is selected. The checked only indicates if this radio button is selected or not.

docs/src/examples/components/Radio/Groups/RadioExampleGroup.shorthand.tsx
```diff
@@ -0,0 +1,26 @@
...
+  }
+  render() {
+    const { selectedValue } = this.state
+    return (
+      <div>
+        The selected value is: {selectedValue}
+        <Divider />
+        <RadioGroup defaultCheckedIndex={0} items={items} onChange={this.handleChange} />
+      </div>
+    )
```
mnajdova: I see that we don't have circular navigation through the radio buttons in group when navigating with the keyboard arrows. Is this how they should behave?

docs/src/examples/components/RadioGroup/Item/RadioGroupItemExample.shorthand.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import { RadioGroup } from '@stardust-ui/react'
+
+const handleChange = () => {
+  alert('The radio checked value was changed!')
+}
+
+const RadioGroupItemExample = () => (
+  <RadioGroup.Item label="Make your choice" onChange={handleChange} />
+)
```
mnajdova: If the RadioGroupItem should not be defined without the RadioGroup, I would suggest even in this examples, that it should be wrapped with RadioGroup, because the user is most likely to copy-paste the example without reasoning that it should be inside RadioGruop.

src/components/Button/Button.tsx
```diff
@@ -196,9 +194,7 @@ class Button extends UIComponent<Extendable<IButtonProps>, any> {
   }
 
   private handleFocus = (e: React.SyntheticEvent) => {
-    const isFromKeyboard = whatInput.ask() === 'keyboard'
-
-    this.setState({ isFromKeyboard })
+    this.setState(isFromKeyboard.state())
```
mnajdova, line 198: Are these changes suppose to be part of this PR?

docs/src/examples/components/RadioGroup/Types/RadioGroupExample.shorthand.tsx
```diff
@@ -0,0 +1,35 @@
...
+        <Divider />
+        <RadioGroup
+          defaultCheckedValue="capricciosa"
+          items={items}
+          checkedValueChanged={this.handleChange}
+        />
+      </div>
+    )
+  }
+}
```
mnajdova, line 38: I'm confused why the exactly same result, degining the RadioGroup.items as a components is not working. I am testing with this example and expect to achieve the same: 
```jsx
import React from 'react'
import { Divider, RadioGroup } from '@stardust-ui/react'
const props = [
  { name: 'pizza', key: 'Capricciosa', label: 'Capricciosa', value: 'capricciosa' },
  { name: 'pizza', key: 'Prosciutto', label: 'Prosciutto', value: 'prosciutto', disabled: true },
  {
    name: 'pizza',
    key: 'Margherita',
    label: 'Margherita',
    value: 'margherita',
  },
]
const items = [
  <RadioGroup.Item {...props[0]} />,
  <RadioGroup.Item {...props[1]} />,
  <RadioGroup.Item {...props[2]} />
]

class RadioGroupExample extends React.Component {
  state = { selectedValue: '' }
  handleChange = (e, props) => {
    this.setState({ selectedValue: props.value })
  }
  render() {
    const { selectedValue } = this.state
    return (
      <div>
        The selected value is: {selectedValue}
        <Divider />
        <RadioGroup
          defaultCheckedValue="capricciosa"
          items={items}
          checkedValueChanged={this.handleChange}
        />
      </div>
    )
  }
}
export default RadioGroupExample

```

src/components/Button/Button.tsx
jurokapsiar: I will make a super small PR to separate this. not necessarily part of this one, but as I was using the same code as it's used in button, I just moved it to a separate function

docs/src/examples/components/RadioGroup/Types/RadioGroupExample.shorthand.tsx
jurokapsiar: I'm not sure if passing components in items array is supposed to work. If you would pass the components as children, it would partially work. However similarly to Menu, RadioGroup works correctly only when you use shorthand for passing the items.

docs/src/examples/components/RadioGroup/Item/RadioGroupItemExample.shorthand.tsx
sophieH29: Agree, the examples should be like for the Menu. Only RadioGroup together RadioGroupItem, without RadioGroupItem separately

docs/src/examples/components/RadioGroup/Types/RadioGroupExample.shorthand.tsx
mnajdova: This is still shorthand, it is just another way how the component can be passed. We can take a look together into this. I wasn't able to figure out yesterday what was the problem, but I am assuming is the casting of the item as IRadioGroupItemProps.

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,186 @@
...
+    if (nextItem) {
+      this.trySetState({ checkedValue: nextItem.value })
+      _.invoke(this.props, 'checkedValueChanged', event, nextItem)
+    }
+
+    event.preventDefault()
+  }
+
+  findNextEnabledCheckedItem = (direction): IRadioGroupItemProps => {
+    if (!this.props.items || this.props.items.length === 0) {
```
sophieH29: ```this.props.items.length === 0``` can be simplified to ```!this.props.items.length```

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,186 @@
...
+      newIndex !== currentIndex;
+      newIndex += direction
+    ) {
+      if (newIndex < 0) {
+        newIndex = this.props.items.length - 1
+      } else if (newIndex >= this.props.items.length) {
+        newIndex = 0
+      }
+
+      if (newIndex < 0 || newIndex >= this.props.items.length || newIndex === currentIndex) {
```
sophieH29: ```newIndex < 0 || newIndex >= this.props.items.length``` you have already checked it above, I think ```newIndex === currentIndex``` might be enough

src/components/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+    event.preventDefault()
+  }
+
+  findNextEnabledCheckedItem = (direction): IRadioGroupItemProps => {
+    if (!this.props.items || this.props.items.length === 0) {
+      return undefined
+    }
+
+    const currentIndex = this.props.items.findIndex(
```
sophieH29: ```findIndex``` is not supported in IE. Not sure if we are going to support IE for stardust
or you can use https://lodash.com/docs/4.17.10#findIndex with backward compatibility for IE

src/lib/accessibility/Behaviors/Radio/RadioBehavior.ts
```diff
@@ -0,0 +1,24 @@
...
+const RadioBehavior: Accessibility = (props: any) => ({
+  attributes: {
+    root: {
+      role: 'radio',
+      tabIndex: props.checked ? '0' : '-1',
+      'aria-checked': props.checked,
+      'aria-disabled': props.disabled,
+    },
+    label: {
+      role: 'presentation',
```
sophieH29: I don't see where it used as ```accessibility.attributes.label```

src/lib/accessibility/Behaviors/Radio/RadioGroupBehavior.ts
```diff
@@ -0,0 +1,29 @@
...
+    },
+  },
+
+  keyActions: {
+    root: {
+      nextItem: {
+        keyCombinations: [{ keyCode: keyboardKey.ArrowDown }, { keyCode: keyboardKey.ArrowRight }],
+      },
+      prevItem: {
+        keyCombinations: [{ keyCode: keyboardKey.ArrowUp }, { keyCode: keyboardKey.ArrowLeft }],
```
sophieH29, line 23: Shouldn't we make navigation based on ```ArrowRight``` and ```ArrowLeft``` for horizontal radio group, and ```ArrowUp``` and ```ArrowDown``` for the vertical? So we can define it based on props value ```vertical```

src/lib/accessibility/Behaviors/Radio/RadioBehavior.ts
```diff
@@ -0,0 +1,24 @@
+import { Accessibility } from '../../interfaces'
+
+/**
+ * @description
+ *  Adds role='radio'. This allows screen readers to handle the component as a radio button.
+ *  Adds attribute 'aria-checked=true' based on the property 'checked'.
+ *  Implements roving tabIndex
+ *  Adds role 'presentation' on label element
+ */
+const RadioBehavior: Accessibility = (props: any) => ({
```
sophieH29: And since there is no more Radio component but RadioGroupItem, might be good to rename also this behavior to ```RadioGroupItemBehavior``` to keep consistency. What do you think?

src/components/RadioGroup/RadioGroup.tsx
jurokapsiar: yes, it's checked twice, but it is changed in between these two ifs.

src/lib/accessibility/Behaviors/Radio/RadioGroupBehavior.ts
jurokapsiar: according to ARIA spec, you can use both up/down and left/right to switch between the radio buttons.

src/components/RadioGroup/RadioGroup.tsx
sophieH29: Sorry, I might be confused or blind, but as I see, you change it in between these two ifs to either last item index or first item index, which seems to be safe and no need to check it later again

src/lib/accessibility/Behaviors/Radio/RadioBehavior.ts
jurokapsiar: good catch, thanks

test/specs/commonTests/implementsShorthandProp.tsx
```diff
@@ -3,10 +3,12 @@ import { mount } from './isConformant'
 
 export type ShorthandTestOptions = {
   mapsValueToProp?: string
+  skipArrayOfStrings?: boolean
 }
```
mnajdova: Let's not change this interface, but create new one specific for the collection shorthand, as skipArrayOfStrings is not in any way used in the simple shorthand. The last option how the shorthand can be used is a function with signature: `(Component, props, children) => React.Node`. If we are not supporting this one, maybe we should add config for this one as well.

mnajdova:
As per our discussion the next step would be introducing variables for the RadioGroup with slot for the items, but we are leaving this for dedicated PR. In my opinion this is good to be merged now.

mnajdova (approved):

jurokapsiar:
fixes https://github.com/stardust-ui/react/issues/195

PR #233, opened 2018-09-14
feat(FocusZone): Add embed mode for FocusZone and new Chat behavior

tomasiser:
# Add embed mode for FocusZone and new Chat behavior

Hi, this PR does the following:

- For `renderComponent` and `UIComponent`:
   - Introduced `FocusZoneMode.Embed`, in which case FocusZone is not wrapping the component in an additional `<div>`, but instead uses the component directly in `render`. This is necessary for some components (e.g., `ChatMessage`) not to break generic styling etc.
   - Added `protected` `this.focusZone` reference to `UIComponent` available in components, so they can e.g., call `this.focusZone.focus()`.
- For `FocusZone` directly:
   - Replaced `onFocusNotification` with a regular `onFocus` event callback to pass unit tests with embed.
   - Replaced `ref={this.setRef}` with `this.setRef(this)` in `componentDidMount` to support functional components, which is needed to pass unit tests with embed.
   - Changed `defaultActiveElement` to query only descendants of the focus zone instead of the whole document, which enables to write simpler selectors. Note that we do not lose any functionality by this, because selecting elements outside of focus zone had no effect.
- For unit tests:
   - `isConformant` and `handlesAccessibility` now support embedded FocusZone.
- For `Chat` and `ChatMessage`:
   - Added new action handlers: `focus` for `Chat`, and `preventDefault` for `ChatMessage`.
   - Added new behaviors: wrapped FocusZone for `Chat`, embedded FocusZone for `ChatMessage`.
   - Behaviors are available in 2 variations:
      - Default: use up/down arrow keys to navigate to previous/next chat message, use right arrow to focus inside a message (e.g., to focus links, buttons), use Tab to navigate through inside, use left arrow to return outside a message again.
      - EnterEsc: everything as in Default, except use Enter/Escape to focus inside/outside of chat message, and use arrow keys to navigate through inside a message.
- Adds `<a>` links to `Chat` example in docsite so that we can easily test navigation there.


src/components/Chat/Chat.tsx
```diff
@@ -6,6 +6,8 @@ import { childrenExist, customPropTypes, UIComponent } from '../../lib'
 import ChatMessage from './ChatMessage'
 import { ComponentVariablesInput, IComponentPartStylesInput } from '../../../types/theme'
 import { Extendable, ReactChildren, ItemShorthand } from '../../../types/utils'
+import { Accessibility, AccessibilityActionHandlers } from '../../lib/accessibility/interfaces'
+import ChatBehavior from '../../lib/accessibility/Behaviors/Chat/ChatBehavior'
 
 export interface IChatProps {
   as?: any
```
sophieH29, line 14: please, add ```accessibility?: Accessibility``` to IChatProps

src/components/Chat/ChatMessage.tsx
```diff
@@ -6,6 +6,8 @@ import { childrenExist, createShorthandFactory, customPropTypes, UIComponent } f
 import { ComponentVariablesInput, IComponentPartStylesInput } from '../../../types/theme'
 import { Extendable, ReactChildren, ItemShorthand } from '../../../types/utils'
 import Avatar from '../Avatar'
+import ChatMessageBehavior from '../../lib/accessibility/Behaviors/Chat/ChatMessageBehavior'
+import { Accessibility, AccessibilityActionHandlers } from '../../lib/accessibility/interfaces'
 
 export interface IChatMessageProps {
   as?: any
```
sophieH29, line 26: Same here, add ```accessibility?: Accessibility``` to ```IChatMessageProps```

src/lib/UIComponent.tsx
```diff
@@ -38,10 +41,15 @@ class UIComponent<P, S> extends React.Component<P, S> {
...
         actionHandlers: this.actionHandlers,
+        focusZoneRef: this.setFocusZoneRef,
       },
       this.renderComponent,
     )
   }
+
+  private setFocusZoneRef(focusZone: IFocusZone) {
+    this.focusZone = focusZone
+  }
```
sophieH29, line 51: I think you can do 
```js
const setFocusZoneRef = (focusZone: IFocusZone) => this.focusZone = focusZone
```
So you wouldn't need to make binding in the constructor. 
But it's not necessary to make changes, you can leave as it is

src/lib/accessibility/Behaviors/Chat/ChatBehavior.ts
```diff
@@ -0,0 +1,31 @@
+import { Accessibility, FocusZoneMode } from '../../interfaces'
+import * as keyboardKey from 'keyboard-key'
+import { FocusZoneDirection } from '../../FocusZone'
+
+const CHAT_FOCUSZONE_ATTRIBUTE = 'chat-focuszone'
+
+const ChatBehavior: Accessibility = {
```
sophieH29, line 13: Please, add description to behavior as we're having in others behaviors :)

src/components/Chat/ChatMessage.tsx
```diff
@@ -62,18 +68,35 @@ class ChatMessage extends UIComponent<Extendable<IChatMessageProps>, any> {
...
   static defaultProps = {
+    accessibility: ChatMessageBehavior as Accessibility,
     as: 'li',
   }
 
-  renderComponent({ ElementType, classes, rest, styles }) {
+  actionHandlers: AccessibilityActionHandlers = {
+    preventDefault: event => {
+      event.preventDefault()
+    },
```
sophieH29, line 102: Having concerns regarding such action as it doesn't do some meaningful action. Is there any other way to avoid it here, and use it in event callback handlers?

src/lib/accessibility/FocusZone/FocusZone.tsx
```diff
@@ -261,12 +265,13 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
     return false
   }
 
-  private _onFocus = (ev: React.FocusEvent<HTMLElement>): void => {
-    const { onActiveElementChanged, stopFocusPropagation, onFocusNotification } = this.props
+  private setRef(elem) {
```
miroslavstastny: If you use an arrow function for setRef (see `_onFocus` bellow), you don't need to `bind` it in ctor.
I vote for arrow function which is more obvious when you read the function + it is consistent with the rest of the class.

src/lib/accessibility/FocusZone/FocusZone.tsx
```diff
@@ -290,6 +295,8 @@ export class FocusZone extends React.Component<IFocusZoneProps> implements IFocu
     if (stopFocusPropagation) {
       ev.stopPropagation()
     }
+
+    _.invoke(this.props, 'onFocus', ev)
```
miroslavstastny, line 302: Why have you moved this from the top of the function to the bottom?

src/lib/accessibility/Behaviors/Chat/ChatBehavior.ts
```diff
@@ -0,0 +1,31 @@
...
+    root: {
+      role: 'presentation',
+    },
+  },
+  focusZone: {
+    mode: FocusZoneMode.Wrap,
+    props: {
+      shouldEnterInnerZone: event => keyboardKey.getCode(event) === keyboardKey.ArrowRight,
+      direction: FocusZoneDirection.vertical,
+      defaultActiveElement: `[${CHAT_FOCUSZONE_ATTRIBUTE}] > * > *:last-child`, // select last chat message by default
```
miroslavstastny: Whenever a chat is rendered, focus is set to it's last element.
That means that playing with chat in docsite is hardly usable.
Plus is this the behavior we want in the real app?

src/lib/accessibility/Behaviors/Chat/ChatBehavior.ts
```diff
@@ -0,0 +1,31 @@
...
+      [CHAT_FOCUSZONE_ATTRIBUTE]: '', // allows querying the default active element
+    },
+  },
+  keyActions: {
+    root: {
+      focus: {
```
miroslavstastny, line 30: Is this ever used as a `focus` action? Or is it more something like `regainFocus`?

src/components/Chat/Chat.tsx
```diff
@@ -39,17 +44,38 @@ class Chat extends UIComponent<Extendable<IChatProps>, any> {
...
 
   static Message = ChatMessage
 
-  renderComponent({ ElementType, classes, rest }) {
+  actionHandlers: AccessibilityActionHandlers = {
+    focus: event => {
+      if (this.focusZone) {
+        this.focusZone.focus()
+      }
+    },
```
sophieH29: nit, can be one line 
```js
focus: event => this.focusZone && this.focusZone.focus()
```

src/lib/accessibility/FocusZone/FocusZone.tsx
tomasiser: Good point. I did this because I changed the meaning of this callback to conform to unit tests. The event should now behave as a transparent `onFocus` callback, i.e., you can do `<FocusZone onFocus={...}>...</FocusZone>` and in that case, I personally expect the callback to be called *after* the component itself handles the callback. This is quite similar to bubbling in a sense that the event is first handled inside, then it goes outwards.

We can discuss this offline if you want.

src/components/Chat/ChatMessage.tsx
tomasiser: I added comments explaining why we use `preventDefault` in this case. Apart from that, I could not come up with a better place for this. @jurokapsiar agrees we can keep this.

src/lib/accessibility/Behaviors/Chat/ChatBehavior.ts
tomasiser: It is used when you want to unfocus a message and focus the whole list again, so Juraj and I think that `focus` is an appropriate action name as the only thing it does is it literally focuses the chat. :)

src/lib/accessibility/FocusZone/FocusZone.tsx
miroslavstastny: Makes sense. I just would like to be consistent. Currently, `Portal`, `Input`, maybe others call `onXXX()` before `trySetState`, but that's out of scope of this PR.

miroslavstastny (approved):

PR #236, opened 2018-09-14
[WIP] fix: Layout nested in vertical Layout

levithomason:
Blocking #255 

The goal of this PR is to make the Layout component compliant with our other components:

- Use `styles` prop instead of `rootCSS`, etc.
- Use `Layout.Area` subcomponents instead of hard coded markup
- Use `display: flex` instead of `display: grid` as this is a 2d layout component

The Layout and LayoutArea components should behave just like the rest of our components and should have the same features.


bmdalex:
@levithomason trying to pick this up but merge is very difficult with all the updated; is it worth merging or better start over?

levithomason:
@Bugaa92 We are removing layout from our components and doing more investigation around the `template` prop and this PR is quite old / has conflicts.  I'm closing for housekeeping.

PR #238, opened 2018-09-15
BREAKING: apply styles prop to the root element

levithomason:
# Before

The `styles` prop accepted a map of component slots to styles:

```jsx
<Button 
  styles={{ 
    root: { background: 'blue' },
    icon: { color: 'white' },
   }}
  icon='user'
/>
```

# After

The `styles` prop now styles the root element only.  Use shorthand props to style component slots:

```jsx
<Button
  styles={{ background: 'blue' }}
  icon={{ name: 'user', styles: { color: 'white' } }}
/>
```

### DIFF
--- a/src/lib/renderComponent.tsx
+++ b/src/lib/renderComponent.tsx
@@ -23,6 +23,7 @@ import {
   IAccessibilityDefinition,
   AccessibilityActionHandlers,
   FocusZoneMode,
+  Accessibility,
 } from './accessibility/interfaces'
 import { DefaultBehavior } from './accessibility'
 import getKeyDownHandlers from './getKeyDownHandlers'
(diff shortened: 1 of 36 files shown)


src/lib/renderComponent.tsx
```diff
@@ -126,8 +127,13 @@ const renderComponent = <P extends {}>(
...
         // Resolve styles using resolved variables, merge results, allow props.styles to override
-        const mergedStyles = mergeComponentStyles(componentStyles[displayName], props.styles)
-        const accessibility = getAccessibility(stateAndProps, actionHandlers)
+        const mergedStyles: IComponentPartStylesPrepared = mergeComponentStyles(
+          componentStyles[displayName],
+          {
+            root: props.styles,
+          },
+        )
+        const accessibility: Accessibility = getAccessibility(stateAndProps, actionHandlers)
```
kuzhelov, line 136: 👍

test/specs/commonTests/implementsShorthandProp.tsx
```diff
@@ -49,20 +49,6 @@ export default Component => {
...
-        const wrapper = mount(
-          <Component {...props} styles={{ [shorthandPropertyName]: { foo: 'bar' } }} />,
-        )
-        const shorthandComponentProps = wrapper.find(ShorthandComponent.displayName).props()
-
-        expect(shorthandComponentProps.styles).toBeDefined()
-        expect(shorthandComponentProps.styles.root).toEqual({ foo: 'bar' })
-      })
-
       test(`shorthand's variables may be passed as '${shorthandPropertyName}' prop of ${
```
kuzhelov, line 52: yes, in that case we should seriously think about this `variables` guy as well..

kuzhelov (approved):

PR #239, opened 2018-09-17
[RFC] feat(Accessibility): Add focus trap zone

sophieH29:
This PR introduces ```FocusTrapZone```.

```FocusTrapZone``` is taken from ```ui-fabric``` library as ```FocusZone```, and was changed a bit to be consistent with a Stardust library.

Notes:
- ```FocusTrapZone``` is only used in a ```Portal``` component and is diabled by default.
- ```FocusTrapZone``` will wrap a ```Portal``` component, in case ```focusTrap``` is set to ```true```
- to configure ```FocusTrapZone```, additional prop ```focusTrapZoneProps``` should be passed to ```Portal``` component. Later, in can be configered through accessibility behaviors of the components which will use ```Portal```, like ```Popup```.
- if ```FocusTrapZone``` is enabled, by default it will have ```isClickableOutsideFocusTrap``` prop set to ```true```.
- on ```componentDidMount```, ```FocusTrapZone``` will hide all other elements in the body from accessibility tree, by setting attribute ```aria-hidden=true```, and will remove it on ```componentWillUnmount```

### CODE REVIEW

File: docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.tsx [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+import { Button, Header, Portal } from '@stardust-ui/react'
+
+class PortalExamplePortal extends React.Component {
+  render() {
+    return (
+      <Portal
+        trigger={
+          <Button
+            content={'Toggle portal'}
+            trapFocus={true}
```
jurokapsiar: is this example correct? should the trapFocus be set on the trigger Button?

sophieH29: you're right, it's not correct. Was testing with shorthand example. Thank you that you caught it!

src/components/Portal/Portal.tsx
```diff
@@ -23,7 +24,8 @@ export interface IPortalProps {
   onUnmount?: (props: IPortalProps) => void
   open?: boolean
   trigger?: JSX.Element
-  triggerRef?: (node: HTMLElement) => void
+  trapFocus?: boolean
+  focusTrapZoneProps?: IFocusTrapZoneProps
```
jurokapsiar: we should consider moving these two props to behavior. let's discuss

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,304 @@
+import * as React from 'react'
```
jurokapsiar, line 1: similarly to FocusZone, lets create a CHANGELOG.md file in this directory to summarize what changes we did to FocusTrapZone in comparison to the original fabric version. no need to do it imediatelly but after others review and provide feedback and before we merge this PR.

src/components/Portal/Portal.tsx
sophieH29: yes, we can make it in future PR, once a PopupBehavior get merged

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: ok, will do that, once a code review will pass, before merging

docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.shorthand.tsx
```diff
@@ -0,0 +1,36 @@
...
+            <p tabIndex={0}>Portals have tons of great callback functions to hook into.</p>
+            <p tabIndex={0}>To close, simply click the close button or click away</p>
+            <Button size="small" content="Button" />
+            <Button size="small" content="Another button" />
+          </div>
+        }
+        trigger={<Button content={'Toggle portal'} />}
+        trapFocus={true}
+        focusTrapZoneProps={{
+          isClickableOutsideFocusTrap: true,
```
miroslavstastny: If this is set to `false` any click outside resets focus to the first element in the portal. Is that ok? @kolaps33

miroslavstastny: `true` is the default value. I don't see the point in setting it here explicitly.

sophieH29: yes, I know, there is no point to set it explicitly. Just wanted to show an example of how focus trap zone props are passed to the component. I would then better make an example with ```false``` value

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,304 @@
...
+      previouslyFocusedElement = document.activeElement as HTMLElement
+    }
+
+    return previouslyFocusedElement
+  }
+
+  private _hideContentFromAccessibilityTree = () => {
+    const elements = (document.body && document.body.children) || []
+
+    // loop through all body's children, except the last one - which is the portal
```
miroslavstastny: Don't mention portal here as this should be generic.

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,304 @@
...
+    for (let index = 0; index < elements.length - 1; index++) {
```
miroslavstastny: So the trap zone can be used only for the Component that is directly in body?
This seems like a black magic to me, I would prefer using some kind of id.

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,304 @@
...
+      if (element.nodeName !== 'SCRIPT' && element.nodeName !== 'STYLE') {
+        element.setAttribute('aria-hidden', 'true')
+        element.setAttribute(HIDDEN_FROM_ACC_TREE, 'true')
+      }
+    }
+  }
+
+  private _showContentInAccessibilityTree = () => {
+    const hiddenElements = document.querySelectorAll(`[${HIDDEN_FROM_ACC_TREE}="true"]`)
+    for (let index = 0; index < hiddenElements.length; index++) {
```
miroslavstastny, line 339: If there is an element which has `aria-hidden` set before the trap zone is created, this removes the attribute.

miroslavstastny:
Do we support nested trap zones? Was that tested?

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: Well, currently Portal is appended only to the body, and focus trap is part of it. And seems like it's not a good practice to create HTML element outside of the body.
But to be safe, we expand FocusTrap zone props, and allow to pass a specific element, and select document.body by default. @miroslavstastny  What do you think?

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: That's why I've added specific attribute HIDDEN_FROM_ACC_TREE, which is added only when the popup was mounted and we want to hide all content form Accessibility tree.
And then we remove 'aria-hidden' only for those element who has set ```${HIDDEN_FROM_ACC_TREE}="true"``` and if it was the last mounted popup with trap zone (if you look at the beginning, focus trap zone are added to stack when element is mointed)

sophieH29:
> Do we support nested trap zones? Was that tested?

I have tested such case when you put one popup into another and popups will be appended separately to the body. So the focus gets the last mounted focus trap zone and the rest (including parent popup, which is a sibling in reality) will be hidden from accessibility tree.

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -117,9 +117,16 @@ export class FocusTrapZone extends React.Component<IFocusTrapZoneProps, {}>
...
+    const lastActiveFocusTrap = FocusTrapZone._focusStack[FocusTrapZone._focusStack.length - 1]
 
     if (!FocusTrapZone._focusStack.length) {
       this._showContentInAccessibilityTree()
+    } else if (
+      lastActiveFocusTrap._root.current &&
+      lastActiveFocusTrap._root.current.hasAttribute(HIDDEN_FROM_ACC_TREE)
+    ) {
+      lastActiveFocusTrap._root.current.removeAttribute(HIDDEN_FROM_ACC_TREE)
+      lastActiveFocusTrap._root.current.removeAttribute('aria-hidden ')
```
miroslavstastny: Not sure about the space in `'aria-hidden '`

sophieH29: ha, nice catch. Thanks!

src/components/Portal/Portal.tsx
jurokapsiar: I have created https://github.com/stardust-ui/react/issues/354 for this

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
jurokapsiar: I have added information to FocusZone CHANGELOG.md

src/lib/accessibility/FocusZone/CHANGELOG.md
```diff
@@ -35,4 +35,16 @@ This is a list of changes made to this Stardust copy of FocusZone in comparison
...
-    - Focus related utilities moved to `focusUtilities.ts`.
\ No newline at end of file
+    - Focus related utilities moved to `focusUtilities.ts`.
+
+## FocusTrapZone Changelog
+
+This is a list of changes made to the Stardust copy of FocusTrapZone in comparison with the original [Fabric FocusTrapZone @ 0f567e05952c6b50c691df2fb72d100b5e525d9e](https://github.com/OfficeDev/office-ui-fabric-react/blob/0f567e05952c6b50c691df2fb72d100b5e525d9e/packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx).
+
+### feat(Accessibility): Add focus trap zone [#239](https://github.com/stardust-ui/react/pull/239)
+- using Stardust utils instad of Fabric utilities
```
kuzhelov: nit: typo 'instad'

docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.tsx
```diff
@@ -0,0 +1,51 @@
...
+              position: 'fixed',
+              left: '40%',
+              top: '45%',
+              zIndex: 1000,
+              padding: '10px',
+            }}
+          >
+            <Header>This is a portal with focus trap!</Header>
+            <p tabIndex={0}>Portals have tons of great callback functions to hook into.</p>
+            <p tabIndex={0}>To close, simply click the close button or click away</p>
```
kuzhelov: trying to click away, portal is not closing - seems that we have a bit if contradiction here :)

src/components/Portal/Portal.tsx
kuzhelov: yes, this is necessary to clean up the Portal's API, sooner the better

docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.tsx
```diff
@@ -0,0 +1,51 @@
...
+            style={{
+              background: '#ddd',
+              position: 'fixed',
+              left: '40%',
+              top: '45%',
+              zIndex: 1000,
+              padding: '10px',
+            }}
+          >
+            <Header>This is a portal with focus trap!</Header>
```
kuzhelov: nit: probably, it would be better to make it a bit more apparent with a text like: `This portal traps focus on appearance`, to make it more evident what stays behind `with a focus trap` words

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,322 @@
...
+
+export class FocusTrapZone extends React.Component<IFocusTrapZoneProps, {}>
+  implements IFocusTrapZone {
+  private static _focusStack: FocusTrapZone[] = []
+  private _root: { current: HTMLElement | null } = { current: null }
+  private _previouslyFocusedElementOutsideTrapZone: HTMLElement
+  private _previouslyFocusedElementInTrapZone?: HTMLElement
+  private windowElement: Window | null
+
+  private createRef = elem => (this._root.current = ReactDOM.findDOMNode(elem) as HTMLElement)
```
kuzhelov, line 32: is it intentional that you would like to return dom node from this call? According to the place where method is used, its not. In that case we should consider to use the following form instead:

```js
private createRef = elem => { this._root.current = ReactDOM.findDOMNode(elem) as HTMLElement }
```

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,322 @@
...
+      this.focus()
+    }
+
+    this._hideContentFromAccessibilityTree()
+    this._subscribeToEvents()
+  }
+
+  public render(): JSX.Element {
+    const { className, ariaLabelledBy } = this.props
+    const rest = getUnhandledProps({ handledProps: FocusTrapZone.handledProps }, this.props)
```
kuzhelov: please, consider to reuse approach suggested in `UIComponent` to fetch handle props, to remove them being hardcoded. Also, it would be great to encapsulate this as a lib function

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,322 @@
...
+      >
+        {this.props.children}
+      </ElementType>
+    )
+  }
+
+  public componentWillUnmount(): void {
+    const { ignoreExternalFocusing } = this.props
+
+    FocusTrapZone._focusStack = FocusTrapZone._focusStack.filter((value: FocusTrapZone) => {
```
kuzhelov, line 141: could you, please, suggest which memory issues that will solve? trying to imagine the situation with obsolete reference, but given that `_focusStack` is an instance field it is hard for me to come up with any. Are there any hidden event-based interconnections that could be created?

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,322 @@
...
+      return this !== value
+    })
+
+    const activeElement = document.activeElement as HTMLElement
+    if (
+      !ignoreExternalFocusing &&
+      this._previouslyFocusedElementOutsideTrapZone &&
+      typeof this._previouslyFocusedElementOutsideTrapZone.focus === 'function' &&
```
kuzhelov: which condition (semantically) we would like to check with this expression? The one that `_previouslyFocusedElementOutsideTrapZone` exists/is defined?

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,322 @@
...
+    // loop through all body's children, except the last one - which is the popup
+    for (let index = 0; index < elements.length - 1; index++) {
+      const element = elements[index]
+
+      if (
+        element.nodeName !== 'SCRIPT' &&
+        element.nodeName !== 'STYLE' &&
+        element.getAttribute('aria-hidden') !== 'true'
+      ) {
+        element.setAttribute('aria-hidden', 'true')
```
kuzhelov, line 331: don't we want to introduce constant for `aria-hidden` as well? What is the criteria that drives this decision boundary?

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,322 @@
...
+        element.setAttribute(HIDDEN_FROM_ACC_TREE, 'true')
+      }
+    }
+  }
+
+  private _showContentInAccessibilityTree = () => {
+    const hiddenElements = document.querySelectorAll(`[${HIDDEN_FROM_ACC_TREE}="true"]`)
+    for (let index = 0; index < hiddenElements.length; index++) {
```
kuzhelov, line 339: nit: `hiddenElements.forEach((hiddenElement, index) => ...)`

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,322 @@
...
+      previouslyFocusedElement = elementToFocusOnDismiss
+    } else if (!previouslyFocusedElement) {
+      previouslyFocusedElement = document.activeElement as HTMLElement
+    }
+
+    return previouslyFocusedElement
+  }
+
+  private _hideContentFromAccessibilityTree = () => {
+    const elements = (document.body && document.body.children) || []
```
kuzhelov: it would be better to name this variable as `bodyChildren` - this would make the intent of this method easier to grasp

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,322 @@
...
+
+    if (elements.length && !document.body.contains(this._root.current)) {
+      // In case popup render options will change
+      throw new Error(
+        'Body does not contain trap zone element as expected. If it is done intentionally, please, make sure to update FocusTrapZone.',
```
kuzhelov: sorry - could you, please, elaborate a bit further on what could be the case for it? Just describe the one where you see this could happen. Thanks!

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,322 @@
...
+      )
+    }
+
+    // loop through all body's children, except the last one - which is the popup
+    for (let index = 0; index < elements.length - 1; index++) {
+      const element = elements[index]
+
+      if (
+        element.nodeName !== 'SCRIPT' &&
```
kuzhelov: `<template>` in addition? could you, please, suggest the source/reference of this 'exceptions list' - or, essentially, your intuition behind what should be put into? Just to, potentially, search for the ways how it could be easier maintained

src/lib/accessibility/FocusZone/focusUtilities.ts
```diff
@@ -399,3 +400,46 @@ export function isElementFocusSubZone(element?: HTMLElement): boolean {
...
+  }
+}
+
+/**
+ * Helper to get the window object.
+ *
+ * @public
+ */
+export function getWindow(rootElement?: Element | null): Window | undefined {
+  return rootElement && rootElement.ownerDocument && rootElement.ownerDocument.defaultView
```
kuzhelov: could be simplified: `return rootElement && rootElement.ownerDocument && rootElement.ownerDocument.defaultView || window`

kuzhelov (changes requested):
lets please, discuss/address the comments mentioned

docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.tsx
jurokapsiar: well this used to work when the PR was created :)

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
jurokapsiar: that will unfortunately not work: `querySelectorAll isn't an array, it's a (non-live) NodeList, and not all browsers support forEach on NodeList's`

https://css-tricks.com/snippets/javascript/loop-queryselectorall-matches/

docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.tsx
sophieH29: will take a look

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: right, good catch

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: Usually, we move things to constants when we want to have 1 point for change, in case it will change in the future. 'aria-hidden' will always stay the same and I don't see any reason to create a const instead of a compile-time static string

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: yes, plus for loop is faster then forEach :)

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: agree

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: This one we agreed with @miroslavstastny. So in case, someone will use Portal outside of the body, we have to notify that FocusTrapZone would not work in that case.

docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.tsx
sophieH29: my bad, this example shows that popup woudn't be closed on outside click, only on button "close" click as I set ```isClickableOutsideFocusTrap: false``` (by default it is true). Will update the text

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: Did you mean re-assigning _focusStack value or filtering itself?
There is no any hidden event-based interconnections known to me. This part is originally from ui-fabric lib.

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: that is defined and has focus function.
I think this can be enough

```
     this._previouslyFocusedElementOutsideTrapZone &&
      this._previouslyFocusedElementOutsideTrapZone.focus
```
What do you think?

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: I vote for removing this check as it doesn't have any impact for now

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: ok, will do that

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
kuzhelov: ok - but in that case shouldn't it be provided as a warning? The thing is that current behavior will result in failed rendering for React, and there won't be **_any_** tree rendered.

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
kuzhelov: cannot take `faster` argument when we are using `{...obj, ...anotherObj }` expressions everywhere - and those introduce extremely low overhead in contrast to actual rendering or even requests to DOM elements (like `query` ones) :) At the same time, correctness point is very appealing, thanks!

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
kuzhelov: yes, the question then is `which attributes we are deciding to put to constants?` What is the semantics behind them - because even for `aria-hidden`, _in general case_, it is possible to argue that another day we might decide that another attribute should be put to all these places, so we would like to have single point of change. Just to clarify this moment

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
kuzhelov: yes, that would be better and its intent will be more readable - or, in case if more sophisticated logic is still necessary (like check that the value is function - although, not sure why it would be) - we might consider to encapsulate tis logic by providing function with a meaningful name, like:

```js
const hasFocus = (element) => element.focus === function

// and later use it
if (... && element && hasFocus(element)) { ... }
```

This will make the intent more readable.

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
kuzhelov: specifically mean the assignment, yes - as it is done, most probably, for the sake of preventing references to deleted objects. To me it seems that this assignment won't do anything useful for it - lets take a look on the source code for ui-fabric where it was introduced, just to infer the reason for it.

sophieH29: Actually, there is no difference of _focusStack usage in fabric-ui and current focusTrapZone.tsx
https://github.com/OfficeDev/office-ui-fabric-react/blob/master/packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx

I think the reason they've done that way it's to be safe in case ```componentWillUnmount``` will be invoked not for the last instance of FTZ in the stack. So they just filter out that reference and assign the rest to the stack. And nothing more specific than that.

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: IMHO, the name would be better ```isFocusable```. Cause ```hasFocus``` it's more about a state, not ability. But I now think about skipping this check. As we never done this across app, to keep consistency.

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: I see the answer to this question as we are putting attributes to constants if they are custom or we definitely see it as a will-change attribute. Adding a constant for ```aria-hidden``` as ```ariaHidden``` for me is more like giving names for the sake of names. My vision on this is to move it to constants (it is also related to any change) with the real reason. If we decide to move this to constants, then we should do it for every attribute used across the app, especially in behaviors.

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: agree with you, I will change it to a warning, to not interrupt rendering of React

src/components/Portal/Portal.tsx
```diff
@@ -119,14 +123,20 @@ class Portal extends AutoControlledComponent<IPortalProps, IPortalState> {
   }
 
   private renderPortal(): JSX.Element | undefined {
-    const { children, content } = this.props
+    const { children, content, trapFocus } = this.props
```
kuzhelov, line 127: 👍

src/lib/accessibility/FocusZone/CHANGELOG.md
```diff
@@ -35,4 +35,16 @@ This is a list of changes made to this Stardust copy of FocusZone in comparison
...
+## FocusTrapZone Changelog
+
+This is a list of changes made to the Stardust copy of FocusTrapZone in comparison with the original [Fabric FocusTrapZone @ 0f567e05952c6b50c691df2fb72d100b5e525d9e](https://github.com/OfficeDev/office-ui-fabric-react/blob/0f567e05952c6b50c691df2fb72d100b5e525d9e/packages/office-ui-fabric-react/src/components/FocusTrapZone/FocusTrapZone.tsx).
+
+### feat(Accessibility): Add focus trap zone [#239](https://github.com/stardust-ui/react/pull/239)
+- using Stardust utils instead of Fabric utilities
+  - uses `eventStack`
+  - extends `React.Component` instead of Fabric `BaseComponent`
+  - uses `ReactDOM.findDOMNode` reference instead of `createRef` for `_root`
+- does not use `componentWillMount`
```
kuzhelov: nit: `componentWillMount` is not used anymore

Probably, this would be better, as it better reflects that this is a _change_ that was made.

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,306 @@
...
+    if (
+      !ignoreExternalFocusing &&
+      this._previouslyFocusedElementOutsideTrapZone &&
+      (this._root.current.contains(activeElement) || activeElement === document.body)
+    ) {
+      focusAsync(this._previouslyFocusedElementOutsideTrapZone)
+    }
+
+    this._unsubscribeFromEvents()
+    const lastActiveFocusTrap = FocusTrapZone._focusStack[FocusTrapZone._focusStack.length - 1]
```
kuzhelov: this better be moved to `else if` branch

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,306 @@
...
+      lastActiveFocusTrap._root.current.removeAttribute('aria-hidden')
+    }
+  }
+
+  public focus() {
+    if (!this._root.current) return
+    const { focusPreviouslyFocusedInnerElement, firstFocusableSelector } = this.props
+
+    if (
+      focusPreviouslyFocusedInnerElement &&
```
kuzhelov, line 175: note that these checks duplicate the ones that are provided on line 100 of this file - it would be better to encapsulate it:

```js
tryFocusPrevioslyFocusedElement(insideTrapZone: boolean) { 
  .. if (...) { focusAsync(...) } 
}
```

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,306 @@
...
+      this._previouslyFocusedElementInTrapZone &&
+      this._root.current.contains(this._previouslyFocusedElementInTrapZone)
+    ) {
+      // focus on the last item that had focus in the zone before we left the zone
+      focusAsync(this._previouslyFocusedElementInTrapZone)
+      return
+    }
+
+    const focusSelector =
```
kuzhelov, line 184: nit: this, probably, would be easier to read, as `null` case will be handled for both branches as a first thing

```js
 const focusSelector = firstFocusableSelector &&
      typeof firstFocusableSelector === 'string'  ? firstFocusableSelector : firstFocusableSelector()
```

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,306 @@
...
+      typeof firstFocusableSelector === 'string'
+        ? firstFocusableSelector
+        : firstFocusableSelector && firstFocusableSelector()
+
+    let _firstFocusableChild
```
kuzhelov: never seen naming convention where locally scoped variables were prefixed by underscore - could you, please, suggest the reasons behind that?

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,306 @@
...
+
+    if (focusSelector) {
```
kuzhelov: this could be simplified - as an additional benefit we are eliminating mutable `let` semantics for the variable

```js
const firstFocusableChild = focusSelector
  ? this._root.current.querySelector('.' + focusSelector)
  : getNextElement(
        this._root.current,
        this._root.current.firstChild as HTMLElement,
        true,
        false,
        false,
        true,
      )
    }
```

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,306 @@
...
+    )
+    const _lastTabbableChild = getLastTabbable(
+      this._root.current,
+      this._root.current.lastChild as HTMLElement,
+      true,
+    )
+
+    if (ev.shiftKey && _firstTabbableChild === ev.target) {
+      focusAsync(_lastTabbableChild)
+      ev.preventDefault()
```
kuzhelov, line 237: these two lines are the same for both logical branches - lets not duplicate them and put out of scope of `if-else`

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,306 @@
...
+      ev.stopPropagation()
+    } else if (!ev.shiftKey && _lastTabbableChild === ev.target) {
+      focusAsync(_firstTabbableChild)
```
kuzhelov, line 240: could you, please, suggest what drives decision boundary between sync and async focusing?

kuzhelov (commented):
please, see comments

docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.shorthand.tsx
```diff
@@ -0,0 +1,73 @@
...
+  render() {
+    const { open } = this.state
+    const btnContent = open ? 'Close Portal' : 'Open Portal'
+
+    return (
+      <div>
+        <Button content={btnContent} onClick={this.openPortal} />
+        <Portal
+          trapFocus={{
+            // When 'false', all clicks outside the Portal will be caught and not handled.
```
kuzhelov, line 24: 👍

src/components/Portal/Portal.tsx
```diff
@@ -119,14 +124,20 @@ class Portal extends AutoControlledComponent<PortalProps, PortalState> {
...
+    const contentToRender = childrenExist(children) ? children : content
+    const focusTrapZoneProps = (_.keys(trapFocus).length && trapFocus) || {}
 
     return (
       open && (
         <Ref innerRef={this.handlePortalRef}>
           <PortalInner onMount={this.handleMount} onUnmount={this.handleUnmount}>
-            {childrenExist(children) ? children : content}
+            {trapFocus ? (
+              <FocusTrapZone {...focusTrapZoneProps}>{contentToRender}</FocusTrapZone>
```
kuzhelov, line 137: 👍

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,296 @@
...
+  private static _focusStack: FocusTrapZone[] = []
+  private _root: { current: HTMLElement | null } = { current: null }
+  private _previouslyFocusedElementOutsideTrapZone: HTMLElement
+  private _previouslyFocusedElementInTrapZone?: HTMLElement
+  private windowElement: Window | null
+
+  private createRef = elem => (this._root.current = ReactDOM.findDOMNode(elem) as HTMLElement)
+
+  static propTypes = {
+    as: customPropTypes.as,
```
kuzhelov, line 38: wouldn't we provide any descriptions for these? Thos will be helpful even from the dev perspective

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,296 @@
...
+    } else if (
+      lastActiveFocusTrap._root.current &&
+      lastActiveFocusTrap._root.current.hasAttribute(HIDDEN_FROM_ACC_TREE)
+    ) {
+      lastActiveFocusTrap._root.current.removeAttribute(HIDDEN_FROM_ACC_TREE)
+      lastActiveFocusTrap._root.current.removeAttribute('aria-hidden')
+    }
+  }
+
+  private _findElementAndFocusAsync = () => {
```
kuzhelov, line 170: 👍

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,296 @@
...
+        target: this.windowElement,
+        useCapture: true,
+      })
+    }
+  }
+
+  private _unsubscribeFromEvents = () => {
+    const { forceFocusInsideTrap, isClickableOutsideFocusTrap } = this.props
+    if (forceFocusInsideTrap) {
+      eventStack.unsub('focus', this._handleOutsideFocus, {
```
kuzhelov, line 289: not related to this PR, but this is so prone to accidental mistakes like in the following cases:

#### Case 1
```js
eventStack.sub('focus', this._handleOutsideFocus, { target: window })
...
// nothing is unsubscribed, by default `document` is used as a target
eventStack.unsub('focus', e => this._handleOutsideFocus(e))
```

#### Case 2
```js
eventStack.sub('focus', e => this._handleOutsideFocus(e))
...
// oops, nothing is unsubscribed
eventStack.unsub('focus', e => this._handleOutsideFocus(e))
```

#### Case 3
```js
private _handleOutsideFocus(e) {
  ...
  // oops, I am not 'this' here (thoughts of the type's instance)
  this.handle(e)
}
...
eventStack.sub('focus', this._handleOutsideFocus)
```

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: Descriptions for props is provided here ```FocusTrapZone.types.tsx```. Do you think we should duplicate it?

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,296 @@
...
+      previouslyFocusedElement = elementToFocusOnDismiss
+    } else if (!previouslyFocusedElement) {
+      previouslyFocusedElement = document.activeElement as HTMLElement
+    }
+
+    return previouslyFocusedElement
+  }
+
+  private _hideContentFromAccessibilityTree = () => {
+    const bodyChildren = (document.body && document.body.children) || []
```
kuzhelov, line 317: lets explicitly put in comments somewhere (probably, in comments to `FocusTrapZone` class) the fact that this is a guy who is able to work in browser environment only

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -0,0 +1,296 @@
...
+
+    if (bodyChildren.length && !document.body.contains(this._root.current)) {
+      // In case popup render options will change
+      console.warn(
+        'Body does not contain trap zone element as expected. If it is done intentionally, please, make sure to update FocusTrapZone.',
```
kuzhelov: still, this message is a bit confusing - as we need just to ask client to use ensure that trap zone element is placed inside body. Lets express exactly the same thought in the warning message:

> Body does not contain trap zone element as expected - ensure that trapped DOM element is a child of <body />.

src/lib/accessibility/FocusZone/focusUtilities.ts
```diff
@@ -399,3 +400,46 @@ export function isElementFocusSubZone(element?: HTMLElement): boolean {
...
+      return
+    }
+
+    targetToFocusOnNextRepaint = element
+
+    const win = getWindow(element as Element)
+
+    if (win) {
+      // element.focus() is a no-op if the element is no longer in the DOM, meaning this is always safe
+      win.requestAnimationFrame(() => {
```
kuzhelov, line 426: seems that `setTimeout` would be enough to achieve this goal - why we are using animation frames?

kuzhelov: as an additional benefit of this approach - we might use handler object returned from this method to avoid potential situations where callback will be called multiple times (as multiple subscriptions might be made)

src/lib/accessibility/types.ts
```diff
@@ -140,12 +140,17 @@ export type FocusZoneDefinition = {
...
+  props?: FocusTrapZoneProps
+}
+
 export type KeyActions = { [partName: string]: { [actionName: string]: KeyAction } }
 export interface AccessibilityDefinition {
   attributes?: AccessibilityAttributesBySlot
   keyActions?: KeyActions
   handledProps?: (keyof AccessibilityAttributes)[]
   focusZone?: FocusZoneDefinition
+  focusTrapZone?: FocusTrapZoneDefinition
```
kuzhelov, line 153: note that this becomes harder for the client to immediately get an idea of why these two are separated and what, actually, drives this division - as both are related to focus handling aspects. Maybe we should take it as a note for the future and address this (API interface) problem

kuzhelov (approved):
👍

Issue #231, opened 2018-09-14
Prop-dependent variables


kuzhelov:
### Where we are now

Agreed on the following task (https://github.com/stardust-ui/react/issues/231#issuecomment-421973984) to be solved first by trying to prototype each of the suggested approaches. Here are prototype links to try:

- CSS-based approach prototype (#240)
- JS-based approach prototype (_about to come_)

-------------

### Problem description
After syncing up with Fabric there were some concerns raised around our taken approach to use `props` as variables function' argument:

```jsx
// buttonVariables
(siteVariables, props) => ({
  ...
})
```

### Proposed solution
Although a step related to make styling aspects to be completely a theme's area of responsibility is definitely a step in the right direction, there is need to discuss potential alternatives to proposed solution.

Please, consider this PR where discussion thread has started #207 - all related posts are cited here as well.

kuzhelov:
@levithomason's response:

I'm not sure we should move forward with this pattern just yet.  It is a big change that has many implications.  I also spent some time discussing it with Fabric yesterday and I think we reached a few good conclusions:

It seems there are only two options here.  
1. **Many variables** - Code combinatorial variables for every component (primaryBackground, primaryDisabledBackground, etc) 
1. **Many Branches** - Code combinatorial function branches that return the proper variable values

Using many variables does have advantages over using functions:
1. Hover, active, focus states can be handled with CSS selectors
1. It is JSON serializable
1. We can show edit controls in the doc experience for variables
1. Designers are able to edit variable values and JSON
1. It is more simple

Using variable functions has a few downsides:
1. Handling hover, active, focus requires JS listeners and state updates
1. The logic is not JSON serializable
1. We cannot give a path to customizing the values, unless we give a full featured code editor
1. Only engineers skilled in JS will be capable of customizing it
1. It is more complex

### Example implementation

Here's how we could use the existing `props` passed to our style function to override the icon style when used in the Radio:

```jsx
const radioStyles = {
  root: ({ props, variables }) => ({
    '.ui-icon': {
      color: props.checked
        ? variables.checkedIconBackgroundColor
        : variables.uncheckedIconBackgroundColor
    }
  })
}
```

[...]

Issue #66, opened 2018-08-08
CSS in JS issues


levithomason:
We'll track CSS in JS issues we're having here and propose solutions later.

1. Cannot switch to a different renderer (normal, rtl) during the life of the app #58 
2. We'll need to support multiple React roots on a single page.

robinweser:
@levithomason FWIW, we're thinking about introducing dynamic direction support, adding the ability to switch rtl => ltr dynamically as well as having different modes for different sub trees :) 

PS: I was thrilled to see Fela in the code here. Looks like a very promising UI library. If there's anything I can help with, feel free to hit me up =)

levithomason:
>we're thinking about introducing dynamic direction support

That would be awesome.  This is one of the big hurdles for us now.

The other goal we have is to reduce the number of wrapping components around every UI component.  This is the reason we are using the renderer directly opposed to an HOC.

miroslavstastny:
3. Fela does not prefix keyframe class names - RTL keyframes rewrite LTR ones #516

miroslavstastny:
4. Fela incompatible with `--isolatedModules` #613

layershifter:
After #1459 we will not need to switch renderers.
Fela is also compatible with TS `--isolatedModules` now 🚀

levithomason:
Both the original issues are now solved 💪

PR #240, opened 2018-09-17
CSS-based approach to reflect pseudo-state changes of parent component in child's styles

kuzhelov:
This prototype is provided to test the CSS-based approach for the problem of tweaking styles of child component on parent's pseudo state changes (like `:hover`) - see #231.

**DISCLAIMER: PROTOTYPE ALERT! This PR is just a prototype, so please, dig into the implementation by accepting the fact  that code could be dirty.**

## API suggested

Is provided in the `buttonStyles.js`. Just to frame an idea, this is a proposed way of providing `Icon` styles on `Button` hover:

```js
// buttonStyles
const styles = {
  ...
  onHover: {
    icon: ({ props }) => ({ props.primary ? 'red' : 'blue' }),
    ...  // styles for other slots could be specified here
    }
 },
 on<another_pseudo_state>: {
   ...
 }
}
```

And for the `Button`'s logic the following changes should be applied - please, note that those are generic:

```jsx
const icon = Icon.create(icon, {
      ... 
      className: classes.slots.icon, // all pseudo-state style slots are aggregated into 'slots' property of 'classes' (we should, probably, better name it)
   }
}
```

## How it works

Essentially, provided logic results in the following CSS being added to the page:

```css
.${componentClassName}:${pseudoStateName} .${slotClassName} {
  ...
}

// specifically, simplified version looks like this
.ui-button:hover .slot-icon-76ae8 {
   ...
}
```

and the following HTML being rendered:

```html
<button class='ui-button .. '>
  <span class='... slot-icon-76ae8 ...' >...</span>
</button>
```

Please, note that `76ae8` is added to the class name of slot - this value is a hash that is calculated based on the props applied to the slot. This trick allow us to solve the following problems:
- address the problem of CSS styles scoping - even for the case of nested theme providers
- establish functional correspondence between parent component's state and calculates styles for child, so that each change of the parent's state (JS) or pseudo-state (CSS) will be properly reflected by styles of child
- limit the number of CSS classes necessary

### DIFF
--- a/src/lib/getClasses.ts
+++ b/src/lib/getClasses.ts
@@ -14,11 +14,33 @@ const getClasses = (
   renderer: IRenderer,
   componentStyles: IComponentPartStylesInput,
   styleParam: ComponentStyleFunctionParam,
+  extra: any,
 ): IComponentPartClasses => {
   // root, icon, etc.
   const componentParts: string[] = Object.keys(componentStyles)
 
+  const { componentId, className: componentClassName } = extra
+
   return componentParts.reduce((classes, partName) => {
+    // handle state section differently
+    // DISCLAIMER: dirty code alert!
+    if (partName.startsWith('on')) {
+      const pseudoStateName = partName.substr(2).toLowerCase()
+
+      const slotStyles = (componentStyles[partName] as any)()
+
+      Object.keys(slotStyles).forEach(slotName => {
+        const slotClassName = `slot-${slotName}`
+
+        renderer.renderStatic(
+          callable(slotStyles[slotName])(styleParam),
+          `.${componentClassName}.${componentId}:${pseudoStateName} .${componentId}.${slotClassName}`,
+        )
+      })
+
+      return classes
+    }
+
     classes[partName] = renderer.renderRule(callable(componentStyles[partName]), styleParam)
 
     return classes
--- a/src/lib/getClasses.ts
+++ b/src/lib/getClasses.ts
@@ -6,6 +6,16 @@ import {
   IRenderer,
 } from '../../types/theme'
 
+import * as crypto from 'crypto'
+
+const getHashCode = (fromString: string): string => {
+  return crypto
+    .createHash('md5')
+    .update(fromString)
+    .digest('hex')
+    .substr(0, 5)
+}
+
 /**
  * Returns a string of HTML classes.
  * Renders one or many component styles (objects of component parts) to the DOM.
@@ -19,32 +29,40 @@ const getClasses = (
   // root, icon, etc.
   const componentParts: string[] = Object.keys(componentStyles)
 
-  const { componentId, className: componentClassName } = extra
+  const { className: componentClassName } = extra
 
-  return componentParts.reduce((classes, partName) => {
-    // handle state section differently
-    // DISCLAIMER: dirty code alert!
-    if (partName.startsWith('on')) {
-      const pseudoStateName = partName.substr(2).toLowerCase()
+  return componentParts.reduce(
+    (classes, partName) => {
+      // handle state section differently
+      // DISCLAIMER: dirty code alert!
+      if (partName.startsWith('on')) {
+        const pseudoStateName = partName.substr(2).toLowerCase()
 
-      const slotStyles = (componentStyles[partName] as any)()
+        const slotStyles = (componentStyles[partName] as any)()
 
-      Object.keys(slotStyles).forEach(slotName => {
-        const slotClassName = `slot-${slotName}`
+        Object.keys(slotStyles).forEach(slotName => {
+          const slotStyle = callable(slotStyles[slotName])(styleParam)
+          const slotStyleHash = getHashCode(JSON.stringify(slotStyle))
 
-        renderer.renderStatic(
-          callable(slotStyles[slotName])(styleParam),
-          `.${componentClassName}.${componentId}:${pseudoStateName} .${componentId}.${slotClassName}`,
-        )
-      })
+          const slotClassName = `${slotName}-slot-of-${componentClassName}--${slotStyleHash}`
+
+          renderer.renderStatic(
+            slotStyle,
+            `.${componentClassName}:${pseudoStateName} .${slotClassName}`,
+          )
+
+          classes.slots[slotName] = slotClassName
+        })
+
+        return classes
+      }
+
+      classes[partName] = renderer.renderRule(callable(componentStyles[partName]), styleParam)
 
       return classes
-    }
-
-    classes[partName] = renderer.renderRule(callable(componentStyles[partName]), styleParam)
-
-    return classes
-  }, {})
+    },
+    { slots: {} } as any,
+  )
 }
 
 export default getClasses
--- a/src/themes/teams/components/Button/buttonStyles.ts
+++ b/src/themes/teams/components/Button/buttonStyles.ts
@@ -140,6 +140,13 @@ const buttonStyles: IComponentPartStylesInput = {
     overflow: 'hidden',
     ...(typeof props.content === 'string' && truncateStyle),
   }),
+
+  onHover: {
+    icon: ({ props }) => ({
+      color: 'red',
+      ...(props.primary && { backgroundColor: 'yellow' }),
+    }),
+  } as any,
 }
 
 export default buttonStyles
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
   },
   "dependencies": {
     "classnames": "^2.2.5",
+    "crypto": "^1.0.1",
     "fela": "^6.1.7",
     "fela-plugin-fallback-value": "^5.0.17",
     "fela-plugin-placeholder-prefixer": "^5.0.18",
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -2,6 +2,7 @@ import * as PropTypes from 'prop-types'
 import * as React from 'react'
 import whatInput from 'what-input'
 import * as _ from 'lodash'
+import * as cx from 'classnames'
 
 import { UIComponent, childrenExist, customPropTypes, createShorthandFactory } from '../../lib'
 import Icon from '../Icon'
@@ -44,6 +45,12 @@ export interface IButtonProps {
  *  - if button includes icon only, textual representation needs to be provided by using 'title', 'aria-label', or 'aria-labelledby' attributes
  */
 class Button extends UIComponent<Extendable<IButtonProps>, any> {
+  constructor(props, context) {
+    super(props, context)
+
+    this.componentId = 'button-instance-id'
+  }
+
   static create: Function
 
   public static displayName = 'Button'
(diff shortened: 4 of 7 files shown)


src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -140,6 +140,13 @@ const buttonStyles: IComponentPartStylesInput = {
     overflow: 'hidden',
     ...(typeof props.content === 'string' && truncateStyle),
   }),
+
+  onHover: {
+    icon: ({ props }) => ({
+      color: 'red',
```
kuzhelov, line 146: introduced just as a usage example

src/lib/getClasses.ts
```diff
@@ -14,15 +24,45 @@ const getClasses = (
...
-  return componentParts.reduce((classes, partName) => {
-    classes[partName] = renderer.renderRule(callable(componentStyles[partName]), styleParam)
+  const { className: componentClassName } = extra
+
+  return componentParts.reduce(
+    (classes, partName) => {
+      // handle state section differently
+      // DISCLAIMER: dirty code alert!
+      if (partName.startsWith('on')) {
+        const pseudoStateName = partName.substr(2).toLowerCase()
```
mnajdova, line 39: Totally supporting the work done here. In my opinion we can start to clear the code and move this further. As for the names of these parts, I would suggest adding the prefix (pseudo). I know this is dirty code as an example, but I think after renaming this it would be pretty much ready for review. What are your thoughts on this @kuzhelov ?

kuzhelov: absolutely agree :) still, one thing that I would like to test before making it 'PR ready' is performance - specifically, would like to estimate how costly it is for us to calculate these hashes, will try on the most contrived case (with pseudos being heavily used) and measure the time spent on that (and, essentially, how it compares to overall rerendering time)

src/themes/teams/components/Button/buttonStyles.ts
miroslavstastny: I understand this is just an example, but, anyway, you are not handling `disabled` and `primary` is not a prop but a type:
```jsx
      ...(!props.disabled && {
        color: 'red',
        ...(props.type === 'primary' && { backgroundColor: 'yellow' }),
      }),
```

src/lib/getClasses.ts
```diff
@@ -14,15 +24,45 @@ const getClasses = (
...
+
+        const slotStyles = (componentStyles[partName] as any)()
+
+        Object.keys(slotStyles).forEach(slotName => {
+          const slotStyle = callable(slotStyles[slotName])(styleParam)
+          const slotStyleHash = getHashCode(JSON.stringify(slotStyle))
+
+          const slotClassName = `${slotName}-slot-of-${componentClassName}--${slotStyleHash}`
+
+          renderer.renderStatic(
```
miroslavstastny, line 49: This breaks when you start mixing LTR and RTL renderers - the class name for both directions will be the same (as the hash is calculated before RTL transformation and `renderStatic`-rendered classes do not get renamed with the `selectorPrefix`) but can end up having different styles (as the styles go through rtl plugin).

To see in in docsite, try adding `borderLeft` to the `button.icon:hover` style and switch to RTL.

src/lib/getClasses.ts
```diff
@@ -6,6 +6,16 @@ import {
   IRenderer,
 } from '../../types/theme'
 
+import * as crypto from 'crypto'
+
+const getHashCode = (fromString: string): string => {
+  return crypto
+    .createHash('md5')
```
miroslavstastny, line 13: This will be fun to debug once you get a collision.

src/lib/getClasses.ts
```diff
@@ -14,15 +24,45 @@ const getClasses = (
...
+      if (partName.startsWith('on')) {
+        const pseudoStateName = partName.substr(2).toLowerCase()
+
+        const slotStyles = (componentStyles[partName] as any)()
+
+        Object.keys(slotStyles).forEach(slotName => {
+          const slotStyle = callable(slotStyles[slotName])(styleParam)
+          const slotStyleHash = getHashCode(JSON.stringify(slotStyle))
+
+          const slotClassName = `${slotName}-slot-of-${componentClassName}--${slotStyleHash}`
```
miroslavstastny, line 47: Why are you adding all the parts to the class name? (Ignoring collisions for now) the hash (and a short prefix) would be enough and could reduce number of classes in a page.

src/lib/getClasses.ts
```diff
@@ -14,15 +24,45 @@ const getClasses = (
...
+    (classes, partName) => {
+      // handle state section differently
+      // DISCLAIMER: dirty code alert!
+      if (partName.startsWith('on')) {
+        const pseudoStateName = partName.substr(2).toLowerCase()
+
+        const slotStyles = (componentStyles[partName] as any)()
+
+        Object.keys(slotStyles).forEach(slotName => {
+          const slotStyle = callable(slotStyles[slotName])(styleParam)
```
miroslavstastny, line 44: Now you are adding the style to DOM and class to the element even if `slotStyle === {}`. Is that intended?

src/lib/getClasses.ts
```diff
@@ -14,15 +24,45 @@ const getClasses = (
...
+          const slotStyleHash = getHashCode(JSON.stringify(slotStyle))
+
+          const slotClassName = `${slotName}-slot-of-${componentClassName}--${slotStyleHash}`
+
+          renderer.renderStatic(
+            slotStyle,
+            `.${componentClassName}:${pseudoStateName} .${slotClassName}`,
```
miroslavstastny, line 51: Don't you want to use direct child combinator here?
With this code, if you put `Button` inside another `Button`, hovering over the outer button highlights icon inside the inner button.

miroslavstastny (commented):
I like the approach, although I am don't like that from my point of view encapsulation is broken - you are directly touching styles of one component in another component (even in the same theme).

In the button-icon example I would rather move that logic as a separate exported function to iconStyles and call that function from buttonStyles

miroslavstastny:
Also, author of Fela JS is aware of this project (#66), it could be interesting to hear his opinion on this as I see this as a general css in js issue.

src/themes/teams/components/Button/buttonStyles.ts
kuzhelov: sorry, do not completely understand that - lets discuss..

kuzhelov:
@miroslavstastny 

> I like the approach, although I am don't like that from my point of view encapsulation is broken - you are directly touching styles of one component in another component (even in the same theme).

No objections in regards to 'encapsulation is broken' part - actually, this is closely attributes to the discussion we've had here: #231 . 

With that being said, however, seems that a reasonable strategy would be is the following - as a first thing, try to address this pseudo-selector issue in CSS, and this is what this PR is attempting to solve. The next and, to my mind, mandatory step would be to address this issue of broken encapsulation you've mentioned - have some ideas around how it could be extended to our 'variables' concept.

src/lib/getClasses.ts
kuzhelov: this is the part that attributes to another comment you've left: https://github.com/stardust-ui/react/pull/240#discussion_r218686036 - was experimenting with selectors a bit there. 

I was thinking about this problem as well and was trying to experiment with different approaches. The first obvious one was using direct child selectors in CSS, like the following one (if I've understood correctly, this is the approach you are suggesting in the comment):

```css
.ui-button > ui.icon { .. }
```

Unfortunately, see some problems with that. For example, we could easily imagine the situation where this slot component will be not a direct child of main component, but rather a nested child like in the following pseudo-case:

```jsx
// Button.tsx
// className attributes are explicitly added just to provide an idea

render = ({ ... }) => (
  <ElementType className='ui-button'>
      <Layout renderMain={() => <Icon className='ui-icon' />} />
  </ElementType>
)
```

which will be rendered to 

```html
<button class='ui-button'>
   <div class='layout-main'>
      <i class='ui-icon'></i>
   </div>
</button>
```

Apparently, direct child CSS selector won't work for this case.

src/lib/getClasses.ts
kuzhelov: if you are about the part where just first five characters are taken from calculated hash - have done this only for this prototype - but there are plenty of other strategies we might consider to make this 'id' part readable. 

But if you are talking about hash value in general - yes, agree, it will be extremely 'funny', but good point is that probability of that is low in a reasonable proportion to this 'funny' part :)

src/lib/getClasses.ts
kuzhelov: no, this is not intentional - we should optimize here :)

src/lib/getClasses.ts
kuzhelov: yes - but see this moment as being relatively easy to address, as it is juts about properly handling class name prefixes. Just to simplify prototype's code have omitted this from the scope for now, but surely, this moment should be addressed if approach will be accepted

src/lib/getClasses.ts
kuzhelov: this closely attributes to your comment about the `Icon` reacting on hover state changes of `Button` that is not its direct parent (`Button` wrapped in `Button` story). Was experimenting with different approaches to address this problem, and to this point preferring the one about introducing notion of `componentId` for parent component, so that slot selector will be like

```css
.${component-class-name}-${componentId} .${slotName}-slot-of-${componentId}--${slotStyleHash}`
```

With that the problem of `Button` -> `Button` -> `Icon` case should be solved - however, obviously, there are several questions that need to be answered before this approach could be proposed as **a** solution:
- what are perf implications (not worrying much about that, frankly)
- !! we should prevent bloat of different `componentId`s - otherwise we will end up with explosion of static styles on the page

Another way to solve the problem is to look on CSS side - specifically, to be smarter in terms of selector to use, so that this problem won't happen. Experimenting with that as well

levithomason:
This will work for limited cases of simple pseudo state selectors from within a theme.  However, what is the solution for the consumer who may need a selector like this?

```jsx
<SomeComponent
  styles={{
    '&:hover .my-component:not(.active)': {
      ...
    }
  }}
/>
```

This is a current and anticipated future need.  I'm afraid any abstraction over pseudo selectors really will need to handle all selector types.  In essence, I fear we'd need to rebuild the syntax of the CSS selector as JS conventions in order to support all the necessary features.

levithomason:
@rofrischmann what's your take on the necessary cases where a parent needs to style a child?  Are you still of the mind that the future of Fela will support the [selectors shown in the docs](http://fela.js.org/docs/basics/Rules.html)?

robinweser:
We will continue support for those yeah. You might also want to check out [fela-identifier](https://github.com/rofrischmann/fela/tree/master/packages/fela-identifier) that's tackling that exact issue of "addressing child elements with a constant and unique parameter.

robinweser:
Not sure where to put this, so I'm abusing the comments section (sorry for that): 
I just noticed that two of you (@kuzhelov and @Bugaa92) are going to present Stardust at ReactiveConf!
Would be cool to meet and talk a bit if you want to (guess we'll see each other in the speakers area anyways), would love to hear more on Stardust :)

PR #41, opened 2018-08-02
feat(Chat.Message): add 'author' and 'timestamp' props

kuzhelov:
# Chat.Message ('author' and 'timestamp') props

Basic layout, as well as key properties were introduced.

image

### TODO

- [x] Conformance test
- [x] Minimal doc site example
- [x] Stardust base theme
- [ ] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [ ] Confirm RTL usage
- [ ] [W3 accessibility](https://www.w3.org/standards/webdesign/accessibility) check
- [ ] [Stardust accessibility](https://github.com/stardust-ui/accessibility) check
- [ ] Update glossary props table
- [ ] Update the CHANGELOG.md

# API Proposal

## Content

image

```jsx
<Chat.Message author="..." content="Sure thing! .." timestamp="..." />
```

```html
<span class='ui-text'>Sure thing! ...</span>
```

## Author

image

```jsx
<Chat.Message author="Jane Doe" content="..." timestamp="..." />
```

```html
<div class='ui-layout__main'>
  <h5 class='ui-header'>Jane Doe</h5>
  ..
</div>
```

## Timestamp

image

```jsx
<Chat.Message author="..." content="..." timestamp="Yesterday, 10:15 PM" />
```

```html
<div class='ui-layout__end'>
  <span class='ui-text'>Jane Doe</span>
  ..
</div>
```

PR #242, opened 2018-09-17
feat(chat message): add `author` and `timestamp` props

bmdalex:
# Chat.Message: `author` and `timestamp`

This PR introduces `author` and `timestamp` props for `Chat.Message` subcomponent

### TODO

- [x] Conformance test
- [ ] Stardust base theme
- [x] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [x] Confirm RTL usage
- [x] Update the CHANGELOG.md

## `author`

Rendered only in "their" messages (`mine=false`).

screen shot 2018-09-17 at 20 23 51

```jsx
<Chat.Message author="John Doe" content="Hello there"  />
```
renders
```html
<li class="ui-layout ui-chat__message">
  <div class="ui-layout__main">
    <div class="ui-layout">
      <div class="ui-layout__start">
        <span class="ui-text a">John Doe</span>
      </div>
      <div class="ui-layout__main">Hello there</div>
    </div>
  </div>
</li>
```

## `timestamp`
screen shot 2018-09-17 at 20 23 42

```jsx
<Chat.Message timestamp="Yesterday, 10:15 PM" content="Hello there"  />
```
renders
```html
<li class="ui-layout ui-chat__message">
  <div class="ui-layout__main">
    <div class="ui-layout">
      <div class="ui-layout__start">
        <span class="ui-text a">Yesterday, 10:15 PM</span>
      </div>
      <div class="ui-layout__main">Hello there</div>
    </div>
  </div>
</li>
```


CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Features
 - Add `FocusZone` to `renderComponent`, change `Menu` behavior to support arrow keys @tomasiser ([#116](https://github.com/stardust-ui/react/pull/116))
 - Add `value`, `disabled`, `checked`, `defaultChecked` and `onChange` props to `Radio` component @mnajdova ([#206](https://github.com/stardust-ui/react/pull/206))
+- Add `author` and `timestamp` props for `Chat.Message` component @Bugaa92 ([#242](https://github.com/stardust-ui/react/pull/242))
```
kuzhelov: consider to move this entry to 'Unreleased' section

src/components/Chat/ChatMessage.tsx
```diff
@@ -46,50 +64,91 @@ class ChatMessage extends UIComponent<Extendable<IChatMessageProps>, any> {
...
+      />
+    )
+  }
+
+  private renderText = (content: string, timestamp?: boolean) => (
+    <Text
+      timestamp={timestamp}
+      content={content}
+      size="sm"
+      styles={{ root: { marginRight: pxToRem(10) } }}
```
kuzhelov: we shouldn't hardcode styles in component's implementation code. Lets rather introduce `timestamp` part in the ChatMessage's styles, and provide this margin there

src/components/Chat/ChatMessage.tsx
```diff
@@ -46,50 +64,91 @@ class ChatMessage extends UIComponent<Extendable<IChatMessageProps>, any> {
...
+          </>
+        }
+        main={content}
+      />
+    )
+  }
+
+  private renderText = (content: string, timestamp?: boolean) => (
+    <Text
+      timestamp={timestamp}
```
kuzhelov: lets use two separate components for that - one for displaying `timestamp`, another one - to display `content`. This is necessary because `Text` component is seen to be a basic one, so it would support either `timestamp` variant, or regular `content` one. More than that, it is highly likely that `timestamp` variant won't be introduced as part of core Stardust

src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -3,6 +3,7 @@ import { IChatMessageProps } from '../../../../components/Chat/ChatMessage'
 import { IChatMessageVariables } from './chatMessageVariables'
 import { pxToRem } from '../../../../lib'
 
+const rem10 = pxToRem(10)
```
kuzhelov: would rather suggest name like `px10asRem` - otherwise the name is a bit misleading and reads like '10 rem'

src/themes/teams/components/Chat/chatVariables.ts
```diff
@@ -0,0 +1,7 @@
+export interface IChatVariables {
+  backgroundColor: string
```
kuzhelov, line 2: this is better to be handled by the code of examples for now - more than that, we could adjust padding aspects there as well

src/components/Chat/ChatMessage.tsx
```diff
@@ -2,19 +2,34 @@ import * as React from 'react'
...
   as?: any
+  author?: string
   avatar?: ItemShorthand
   children?: ReactChildren
   className?: string
   content?: any
   mine?: boolean
   styles?: IComponentPartStylesInput
+  timestamp?: string
   variables?: ComponentVariablesInput
```
mnajdova: Why don't we make the text and the author shorthands for the Text component?

mnajdova:
Can we close https://github.com/stardust-ui/react/pull/41 in favor of this one?

kuzhelov:
sure, lets do that

src/components/Chat/ChatMessage.tsx
kuzhelov: yeah - have discussed with @mnajdova, here proposal is essentially to introduce `author` and `timestamp` shorthands (slots) for the `ChatMessage` component, and use `Text` as a default component for these.

Totally support this idea 👍

src/themes/teams/components/Chat/chatVariables.ts
mnajdova: Actually, I added the backgroundColor and the padding to the variables with some defaults that would match the Team's theme, as we are developing this theme, so why not have this look by default. Please share your thoughts on this. Also, I added avatar slot in the ChatMessage, for configuring the statusBorderColor to match the background. Actually I have one thing that I am not 100% sure that should be implement this way. The Chat component has it's own backgroundColor and this color should match the Avatar's statusBorderColor. The problem is that, the Avatar is inside the ChatMessage component, and I am not confident that the Chat component should define the avatar's slot in the ChatMessage component. For now, those colors are independent variables in the Chat, and the ChatMessage's avatar slot component. What do you think about this @kuzhelov? This is how the default look of the examples looks now:
image

CHANGELOG.md
```diff
@@ -17,6 +17,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### BREAKING CHANGES
+- Fixed `Divider` wrong usage of the `typeSecondary{color, backgroundColor}` and `default{color, backgroundColor}` variables; renamed `default{color, backgroundColor}` variables to `color` and `backgroundColor` @mnajdova ([#234](https://github.com/stardust-ui/react/pull/234))
```
kuzhelov: 👍

src/components/Chat/ChatMessage.tsx
```diff
@@ -2,19 +2,33 @@ import * as React from 'react'
...
+  ComponentPartStyle,
+} from '../../../types/theme'
 import { Extendable, ReactChildren, ItemShorthand } from '../../../types/utils'
 import Avatar from '../Avatar'
+import Layout from '../Layout'
+import Text from '../Text'
 
 export interface IChatMessageProps {
   as?: any
+  author?: ItemShorthand
```
kuzhelov, line 24: 👍

src/themes/teams/components/Chat/chatVariables.ts
kuzhelov: sounds quite reasonable, thanks!

kuzhelov (approved):

src/themes/teams/components/Chat/chatVariables.ts
kuzhelov: agreed to not introduce these two as variables for now, just inline them into styles directly

mnajdova: We are going to stick with the backgroundColor as variable in order to reuse the siteVariables. Paddings are deleted from variables

Issue #230, opened 2018-09-14
Theme-defined icons support (including SVG, font-based)


kuzhelov:
### Problem description
Currently it is impossible for theme author to define set of icons that could be rendered by `Icon` component. Specifically, if theme author would like to introduce icon `foo` and consume it in the following way with her theme being applied, currently there is no way to achieve that

```jsx
<Icon name='foo'>
```

### Proposed solution
Here is the plan we see to address this issue

- ~~introduce mechanics of consuming icons from theme~~
  - ~~now is introduced for SVGs (#183)~~, should be generalised for font-based icons as well (#243)
- ~~resolve aspects related to sizing (it should be possible to pixel-perfectly customize size of the icon)~~
- ~~support icon's outline/filled variants~~
- ~~support changing look of the icon on hover state~~

kuzhelov:
closing as all the mentioned steps are made :)

PR #243, opened 2018-09-17
feat(theme): add font icons support for themes

kuzhelov:
This PR is introduced as one of the steps necessary to address #230.

## Overview

Support for font icons is introduced for themes, icon types are generalized (no need to specify `svg` prop to switch between SVG and font-based icons).

#### Before
```jsx
// SVG icon
<Icon svg name='umbrella' />   

// font icon
<Icon name='user' /> 
```

#### Now
```jsx
// SVG icon
<Icon name='umbrella' />

// font icon - exactly the same API
<Icon name='user' />
```

## Font icons support for themes
Theme authors are able to introduce custom set of icons, both font-based and SVGs are supported.

```jsx
// theme.js
const icons = {
  'umbrella': () => <svg>...</svg>,
  'user': { fontFamily: 'Icons', content: '"\\f0152" }
}
...

export default {
  ...
  icons: icons
}
```


hughreeling (commented):
Looks good.

src/lib/renderComponent.tsx
```diff
@@ -120,7 +122,8 @@ const renderComponent = <P extends {}>(
       }: IThemeInput | IThemePrepared = {}) => {
         const ElementType = getElementType({ defaultProps }, props)
 
-        const stateAndProps = { ...state, ...props }
+        const derivedProps = getExtraProps({ icons })
```
miroslavstastny: So `derivedProps` or `extraProps`?

src/components/Icon/Icon.tsx
```diff
@@ -130,10 +136,26 @@ class Icon extends UIComponent<Extendable<IIconProps>, any> {
...
+      isFontBased,
+      fontIconFromTheme: isFontBased && icons[this.props.name],
+    }
+  }
+
   renderComponent({ ElementType, classes, rest, accessibility, icons }) {
-    return this.props.svg
-      ? this.renderSvgIcon(ElementType, icons, classes, rest, accessibility)
-      : this.renderFontIcon(ElementType, classes, rest, accessibility)
+    return this.isFontBased(icons)
```
miroslavstastny: I don't like the fact that you do similar calculation to what you've already done in `getExtraProps()`. Primarily not because of the performance but because for me (as a component developer) it is difficult to understand what the `extraProps` are good for - they are used to calculate styles, accessibility... why there are not used to render the component?
`renderComponent()` should receive `extraProps` as a parameter.

src/components/Icon/Icon.tsx
```diff
@@ -130,10 +136,26 @@ class Icon extends UIComponent<Extendable<IIconProps>, any> {
     )
   }
 
+  isFontBased(themeIcons: ThemeIcons): boolean {
+    const { name } = this.props
```
miroslavstastny: nit: from the function signature, it is not obvious that it checks current icon (in this).
I would change the signature to `isFontBased(name, themeIcons)`

src/components/Icon/Icon.tsx
levithomason: Agreed, we can remove `getExtraProps` entirely if we check the type of the icon instead.  To limit the number of ideas and patterns in the library, I propose we also allow values or functions (as with variables and styles):

```jsx
const icons = {
  // value form
  umbrella: <svg></svg>,                             
  user: { fontFamily: 'Icons', content: "\\f0152" },

  // function form
  umbrella: () => <svg></svg>,
  user: () => ({ fontFamily: 'Icons', content: "\\f0152" }),
}

React.isValidElement(callable(icons.umbrella)()) ? isSVG : isFONT
```

This would allow us to pass values to the icon later.  It also removes the need for the `getExtraProps` function.

src/components/Icon/Icon.tsx
```diff
@@ -130,10 +136,26 @@ class Icon extends UIComponent<Extendable<IIconProps>, any> {
...
+  getExtraProps({ icons }): IconExtraProps {
+    const isFontBased = this.isFontBased(icons)
+
+    return {
+      isFontBased,
+      fontIconFromTheme: isFontBased && icons[this.props.name],
+    }
+  }
+
   renderComponent({ ElementType, classes, rest, accessibility, icons }) {
```
levithomason: Propose we pass the entire `theme` to `renderComponent` as it is more generic and seems more useful in a wider range of scenarios.  Icons alone probably are not that useful to many other components.

miroslavstastny (approved):

kuzhelov:
agreed with @miroslavstastny to reduce amount of style parts exposed for SVG icons to absolute necessary minimum (instead of `path1, path2, path3 ... path9` we have now) before merging changes into `master`

src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -103,7 +98,9 @@ const iconStyles = {
       width: '1em',
       height: '1em',
 
-      ...(isFontBased ? getFontStyles(font, name, size) : {}),
+      ...(isFontBased
```
kuzhelov: simplify this one

types/theme.d.ts
```diff
@@ -249,10 +250,24 @@ export interface IFontFace {
...
 // ========================================================
-// SVG Icons
+// Icons
 // ========================================================
 
 type Classes = { [iconPart: string]: string }
-export type RenderSvgIconFunction = (classes: Classes) => React.ReactNode
+type SvgIconFuncArg = {
+  classes: Classes
+  props: any
```
kuzhelov: `props` should be eliminated for now

PR #245, opened 2018-09-18
feat(test): add accessibility behavior tests

kolaps33:
# Unit tests generating from description of "behavior files"

Adding unit test which are generating from description of each behavior file.

There are regex definition which are validated against the behavior description.
When regex match, then parameters are taken and test is generated and executed.

Currently is cover most of the cases for behavior files (will create additional ticket to cover other scenarious) :
image


mnajdova:
I am really worried about generating tests depending on comments provided for the implementation. Can we use some kind a configuration for making the tests be easily generated instead of using comments. I am worried that if we have some kind of a typo inside the comments, we may miss testing some features in the implementation.

kuzhelov:
and, what is worse, now we could have build being broken because of some typo in comment. More than that, while now (when PR is just introduced) it is apparent which words should be used in comments, this intuitiveness will evanesce quite quickly with time - and while there could be suggestion to follow the path taken for other components (in terms of mimicking their @description section), in cases when new kind of word construct will be needed we might find the situation when it will collide with the generator's assumptions and break the build.

Agree with @mnajdova thoughts on that.

jurokapsiar:
There are two reasons for this:
1) quality of the code
2) quality of the descriptions (ensuring that what is written there is true and the wording is consistent)

This is a first step, in a following PR we will also fail the tests if none of the patterns matched a particular line. That will require having two sections - specification and description. Description will not be tested.

If there is a typo, it is good that we know about it. Also, a typo won't appear out of nowhere, the user that changes the text will see his changes. Once we will have more test definitions, we will categorize them to make sure it is easier to find them.

mnajdova:
@jurokapsiar I totally support both of the reasons mention, but we can easily achieve both with taking different approach. We can create description object in the behaviors, that will in some structure reflect the comments, and use this object in the tests, as well as in the docs description of the behavior.  We are either way creating `behaviorXXX.info.json` files, so these can be easily used for generating the documentation, while still ensuring that the same is implemented using the tests. The comments then can reflects the same, even with providing some additional thoughts if needed. I will be happy to help with the implementation of it, if needed.

kuzhelov:
have discussed it with @jurokapsiar, @kolaps33, @mnajdova and @alinais. 

Speaking of the pros of this approach (and the reason why it is rather something that doesn't violate common practices) is that for JS Docs there is special syntax used, for the content that is also provided in docs (@jurokapsiar). Also, there was a proposal to introduce dedicated section for expressing these rules, so it would be harder for the developer to be trapped in a situation where changes in the comments will result in broken tests.

Another research vector suggested is to take a look on the TS attributes - in a hope that they will be able to provide a way to specify rules in a formal language, as well as will be able to serve as a source for docs description, - with as little overhead possible.

jurokapsiar:
TS decorators (attributes) cannot be applied on plain functions - since behaviors are functions and we do not want to add complexity by making them classes, we cannot use decorators.

test/specs/behaviors/testHelper.tsx
```diff
@@ -0,0 +1,108 @@
...
+    if (!importedBehavior) {
+      throw 'Behavior file was not find, probably was not imported. Import file and add behavior.'
+    }
+
+    return typeof importedBehavior === 'function'
+      ? (importedBehavior as AccessibilityDefinitionFunction)
+      : (importedBehavior as IAccessibilityDefinition)
+  }
+
+  public convertToBooleenIfApplicable(stringToConvert: string) {
```
miroslavstastny: ...Boolean...

test/specs/behaviors/testHelper.tsx
```diff
@@ -0,0 +1,108 @@
...
+    })
+  }
+
+  public getBehaviorWithProperType(
+    behaviorName: string,
+  ): AccessibilityDefinitionFunction | IAccessibilityDefinition {
+    const baseBehaviorName = behaviorName.replace('.ts', '')
+    const importedBehavior = this.behaviors.get(baseBehaviorName)
+    if (!importedBehavior) {
+      throw 'Behavior file was not find, probably was not imported. Import file and add behavior.'
```
miroslavstastny: ...found...

miroslavstastny (commented):
To make sure that I understand the expected final state correctly:
1. all "test sentences" will be moved to `@specification` section.
2. each test sentence will either match a regex (and then it will be tested) or fail the test.
Considering we cannot use TS decorators and we need the specifications to be part of the behavior jsdoc, I am ok with that.

miroslavstastny:
Looks ok for the first version, just two minor comments:

1. after the renaming storm, `Behavior-test.tsx` should be `behavior-test.tsx`,
2. I would add a comment to `Behavior-test.tsx` that before running the test you must run `gulp build:docs:component-menu-behaviors`

kolaps33:
Issues mentioned in previous comment, should be fixed now :)

miroslavstastny (approved):

PR #248, opened 2018-09-18
feat(test): add test React app to verify Stardust is consumable

kuzhelov:
This PR aims to address the problem of published Stardust package being unconsumable on the client side. For the sake of preventing this situation a check for the following scenario has been introduced:
- create simple TSX React app (by means of `create-react-app` util)
- add Stardust as this app's dependency
- use some Stardust components in the app's main file (`App.tsx`)
- try to build the app

To verify this scenario the `test:projects:cra-ts` yarn script has been introduced, as well as added to the list of CI steps performed on each commit.


package.json
```diff
@@ -135,6 +135,7 @@
     "leven": "^2.1.0",
     "lint-staged": "^7.0.2",
     "merge2": "^1.2.2",
+    "mkdirp": "^0.5.1",
```
kuzhelov: we could omit adding this dependency

build/gulp/tasks/test-projects.tsx
```diff
@@ -0,0 +1,135 @@
...
+  ////////
+
+  try {
+    log('STEP 1. Create test React project with TSX scripts..')
+    const testAppDir = path.resolve(tmp.dirSync().name, 'test')
+    const runInTestApp = cmd => sh(`cd ${testAppDir} && ${cmd}`)
+
+    fs.mkdirSync(testAppDir)
+
+    await runInTestApp(`yarn create react-app . --scripts-version=react-scripts-ts`)
```
levithomason: Heads up, this appears to install a global on the users machine:
https://yarnpkg.com/lang/en/docs/cli/create/

build/gulp/tasks/test-projects.tsx
```diff
@@ -0,0 +1,135 @@
...
+    log("STEP 4. Reference Stardust components in test project's App.tsx")
+    fs.writeFileSync(path.resolve(testAppDir, 'src', 'App.tsx'), appTSX)
+
+    ////////////
+
+    log('STEP 5. Build test project..')
+    await runInTestApp(`yarn build`)
+    log('Test project is built successfully!')
+  } finally {
+    await sh(`rm ${stardustPackageFilename}`)
```
levithomason: Perhaps ignore this tarball instead.  That way on fail, it is not an issue that this is not run.

levithomason (approved):
Looks great, thanks for the improvements and updates.  Address comments as you see fit and merge at will.

PR #249, opened 2018-09-18
fix(Divider): Matching divider Teams theme to redlines

codepretty:
Matched divider styles for the Teams theme to the redlines. This includes updating colors, variable names, spacing.

In Teams theme, there are only 2 types of dividers

**1. Default divider w/text**
image

**2. Primary, important divider w/text**
image


kuzhelov (approved):

src/themes/teams/components/Divider/dividerVariables.ts
```diff
@@ -1,10 +1,11 @@
 export default (siteVars: any) => {
   return {
-    color: siteVars.gray02,
-    backgroundColor: siteVars.gray10,
-    typePrimaryColor: siteVars.brand,
-    typePrimaryBackgroundColor: siteVars.brand,
-    typeSecondaryColor: siteVars.gray04,
-    typeSecondaryBackgroundColor: siteVars.gray08,
+    dividerColor: siteVars.gray09,
+    textColor: siteVars.gray03,
```
mnajdova, line 4: Please mention in the CHANGELOG that we are renaming this variables, so that the user can be aware if they were using them.

src/themes/teams/components/Divider/dividerStyles.ts
```diff
@@ -33,43 +33,36 @@ const dividerStyles: IComponentPartStylesInput = {
...
     const { children, size, type, important, content } = props
     return {
-      marginTop: pxToRem(5 + size * 7.5),
-      marginBottom: pxToRem(5 + size * 7.5),
-      fontWeight: 400,
+      color: variables.textColor,
+      display: 'flex',
+      alignItems: 'center',
+      ...(type === 'primary' && {
+        color: variables.primaryColor,
```
notandrew: does this set color twice then?

mnajdova: No, if the type is primary, the second will override the first definition.

mnajdova (approved):

mnajdova:
Little bit worried about not applying any height on the Divider, as currently we used it also as an invisible divider between elements. Please take a look into this visual regression: https://screener.io/v2/states/rkGqtccHX.stardust-ui-react/fix%2Fdivider-redlines/1024x768/Chrome/portalexampleshorthandtsx-0 
Please reason for this and update the styles accordingly, if you think they should be changed.

notandrew:
Maybe we add an optional buffer size between elements, but I'm not sure that we want a height set. That may make it harder to modify in the case where we want smaller spacers between objects.

codepretty:
I don't think we want to be setting a height either, but for dividers in the Teams theme it's probably okay to have top/bottom spacing. At least for now until we see other use cases for it and then we can make spacing optional if that is needed.

So now we can at least have 2 dividers in succession that don't sit on top of each other...
image

PR #252, opened 2018-09-19
docs: more theming examples

alinais:
This PR adds more examples for the Theming guide.


docs/src/views/ThemingExamples.tsx
```diff
@@ -0,0 +1,400 @@
...
+import { NavLink } from 'react-router-dom'
+import { Header } from 'semantic-ui-react'
+import { Button, Divider, Icon, Label, Provider } from '@stardust-ui/react'
+
+import DocPage from '../components/DocPage/DocPage'
+import ExampleSnippet from '../components/ExampleSnippet/ExampleSnippet'
+
+export default () => (
+  <DocPage title="Theming Examples">
+    <Header as="h2" content="Theming needs" />
```
hughreeling: needs doesn't seem right here.  Theming approaches?  Theming mechanisms?  Theming techniques?

docs/src/views/ThemingExamples.tsx
```diff
@@ -0,0 +1,400 @@
...
+
+    <p>
+      A Theme defines default values and functions to generate variables that will eventually be
+      used to generate styles for individual components.
+    </p>
+    <p>Stardust is intentional in supporting three levels of theming needs:</p>
```
hughreeling: There seem to be four...

docs/src/views/ThemingExamples.tsx
```diff
@@ -0,0 +1,400 @@
...
+            typeSecondaryBackgroundColor: 'charcoal',
+            paddingLeftRightValue: 20,
+          }}
+        />
+      )}
+    />
+
+    <Header as="h3" content="Theme Level Styling" />
+    <p>
+      Talking about pixel perfect output, one might need to obtain more detailed changes applied to
```
hughreeling: To achieve pixel perfect output

docs/src/views/ThemingExamples.tsx
```diff
@@ -0,0 +1,400 @@
...
+      the entire suite of styles. The components styling can be changed at different levels of the
+      theme:
+    </p>
+
+    <ol>
+      <li>
+        <code>siteVariables</code> - when more general changes are requested (colors, page container
+        sizes, default fonts that can be inherited or not at the detailed component variables
```
hughreeling: I think something like:  Typically these control the page colors, sizes, default fonts etc, and allow for easier restyling of the whole app.  Individual components can choose to inherit these values.

docs/src/views/ThemingExamples.tsx
```diff
@@ -0,0 +1,400 @@
...
+        level),
+      </li>
+      <li>
+        <code>componentVariables</code> - when all the components of certain kind have to have same
```
hughreeling: Styles can be applied to all instances of particular component types, and these styles can also be parameterized using componentVariables.

docs/src/views/ThemingExamples.tsx
```diff
@@ -0,0 +1,400 @@
...
+              <Divider content="Secondary Divider" type="secondary" />
+            </div>
+          </Provider>
+        </div>
+      )}
+    />
+
+    <Header as="h3" content="Nesting Providers" />
+    <p>
+      If you need to customize a bigger part of the theme, you can nest the second provider and
```
hughreeling: If you have areas of an application that require different theming, you can achieve that using nested providers.

docs/src/views/ThemingExamples.tsx
```diff
@@ -0,0 +1,400 @@
...
+              <br />
+              <Divider content="Primary Divider" type="primary" />
+              <Divider content="Secondary Divider" type="secondary" />
+            </div>
+          </Provider>
+        </div>
+      )}
+    />
+
+    <Header as="h3" content="Nesting Providers" />
```
kuzhelov: would suggest `Nesting Themes` as more domain-related name (as Providers is rather an implementation detail)

docs/src/views/ThemingExamples.tsx
```diff
@@ -0,0 +1,400 @@
...
+        <strong>Default</strong> - Users who just need a good first run experience.
+      </li>
+      <li>
+        <strong>Component level styling</strong> - Users who need to change a little or a lot.
+      </li>
+      <li>
+        <strong>Theme level styling</strong> - Users who have pixel perfect design requirements.
+      </li>
+      <li>
+        <strong>Nesting providers</strong> - Users who need some styling at the upper theme level
```
kuzhelov: probably, `Nesting themes` would better reflect reader's needs..

docs/src/views/ThemingExamples.tsx
```diff
@@ -0,0 +1,400 @@
...
+        but also at component level.
+      </li>
+    </ol>
+
+    <Header as="h3" content="Default" />
+    <p>If you do not need any custom theming, just import some components and start building.</p>
```
kuzhelov: maybe `start building` -> `render them`?

docs/src/views/ThemingExamples.tsx
```diff
@@ -0,0 +1,400 @@
...
+          <br />
+          <Divider content="Primary Divider" type="primary" />
+          <Divider content="Secondary Divider" type="secondary" />
+        </>
+      )}
+    />
+
+    <Header as="h3" content="Component Level Styling" />
+    <p>
+      When you need only small amount of styling isolated per component, you can opt-in to change
```
kuzhelov: I would suggest something like the following - seems that the key thing to emphasize here is that this is *per-instance* customization mechanism:

> When you need to tweak styles of specific component's instance, you can opt-in to change its styles and theme-defined variables.

After that we could mention that this technique could be used to create component wrappers (the thing provided examples are about).

docs/src/views/ThemingExamples.tsx
```diff
@@ -0,0 +1,400 @@
...
+      certain local styles or the variables behind it.
+    </p>
+
+    <Header as="h4" content="Changing local styles" />
```
kuzhelov: would suggest `Changing component styles`, similar to the header of the next section

docs/src/views/ThemingExamples.tsx
```diff
@@ -0,0 +1,400 @@
...
+
+    <Header as="h3" content="Theme Level Styling" />
+    <p>
+      Talking about pixel perfect output, one might need to obtain more detailed changes applied to
+      the entire suite of styles. The components styling can be changed at different levels of the
+      theme:
+    </p>
+
+    <ol>
+      <li>
```
kuzhelov, line 170: seems to be a good place to paste picture @hughreeling was talking about. Probably, for now we could use `ul` for enumerating these options

docs/src/views/ThemingExamples.tsx
```diff
@@ -0,0 +1,400 @@
...
+        <code>siteVariables</code> - when more general changes are requested (colors, page container
```
kuzhelov: would suggest to add precise definition of the scope each variables group provides effect for:
- site variables - those define theme-wide styling parameters that could affect styles of all the components. One might want consider to tweak them when ...
- component variables - define variables that will affect styles os specific component type only. ...
- component styles - define CSS properties that will be applied to components of specific type. Those have the same scope as `component variables`, but serve as a lower level styling abstraction (and, thus, are more specific). Consider to use them ...

docs/src/views/ThemingExamples.tsx
kuzhelov: probably, it would be good to mentioning common use cases where one will need to preserver most of theming aspects, but at the same time achieve one or more of the following
- tweak some specific styling aspects of existing theme - so to speak, build theme on top of existing one
- introduce theme changes that will be applied to specific components' subtree only

This will immediately provide reader with an idea whether this section is something she is looking for

docs/src/views/ThemingExamples.tsx
```diff
@@ -0,0 +1,415 @@
...
+          <Divider content="Primary Divider" type="primary" />
+          <Divider content="Secondary Divider" type="secondary" />
+
+          <Provider
+            theme={{
+              siteVariables: { brand: 'darkred' },
+            }}
+          >
+            <>
+              <Header as="h3" content="First provider theming" />
```
kuzhelov: we might consider to name it just as `First nested theming` - i.e., to omit `provider` word, as it is rather Stardust term not closely related to the scenario user is interested in (nested theming)

kuzhelov (approved):

PR #254, opened 2018-09-19
feat(accessibility): Add FocusZone for ButtonGroup, TabList and Toolbar behaviors

jurokapsiar:
This PR adds FocusZone handling for Button Group, TabList and Toolbar behaviors.


src/lib/accessibility/Behaviors/Toolbar/ToolbarBehavior.ts
```diff
@@ -1,11 +1,17 @@
...
   attributes: {
     root: {
       role: 'toolbar',
     },
   },
+  focusZone: {
+    mode: FocusZoneMode.Wrap,
+    props: {
+      isCircularNavigation: false,
+    },
```
sophieH29, line 19: What about setting direction? As far as I remember it will be bidirectional by default (which has a bigger performance impact). Might be worth to set to the horizontal direction.
Same also applies for tablist, I think

jurokapsiar: this is just initial implementation, we will need to get feedback on the arrow keys. For user that can see, that a toolbar is vertical/horizontal it makes sense to use only up/down or left/right. but screen reader user does not see the orientation of the toolbar/tabbar so it might be confusing for him.

src/lib/accessibility/Behaviors/Tab/TabBehavior.ts
```diff
@@ -12,9 +24,17 @@ const TabBehavior: Accessibility = (props: any) => ({
       'aria-label': props['aria-label'],
```
hughreeling, line 25: aria-describedby provides similar capabilities - should we handle that too?

hughreeling (approved):

miroslavstastny (approved):

PR #255, opened 2018-09-20
feat(Chat): add different chat item types

mnajdova:
# ChatMessage

This PR aims to add different types of messages for the chat. We should support the following types:
- control messages 
image
- bubble messages (this is currently the implementation of the ChatMessage itself)
image
- divider messages
image

### TODO

- [x] Conformance test
- [ ] Stardust base theme
- [x] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [ ] Confirm RTL usage
- [x] Update the CHANGELOG.md

I have three proposals and would like to hear your opinion on.

# API Proposal 1

## type: 'bubble' | 'control' | 'divider'

The type property will be used for rendering one of the three different components. The properties will remain the same, we would just add additional icon shorthand for the control component.

## Usage:
```jsx
<ChatMessage type='bubble' author={...} ... />
<ChatMessage type='control' icon={...} ... />
<ChatMessage type='divider' content={'Yesterday'} ... />
```
This would be easy for use, but not much open for fully customizing the generated component.

# API Proposal 2

## divider: shorthand for the Message.Divider component

Open for customization and consistent with the other usages of the shorthand. One problem is that, a message cannot be at the same time two of these types, so we will kind a have to decide behind the scene which would take precedence. 

## Usage
```jsx
<ChatMessage bubble={{author: 'John Doe', ...}} ... />
<ChatMessage control={{icon: {{...}}, content='Some content...'}} ... />
<ChatMessage divider={{content: '.Yesterday}}... />
```
# API Proposal 3

This proposal is combination of the previous two. We should use the type for deciding which component will be rendered, but will use the shorthand for generating the component itself.

## divider: shorthand for the Message.Divider component

```jsx
<ChatMessage type='bubble' bubble={{author: 'John Doe', ...}} ... />
<ChatMessage type='control'  control={{icon: {{...}}, content='Some content...'}} ... />
<ChatMessage type='divider' divider={{content: '.Yesterday}}... />
```


kuzhelov:
From those options that are provided would rather vote against the third one, due to that it opens way to (accidentally) do things improperly on the consumer's side:

```jsx
<ChatMessage type='divider' bubble={{author: 'John Doe', ...}} ... />
```

Speaking of the rest two (Proposal 1 and Proposal 2) - the safest approach would be to start with just different type values being supported (no shorthands for now), and see how it will go afterwards; although I should notice that I do really like a shorthand-based approach (i.e. Proposal 2) as well 👍

mnajdova:
Got it, I agree it's the easiest from user perspective as well. Will start with the Proposal 1, we can easily migrate to 2 if we see it would be needed.

mnajdova:
Actually, let's go directly with the Proposal 2, because even now I can see some problems with allowing the user to define different types for the Divider for example. I would like to avoid having to change the whole API rather soon...

alinais:
I'd go also with the second proposal.
One thing though: why do we need the divider type of the message? Can't divider be simply a divider and used where needed?

mnajdova:
Not really in this moment, as the messages are collection shorthand inside the chat component, so we aren't really capable of adding some other component in between..

mnajdova:
To be more specific the chat definition would look something like this:
```jsx
<Chat items={[
    {{bubble: {avatar: {...}, content: '...'}}},
    {{bubble: {avatar: {...}, content: '...'}}},
    {{control: {icon: {...}, content: 'Meeting started...'}}}, //now that I think about this, maybe the better naming would be action
    {{divider: {content: 'Today'}}},
    {{bubble: {avatar: {...}, content: '...'}}},
  ]}
```

So the general change is that, the chat messages will be transformed to chat items, which can be messages, dividers, action etc..

mnajdova:
Still work in progress, but this is how it currently looks:
image


And here is the code for that example: 
```jsx
import React from 'react'

import { Chat } from '@stardust-ui/react'

const janeAvatar = {
  src: 'public/images/avatar/small/ade.jpg',
  status: {
    color: 'green',
    icon: 'check',
    title: 'Available',
  },
}

const items = [
  {
    key: 1,
    bubble: { content: 'Hello', author: 'John Doe', timestamp: 'Yesterday, 10:15 PM', mine: true },
  },
  {
    key: 2,
    bubble: {
      content: 'Hi',
      author: 'Jane Doe',
      timestamp: 'Yesterday, 10:15 PM',
      avatar: janeAvatar,
    },
  },
  {
    key: 3,
    bubble: {
      content: 'Would you like to grab a lunch?',
      author: 'John Doe',
      timestamp: 'Yesterday, 10:16 PM',
      mine: true,
    },
  },
  {
    key: 4,
    bubble: {
      content: "Sure! Let's try the new place downtown",
      author: 'Jane Doe',
      timestamp: 'Yesterday, 10:16 PM',
      avatar: janeAvatar,
    },
  },
  { key: 5, divider: { content: 'Today', type: 'primary', important: 'true' } },
  {
    key: 6,
    bubble: {
      content: "Let's have a call",
      author: 'John Doe',
      timestamp: 'Today, 11:15 PM',
      mine: true,
    },
  },
  {
    key: 7,
    action: { icon: 'record', content: 'Meeting started', timestamp: 'Today, 11:15 PM' },
  },
]

const ChatExampleShorthand = () => <Chat items={items} />

export default ChatExampleShorthand

```

@kuzhelov @alinais please give me your feedback about the API.

docs/src/examples/components/Chat/index.tsx
```diff
@@ -1,11 +1,9 @@
 import React from 'react'
 import Types from './Types'
-import Variations from './Variations'
 
 const MessageExamples = () => (
```
bmdalex: I know it's out of scope of PR, but we might want to rename to `ChatExamples` to avoid confusion

src/components/Chat/ChatItem.tsx
```diff
@@ -0,0 +1,130 @@
...
+  static className = 'ui-chat__item'
+
+  static create: Function
+
+  static displayName = 'ChatItem'
+
+  static propTypes = {
+    as: customPropTypes.as,
+
+    /** Shorthand for the bubble message. */
```
bmdalex: Since `bubble`, `divider` and `action` props are mutually exclusive, we need a better way to prevent users try more than one at a time; I'd suggest something like this:
```typescript
/** Shorthand for the bubble message. */
bubble: customPropTypes.disallow(['action', 'divider']),

/** Shorthand for the divider message. */
divider: customPropTypes.disallow(['action', 'bubble']),

/** Shorthand for the control message. */
action: customPropTypes.disallow(['bubble', 'divider']),
```

src/components/Chat/ChatItem.tsx
```diff
@@ -0,0 +1,130 @@
...
+    )
+  }
+
+  private renderMessage = (
+    styles: IComponentPartStylesPrepared,
+    variables: ComponentVariablesInput,
+  ): React.ReactNode => {
+    const { bubble, action, divider } = this.props
+
+    return (
```
bmdalex: right now we assume priorities `bubble` > `action` > `divider` and some users would be confused if they try to mix them (use `bubble` and `action` at the same time and only `bubble` renders); we need to make it obvious for user these are mutually exclusive, see my suggestion above

src/components/Chat/ChatBubble.tsx
```diff
@@ -93,7 +93,7 @@ class ChatMessage extends UIComponent<Extendable<IChatMessageProps>, any> {
     rest,
     styles,
     variables,
-  }: IRenderResultConfig<IChatMessageProps>) {
+  }: IRenderResultConfig<IChatBubbleProps>) {
```
bmdalex: hmm.. if you take a look at the proposal in our internal ticket, the idea was that the `ChatBubble` would contain the body of a message + media and avatar would be part of `ChatItem`; is that achievable?

image

UPDATE:
Taking a closer look, we add `avatar` prop for `ChatBubble` and `icon` prop for `ChatAction`; so if we're certain that's always the case (bubbles can have avatar and actions can have icon) I think this is a good separation and you can disregard the comment

docs/src/examples/components/Chat/index.tsx
mnajdova: Sure, I will restructure to example containing all different chat items. Will consider this as well.

src/components/Chat/ChatItem.tsx
mnajdova: Just the thing I was missing. Thanks a lot for the suggestion Alex!

src/components/Chat/ChatItem.tsx
mnajdova: Taken into consideration.

src/components/Chat/ChatBubble.tsx
mnajdova: Yes, we needed actually different components for the structure of the different types of chat items, that's why I separate those.

src/components/Chat/ChatAction.tsx
```diff
@@ -0,0 +1,133 @@
...
+  renderComponent({
+    ElementType,
+    classes,
+    rest,
+    styles,
+    variables,
+  }: IRenderResultConfig<IChatActionProps>) {
+    const { as, icon, children } = this.props
+
+    return childrenExist(children) ? (
```
bmdalex: @mnajdova - it looks like the `Chat.Action` subcomponent is more complex (per @codepretty 's comments in our chats, we can have an array of actions with more complicated logic on what's rendered); I'm not even sure it falls under Stardust principles, maybe we should be more generic with the `action` prop in `Chat` component

src/components/Chat/ChatBubble.tsx
bmdalex: got it, makes sense, thanks!

src/components/Chat/ChatAction.tsx
mnajdova: In my opinion, we shouldn't alter this component at this moment. We can reason for this in the following manner:
- Chat.Action can have children which can define whatever the user wants inside - in this case we won't change anything right now
- We can add actions as a list and make the Chat.Action use the Accordion component, but in that case it would be in my opinion very much Team's specific rather then common web (as @Bugaa92 said, it doesn't really fall under stardust principles)

My final thought on this is, leaving the Chat.Action component as is for this now, and address this later if we think would be necessary. First, I would like after this is merged, to try and address this in the chat pane and we will see how much complication it would bring.

docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx
```diff
@@ -2,25 +2,63 @@ import React from 'react'
...
+    bubble: {
+      content: "Let's have a call",
+      author: 'John Doe',
+      timestamp: 'Today, 11:15 PM',
+      mine: true,
+    },
+  },
+  {
+    key: 7,
+    action: { icon: 'record', content: 'Meeting started', timestamp: 'Today, 11:15 PM' },
```
bmdalex: nit:

image

mnajdova: Changed, thanks for the suggestion!

mnajdova:
We have some doubts whether the `Divider` should be define as an attribute to the `ChatItem`, as it is not really a message in the context of the `Chat` component. We can provide props like `lastRead` and `firstMessageOnDay` together with some shorthand for defining the `Divider` before or after the message. Although this provide more 'chat like' API, my worries are that those are kind a restrictive to just these two scenarios. We could define before and after dividers too, but that would be too much generic for the `ChatItem` (not much different than the `List`). I would like to hear your opinion on this, and what makes more sense from user's perspective.

sophieH29:
> We have some doubts whether the `Divider` should be define as an attribute to the `ChatItem`, as it is not really a message in the context of the `Chat` component. We can provide props like `lastRead` and `firstMessageOnDay` together with some shorthand for defining the `Divider` before or after the message. Although this provide more 'chat like' API, my worries are that those are kind a restrictive to just these two scenarios. We could define before and after dividers too, but that would be too much generic for the `ChatItem` (not much different than the `List`). I would like to hear your opinion on this, and what makes more sense from user's perspective.

I like an idea to make a divider as an attribute to a message. But maybe, to avoid restrictions just to those 2 mentioned scenarios (often chats have 'yesterday' and 'someday' dividers), we can provide just props ```before``` and ```after```, and a user will define the content ('last read', 'yesterday', 'today', etc).

src/components/Chat/ChatBubble.tsx
```diff
@@ -6,20 +6,20 @@ import {
...
   IComponentPartStylesInput,
 } from '../../../types/theme'
-import { Extendable, ReactChildren, ItemShorthand } from '../../../types/utils'
+import { Extendable, ItemShorthand, ReactChildren } from '../../../types/utils'
 import Avatar from '../Avatar'
 import Layout from '../Layout'
 import Text from '../Text'
 
-export interface IChatMessageProps {
+export interface IChatBubbleProps {
```
codepretty: Not sure how much of this belongs in Stardust, but seems like we might be missing a couple props for message bubbles.

1. How should we handle a succession of messages? It seems to be common UI practice to show those messages different from message with dividers or other messages between them...
**From FB messenger:**
image

**From Teams:**
image

2. What about _Edited_ or _translated_ messages?
**From Telegram:**
image

**From Teams:**
image

3. Message actions and important/mentioned indicators.
**From FB messenger:**
image

**From Teams:**
image

levithomason:
Recapping our offline chat about Stardust's guidelines for creating components and APIs:

1. Syntax is evil, avoid creating it at all costs.
1. Prefer patterns, composition, and reuse over syntax and APIs.
1. Stardust will not contain any original components, component slots, or layouts.
1. Stardust only creates components and APIs when we can find a concept in widespread usage in many well-known UIs.
1. Never invent a component or API based on a single use case or our feelings.

The conclusion to the Chat API for now was to provide mostly or only the children API for now.  As we have more time and use cases to standardize on we will.  Right now, the correct abstraction is not clear so we'll let users compose their own abstractions using our lower level components for Chat.

src/components/Chat/ChatBubble.tsx
levithomason: 1. Yes, nice call.  We currently have a `mine` prop for styling messages that are yours vs others'.  Perhaps there is a `grouped` or `consecutive` prop that we have for delineating messages that should be styled like a group or consecutively sent?

1. The edited identifier is common, however, the placement is not always common.  I'm not sure the best way to implement this due to that fact, since, right now we don't allow you to change the component "template" (where slots are rendered).  We only allow changing the contents of the slots themselves.  Here's an example of the `edited` being rendered elsewhere:

    image
Because of this, I think we should leave it to the consumer to place this where needed.

1. As with the `edited` identifier, the consumer will place their own actions where necessary for now.  There is again more variance than consistency here that I'm not sure we can reasonably create an abstraction for.  Most likely, the consumer will want very specific implementations of these parts of a message (animations, analytics, etc).

mnajdova:
So, after the discussions we had, the components are updated to provide the user to use the Chat with different components that should be used as children of the Chat.Item component that represents just a row inside the Chat. For example, here is what the code would look like for rendering the Chat provided in the picture.

image

```jsx
import React from 'react'

import { Chat, Divider } from '@stardust-ui/react'

const janeAvatar = {
  image: 'public/images/avatar/small/ade.jpg',
  status: {
    color: 'green',
    icon: 'check',
    title: 'Available',
  },
}

const ChatExample = () => (
  <Chat>
    <Chat.Item>
    </Chat.Item>
    <Chat.Item>
      <Chat.Message
        content="Hi"
        author="Jane Doe"
        avatar={janeAvatar}
      />
    </Chat.Item>
    <Chat.Item>
      <Chat.Message
        content="Would you like to grab a lunch?"
        author="John Doe"
        timestamp="Yesterday, 10:16 PM"
        mine
      />
    </Chat.Item>
    <Chat.Item>
      <Chat.Message
        content="Sure! Let&apos;s try the new place downtown"
        author="Jane Doe"
        avatar={janeAvatar}
      />
    </Chat.Item>
    <Chat.Item>
      <Divider content="Today" type="primary" important />
    </Chat.Item>
    <Chat.Item>
      <Chat.Message
        content="Let&apos;s have a call"
        author="John Doe"
        timestamp="Today, 11:15 PM"
        mine
      />
    </Chat.Item>
  </Chat>
)

export default ChatExample

```

For generating the Action messages, the user can use the Accordion component, or the Chat.Message component with providing transparent background and using the avatar as an icon. Moreover, the Divider can be used together with the Chat.Message, or as a separate Chat.Item. More example will be provided in the Chat Pane Prototype.

## Question
One issue we have now, after removing the usage of the slots inside the styles property is that I had to add the `mainStyle` and `avatarStyle` properties inside the Chat.Message, because the layout is not a shorthand, and we cannot pass these styles any other way. Am I missing something, do we have another way to pass these?

src/components/Chat/ChatItem.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  static displayName = 'ChatItem'
+
+  static propTypes = {
+    as: customPropTypes.as,
+
+    /** Child content. */
+    children: PropTypes.node,
+
+    /** Additional classes. */
```
alinais: Please change the description to align to the rest of the components propTypes:
"Additional CSS class name(s) to apply."

src/components/Chat/ChatMessage.tsx
```diff
@@ -6,15 +6,16 @@ import {
...
   createShorthandFactory,
   customPropTypes,
-  UIComponent,
   IRenderResultConfig,
+  UIComponent,
 } from '../../lib'
 import {
-  ComponentVariablesInput,
   ComponentPartStyle,
+  ComponentVariablesInput,
```
alinais, line 14: I like the ordered dependencies :)

alinais (approved):

levithomason:
Note, we'll remove the temporary styles properties in this PR.   Once the Layout is fixed, https://github.com/stardust-ui/react/pull/236, we can use the shorthand styles properly for the Layout areas.

mnajdova:
This is ready for final review @levithomason @kuzhelov @miroslavstastny

src/components/Chat/ChatItem.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+  static defaultProps = {
+    as: 'li',
+  }
+
+  renderComponent({
+    ElementType,
+    classes,
+    rest,
+    styles,
```
kuzhelov: nit: seems that these two guys could be omitted, as those are not used anywhere

src/components/Chat/ChatItem.tsx
```diff
@@ -0,0 +1,70 @@
...
+    variables,
+  }: IRenderResultConfig<IChatItemProps>) {
+    const { children, content } = this.props
+
+    return (
+      <ElementType {...rest} className={cx(classes.root, classes.content)}>
```
kuzhelov: we have potentially unreliable styling logic here - could you, please, suggest what are our plans to address it? Would rather expect this merge being performed in styles file, and exposed as `item` styles then

src/components/Chat/ChatItem.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+class ChatItem extends UIComponent<Extendable<IChatItemProps>, any> {
+  static className = 'ui-chat__item'
+
+  static create: Function
+
+  static displayName = 'ChatItem'
+
+  static propTypes = {
+    as: customPropTypes.as,
```
kuzhelov, line 32: nit: description for this prop is missing (as well s for a few other components of this PR, like `ChatMessage`) - is it intentional?

docs/src/examples/components/Chat/Types/ChatExample.tsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react'
```
kuzhelov, line 1: shorthand API currently is not working correctly in docs examples - the problem is apparent if one would try to click on 'Shorthand API' button:

image

Do we really want to dismiss shorthand case for `Chat`? This seems to be confusing, as by looking on `Chat`'s implementation it is clearly visible that `items` prop is exposed - thus, it is possible to utilize shorthand advantages for this component. More than that, it seems that we won't be able to get rid of Shorthand API support, due to in this case there won't be clear mechanisms for handling `Chat` component's accessibility concerns.

src/components/Chat/ChatMessage.tsx
```diff
@@ -78,7 +78,7 @@ class ChatMessage extends UIComponent<Extendable<IChatMessageProps>, any> {
 
   static defaultProps = {
     accessibility: ChatMessageBehavior as Accessibility,
-    as: 'li',
+    as: 'div',
```
kuzhelov: have take a look on how `children` prop is handled for this component - seems that there is a problem with that, as `ElementType` will be always a wrapper for the content provided as `children`. Could we address this problem in this PR, or you think it would be better to file an issue for that?

src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -18,6 +18,9 @@ const chatMessageStyles: IComponentPartStylesInput<IChatMessageProps, IChatMessa
           marginRight: 'auto',
         }),
     maxWidth: v.messageWidth,
+    width: 'max-content',
```
kuzhelov: could you, please, suggest what would be the styling problem if this CSS property won't be applied? The reason I am asking is because this one is marked as experimental: https://developer.mozilla.org/en-US/docs/Web/CSS/width

src/components/Chat/Chat.tsx
```diff
@@ -35,8 +36,8 @@ class Chat extends UIComponent<Extendable<IChatProps>, any> {
 
     children: PropTypes.node,
 
-    /** Shorthand array of messages. */
-    messages: PropTypes.arrayOf(PropTypes.any),
+    /** Shorthand array of the items inside the chat. */
+    items: PropTypes.arrayOf(PropTypes.any),
```
kuzhelov: nit: seems that we could be more specific here: 

```js
items:  PropTypes.arrayOf(customPropTypes.itemShorthand)
```

kuzhelov (commented):
overall the vector of changes taken looks good to me 👍 - few moments that have raised my concerns, left comments for them.

src/components/Chat/ChatItem.tsx
mnajdova: This is a wrong copy/paste error, thanks for noticing. Actually we will have just the classes.root here.

src/components/Chat/ChatItem.tsx
mnajdova: It's not intentional, will provide description for all prop types in the components added/changed in this PR.

src/themes/teams/components/Chat/chatMessageStyles.ts
mnajdova: Without that prop this is how the chat looks:
image

I will try to implement this behavior using some other styling then...

docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx
```diff
@@ -0,0 +1,51 @@
...
+    content: (
+      <Chat.Message
+        content="Hi"
+        author="Jane Doe"
+        timestamp="Yesterday, 10:15 PM"
+        avatar={janeAvatar}
+      />
+    ),
+  },
+  <Chat.Item>
```
kuzhelov: frankly, a bot unsure whether we should provide sort of heterogenous approach in the example - probably, it would be more intuitive if all the items will be the same (especially given that those that are at the end have different content - won't like to mix this differentiation aspect with another one that is about using `Chat.Item` directly instead of content prop)

src/components/Chat/ChatItem.tsx
```diff
@@ -48,17 +48,11 @@ class ChatItem extends UIComponent<Extendable<IChatItemProps>, any> {
...
-    rest,
-    styles,
-    variables,
-  }: IRenderResultConfig<IChatItemProps>) {
+  renderComponent({ ElementType, classes, rest }: IRenderResultConfig<IChatItemProps>) {
     const { children, content } = this.props
 
     return (
-      <ElementType {...rest} className={cx(classes.root, classes.content)}>
+      <ElementType {...rest} className={classes.root}>
```
kuzhelov, line 55: 👍

kuzhelov (approved):

PR #256, opened 2018-09-20
feat(Accessibility): Added focus handling for List component

smykhailov:
Let me introduce focus handling for List component, which is based on the idea of and roving tabindex technique. Also, the idea is based on the current implementation of accessibility behaviors and key handlers. This implementation is based on the research and PoC located in the [accessibility repository](https://github.com/stardust-ui/accessibility/tree/master/playground)

### Container/AtomicItem hierarchy 
A lot of components can be presented as a hierarchical structure as container and item(s), each item may also contain a nested container with its own items. Having this in mind, we can encapsulate focus handling into two components and inject the corresponding functionality into the React component. When container created it can set the first (or any other) item to be focused. The item itself handles the pressed key and fires the corresponding event. Container receives this event and does what is requested as it knows about children, how to handle them and when they have to be updated and how.

### Roving tabindex technique

Setting the `tabindex` of the focused element to "0" ensures that if the user tabs away from the widget and then return, the selected item within the group retains focus. Note that updating the `tabindex` to "0" requires also updating the previously selected item to `tabindex="-1"`. This technique involves programmatically moving the focus in response to key events and updating the `tabindex` to reflect the currently focused item.

```JSX
<radio tabindex="-1">1</radio>
<radio tabindex="-1">2</radio>
<radio tabindex="0">3</radio>
<radio tabindex="-1">4</radio>
<radio tabindex="-1">5</radio>
```

[more details about roving tabindex](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Keyboard-navigable_JavaScript_widgets#Technique_1_Roving_tabindex)


src/lib/accessibility/FocusHandling/AtomicItemFocusHandler.tsx
```diff
@@ -0,0 +1,109 @@
...
+export type SetStateDelegate<P, S> = <K extends keyof S>(
+  state:
+    | ((prevState: Readonly<S>, props: Readonly<P>) => Pick<S, K> | S | null)
+    | (Pick<S, K> | S | null),
+  callback?: () => void,
+) => void
+
+export class AtomicItemFocusHandler<P extends IAtomicItemProps, S extends IAtomicItemState> {
+  constructor(
+    private getProps: () => P,
```
sophieH29: Is there any reason why not just passing props but wrap it in function?

kuzhelov: yes - otherwise just reference will be passed, so it won't be updated with update of component props

src/lib/accessibility/Behaviors/List/BasicListItemBehavior.ts
```diff
@@ -6,12 +7,31 @@ import { IAccessibilityDefinition } from '../../interfaces'
  * The 'listitem' role is used to identify an element that is a single item in a list.
  */
 
-const BasicListItemBehavior: IAccessibilityDefinition = {
```
jurokapsiar: please add keyboard handling o selectable list only - simple list is not navigable/focusable

smykhailov: fixed

jurokapsiar:
proposal: 
- rename IAtomicItem to IFocusable or IFocusItem
- rename Container to FocusContainer

src/components/List/List.tsx
```diff
@@ -86,6 +90,15 @@ class List extends UIComponent<Extendable<IListProps>, any> {
   // List props that are passed to each individual Item props
   static itemProps = ['debug', 'selection', 'truncateContent', 'truncateHeader', 'variables']
 
+  private containerFocusHandler = new ContainerFocusHandler(
```
jurokapsiar: can we replace this with something like
```
focusHandler.createContainer(this)
```

src/lib/accessibility/FocusHandling/AtomicItemFocusHandler.ts
```diff
@@ -0,0 +1,109 @@
...
+    this.setState({
+      shouldSubContainerBeOpened: true,
+      isLastOpened: true,
+    })
+
+    this.getProps().onEnter()
+  }
+
+  public space() {
+    if (!this.getProps().isFocused) {
```
hughreeling: Could we move the isFocused tests up a level somehow - it's repeated code for every handler...

src/lib/accessibility/FocusHandling/ContainerFocusHandler.ts
```diff
@@ -0,0 +1,71 @@
...
+  ) {
+    this.initState({ focusItemOnIdx: 0 })
+  }
+
+  public assignAtomicItemsProps(idx: number, itemsLength: number): IAtomicItemProps {
+    const itemProps: IAtomicItemProps = {
+      isFocused: idx === this.getState().focusItemOnIdx && this.getState().focusItemOnIdx !== -1,
+      isFirstElement: idx === 0,
+      isLastElement: idx === itemsLength - 1,
+      onMovePrevious: this.movePrevious.bind(this),
```
hughreeling: Is the lifecycle of the keymapping assignments and the isXXXXX variables the same?  Seems like we should be recalculating isXXXX each keypress?

src/lib/accessibility/Behaviors/List/SelectableListItemBehavior.ts
```diff
@@ -11,6 +12,23 @@ const SelectableListItemBehavior: (props: any) => IAccessibilityDefinition = (pr
     root: {
       role: 'option',
       'aria-selected': !!props['active'],
+      tabIndex: props.atomicItemProps.isFocused ? '0' : '-1',
+    },
+  },
+  keyActions: {
```
jurokapsiar: did you try moving the keyActions to the container? it looks like that might simplify the logic and reduce the number of callbacks sent to the item as this is updating the container state anyway

src/lib/accessibility/Behaviors/List/SelectableListItemBehavior.ts
```diff
@@ -1,3 +1,4 @@
+import * as keyboardKey from 'keyboard-key'
```
jurokapsiar: thinking about the list use case again - probably list will be used with some kind of virtualized list in the future, which would require support of children API. maybe radiogroup and radioitem would be better use cases for this.

src/lib/accessibility/FocusHandling/AtomicItemFocusHandler.ts
smykhailov: It doesn't make a lot of sense, as we need to return from the function, based on this condition. If we extract this logic to the method, it will look the following:
```TS
  private isFocused(): boolean {
    return this.getProps().isFocused
  }

  public space() {
    if (!this.isFocused()) {
      return
    }
```
which is almost the same in terms of maintainability, but adds one extra function call.

src/lib/accessibility/Behaviors/List/SelectableListItemBehavior.ts
smykhailov: yes, that is what we want to support, I haven't yet done the full investigation of how (and if) it will be possible to do. But that is definitely the thing I'll do the next.

src/lib/accessibility/FocusHandling/ContainerFocusHandler.ts
smykhailov: Not exactly on each, but only on that we should handle, and all of them are changing the state, and change state causes component rerendering, which will recalculate the isXXX on this phase, and that's exactly what is needed

src/components/List/List.tsx
smykhailov: This would be nice to have, and that's was the initial way how I did it, but there is a reason, why not to do it:
- FocusContainer/FocusableItem will have a dependency on React.Component, which I want to avoid, to keep the focusable logic framework agnostic

jurokapsiar: we can have the framework agnostic, but still provide factory functions specific for react. for other frameworks, we can create different factory functions.

src/lib/accessibility/FocusHandling/FocusContainer.ts
```diff
@@ -0,0 +1,75 @@
...
+      onMovePrevious: this.movePrevious.bind(this),
+      onMoveNext: this.moveNext.bind(this),
+      onMoveFirst: this.moveFirst.bind(this),
+      onMoveLast: this.moveLast.bind(this),
+      onEnter: this.enter.bind(this),
+      onSpace: this.space.bind(this),
+      onEsc: this.esc.bind(this),
+    }
+
+    return itemProps
```
sophieH29: nit, as you're not using ```itemProps``` anywhere, you can immediately return an object

```
return {
      isFocused: idx === this.getState().focusItemOnIdx && this.getState().focusItemOnIdx !== -1,
      isFirstElement: idx === 0,
      isLastElement: idx === itemsLength - 1,
      onMovePrevious: this.movePrevious.bind(this),
      onMoveNext: this.moveNext.bind(this),
      onMoveFirst: this.moveFirst.bind(this),
      onMoveLast: this.moveLast.bind(this),
      onEnter: this.enter.bind(this),
      onSpace: this.space.bind(this),
      onEsc: this.esc.bind(this),
    }
```

src/components/List/ListItem.tsx
```diff
@@ -94,7 +108,22 @@ class ListItem extends UIComponent<Extendable<IListItemProps>, any> {
...
   }
 
-  state: any = {}
+  private itemRef = React.createRef<HTMLElement>()
+
+  private focusableItem = new FocusableItem(
+    () => this.props.focusableItemProps,
+    this.setState.bind(this),
+    state => {
+      this.state = { ...{ isHovering: false }, ...state }
```
sophieH29: why do you need here this ```...{ isHovering: false }```?

It leads to one more ```Object.assign``` call, I think might be ok just ```this.state = { isHovering: false, ...state }```

Also, if ```...state``` will have ```isHovering:true``` it will override ```isHovering: false``` at the beginning.
Is it intended?

src/components/List/ListItem.tsx
```diff
@@ -104,6 +133,10 @@ class ListItem extends UIComponent<Extendable<IListItemProps>, any> {
     this.setState({ isHovering: false })
   }
 
+  componentDidUpdate() {
+    this.focusableItem.focus(ReactDOM.findDOMNode(this.itemRef.current!) as HTMLElement)
+  }
```
sophieH29, line 112: At first look, it's a bit confusable, as seems like it is going to be focused anyway. But inside ```this.focusableItem.focus``` method there is a check which actually decides focus or not.
Saying that, can we rename it to ```tryFocus```?

src/lib/accessibility/FocusHandling/FocusableItem.ts
```diff
@@ -0,0 +1,113 @@
...
+      shouldSubContainerBeOpened: true,
+      isLastOpened: true,
+    })
+
+    this.getProps().onEnter()
+  }
+
+  private isFocused(): boolean {
+    return this.getProps().isFocused
+  }
```
sophieH29, line 22: you have this method, but still using ```this.getProps().isFocused``` everywhere

src/lib/accessibility/FocusHandling/FocusableItem.ts
```diff
@@ -0,0 +1,113 @@
...
+    })
+  }
+
+  public focus(focusableElement: HTMLElement) {
+    if (this.getProps().isFocused) {
+      focusableElement.focus()
+    }
+  }
+
+  public movePrevious() {
```
miroslavstastny: Change all these methods to arrow functions `public movePrevious = () => {` so you get rid of all the `bind()` calls.

src/lib/accessibility/FocusHandling/FocusContainer.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  private movePrevious(): void {
+    this.setState(prev => {
+      return { focusItemOnIdx: prev.focusItemOnIdx - 1 }
```
miroslavstastny, line 53: How do you plan to modify this to support disabled items in a list?

src/lib/accessibility/FocusHandling/FocusContainer.ts
```diff
@@ -0,0 +1,75 @@
...
+    private setState: SetStateDelegate<P, S>,
+    private initState: (state: IFocusContainerState) => void,
+    private getState: () => S,
+  ) {
+    this.initState({ focusItemOnIdx: 0 })
+  }
+
+  public assignAtomicItemsProps(idx: number, itemsLength: number): IFocusableItemProps {
+    const itemProps: IFocusableItemProps = {
+      isFocused: idx === this.getState().focusItemOnIdx && this.getState().focusItemOnIdx !== -1,
```
miroslavstastny, line 41: This is the only place where you are currently using `getState()`. As this method is called directly from the container (`List`) I would rather remove `getState` from `ContainerFocusHandler` and pass the state as a parameter of `assignAtomicItemsProps()`.

src/lib/accessibility/FocusHandling/FocusableItem.ts
```diff
@@ -0,0 +1,113 @@
...
+  public moveLast() {
+    if (this.getProps().isLastElement || !this.getProps().isFocused) {
+      return
+    }
+
+    this.getProps().onMoveLast()
+  }
+
+  public enter() {
+    this.setState({ isLastOpened: false })
```
miroslavstastny: This is currently dead code so it is not possible to verify it is correct.
When does it happen that `enter()` is called on an item that is not focused?

src/lib/accessibility/FocusHandling/FocusableItem.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    if (!this.getProps().isFocused) {
+      return
+    }
+
+    this.setState({
+      shouldSubContainerBeOpened: true,
+      isLastOpened: true,
```
miroslavstastny: Also as these are not used anywhere at the moment, it is difficult for me to understand their purpose.

miroslavstastny (commented):
Tested it on accessibility prototype, it does not work as I would expect for nested containers:

tabbehavior

src/components/List/List.tsx
smykhailov: Agreed, added the static methods to create it

src/components/List/ListItem.tsx
smykhailov: good, catch. my bad. I've changed approach completely

src/components/List/ListItem.tsx
smykhailov: like it, renamed

src/lib/accessibility/FocusHandling/FocusContainer.ts
smykhailov: there are two ways of doing this: 
 - add a loop, which will verify the item state
 - wrap items into smart LinkedList, which will handle this automatically

but before doing this, we have to introduce component state (the state which reflects if the component is disabled/active/focused etc). This will be done as a separate PR and after we can proceed with it

src/lib/accessibility/FocusHandling/FocusContainer.ts
smykhailov: I've changed approach, to handle item focusing to be handled by a container and now there are a few of places, where it is called

src/lib/accessibility/FocusHandling/FocusableItem.ts
smykhailov: good point, Removed this part for now, until we have the real example to use it

src/lib/accessibility/FocusHandling/FocusableItem.ts
smykhailov: good point, Removed this part for now, until we have the real example to use it

src/lib/accessibility/FocusHandling/FocusContainer.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    this.setState(prev => {
+      return { focusItemOnIdx: prev.focusItemOnIdx - 1 }
+    })
+  }
+
+  public moveNext(): void {
+    if (
+      !this.getProps().items ||
+      this.getState().focusItemOnIdx >= this.getProps().items.length - 1
```
sophieH29, line 60: Are we going to support circular navigation?

test/specs/components/List/List-test.ts
```diff
@@ -6,7 +6,7 @@ import ListItem from 'src/components/List/ListItem'
 
 const listImplementsCollectionShorthandProp = implementsCollectionShorthandProp(List)
 
-describe('List', () => {
+xdescribe('List', () => {
```
sophieH29: It might be mistakenly appended this ```x```  to ```describe``` :)

src/lib/accessibility/FocusHandling/FocusContainer.ts
smykhailov: in later stages, not now

test/specs/components/List/List-test.ts
smykhailov: fixed

src/components/List/ListItem.tsx
```diff
@@ -75,7 +87,17 @@ class ListItem extends UIComponent<Extendable<IListItemProps>, any> {
     accessibility: listItemBehavior as Accessibility,
   }
 
-  state: any = {}
+  constructor(props: IListItemProps, state: IListItemState) {
```
kuzhelov: there is no `state` passed to component constructor, second arg should be removed

src/components/List/List.tsx
```diff
@@ -76,11 +80,25 @@ class List extends UIComponent<Extendable<IListProps>, any> {
   // List props that are passed to each individual Item props
   static itemProps = ['debug', 'selection', 'truncateContent', 'truncateHeader', 'variables']
 
+  private focusContainer = ContainerFocusHandler.create(this)
+
+  actionHandlers: AccessibilityActionHandlers = {
+    moveNext: this.focusContainer.moveNext.bind(this.focusContainer),
```
kuzhelov, line 102: nit: just as a matter of subsequent refactoring, we could provide a method that will fetch all four from focus container as an object, so that the code will look something like

```js
actionHandlers: AccessibilityActionHandlers = {
  ...this.focusHandler.getNavigationHandlers()
}
```

src/components/List/ListItem.tsx
smykhailov: fixed

src/components/List/List.tsx
smykhailov: let's do it as the next stage of refactoring, as this is more complex than it looks. MoveFrist, MoveLast might be not used in some of the scenarios, some other scenarios might require PageUp and PageDown in addition to what is already defined.

src/components/List/List.tsx
```diff
@@ -2,29 +2,33 @@ import * as _ from 'lodash'
...
-  items?: ItemShorthand[]
   selection?: boolean
   truncateContent?: boolean
   truncateHeader?: boolean
   styles?: ComponentPartStyle
   variables?: ComponentVariablesInput
 }
 
-class List extends UIComponent<Extendable<IListProps>, any> {
+class List extends UIComponent<Extendable<IListProps>, IFocusContainerState> {
```
kuzhelov, line 37: one observation is that, formally, we don't need to translate state back from the `FocusContainer` to component. Essentially, we could keep this state being encapsulated in `FocusContainer`, and the only thing that we would need is some notification mechanism to let component note that it should update. We could further elaborate on this idea to make `FocusContainer` being framework agnostic

src/components/List/List.tsx
```diff
@@ -90,7 +108,13 @@ class List extends UIComponent<Extendable<IListProps>, any> {
     const { items } = this.props
     const itemProps = _.pick(this.props, List.itemProps)
 
-    return _.map(items, item => ListItem.create(item, { defaultProps: itemProps }))
+    return _.map(items, (item, idx) => {
+      itemProps.focusableItemProps = this.focusContainer.assignAtomicItemsProps(idx, items.length)
```
kuzhelov: probably, `prepareItemProps` or `createItemProps` (with `Item` being a singular form) would be better, due to this method
- prepares props for single item element
- there is no assignment for the result props object made there
- probably, we don't need to repeat `Atomic` in method's names

src/components/List/ListItem.tsx
```diff
@@ -1,15 +1,22 @@
...
+  FocusableItem,
+  IFocusableItemProps,
+} from '../../lib/accessibility/FocusHandling/FocusableItem'
 import { ComponentVariablesInput, ComponentPartStyle } from '../../../types/theme'
 import { Extendable } from '../../../types/utils'
 
 export interface IListItemProps {
   accessibility?: Accessibility
   as?: any
+  focusableItemProps?: IFocusableItemProps
```
kuzhelov: nit: for the sake of consistency, lets preserve alphabetical order

src/components/List/ListItem.tsx
```diff
@@ -35,6 +46,7 @@ class ListItem extends UIComponent<Extendable<IListItemProps>, any> {
 
   static propTypes = {
     as: customPropTypes.as,
+    focusableItemProps: PropTypes.object,
```
kuzhelov: nit: maybe we should place it closer to `accessibility` prop

src/components/List/ListItem.tsx
```diff
@@ -75,7 +87,17 @@ class ListItem extends UIComponent<Extendable<IListItemProps>, any> {
     accessibility: listItemBehavior as Accessibility,
   }
 
-  state: any = {}
+  constructor(props: IListItemProps, context: any) {
```
kuzhelov: `context` should be omitted as well :) We are not supporting legacy React API. Also, in either case, there is no need to provide this arg here

src/components/List/ListItem.tsx
```diff
@@ -85,6 +107,10 @@ class ListItem extends UIComponent<Extendable<IListItemProps>, any> {
     this.setState({ isHovering: false })
   }
 
+  componentDidUpdate() {
+    this.focusableItem.tryFocus(ReactDOM.findDOMNode(this.itemRef.current!) as HTMLElement)
```
kuzhelov: this is better to be rewritten the following way, as we will reduce amount of casts used by one:

```js
ReactDOM.findDOMNode(this.itemRef.current) as HTMLElement
```

src/components/List/List.tsx
```diff
@@ -76,11 +80,25 @@ class List extends UIComponent<Extendable<IListProps>, any> {
   // List props that are passed to each individual Item props
   static itemProps = ['debug', 'selection', 'truncateContent', 'truncateHeader', 'variables']
 
+  private focusContainer = ContainerFocusHandler.create(this)
```
kuzhelov, line 99: a bit worrying about the fact that we are using reference on the object that is not fully initialized at this point. While it is fine (as only `props` are used by this method, and those come already initialized), we might still consider to make this call in constructor for better safety.

src/lib/accessibility/Behaviors/List/selectableListItemBehavior.ts
```diff
@@ -11,6 +11,7 @@ const selectableListItemBehavior: Accessibility = (props: any) => ({
     root: {
       role: 'option',
       'aria-selected': !!props['active'],
+      tabIndex: props.focusableItemProps.isFocused ? '0' : '-1',
```
kuzhelov: seems that we should at least specify type for `props` here, so that it will be clear from which context `focusableItemProps` come from. Also, we should think about following thing - what if this behavior will be applied to the component that doesn't have this prop support (for some reason, like some custom component that haven't introduced support for `FocusableContainer` - to support some custom focus handling logic, for instance)? In that case we shouldn't make `tabIndex` assignments here - otherwise behavior for `selectableListItem` will be tightly coupled with the focus handling one. 

May suggest to either
- consider to merge focus handling to be a thing that is defined by accessibility behavior (so that there won't be any code like `FocusContainer.create` in the component's code - all this will come as accessibility behavior concern). See this option is a preferable one
- or decouple these two completely - so that focus handling logic will be entirely responsible for handling `tabIndex` attributes as well

src/components/List/List.tsx
smykhailov: we can postpone it for the next round of refactoring

src/components/List/List.tsx
smykhailov: renamed

src/components/List/ListItem.tsx
smykhailov: done

src/components/List/ListItem.tsx
smykhailov: done

src/components/List/ListItem.tsx
smykhailov: done

src/components/List/ListItem.tsx
smykhailov: done

src/components/List/List.tsx
smykhailov: for future refactoring

src/lib/accessibility/Behaviors/List/selectableListItemBehavior.ts
smykhailov: For the next round of refactoring

kuzhelov:
agreed with Sergey to address several follow-up points by means of PRs that will come next. For the sake of making the roadmap clear, will file an issue item where all these findings will be summarised - and after that merge the PR 👍 

In general, lets scope the area to which this approach is applied to `List` and `ListItem` components, till the moment when all the necessary refactoring steps will be made and, as a result, general strategy of applying this functionality will reach its maturity point.

miroslavstastny (approved):

kuzhelov (approved):

kuzhelov (changes requested):
provided functionality raise the following error on switch between Shorthand API and Children API modes of List Selection example. Let sync tomorrow and fix before merging

image

kuzhelov (approved):

PR #258, opened 2018-09-20
fix(RadioGroup): fix imports of theme and utils in RadioGroup

jurokapsiar:
This fixes wrong imports in RadioGroup components


kuzhelov (approved):

jurokapsiar:
this issue occurs only in yarn link scenario when skipLibCheck is not used. postponing for now, but we can merge it if necessary

mnajdova (approved):

mnajdova:
I don't see why we won't merge it. We would at lest be consistent of how the imports are made. This was exactly the same problem with the Portal component problem, that was making version 0.5.1 not consumable. I would vote for merging.

PR #259, opened 2018-09-20
chore: Publish all master builds to npm with @next tag

miroslavstastny:
Every commit (every successful build) to master branch is automatically released to npmjs.com under `current` tag.

### How to consume master build in a project
`yarn add @stardust-ui/react@current`

### Package version format

Let's assume that version in `package.json` is set to **0.5.2**. Then the version of the master package will be something like:

`0.5.3-1721.master.51f0fc6.0`

- **0.5.3** - version higher than the current version in `package.json`,
- **- (hyphen)** - this is prerelease version, semantically before *0.5.3*,
- **1721** - CircleCI build number (assures that any build has unique version and that any new version is higher than the previous one,
- **.master.** - name of the branch (currently master only),
- **51f0fc6** - commit SHA - not necessary for the versioning but will help us easily map npm to commit,
- **.0** - always the same, automatically added by `npm version prerelease`

### Deleting old versions
Currently these prerelease versions are only added and never deleted. When this becomes a problem we can write a simple script that will delete old versions from npmjs.

### TODO
For the publish to work `npm_TOKEN` environment variable must be set in CircleCI. If it is not set, publish step is skipped.


.circleci/config.yml
```diff
@@ -55,3 +55,11 @@ jobs:
       - run:
           name: Project Tests
           command: yarn test:projects:cra-ts
+      - run:
+          name: Publish npm package
+          command: |
+            if [ ! -z ${npm_TOKEN+x} ] && [ "${CIRCLE_BRANCH}" == "master" ]; then
```
miroslavstastny: Check for `npm_TOKEN` can be used to disable publishing without any commit if something goes wrong.
It is only temporary and will be removed once the step proofs it is stable.

miroslavstastny:
### Resources
- [Semantic versioning](https://semver.org/)
- [Semver calculator](https://semver.npmjs.com/)

.circleci/config.yml
kuzhelov: couple of comments:
- why `+x` trick is necessary for parameter expansion in this condition? seems that `if [ -n "${npm_TOKEN}" ]` would do the same test (and is easier to read)
- would suggest to use double quotes for wrapping this expression

.circleci/config.yml
```diff
@@ -55,3 +55,11 @@ jobs:
...
+              echo "//registry.npmjs.org/:_authToken=${npm_TOKEN}" > ~/project/.npmrc
+              npm version prerelease --preid=${CIRCLE_BUILD_NUM}.${CIRCLE_BRANCH}.`git rev-parse --short HEAD` --no-git-tag-version
```
kuzhelov: would suggest to use quotes here as well:

```sh
 npm version prerelease --preid="${CIRCLE_BUILD_NUM}.${CIRCLE_BRANCH}.$(git rev-parse --short HEAD)" --no-git-tag-version
```

.circleci/config.yml
miroslavstastny: I am used to add `+x` to distinguish between unset and empty string. Although it is not necessary here, I consider it a defensive programming technique.
Will switch the condition to `-n` and add quotes.

.circleci/config.yml
```diff
@@ -55,3 +55,11 @@ jobs:
       - run:
           name: Project Tests
           command: yarn test:projects:cra-ts
+      - run:
+          name: Publish npm package
+          command: |
+            if [ -n "${npm_TOKEN+x}" ] && [ "${CIRCLE_BRANCH}" == "master" ]; then
+              echo "//registry.npmjs.org/:_authToken=${npm_TOKEN}" > ~/project/.npmrc
+              npm version prerelease --preid="${CIRCLE_BUILD_NUM}.${CIRCLE_BRANCH}.`git rev-parse --short HEAD`" --no-git-tag-version
```
kuzhelov, line 63: my initial thought was about not introducing any Circle CI dependencies to the version tag (like `CRICLE_BUILD_NUM`), but have a guess that you are doing this intentionally for the sake of providing some mechanism to transparently upgrade 'experimental' packages on the client side (and, thus, you do need some number that would be monotonically incremented in the version), right?

miroslavstastny: Yes, that was the idea - whatever happens you want the new build to be "semantically newer".

miroslavstastny: And the fact that this is tightly coupled with circle ci was the reason why I added it directly to circle ci config instead of yarn/gulp.

kuzhelov: still, according to the documentation, tags complements semantic versioning (in a sense that those serve as a substitution for version number) - so, the question is whether versioning will take effect for different packages published under the same tag? or the published package (with the same tag) will be just replaced in the NPM registry?

levithomason (approved):
This looks good to me.  The only things I'm reconsidering is the use of `next` as the tag.  This name is often used for the next major version of a library.

I'm wondering if we use something like `live` or `ci` or something else that indicates it is a constantly new version from latest successful build.  Thoughts?

levithomason:
Solved offline, let's go with `current` for the tag.

miroslavstastny:
Tag changed to `@current`, PR description updated

mnajdova:
Is there something blocking this PR from being merged?

PR #262, opened 2018-09-21
docs(comments): Improved docsite comments for 'variables'

hughreeling:
Simple find and replace for inline documentation about 'variables'.

Changed 
/** Custom variables to be applied for component. */
to
/** Override for theme variables to allow modifications of component styling via themes. */


src/components/Accordion/Accordion.tsx
```diff
@@ -92,7 +92,7 @@ class Accordion extends AutoControlledComponent<Extendable<IAccordionProps>, any
     /** Custom styles to be applied for component. */
     styles: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
 
-    /** Custom variables to be applied for component. */
+    /** Override for theme variables to allow modifications of component styling via themes. */
```
kuzhelov: would suggest `Override for theme site variables.. `, just to be consistent with current naming used.

kuzhelov (approved):

Issue #263, opened 2018-09-21
react-popper has a forbidden dependency for Microsoft


bmdalex:
## Bug Report

We're using `react-popper` npm package to create `Popup` components with correct positioning around the trigger elements.

This package has a dependency on `create-react-context`, a package that **specifically excludes Microsoft (among other corporations) from the list of users** : [react-popper license](https://github.com/jamiebuilds/license).

**Therefore, Microsoft cannot use Stardust starting with `v0.5.2`.**

#### 1. Use `react-popper v0.9.2`

`v0.9.2` is the last stable `react-popper` version that didn't have a dependency on `create-react-context`, as you can see in: [feat: New react-popper API #117](https://github.com/FezVrasta/react-popper/pull/117).

We can update our `package.json` `"react-popper": "^1.0.2",` to `"react-popper": "^0.9.2",` and go from there.

**Drawbacks:**
- the API has changed  drastically from `v0.9.2` to `v1.0.2` (hence the major change) so we need to reimplement the `Popper` integration in our `Popup` implementation;
- no support, we're exposing ourselves to existing bugs in `react-popper v0.9.2`;
- we won't have the fixes for the bugs `react-popper` had until `v0.9.2`;

#### 2. Use `popper.js` directly and reimplement `react-popper` React layer
We can take a look at `react-popper` implementation of the `React` layer on top of `popper.js` and try implement ourselves only the bits we need for our `Popup` scenarios.

**Drawbacks:**
- will take the most time as there is plenty of work to be done here; this will be like a small library;
- we're basically reinventing the wheel since it's the same work as `react-popper`.

#### 3. Implement custom `Popup` positioning
`Semantic UI React` already implements custom positioning for `Popup` and we were initially relying on that: [SUIR Poppup.js](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Popup/Popup.js)

**Drawbacks:**
- the implementation in SUIR is suboptimal and incomplete;
- needs further work to handle all positioning cases (RTL, fitting in viewport, etc) and extensive testing.

bmdalex:
Feedback much appreciated
@levithomason @mnajdova @alinais @miroslavstastny @kuzhelov

kuzhelov:
@Bugaa92 here is a commit where he license of this package has been reverted back to true MIT one:

[...]

PR #264, opened 2018-09-21
fix(popup): fix licensing issue for create react context

bmdalex:
# Popup `create-react-context` license fix

This PR fixes the #263 issue by referencing correct package versions unaffected by the license file that prohibits Microsoft from using it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -169,5 +169,8 @@
   "peerDependencies": {
     "react": ">=0.14.0 <= 16",
     "react-dom": ">=0.14.0 <= 16"
+  },
+  "resolutions": {
+    "create-react-context": "0.2.2"
   }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -1592,9 +1592,9 @@ create-hmac@^1.1.0, create-hmac@^1.1.2, create-hmac@^1.1.4:
     safe-buffer "^5.0.1"
     sha.js "^2.4.8"
 
-create-react-context@^0.2.1:
-  version "0.2.3"
-  resolved "https://registry.yarnpkg.com/create-react-context/-/create-react-context-0.2.3.tgz#9ec140a6914a22ef04b8b09b7771de89567cb6f3"
+create-react-context@0.2.2, create-react-context@^0.2.1:
+  version "0.2.2"
+  resolved "https://registry.yarnpkg.com/create-react-context/-/create-react-context-0.2.2.tgz#9836542f9aaa22868cd7d4a6f82667df38019dca"
   dependencies:
     fbjs "^0.8.0"
     gud "^1.0.0"


yarn.lock
```diff
@@ -1592,7 +1592,7 @@ create-hmac@^1.1.0, create-hmac@^1.1.2, create-hmac@^1.1.4:
     safe-buffer "^5.0.1"
     sha.js "^2.4.8"
 
-create-react-context@^0.2.1:
+create-react-context@^0.2.3:
```
kuzhelov: heads up, this won't fix the issue - consider to review text of LICENSE for this version, we won't be able to comply. Consider to enforce version 0.2.2 instead (or lower) - and as the newer version will be published (assuming the LICENSE would still be MIT) we will be able to update the version here.

bmdalex: I don't understand , 0.2.3 has MIT license..

bmdalex: oh... the license was introduced in 0.2.3...and the removal directly to master and was not released yet...

kuzhelov: `create-react-context@0.2.2`

kuzhelov: agreed to make changes in package.json instead, to include `resolutions` section: https://yarnpkg.com/lang/en/docs/selective-version-resolutions/

bmdalex: done, @kuzhelov can u please take another look?

yarn.lock
```diff
@@ -1592,9 +1592,9 @@ create-hmac@^1.1.0, create-hmac@^1.1.2, create-hmac@^1.1.4:
     safe-buffer "^5.0.1"
     sha.js "^2.4.8"
 
-create-react-context@^0.2.1:
-  version "0.2.3"
-  resolved "https://registry.yarnpkg.com/create-react-context/-/create-react-context-0.2.3.tgz#9ec140a6914a22ef04b8b09b7771de89567cb6f3"
+create-react-context@0.2.2, create-react-context@^0.2.1:
```
kuzhelov, line 1595: this change seems to be unnecessary - would rather avoid touching yarn artefacts, Let me try this out..

package.json
```diff
@@ -94,6 +94,9 @@
     "tmp": "^0.0.33",
     "what-input": "^5.1.2"
   },
+  "resolutions": {
```
kuzhelov: nit: it is better to put this section after dependencies sections (i.e. `dependencies`, `devDependencies`, `peerDependencies`) - that's the general practice

yarn.lock
bmdalex: this was generated by yarn itself, only change I made was in `package.json`

package.json
bmdalex: will change

bmdalex: done

kuzhelov (approved):

PR #269, opened 2018-09-24
docs: classname and style clarification

hughreeling:
Attempting to better describe how classname and style properties are used.

## className

*before*
```js
    /** Additional classes. */
    className: PropTypes.string,
```

*after*
```js
   /** Additional css class name or space separated class names to apply **/
    className: PropTypes.string,
```

## styles
*before*
```js
    /** Custom styles to be applied for component. */
    styles: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
```

*after*
```js
    /** Styles to apply to this component instance **/
    styles: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
```

### DIFF
--- a/src/components/Accordion/Accordion.tsx
+++ b/src/components/Accordion/Accordion.tsx
@@ -55,7 +55,7 @@ class Accordion extends AutoControlledComponent<Extendable<IAccordionProps>, any
     /** Primary content. */
     children: PropTypes.node,
 
-    /** Additional css class name or space separated class names to apply **/
+    /** Additional CSS class name(s) to apply.  */
     className: PropTypes.string,
 
     /** Initial activeIndex value. */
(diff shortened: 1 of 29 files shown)


src/components/Accordion/Accordion.tsx
```diff
@@ -89,7 +89,7 @@ class Accordion extends AutoControlledComponent<Extendable<IAccordionProps>, any
     /** Accessibility behavior if overridden by the user. */
     accessibility: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
 
-    /** Custom styles to be applied for component. */
+    /** Styles to apply to this component instance **/
```
kuzhelov: seems that we might consider a more explicit way to express the same thing, like
`Additional CSS styles to apply to ... `

This will make this comment more descriptive, as well as will go along the content provided in other comments in terms of specificity (`className` as an example)

src/components/Accordion/Accordion.tsx
```diff
@@ -55,7 +55,7 @@ class Accordion extends AutoControlledComponent<Extendable<IAccordionProps>, any
     /** Primary content. */
     children: PropTypes.node,
 
-    /** Additional classes. */
+    /** Additional css class name or space separated class names to apply **/
```
kuzhelov: this might be simplified a bit, as `className` is React concept, so the `space-separated` aspects could be omitted from the description:

`Additional css class name(s) to apply`

kuzhelov: nit: just for the sake of consistency, we should finish these comments with dot

kuzhelov (approved):

PR #270, opened 2018-09-24
feat(Avatar): added image&label shorthands

mnajdova:
# Avatar

Introducing Avatar's image and label shorthands. Introduced some improvements in the styles and variables definition of the component.

### TODO

- [x] Conformance test
- [ ] Stardust base theme
- [x] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [x] Confirm RTL usage
- [x] Update the CHANGELOG.md

# API Proposal

## image: ItemShorthand

Shorthand for image inside the Avatar component. Provided examples about how this shorthand can be used for generating Icon instead of Image.

```jsx
import React from 'react'
import { Avatar, Icon } from '@stardust-ui/react'

const iconAsImage = (Image, props, children) => {
  return <Icon name={'user'} circular variables={{ color: 'blue' }} styles={{ fontSize: '16px' }} />
}

const AvatarExampleImageCustomizationShorthand = () => (
  <>
    <Avatar
      image={{ src: 'public/images/avatar/small/matt.jpg', alt: 'Profile picture of John Doe' }}
      status={{ color: 'green', icon: 'check', title: 'Available' }}
    />
    &emsp;
    <Avatar image={iconAsImage} status={{ color: 'green', icon: 'check', title: 'Available' }} />
  </>
)

export default AvatarExampleImageCustomizationShorthand
```

```html
<div class="ui-avatar">
  <img src="public/images/avatar/small/matt.jpg" alt="Profile picture of John Doe" class="ui-image" />
  <span title="Available" class="ui-status">
  </span>
</div>
<div class="ui-avatar">
  <span class="ui-icon ab" aria-hidden="true"></span>
  <span title="Available" class="ui-status">
  </span>
</div>
```

image

## label: ItemShorthand
Shorthand for label inside the Avatar component. 

```jsx
<Avatar name="John Doe" label={{ variables: { backgroundColor: 'pink' } }} />
```

```html
<div class="ui-avatar">
  <div title="John Doe" class="ui-label">
    <div class="ui-layout">
      <div class="ui-layout__main">JD</div>
    </div>
  </div>
</div>
```

image

### DIFF
--- a/docs/src/examples/components/Avatar/Variations/AvatarExampleImageCustomization.shorthand.tsx
+++ b/docs/src/examples/components/Avatar/Variations/AvatarExampleImageCustomization.shorthand.tsx
@@ -1,0 +1,19 @@
+import React from 'react'
+import { Avatar, Icon } from '@stardust-ui/react'
+
+const iconAsImage = (Image, props, children) => {
+  return <Icon name={'user'} circular variables={{ color: 'blue' }} styles={{ fontSize: '16px' }} />
+}
+
+const AvatarExampleImageCustomizationShorthand = () => (
+  <>
+    <Avatar
+      image={{ src: 'public/images/avatar/small/matt.jpg', alt: 'Profile picture of John Doe' }}
+      status={{ color: 'green', icon: 'check', title: 'Available' }}
+    />
+    &emsp;
+    <Avatar image={iconAsImage} status={{ color: 'green', icon: 'check', title: 'Available' }} />
+  </>
+)
+
+export default AvatarExampleImageCustomizationShorthand
--- a/docs/src/examples/components/Avatar/Variations/AvatarExampleImageCustomization.shorthand.tsx
+++ b/docs/src/examples/components/Avatar/Variations/AvatarExampleImageCustomization.shorthand.tsx
@@ -2,7 +2,7 @@ import React from 'react'
 import { Avatar, Icon } from '@stardust-ui/react'
 
 const iconAsImage = (Image, props, children) => {
-  return <Icon name={'user'} circular variables={{ color: 'blue' }} styles={{ fontSize: '16px' }} />
+  return <Icon name="user" circular variables={{ color: 'blue' }} styles={{ fontSize: '16px' }} />
 }
 
 const AvatarExampleImageCustomizationShorthand = () => (
--- a/src/components/Avatar/Avatar.tsx
+++ b/src/components/Avatar/Avatar.tsx
@@ -47,7 +47,7 @@ class Avatar extends UIComponent<Extendable<IAvatarProps>, any> {
     /** An element type to render as (string or function). */
     as: customPropTypes.as,
 
-    /** Additional classes. */
+    /** Additional CSS class name(s) to apply.  */
     className: PropTypes.string,
 
     /** The name used for displaying the initials of the avatar if the image is not provided. */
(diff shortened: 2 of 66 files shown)


hughreeling (approved):

src/themes/teams/components/Avatar/avatarStyles.ts
```diff
@@ -11,10 +11,10 @@ const avatarStyles: IComponentPartStylesInput<IAvatarPropsWithDefaults, any> = {
     height: pxToRem(size),
     width: pxToRem(size),
   }),
-  imageAvatar: (): ICSSInJSStyle => ({
+  image: (): ICSSInJSStyle => ({
```
kuzhelov, line 14: 👍

docs/src/examples/components/Avatar/Variations/AvatarExampleImageCustomization.shorthand.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import { Avatar, Icon } from '@stardust-ui/react'
+
+const iconAsImage = (Image, props, children) => {
+  return <Icon name={'user'} circular variables={{ color: 'blue' }} styles={{ fontSize: '16px' }} />
```
kuzhelov: seems that the expression for `name` prop could be safely simplified to use just quotes

src/components/Avatar/Avatar.tsx
```diff
@@ -101,30 +101,35 @@ class Avatar extends UIComponent<Extendable<IAvatarProps>, any> {
...
+          defaultProps: {
+            fluid: true,
+            avatar: true,
+            title: name,
+            styles: styles.image,
+            variables: variables.image,
+          },
+        })}
+        {!image &&
+          Label.create(label || {}, {
```
kuzhelov: currently it is possible to use component in the following way:

```jsx
<Avatar name='John Doe' label={{ content: 'A' }} />
```

In that case the `name`'s value will be completely ignored. Could we consider `name` to be a shorthand slot (with `Label` being a default component)?

src/themes/teams/components/Avatar/avatarVariables.ts
```diff
@@ -1,4 +1,7 @@
 export default () => ({
-  statusBorderColor: 'white',
-  statusBorderWidth: 2,
+  label: { padding: '0px' },
+  status: {
```
kuzhelov: I would rather suggest to refrain from utilizing this pattern of nested (child) variables for now, as we are not completely sure if it is a way to go for us (couple of pending prototypes are under review currently). It seems that we might consider to hardcode these styles for now (in `avatarStyles` file) - although I do understand that we will lose a bit in terms of customization capabilities for the examples, at the same time we 
- won't restrict the functionality for Teams as a consumer
- would be at much safer spot in terms of using only the approaches that we have agreed upon

src/components/Avatar/Avatar.tsx
mnajdova: The name and label are two different things. The name is used as a title for the Label and Image components inside the Avatar, but the content of the Label is actually the result of getInitials method (prop), that uses the names for generating the initials. The use can however override the content of the Label using the shorthand, and the name can still reflects the actual user's name.

src/themes/teams/components/Avatar/avatarVariables.ts
mnajdova: Okay, will try to replace then all the variables that are referring some components slots with the slot's inside the styles.

mnajdova: I reverted the changes I introduced for the status slot, and removed the label's variables. I wasn't able to get rid of the variables for the status, as these variables are used in multiple places for different calculation inside the statusStyles.

src/components/Avatar/Avatar.tsx
kuzhelov: frankly, a bit unsure whether we do need both `label` and `name` at the same time - it seems that for almost all cases it would be sufficient to have just one prop. Let's please, discuss this tomorrow

kuzhelov (approved):

PR #271, opened 2018-09-24
fix: Generate correct paths in es/../Provider.d.ts

miroslavstastny:
Wrong path in v0.6.0:
dist/es/components/Provider/Provider.d.ts:
```jsx
import("../../../../src/components/Provider/ProviderConsumer")
```

Fixed by this PR:
dist/es/components/Provider/Provider.d.ts:
```jsx
import("./ProviderConsumer")
```

### DIFF
--- a/build/tsconfig.es.json
+++ b/build/tsconfig.es.json
@@ -2,7 +2,6 @@
   "extends": "./tsconfig.common.json",
   "compilerOptions": {
     "module": "esnext",
-    "target": "es5",
-    "outDir": "../dist/es"
+    "target": "es5"
   }
 }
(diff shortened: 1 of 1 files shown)


hughreeling (approved):

kuzhelov (approved):

build/tsconfig.es.json
```diff
@@ -2,7 +2,6 @@
   "extends": "./tsconfig.common.json",
   "compilerOptions": {
     "module": "esnext",
-    "target": "es5",
-    "outDir": "../dist/es"
+    "target": "es5"
```
kuzhelov: sorry, I remember that it was my suggestion - but it seems that there is already the following line in the `tsconfig.common.json`: `"target": "es5"`. So, it seems that there is no reason to duplicate this :) will fix

PR #272, opened 2018-09-24
docs: hiding children props in doc site

hughreeling:
In current docSite in Stardust we have 

Name | Default | Type | Description
-- | -- | -- | --
accessibility | accessibility_1.ButtonBehavior | {object\|func} | Accessibility behavior if overridden by the user.
as | 'button' | {custom} | An element type to render as (string or function).
children |   | {node} | Primary content.
circular |   | {bool} | A button can appear circular.

The children property here is confusing - it's used internally for the Children API implementation, but should not be called by consumers of the library.

Adding @docsiteIgnore tag to the property, so in component it's marked

    /**
     *  Used to set content when using childrenApi - internal only
     *  @docSiteIgnore
     */
    children: PropTypes.node,

which will hide this in the generated docSite.


docs/src/components/ComponentDoc/ComponentTable/ComponentTableRow.tsx
```diff
@@ -49,4 +51,13 @@ export default class ComponentTableRow extends React.Component<any, any> {
...
     )
   }
+
+  docSiteHidden(tags) {
+    const hide = _.filter(tags, { title: 'docSiteIgnore' })
+    return !!hide.length
+  }
 }
+
+export const hidden = `
```
hughreeling: Remove this - un-used

docs/src/components/ComponentDoc/ComponentTable/ComponentTableRow.tsx
```diff
@@ -49,4 +51,9 @@ export default class ComponentTableRow extends React.Component<any, any> {
       </Table.Row>
     )
   }
+
+  docSiteHidden(tags) {
+    const hide = _.filter(tags, { title: 'docSiteIgnore' })
```
miroslavstastny:
```jsx
return _.some(tags, { title: 'docSiteIgnore' })
```

miroslavstastny (approved):

mnajdova (approved):

PR #273, opened 2018-09-24
fix(test): disable skipLibCheck in test app

kuzhelov:
We've agreed to remove `skipLibCheck` flag from `tsconfig` that is used for Stardust test app - in that case we will be able to precisely follow the following plan of consuming Stardust on the client side:
- create app by `create-react-app` util
- add Stardust as a dependency
- use some Stardust components in the app
- try to build the app


kuzhelov:
closing for now as this will result in guaranteed failed test, due to #276.

PR #107, opened 2018-08-17
RFC: generate PropTypes, handledProps

layershifter:
# RFC

Hey there!

I've made a dirty POC to check that it's possible. My proposal contains some usefull things.

### `propTypes`
Addresses #7. I almost sure that we should generate `propTypes` from interfaces, this PR is an initial POC for this. We also can to add there the environment checks.

### `handledProps`

As we use AST transforms, we can easily add `handledProps`, too 👍 

----

This PR heavily relies on [ttypescript](https://github.com/cevek/ttypescript), please let me know what you think about this.

PR #275, opened 2018-09-24
feat: remove handled props declarations

kuzhelov:
**TL;DR** This PR aims to remove source of huge code duplication, for the sake of significantly alleviating maintainability scenarios. All the previous functionality (as well as set of supported scenarios) remains to be the same.

Follows discussion opened in https://github.com/stardust-ui/react/pull/107#issuecomment-414513996

---------------

The problem that we've had before is that during component development cycle we need to introduce, essentially, the same information about component's props in three different places:

```js
// Button
 public static propTypes = {
    as: customPropTypes.as,
   ...
}
...
public static handledProps = [ ..., 'as', ... ]
...
export interface IButtonProps {
  as?: any
  ...
}
....
```

This PR proposes to infer `handledProps` from the `propTypes` static property, as the set of `handledProps` values  in most cases just duplicates the keys of `propTypes`. In case when it would be necessary to pass some property to children element (and, thus, to mark this specific property as `unhandled`), there is `unhandledProps` array that component could provide the values for:

```js
// Button
 public static propTypes = {
    /** Suppose that this property should be passed through to children */
    foo: customPropTypes.as,
   ...
}

public static unhandledProps = [ 'foo' ]
```

It is worth to notice that this case of `unhandledProps` is extremely rare (so it is quite unlikely that developer would need to introduce this array for component in average case) - thus, with this strategy being applied, we would be able to make huge savings in typing, as well as avoid code duplication that imposes maintainability burden.


hughreeling (approved):
Like it!

alinais (approved):

src/lib/UIComponent.tsx
```diff
@@ -7,7 +8,21 @@ class UIComponent<P, S> extends React.Component<P, S> {
...
+
+  static propTypes: any
+
+  /** Array of props to exclude from list of handled ones. */
+  static unhandledProps: string[] = []
+
+  private static _handledPropsCache: string[] = undefined
+  static get handledProps() {
+    if (!this._handledPropsCache) {
+      this._handledPropsCache = _.pullAll(_.keys(this.propTypes), this.unhandledProps).sort()
```
miroslavstastny: It makes no difference at the end, but wouldn't using `difference()` instead of `pullAll()` be cleaner? (difference creates a new array while pullAll modifies the source one)

miroslavstastny (approved):

levithomason (approved):
<3

PR #277, opened 2018-09-25
feat(Text): add "me" option to `atMention` prop

codepretty:
atMention text can be of 2 types, myself or another person

**Example**
image

**Usage**
```jsx
<Text atMention="me">@Mention Me</Text>
<Text atMention="other">@Mention Another</Text>
```

### DIFF
--- a/src/components/Text/Text.tsx
+++ b/src/components/Text/Text.tsx
@@ -42,7 +42,7 @@ class Text extends UIComponent<Extendable<ITextProps>, any> {
     as: customPropTypes.as,
 
     /** At mentions can be formatted to draw users' attention. Mentions for "me" can be formatted to appear differently. */
-    atMention: PropTypes.oneOfType([PropTypes.bool, 'me']),
+    atMention: PropTypes.oneOfType([PropTypes.bool, PropTypes.string]),
 
     /** Additional CSS class name(s) to apply.  */
     className: PropTypes.string,
--- a/src/components/Text/Text.tsx
+++ b/src/components/Text/Text.tsx
@@ -8,7 +8,7 @@ import { ComponentVariablesInput, ComponentPartStyle } from '../../../types/them
 
 export interface ITextProps {
   as?: any
-  atMention?: boolean
+  atMention?: 'me' | 'other'
   className?: string
   content?: any
   disabled?: boolean
@@ -41,8 +41,8 @@ class Text extends UIComponent<Extendable<ITextProps>, any> {
     /** Change the default element type of the Text component */
     as: customPropTypes.as,
 
-    /** Set as @mention Text component */
-    atMention: PropTypes.bool,
+    /** An @mention Text component can be formatted for myself mentioned or an other person mentioned */
+    atMention: PropTypes.oneOf(['me', 'other']),
 
     /** Additional CSS class name(s) to apply.  */
     className: PropTypes.string,
--- a/src/components/Text/Text.tsx
+++ b/src/components/Text/Text.tsx
@@ -41,8 +41,8 @@ class Text extends UIComponent<Extendable<ITextProps>, any> {
     /** Change the default element type of the Text component */
     as: customPropTypes.as,
 
-    /** An @mention Text component can be formatted for myself mentioned or an other person mentioned */
-    atMention: PropTypes.oneOfType(['me', PropTypes.bool]),
+    /** At mentions can be formatted to draw users' attention. Mentions for "me" can be formatted to appear differently. */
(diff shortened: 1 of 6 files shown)


kuzhelov:
it seems that at this point we should really be  concerned about Teams-domain aspects propagating into core Stardust API. Probably, this is the right point to introduce wrapper for the `Text` component on the side of Teams, (let me name it `TextWrapper`), and introduce `atMention` property with all the necessary values for it

src/components/Text/Text.tsx
```diff
@@ -8,7 +8,7 @@ import { ComponentVariablesInput, ComponentPartStyle } from '../../../types/them
 
 export interface ITextProps {
   as?: any
-  atMention?: boolean
+  atMention?: 'me' | 'other'
```
levithomason: I would think that we would still support the `boolean` case here.  Some themes (like GitHub) don't have the concept of `me` vs `other`.  In those cases, themes have the need of styling based on the presence of this prop alone (boolean).

```jsx
<Text atMention>@someone</Text>
<Text atMention="me">@me</Text>
<Text atMention="other">@other</Text>
```

What these `atMention` values look like or whether or not they look different between their usages is up to the theme.

src/components/Text/Text.tsx
```diff
@@ -41,8 +41,8 @@ class Text extends UIComponent<Extendable<ITextProps>, any> {
     /** Change the default element type of the Text component */
     as: customPropTypes.as,
 
-    /** Set as @mention Text component */
-    atMention: PropTypes.bool,
+    /** An @mention Text component can be formatted for myself mentioned or an other person mentioned */
```
levithomason: Phrasing is a bit odd here.  There is no "@mention Text component".  There is a "Text component" that can be used to show an "at mention".  One potential rewrite here might be:

```jsx
/**
 * At mentions can be formatted to draw users' attention.
 * Mentions for "me" can be formatted to appear differently.
 */
```

src/themes/teams/components/Text/textStyles.ts
```diff
@@ -27,7 +27,15 @@ export default {
   }: TextStylesParams): ICSSInJSStyle => {
     return {
       ...(truncated && truncateStyle),
-      ...(atMention && { color: v.atMentionTextColor }),
+      ...(atMention &&
+        atMention === 'other' && {
```
levithomason: I would think that we would want to also handle the boolean case here as the "default" at mention formatting.  The exception to the `atMention` usage would be mentions to `me`.  So my proposal would be the majority case (just a boolean prop) would use the majority styling (other).

```jsx
atMention === true || atMention === 'other' && ...
```

src/themes/teams/components/Text/textStyles.ts
```diff
@@ -27,7 +27,15 @@ export default {
...
+          color: v.atMentionOtherTextColor,
+        }),
+      ...(atMention &&
+        atMention === 'me' && {
```
levithomason: No need to check for `atMention &&` first.  If there is no `atMention` then the next logical check `atMention === 'me'` will fail.  It is sufficient to only have this one check.

levithomason (changes requested):

levithomason:
Updating everyone so we're on the same page:

> it seems that at this point we should really be concerned about Teams-domain aspects propagating into core Stardust API. Probably, this is the right point to introduce wrapper for the `Text` component on the side of Teams, (let me name it `TextWrapper`), and introduce `atMention` property with all the necessary values for it

**Response:** In Bellevue we initially shared these same thoughts.  However, after considering the "mention" usage in other chat applications the concept of "my mentions" vs "regular mentions" does exist.  Since we already have the `atMention` prop, we decide it makes sense to allow this distinction at the API level.

**Conclusion:** At the time of writing, both Prague and Bellevue have been updated on this and we're on the same page as far as accepting these Text changes for now.

**Forward thinking:** There is probably a refactor warranted to the Text component at some future time as it is currently taking on a lot of responsibility.  It is needs to have props for various _usages_ (important, etc) as well as props for _utility_ (fontWeight, etc).  However, the `atMention` prop may very well become a separate component entirely.

src/components/Text/Text.tsx
levithomason: Actually, thinking through https://github.com/stardust-ui/react/pull/277#discussion_r220693137 more carefully I think we only need `boolean` and `me`, correct?  An `atMention` is a generic mention while `atMention="me"` is the only specifier needed for custom formatting.

levithomason (approved):

PR #278, opened 2018-09-25
fix: (ListItem) - Fix endMedia to not be removed from DOM on mouseleave for ListItem

musingh1:
Earlier, we were listening for mouseEnter and mouseLeave, maintaining the isHovering state and only showing the endMedia when the item is hovered. This breaks the scenarios where endMedia is a Popup. Once the Popup is triggered we want the content to stay even after the user leaves the list item (e.g. points to the popup content).
So now, if provided we will always render all the media's and in Teams theme control the display of endMedia, headerMedia etc through styles.


src/themes/teams/components/List/listItemStyles.ts
```diff
@@ -37,12 +50,10 @@ const listItemStyles: IComponentPartStylesInput<IListItemProps, any> = {
     lineHeight: variables.headerLineHeight,
   }),
   headerMedia: ({ variables }): ICSSInJSStyle => ({
-    color: variables.headerMediaColor,
```
musingh1, line 40: @levithomason , please let me know if this is still needed? I can add it in the variable and use the override from the root styles. But anyways, I am not sure if having headerMediaColor in the variables is needed here since it should anyways get inverted when hovering for selection...

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixes
+- Fix endMedia to not be removed from DOM on mouseleave for `ListItem` ([#278](https://github.com/stardust-ui/react/pull/278))
```
miroslavstastny: This is missing your github @username.

musingh1: Thanks for catching this. Will update. 

Somehow I had copied and updated the markup from the only entry which did not have it. :)
### Fixes
- Fixed fonts to support zwj gender emojis ([#215](https://github.com/stardust-ui/react/pull/215))

jurokapsiar:
@musingh1 the same would need to be done for focus state if the user is using keyboard. Would you mind adding following commit to the PR: https://github.com/stardust-ui/react/compare/feature/mukul/fix-listitem-styles...feat/listitem-focus-styles

or should we create a followup PR after this one is merged?

src/themes/teams/components/List/listItemStyles.ts
```diff
@@ -2,16 +2,32 @@ import { pxToRem } from '../../../../lib'
 import { IComponentPartStylesInput, ICSSInJSStyle } from '../../../../../types/theme'
 import { IListItemProps } from '../../../../components/List/ListItem'
 
+const hoverStyle = variables => ({
+  background: variables.selectionHoverBackgroundColor,
+  color: variables.selectionHoverColor,
+  cursor: 'pointer',
+
+  '& .ui-item-layout__header': { color: 'inherit' },
```
kuzhelov, line 10: this will be rendered by Fela as the following CSS selector
```css
.a .ui-item-layout__header
```

where `.a` is class of `ListItem`'s  DOM element. The problem I see here is that these styles will be applied to all `.ui-item-layout__header` subelements that happen to be **_anywhere_** in the child tree of `ListItem` - although our intent was to provide these styles for direct `.ui-item-layout__header` child only. And with this encapsulation principle being broken, we might end up with unexpected styling effects

musingh1: We could use '>' selector instead of ' '. I will check that later today.

kuzhelov: direct child selector, unfortunately, won't work in this case, because rendered DOM tree is like the following for the list item:
```html
<div .ui-list__item>
  ..
  <div .ui-layout__main>  // this one prevents direct child selector to work
     ...
     <div .ui-item-layout__header>
```

musingh1: As discussed offline - "we are not making things any worse"

docs/src/examples/components/List/Content/ListExampleEndMedia.tsx
```diff
@@ -4,18 +4,21 @@ import { List } from '@stardust-ui/react'
 const ellipsis = <span>&hellip;</span>
 
 const ListExample = () => (
-  <List>
+  <List selection>
```
kuzhelov: 👍

.github/CONTRIBUTING.md
```diff
@@ -11,6 +11,7 @@ CONTRIBUTING
 - [Accessibility](#accessibility)
   - [Role and aria props](#role-and-aria-props)
   - [Focus](#focus)
+  - [Keyboard handling](#keyboard-handling)
```
kuzhelov: not sure how this is here

src/themes/teams/components/List/listItemStyles.ts
```diff
@@ -2,16 +2,32 @@ import { pxToRem } from '../../../../lib'
...
     ...(selection && {
       position: 'relative',
 
-      ':hover': {
-        background: 'rgba(98, 100, 167, .8)',
-        color: '#fff',
-        cursor: 'pointer',
-      },
+      // show the end media on hover
+      '& .ui-item-layout__endMedia': { display: 'none' },
```
kuzhelov, line 27: seems that comment contradicts the code here

src/themes/teams/components/List/listItemStyles.ts
```diff
@@ -2,16 +2,32 @@ import { pxToRem } from '../../../../lib'
...
+
+  '& .ui-item-layout__header': { color: 'inherit' },
+  '& .ui-item-layout__content': { color: 'inherit' },
+
+  // hide the header media and content media on hover
+  '& .ui-item-layout__headerMedia': { display: 'none', color: 'inherit' },
+  '& .ui-item-layout__contentMedia': { display: 'none', color: 'inherit' },
+
+  // show the end media on hover
+  '& .ui-item-layout__endMedia': { display: 'unset', color: 'inherit' },
```
kuzhelov: heads up, this `unset` value is not supported for IE. is there any other strategy how we could handle it? Seems that we have all the necessary data here to apply proper `display` value directly

src/themes/teams/components/List/listItemStyles.ts
```diff
@@ -2,16 +2,32 @@ import { pxToRem } from '../../../../lib'
 import { IComponentPartStylesInput, ICSSInJSStyle } from '../../../../../types/theme'
 import { IListItemProps } from '../../../../components/List/ListItem'
 
+const hoverStyle = variables => ({
```
kuzhelov, line 5: do we see these hover styles being generically applied across Teams theme to all List items? Frankly, would doubt that (but please, feel free to object and provide the points where I am wrong). To me it seems to be quite reasonable to expect that at some places in Teams the ListItem would not require to hide header and content media on hover. 

If this assumption is correct, shouldn't we introduce component wrapper for `ListItem` instead (`TeamsMessageListItem`) - and provide these styles to be merged on top of the ones of `ListItem`?

kuzhelov (commented):
lets address comments before merging that. 

Also, there are couple problems appearing when List Item example is rendered in Docs:
- Received `true` for a non-boolean attribute `selection`. ..
- child in an array or iterator should have a unique "key" prop

Lets address them as well before merging (I may assist in that :).

.github/CONTRIBUTING.md
bmdalex: it's because this change always shows up in the git changelog whenever we switch to a new branch... a lot of us are experiencing the same... @alinais any ideas why it's happening?

levithomason: There is a script `yarn build:docs` which eventually runs `gulp build:docs:toc`.  That gulp task is responsible for generating the table of contents for markdown files.

If you change a markdown file's headings, the table of contents will be out of date.  Since we don't run doc builds prepush (it is super slow), there is a chance that you can edit a markdown's headings and merge a PR without having first run a doc build in your PR.  A subsequent PR will likely update since users run the docs after pulling the latest master changes for the base of their PR.

### The fix

We could introduce some logic to check if markdown files were changed and if so, run `gulp build:docs:toc` to regenerate the table of contents.  Low priority and not that big of an issue at present.  Let's save this discussion and work for another PR.

src/themes/teams/components/List/listItemStyles.ts
musingh1: switching to using display:block

docs/src/examples/components/List/Content/ListExampleEndMedia.tsx
musingh1: Cannot make endMedia example selectable because of the issue - https://github.com/stardust-ui/react/issues/370

src/themes/teams/components/List/listItemStyles.ts
musingh1: We checked the chat list and calling roster list behavior. This seems to be the default styles, so will keep it like this and in case we want a different style (Teams/Channels list), we would add a wrapper there.

src/themes/teams/components/List/listItemStyles.ts
musingh1: Fixed. Thanks.

docs/src/examples/components/List/Content/ListExampleEndMedia.tsx
```diff
@@ -8,14 +8,17 @@ const ListExample = () => (
     <List.Item
       content="Program the sensor to the SAS alarm through the haptic SQL card!"
       endMedia={ellipsis}
+      selection
```
kuzhelov, line 11: ideally this prop should be propagated from `List` to `List.Item`s when is set for `List`

docs/src/examples/components/List/Variations/ListExampleTruncate.shorthand.tsx
```diff
@@ -3,20 +3,23 @@ import { List, Image } from '@stardust-ui/react'
 
 const items = [
   {
+    key: 'irving',
```
kuzhelov, line 6: 👍

kuzhelov (approved):

docs/src/examples/components/List/Content/ListExampleEndMedia.tsx
musingh1: When using children api, we are directly rendering them, and not propagating any props.
{childrenExist(children) ? children : this.renderItems()}

@kuzhelov , @levithomason , @alinais , I will open a separate bug for this.

musingh1: https://github.com/stardust-ui/react/issues/412

PR #279, opened 2018-09-26
feat(list): enter and spacebar keyboard handling for List and Listitem

jurokapsiar:
This PR adds focus zone and Enter/Space keyboard handlers to selection list.


src/lib/accessibility/Behaviors/List/SelectableListItemBehavior.ts
```diff
@@ -11,6 +13,14 @@ const SelectableListItemBehavior: (props: any) => IAccessibilityDefinition = (pr
     root: {
       role: 'option',
       'aria-selected': !!props['active'],
+      [IS_FOCUSABLE_ATTRIBUTE]: true,
+    },
+  },
+  keyActions: {
+    root: {
+      performClick: {
+        keyCombinations: [{ keyCode: keyboardKey.Enter }, { keyCode: keyboardKey.Spacebar }],
```
tomasiser: Spacebar scrolls page down, so we should add some clever way of preventing default of certain key actions. There is already the same problem with `Menu` (you can try in docsite). What do you think?

src/themes/teams/components/List/listItemStyles.ts
```diff
@@ -1,12 +1,21 @@
...
 
 const listItemStyles: IComponentPartStylesInput<IListItemProps, any> = {
   root: ({ props: { selection, important } }): ICSSInJSStyle => ({
     ...(selection && {
       position: 'relative',
 
+      ':focus': {
+        ...(isFromKeyboard && {
+          background: 'rgba(98, 100, 167, .8)',
+          color: '#fff',
```
bmdalex: we should use variables for this; steps:
- look in `siteVariables.ts` and find the colors there, for `#fff` it's `white`;
- go to `listItemVariables.ts` and define your variables there, like:
```typescript:
contentColorFocus: siteVars.xxx,
backgroundColorFocus: siteVars.white,
```
- add `variables` prop to the argument of `root` slot in this file on line 7 and replace the hardcoded colors with the defined variables

src/themes/teams/components/List/listItemStyles.ts
```diff
@@ -1,12 +1,21 @@
...
+        }),
+        outline: 0,
+      },
+
       ':hover': {
         background: 'rgba(98, 100, 167, .8)',
```
bmdalex: looks like we need a color for this, can we find the hex value in teams colors and add it to `siteVariables.ts` if not there?

src/lib/accessibility/Behaviors/List/SelectableListItemBehavior.ts
jurokapsiar: yes - the user will need to do preventDefault for now, we do not have a generic solution yet.

src/themes/teams/components/List/listItemStyles.ts
jurokapsiar: that is already done in https://github.com/stardust-ui/react/pull/278 so I would wait for that one to be merged

kolaps33:
When I tried to navigate into the list with 
Virtual cursor/JAWS and Browse Mode/NVDA list did not get focus at all.

Figure out that it causing the role='presentation' which is generated on the div about the list. 

Find out if JAWS user navigate to listbox with virtual cursor, then need to press space activate navigation in listbox:
https://jfw.groups.io/g/main/topic/selecting_in_a_list_box_jaws/5608353?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,5608353

src/themes/teams/components/List/listItemStyles.ts
jurokapsiar: this is not part of the PR anymore, was covered by other changes.

sophieH29 (approved):

test/specs/components/List/ListItem-test.tsx
```diff
@@ -0,0 +1,31 @@
...
+      listItem.simulate('keydown', { keyCode: 13 })
+      expect(onClick).toHaveBeenCalled()
+    })
+
+    test('is executed when Spacebar is pressed', () => {
+      const onClick = jest.fn()
+      const listItem = mountWithProvider(
+        <ListItem accessibility={selectableListItemBehavior} onClick={onClick} />,
+      ).find('ListItem')
+      listItem.simulate('keydown', { keyCode: 32 })
```
layershifter: We can use here `keyboard-key` it will be much obvious

test/specs/components/List/ListItem-test.tsx
```diff
@@ -0,0 +1,31 @@
...
+      expect(onClick).toHaveBeenCalled()
```
layershifter: Can we make this assert more strict?

```suggestion
      expect(onClick).toHaveBeenCalledTimes(1)
```

test/specs/components/List/ListItem-test.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { mountWithProvider } from 'test/utils'
+
+import ListItem from 'src/components/List/ListItem'
+import { selectableListItemBehavior } from 'src/lib/accessibility'
+
+describe('ListItem', () => {
+  isConformant(ListItem)
+  handlesAccessibility(ListItem, { defaultRootRole: 'listitem' })
+
+  describe('selectable list handleClick', () => {
```
layershifter: `describe` should contain something more meaningful.

src/components/List/ListItem.tsx
```diff
@@ -94,6 +103,14 @@ class ListItem extends UIComponent<Extendable<ListItemProps>, ListItemState> {
 
   private itemRef = React.createRef<HTMLElement>()
 
+  protected actionHandlers: AccessibilityActionHandlers = {
+    performClick: event => this.handleClick(event),
```
layershifter: `e` vs `event`: 
```
performClick: event =>
//
handleClick = e => {
```

Can we be more consistent there?

layershifter (approved):
Left some NITs, we need to resolve merge conflicts there and ship it 👍

PR #282, opened 2018-09-26
feat(Popup): refactor and add controlled mode

kuzhelov:
### TODO
- [x] add CHANGELOG entry
- [x] add CHANGELOG entry for the #218 

--------------

The following issues are aimed to be addressed by the proposed PR:
- simplify logic of `Popup` component
  - reduce amount of abstraction layers (primary reason for that is to simplify logic of refs capturing)
  - eliminate unnecessary props
- address subtle problems with logic
  - it turns out that in RTL mode we were 'flipping' CSS position properties three times (instead of just one), and this RTL positioning could be easily broken with custom component/DOM element being provided as value for `as` prop
  - custom `as` components/elements could cause breaks in ref capturing logic
- introduce controlled mode to `Popup`
  - by adding `open` and `defaultOpen` props


docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
```diff
@@ -1,145 +1,67 @@
 import React from 'react'
 import { Button, Grid, Popup } from '@stardust-ui/react'
 
+class PopupArrowExample extends React.Component<any, any> {
```
sophieH29: why the name is ```PopupArrow```? :)

src/components/Popup/Popup.tsx
```diff
@@ -2,73 +2,78 @@ import * as React from 'react'
...
 
-    /** An element type to render as (string or function). */
-    as: customPropTypes.as,
-
-    /** Basic CSS styling for the popup. */
-    basic: PropTypes.bool,
-
     /** The popup content (deprecated). */
-    children: customPropTypes.disallow(['children']),
+    children: PropTypes.any, // customPropTypes.disallow(['children']),
```
sophieH29: is it intended?

src/components/Popup/Popup.tsx
```diff
@@ -80,78 +85,61 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
-  public state = { triggerRef: undefined, popupOpened: false }
+  public static autoControlledProps = ['open']
 
   protected actionHandlers: AccessibilityActionHandlers = {
-    performClick: e => this.onTriggerClick(e),
-    closeAndFocusTrigger: e => this.handlePopupState(e, () => false, this.focusTrigger),
+    toggle: e => _.invoke(this.props, 'onOpenChange', e, !this.props.open),
+    closeAndFocusTrigger: e => {
+      _.invoke(this.props, 'onOpenChange', false)
+      _.invoke(this.state.triggerRef, 'focus')
```
sophieH29: So, in case of user won't provide ```onOpenChange``` function which will handle open/close state, the Popup won't be closed, but the trigger would be focused. It seems like a not expected scenario.
And seems like same for toggle (in case it's not a button).
Let's discuss it.

src/components/Popup/Popup.tsx
kuzhelov: to remove `disallow` statement - yes, this is intended :) but at the same time (other than cleaning up commented code), PropTypes.node would be a better substitute here - let me address that.

src/components/Popup/Popup.tsx
kuzhelov: this is a good catch - lets discuss `closeAndFocus` case. Speaking of the `toggle` one - could you, please, suggest what the problem is exactly about?

src/components/Popup/Popup.tsx
bmdalex: why do we remove it? we support children API now?

src/components/Popup/Popup.tsx
```diff
@@ -80,78 +85,61 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
-          : this.renderContent(
-              computePopupPlacement({ align, position, rtl }),
-              this.renderPopperChildren.bind(this, rtl, accessibility),
-            )}
-      </ElementType>
+      <>
+        <Ref
+          innerRef={domNode => {
+            this.setState({ triggerRef: domNode })
+          }}
```
bmdalex, line 131: we should define a method for this and not define it here every time since it's hurting perf and it's not a good React practice

src/components/Popup/Popup.tsx
kuzhelov: yes - as this is quite reasonable for the Popup case, see:

```jsx
<SomeTree>
  <Popup content={popupContent}>
    <button>Click me to trigger popup</button>
  </Popup>
</SomeTree>
```

Now to introduce popup functionality it is necessary just to wrap the component _that previously was in the tree_ - to augment it with popup functionality (and make it a trigger)

src/components/Popup/Popup.tsx
kuzhelov: here is the excerpt from official docs :) 

https://reactjs.org/docs/faq-functions.html

image

And now it seems that we cannot claim that there are any perf issues because of that. I will change this, but it seems that it is better to obtain measurements first, to avoid premature optimisations with unclear benefit.

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
bmdalex: +1 name should reflect name of file

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
```diff
@@ -1,145 +1,67 @@
...
-      align="bottom"
-      position="after"
-      trigger={triggers[11]}
-      content={
-        <p>
-          The popup is rendered after the trigger<br />aligned to the bottom.
-        </p>
-      }
-    />
+    {triggers.map(({ position, align, icon, padding }) => (
```
bmdalex, line 54: @levithomason thought we want to make these kind of examples explicit; at the same time we want to keep them short; what should be the approach here?

src/components/Popup/Popup.tsx
```diff
@@ -2,73 +2,78 @@ import * as React from 'react'
...
   accessibility?: Accessibility
   align?: Alignment
-  as?: any
-  basic?: boolean
   children?: ReactChildren
   className?: string
   content?: ItemShorthand | ItemShorthand[]
+  defaultOpen?: boolean
+  open?: boolean
+  onOpenChange?: () => void
```
bmdalex: should't we expose the `open` flag here as well; what's the advantage of this on having 2 props as in SUIR:
`onOpen` and `onClose`?

src/components/Popup/Popup.tsx
sophieH29: Speaking of a toggle, if a trigger is not ```button``` element,  you won't be able to toggle popup with ```Enter```\ ```Space``` button. So, do we expect here that a user should use ```onOpenChange`` function in case to handle toggling?

src/components/Popup/Popup.tsx
```diff
@@ -80,78 +85,61 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
+          innerRef={domNode => {
+            this.setState({ triggerRef: domNode })
+          }}
+        >
+          {React.cloneElement(childrenExist(children) ? children : (trigger as any), {
+            ...accessibility.attributes.trigger,
+            ...accessibility.keyHandlers.trigger,
+          })}
+        </Ref>
+        {this.props.open && this.renderPopupContent(rtl, accessibility)}
```
bmdalex: nit: add `open` to destructured object above and use from there

src/components/Popup/Popup.tsx
```diff
@@ -160,41 +148,22 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
-        innerRef={ref}
-        basic={basic}
-        {...rtl && { dir: 'rtl' }}
-        styles={computedStyle}
-        {...accessibility.attributes.popup}
-        {...accessibility.keyHandlers.popup}
-      >
-        {content}
-      </Popup.Content>
+      <Ref innerRef={domElement => ref(domElement)}>
```
bmdalex, line 168: we should define a method for this and not define it here every time since it's hurting perf and it's not a good React practice

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
```diff
@@ -1,145 +1,67 @@
...
+
+    return (
+      <Popup
+        open={this.state.popupOpen}
+        align={align}
+        position={position}
+        trigger={
+          <Button
+            onClick={() => this.togglePopupState()}
+            onMouseLeave={() => this.setState({ popupOpen: false })}
```
bmdalex, line 22: why do we close the popup on mouseLeave? I thought it's a bug. What is the user scenario?
also, same as above, we redefine the function on every `onMouseLeave`, bad for perf and not good react practice

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
```diff
@@ -1,145 +1,67 @@
 import React from 'react'
 import { Button, Grid, Popup } from '@stardust-ui/react'
 
+class PopupArrowExample extends React.Component<any, any> {
+  state = { popupOpen: false }
+
+  togglePopupState = () => {
+    this.setState(prev => ({ popupOpen: !prev.popupOpen }))
```
bmdalex, line 8: why not just
```typescript
this.setState({ popupOpen: !this.state.popupOpen })
```
?

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
```diff
@@ -1,145 +1,67 @@
...
+  }
+
+  render = () => {
+    const { position, align, icon, padding } = this.props
+
+    const buttonStyles = { padding, height: '38px', minWidth: '64px' }
```
bmdalex: nit: u can inline the styles to save space in the example

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
```diff
@@ -1,145 +1,67 @@
...
+
+    return (
+      <Popup
+        open={this.state.popupOpen}
+        align={align}
+        position={position}
+        trigger={
+          <Button
+            onClick={() => this.togglePopupState()}
```
bmdalex, line 21: we redefine the function on every click, bad for perf and not good react practice; since `togglePopupState` is already an arrow function u can do
```typescript
onClick={this. togglePopupState}
```

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
```diff
@@ -1,145 +1,67 @@
 import React from 'react'
 import { Button, Grid, Popup } from '@stardust-ui/react'
 
+class PopupArrowExample extends React.Component<any, any> {
+  state = { popupOpen: false }
+
+  togglePopupState = () => {
```
bmdalex, line 7: you removed the examples that show that the popup state (open/closed reflected by `open` prop) can also be handled by the popup internally without adding an event handler for the trigger; we should keep that example since `Popup` is now an autocontrolled component

bmdalex (changes requested):
pls take a look at comments

src/components/Popup/Popup.tsx
```diff
@@ -160,41 +148,22 @@ export default class Popup extends UIComponent<Extendable<IPopupProps>, IPopupSt
...
+        <Popup.Content
+          {...rtl && { dir: 'rtl' }}
+          style={popupPlacementStyles}
+          {...accessibility.attributes.popup}
+          {...accessibility.keyHandlers.popup}
+          data-placement={placement}
```
bmdalex: why is this needed? accessibility / w3c standards? it's not adding any functionality

kuzhelov: these lines reflect changes that were merged by another PR - please, take a look on the current state of master. So, those are necessary to project this already introduced functionality on top of changes made in this PR

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
kuzhelov: sorry - you mean, to add two examples with different values provided to the `open` prop?

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
kuzhelov: have already adressed that in comment, would leave it as it is as it goes along with official React docs

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
kuzhelov: because `setState` applies changes asynchronously, so the way you are suggesting won't be reliable in a sense that we might read a 'stale' value for `popupOpen` prop. Please, take a look here: https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
kuzhelov: this is fine as those are docs examples, as well as because of what the official React doc is saying for that: 

https://reactjs.org/docs/faq-functions.html

image

------------

This is the logic that is provided on the client side, so it is safe for us to advertise it. A thing that I would like prevent is the following - and this seems to be a fair way to do that:

image

src/components/Popup/Popup.tsx
kuzhelov: yes - but this leads us to more general problem that is rather accessibility related. In this PR I would like to focus on the following things, **with all the previously introduced functionality being untouched**:
- refactor and simplify `Popup` base implementation
- introduce controlled mode to it

So, these are the main reasons I've left `PopupBehavior`'s functionality as it was introduced before. so that If we would like to address/discuss some problems related to this aspect exclusively, we would be able to do it by means of a dedicated PR.

Please, let me know about your thoughts on that :)

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
kuzhelov: I would rather follow the approach you've initially introduced for _these_ examples - as it seems the most reasonable one to me. Otherwise we will end up with extremely verbose code (once again, for this particular case) that will be very hard to grasp by the client. So, this is the reason I've preferred to use `map` and encapsulate not-changing parts

src/components/Popup/Popup.tsx
kuzhelov: have replied on that in the first comment about arrow functions

src/components/Popup/Popup.tsx
mnajdova: +1 for adding handlers onOpen and onClose that would be invoked based on the value of the open. It is much cleaner API for the user.

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
```diff
@@ -1,145 +1,65 @@
...
+  { position: 'below', align: 'end', icon: 'arrow circle down', padding: '18px 5px 5px 42px' },
+  { position: 'before', align: 'top', icon: 'arrow circle left', padding: '5px 42px 18px 5px' },
+  { position: 'before', align: 'center', icon: 'arrow circle left', padding: '5px 42px 5px 5px' },
+  { position: 'before', align: 'bottom', icon: 'arrow circle left', padding: '18px 42px 5px 5px' },
+  { position: 'after', align: 'top', icon: 'arrow circle right', padding: '5px 5px 18px 42px' },
+  { position: 'after', align: 'center', icon: 'arrow circle right', padding: '5px 5px 5px 42px' },
+  { position: 'after', align: 'bottom', icon: 'arrow circle right', padding: '18px 5px 5px 42px' },
+]
 
 const PopupExamplePosition = () => (
```
mnajdova, line 52: This const and the class both have the same name which is casing some compiler errors: 
```
[at-loader] Checking finished with 2 errors
[at-loader] ./docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx:4:7
    TS2300: Duplicate identifier 'PopupExamplePosition'.

[at-loader] ./docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx:52:7
    TS2300: Duplicate identifier 'PopupExamplePosition'.
```
Please rename one of them.

kuzhelov: done 👍

mnajdova:
I think there is some issue with the RTL. Please take a look into the gif: 
ezgif com-gif-maker 1

kuzhelov:
> I think there is some issue with the RTL. Please take a look into the gif:
> ezgif com-gif-maker 1

This is behavior that is not introduced by this PR, but rather the one that we currently have. Thanks @mnajdova for spotting it - let me, please, open an issue: #288

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
bmdalex: that should not happen; clicking on another item should act like document click and close the opened ones; I don't think that we should add custom logic to examples for closing the popups (in order to avoid this case u mentioned) but it should be part of functionality

src/components/Popup/Popup.tsx
bmdalex: any thoughts here?

mnajdova (approved):

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
bmdalex: oh, disregard my comment

I knew about that, but I think it's relevant for complex state updates from different entry points that might be batched together; however, I agree it's better like u suggest; first time I read that part I misunderstood some of it, thx for reminder ;)

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
bmdalex: I mean examples like this one below where we do not have to add custom logic or handlers for handling basic popup functionality; this used to work in SUIR and I believe we should still have it

```
import React from 'react'
import { Button, Popup } from '@stardust-ui/react'

const PopupExample = () => (
  <Popup trigger={<Button icon="expand" />} content="Add users to your feed." />
)

export default PopupExample
```

src/components/Popup/Popup.tsx
bmdalex: I was talking about line 162:
```typescript
data-placement={placement}
```

kuzhelov: oh, got it - let me address..

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
kuzhelov: as long as `Popup` is controlled component, it shouldn't - lets discuss, please

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
kuzhelov: Here is the problem - there is a need to make Popup to be a controlled component. Stardust suggests to achieve this by extending `AutoControlledComponent` type (for now), and this type enforces controlled mode for the component: https://reactjs.org/docs/forms.html#controlled-components

Any attempt to duplicate `open` prop in state of the `Popup` (and, note, that this is essentially the only way to handle `open` state changes by `Popup` itself) will raise an error, and this policy is enforced by `AutoControlledComponent` by intent.

Thus, _for now_ only controlled mode is supported, and this mode will provide all the necessary functionality for the client. Still, we would be able to extend the approach taken and introduce uncontrolled mode to `Popup` in future - specifically, once `AutoControlledComponent` approach will be reconsidered.

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
kuzhelov: please, see my comment where the problem (and reasons for the approach taken) is further explained: https://github.com/stardust-ui/react/pull/282#discussion_r221593766

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
bmdalex: ah, so the uncontrolled mode was taken down because of the new `AutoControlledComponent` approach was not decided; I thought it was an oversight, got it; thx!

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
bmdalex: got it

bmdalex (approved):
approved

PR #287, opened 2018-09-27
fix(radio-group): replaced label with content

bmdalex:
# Radio group

This PR contains:
- replacing `label` prop with `content`
- making `content` accept react nodes as value

### TODO

- [x] Conformance test
- [ ] Stardust base theme
- [ ] Teams Light theme
- [x] Teams Dark theme
- [ ] Teams Contrast theme
- [ ] Confirm RTL usage
- [x] Update the CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix generation of `key` for the `Accordion.Content` @mnajdova ([#305](https://github.com/stardust-ui/react/pull/305))
 - Ensure `Popup` is rendered as direct child of `body` element in the DOM tree @kuzhelov ([#302](https://github.com/stardust-ui/react/pull/302))
 - Fix toggle logic of `Popup` as reaction on key press events @kuzhelov ([#304](https://github.com/stardust-ui/react/pull/304))
+- Fix for `RadioGroup`: made `label` accept react nodes as value and fixed keyboard navigation @Bugaa92 ([#287](https://github.com/stardust-ui/react/pull/287))
 
 ### Features
 - Add focus styles for `Menu.Item` component @Bugaa92 ([#286](https://github.com/stardust-ui/react/pull/286))
--- a/docs/src/examples/components/RadioGroup/Types/RadioGroupExample.shorthand.tsx
+++ b/docs/src/examples/components/RadioGroup/Types/RadioGroupExample.shorthand.tsx
@@ -1,26 +1,9 @@
 import React from 'react'
-import { Divider, RadioGroup } from '@stardust-ui/react'
-const props = [
-  { name: 'pizza', key: 'Capricciosa', label: 'Capricciosa', value: 'capricciosa' },
-  { name: 'pizza', key: 'Prosciutto', label: 'Prosciutto', value: 'prosciutto', disabled: true },
-  {
-    name: 'pizza',
-    key: 'Margherita',
-    label: 'Margherita',
-    value: 'margherita',
-  },
-]
-const items = [
-  <RadioGroup.Item {...props[0]} />,
-  <RadioGroup.Item {...props[1]} />,
-  <RadioGroup.Item {...props[2]} />,
-]
+import { Divider, RadioGroup, Input, Text } from '@stardust-ui/react'
 
-class RadioGroupExample extends React.Component {
-  state = { selectedValue: '' }
-  handleChange = (e, props) => {
-    this.setState({ selectedValue: props.value })
-  }
+class RadioGroupVerticalExample extends React.Component {
+  state = { selectedValue: '', inputTabIndex: '-1' }
+
   render() {
     const { selectedValue } = this.state
     return (
@@ -29,11 +12,42 @@ class RadioGroupExample extends React.Component {
         <Divider />
         <RadioGroup
           defaultCheckedValue="capricciosa"
-          items={items}
+          items={this.getItems()}
           checkedValueChanged={this.handleChange}
         />
       </div>
     )
   }
+
+  getItems() {
+    return [
+      { name: 'pizza', key: 'Capricciosa', label: 'Capricciosa', value: 'capricciosa' },
+      {
+        name: 'pizza',
+        key: 'Prosciutto',
+        label: 'Prosciutto',
+        value: 'prosciutto',
+        disabled: true,
+      },
+      {
+        name: 'pizza',
+        key: 'Custom',
+        label: (
+          <Text>
+            Choose your own <Input inline placeholder="flavour" />
+          </Text>
+        ),
+        value: 'custom',
+        checkedChanged: this.handleCustomCheckedChange,
+        'aria-label': 'Press Tab to change flavour',
+      },
+    ]
+  }
+
+  handleChange = (e, props) => this.setState({ selectedValue: props.value })
+
+  handleCustomCheckedChange = (e, props) =>
+    this.setState({ inputTabIndex: props.checked ? '0' : '-1' })
 }
-export default RadioGroupExample
+
+export default RadioGroupVerticalExample
--- a/src/components/RadioGroup/RadioGroupItem.tsx
+++ b/src/components/RadioGroup/RadioGroupItem.tsx
@@ -24,11 +24,12 @@ export interface IRadioGroupItemProps {
   checkedChanged?: ComponentEventHandler<IRadioGroupItemProps>
   children?: ReactChildren
   className?: string
+  label?: React.ReactNode
   defaultChecked?: boolean
   disabled?: boolean
   icon?: ItemShorthand
-  label?: string
   name?: string
+  shouldFocus?: boolean // TODO: RFC #306
   styles?: ComponentPartStyle
   value?: string | number
   variables?: ComponentVariablesInput
@@ -49,6 +50,8 @@ class RadioGroupItem extends AutoControlledComponent<
   Extendable<IRadioGroupItemProps>,
   IRadioGroupItemState
 > {
+  private elementRef: HTMLElement
+
   static create: Function
 
   static displayName = 'RadioGroupItem'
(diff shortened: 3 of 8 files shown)


src/components/RadioGroup/RadioGroupItem.tsx
```diff
@@ -62,6 +62,9 @@ class RadioGroupItem extends AutoControlledComponent<Extendable<IRadioGroupItemP
     /** Additional CSS class name(s) to apply.  */
     className: PropTypes.string,
 
+    /** Shorthand for content of the radio item. */
+    content: customPropTypes.contentShorthand,
```
kuzhelov: `shorthand` test should be added for this prop (`implementsShorthandProp`) then

kuzhelov: but, in fact, this is not a shorthand.. seems that propType name (`contentShorthand`) is misleading - please, ignore the comment about tests necessary

src/components/RadioGroup/RadioGroupItem.tsx
```diff
@@ -172,7 +172,7 @@ class RadioGroupItem extends AutoControlledComponent<Extendable<IRadioGroupItemP
         onClick={this.handleClick}
         className={classes.root}
       >
-        <Label styles={styles.label}>
+        <Label styles={styles.content}>
```
kuzhelov: a bit worrying about semantics behind this nesting here. Do we really want the content (that is quite generic by its nature) to be always a children of `Label` (and, thus, to always reflect its styles and, what is more important, behavioural and accessibility aspects - for _any arbitrary value of `content`_ that will be passed in)?

bmdalex: I'm not changing the implementation as you can see, just the name of the prop (I should have kept the name of the styles slot to `label`, will revert that part).

I think this is a legitimate concern, but the nature of the radio button dictates to us that we need an `icon` with some `content` after it; maybe we should not use `Label` to wrap these, any suggestions?

@kuzhelov @mnajdova

src/components/RadioGroup/RadioGroupItem.tsx
bmdalex: any changes you'd suggest?

kuzhelov: no-no, nothing - let me close this one, sorry for misleading you

src/components/RadioGroup/RadioGroupItem.tsx
kuzhelov: yes, I would suggest exactly the same path as you've done - specifically, to not wrap Icon with Label (especially if we would take into account where Icon is used - it is used to 'fill' the circle area of radio input). Then it would be much easier to decouple all these parts from each other.

Not sure if we should do it in this PR or create a dedicated one as a close follow-up, but the most important thing would be, actually, if we all agree on that proposal :)

bmdalex: I'd rather do it in another PR; I'll discuss with @jurokapsiar as well to understand what are the concerns (if any) and proceed with removing Label

src/components/Input/Input.tsx
```diff
@@ -180,7 +180,7 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
 
     return (
       <ElementType className={classes.root} {...restProps} onChange={onChange}>
-        {createHTMLInput(type, {
+        {createHTMLInput(input || type, {
           defaultProps: htmlInputProps,
```
mnajdova: Did we fix here the input prop not to be propagated as a prop to the ElementType?

src/components/RadioGroup/RadioGroupItem.tsx
```diff
@@ -24,11 +23,12 @@ export interface IRadioGroupItemProps {
   checkedChanged?: ComponentEventHandler<IRadioGroupItemProps>
   children?: ReactChildren
   className?: string
+  content?: React.ReactNode
   defaultChecked?: boolean
   disabled?: boolean
   icon?: ItemShorthand
-  label?: string
   name?: string
+  shouldFocus?: boolean
```
kuzhelov: not sure whether we would like to expose it as a prop - most probably this aspect will be handled by accessibility behavior. Could we omit it for this PR, or there is specific urgent need for this functionality to be introduced?

src/components/Input/Input.tsx
```diff
@@ -180,7 +180,7 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
 
     return (
       <ElementType className={classes.root} {...restProps} onChange={onChange}>
-        {createHTMLInput(type, {
+        {createHTMLInput(input || type, {
```
kuzhelov: have we agreed to have this `input` prop for the `Input` element (like as it is in SUIR)? We've had some concerns - specifically, the fact that it makes it non-intuitive for the client what `input` prop of `Input` component is about. Maybe we should leave it for a dedicated PR?

kuzhelov (commented):
generally agree with the changes made, but it seems that there is more changes than necessary to address the goal of this PR. Would rather suggest to do it step-by-step - or to discuss those additional concepts introduced, as those might raise some concerns (have expressed them in comments)

src/components/Input/Input.tsx
bmdalex: then we won't be able to achieve functionality of `RadioGroup` items that contain inputs.. Marija told we we're expected to merge a dedicated PR as well; I'm ok with removing it from here but the examples with `input` in this PR won't work correctly

@kuzhelov @mnajdova

src/components/RadioGroup/RadioGroupItem.tsx
bmdalex: this is solving the bug I was telling you about.

#### Repro:
- create a `RadioGroup` with a `RadioGroupItem` with input (as in the examples of this PR);
- click on input;

#### Expected:
- item is checked and focus stays on input (see `Teams` `Add people` dialog in a group chat)

#### Actual:
- the radio item steals the focus;

This is the reason we're using this `shouldFocus` prop in order to set focus only when it's coming from keyboard; 

Unfortunately accessibility functionality spilled in our component logic and this became necessary now to do it like this. Alternative (and probably better approach) would be to rethink the whole accessibility part

What do you think?
@kuzhelov @mnajdova @jurokapsiar

bmdalex: fyi @smykhailov

bmdalex: adding this prop is critical to enable keyboard navigation for this component ASAP; I created an issue #306  for implementing it properly once @smykhailov 's work in #256 is merged

fyi @mnajdova @kuzhelov @jurokapsiar 

It'd be great to approve this ASAP so we can unblock customers

docs/src/examples/components/RadioGroup/Types/RadioGroupExample.shorthand.tsx
```diff
@@ -29,11 +12,43 @@ class RadioGroupExample extends React.Component {
...
+        value: 'prosciutto',
+        disabled: true,
+      },
+      {
+        name: 'pizza',
+        key: 'Custom',
+        label: (
+          <Text>
+            Choose your own{' '}
+            <Input input={{ tabIndex: this.state.inputTabIndex }} inline placeholder="flavour" />
```
mnajdova: I am bit confuse here... Can we still use the input shorthand, now that we don't have it in the Input component?

mnajdova: Watch out, this definition that you have is different then having 
```
label: {
  content: (
    <Text>
      Choose your own{' '}
      <Input input={{ tabIndex: this.state.inputTabIndex }} inline placeholder="flavour" />
     </Text>
  )
}
``` 
With your definition you are just having this component rendered, so any default or overridden props that you have for the label component will not be applied.

src/components/RadioGroup/RadioGroupItem.tsx
mnajdova: I agree with @Bugaa92 on this. We can come back and refactor this, otherwise this basic functionality will be blocked for at least a week from now unfortunately.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Fixes
 - Fix Attachment `styles` prop typing @levithomason ([#299](https://github.com/stardust-ui/react/pull/299))
 - Fix generation of `key` for the `Accordion.Content` @mnajdova ([#305](https://github.com/stardust-ui/react/pull/305))
+- Fix for `RadioGroup`: replaced `label` with `content` and made it accept react nodes as value @Bugaa92 ([#287](https://github.com/stardust-ui/react/pull/287))
```
mnajdova: Please update this, as we are not really replacing the label after all :)

docs/src/examples/components/RadioGroup/Types/RadioGroupExample.shorthand.tsx
bmdalex: > I am bit confuse here... Can we still use the input shorthand, now that we don't have it in the Input component?

no, this will work correctly once we fix the input shorthand, but it does no harm now, just key nav is not working correctly for the case when we tab from an item to the next component

> With your definition you are just having this component rendered, so any default or overridden props that you have for the label component will not be applied.

I don't get it, that's what I want.. the label slot is already inside `Label` component

mnajdova: > I don't get it, that's what I want.. the label slot is already inside `Label` component

Sorry, my mistake, I thought this contain `Label.create(label ...` It is defined as it should.

mnajdova: Let's get rid of the input shorthand, and we will add it in the Input's PR

kuzhelov (approved):

CHANGELOG.md
bmdalex: done

src/components/Input/Input.tsx
bmdalex: we will do it in another PR

docs/src/examples/components/RadioGroup/Types/RadioGroupExample.shorthand.tsx
bmdalex: done

PR #292, opened 2018-10-01
fix: make theme variables and styles extensible

levithomason:
Currently, consumers cannot add custom component display names to a theme's `componentVariables` and `componentStyles`.  This is a problem as they will need to add their component's to the theme as well.

I've also added the unlisted component display names the theme.

### DIFF
--- a/types/theme.d.ts
+++ b/types/theme.d.ts
@@ -60,6 +60,9 @@ export interface ICSSPseudoElementStyle extends ICSSInJSStyle {
 }
 
 export interface ICSSInJSStyle extends React.CSSProperties {
+  // TODO Questionable: how else would users target their own children?
+  [key: string]: any
+
   // missing React.CSSProperties
   speak?: CSSType.Globals | 'none' | 'normal' | 'spell-out'
 
@@ -163,7 +166,7 @@ export interface IThemePrepared {
 }
 
 export interface IThemeComponentStylesInput {
-  [key: string]: any
+  [key: string]: IComponentPartStylesInput
 
   Accordion?: IComponentPartStylesInput
   Attachment?: IComponentPartStylesInput
@@ -198,7 +201,7 @@ export interface IThemeComponentStylesInput {
 }
 
 export interface IThemeComponentStylesPrepared {
-  [key: string]: any
+  [key: string]: IComponentPartStylesPrepared
 
   Accordion?: IComponentPartStylesPrepared
   Attachment?: IComponentPartStylesPrepared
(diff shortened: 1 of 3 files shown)


types/theme.d.ts
```diff
@@ -60,6 +60,9 @@ export interface ICSSPseudoElementStyle extends ICSSInJSStyle {
 }
 
 export interface ICSSInJSStyle extends React.CSSProperties {
+  // TODO Questionable: how else would users target their own children?
+  [key: string]: any
```
mnajdova, line 64: Is it wrong to make this `[key: string]: CssInJssStyle`? I am a bit worried with the any there..

types/theme.d.ts
```diff
@@ -163,72 +166,142 @@ export interface IThemePrepared {
 }
 
 export interface IThemeComponentStylesInput {
+  [key: string]: any
+
```
mnajdova, line 170: As in the previous comment, can we make this: `[key: string]: IComponentPartStylesInput`? What else can be the value of this?

types/theme.d.ts
levithomason: Ah, yep. Good call.

levithomason: This is actually quite a bit trickier.  Not every string property is an ICSSInJSStyle.  Some have limited values, such as `speak`, so they cannot be typed as ICSSInJSStyle.

Since this is currently blocking and broken, let's continue with a more permissive object (as is) for now.  We can always improve typings.

types/theme.d.ts
levithomason: Yes, here we can since all keys have the same value type.  Same with IThemeComponentStylesPrepared below which I have updated as well.

types/theme.d.ts
mnajdova: Got it! Approved.

mnajdova (approved):

types/theme.d.ts
```diff
@@ -163,72 +166,142 @@ export interface IThemePrepared {
...
+  PopupContent?: IComponentPartStylesPrepared
+  RadioGroup?: IComponentPartStylesPrepared
+  RadioGroupItem?: IComponentPartStylesPrepared
+  Segment?: IComponentPartStylesPrepared
+  Status?: IComponentPartStylesPrepared
   Text?: IComponentPartStylesPrepared
 }
 
 export interface IThemeComponentVariablesInput {
+  [key: string]: any
```
mnajdova, line 239: I guesss the same can be applied here as well as in the `IThemeComponentVariablesPrepared`? `[key: string]: ComponentVariablesInput` & `[key: string]:  ComponentVariablesPrepared`? Otherwise we have some typings errors already..

kuzhelov (approved):

Issue #294, opened 2018-10-02
Popup appearance in the DOM tree


kuzhelov:
## Bug Report
The issue is related to where `Popup`'s rendered HTML is placed in the DOM tree.

### Steps
- open Stardust docs, navigate to Popup page: https://stardust-ui.github.io/react/components/popup
- toggle popup for the first example
- observe where Popup's HTML is rendered in the DOM tree

### Expected Result
It should be rendered as body's last child. This requirement is necessary for Stardust to properly handle accessibility aspects.

![image](https://user-images.githubusercontent.com/9024564/46355864-4ddd9380-c66a-11e8-8d0f-b06ef276e79f.png)

### Actual Result
It is rendered next to trigger element (`<button />`)

![image](https://user-images.githubusercontent.com/9024564/46357101-c34a6380-c66c-11e8-8c1d-8d1a90425ac7.png)

### Version
0.8.0

kuzhelov:
this is addressed by #302

PR #302, opened 2018-10-02
fix(Popup): render element as body children

kuzhelov:
This PR fixes issue #294 by ensuring that `Popup` component's DOM tree is rendered as `body`'s child element.


src/components/Popup/Popup.tsx
```diff
@@ -135,7 +138,9 @@ export default class Popup extends AutoControlledComponent<Extendable<IPopupProp
             ...accessibility.keyHandlers.trigger,
           })}
         </Ref>
-        {open && this.renderPopupContent(rtl, accessibility)}
+        {open &&
+          Popup.isBrowserContext &&
+          createPortal(this.renderPopupContent(rtl, accessibility), document.body)}
```
miroslavstastny, line 146: This is probably more related to #282 but what's the reason to use React Portal directly instead of using stardust Portal component?
The main benefit of using stardust Portal component would be a reuse of `onOutsideClick` (close popup when user clicks outside the popup/trigger) and similar.

sophieH29: +1, also stardust Portal component is going to have Focus Trap Zone, which can be used in Popup too, by just passing prop to enable it.
Otherwise, we would need to duplicate logic with focus trap in Popup component

kuzhelov: would rather like to address it by means of separate PR. The reason for that is, primarily, the following: apart from introducing this common functionality, the `Portal` component introduces quite a few additional conventions, such as _render as direct body's children_, etc - and there are not intuitive for the client given just the `Portal`'s name and API - probably, it should be renamed first.

So, my intuition is to introduce the functionality that is necessary as a first thing, and after that generalize by making refactoring of `Portal` component - in a way of making this component itself more generic and properly translate its functionality and application strategies via its name and API (`BodyPortal` is something that comes to my head right now, but again, would rather like to see a clear picture first). More than that, now it seems that only these two aspects are addressed by Portal - and thus it is quite easy to introduce a dedicated function instead of component for that. One of the pros of this approach is that we will save one component from the Popup's rendered tree.

Please, let me know about your thoughts on that.

mnajdova: If these changes are unblocking some other pull request, I am okay with them being merged and then the Portal can be replace with Stardust's Portal. In that way we can make progress on both aspects.

mnajdova (approved):

src/components/Popup/Popup.tsx
kuzhelov: have discussed with @sophieH29 - essentially, have agreed on the same plan. Lets wait on @miroslavstastny's comments and merge if there won't be anything critical.

miroslavstastny: Go for now.

miroslavstastny (approved):

jurokapsiar (approved):

Issue #295, opened 2018-10-02
Popup is not opening on trigger's 'enter' key press


kuzhelov:
## Bug Report

Popup is not displayed when `Enter` is pressed while trigger element is focused.

### Steps
- open Stardust Docs page, navigate to Popup: https://stardust-ui.github.io/react/components/popup
- for the first (default) example - focus on button and press `Enter`

### Expected Result
- alert message is displayed
- popup is opened

### Actual Result
- alert is displayed
- popup IS NOT opened

### Version
0.8.0

jurokapsiar:
in my case, when I commented the alert() in the example, even space started to behave in a strange way - it opened the popup and closed it immediately. please check this as well.

kuzhelov:
closing as this one is fixed by #304

PR #304, opened 2018-10-03
fix(Popup): keypress toggle logic

kuzhelov:
## TODO 
- [x] add entry to changelog

--------

This PR is aimed to address #295. 

The main reason for improper response of Popup's open state on key press events was the fact that for trigger component event of click was handled two times when `click` was initiated by key press:
- once in `onClick()` handler (we need this one for controlled Popup)
- another in `PopupBehavior` handler

Both these handlers were toggling `open` state of the Popup - and, as a result, there were no visible changes to the `open` state of the Popup, as two toggles resulted in `open`'s initial value.

The core of the problem was the fact that by introducing key handlers `PopupBehavior` had overlapped the functionality that is already covered by browsers (specifically, specific keys handling logic). To prevent this, trigger key actions were removed from `PopupBehavior` - as those are not necessary: this `PopupBehavior` is designed to specifically target the case where button serves as a trigger (something like rather `ButtonPopupBehavior`) - and in that case there is no need to augment set of key handlers with `space` and `enter` keys.


jurokapsiar (approved):

mnajdova (approved):

docs/src/examples/components/Popup/Types/PopupExample.tsx
```diff
@@ -8,17 +8,15 @@ class PopupExample extends React.Component<any, any> {
...
   render() {
     return (
       <Popup
         open={this.state.popupOpen}
-        onOpenChange={this.handleOpenChange.bind(this)}
-        content="Add users to your feed."
+        onOpenChange={(e, newProps) => {
+          alert(`Popup is requested to change its open state to "${newProps.open}".`)
+          this.setState({ popupOpen: newProps.open })
+        }}
```
mnajdova, line 18: +1 for simplicity :)

docs/src/examples/components/Popup/Types/index.tsx
```diff
@@ -7,7 +7,7 @@ const Types = () => (
   <ExampleSection title="Types">
     <ComponentExample
       title="Default"
-      description="A default popup. Note that Popup is a controlled component, and its 'open' prop value could be changed either by parent component, or by user actions (keypress) - thus it is necessary to handle 'onOpenChanged' event. Try to press space key while button is focused to see the effect."
+      description="A default popup. Note that Popup is a controlled component, and its 'open' prop value could be changed either by parent component, or by user actions (e.g. key press) - thus it is necessary to handle 'onOpenChanged' event. Try to type some text into popup's input field and press ESC to see the effect."
```
sophieH29, line 10: like it!

PR #305, opened 2018-10-03
fix(Accordion): generate key for the Accordion.Content

### LINKED ISSUE #291: Missing key in AccordionContent
## Bug Report
Creating Accordion using panels props causes missing key
### Steps
1. Create Accordion using the panels props. Example (from the docs):
```const panels = [
  {
    key: "p",
    title: "What is a point?",
    content:
      "Use the haptic SDD circuit, then you can index the redundant pixel!"
  },
  {
    key: "d",
    title: "What is a dimension of a point?",
    content: "We need to copy the primary USB firewall!"
  }
];
```
### Expected Result
Both AccordionTitle and AccordionContent have key prop related to the specified key.

### Actual Result
AccordionTitle has key related to its content (for a string it duplicates the string value). AccordionContent has no key.

image

### Version
0.5.0-0.7.0

mnajdova:
# Fix

This PR fixes https://github.com/stardust-ui/react/issues/291 

### TODO

- [x] Conformance test
- [x] Update the CHANGELOG.md

There was a bug in the creation of the Content which prevented generation of keys, and a small bug where the complex content for the Accordion.Content was created. Those two are fixed now. 
## Important note
We support generation of keys, for the component that as content has some primitive value (string, number etc). If the content is something more complex (as in the last example of the docs), that the user is responsible for creating the key.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixes
 - Fix Attachment `styles` prop typing @levithomason ([#299](https://github.com/stardust-ui/react/pull/299))
-- Fix generation of key for the Accordion.Content @mnajdova ([#305](https://github.com/stardust-ui/react/pull/305))
+- Fix generation of `key` for the `Accordion.Content` @mnajdova ([#305](https://github.com/stardust-ui/react/pull/305))
 
 ### Features
 - Add focus styles for `Menu.Item` component @Bugaa92 ([#286](https://github.com/stardust-ui/react/pull/286))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixes
 - Fix Attachment `styles` prop typing @levithomason ([#299](https://github.com/stardust-ui/react/pull/299))
+- Fix generation of key for the Accordion.Content @mnajdova ([#305](https://github.com/stardust-ui/react/pull/305))
 
 ### Features
 - Add focus styles for `Menu.Item` component @Bugaa92 ([#286](https://github.com/stardust-ui/react/pull/286))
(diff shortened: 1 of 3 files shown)


layershifter:
@mnajdova may be the better idea will be to pick the `AccordionPanel` component for SUIR? PR #3032 contains the updated component that uses `Fragment` API ([AccordionPanel.js](https://github.com/Semantic-Org/Semantic-UI-React/blob/ceecf00ba1c042a3cecf2134262af9971a79f820/src/modules/Accordion/AccordionPanel.js)).

This allow handle keys on panels:
```jsx
const panels = [
 {
    key: 'what-is-dog',
    title: 'What is a dog?',
    content: 'A dog is a type of domesticated animal.',
  },
  {
    key: 'kinds-of-dogs',
    title: 'What kinds of dogs are there?',
    content: 'There are many breeds of dogs.',
  },
]
```

mnajdova:
@layershifter thanks for the reference on the AccordionPanel. I would not make that big changes at this moment, as the only issue we hit so far was the generation of the content's key which if fix with this PR. We can however consider refactoring this in the future, but I see having more things that should be considered, as supporting the `as` property, which would adding additional wrapper that the user doesn't really need. Will keep in mind this, if this component shows some other issues in the future.

alinais:
@mnajdova @layershifter please create an RFC for further discussion around Accordion panels.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixes
 - Fix Attachment `styles` prop typing @levithomason ([#299](https://github.com/stardust-ui/react/pull/299))
+- Fix generation of key for the Accordion.Content @mnajdova ([#305](https://github.com/stardust-ui/react/pull/305))
```
alinais: `Accordion.Content` ;)

alinais: meaning, that you can add the `

alinais (approved):

CHANGELOG.md
mnajdova: Thanks, pushed :)

miroslavstastny (approved):

PR #309, opened 2018-10-04
fix(docs): fix className to receive string

bmdalex:
# Docsite

### Description

This PR contains a one line fix for the doc site. Whenever we open the code editor there is a console error coming from the API code menu (`Shorthand API | Children API`) items.

We are sending  boolean values instead of string values to `className` prop for menu items, hence the error:
docsite_error

### Fix:

Always send string as `className` prop value.


bmdalex:
quick one, fyi @kuzhelov @mnajdova @miroslavstastny

miroslavstastny (approved):

PR #310, opened 2018-10-04
fix(Input): remove duplcated 'onChange' handling

kuzhelov:
Currently for each change of the input value we have `onChange` handler logic being called twice. This is happening due to the same event handling function is provided both for `<input>` DOM element, as well as for the wrapping component (`default is <div>`):

```jsx
renderComponent({ ElementType, classes, styles, variables }) {
   ...

    return (
      <ElementType className={classes.root} {...restProps} onChange={onChange}>  // <----- HERE
        {createHTMLInput(type, {
          defaultProps: htmlInputProps, // <------ AND HERE, as htmlInputProps contain 'onChange' one
          overrideProps: {
            className: inputClasses,
            ref: this.handleInputRef,
          },
        })}
```

image

### DIFF
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -174,12 +174,10 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
     const { type } = this.props
     const [htmlInputProps, restProps] = this.partitionProps()
 
-    const { onChange } = htmlInputProps as any
-
     const inputClasses = classes.input
 
     return (
-      <ElementType className={classes.root} {...restProps} onChange={onChange}>
+      <ElementType className={classes.root} {...restProps}>
         {createHTMLInput(type, {
           defaultProps: htmlInputProps,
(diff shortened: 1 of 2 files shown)


bmdalex (approved):
looks good @mnajdova @alinais can one of you approve too?

src/components/Input/Input.tsx
```diff
@@ -174,12 +174,10 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
     const { type } = this.props
     const [htmlInputProps, restProps] = this.partitionProps()
 
-    const { onChange } = htmlInputProps as any
-
     const inputClasses = classes.input
 
     return (
-      <ElementType className={classes.root} {...restProps} onChange={onChange}>
+      <ElementType className={classes.root} {...restProps}>
```
mnajdova, line 180: +1

mnajdova (approved):

PR #311, opened 2018-10-04
fix(Popup): rename 'onOpenChange' -> 'onChange'

kuzhelov:
### TODO 
- [ ] update CHANGELOG

-----------

This BREAKING change is necessary to ensure the naming is consistent with other components (such as Input). Although it is a breaking one, it is better to introduce it sooner than later.

### DIFF
--- a/docs/src/examples/components/Popup/Types/PopupExample.shorthand.tsx
+++ b/docs/src/examples/components/Popup/Types/PopupExample.shorthand.tsx
@@ -12,7 +12,7 @@ class PopupExample extends React.Component<any, any> {
     return (
       <Popup
         open={this.state.popupOpen}
-        onOpenChange={(e, newProps) => {
+        onChange={(e, newProps) => {
           alert(`Popup is requested to change its open state to "${newProps.open}".`)
           this.setState({ popupOpen: newProps.open })
         }}
--- a/docs/src/examples/components/Popup/Types/PopupExample.tsx
+++ b/docs/src/examples/components/Popup/Types/PopupExample.tsx
@@ -12,7 +12,7 @@ class PopupExample extends React.Component<any, any> {
     return (
       <Popup
         open={this.state.popupOpen}
-        onOpenChange={(e, newProps) => {
+        onChange={(e, newProps) => {
           alert(`Popup is requested to change its open state to "${newProps.open}".`)
           this.setState({ popupOpen: newProps.open })
         }}
--- a/docs/src/examples/components/Popup/Types/index.tsx
+++ b/docs/src/examples/components/Popup/Types/index.tsx
@@ -7,7 +7,7 @@ const Types = () => (
   <ExampleSection title="Types">
     <ComponentExample
       title="Default"
-      description="A default popup. Note that Popup is a controlled component, and its 'open' prop value could be changed either by parent component, or by user actions (e.g. key press) - thus it is necessary to handle 'onOpenChanged' event. Try to type some text into popup's input field and press ESC to see the effect."
+      description="A default popup. Note that Popup is a controlled component, and its 'open' prop value could be changed either by parent component, or by user actions (e.g. key press) - thus it is necessary to handle 'onChange' event. Try to type some text into popup's input field and press ESC to see the effect."
       examplePath="components/Popup/Types/PopupExample"
     />
   </ExampleSection>
--- a/src/components/Popup/Popup.tsx
+++ b/src/components/Popup/Popup.tsx
@@ -35,7 +35,7 @@ export interface IPopupProps {
   content?: ItemShorthand | ItemShorthand[]
   defaultOpen?: boolean
   open?: boolean
-  onOpenChange?: ComponentEventHandler<IPopupProps>
+  onChange?: ComponentEventHandler<IPopupProps>
   position?: Position
   trigger?: JSX.Element
 }
@@ -86,7 +86,7 @@ export default class Popup extends AutoControlledComponent<Extendable<IPopupProp
      * @param {SyntheticEvent} event - React's original SyntheticEvent.
      * @param {object} data - All props and proposed value.
      */
-    onOpenChange: PropTypes.func,
+    onChange: PropTypes.func,
 
     /**
      * Position for the popup. Position has higher priority than align. If position is vertical ('above' | 'below')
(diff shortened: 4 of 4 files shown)


levithomason (changes requested):
I feel strongly that this should be `onOpen` and `onClose`.  The guide for our API decisions is natural language, how people talk about UI to each other.  It seems reasonable that most people say "open the popup" or "close the popup" opposed to "change the popup [open state]".

kuzhelov:
@levithomason, this should be a pair of additional props added **in addition** to that - and I agree with your suggestions. At the same time this requirement shouldn't block this PR to be merged, as this one introduces method that is consistent with all other controlled mode implementations.

>  It seems reasonable that most people say "open the popup" or "close the popup" opposed to "change the popup [open state]".

For the first part of the sentence - lets consider controlled mode of the Popup, see:

```jsx
<Popup open={this.state.open}>
  <Button onClick={() => { this.state.open = !this.state.open }} />   // sloppy, but just to indicate the idea
</Popup>
```

Here is the reason I would disagree - **in the controlled mode** for me, as a client, it is much more reasonable to _declaratively_ set its `open` state.

-----
Further progressing with that. For the sake of handling some accessibility aspects, on the acc behavior's side we need to **_ask client for changing `Popup` state as a reaction on some key press events, such as ESC_** - for these exact cases we are exposing `onChange` prop to the client, as a way to observe that Popup state had been requested to change. This is done via `onChange`, the same way it is done for standard React controlled components:

```jsx
<Popup open={this.state.open} onChange={(_, newState) => { this.state.open = newState.open}}>
  <Button onClick={() => { this.state.open = !this.state.open }} />   // sloppy, but just to indicate the idea
</Popup>
```

Please, consider how this API mirrors the API exposed by `input` (`select`, `textarea`) component, making our Popup controlled component consistent:

```jsx
<input 
  value={`${this.state.popupOpen}`} 
  onChange={(e) => console.log('ASK client to change input value to', e.target.value)}/>

<Popup
  open={this.state.popupOpen}
  onChange={(e, newProps) => console.log('ASK client to change open state to ', newProps.open)}
    content={...}
  >
    <Button icon="expand" onClick={() => this.togglePopup()} />
</Popup>

```

In contrast, with this being replaced by `onOpen` and `onClose`:

```jsx
<Popup 
  open={this.state.open} 
  onOpen={( ? what here? ) => { this.state.open = true }}  // should we set the state here?
  onClose={() =>  { this.state.open = false }}    // name suggests that  no - we need just to REACT on state change
>
  <Button onClick={() => { this.state.open = !this.state.open }} />   // sloppy, but just to indicate the idea
</Popup>
```

And here we will have to introduce two subscriptions, instead of just one - without any good reason for that, as well as with no reason to increment the entropy of prop names used in Stardust.

Also, we are introducing the question about semantics of these methods - while semantics of `onChange` is to rather **ask** client to handle state changes, `onOpen` and `onClose`, in contrast, **notify** client about change being made. For this reason see the concepts being different.

### Conclusion
So, once again, see these two methods proposed as an addition to the `onChange`, and, thus, it shouldn't block current set of changes.

mnajdova:
I am also for introducing the `onOpen` and `onClose` props, but as an addition to the `onChange`. And I support the renaming of the `onOpenChange` to `onChange` (this is something usual for the controlled components). So I don't see at this moment point for blocking the renaming of this prop, and we will introduce the additional properties in a dedicated PR.

mnajdova (approved):

docs/src/examples/components/Popup/Types/PopupExample.shorthand.tsx
```diff
@@ -12,7 +12,7 @@ class PopupExample extends React.Component<any, any> {
     return (
       <Popup
         open={this.state.popupOpen}
-        onOpenChange={(e, newProps) => {
+        onChange={(e, newProps) => {
           alert(`Popup is requested to change its open state to "${newProps.open}".`)
```
mnajdova, line 16: Exactly in this case, I would feel awkward to introduce onOpen and onClose for setting the value of the state to true and false, rather then having the onChange for handling both cases.

mnajdova (approved):

levithomason (changes requested):
For clarity, we discussed this PR offline.  I do not want to render the trigger as children of the Popup.

All Popup packages and libraries we can find online render the children of the Popup as the content of the Popup.  I think the user expects the children to always be the main content of the component as well.

We are taking some time to consider the merging of `onOpen` and `onClose` into `onChange`.  I'm currently opposed to this for the following:
- Popups don't "change", they "open" and "close".  Inputs do "change" so it makes sense.
- The "onChange" React prop and HTML attribute are used exclusively for value changes, not other props
- APIs should follow natural language, non-engineers wouldn't say "change the popup" but "open it" or "close it"
- There are precedent concerns in that all future components that have binary operation changes should use a single change callback.  Modals, Confirms, Transitions, Dropdowns, ...
- The API isn't consistent with Popups in the community

None the less consideration will be taken and my biases examined.  At present, however, I do not want to merge this.

kuzhelov:
**TL;DR** In general to me it seems a bit strange that we are spending so much time on this refactoring change commit, especially given that it **doesn't make things any worse** - while striving to be consistent with other controlled components examples. More than that, this is a thing that is so easy to change in future. 

At the same time, let me, please, respond to each of the points raised - by focusing on the moments that are introduced by this PR (and, thus, omitting the question of 'trigger as children')

-------

> I'm currently opposed to this for the following:
> - Popups don't "change", they "open" and "close".  Inputs do "change" so it makes sense.

Please, specify exact meaning of `change` then, because if we will follow the semantics of `change` stays for **state changes**, then Popup adheres to this concept well

> - The "onChange" React prop and HTML attribute are used exclusively for value changes, not other props

Those are used to **let user handle changes advertised by the component** (for `<input>`, as an example, it is about <input> asking the client code to change the `value`'s value in the controlled mode)

> - APIs should follow natural language, non-engineers wouldn't say "change the popup" but "open it" or "close it"

Ok, but for the alternative case of using `onOpen` and `onClose` events - is the answer on question "what these two stay for?" immediately clear for the user. Specifically, do they have either semantics of:
-  should the logic take place **when `Popup` has just become to be opened/closed**, so that this logic will be executed right after (like, for example, for button's `onClick`, `onMouseEnter`)?
- **should the logic itself make `Popup` to be opened/closed?** Note that it contradicts semantics of regular DOM elements event handlers

Alternatively, when `onChange` would be used for controlled mode, it would follow the same patterns that client of the code is used to - specifically, the patterns of regular controlled components. Thus, I would argue that this option is rather more natural for the client, rather than event handlers that have names with confusing semantics.


> - There are precedent concerns in that all future components that have binary operation changes should use a single change callback.  Modals, Confirms, Transitions, Dropdowns, ...

There is no problem with that to my mind, have studied Semantics components (`Transitions`, `TransitionablePortals`, etc. ) - please, provide the examples.

> - The API isn't consistent with Popups in the community

What are the exact examples? Please, provide them for **controlled modes** of the Popup analogs.

PR #315, opened 2018-10-04
feat(proto): Avatar showing employee card 

mnajdova:
# Prototype

This PR created prototype for EmployeeCard, as well as Avatar component, that shows the EmployeeCard on hovering. Here is how it looks:

image


docs/src/prototypes/employeeCard/Text.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from 'react'
+import { Text as StardustUIText } from '@stardust-ui/react'
+
+const Text = props => {
+  const { muted, ...rest } = props
+  return <StardustUIText {...rest} styles={{ ...(muted && { color: '#888' }) }} />
```
kuzhelov: if `StardustUIText` is an only option here, then this is an important point to reflect as a concern to address. Lets mention it in the PR's description

kuzhelov (approved):

Issue #301, opened 2018-10-02
`debug` package not provided as dependency (should it be?)


dvdzkwsk:
## Bug Report

`debug` is [imported](https://github.com/stardust-ui/react/blob/ec3f3185eb1282d35466358ccbdeb5f0cc793c00/src/lib/debug.tsx#L1) but not listed as a dependency. This should be an easy fix, depending on what you want to do.

![screen shot 2018-10-02 at 3 19 05 pm](https://user-images.githubusercontent.com/6439050/46380280-8d9a8000-c656-11e8-9b82-22585563dd56.png)

### Steps

1. Import stardust
1. Try to use a component
1. Immediate runtime error

### Expected Result

The package to be imported properly.

### Actual Result

Runtime error trying to find `debug`.

### Version
0.7.0

### Testcase

https://codesandbox.io/s/7jp4w964v1

PR #318, opened 2018-10-05
fix(debug): make `debug` a runtime dependency

dvdzkwsk:
Fixes https://github.com/stardust-ui/react/issues/301. Ideally we could get rid of this dependency if in production environments, but this seems to be a fine solution in the meantime.


kuzhelov (approved):

levithomason:
Yep, in SUIR we strip debug statements in the prod build.  This is part of the reason we'll integrate babel back into Stardust soon.

PR #319, opened 2018-10-05
feat(Popup): autocontrolled mode

kuzhelov:
### TODO
- [x] update CHANGELOG
-------
This PR introduces uncontrolled mode to the `Popup`, so now it is possible to consume it in the following way for the simplest scenarios:

```jsx
<Popup content='Popped up!'>
  <Button>Click to open the popup</Button>
</Popup>
```

## FAQ

- **In Uncontrolled mode, we have Popup reacting on `click` trigger's event only - is it usual the popup to be open always on click?**

No, it is not - and **to parametrise this behavioral aspect, we should consider to introduce additional prop (like `triggerAction`) to the API of `Popup`**:

```jsx
// where click, presumably, would be default value
<Popup triggerAction='click' content=..>
   <Button />
</Popup>
```

Also, this will introduce another benefit to us. Now related reasonable question could be asked - is it correct that we have a `PopupBehavior` as a default one (note that `PopupBehavior`'s implementation now is tightly coupled with the click trigger concept - and, thus, should probably be renamed to something that will contain 'click' in its name!). Surely, no. But the reason it is introduce as a default one for now is that it covers the only uncontrolled case we have now. At the same time, **with `triggerAction` prop introduced it would be possible to set accessibility behavior aspects for the `Popup`.**

See this as a step that should be preempted by the set of careful investigations and experiments - and, thus, a separate dedicated PR should be introduced for that.


mnajdova:
I like the idea for introducing the triggerAction. We should investigate more about this.

mnajdova (approved):

PR #320, opened 2018-10-05
fix(Icon): handle brand icons

levithomason:
This PR simply modifies the icon definitions to specify the correct family for each font awesome icon.

image


src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -20,10 +20,7 @@ const sizes = new Map([
 ])
 
 const getDefaultFontIcon = (iconName: string) => {
```
kuzhelov, line 22: it is better to explicitly specify the return type - otherwise we will lose this knowledge

PR #321, opened 2018-10-05
fix(Accessibility): Fix accessibility behaviors' src in tests to accept lower case in naming

sophieH29:


build/gulp/tasks/docs.ts
```diff
@@ -66,7 +66,7 @@ task(
 // ----------------------------------------
 
 const componentsSrc = [`${paths.posix.src()}/components/*/[A-Z]*.tsx`]
-const behaviorSrc = [`${paths.posix.src()}/lib/accessibility/Behaviors/*/[A-Z]*.ts`]
+const behaviorSrc = [`${paths.posix.src()}/lib/accessibility/Behaviors/*/[A-Z a-z]*.ts`]
```
kuzhelov: space should be omitted between `A-Z` and `a-z`, otherwise it will literally match the space character (something that is rather impossible for file names, but the intent should be clear in the expression)

kuzhelov (approved):
please, address comment before merging

build/gulp/tasks/docs.ts
kuzhelov: another question is whether we should still preserve upper case letters as starting ones - as far as I understand, with the rename refactoring there shouldn't be any behaviors starting with capital letter anymore?

kuzhelov (approved):

build/gulp/tasks/docs.ts
sophieH29: I think you're right. We don't need uppercase any more. Thank you!

Issue #191, opened 2018-09-06
Menu item `href` breaks focus (necessary to TAB twice)


tomasiser:
### Steps
1. Go to https://stardust-ui.github.io/react/components/menu, version 0.5.0.
1. Edit the first example by adding `href: '#'`, e.g., with the following code:
   ```
   import React from 'react'
   import { Menu } from '@stardust-ui/react'

   const items = [
     { key: 'editorials', content: 'Editorials', href: '#' },
     { key: 'review', content: 'Reviews', href: '#' },
     { key: 'events', content: 'Upcoming Events', href: '#' },
   ]

   const MenuExample = () => <Menu defaultActiveIndex={0} items={items} />

   export default MenuExample
   ```
1. Try keyboard navigation (`TAB` key) on the menu.

### Expected Result
Pressing `TAB` moves you to the next menu item, `Shift+TAB` to the previous one.

### Actual Result
It is necessary to press `TAB` or `Shift+TAB` **twice** for each menu item. The reason is that using `href` changes the menu item into `<a>`, so there are now two nested `<a>` taking focus one by one.

### Version
0.5.0

miroslavstastny:
`getElementType()` takes `getDefault` parameter but `renderComponent()` never sets that. There is no way how a Component could override that.
@levithomason, @mnajdova, @kuzhelov - before stardust-ui/react-old#21 there was such a possibility. Can we add it back?

kuzhelov:
Agree that using `getDefault`parameter of `getElementType()` function might be a first step to resolve this problem. 

At the same time, the following line of `getElementType` seems like a black magic to me:

```js
// ----------------------------------------
// infer anchor links

if (props.href) return 'a'
```

**_Where this functionality might come in need?_** It seems to be always possible for a client to specify `as='a'` alongside with `href`, so that desired effect would be achieved - and achieved with reasonable explicitness. And the fact that in case of `href` being specified component type suddenly changes its dress to appear as `a` - this is really weird and unexpected thing to me (at least certainly for those cases where component's `as` value is specified in its `defaultProps`).

So, my suggestion would be to avoid this 'implicit "a"', and as further steps - introduce `href` as handled prop of `MenuItem`, so that component will take care of where this attribute should be plaved in its components' tree:

```jsx
// MenuItem
 renderComponent({ ElementType, classes, accessibility, rest }) {
    const { children, content, icon, href } = this.props

    const hasChildren = childrenExist(children)

    return (
[...]

PR #283, opened 2018-09-26
feat(Menu): MenuItemLink component

miroslavstastny:
This is a Proof-of-Concept of splitting `MenuItem` component which renders as `li > a`  to `MenuItem` which renders as `li` and `MenuItemLink` which renders as `a`.
The main motivation for this are issues #79 and #191.

**BEWARE: This is just POC with dirty code and disabled UT's, it must be reworked if we decide to take this path!**

All existing Menu examples render without any change with no visual regressions.

## API

Consider following Menu example which uses MenuItem shorthands:
```jsx
<Menu defaultActiveIndex={0} items={items} />
```

`MenuItem.content` is a shorthand for `MenuItemLink`.

:+1: It can be a string `{ content: 'Home' }`:
```jsx
<li class="ui-menu__item" role="presentation">
  <a class="ui-menu__itemlink" role="menuitem" tabindex="0" data-is-focusable="true">Home</a>
</li>
```

:+1: It can be an object `{ content: {content: 'Home'}}` - the same result as the previous one:
```jsx
<li class="ui-menu__item" role="presentation">
  <a class="ui-menu__itemlink" role="menuitem" tabindex="0" data-is-focusable="true">Home</a>
</li>
```

:+1: It can contain icon (shorthand) `{ content: {content: 'Home', icon: 'home'}}`:
```jsx
<li class="ui-menu__item" role="presentation">
  <a class="ui-menu__itemlink" role="menuitem" tabindex="0" data-is-focusable="true">
    <span class="ui-icon" aria-hidden="true"></span>
    Home
  </a>
</li>
```

:+1: It can contain icon only `{ content: {icon: 'home' }}`:
```jsx
<li class="ui-menu__item" role="presentation">
  <a class="ui-menu__itemlink" role="menuitem" tabindex="0" data-is-focusable="true">
    <span class="ui-icon" aria-hidden="true"></span>
  </a>
</li>
```

:-1: For backwards compatibility, the icon can be specified directly as MenuItem prop `{ content: {content: 'Home'}, icon: 'home' }`.
This can be confusing as icon can be specified both in content and aside the content (the one in content takes precedence). This functionality can be removed if desired.

:+1: It is possible to specify href (and other link attributes for the `a` element `{ content: {content: 'github.com', href:'https://github.com'}}`:
```jsx
<li class="ui-menu__item" role="presentation">
  <a class="ui-menu__itemlink" role="menuitem" tabindex="0" data-is-focusable="true" href="https://github.com">github.com</a>
</li>
[...]

PR #323, opened 2018-10-05
feat(Menu): Add wrapper prop to MenuItem

miroslavstastny:
### Previous approach (MenuItemLink)
This is an alternative proposal to #283.
The original proposal splits `MenuItem` (`ElementType > a` = `li > a`) to `MenuItem` (`li`) and `MenuItemLink` (`a`).
That gives us a lot of flexibility, but makes Menu API more difficult:
- user needs to use nested objects: `{ content: {content: 'github.com', href:'https://github.com'}}`
- it is not clear which properties should go to which object: `{ content: {icon: 'call'}, icon: 'home' }`
- user needs to know where to put accessibility attributes or accessibility behavior needs to 'move' the attributes to the inner `a` (but then it is not possible to specify the attribute for `li`)

That approach also adds additional React component to the tree.

### `MenuItem` `wrapper` slot
This PR keeps `MenuItem` as a single element but renders it as `wrapper > ElementType`
That means that the `a` is now the 'main element'

### Wrapper benefits
Considering `ul > li > a` menu structure, users would need to customise the `a` more often than `li`. This proposal makes customisations to `a` straightforward while keeping customisations to the `li` possible.

### MenuItemLink vs Wrapper comparison

  | MenuItemLink | wrapper
-- | -- | --
content |` {content: 'Home' }` |` {content: 'Home' }`
icon|`{content: {content: 'Home'}, icon: 'home' }` <br> `{content: {content: 'Home', icon: 'home'}}`|` { content: 'Home', icon: 'home'}`
href|`{ content: {content: 'github.com', href:'https://github.com'}}`|`{ content: 'github.com', href:'https://github.com'}`
`as` for inner|`{ content: {as: Button, content:'Home'}}`|`{as: Button, content:'Home'}`
`as` for outer|`{as: 'div', content:'Home'}`|`{content: 'Home', wrapper: <div/>}`
custom outer||`{content: 'Home', renderWrapper: (Component, props) => (<Component {...props} style={{border: '1px solid red'}} />)}`
remove outer||`{content: 'Home', wrapper: <React.Fragment/>}`
custom `role` for outer||`{ content: 'Home', wrapper: {role: 'custom-role'} }`

### Tests
Both `isConformant` and `handlesAccessibility` now support components with wrapper slot using `usesWrapperSlot` config.


kuzhelov:
share the same sentiments around benefits of the proposal (as well as about the problems it addresses), need to make couple of tests to verify it further. Speaking of the tests - this is the thing that is easily adjustable, as long as we will agree to update already taken principles :) But, once again, as it is relatively drastic change, it should be carefully tested before.

**Just for the sake of having a fair overview**: in regards to impossibility of addressing `customOuter` case with the shorthand - is it really the case? Won't it be possible to achieve this functionality by using function shorthand, like

```jsx
...={() => (Component, props) => (<li {...props} style={{border: '1px solid red'}}>{props.children}</li>)}}
```

The same would apply to `removeOuter`'s case.

bmdalex (approved):
Great work @miroslavstastny

test/specs/commonTests/isConformant.tsx
```diff
@@ -28,7 +28,12 @@ export const mount = (node, options?) => {
  * @param {Object} [options.requiredProps={}] Props required to render Component without errors or warnings.
  */
 export default (Component, options: any = {}) => {
-  const { eventTargets = {}, requiredProps = {}, rendersPortal = false } = options
+  const {
+    eventTargets = {},
+    requiredProps = {},
+    rendersPortal = false,
+    usesWrapperSlot = false,
```
bmdalex, line 44: nit: add this to the function's doc above

src/lib/factories.tsx
```diff
@@ -169,3 +169,4 @@ export const createHTMLImage = createShorthandFactory('img', val => ({ src: val
 export const createHTMLInput = createShorthandFactory('input', val => ({ type: val }))
 export const createHTMLLabel = createShorthandFactory('label', val => ({ children: val }))
 export const createHTMLParagraph = createShorthandFactory('p', val => ({ children: val }))
+export const createHTMLLi = createShorthandFactory('li', val => ({ children: val }))
```
jurokapsiar: createHTMLListItem would be more consistent with other factory names

jurokapsiar (approved):

test/specs/commonTests/isConformant.tsx
miroslavstastny: added

miroslavstastny:
One additional note -  As discussed with @jurokapsiar offline, if this approach gets merged, we will review MenuItem behaviors and simplify them as a lot of their logic is no longer necessary after this change.
We will be probably able to completely remove `handledProps` concept from accessibility.

This PR fixes #191.
This PR obsoletes #283.

src/components/Menu/MenuItem.tsx
```diff
@@ -121,51 +129,71 @@ class MenuItem extends UIComponent<Extendable<IMenuItemProps>, IMenuItemState> {
...
+          Icon.create(this.props.icon, {
+            defaultProps: { xSpacing: !!content ? 'after' : 'none' },
+            render: renderIcon,
+          })}
+        {content}
       </ElementType>
     )
+
+    if (wrapper) {
+      return createHTMLListItem(wrapper, {
```
bmdalex: don't we want this to be an actual `Slot` shorthand so it's being passed through the Fella renderer? look for `Slot.create` references; we can discuss tomorrow

src/components/Menu/MenuItem.tsx
```diff
@@ -121,51 +129,71 @@ class MenuItem extends UIComponent<Extendable<IMenuItemProps>, IMenuItemState> {
...
+        defaultProps: {
+          className: classes.root,
```
bmdalex: there are a few problems caused by this; when we pass custom `styles` prop at the `MenuItem` level, they will be applied to the `li` wrapper instead of the `a` link

screen shot 2018-10-10 at 22 24 53

It can be fixed by renaming the `root` slot to something like wrapper here and add `classes.root` to the `a` link; here are the changes:

screen shot 2018-10-10 at 22 39 29

and here is the expected result (contains the changes):

screen shot 2018-10-10 at 22 38 14

bmdalex: problem is the tests will fail now..

miroslavstastny: Good catch, thanks!
The main problem is that `ui-menu__item` class now goes to the `a` and nothing to the `li`.
Not sure if it would be ok to have `ui-menu__item` on the inner `a` and `ui-menu-wrapper` on the outer `li`.
Plus similar change should be done in behaviors.
I will look into that.

bmdalex: well we can definitely rename `ui-menu__item` to `ui-menu__item__link` or something similar and use `item` instead of `wrapper` an name that `ui-menu__item` instead of `ui-menu-wrapper`

bmdalex: here is another example that shows that `styles` prop and `wrapper.style` are styling the same element, the wrapper and `wrapper.style` with the latter having higher priority since it's applied to the element style directly; currently we cannot target the `a` link to style it:

screen shot 2018-10-11 at 12 44 55

src/components/Menu/MenuItem.tsx
bmdalex: To expand here, we need to pass the content of `wrapper` prop through the Fella renderer in order to leverage Fella's functionality (rtl, styles applied as classnames, etc)

The way to go now is by using `Slot`; I'm doing the same for `Input` in #326 

Right now we need to use `wrapper.style` instead of `wrapper.styles` to style the wrapper and these are applied  directly to `element.style` and not through `classnames`:

screen shot 2018-10-11 at 13 01 58

`wrapper.styles` won't work (it will be added directly to the `DOM` as an attribute), need to use `Slot` for that:

screen shot 2018-10-11 at 13 02 15

miroslavstastny:
I've updated the PR as follows:
- `Slot` is used for wrapper
- in styles, inner anchor is considered 'root', wrapper is 'wrapper'
- inner anchor class is `ui-menu__item`, wrapper class is `ui-menu__item__wrapper`
- every prop applied on `MenuItem` component goes to anchor (including `style` and `styles`), use `wrapper` prop to alter the `li`:

Example
```jsx
const items = [
...
  {
    key: 'review',
    content: 'Reviews',
    styles: { borderLeft: '5px solid red' },
    style: { background: 'pink' },
    className: 'CLASS-ANCHOR',
    'data-custom': 'DATA-ANCHOR',
    wrapper: {
      styles: { paddingLeft: '50px' },
      style: { background: 'yellow' },
      className: 'CLASS-WRAPPER',
      'data-custom': 'DATA-WRAPPER'
    },
  },
...
]
```

Renders as:
```html
<li role="presentation"
    style="background:yellow"
    data-custom="DATA-WRAPPER"
    class="ui-slot dy ui-menu__item__wrapper CLASS-WRAPPER">

     <a class="ui-menu__item ea CLASS-ANCHOR"
        role="menuitem"
        tabindex="0"
        data-is-focusable="true"
        style="background:pink"
        data-custom="DATA-ANCHOR">
          Reviews
     </a>
</li>
```

image

@Bugaa92 would you mind reviewing again? Thanks!

src/components/Menu/MenuItem.tsx
```diff
@@ -121,51 +124,71 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
     /** Override for theme site variables to allow modifications of component styling via themes. */
     variables: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
+
+    /** Shorthand for the wrapper component */
+    wrapper: PropTypes.oneOfType([PropTypes.node, PropTypes.object]),
   }
 
   static defaultProps = {
-    as: 'li',
+    as: 'a',
```
bmdalex, line 144: when we discussed changes for `Input` component we had a consensus that **`as` and `styles` props always go to the root element**; we might want the same for `MenuItem`; let's see what others say @levithomason @kuzhelov

src/components/Menu/MenuItem.tsx
```diff
@@ -121,51 +124,71 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
   renderComponent({ ElementType, classes, accessibility, rest }) {
-    const { children, content, icon, renderIcon } = this.props
+    const { children, content, icon, renderIcon, renderWrapper, wrapper } = this.props
 
-    return (
+    const menuItemInner = childrenExist(children) ? (
+      children
+    ) : (
       <ElementType
         className={classes.root}
```
bmdalex, line 158: when we discussed changes for `Input` component we had a consensus that **`styles` prop always goes to the root element**; we might want the same for MenuItem; this will break that convention when `wrapper` prop is defined; let's see what others say @levithomason @kuzhelov

src/lib/factories.tsx
```diff
@@ -166,3 +166,4 @@ export const createHTMLImage = createShorthandFactory('img', val => ({ src: val
 export const createHTMLInput = createShorthandFactory('input', val => ({ type: val }))
 export const createHTMLLabel = createShorthandFactory('label', val => ({ children: val }))
 export const createHTMLParagraph = createShorthandFactory('p', val => ({ children: val }))
+export const createHTMLListItem = createShorthandFactory('li', val => ({ children: val }))
```
bmdalex: no need to add this as we're removing HTML factories in #376

src/components/Slot/Slot.tsx
```diff
@@ -57,6 +57,7 @@ class Slot extends UIComponent<Extendable<SlotProps>, any> {
 
   static create = createSlotFactory(Slot.defaultProps.as, content => ({ content }))
   static createHTMLInput = createSlotFactory('input', type => ({ type }))
+  static createHTMLListItem = createSlotFactory('li', children => ({ children }))
```
bmdalex: no need to add this as we're removing HTML factories in #376
use `createSlotFactory('li', children => ({ children }))` directly where you need it

src/components/Menu/MenuItem.tsx
```diff
@@ -121,51 +124,71 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
+          Icon.create(this.props.icon, {
+            defaultProps: { xSpacing: !!content ? 'after' : 'none' },
+            render: renderIcon,
+          })}
+        {content}
       </ElementType>
     )
+
+    if (wrapper) {
+      return Slot.createHTMLListItem(wrapper, {
```
bmdalex: this was discussed in previous sync meetings; pls use `createSlotFactory('li', children => ({ children }))` directly; see comments in `Slot.tsx`

test/specs/commonTests/isConformant.tsx
```diff
@@ -58,6 +61,14 @@ export default (Component, options: Conformant = {}) => {
         component = component.childAt(0) // skip the additional wrap <div> of the FocusZone
       }
     }
+
+    if (usesWrapperSlot) {
```
bmdalex, line 65: @kuzhelov @levithomason 
I think we should avoid changing `isConformant` tests for the elements with wrapper slot; we had this discussion for `Input` component.

@miroslavstastny AFAIR, the tests that fail are related to `as` and `className` props not being set correctly; these can be solved by addressing the comments I had in `MenuItem.tsx` where `as` and `styles` should always be passed to the root element, being that the `a` or the `li` for the `MenuItem` case.

This probably requires a session where we can discuss options in more detail; @miroslavstastny feel free to schedule a meeting with me and @kuzhelov

bmdalex (changes requested):
pls take a look at comments, thanks

test/specs/commonTests/isConformant.tsx
miroslavstastny: Thank you very much for the review, @Bugaa92.

So the main question is whether  we should split props between `li` and `a` as we do in Input or pass everything to the `a`.

I am not strictly against doing the split, but when I was considering this I thought it would make more sense to do 'everything to `a`' although it is different approach than in `Input`.

We should consider usability and API consistency first and test after that.

Agree to discuss this offline.

kuzhelov: thanks for sharing your thoughts, @miroslavstastny, @Bugaa92 

> So the main question is whether  we should split props between li and a as we do in Input or pass everything to the a.

If I would answer this question only **for the MenuItem component** I would rather choose to not split and apply all the props to wrapped `<a>` element. This would make the following usage case be quite more apparent to me as a user, _given that I am aware of wrapper concept_

```jsx
<MenuItem styles={..} href={..} />
```

It would be really confusing and non-intuitive to me if `href` will be applied to `<a>` element, while, at the same time, `styles` would be applied to `<li>`. In this situation I would rather expect consistent behavior, i.e. both props being applied to `<a>`.

--------------

At the same time, agree with the @Bugaa92's reasons for the concerns he has expressed, although see `Input's` situation to be a bit different: while for the menu's case it is a **web standard** that declares `li` to be a wrapper, in contrast, for the `Input` component's case it is a Stardust's need to use wrapper `div` (what is more interesting, this one is necessary for _styling_ aspects, a bit smelly fact to me). 

Agree to strive for consistent and intuitive API, and for this sake, let's, please, schedule a meeting as you've proposed - I will be back staring from 7th of November, but let me know if it should be decided earlier, will do all my best to allocate time for that. 

Thank you!

bmdalex (approved):

src/components/Menu/MenuItem.tsx
levithomason: IIRC there is an accessibility requirement that the click listener needs to be on the anchor tag.  The primary use case for `as` on the menu item is to allow adding router functionality.  This means the router's navigation handler needs to be applied to the anchor:

```jsx
import { NavLink } from 'react-router'

<Menu items={[{ as: NavLink, to: '/', content: 'Home' }]} />
```

If the `as` prop goes to the root, it will still work as the event will bubble from the anchor up to the `li`, however, it would not be the correct markup and behavior for the accessibility requirement.

levithomason (approved):

src/components/Menu/MenuItem.tsx
```diff
@@ -123,46 +136,56 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
+    if (wrapper) {
+      return Slot.create(wrapper, {
+        defaultProps: {
+          className: cx('ui-menu__item__wrapper', classes.wrapper),
+          ...accessibility.attributes.root,
+          ...accessibility.keyHandlers.root,
+        },
+        render: renderWrapper,
+        overrideProps: () => ({
+          children: menuItemInner,
```
bmdalex, line 184: very nice usage of `overrideProps` for `children` in the wrapper; basically we achieve not allowing the user to override `children` through `wrapper` prop 👍 
I'll create a PR to do the same for `Input` component

src/components/Menu/MenuItem.tsx
```diff
@@ -123,46 +136,56 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
     renderIcon: PropTypes.func,
+    wrapper: PropTypes.oneOfType([PropTypes.node, PropTypes.object]),
+    renderWrapper: PropTypes.func,
   }
 
   static defaultProps = {
-    as: 'li',
+    as: 'a',
     accessibility: menuItemBehavior as Accessibility,
+    wrapper: { as: 'li' },
```
bmdalex, line 146: 👍

bmdalex (approved):

src/components/Menu/MenuItem.tsx
layershifter: In this case we can use `overrideProps` as object

PR #324, opened 2018-10-05
docs(Architecture): Added architecture guide

smykhailov:
The guide which describes core architecture principles used in the SUI. 
Especially, theming and accessibility.


docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
...
...41l7urXtbudO2OXK39NMfr8oflK5MhFSZQ75tYOBR3vsce6KjRZtjdaeaH/0hOuCHuxpaSQkWdIMSRYhJiRZhFphaWlZycrF8AtW1XiMjHQ/lNl0ryMz+6XOWe/lvffajgxaefyXoG1WXRL2ufTCXvfOwpyogz6cgPnJJ57z4W/iKeR6e7k4cWKWu40Lf2aKb8cfw9/Uk+9jr3tXbLPpELXqRMugo05Tgs76LG2512ZoHfZUNB6SLGmGJIsQjW+++TqN++HBpTVf//zzT2xYfTFf//rrL3ay+mF5vcHgChUq8HV3vt7guwZNq1atwtf51OBSoU7dWjn1X7l88NMvTW35er9Bv3u4uZ36qEOX1rv5euMm33lYuR77pVPXNkMrV67oUbFiBY+LTkbzO3Zv06xWrRpCH2bnDQ5qa2tX+q7RV0J9wx/6btxPrs8mdnx9wOCeHqdO7fqofefWBny9SdOGHhfsTFp27tZqRMWKXJ9cvxedTea079aqeU3W5wmLvQf4K9S+ayjvc+P2ZS7czw9++rmJPV8fNEzLw97eqHrHTq328vUmzRp6WDmY/tSxS5tRfJ+VK1f2sHQ0mtWly68tatZkfZ7ft79NmzYffvvdl0J9885lztzP3D6HjuzjYWGxp1qHTq328/WmLRp52Dofa9Gpa6sxfJ8ffvihh5WTyYzREwf+9PEndYU+/jBYtbdfv37/a9mqmVBfsGQaf/fbD3podXLg61p9OnuYmcmqjhzT/wBfb9joa4+9h9c2154weBz/evJ97j0imzZq/NCfP/64jtDHtr0rDXR0ulfl3hOhvmj5dL6vD3r06uDI13v37+Yhk8mqjhjd7xBfb9T4W66PTU25PifyfVapwvV5VDZl7OTBLeuyPrnnuWfUqFHVfmyZ0+cM/nfO7bPPgN899u9f8L/ho/oe4ev8+777oKwJ97tO5n93/u/H4MiGSdrjB/xal/3tbN+7ereu7qDqP/7URKgvXjGDf78/6N6zgxP3+nt17NyahguLwOP6wY8drxpoGdiO1lpq9KvWymO/asnOd9A677v833M+S7HHrr+QP6y6Y/XpH7Hu7C/Ybdsnd/sh51Gw8F2JLRe79pWZd+D6+FnriONULbewXR+xhyAK0PbQ+v/ammwGRWIx3IgOJFlEmfPBB/8HSefvMf4S+NwAAAAASUVORK5CYII=" />
```
kuzhelov: it is better to introduce this image to the repo and reference it by path - otherwise it is impossible to review just the image's content

smykhailov: done

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
...
+      The library encapsulates accessibility into <strong>Behaviors</strong>. The behaviors returns
+      the recommended set of attributes and roles for the standard components (e.g. Image, DropDown,
+      Dialog, etc), based on their state (e.g. active, disabled, expanded, etc). Also, it provides
+      the expected keyboard navigation and focus handling for the components (e.g. List, Grid, Radio
+      Group, etc).
+    </p>
+    <p>
+      Library provides the set of accessibility behaviors and keyboard handlers for the components
+      and allows to create and inject your own, if it is necessary for some custom components. In a
+      most of the cases there should be enough to use the existing from the library.
```
hughreeling: In most cases it should be enough to use the behaviors provided in the libary, but, of course, you can provide your own or extend existing behaviors.

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
...
+    <Header as="h3">Accessibility</Header>
+    <p>
+      One of the main purposes of the library is to make component accessible and provide ease
+      accessibility integration for the entire projects which will be consumer of the library.
+    </p>
+    <p>
+      The library encapsulates accessibility into <strong>Behaviors</strong>. The behaviors returns
+      the recommended set of attributes and roles for the standard components (e.g. Image, DropDown,
+      Dialog, etc), based on their state (e.g. active, disabled, expanded, etc). Also, it provides
+      the expected keyboard navigation and focus handling for the components (e.g. List, Grid, Radio
```
hughreeling: We've started off stating that everything is included in behaviors, and then sneak in the keyboard handlers.  let's reword these two to make clearer.

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
...
+      Group, etc).
+    </p>
+    <p>
+      Library provides the set of accessibility behaviors and keyboard handlers for the components
+      and allows to create and inject your own, if it is necessary for some custom components. In a
+      most of the cases there should be enough to use the existing from the library.
+    </p>
+    <p>
+      In addition to behaviors and keyboard handlers, the toolkit also provides FocusZone component,
+      which can be used to add focus handling for the wrapped component. There is also FocusTrap
```
hughreeling, line 99: Let's add explanation of what focus handling is in the glossary.  We use the term, but people outside the team might have different understanding.  Then different between focus handling and focus trap will be clearer.

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
...
+      component, which allows to trap the focus inside some area, which is useful for such a
+      component as Dialog.
+    </p>
+
+    <Header as="h3">Theming</Header>
+    <p>
+      Component logic allows to inject and switch the component theme (e.g. Light, Dark or High
```
hughreeling: How about changing to 'You can inject different themes into Stardust components'

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
...
+      Contrast). To achieve this, we use CSS in JS technique, which is provided by{' '}
+      <a href="http://fela.js.org/" target="_blank">
+        Fela library
+      </a>.
+    </p>
+    <p>
+      Each component has its own set of themes defined. Component logic handles theme switching and
+      style overriding by providing ability to set <strong>variables</strong> or override component
+      (ot its children) styles using <code>styles</code> property.
```
hughreeling: typo

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
...
+    </p>
+    <p>
+      Components are relaying on the theme provider, which has to wrap components to provide default
```
hughreeling: Component styling is managed by the theme provider

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
...
+      theme and theme switching. Providers can be nested, this is necessary in case if you need to
+      use different theme in children and parent components.
```
hughreeling: Link to themeing documentation page?

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
...
+    <p>
+      We recommend to use <code>UIComponent</code> instead of <code>React.Component</code> when you
+      create your own primitive or composed component, to leverage ability of injecting
+      accessibility and theming, instead of implementing your own mechanisms for handling these
+      parts.
+    </p>
+
+    <Header as="h3">AutoControlledComponent</Header>
+    <p>
+      The <code>AutoControlledComponent</code> is the base component, which maintains a component
```
hughreeling: Can't use 'is the base component' again.  AutoControlledComponent is another base component that is used when it's necessary to maintain state

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
...
+      state and allows to set its initial state based on default properties and allows to override
+      state by property with the same name.
+    </p>
+
+    <Header as="h2">Component Library</Header>
+    <p>
+      The Component Library is a catalog of components, primitive and composed, which has predefined
+      behavior, themes and accessibility and can be customized using their properties or with
```
hughreeling: or by

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
...
+      injecting additional themes and/or children components.
+    </p>
+
+    <Header as="h3">Components</Header>
+    <p>
+      Each component is inherited from <code>UIComponent</code> or{' '}
```
hughreeling, line 155: remove {' '} stuff

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
...
+      <code>AutoControlledComponent</code> and includes the logic to handle <strong>theming</strong>{' '}
+      and <strong>accessibility</strong>. Each component has default{' '}
+      <strong>accessibility behavior</strong>, which can be overridden by consumer of the library.
+      The same can be done with component look and feel by overriding theme styles or defining theme
```
hughreeling: not defining theme variables - perhaps setting, or overriding again?

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
...
+      variables.
+    </p>
+    <p>
+      Additional info about how to use themes or override styles can be found in the{' '}
+      <a href="themeing">Theming guide</a>.
+    </p>
+
+    <Header as="h3">Accessibility Behaviors</Header>
+    <p>
+      <strong>Accessibility Behavior</strong> is the function, which returns the set of the
```
hughreeling: An Accessibility Behavior is a function that

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
...
+        `      'aria-disabled': !!props['disabled'],`,
+        `    },`,
+        `  },`,
+        `})`,
+        ``,
+        `export default ButtonBehavior`,
+      ].join('\n')}
+    />
+    <p>
+      Each component has default accessibility behavior, which can be easily overridden by injecting
```
hughreeling: has a default

hughreeling: Remove easily - doesn't add anything

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,213 @@
+import * as React from 'react'
+import { NavLink } from 'react-router-dom'
+import { Header, Divider } from 'semantic-ui-react'
```
hughreeling: Let's use Stardust where we can

docs/src/views/Architecture.tsx
smykhailov: added

docs/src/views/Architecture.tsx
smykhailov: this is the code added by pretier, or some similar tool, to add the space between the words when they are located on separate lines. So, it can't be removed :)

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,224 @@
...
+        <strong>Accessibility keyboard handlers</strong> - provides the logic to handle components
+        navigation and focus.
+      </li>
+      <li>
+        <strong>Themes</strong> - the logic to inject and handle component visual representation
+        (theme).
+      </li>
+      <li>
+        <strong>Renderer</strong> - to render the HTML output, the toolkit uses renderer, which is
+        now React.
```
hughreeling: to render the HTML Output.  Current implementation uses React, but it can be extended to other frameworks

hughreeling (approved):

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,224 @@
...
+        <strong>Accessibility keyboard handlers</strong> - provides the logic to handle components
+        navigation and focus.
+      </li>
+      <li>
+        <strong>Themes</strong> - the logic to inject and handle component visual representation
+        (theme).
+      </li>
+      <li>
+        <strong>Renderer</strong> - to render the HTML Output. Current implementation uses React,
+        but it can be extended to other frameworks.
```
levithomason, line 60: This is the React docs site. Let's not discuss other renderers here.  That conversation belongs at the top level of the Stardust UI org where the Stardust UI concept as a whole is explained.

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,224 @@
...
+      UI framework agnostic and written in pure TypeScript. Stardust UI uses this toolkit to create
+      our own set of UI Components and allows other consumers to create their own primitive or
+      composed components. This architecture allows to combine Stradust UI primitive or composed
+      components with the custom primitive or composed components made by third parties, who were
+      using the toolkit to create their own components.
+    </p>
+    <p>The toolkit combines the following parts:</p>
+    <ul>
+      <li>
+        <strong>Component Logic</strong> - the framework core logic, which couples all other parts
```
levithomason, line 40: We have no tools for component logic.  Component logic is implemented by each component on an as needed basis. I don't think we should include this in the docs.

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,224 @@
...
+        accessibility roles and <code>aria-*</code> attributes for the given component based on the
+        component state.
+      </li>
+      <li>
+        <strong>Accessibility keyboard handlers</strong> - provides the logic to handle components
+        navigation and focus.
+      </li>
+      <li>
+        <strong>Themes</strong> - the logic to inject and handle component visual representation
+        (theme).
```
levithomason, line 56: Injecting themes is done by the Provider.  We do not have a toolkit for themes.  There is also no logic involved in themes.

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,224 @@
...
+      component as Dialog.
+    </p>
+
+    <Header as="h3">Theming</Header>
+    <p>
+      You can inject different themes into Stardust components (e.g. Light, Dark or High Contrast).
+      To achieve this, we use CSS in JS technique, which is provided by{' '}
+      <a href="http://fela.js.org/" target="_blank">
+        Fela library
+      </a>.
```
levithomason, line 110: Fela is an implementation detail that we should not expose.  It is subject to change at any time.  We should have a guide on CSS in JS but it should not refer to internal libraries.  It should show the capabilities of Stardust's CSS in JS handling alone.

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,224 @@
...
+    <p>
+      In addition to behaviors and keyboard handlers, the toolkit also provides FocusZone component,
+      which can be used to add focus handling for the wrapped component. There is also FocusTrap
+      component, which allows to trap the focus inside some area, which is useful for such a
+      component as Dialog.
+    </p>
+
+    <Header as="h3">Theming</Header>
+    <p>
+      You can inject different themes into Stardust components (e.g. Light, Dark or High Contrast).
```
levithomason, line 106: Components cannot receive themes, only the Provider can receive a theme.  Components can only receive variables and styles props.

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,224 @@
...
+      To achieve this, we use CSS in JS technique, which is provided by{' '}
+      <a href="http://fela.js.org/" target="_blank">
+        Fela library
+      </a>.
+    </p>
+    <p>
+      Each component has its own set of themes defined. Component logic handles theme switching and
+      style overriding by providing ability to set <strong>variables</strong> or override component
+      (or its children) styles using <code>styles</code> property.
```
levithomason, line 115: Components do not define themes, only the Provider defines a theme.  Component logic does not handle theme switching, all theme switching is handled solely by the user and what they pass to the Provider's theme prop.  All theming concerns are intentionally kept outside of the component.  Components are more like templates whose template slots are factories for creating components.

Components can override the theme by passing variables and styles as noted, however, this is duplicating what is already laid out in `/theming`.

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,224 @@
...
+    </p>
+
+    <Header as="h3">UIComponent</Header>
+    <p>
+      The <code>UIComponent</code> is the base component, which handles component default
+      properties, accessibility navigation and focus handling and renders the component.
+    </p>
+    <p>
+      We recommend to use <code>UIComponent</code> instead of <code>React.Component</code> when you
+      create your own primitive or composed component, to leverage ability of injecting
```
levithomason, line 134: We **do not** recommend using UIComponent, we'd actually like to get rid of it if possible.  The API for consumers to create their own components is still underway, however, we hope that it does not involved extending our component.  It will likely be something like `createComponent()` or `renderComponent()`.

We do not want to spread knowledge of UIComponent.

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,224 @@
...
+      accessibility and theming, instead of implementing your own mechanisms for handling these
+      parts.
+    </p>
+
+    <Header as="h3">AutoControlledComponent</Header>
+    <p>
+      The <code>AutoControlledComponent</code> is another base component that is used when it's
```
levithomason, line 141: This is also not a solution we support.  It was hastily copied from SUIR as a stop gap.  Instead, Stardust UI promotes the use of vanilla JS state managers to handle state.  This is another area of tech debt that we shouldn't promote as a good idea.

docs/src/views/Architecture.tsx
```diff
@@ -0,0 +1,224 @@
...
+      necessary to maintain state. It allows to set its initial state based on default properties
+      and allows to override state by property with the same name.
+    </p>
+
+    <Header as="h2">Component Library</Header>
+    <p>
+      The Component Library is a catalog of components, primitive and composed, which has predefined
+      behavior, themes and accessibility and can be customized using their properties or by
+      injecting additional themes and/or children components.
```
levithomason, line 150: This is true, however, this is what the entire `/react` package is about.  Something like this could go on the intro page.

levithomason (changes requested):
I have many concerns with this PR.  The majority of the story communicated is not inline with the vision of Stardust UI.  The parts that are inline either duplicate other docs or should be separated into their own doc pages.

If someone will adopt the work of describing the architecture as a whole I propose we schedule a meeting to scaffold the outline of this document and where the information should be presented.  Some information is relevant to consumers, some is relevant to producers, some is relevant to React users and some is relevant only at the root level of the Stardust UI organization.

I'm very happy to outsource writing the docs on our architecture, however, it needs to be cohesive.  Our documentation is inconsistent in many places already (e.g. compare the glossary to our actual implementation).  We also have many areas and ideas that are missing or broken (examples, component descriptions, prop descriptions, etc).  I do not want to ship anymore conflicting information without first normalizing and aligning our vision and direction amongst ourselves.

levithomason:
Note, the vision outline has been started in the `docs/vision` branch.  The markdown outline is here:

https://github.com/stardust-ui/react/blob/docs/vision/VISION.md

We are iterating on this twice a week in our vision and architecture meetings.  Once complete, we should be all clear for some doc contributions based on the outline.

PR #325, opened 2018-10-05
(Docs): Add prototype for chat pane with virtualized list

sophieH29:
This PR adds chat pane prototype with a virtualized list, using a ```react-virtualized``` library.


sophieH29:
Not relevant right now.

PR #326, opened 2018-10-05
feat(input): new Input component with 'input' slot

### LINKED ISSUE #314: [RFC] New Input component
# New Input component

## Problem description

The current `Input` component looks like this:

##### API
```jsx
<Input value="123" icon="search" placeholder="Search..." />
```

##### DOM structure:
```html
<div class="ui-input">
  <input type="text" placeholder="Search..." value="123" />
</div>
```

##### Rendered component:
screen shot 2018-10-04 at 17 08 37

This introduces a few issues. Since it **directly** creates an HTML `<input />` element (implementation detail), it does not support a set of user scenarios:

#### 1. Styling the `<input />`
It's impossible to target the `<input />` element in order style it (e.g.: change the height of the input through styles) because it's not implemented as a shorthand slot.
Also, custom `styles` passed as prop will be set for the wrapping `<div />` instead of `<input />`:
```jsx
<Input placeholder="Search..." styles={{ backgroundColor: 'red' height: '80px' }} />
```
renders:
screen shot 2018-10-04 at 17 24 53
instead of expected:
screen shot 2018-10-04 at 17 37 53

#### 2. Setting HTML attributes for the `<input />`
It's impossible to target the `<input />` element in order to set attributes for it (e.g.: setting `tabIndex="-1"` to the `<input />` element so that it's skipped by keyboard navigation).
- passing `tabIndex="-1"` as prop adds it to the wrapping `<div />` so it doesn't work:
```jsx
<Input placeholder="Search..." tabIndex="-1" />
```
generates
```html
<div class="ui-input" tabindex="-1">
  <input type="text" placeholder="Search..." value="" />
</div>
```
- passing `input={{ tabIndex:"-1" }}` as prop does not target the `<input />` element because it's not implemented as a shorthand slot:
```jsx
<Input placeholder="Search..." input={{ tabIndex:"-1" }} />
```
generates
```html
<div class="ui-input" input="[object Object]">
  <input type="text" placeholder="Search..." value="" />
</div>
```
instead of expected
```html
<div class="ui-input">
  <input type="text" placeholder="Search..." value="" tabindex="-1" />
</div>
```

#### 3. Setting HTML attributes for the wrapping `<div />`
Because of implementation details, some HTML attributes (the ones specific to `<input />`) end up in the `<input />` element instead of the wrapping `<div />`:

```jsx
<Input placeholder="Search..." role="presentation" />
```
generates
```html
<div class="ui-input">
  <input type="text" placeholder="Search..." role="presentation" value="" />
</div>
```
instead of expected
```html
<div class="ui-input" role="presentation">
  <input type="text" placeholder="Search..." value="" />
</div>
```

## Proposed solutions

Both solutions rely on 2 main concepts:

#### i. Creating an additional standalone `InputBase` component

`InputBase` component will have the following structure:
##### Props:
```typescript
interface IInputBaseProps {
  as?: any // TODO: decide if we want it
  className?: string
  defaultValue?: React.ReactText
  fluid?: boolean
  inline?: boolean
  onChange?: ComponentEventHandler<IInputProps>
  placeholder?: string
  value?: React.ReactText
  type?: string
}
```

##### API
```jsx
<Input placeholder="Search..." />
````

##### DOM structure
```html
<input type="text" placeholder="Search..." value="" />
```

For simple scenarios where other actions / elements are not needed, `InputBase` is all the user needs. It will also fix all scenarios mentioned in the `Problem description`:

##### 1. Styling the `<input />`
```jsx
<InputBase placeholder="Search..." styles={{ backgroundColor: 'red' height: '80px' }} />
```

##### 2. Setting HTML attributes for the `<input />`
```jsx
<InputBase placeholder="Search..." tabIndex="-1"  />
```

#### ii. Reimplementing `Input` using `InputBase` as shorthand
For scenarios like:
- input with `search` icon;
- clearable input;
- clearable input with `search` icon;

### Solution 1: `wrapper` prop

`Input` will leverage `InputBase` and use it internally.
All the `props` sent as `<Input ...{props} />` will be passed to the `InputBase` unless the user clearly specifies they are for the wrapping element using the `wrapper` prop.

##### Props:
```typescript
interface IInputProps extends IInputBaseProps {
  clearable?: boolean
  icon?: ItemShorthand
  wrapper?: ItemShorthand
}
```

##### API
```jsx
<Input {...inputProps} wrapper={wrapperProps} />
````
e.g.: 
```jsx
<Input
  placeholder="Search..."
  icon="search"
  clearable
  wrapper={{ role: "presentation" }}
/>
````

##### Component structure
```html
<ElementType ...> // <div />
  <InputBase ... />  // <input />
  <Icon ...>  // <span />
</ElementType>
```

##### DOM structure
```html
<div class="ui-input" role="presentation">
  <input type="text" placeholder="Search..." value="" />
</div>
```

This will solve all the scenarios mentioned above in the following fashion:

##### 1. Styling the `<input />`
```jsx
<Input
  placeholder="Search..."
  icon="search"
  clearable
  styles={{ backgroundColor: 'red' height: '80px' }}
/>
```

##### 2. Setting HTML attributes for the `<input />`
```jsx
<Input placeholder="Search..." tabIndex="-1" clearable />
```

##### 3. Setting HTML attributes for the wrapping `<div />`
```jsx
<Input placeholder="Search..." wrapper={{ role: "presentation" }} />
```

### Solution 2: `input` slot

`Input` will leverage `InputBase` and use it as a shorthand slot, the same it does for `Icon`. The structure will be:

##### Props:
```typescript
interface IInputProps extends IInputBaseProps {
  clearable?: boolean
  icon?: ItemShorthand
  input?: ItemShorthand
}
```
##### API
- preferred:
```jsx
<Input placeholder="Search..." icon="search" clearable />
````
- shorthand alternative 1 or when we need to target the `<input />`:
```jsx
<Input input={{ placeholder: "Search..." }} icon="search" clearable />
````
- shorthand alternative 2 or when we need to target the `<input />`:
```jsx
<Input input={<InputBase placeholder="Search..." />} icon="search" clearable />
````

##### Component structure
```html
<ElementType ...> // <div />
  <InputBase ... />  // <input />
  <Icon ...>  // <span />
</ElementType>
```

##### DOM structure
```html
<div class="ui-input">
  <input type="text" placeholder="Search..." value="" />
</div>
```

This will solve all the scenarios mentioned above in the following fashion:

##### 1. Styling the `<input />`
```jsx
<Input input={{ placeholder: "Search...", styles: { backgroundColor: 'red' height: '80px' } }} icon="search" clearable />
```

##### 2. Setting HTML attributes for the `<input />`
```jsx
<Input input={{ placeholder: "Search...", tabIndex: "-1" }} clearable />
```

##### 3. Setting HTML attributes for the wrapping `<div />`
```jsx
<Input placeholder="Search..." role="presentation" />
```

bmdalex:
# [RFC] Input

**New input component according following the strategy in #314** (closes #314)
This PR contains:
- new implementation of input (using `input` and `wrapper` slots);
- `Slot` factory `v1`;
- fix for input height in the chat prototype;
- fix for tab keyboard navigation for radio group with inline input;
- fixes for generic unit tests for shorthands;
- lots of unit tests to cover `Slot` factory, `input` and `wrapper` slots for `Input` component, `controlled`/`uncontrolled` mode for `Input`, etc

### TODO

- [x] Conformance test
- [ ] Stardust base theme
- [x] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [x] Confirm RTL usage
- [x] Update the CHANGELOG.md

## Proposed solution

- create an `Input` component consisting of an `<input />` element wrapped with a `<div />`;
- use `partitionHTMLProps` to decide what `props` (`<Input {...props} />`) go to the `<input />` element or wrapping `<div />`;
- use `input` slot to target the `<input />` element  and `wrapper` slot to target wrapping `<div />` when `partitionHTMLProps` does not work as expected:
```jsx
<Input {...partitionedProps} input={inputProps} wrapper={wrapperProps} />
```
e.g.:
```jsx
<Input
  placeholder='Search...'
  dir='ltr'
  tabIndex={2}
  disabled={false}
  role='presentation'
  input={{ dir: 'rtl', tabIndex: 0 }}
  wrapper={{ disabled: true, role: 'checkbox' }}
/>
```

`Input` component has the following structure:
##### Props:
```typescript
export interface IInputProps {
  as?: any
  className?: string
  clearable?: boolean
  defaultValue?: React.ReactText
  fluid?: boolean
  icon?: ShorthandValue
  inline?: boolean
  input?: ShorthandValue
  onChange?: ComponentEventHandler<IInputProps>
  renderInput?: ShorthandRenderFunction
  renderWrapper?: ShorthandRenderFunction
  styles?: ComponentPartStyle<IInputProps, any>
  type?: string
  value?: React.ReactText
  wrapper?: ShorthandValue
}
```

##### API
```jsx
<Input {...inputProps} wrapper={wrapperProps} />
````
e.g.: 
```jsx
<Input
  placeholder="Search..."
  icon="search"
  clearable
  wrapper={{ role: "presentation" }}
/>
````

##### Component structure
```html
<ElementType ...> // <div />
  <InputBase ... />  // <input />
  <Icon ...>  // <span />
</ElementType>
```

##### DOM structure
```html
<div class="ui-input" role="presentation">
  <input type="text" placeholder="Search..." value="" />
</div>
```

This will solve all the scenarios mentioned above in the following fashion:

##### 1. Styling the `<input />`
```jsx
<Input
  placeholder="Search..."
  input={{ styles: { backgroundColor: 'red' height: '80px' } }}
/>
```

##### 2. Setting HTML attributes for the `<input />`
```jsx
<Input placeholder="Search..." input={{ tabIndex='-1' }} />
```

##### 3. Setting HTML attributes for the wrapping `<div />`
```jsx
<Input placeholder="Search..." wrapper={{ role: "presentation" }} />
```

##### Rendered component:
screen shot 2018-10-05 at 17 50 53


bmdalex:
@levithomason @kuzhelov @miroslavstastny @mnajdova review needed

bmdalex:
I had to comment out conformance test for wrapped input, since they are failing (expectedly) on the tests related to the `as` and `className` props not being set to the root element; `as` and `classname` are set to the inner input element.

In oder to set them to the root element, one needs to use the `wrapper` slot. This is a different paradigm for components that accept `wrapper` props and we have to decide if it adheres to Stardust principles and if it does, we need to write dedicated UTs for components acception `wrapper`.

screen shot 2018-10-12 at 16 55 59

how should we proceed here?
@levithomason @kuzhelov @miroslavstastny

kuzhelov:
Just as an alternative, to have a full picture, let me provide what we would have with this component being split on two:
- `Input`, the basic one, is direct substitute for DOM `<input />`

image



- `InputField` - more complex one, will provide `icon` and `clearable` (and more, if would be necessary) functionality on top of the `Input`
  - this component should expose `input` as a slot

```jsx
<InputField ... input={{ ... }}>
```

image


---------------

With that we will solve the following problems:
- each of these components will be safe to use by the client, as **each will fully support the set of props exposed**
- we won't introduce `wrapper` semantics to the client which could be a source for confusion
  - for `Input` styles (approach of merged component) we now don't have `root` slot - it is now `input`, `wrapper`, ... slots:
```js
// input styles
{
  wrapper: (...) => { .. wrapper styles .. }
  input: (...) => { .. input styles .. }
  /* no 'root' anymore */
 ...
}
```
This is inconsistent with most other components where `root` slot is provided - and it could result in **client's confusion about why styles are not properly applied**. **This point of concern is valid for accessibility domain as well.**

levithomason:
Consensus for now, do not add special logic for `wrapper={false}`.  If it becomes an issue or request later, we may support it and only render the non-wrapped input.  Features like icon and clearable would not work when not wrapped.

bmdalex:
finally ready for final review

@kuzhelov @mnajdova @levithomason

package.json
```diff
@@ -97,6 +97,7 @@
   },
   "devDependencies": {
     "@types/classnames": "^2.2.4",
+    "@types/enzyme": "^3.1.14",
```
mnajdova, line 100: Is this intentional?

bmdalex: yes, these typings are very useful in the tests for autocomplete the API suggestions

src/components/Slot/Slot.tsx
```diff
@@ -43,7 +55,10 @@ class Slot extends UIComponent<Extendable<ISlotProps>, any> {
     as: 'div',
   }
 
-  renderComponent({ ElementType, classes, rest }) {
+  static create = createSlotFactory(Slot.defaultProps.as, content => ({ content }))
+  static createHTMLInput = createSlotFactory('input', type => ({ type }))
```
mnajdova, line 59: Is there really need for creating this method? If we start creating this for the input, then we will continue creating one for all elements. It is pretty easily used generic.

src/components/Input/Input.tsx
```diff
@@ -104,120 +110,123 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
...
+
+    return Slot.create(wrapper, {
+      defaultProps: {
+        as: ElementType,
+        className: cx(Input.className, className),
+        styles: styles.root,
+        ...rest,
+      },
+      overrideProps: {
+        children: (
```
mnajdova, line 161: Why are we always overriding the wrapper's children? How will we allow the user to add custom children if they want? Or we don't support children API here?

src/components/Input/Input.tsx
```diff
@@ -104,120 +110,123 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
...
   static defaultProps = {
     as: 'div',
     type: 'text',
+    wrapper: 'div',
   }
 
   static autoControlledProps = ['value']
 
-  inputRef: any
+  state = { value: this.props.value || this.props.defaultValue || '' }
```
mnajdova: Shouldn't AutoControlled component handle this?

docs/src/prototypes/chatPane/chatPaneHeader.tsx
```diff
@@ -87,7 +86,7 @@ class ChatPaneHeader extends React.PureComponent<IChatPaneHeaderProps> {
             styles={{
               fontWeight: 100,
               ...(!index && { marginRight: '1.6rem' }),
-              marginTop: pxToRem(8),
+              marginTop: '8px',
```
mnajdova: Just out of curiosity what is the need of this? Is the result some decimal number?

docs/src/examples/components/Input/Variations/InputExampleTargeting.shorthand.tsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react'
+import { Grid, Input, Text } from '@stardust-ui/react'
+
+const inputStyles = { color: 'blue', background: 'yellow' }
+const wrapperStyles = { padding: '5px', background: 'red' }
+const partitionedProps = {
```
hughreeling: Not sure that partitioned is right word here.  What exact;y are we trying to express?

src/components/Input/Input.tsx
```diff
@@ -72,6 +78,9 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
     /** Optional Icon to display inside the Input. */
     icon: customPropTypes.itemShorthand,
 
+    /** Shorthand for the input component */
```
hughreeling: nit - end with '.'

src/components/Input/Input.tsx
```diff
@@ -104,120 +110,123 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
...
+
+  renderComponent({
+    ElementType,
+    classes,
+    rest: restProps,
+    styles,
+    variables,
+  }: IRenderResultConfig<IInputProps>) {
+    const { className, input, renderIcon, renderInput, renderWrapper, type, wrapper } = this.props
+    const { value = '' } = this.state
```
hughreeling, line 154: Do we need default value - aren't we doing already in 

state = { value: this.props.value || this.props.defaultValue || '' }

src/components/Input/Input.tsx
```diff
@@ -104,120 +110,123 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
...
+        ...rest,
+      },
+      overrideProps: {
+        children: (
+          <>
+            <Ref innerRef={this.handleInputRef}>
+              {Slot.createHTMLInput(input || type, {
+                defaultProps: {
+                  ...htmlInputProps,
+                  type,
```
hughreeling, line 171: Do we need 'type' here too?

src/components/Input/Input.tsx
bmdalex: for an input the children API does not make sense;
we can definitely pass children as defaultProps but then **we will open up for the users scenarios where they can do:**

```jsx
<Input placeholder="Search..." />
<Input placeholder="Search..." wrapper={{ children: <Text content='random text instead of input' /> }} />
<Input placeholder="Search...">
  <input placeholder="Override haha" />
</Input>
<Input placeholder="Search...">
  <span>random span instead of input</span>
</Input>
```
**that would render:**
screen shot 2018-10-17 at 15 05 52

**by using `overrideProps` to pass `children` we do not allow the weird scenarios above and for the same inputs the rendered elements are:**
screen shot 2018-10-17 at 15 07 24

src/components/Input/Input.tsx
bmdalex: good catch, it was copy/paste from previous implementation, removed

src/components/Slot/Slot.tsx
bmdalex: we're trying to keep consistency with the methods in `factories.tsx`; also, I think this is valuable because it's always this method we should use for creating an input same for the ones in `factories.tsx`

```tsx
// ============================================================
// HTML Factories
// ============================================================
export const createHTMLDivision = createShorthandFactory('div', val => ({ children: val }))
export const createHTMLIframe = createShorthandFactory('iframe', src => ({ src }))
export const createHTMLImage = createShorthandFactory('img', val => ({ src: val }))
export const createHTMLInput = createShorthandFactory('input', val => ({ type: val }))
export const createHTMLLabel = createShorthandFactory('label', val => ({ children: val }))
export const createHTMLParagraph = createShorthandFactory('p', val => ({ children: val }))
```

that will probably appear in `Slot.tsx` as:
```tsx
// ============================================================
// HTML Slot Factories
// ============================================================
static createHTMLDivision = createSlotFactory('div', children => ({ children }))
static createHTMLIframe = createSlotFactory('iframe', src => ({ src }))
static createHTMLImage = createSlotFactory('img', src => ({ src }))
static createHTMLLabel = createSlotFactory('label', children => ({ children }))
static createHTMLParagraph = createSlotFactory('p'', children => ({ children }))
```

docs/src/prototypes/chatPane/chatPaneHeader.tsx
bmdalex: the space was not correct in initial implementation + consistency in the prototypes, the devs using them might don't have access to the pxToRem utility

docs/src/examples/components/Input/Variations/InputExampleInputSlot.shorthand.tsx
```diff
@@ -0,0 +1,24 @@
...
+const inputStyles = { color: 'blue', background: 'yellow' }
+const inputOverrides = { placeholder: 'Placeholder Override...', role: 'checkbox' }
+
+const InputExampleInputSlot = () => (
+  <Grid columns="1fr 2fr" styles={{ justifyItems: 'start', alignItems: 'center', gap: '10px' }}>
+    <Text content="Input default:" />
+    <Input {...inputProps} />
+
+    <Text content="Input with input slot as props:" />
+    <Input {...inputProps} input={{ ...inputOverrides, styles: inputStyles }} />
```
kuzhelov: would suggest to inline props of the `Input` component for each example. Also, for the `input` slot would suggest to do the same - from those that are extracted as an object we might think of only `inputStyles` being intuitive for the client, as this is a general concept for all Stardust component. All other props are `Input` specific.

```jsx
 <Input {...inputProps} />
...
<Input placeholder='Search...' role='presentation' input={{ 
    // override component's 'placeholder' value
    placeholder='Placeholder Override...'   
   
    // override component's 'role' value
    role='checkbox'   
    
    // set custom styles for input DOM element
    styles: customInputStyles 
}} />
```

docs/src/examples/components/Input/Variations/InputExampleTargeting.shorthand.tsx
kuzhelov: actually, to me it seems that we could simplify this example a bit - there are quite a lot of props involved, but in our examples we should rather target specific aspect of the component - and, thus, we should use the least amount of properties necessary for that. Frankly, would expect that even for Stardust developer it might be not immediately visible what is the goal solved by that

docs/src/prototypes/chatPane/chatPaneHeader.tsx
kuzhelov: not sure if we should advertise the pixel-based values in prototypes. The question we need to ask ourselves is the following - **which units would be correct here, pixels or rems**? It seems that `rems` would be a better answer, given that
-  all Stardust components scales in accordance to `rem` units
- thus, client's intent here would really to introduce `rem`-based value here, so this indent would scale with the component being scaled
- as a result, I would expect that if not the `pxToRem`, but at least `rem`-based value will be specified here

docs/src/prototypes/chatPane/composeMessage.tsx
```diff
@@ -21,6 +21,7 @@ class ComposeMessage extends React.Component {
       <Input
         fluid
         placeholder="Type a message"
+        input={{ styles: { height: '44px' } }}
```
kuzhelov: same considerations about `px vs rem` units (https://github.com/stardust-ui/react/pull/326/files#r225954685) apply here as well

src/components/Input/Input.tsx
```diff
@@ -1,54 +1,60 @@
...
+import Slot from '../Slot'
+import Ref from '../Ref'
 
 export interface IInputProps {
   as?: any
-  children?: ReactChildren
   className?: string
   clearable?: boolean
-  defaultValue?: string | number
+  defaultValue?: React.ReactText
```
kuzhelov, line 27: I would vote for leaving it as it was before, because of its explicitness

docs/src/examples/components/Input/Variations/InputExampleTargeting.shorthand.tsx
bmdalex: the fact that these props can go either to the wrapping `div` (`dir`, `tabIndex`, `styles`)  or the `input` element (`placeholder`, `id`, `role`) and the examples suggest how clients can use `wrapper` and `input` slots to target the wrapping `div` and the `input` element, respectively to set props that are sent by default to the other slot; I pushed a fix trying to make the example more clear.

src/components/Input/Input.tsx
bmdalex: done

src/components/Input/Input.tsx
bmdalex: I had the same thought, but I get this weird warning in the tests without the default value:

> Warning: A component is changing a controlled input of type text to be uncontrolled. Input elements should not switch from controlled to uncontrolled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
        in input (created by FelaTheme)
        in FelaTheme (created by Slot)
        in Slot (created by FelaTheme)
        in Ref (created by FelaTheme)
        in div (created by FelaTheme)
        in FelaTheme (created by Slot)

I think it might be a bug in the `AutoControlledComponent` class. Therefore I'd rather keep it even though it definitely looks redundant

src/components/Input/Input.tsx
bmdalex: yes, input needs a default type in case it is not set by the user in the input slot;
e.g.:
```
<Input placeholder="Search..." input={{ placeholder: 'asdsa' }} />
```
would render:
```html
<div class="ui-slot ui-input">
  <input placeholder="asdsa" value="" class="ui-slot" />
</div>
```
instead of
```html
<div class="ui-slot ui-input">
  <input type="text" placeholder="asdsa" value="" class="ui-slot" />
</div>
```

docs/src/prototypes/chatPane/chatPaneHeader.tsx
bmdalex: fixed by providing margin auto

docs/src/prototypes/chatPane/composeMessage.tsx
bmdalex: that would be `3.1429rem` are we sure it's better to have this value?

src/components/Input/Input.tsx
bmdalex: we'd duplicate this `string | number` type everywhere in the file (3 places); I vote for consistency over explicitness if u are not really against it

docs/src/examples/components/Input/Variations/InputExampleInputSlot.shorthand.tsx
bmdalex: done

hughreeling (approved):

Bugaa92:
I have a [ci/circleci failure](https://circleci.com/gh/stardust-ui/react/2628?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) that I don't understand.

@kuzhelov 
@mnajdova 
@miroslavstastny

src/components/Input/Input.tsx
bmdalex: discussed with @kuzhelov and came to the point that we should keep the consistency of the API and let the user pass `children` if they want; so I'm moving children to `defaultProps`;

no need to handle this with case `childrenExist` utility in the code as it's a side effect of the `Slot.create` method

mnajdova (approved):

docs/src/examples/components/Input/Variations/InputExampleClearable.shorthand.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
 import { Input } from '@stardust-ui/react'
 
-const InputExampleClearableShorthand = () => <Input clearable placeholder="Search..." />
+const InputExampleClearable = () => <Input clearable placeholder="Search..." />
```
kuzhelov, line 4: 👍

docs/src/examples/components/Input/Variations/InputExampleInputSlot.shorthand.tsx
```diff
@@ -0,0 +1,49 @@
...
+        // set custom styles for input DOM element
+        styles: inputStyles,
+      }}
+    />
+
+    <Text content="Wrapped Input with existing component:" />
+    <Input
+      placeholder="Search..."
+      role="presentation"
+      input={
```
kuzhelov, line 30: 👍

docs/src/examples/components/Input/Variations/InputExampleTargeting.shorthand.tsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import { Grid, Input, Text } from '@stardust-ui/react'
+
+const propsForInput = { placeholder: 'Search...', id: 'inputId', role: 'checkbox' }
```
kuzhelov, line 5: nit: just to make the idea more visible, I would suggest to introduce comment like `this object contains properties that will be applied to the input`. But another point that I would like to make, just to share my thoughts on this - it seems to be really confusing to the client in a sense which prop belongs to which slot, and **we would need to not forget to maintain this example** with each change to this decision being made, as well as with each new prop being introduced. Frankly, to me it seems like a serious problem we should further think on (another PR, of course)

docs/src/examples/components/Input/Variations/InputExampleWrapperSlot.shorthand.tsx
```diff
@@ -0,0 +1,52 @@
...
+    />
+
+    <Text content="Wrapped Input with props:" />
+    <Input
+      placeholder="Search..."
+      dir="ltr"
+      tabIndex={2}
+      styles={{ color: 'blue', backgroundColor: 'yellow' }}
+      wrapper={{
+        // override component's 'dir' attribute
```
kuzhelov: nit: just to explain from the _result_'s point of view to the client, not the _intent_ - maybe we might consider to use `will override component's 'dir' attribute` form for this comment instead

docs/src/examples/components/Input/Variations/InputExampleWrapperSlot.shorthand.tsx
```diff
@@ -0,0 +1,52 @@
...
+      dir="ltr"
+      tabIndex={2}
+      styles={{ color: 'blue', backgroundColor: 'yellow' }}
+      wrapper={<Text dir="rtl" tabIndex={0} styles={{ padding: '5px', backgroundColor: 'red' }} />}
+    />
+
+    <Text content="Wrapped Input with custom element:" />
+    <Input
+      placeholder="Search..."
+      dir="ltr"
```
kuzhelov: just for the sake of focusing on more common stuff for these scenarios - maybe it would be fine if we'll remove `dir` from the example's scope?

docs/src/examples/components/Input/Variations/index.tsx
```diff
@@ -29,6 +29,26 @@ const Variations = () => (
       description="An input can be used inline with text."
       examplePath="components/Input/Variations/InputExampleInline"
     />
+    <ComponentExample
+      title="Inline, clearable with icon"
+      description="An input with a given icon can be clearable and can be used inline with text."
```
kuzhelov: simplify to `An input can be clearable, with icon and inlined into text.`?

docs/src/examples/components/RadioGroup/Types/RadioGroupExample.shorthand.tsx
```diff
@@ -34,7 +34,8 @@ class RadioGroupVerticalExample extends React.Component {
         key: 'Custom',
         label: (
           <Text>
-            Choose your own <Input inline placeholder="flavour" />
+            Choose your own{' '}
+            <Input input={{ tabIndex: this.state.inputTabIndex }} inline placeholder="flavour" />
```
kuzhelov, line 38: 👍

docs/src/prototypes/chatPane/composeMessage.tsx
kuzhelov: yes, as it will correctly apply scaling behavior to this component. At the same time, we might consider to round this value to more readable one, like `3.15rem` - or provide the one you've suggested with comment that it corresponds to 44px for default font size (`16px`)

src/components/Input/Input.tsx
```diff
@@ -104,120 +110,119 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
...
+          </>
+        ),
+        styles: styles.root,
+        ...rest,
+      },
+      render: renderWrapper,
+    })
+  }
+
+  private handleInputRef = (c: HTMLInputElement) => (this.inputRef = c)
```
kuzhelov: nit: this could be safely inlined. Also, it would be better to use more meaningful name for function argument :)

```jsx
 <Ref innerRef={inputDomElement => (this.inputDomElement = inputDomElement)}>
```

src/components/Input/Input.tsx
```diff
@@ -1,54 +1,60 @@
...
  * For good screen reader experience set aria-label or aria-labelledby attribute for input.
  *
- *
  * Other considerations:
  *  - if input is search, then use "role='search'"
- *
  */
-class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
+class Input extends AutoControlledComponent<Extendable<IInputProps>, IInputState> {
+  private inputRef: HTMLInputElement
```
kuzhelov: this is not ref, name like `inputElement` or, better, `inputDomElement` would be more descriptive here

src/components/Input/Input.tsx
```diff
@@ -104,120 +110,119 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
...
-            defaultProps: {
-              styles: styles.icon,
-              variables: variables.icon,
-            },
-            overrideProps: this.handleIconOverrides,
-            render: renderIcon,
-          })}
-      </ElementType>
-    )
+    return icon || null
```
kuzhelov, line 227: here we will ignore any overrides provided to icon slot in case we have `clearable` input with non-empty value - this is not what client would expect, especially given that there might be no `close` icon provided by user's theme

src/index.ts
```diff
@@ -5,29 +5,23 @@ export { themes }
 export { default as Accordion } from './components/Accordion'
 export { default as Attachment } from './components/Attachment'
 export { default as Avatar } from './components/Avatar'
-export { default as Button } from './components/Button'
-export { ButtonGroup } from './components/Button'
+export { default as Button, ButtonGroup } from './components/Button'
```
kuzhelov, line 8: 👍

src/lib/factories.tsx
```diff
@@ -6,12 +6,9 @@ import {
   ShorthandPrimitive,
   ShorthandRenderFunction,
   ShorthandValue,
+  IProps,
 } from '../../types/utils'
 
-interface IProps {
```
kuzhelov, line 11: 👍

src/themes/teams/components/Input/inputVariables.ts
```diff
@@ -1,21 +1,31 @@
...
-  vars.inputPadding = `${pxToRem(6)} ${pxToRem(24)} ${pxToRem(6)} ${pxToRem(12)}`
-  vars.inputFocusBorderColor = siteVars.brand
-  vars.inputFocusBorderRadius = `${pxToRem(3)} ${pxToRem(3)} ${pxToRem(2)} ${pxToRem(2)}`
+  fontColor: siteVars.bodyColor,
+  fontSize: siteVars.fontSizes.medium,
 
-  vars.iconPosition = 'absolute'
-  vars.iconRight = `${pxToRem(2)}`
+  iconPosition: 'absolute',
+  iconRight: pxToRem(2),
```
kuzhelov: `px2asRem`?

src/themes/teams/components/Input/inputVariables.ts
```diff
@@ -1,21 +1,31 @@
...
+  inputFocusBorderRadius: string
+}
 
-export default (siteVars: any) => {
-  const vars: any = {}
-
-  vars.borderRadius = `${pxToRem(3)}`
-  vars.borderBottom = `${pxToRem(2)} solid transparent`
-  vars.backgroundColor = siteVars.gray10
+const [px2asRem, px3asRem, px6asRem, px12asRem, px24asRem] = [2, 3, 6, 12, 24].map(v => pxToRem(v))
```
kuzhelov: maybe the following naming pattern will result in more readable strings at the place of usage:

```js
px2asRem -> _2px_asRem

///
inputFocusBorderRadius: `${px3asRem} ${px3asRem} ${px2asRem} ${px2asRem}`,

// vs
inputFocusBorderRadius: `${_3px_asRem} ${_3px_asRem} ${_2px_asRem} ${_2px_asRem}`,
```

test/specs/commonTests/implementsShorthandProp.tsx
```diff
@@ -18,34 +18,47 @@ export default Component => {
     const { mapsValueToProp } = options
     const { displayName } = ShorthandComponent
 
+    /**
+     * @param {ReactWrapper} wrapper [mounted Component]
+     * @param {(n: ReactWrapper) => boolean} propsCheckFn [boolean function that applies a boolean check on a mounted ShorthandComponent]
+     * return true if the ShorthandComponent that matches propsCheckFn is found
```
kuzhelov: `return` -> `@returns`

test/specs/commonTests/implementsShorthandProp.tsx
```diff
@@ -18,34 +18,47 @@ export default Component => {
     const { mapsValueToProp } = options
     const { displayName } = ShorthandComponent
 
+    /**
+     * @param {ReactWrapper} wrapper [mounted Component]
```
kuzhelov: lets use `-` as a separator - for the sake of being consistent with other places where JS doc is used

```js
/**
 * @param {ReactWrapper} wrapper - mounted Component
```

test/specs/commonTests/implementsWrapperProp.tsx
```diff
@@ -0,0 +1,60 @@
...
+) => {
+  const { wrapppedComponentSelector, wrappperComponentSelector = Slot.defaultProps.as } = options
+
+  const wrapperTests = (wrapper: ReactWrapper) => {
+    expect(wrapper.length).toBeGreaterThan(0)
+    expect(wrapper.find(wrapppedComponentSelector).length).toBeGreaterThan(0)
+  }
+
+  describe('"wrapper" prop', () => {
+    it('does not wrap the component when set to false', () => {
```
kuzhelov: we should eliminate this test, as we've agreed to not support this scenario for now (`element should always be wrapped`)

test/specs/commonTests/implementsWrapperProp.tsx
```diff
@@ -0,0 +1,60 @@
...
+    it('wraps the component by default', () => {
+      wrapperTests(mount(<Component />).find(wrappperComponentSelector))
+    })
+
+    it('wraps the component with a custom element', () => {
+      wrapperTests(mount(<Component wrapper={<span />} />).find('span'))
+    })
+
+    it('wraps the component with a custom element using "as" prop', () => {
+      const customElem = 'p'
```
kuzhelov: this could be inlined

test/specs/commonTests/implementsWrapperProp.tsx
```diff
@@ -0,0 +1,60 @@
...
+      wrapperTests(mount(<Component wrapper={{ as: customElem }} />).find(customElem))
+    })
+
+    it('wraps the component with another stardust component as wrapper', () => {
```
kuzhelov: this is, essentially, the same test that is done for `<span/>` case - in both cases you are providing just React element, and what you are testing is how React element handled by the wrapper. It seems redundant to have these two at the same time

test/specs/commonTests/implementsWrapperProp.tsx
```diff
@@ -0,0 +1,60 @@
...
+      wrapperTests(mount(<Component wrapper={<Segment />} />).find(Segment))
+    })
+
+    it('wraps the component with a custom component as wrapper', () => {
```
kuzhelov: > this is, essentially, the same test that is done for <span/> case - in both cases you are providing just React element, and what you are testing is how React element handled by the wrapper. It seems redundant to have these two at the same time

the same thought applies here as well

test/specs/commonTests/implementsShorthandProp.tsx
```diff
@@ -18,34 +18,47 @@ export default Component => {
...
+     * @param {(n: ReactWrapper) => boolean} propsCheckFn [boolean function that applies a boolean check on a mounted ShorthandComponent]
+     * return true if the ShorthandComponent that matches propsCheckFn is found
+     */
+    const testShorthandComponentProps = (
+      wrapper: ReactWrapper,
+      propsCheckFn: (n: ReactWrapper) => boolean,
```
kuzhelov: it is hard to reason about what does `n` stands for with only its name being given :)

src/components/Input/Input.tsx
```diff
@@ -104,120 +110,119 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
...
+    type: PropTypes.string,
+
     /** The value of the input. */
     value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
 
     /** Override for theme site variables to allow modifications of component styling via themes. */
     variables: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
+
+    /** Shorthand for the wrapper component. */
+    wrapper: customPropTypes.itemShorthand,
```
kuzhelov: note that now, with this Prop type being defined in a way it is, we have the following error in console output when component is rendered:

> Failed prop type: Prop `wrapper` in `Input` conflicts with props: `children`. They cannot be defined together, choose one or the other.

test/specs/commonTests/implementsShorthandProp.tsx
```diff
@@ -18,34 +18,47 @@ export default Component => {
     const { mapsValueToProp } = options
     const { displayName } = ShorthandComponent
 
+    /**
+     * @param {ReactWrapper} wrapper [mounted Component]
+     * @param {(n: ReactWrapper) => boolean} propsCheckFn [boolean function that applies a boolean check on a mounted ShorthandComponent]
+     * return true if the ShorthandComponent that matches propsCheckFn is found
+     */
+    const testShorthandComponentProps = (
```
kuzhelov: this method is only used to check the following thing - essentially, to check that 
- there is a component in the rendered tree **of specific type**
- that has specific set of props being set
- and there is only one such component rendered

What if we would omit the predicate (that introduces generic use to the function, as it is not needed), and would just do it the following way

```ts
const expectContainsSingleShorthandElement = (wrapper, withProps) => ... // use ShorthandType here
```

Then it would be possible to use this function the following way for the test cases provided
```jsx
 test(`string value is handled as ${displayName}'s ${mapsValueToProp}`, () => {
   ...
   const shorthandProps = ...
   expectContainsSingleShorthandElement(
      mount(<Component {...props} />), 
      shorthandProps
   )
}
```

docs/src/examples/components/Input/Variations/InputExampleTargeting.shorthand.tsx
bmdalex: done with the comment, pls suggest ways to tackles this so we can track it in a VSTS bug / issue

docs/src/examples/components/Input/Variations/InputExampleWrapperSlot.shorthand.tsx
bmdalex: hmm.. that test was your suggestion from yesterday's comment.. added the "will " in front of text; please make it more clear next time so we save up some time..

docs/src/examples/components/Input/Variations/InputExampleWrapperSlot.shorthand.tsx
bmdalex: not sure I understand; this is relevant to point out that `dir` goes to input but we can target the `wrapper` slot; it's also caught by the visual tests so it'd catch regressions; I vote for keeping it unless u have strong arguments on how this can be dangerous/misleading

docs/src/examples/components/Input/Variations/index.tsx
bmdalex: done, thx

docs/src/prototypes/chatPane/composeMessage.tsx
bmdalex: ok, done, thx

src/components/Input/Input.tsx
bmdalex: I cannot see this error; can you add repro steps?

src/components/Input/Input.tsx
bmdalex: done

_Note: This code didn't change from initial review. We should avoid starting new conversations in the phase of addressing comments unless that code changed_

src/components/Input/Input.tsx
bmdalex: _Note: This is not in the scope of this PR and we can open separate issue if needed after having this conversation_

> here we will ignore any overrides provided to icon slot in case we have clearable input with non-empty value - this is not what client would expect

It actually is what the client expects if he provides `clearable` and `icon`; for me that's just very intuitive; do you have any examples to suggest otherwise?

> especially given that there might be no close icon provided by user's theme

this is a good point, we can discuss it in a separate issue; it'd be great if you can open one

test/specs/components/Slot/Slot-test.ts
```diff
@@ -1,6 +1,69 @@
...
-  isConformant(Slot, { exportedAtTopLevel: false })
+  const createSlot = (factoryFn: Function, val, options?) =>
+    mount(factoryFn(val, options)).find(Slot)
+
+  describe('is conformant', () => {
+    isConformant(Slot, { exportedAtTopLevel: false })
+  })
+
+  describe('createSlotFactory', () => {
+    it('calls createShorthand from lib', () => {
```
kuzhelov: this is about implementation details - we should avoid testing this

test/specs/components/Slot/Slot-test.ts
```diff
@@ -1,6 +1,69 @@
...
+        mapValueToProp,
+        value,
+        expect.any(Object),
+      )
+    })
+
+    it('sets correct "as" prop in defaultProps', () => {
+      const as = 'span'
+      const options = { testOption: 'test option value' }
+      const createShorthandSpy = spyOn(lib, 'createShorthand')
```
kuzhelov: here we are checking the implementation details - we should rather check method's result instead

src/themes/teams/components/Input/inputVariables.ts
bmdalex: I think it's good, done

_Note: This code didn't change from initial review. We should avoid starting new conversations in the phase of addressing comments unless that code changed_

test/specs/components/Slot/Slot-test.ts
```diff
@@ -1,6 +1,69 @@
...
+      expect(optionsArg).toEqual({ defaultProps: { as }, ...options })
+    })
+
+    it('overrides "as" prop in defaultProps', () => {
+      const as = 'span'
+      const asOverride = 'p'
+      const createShorthandSpy = spyOn(lib, 'createShorthand')
+      createSlotFactory(as, () => ({}))('testValue', { defaultProps: { as: asOverride } }) // clone of the options
+
+      const optionsArg = createShorthandSpy.calls.mostRecent().args[3]
```
kuzhelov: this test is very brittle - if the next day we will check the arguments that `createShorthand` accepts, this will fail this test - while this failure won't surely guarantee that the logic of `createSlotFactory` has become to be incorrect

src/themes/teams/components/Input/inputVariables.ts
bmdalex: nice catch, thx

test/specs/commonTests/implementsShorthandProp.tsx
bmdalex: done, thx

test/specs/commonTests/implementsShorthandProp.tsx
bmdalex: I was sleeping I thing (facepalm); thx, nice catch

test/specs/components/Slot/Slot-test.ts
```diff
@@ -1,6 +1,69 @@
...
+import { mount } from 'enzyme'
+
+import * as lib from 'src/lib'
 import Slot from 'src/components/Slot'
+import { createSlotFactory } from 'src/components/Slot/Slot'
+import { isConformant } from 'test/specs/commonTests'
 
 describe('Slot', () => {
-  isConformant(Slot, { exportedAtTopLevel: false })
+  const createSlot = (factoryFn: Function, val, options?) =>
```
kuzhelov, line 7: lets move it to `describe` block where it is used

test/specs/components/Slot/Slot-test.ts
bmdalex: agreed with @kuzhelov  to add the tests in the dedicated PR for Slot factory that's about to come

test/specs/components/Slot/Slot-test.ts
bmdalex: agreed with @kuzhelov  to add the tests in the dedicated PR for Slot factory that's about to come

test/specs/components/Slot/Slot-test.ts
bmdalex: agreed with @kuzhelov  to add the tests in the dedicated PR for Slot factory that's about to come

test/specs/components/Slot/Slot-test.ts
bmdalex: already there

test/specs/commonTests/implementsShorthandProp.tsx
bmdalex: not sure about this, the 2nd usage is different, let's discuss

test/specs/commonTests/implementsShorthandProp.tsx
bmdalex: `n` -> `node` is this ok?

_Note: This code didn't change from initial review. We should avoid starting new conversations in the phase of addressing comments unless that code changed_

test/specs/commonTests/implementsWrapperProp.tsx
bmdalex: we agreed we won't add any condition to prevent the user from passing `wrapper={false}`;
this is just testing what will happen **by default** (not handled) if the user sets the wrapper to false

test/specs/commonTests/implementsWrapperProp.tsx
bmdalex: I'm reusing it in the expectation

test/specs/commonTests/implementsWrapperProp.tsx
bmdalex: ok, removed

test/specs/commonTests/implementsWrapperProp.tsx
bmdalex: ok, removed

src/components/Input/Input.tsx
bmdalex: > this is not ref

not sure about that; from react typings:
```tsx
refs: {
  [key: string]: ReactInstance
};

type ReactInstance = Component<any> | Element;

interface Element { }

interface HTMLElement extends Element { }
```
Also, this is not in the scope of the PR, I'm just adding the type. Made the change as requested anyways.

_Note: This code didn't change from initial review. We should avoid starting new conversations in the phase of addressing comments unless that code changed_

bmdalex (commented):
addressed @kuzhelov 's comments

_Note: Please provide all the initial feedback on the first review so I can address it from beginning. It'd be better not to start new conversations in the phase of addressing review comment (after initial code review from specific dev) unless there are code changes in that area, of course_

docs/src/examples/components/Input/Variations/InputExampleWrapperSlot.shorthand.tsx
bmdalex: agreed to remove `dir`

test/specs/commonTests/implementsWrapperProp.tsx
bmdalex: agreed to remove the test

test/specs/commonTests/implementsWrapperProp.tsx
bmdalex: agreed to inline

src/components/Input/Input.tsx
bmdalex: fixing

test/specs/commonTests/implementsShorthandProp.tsx
bmdalex: will address

kuzhelov (approved):

test/specs/commonTests/implementsShorthandProp.tsx
bmdalex: N/A anymore

test/specs/commonTests/implementsWrapperProp.tsx
bmdalex: done

src/components/Input/Input.tsx
bmdalex: done

PR #327, opened 2018-10-05
fix(divider): use primary/secondary prop instead of type prop

codepretty:
# API Update

## Removing type prop
Instead of using the prop _type_, we want to be able to use the prop _primary_or _secondary_

**Before**
```jsx
<Divider type="primary" />
<Divider type="secondary" />
```

**Now**
```jsx
<Divider primary />
<Divider secondary />
```


alinais (changes requested):
Don't we need to preserve the styles for the secondary type?

kuzhelov (changes requested):
First thing to note is that I totally understand the motivation for this change. At the same time, **if we would like to preserve both `primary` and `secondary` variants**, we should notice that these two are mutually exclusive ones for the component. It was an intentional design decision made that 

> for all mutually exclusive values we would use single enum prop, to prevent accidental collisions of multiple values provided for the same component

So, here I see the following ways to achieve the goal of this PR (to make API more convenient):
- leave things as they are for Stardust component, introduce wrapper for Divider in Teams (TeamsDivider) and translate its `primary` flag prop to Divider's `type`:
```jsx
const TeamsDivider = ({ primary }) => <Divider {...(primary && { type: 'primary' }) }>

// and later use it like:
<TeamsDivider primary />
```
- agree that Stardust will expose only `primary` **for all components** - this is much serious design decision where we should agree that `secondary` has semantics of `default` for majority of broad set of use cases (not only Teams-specific)

src/components/Divider/Divider.tsx
```diff
@@ -46,8 +47,11 @@ class Divider extends UIComponent<Extendable<IDividerProps>, any> {
     /** Size multiplier (default 0) * */
     size: PropTypes.number,
 
-    /** A Divider can be formatted to show different levels of emphasis. */
-    type: PropTypes.oneOf(['primary', 'secondary']),
+    /** A Divider can appear to show a primary level of emphasis. */
+    primary: PropTypes.bool,
```
mnajdova, line 51: If we even decide to go with introducing these props, we should make sure that those are not going to be used together. In order to achieve this, please use the `customPropTypes.disallow` in the definition of the properties accordingly.

kuzhelov: this would be definitely a good step, but would like to make it clear - it won't result in 'disallowing these props to be used together', but will just introduce a error message printed to console in case if this will happen. The problem is that, actually, at the API level we _cannot_ prevent these two to be specified at the same time.

codepretty:
It doesn't really makes sense to have a primary/secondary divider. We should really just have a color prop.

PR #330, opened 2018-10-06
chore(docs): allow to use .mdx format in docs

layershifter:
I made similar thing for SUIR recently, now I want to ship it there and and simplify your life :cat: 

### Pros

* much simplier format, you can compare how changed _Quick Start_ page, this will simplify editing of doc pages
* formatting will be supported in the next release of Prettier

### Cons

* Babel is there again. I couldn't make `awesome-typescript-loader` or even `ts-loader` work, it always fails with https://github.com/s-panferov/awesome-typescript-loader/issues/561


levithomason:
We need babel to wrap prop types again, strip debug statements from prod builds, and generate handled props.  So I think this is inevitable.

I like the idea of easily editable docs.  However, we also loose typings / highlighting in doc pages by doing this.  In comparing the added dependencies/complexity to use markdown in docs compared to how many doc edits we get, I'm not sure the tradeoff is worth it.  I think I'd prefer less deps and a simpler build opposed to easier contributing path.  Why?  I don't think we'll have that many contributions compared to our own wrangling of the extra code and logic.

Feedback?

layershifter:
> However, we also loose typings / highlighting in doc pages by doing this. In comparing the added dependencies/complexity to use markdown in docs compared to how many doc edits we get, I'm not sure the tradeoff is worth it.

It's not usual markdown, this is mdx format.

Formatting will be support by Prettier in the next release, VSCode has already a plugin (https://github.com/mdx-js/mdx/issues/119), I'm think that Webstorm will support it soon, too.

hughreeling:
https://www.docz.site/

layershifter:
MDX support by `prettier@1.15` (is not released yet):
https://deploy-preview-5296--prettier.netlify.com/blog/2018/10/00/1.15.0.html
https://github.com/prettier/prettier/pull/4975

----

@hughreeling I dont tried Docz, it may be an option.

layershifter:
I've updated `prettier` and resolved merge conflicts.

layershifter:
We will definitely go with a prebuilt solution, most of them support MDX out of the box.

PR #331, opened 2018-10-08
feat(proto): Meeting options page

mnajdova:
# Meeting options prototype

This PR create a prototype for a Meeting options page inspired from the Microsoft Teams application.

Here is how the prototype looks like:

image

The page consists of three main components:
- `MSTeamsHeader`
- `MSTeamsMeetingOptions`
- `MSTeamsFooter`

Additional components created for the prototype are:
- `MSTeamsLogo` - component that receives an icon shorthand and a company name for generating a logo
- `MSTeamsDivider` - component that is wrapper around the stardust `Divider` and introduces the transparent prop for setting the `dividerColor` variable to `transparent`
- `MSTeamsLink` - general components for creating MSTeams links (as the ones in the footer)

Found issues:
1. The Layout alignment options are limit. The component doesn't allow the user to customize the content to appear in the middle or the main slot to have some user defined width without applying some styles. If the user himself apply these styles, then there is no need for using the Layout component at all.
2. We need to come up with better mechanism to integrate user's styles into theme (so that `siteVariables` and, maybe, `variables` will be consumed)


docs/src/prototypes/meetingOptions/components/MSTeamsDivider.tsx
```diff
@@ -0,0 +1,10 @@
+import { Divider as StardustUIDivider } from '@stardust-ui/react'
+
+export default props => {
```
kuzhelov: would rather name it as `InvisibleDivider` or `TransparentDivider` - this seems to better reflect the idea behind it. Also, we won't need to introduce `transparent` prop with that.

docs/src/prototypes/meetingOptions/components/MSTeamsFooter.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+import { Divider, Layout } from '@stardust-ui/react'
+
+export default props => {
+  return (
+    <div style={{ position: 'fixed', width: 'inherit', bottom: '0px' }}>
```
kuzhelov, line 6: raises couple of concern points (should be mentioned in the PR description maybe):
- how we could introduce these styles as part of theme?
- `Layout` alignment options are limited

docs/src/prototypes/meetingOptions/styles.ts
```diff
@@ -0,0 +1,13 @@
+import { CSSProperties } from 'react'
+
+export const middleColumnStyles = {
```
kuzhelov, line 3: will repeat one of the concerns point - as this is a clear indicator that we need to come up with better mechanism to integrate user's styles into theme (so that `siteVariables` and, maybe, `variables` will be consumed)

docs/src/prototypes/meetingOptions/styles.ts
```diff
@@ -0,0 +1,13 @@
...
+  width: '40%',
+  margin: '0 auto',
+}
+
+export const mainStyle: CSSProperties = {
+  backgroundColor: '#f3f2f1',
+  width: 'calc(100% - 250px)',
```
kuzhelov, line 10: this `250px` definitely should come as one of the value from theme's `siteVariables`

docs/src/prototypes/meetingOptions/components/MSTeamsDivider.tsx
mnajdova: Yep, make sense.

docs/src/prototypes/meetingOptions/components/MSTeamsFooter.tsx
mnajdova: Will mention them.

docs/src/prototypes/meetingOptions/styles.ts
mnajdova: This is actually something regarding just the layout on the docs page (the sidebar). It is not related to the theme.

docs/src/prototypes/meetingOptions/styles.ts
mnajdova: Mention in the PR description

kuzhelov (approved):
👍 we need to be sure these follow-up steps (to address issues discovered) won't be discovered

PR #333, opened 2018-10-08
feat(divider): add fitted prop

gopalgoel19:
# Divider ```fitted``` prop

A divider can be ```fitted``` to render a divider that is simply a line with no space above or below it.

### TODO

- [x] Conformance test
- [x] Stardust base theme
- [x] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [ ] Confirm RTL usage
- [x] [W3 accessibility](https://www.w3.org/standards/webdesign/accessibility) check
- [x] Update the CHANGELOG.md

# API Proposal

## ```fitted```
screenshot 52

## Use cases:

A ```fitted``` divider can be used inside a List component to separate few list items from another.

```jsx
<List>
    <List.Item
      media={<Image src="public/images/avatar/small/matt.jpg" avatar />}
      header="Irving Kuhic"
    />
    <List.Item
      media={<Image src="public/images/avatar/small/steve.jpg" avatar />}
      header="Skyler Parks"
      content="Use the online FTP application to input the multi-byte application!"
    />
    <Divider fitted size={1}/>
    <List.Item
      media={<Image src="public/images/avatar/small/nom.jpg" avatar />}
      header="Dante Schneider"
      content="The GB pixel is down, navigate the virtual interface!"
    />
  </List>
```
The above code give us something like,
screenshot 53


src/themes/teams/components/Divider/dividerStyles.ts
```diff
@@ -25,13 +25,13 @@ const beforeAndAfter = (size, type, variables): ICSSPseudoElementStyle => ({
 
 const dividerStyles: IComponentPartStylesInput<IDividerPropsWithDefaults, any> = {
   root: ({ props, variables }): ICSSInJSStyle => {
-    const { children, size, type, important, content } = props
+    const { children, fitted, size, type, important, content } = props
     return {
       color: variables.textColor,
       display: 'flex',
       alignItems: 'center',
-      paddingTop: pxToRem(variables.dividerPadding),
```
codepretty: The default padding should already be `0`, so I think you only need to apply padding for the non-fitted case.

```
...(!fitted && {
    paddingTop: pxToRem(variables.dividerPadding),
    paddingBottom: pxToRem(variables.dividerPadding),
}),
```

gopalgoel19: done

hughreeling:
You should NOT do this:

```
 <List>
     <List.Item
       media={<Image src="public/images/avatar/small/matt.jpg" avatar />}
       header="Irving Kuhic"
     />
     <List.Item
       media={<Image src="public/images/avatar/small/steve.jpg" avatar />}
header="Skyler Parks"
       content="Use the online FTP application to input the multi-byte application!"
     />
     **<Divider fitted size={1}/>**
     <List.Item
       media={<Image src="public/images/avatar/small/nom.jpg" avatar />}
       header="Dante Schneider"
       content="The GB pixel is down, navigate the virtual interface!"
     />
   </List>
 
```

A list should not contain any direct children apart from list items.  This will definitely break things.

gopalgoel19:
@hughreeling What would be the best way to have a similar line in between list items?

Considering what you mentioned above, my present options are:
1. Add a ```border``` to the bottom of the list item when required.
2. Insert a ```Divider``` inside the ```ListItem```.
3. Make a ```ListItem``` serve as a divider itself. E.g. ```<List.Item divider/>``` renders a list item which look and feel like the ```Divider```.
4. Use two separate ```List``` component.

Please let me know if there are other options as well

hughreeling:
Lists definitely need to contain only list items.  (
https://dev.w3.org/html5/html-author/#the-ol-element)

I think the options are for the list item to contain the divider, or to
style the list item.

Currently you could so something like <Divider as='li' />  but I highly
recommend NOT doing this, it's pretty much certain to make accessibility /
keyboard navigation.

hughreeling:
it's pretty much certain to break accessibility / keyboard navigation.

docs/src/examples/components/Divider/Variations/DividerExampleFitted.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
+import { Divider } from '@stardust-ui/react'
+
+const DividerExampleFitted = () => (
```
kuzhelov: this example could be eliminated - as there is no Children API variation for this case

src/components/Divider/Divider.tsx
```diff
@@ -43,6 +43,9 @@ class Divider extends UIComponent<Extendable<IDividerProps>, any> {
     /** Shorthand for primary content. */
     content: customPropTypes.contentShorthand,
 
+    /** A divider can be fitted, without any space above or below it.  */
+    divider: PropTypes.bool,
```
kuzhelov: this should be name as `fitted`. Also, please, make sure that `IDividerProps` reflects the changes you've introduced for the API.

In general, would suggest to make tests in docs (probably, you've done that), but with dev console being opened - in that case you won't let these problems pass your checks :)

kuzhelov (changes requested):
please, address the changes raised in comments before merging

src/components/Divider/Divider.tsx
gopalgoel19: Ah, a silly one. Fixing it right away.

kuzhelov (approved):

PR #244, opened 2018-09-18
Prototype for the Chat People Picker

silviuaavram:
WIP

PR #334, opened 2018-10-08
Proto/multi dropdown

silviuaavram:
This PR contains the changes from https://github.com/stardust-ui/react/pull/244 since it became pretty complicated to rebase master changes. Will update on its progress.

Do not merge yet! It still contains a work-around for Input in order to pass` role="presentation"` to its `<div>` so the other Input examples are at the moment broken.

- [ ] Conformance test
- [ ] Stardust base theme
- [ ] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [ ] Confirm RTL usage
- [ ] Update the CHANGELOG.md


docs/src/prototypes/ChatPeoplePicker/ChatPeoplePickerExample.tsx
```diff
@@ -0,0 +1,164 @@
...
+  return getUnselectedItems(selected).filter(
+    item => !inputValue || item.name.toLowerCase().includes(inputValue.toLowerCase()),
+  )
+}
+
+const PeoplePickerExampleShorthand = () => (
+  <div>
+    <ChatPeoplePicker
+      source={peopleSupplier}
+      styles={{ width: '30rem', margin: '4rem auto 0 4rem' }}
```
kuzhelov: it would be better to apply these styles to outer container (especially `margin`) - the same way it is done for the next `div` section

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePickerExample.tsx
```diff
@@ -0,0 +1,164 @@
...
+      <ul>
+        <li>Providing the autosuggestions when you type something in the text field.</li>
+        <li>Highlight option in dropdown on mouse hover and on keyboard navigation.</li>
+        <li>
+          Ensure keyboard navigation works in the dropdown, by keeping focus on edit text and using
+          aria active descendant attribute.
+        </li>
+        <li>Open/Close dropdown on click and keyboard Enter/Up/Down/Esc.</li>
+        <li>Provide a callback for handling the selection of an element from the list.</li>
+        <li>Append aria roles and other attributes to make it accessible with screen reader.</li>
```
kuzhelov: nit: aria -> ARIA

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePickerExample.tsx
```diff
@@ -0,0 +1,164 @@
...
+      </ul>
+      <p>
+        Downshift provides a children callback in which we can render whatever we want, as long as
```
kuzhelov: it is better to use impersonalized text in docs:  
`Downshift provides a children callback in which we can render whatever we want, as long as we use...` -> `Downshift provides a children callback one is able to introduce any necessary customisations, as long as the following requirements are met: ...`

src/components/Input/Input.tsx
```diff
@@ -219,6 +221,31 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
...
     )
   }
+  */
+
+  renderComponent({ ElementType, classes, styles, variables, rest }) {
+    const { input, onChange } = this.props
+    const inputClasses = classes.input
+
+    return (
+      <ElementType className={classes.root} onChange={onChange} {...rest}>
```
kuzhelov: note that for `onChange` event here we are not providing the callback's signature that is advertised in prop types (namely, `(e, props) => ...`): here we are not providing any component's `props` to the callback

src/components/Input/Input.tsx
```diff
@@ -219,6 +221,31 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, any> {
...
+        {createHTMLInput(input, {
+          overrideProps: {
+            className: inputClasses,
+            ref: this.handleInputRef,
+          },
+        })}
+        {this.computeIcon() &&
```
kuzhelov: note, `renderIcon` support is lost with suggested changes. The same applies to `renderInput`

kuzhelov (changes requested):
very good prototype, lets address comments before merging

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePicker.tsx
```diff
@@ -0,0 +1,360 @@
+import * as React from 'react'
+import Downshift, { DownshiftState, StateChangeOptions } from 'downshift'
+import { Label, Input, Button, Image, MenuItem, List, Provider, Text } from '@stardust-ui/react'
+import * as _ from 'lodash'
+import keyboardKey from 'keyboard-key'
+import { pxToRem } from 'src/lib'
+import { IThemePrepared, ICSSInJSStyle } from 'types/theme'
+
+type TChatPeoplePickerStyles = {
```
mnajdova: Please move the styles in a separate file - chatPeoplePickerStyles.ts

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePicker.tsx
```diff
@@ -0,0 +1,360 @@
...
+    width: '1px',
+    position: 'absolute',
+  },
+}
+
+interface IPeoplePickerProps {
+  source: (
+    inputValue: string,
+    selected: any[],
+  ) => { name: string; image: string; position: string }[]
```
mnajdova, line 13: Is there any reason why the image here is not ShorthandValue? (Not saying that it should be, just out of curiosity) Maybe we should consider making a type out of the result of the source method.

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePicker.tsx
```diff
@@ -0,0 +1,360 @@
...
+            getInputProps,
+            getItemProps,
+            getMenuProps,
+            isOpen,
+            highlightedIndex,
+            selectItemAtIndex,
+            getLabelProps,
+          }) => {
+            return (
+              <div>
```
mnajdova, line 206: Restructure this return statement into multiple methods that you would invoke here so that it can be easier to read. It's pretty long, and having the conditions is hard to read.

src/components/Input/Input.tsx
mnajdova: We can iterate again on the Input once the new PR for it is merged, and can see if there is any other issues. This are just temporary fixes anyway.

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePickerExample.tsx
```diff
@@ -0,0 +1,164 @@
...
+    <ChatPeoplePicker
+      source={peopleSupplier}
+      styles={{ width: '30rem', margin: '4rem auto 0 4rem' }}
+    />,
+    <div style={{ width: '50rem', margin: '1rem auto 1rem 4rem' }}>
+      <h3>Chat People Picker</h3>
+      <h4>Contents</h4>
+      <p>
+        Prototype built with the help of PayPal'a Downshift component and Stardust components.
+        Downshift is responsible for:
```
mnajdova: PayPal'a -> Paypal's

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePickerExample.tsx
```diff
@@ -0,0 +1,164 @@
+import * as React from 'react'
+import ChatPeoplePicker from './ChatPeoplePicker'
+
+const items = [
+  {
```
mnajdova: Adding more items causes the opened element to be cut in the middle of some item. Maybe we should consider having some max heights on the element displaying the items in the dropdown.
image

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePickerExample.tsx
silviuaavram: addressed, you can check again. I moved margin upwards, although I am not sure if I understood your padding related comment

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePickerExample.tsx
silviuaavram: did some changes in the docs. thanks!

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePicker.tsx
silviuaavram: moved the styles, wanted to do that at some point. thanks for pointing it out too!

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePickerExample.tsx
silviuaavram: thank you!

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePicker.tsx
silviuaavram: created some separated render methods. thank you!

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePicker.tsx
silviuaavram: my example is pretty specific in this case, `  {
    name: 'Imperator Furiosa',
    image: 'public/images/avatar/small/veronika.jpg',
    position: 'Boss',
  },`, and maybe we should address this in a future Component PR? where we can make it general, API-like. do you agree @mnajdova ?

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePickerExample.tsx
silviuaavram: added default maxWidth and width. Also the prototype will accept these as props.

levithomason:
@silviuavram Note, if you slightly adjust the PR description then this PR will automatically close the associated issue once it is merged. See here:

https://help.github.com/articles/closing-issues-using-keywords

Associated issue is #388

docs/src/prototypes/ChatPeoplePicker/ChatPeoplePicker.tsx
```diff
@@ -0,0 +1,330 @@
...
+                  content="Add people"
+                  variables={{ backgroundColor: peoplePickerStyles.addPeopleLabel.backgroundColor }}
+                  {...getLabelProps()}
+                />
+                <div style={containerStyles} onClick={this.onContainerClick.bind(this, isOpen)}>
+                  <span aria-live="assertive" style={peoplePickerStyles.ariaLive}>
+                    {this.state.message}
+                  </span>
+                  {this.renderSelected()}
+                  <Input
```
kuzhelov, line 217: this `Input` element can be wrapped by `Ref` that is now exported from Stardust - no need to change `Input` component's API in that case

silviuaavram:
This PR preceded the Dropdown component, by having this prototype as reference. closing as the Dropdown component has been implemented.

PR #335, opened 2018-10-08
[RFC] feat(slot): create generic slot component

bmdalex:
# Slot

This PR contains:
- implementation of generic `Slot` component;
- using it to convert `content` prop to shorthand slots

### TODO

- [x] Conformance test
- [ ] Stardust base theme
- [x] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [x] Confirm RTL usage
- [ ] Update the CHANGELOG.md

# API Proposal

## `content` prop as shorthand
We are now able to use `content` prop as real shorthand for our components if it's needed. Here are a couple of scenarios where this is useful:

### 2. Usage of `content` prop for `Button` component that was transformed to shorthand in this PR:

screen shot 2018-10-08 at 17 29 43

```jsx
<div>
  <Button content="Basic" />
  <Button content="Styled" styles={{ color: 'red', background: 'yellow' }} />
  <Button content={{ content: 'Shorthand styled', styles: { color: 'red', background: 'yellow' }}} />
</div>
```

```html
<div>
    <span class="ui-slot">Basic</span>
  </button>
  <button class="ui-button cf cj" aria-disabled="false">
    <span class="ui-slot">Styled</span>
  </button>
    <span class="ui-slot cf cj">Shorthand styled</span>
  </button>
</div>
```

### DIFF
--- a/docs/src/examples/components/Slot/index.tsx
+++ b/docs/src/examples/components/Slot/index.tsx
@@ -1,0 +1,9 @@
+import React from 'react'
+
+export default () => (
+  <div>
+    A <code>Slot</code> is a basic component (no default styles) used mainly as a way to create
+    shorthand elements as a part of more complex components (e.g.: <code>content</code> prop for{' '}
+    <code>Button</code> and other components)
+  </div>
+)
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -4,6 +4,7 @@ import * as _ from 'lodash'
 
 import { UIComponent, childrenExist, customPropTypes, createShorthandFactory } from '../../lib'
 import Icon from '../Icon'
+import Slot from '../Slot'
 import { buttonBehavior } from '../../lib/accessibility'
 import { Accessibility } from '../../lib/accessibility/interfaces'
 import { ComponentVariablesInput, ComponentPartStyle } from '../../../types/theme'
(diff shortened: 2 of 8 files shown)


src/components/Button/Button.tsx
```diff
@@ -160,7 +161,9 @@ class Button extends UIComponent<Extendable<IButtonProps>, IButtonState> {
       >
         {hasChildren && children}
         {!hasChildren && iconPosition !== 'after' && this.renderIcon(variables, styles)}
-        {!hasChildren && content && <span className={classes.content}>{content}</span>}
+        {!hasChildren &&
+          content &&
+          Slot.create(content, { defaultProps: { as: 'span', className: classes.content } })}
```
levithomason: Short circuit checks not required for .create() methods.  The Slot will return nothing if content does not exist.

levithomason (approved):

src/components/Button/Button.tsx
bmdalex: removed check

docs/src/examples/components/Slot/index.tsx
```diff
@@ -0,0 +1,3 @@
+import React from 'react'
+
+export default () => <div>No examples</div>
```
mnajdova: Maybe we can add here some description of the component and why it was introduced. This is not blocking this PR from being merged however.

mnajdova (approved):

docs/src/examples/components/Slot/index.tsx
bmdalex: will do, thx

PR #336, opened 2018-10-08
BREAKING: Button - implement dark theme and match redlines

bcalvery:
affects light and dark

# Button

- Modified Teams Dark Theme Button styles to match redline.
- Updating Teams default theme to match redline spec.
- removed "type" prefix for button style variables

## API Change
*"type" prefix removed from buttonVariables:*
image

## Before
image
image

## After
(For general reference, hover, pressed, keyboard focus states not shown)
image
image
image
image


alinais:
Lovely to see this one being worked on.

levithomason (approved):
Please remove secondaryTinted button type and replace it with a customization example where the variables are used to create this custom usage.  This way, the API is not required by all themes.

src/components/Button/Button.tsx
```diff
@@ -34,7 +34,7 @@ export interface IButtonProps {
   onFocus?: ComponentEventHandler<IButtonProps>
   renderIcon?: ShorthandRenderFunction
   text?: boolean
-  type?: 'primary' | 'secondary'
+  type?: 'primary'
```
kuzhelov: do we really want to remove for all the Stardust clients? What is the point to have `type` being a string with `secondary` removed - it should be a `boolean` flag then. However, would disagree that we should remove `secondary` from the list of types provided.

kuzhelov (changes requested):
lets agree on whether `secondary` should be removed as `type` enum value for `Button` @levithomason

src/components/Button/Button.tsx
bcalvery: What does "secondary" mean though, vs just a "default" button? Or conversely, if we have a "primary" and "secondary" button visual treatment, then the "neither primary nor secondary button" becomes a liability (in the context of Teams at least) because there is no such thing as a button that isn't primary or secondary... do we make "primary" and "secondary" non-optional for button? That seems onerous.

I agree that perhaps "primary" should be a Boolean rather than a string... This might also be useful in "groups" of buttons, to make some sort of rule that breaks if there is more than one primary...

kuzhelov (approved):

PR #343, opened 2018-10-09
chore: use eval-source-map

miroslavstastny:
Switching to `eval-source-map` allows us to debug original sources.

Before:
image

After:
image

Details:
https://webpack.js.org/configuration/devtool/


kuzhelov:
What are the build time implications estimated for this change? According to the reference you've provided, this change implies serious degrade in build time (speaking of symbols legend used there) - and the question is, how serious it is if we would speak about seconds? 

Another thing that I might suggest is to use this mode for `watch` development mode only - due to we don't need to pay this cost for regular CI builds.

miroslavstastny:
That's fair concern, @kuzhelov.
`webpack.config` is currently used in `build:docs:webpack` and `serve:docs` gulp tasks and sourcemaps are enabled only for *development* environment.
Local development (yarn start) as well as ci run both tasks.

`build:docs:webpack` takes ~20s with source maps disabled, ~21s with cheap-source-map and ~25s with eval-source-map. Is this the price we are willing to pay in order to be able to debug?

What documentation says about the two options:
**eval-source-map** - Each module is executed with eval() and a SourceMap is added as a DataUrl to the eval(). Initially it is slow, but it provides fast rebuild speed and yields real files. Line numbers are correctly mapped since it gets mapped to the original code. It yields the best quality SourceMaps for development.

**cheap-source-map** - A SourceMap without column-mappings ignoring loader Source Maps.
...not ideal for development nor production. They are needed for some special cases, i. e. for some 3rd party tools.

kuzhelov:
thanks @miroslavstastny - just to be clear, yes, 'build' times, not 'rebuild' ones were my concern. So, just to be sure (sorry if I've misunderstood this sentence) - do you mean that tasks affected by this change are run in CI as well?

> webpack.config is currently used in build:docs:webpack and serve:docs gulp tasks and sourcemaps are enabled only for development environment.
Local development (yarn start) as well as ci run both tasks.

---------

>  Is this the price we are willing to pay in order to be able to debug?

The price of couple seconds is pretty fair, totally agree - although I wouldn't say that it was _impossible_ to debug before - transpiled files used to work pretty fine for me. At the same time, might agree that it introduces convenience points for majority of scenarios (where you don't need the _actual_ transpiled code to dig into).

Let me, please, make couple of tests on my machine.

kuzhelov (approved):

PR #351, opened 2018-10-11
test: Adding/consolidating unit tests/description 

kolaps33:
To reach high coverage
    - adding unit tests 
    - editing existing tests 
- consolidate 
    - description to be consistent across the behaviors 
    - code to be consistent across the behaviors 
    - regex expression to hit properly description

The behaviours like "chatBehavior" and "popupBehavior" have not fully coverage as:
- chatBehaviour should be moved away from the repo
- popupBehavior will be edited 

Current coverage:
image


silviuaavram (approved):

jurokapsiar (approved):

src/lib/accessibility/Behaviors/Button/buttonBehavior.ts
```diff
@@ -3,14 +3,14 @@ import { Accessibility } from '../../interfaces'
 /**
  * @description
  *  Adds role='button' if element type is other than 'button'. This allows screen readers to handle the component as a button.
- *  Adds attribute 'aria-disabled=true' based on the property 'disabled'.
+ *  Adds attribute 'aria-disabled=true' based on the property 'disabled'. This can be overriden by directly providing 'aria-disabled' property to the component.
```
kuzhelov: while do understand that it could be a rather subjective matter, at the same it seems that 
> This can be overriden by providing 'aria-disabled' property directly to the component.

sounds more natural

src/lib/accessibility/Behaviors/Button/buttonBehavior.ts
```diff
@@ -3,14 +3,14 @@ import { Accessibility } from '../../interfaces'
...
  */
 
 const buttonBehavior: Accessibility = (props: any) => ({
   attributes: {
     root: {
       role: props.as === 'button' ? undefined : 'button',
-      'aria-disabled': !!props['disabled'],
+      'aria-disabled': 'aria-disabled' in props ? props['aria-disabled'] : props['disabled'],
```
kuzhelov: why `in` check is necessary? seems that `props['aria-disabled'] || props['disabled']` would be sufficient enough, as this will make overrides only when `aria-disabled` prop is set

test/specs/behaviors/testDefinitions.ts
```diff
@@ -1,11 +1,12 @@
 import { TestDefinition, TestMethod, TestHelper } from './testHelper'
+import * as keyboardKey from 'keyboard-key'
 
 const definitions: TestDefinition[] = []
 const testHelper = new TestHelper()
 
 // Example:  Adds attribute 'aria-pressed=true' based on the property 'active'
 definitions.push({
-  regexp: /Adds attribute '([\w\-\w \s*]+)=([a-z]+)' based on the property '([a-z]+)'\./g,
+  regexp: /Adds attribute '([\w\-\w \s*]+)=([a-z]+)' based on the property '([a-z]+)'\.$/g,
```
kuzhelov: why there is a separation implied for cases when only this sentence is used vs cases where this sentence can be followed by additional statements (like `This can be overriden by directly providing '([a-z -]+)' property to the component\.`)? Seems like the same set of tests should be run in both cases

test/specs/behaviors/testDefinitions.ts
```diff
@@ -86,6 +87,81 @@ definitions.push({
   },
 })
 
+// Adds attribute 'aria-selected=true' to 'anchor' component's part based on the property 'active'. This can be overriden by directly providing 'aria-selected' property to the component.
+definitions.push({
+  regexp: /Adds attribute '([a-z A-Z -]+)=([a-z 0-9]+)' to '([a-z -]+)' component's part based on the property '([a-z]+)'\. This can be overriden by directly providing '([a-z -]+)' property to the component\./g,
```
kuzhelov: why we won't test for `This can be overriden by directly providing '([a-z -]+)' property to the component\.` expression only and introduce tests for verifying that only? Just imagine how each new sentence would complicate the overall picture if we would need to introduce new tests for each possible combination of conditions - in contrast we could just implement modular approach to solve this problem where each sentence would be responsible for a set of tests aimed to verify this sentence's condition only

test/specs/behaviors/testDefinitions.ts
```diff
@@ -86,6 +87,81 @@ definitions.push({
...
+    ][attributeToBeAdded]
+    expect(testHelper.convertToBooleanIfApplicable(expectedResultAttributeDefined)).toEqual(
+      testHelper.convertToBooleanIfApplicable(valueOfAttributeToBeAdded),
+    )
+  },
+})
+
 // Example: Adds attribute 'aria-hidden=true', if there is no 'alt' property provided.
 definitions.push({
   regexp: /Adds attribute '([\w\-\w \s*]+)=([a-z]+)', if there is no '[a-z]+' property provided\.+/g,
```
kuzhelov: don't see any place where the match of the third 'group' (`[a-z]+` - yes, this is not a group in the expression, thus I am using quotes) is used. Is it really irrelevant what we would match there for tests' logic? According to the text of the statement, it is not

test/specs/behaviors/testDefinitions.ts
```diff
@@ -96,6 +172,14 @@ definitions.push({
     expect(testHelper.convertToBooleanIfApplicable(expectedResult)).toBe(
       testHelper.convertToBooleanIfApplicable(attributeExpectedValue),
     )
+
+    const propertyWithAlt = { alt: 'mockText' }
```
kuzhelov, line 155: here `alt` prop is used for verifying the statement, but statement's logic doesn't claim specifically that `alt` property will invalidate the attribute value

kuzhelov (changes requested):
please, consider to review comments left

test/specs/behaviors/testDefinitions.ts
kolaps33: changed regex to hardcoded 'alt'. Please resolve comment if this change is enough ?

src/lib/accessibility/Behaviors/Button/buttonBehavior.ts
silviuaavram: no, actually the check is ok. our first option for `aria-disabled` is the manually specified `aria-disabled` on the component. If the component user does not specify it via props, we give `aria-disabled` the value of the `disabled` prop, which is always available.

The condition above will not work when, for instance you use `<Component aria-disabled="false" disabled="true">`. The `aria-disabled` value applied to the HTML element will be `true` in this case, which is not what we wanted.

src/lib/accessibility/Behaviors/Button/toggleButtonBehavior.ts
```diff
@@ -4,14 +4,14 @@ import { Accessibility } from '../../interfaces'
...
+ * Adds attribute 'aria-disabled=true' based on the property 'disabled'. This can be overriden by providing 'aria-disabled' property directly to the component.
  */
 
 const toggleButtonBehavior: Accessibility = (props: any) => ({
   attributes: {
     root: {
       role: props.as === 'button' ? undefined : 'button',
-      'aria-disabled': !!props['disabled'],
+      'aria-disabled': props['aria-disabled'] || props['disabled'],
       'aria-pressed': !!props['active'],
```
levithomason: Should `active` not be updated in the same way as `disabled`?  Currently, we coerce to boolean so it is always present.

src/lib/accessibility/Behaviors/Button/buttonBehavior.ts
```diff
@@ -3,14 +3,14 @@ import { Accessibility } from '../../interfaces'
...
- *  Adds attribute 'aria-disabled=true' based on the property 'disabled'.
+ *  Adds attribute 'aria-disabled=true' based on the property 'disabled'. This can be overriden by providing 'aria-disabled' property directly to the component.
  */
 
 const buttonBehavior: Accessibility = (props: any) => ({
   attributes: {
     root: {
       role: props.as === 'button' ? undefined : 'button',
-      'aria-disabled': !!props['disabled'],
+      'aria-disabled': props['aria-disabled'] || props['disabled'],
```
levithomason: The description indicates that passing `aria-disabled` will override the default behavior, however, the logic here only works if you pass `true`.  The short circuit will not respect a value of `false`.

An override would respect any value the user passes:

```suggestion
      'aria-disabled': typeof props['aria-disabled'] !== 'undefined' ? props['aria-disabled'] : props['disabled'],
```


Or possibly only not `undefined` nor `null` values:

```suggestion
      'aria-disabled': !_.isNil(props['aria-disabled']) ? props['aria-disabled'] : props['disabled'],
```

Not sure what the expected intent is here.  My intuition says the second example would seem most correct to a React user, since `null` and `undefined` are both considered "not present" in React.

kuzhelov (approved):

PR #353, opened 2018-10-11
feat(Form): base implementation

mnajdova:
This PR aims to create the base implementation for the Form component. Here is a short RFC for the proposed API.

# Form

This component will be used for displaying a set of related user input fields in a structured way. In the implementation the Form component will use the Grid component for customizing the appearance of the fields inside.

### TODO

- [x] Conformance test
- [ ] Stardust base theme
- [ ] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [ ] Confirm RTL usage
- [ ] Update the CHANGELOG.md

# API Proposal

## as: defaultValue: 'form'

An element type to render as (string or function).

## className

Additional classes.

## action

The HTML form action.

## onSubmit

The HTML form submit handler. 

## fields

Array shorthand for the fields inside the form.

## message: Segment shorthand prop

Will show a message inside the form (error, success, warning etc.)
**UPDATE**: the message prop will be implemented to the Form.Field component instead of here.

## size:  enum: sizes

Size of the text and elements inside the form. Should be consistent with the Text's sizes.
**UPDATE**: this will be addressed in a consequent PR, as we still haven't decided on the way it should be implemented.

## loading: PropTypes.bool

Automatically show a loading indicator.
**UPDATE**: after discussion, decided that this prop will not be introduced. See comment: https://github.com/stardust-ui/react/pull/353#discussion_r225130590

## variables?: ComponentVariablesInput

# Form.Field

A field is a form element containing a label and an input.

## as

An element type to render as (string or function).

## className

Additional classes.

## content: customPropTypes.contentShorthand
Shorthand for primary content.

## control: customPropTypes.some([PropTypes.func, PropTypes.oneOf(['button', 'input', 'select', 'textarea'])])
A form control component (i.e. Input, RadioGroup) or HTML tagName (i.e. 'input').
Extra FormField props are passed to the control component.
Mutually exclusive with children.

## error: PropTypes.bool

Individual fields may display an error state.
UPDATE: this prop will be added in a consequent PR that will address the validation aspects. 

## inline: PropTypes.bool

A field can have its label next to instead of above it.

## label: Text with default as='label' shorthand

Mutually exclusive with children.

## required: PropTypes.bool

A field can show that input is mandatory.

## variables?: ComponentVariablesInput

By default the Form field will render the following HTML structure giving label and a input control:

```html
<div> <label for='inputID'>Label text </label><input id='inputID'.../> </div>
```

For the components that doesn't need label (like the Button), or are providing the label together with the component the label will be rendered according to that component implementation.


hughreeling:
Can you remove children lines - they're just cut and paste with old format.

hughreeling:
Side or size?

hughreeling:
Loading indicator - is that specified in theme?  or how would user customize?

mnajdova:
Children removed, side -> size fixed. The loader indicator will be indeed styled by the theme. We might decide for adding Loading component in the future and make this shorthand, but in the beginning this will be some slot inside the Layout with some styles defined in the component's styles.

mnajdova:
# Update

The Form is implemented using the Grid component with default `as` to be 'form'. This allows the user to be able to fully customize the appearance of the elements inside the form. For now, the Grid's properties are added as a props inside the Form in order for them to be advertised to the user.

## Question
Do we want to support fields shorthand inside the form for all elements inside? If we allow it would be easy to apply some props to all fields inside (for example the size property which will allow different sizing to all elements inside the form). We can achieve the same using context, but as this is not used anywhere so far wasn't sure if we want to go on that path.

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,147 @@
...
+      render: renderLabel,
+    })
+
+    const controlElement = control && React.createElement(control, rest)
+
+    // ----------------------------------------
+    // No Control
+    // ----------------------------------------
+
+    if (_.isNil(control)) {
```
kuzhelov: nit: seems that it could be simplified to `if (control) ... ` without forfeiting correctness - to make this check be consistent with other ones used in this file

docs/src/examples/components/Form/Types/FormExample.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react'
+import { Form, Input, Button } from '@stardust-ui/react'
+
+const FormExample = () => (
```
kuzhelov, line 4: we should consider to support Shorthand API as well - probably, something that you've already thought of, mentioning just for the sake of align our thoughts on that :) In that case Form will provide an `items` prop (or `fields` one) that will accept an array of `FormField` shorthands.

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,133 @@
...
+     *  Button content for childrenApi
+     *  @docSiteIgnore
+     */
+    children: PropTypes.node,
+
+    /** Additional CSS class name(s) to apply.  */
+    className: PropTypes.string,
+
+    /** The columns of the grid with a space-separated list of values. The values represent the track size, and the space between them represents the grid line. */
+    columns: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
```
kuzhelov: I would rather consider to remove layout-related props from the Form's API for now, as am pretty sure that we would be able to address quite wide set of scenarios by
- introducing base 'vertical form' layout support as a default one
- ability to customize form styles for the client - not that with that it would be possible for the client to provide custom layout styles for each Form.Field element
- potentially add ability to introduce support for horizontal forms

Not saying that **_for sure we won't need Grid support for form's layout_** - the only thing that stating is that _for now_ we, probably, don't need it

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * The HTML form submit handler.
+     * @param {SyntheticEvent} event - React's original SyntheticEvent.
+     * @param {object} data - All props.
+     */
+    onSubmit: PropTypes.func,
+
+    /** Automatically show a loading indicator. */
+    loading: PropTypes.bool,
```
kuzhelov: emm, not sure that it should be part of the default form interface. To me `loading` seems to be something that introduces behavioural aspects to the form - and, thus, see this `loading` aspect to be used as rather _decorator_ for the general `Form` component.

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+    /** The HTML form action. */
+    action: PropTypes.string,
+
+    /** The size for the Text component */
+    size: PropTypes.oneOf(['smallest', 'smaller', 'small', 'medium', 'large', 'larger', 'largest']),
```
kuzhelov: not sure that it should be exposed for `Form` component, especially given that this prop, in fact, defines `Text` size (we should ensure prop types consistency here, by the way then). Pretty sure that it was added for reason - could you, please, suggest which case you've tried to address with that?

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,147 @@
...
+} from '../../../types/utils'
+import Text from '../Text'
+
+export interface IFormFieldProps {
+  as?: any
+  children?: ReactChildren
+  className?: string
+  label?: ShorthandValue
+  renderLabel?: ShorthandRenderFunction
+  control?: React.ReactType<any>
```
kuzhelov: yes, agree, this should considered to be a shorthand.

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,147 @@
...
+import { ComponentVariablesInput, ComponentPartStyle } from '../../../types/theme'
+import {
+  Extendable,
+  ReactChildren,
+  ShorthandValue,
+  ShorthandRenderFunction,
+} from '../../../types/utils'
+import Text from '../Text'
+
+export interface IFormFieldProps {
```
kuzhelov, line 16: one prop that I see missing here is, probably, `errorMessage` one (or general `message`) - the text that will be displayed if field value's validation had failed/passed (text below the input field on the picture): 

image

image


Might expect that it will be part of the 'extended' `Input` component, but not completely sure, as validation stuff seems to be something that is rather form-related. In either case, do we have any plans to support it?

src/components/Form/Form.tsx
mnajdova: The idea for adding this prop was providing out of the box sizing mechanism of all items inside the form. For example, if the user wants the form to have bigger size then regular, applying only one prop on the form, should scale all elements inside (for now it is just the Text components - which are used for the labels, but we should further adjust the sizes of the input and other elements which are going to be used as FormFields) If this property is not added here, then the user should be himself responsible for correct sizing of all elements inside the Form. Please see the latest commit on the PR, which is introducing the SizeContext that is used inside the FormField component.

kuzhelov:
I do like the proposal, totally agree with structuring `Form` component as `Form` -> `Form.Filed` collection. Probably, this is the key thing that we should agree upon and introduce by this PR.

Have left some comments, but those are, mainly, necessary to clarify some bits.

src/components/Form/Form.tsx
mnajdova: Make sense, will remove this prop for now. It seems like a good concept, but in the end introducing this here would then bring out question as why only this component has the loading indicator, so for this moment I agree that it should not be part of the Form component.

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,136 @@
...
+    ElementType,
+    classes,
+    accessibility,
+    variables,
+    styles,
+    rest,
+  }): React.ReactNode {
+    const { as, rows, columns, action, children, content, size } = this.props
+    return (
+      <SizeContext.Provider value={{ size }}>
```
kuzhelov: would rather try to avoid this in general, and certainly at the early stages. Could we list the cases where we need this sizing, and after that think about whether there are alternative ways to achieve the same goal?

docs/src/examples/components/Form/Types/FormExample.tsx
mnajdova: Already talk with Hugh on this. This was my initial idea (having fields array for defining everything inside the Form), but he had a good argument: 'Form is effectively just wrapper around some controls - so what's inside is pretty free.' I am okay with adding this and we can see further if there will be any need for this prop.

src/components/Form/Form.tsx
mnajdova: This was the reason why I added the sizing mechanism: https://github.com/stardust-ui/react/pull/353#discussion_r225133012 I agree that we may want to address this in some consequent PR, and not the initial implementation.

src/components/Form/FormField.tsx
mnajdova: The logic for generating the control is a bit tricky. The control can be some component (we still don't have all components that can appear in the form implemented) or some HTML element (checkbox, button etc.). This is why this is just a type, and for consistency all properties available in the FormField are applied to the control element. What would be your proposal for using the shorthand in this case? Would be use the shorthand as one specific component and then all other components will always send the <Component ...props /> or some function and be responsible for applying all props inside that component?

src/components/Form/Form.tsx
mnajdova: This seemed like the easier (for implementation and usage) layouting mechanism support out of the box for the form component. With this by default we are supporting the vertical, multi-column and horizontal layout. We might introduce this as a consequent PR too, after the initial implementation is merged. What are your thoughts on this? And, yes, I would definitelly use the Grid in the background anyway for aligning the elements

src/components/Form/FormField.tsx
mnajdova: I introduced the message prop on the form itself, but you are right, it makes more sense to have this inside the field, as different fields might introduce different text for each of them. I will add this as a plain text shorthand that will be displayed below the input for now, and we can iterate on it later, once the validation aspects will be added.

src/components/Form/Form.tsx
mnajdova: Reverted the changes for the sizing mechanism from this PR. We can come back to these in some consequent PR.

docs/src/examples/components/Form/Types/FormExample.tsx
mnajdova: Added fields prop in the form.

src/components/Form/FormField.tsx
mnajdova: Added message prop for the form field component that appears below the input.

docs/src/examples/components/Form/Types/FormExample.tsx
kuzhelov: yes, but there is a counterargument to this point - we could add _anything_ as Form.Field, as long as it is a shorthand. So, without limiting the cases covered, we would introduce the `fields/items` API that would help user in vast majority of cases. 

Another thing that goes in favour of that - consider what we do have for `Chat` now, with `ChatItem` component - essentially, the story is exactly the same, as each `ChatItem` item could be heavily customized and, thus, doesn't limit ability to generalize for the component :)

src/components/Form/Form.tsx
kuzhelov: yes - my point is rather about how this layou mechanism would be exposed. And, given that this is a question that should involve some consideration phases, I would not introduce any layout aspects to the API in this PR

src/components/Form/FormField.tsx
kuzhelov: Here is the thing that I am concerned about - how we would be able to customize styles of this component? Yes, while it would be possible (with current approach) to pass `styles` to `Form.Field`, and those, if I've understood you correctly, will be passed to `content` element - for regular HTML element, however, we will have a problem that we would provide unknown `styles` attribute. This is the reason why I see this boundary should be clearly advertised - specifically, by providing the following semantics to the client: here is the `control` slot, please, provide all the necessary component props and style customizations there.

So, I would suggest to use `createHTMLInput` method from `factories.tsx` as a value for `content` - and even given that we have some problems with its implementation now, with this being fixed we should end up with correct implementation for `Form.Field` as well

mnajdova: The `control` is converted to shorthand, and `controlType` prop is added that is the previous `control` prop. Depending on the `controlType`, different slot shorthand factories are created. @Bugaa92 I need your input whether this is the right approach. And yeah, I will need to add some custom tests for this prop, after we decide on the API used.

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,221 @@
...
+    className: PropTypes.string,
+
+    /**
+     * A form control component (i.e. Input) or HTML tagName (i.e. 'input').
+     * Extra FormField props are passed to the control component.
+     * Mutually exclusive with children.
+     */
+    controlType: customPropTypes.some([
+      PropTypes.func,
+      PropTypes.oneOf(['button', 'input', 'select', 'textarea']),
```
bmdalex: can we use these to define the `controlType ` prop in the interface?

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,221 @@
...
+      renderLabel,
+      id,
+      type,
+      required,
+      message,
+      renderMessage,
+      controlType,
+    } = this.props
+
+    let labelContent = label
```
bmdalex:
```suggestion
    const labelContent = required
      ? typeof label === 'string'
        ? `${label} *`
        : typeof label === 'object' && (label as any).content
          && { ...label, content: `${(label as any).content} *` }
      : label
```

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,221 @@
...
+    })
+
+    const messageElement = Text.create(message, {
+      defaultProps: {
+        styles: styles.message,
+      },
+      render: renderMessage,
+    })
+
+    const factoryMethod = createSlotFactory(controlType, name => ({ name }))
```
bmdalex: I think this is ok for now, but we might move this `createSlotFactory` to `lib/factories.tsx` or something similar; I understand the usage now

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,221 @@
...
+          {messageElement}
+        </ElementType>
+      )
+    }
+
+    // --------------------------------------------------------
+    // Check box or radio (label should appear after the input)
+    // --------------------------------------------------------
+
+    if (controlType && (type === 'checkbox' || type === 'radio')) {
```
bmdalex: `controlType` check not needed
```suggestion
    if (type === 'checkbox' || type === 'radio') {
```

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,221 @@
...
+    // ----------------------------------------
+    // No Control
+    // ----------------------------------------
+
+    if (!controlType) {
+      if (!label) {
+        return (
+          <ElementType {...rest} className={classes.root}>
+            {childrenExist(children) && children}
+            {messageElement}
```
bmdalex: I thought we're not supposed to change the `children` if they are provided; with this condition, `{messageElement}` is added even if `children` are defined

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,221 @@
...
+    const controlElement = factoryMethod(control || {}, {
+      defaultProps: { required, ...rest },
+      render: renderControl,
+    })
+
+    // ----------------------------------------
+    // No Control
+    // ----------------------------------------
+
+    if (!controlType) {
```
bmdalex: this if branch looks strange; can we do:
```jsx
if (!controlType) {
  return (
      {label ? labelElement : (childrenExist(children) && children)}
      {messageElement}
    </ElementType>
  )
}

File: src/components/Slot/Slot.tsx [hunk 15]
```diff
@@ -12,12 +12,18 @@ export interface ISlotProps {
   variables?: ComponentVariablesInput
 }
 
+export const createSlotFactory = (as: any, mapValueToProps: MapValueToProps) => (
+  val,
+  options: ObjectOf<any> = {},
```
bmdalex: this type changed a bit, pls take a look again at #326, and also apply the fix from `factories.tsx` cause you will have a weird type error related to `IProps`

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,221 @@
...
+    } = this.props
+
+    let labelContent = label
+
+    if (required) {
+      labelContent =
+        typeof label === 'string'
+          ? `${label} *`
+          : typeof label === 'object' && (label as any).content
+            ? { ...label, content: `${(label as any).content} *` }
```
bmdalex: I don't think we should make so many presumptions on the content of the `label` shorthand; this looks like a candidate for a smaller component that will handle all this logic

types/theme.d.ts
```diff
@@ -175,6 +175,8 @@ export interface IThemeComponentStylesInput {
   ChatItem?: IComponentPartStylesInput
   ChatMessage?: IComponentPartStylesInput
   Divider?: IComponentPartStylesInput
+  Form?: IComponentPartStylesInput
```
bmdalex, line 178: never have i ever update this file... I didn't even know about it; we should have a mechanism to do it for us automatically...

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,132 @@
...
+  handleSubmit = (e, ...args) => {
+    const { action } = this.props
+
+    // Heads up! Third party libs can pass own data as first argument, we need to check that it has preventDefault()
+    // method.
+    if (typeof action !== 'string') _.invoke(e, 'preventDefault')
+    _.invoke(this.props, 'onSubmit', e, this.props, ...args)
+  }
+
+  renderFields = () => {
```
bmdalex: pls make private

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,132 @@
...
+    variables: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
+  }
+
+  public static defaultProps = {
+    as: 'form',
+  }
+
+  static Field = FormField
+
+  handleSubmit = (e, ...args) => {
```
bmdalex: pls make private

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+    /** Override for theme site variables to allow modifications of component styling via themes. */
+    variables: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
+  }
+
+  public static defaultProps = {
+    as: 'form',
+  }
+
+  static Field = FormField
```
bmdalex: let's keep consistency within the file at least when it comes to modifiers; if we explicitly use `public` let's add it to every public member

bmdalex (changes requested):
added comments and suggestions

src/components/Form/FormField.tsx
mnajdova: We don't know all components that can be defined there upfront, that is why the type is `React.ReactType<any>`. We know just for the primitive types that can be one of those.

types/theme.d.ts
mnajdova: Yep, this file was updated in some of the recent refactoring and adding typings for the styling. Agreed that there should be some mechanism for this.

src/components/Form/FormField.tsx
mnajdova: I think I over-complicated things here :\ I will just add additional text component after the label if the field is required, without changing the actual text inside the label.

src/components/Form/FormField.tsx
mnajdova: Will refactor this using two different elements

src/components/Form/FormField.tsx
bmdalex: nit: here's what I meant: since
```tsx
type ReactType<P = any> = string | ComponentType<P>
```
our type could be more precise:
```tsx
controlType: ComponentType<any> | 'button' | 'input' | 'select' | 'textarea'
```
so that the user gets better suggestions; still nit, no need to address

docs/src/examples/components/Form/Types/FormExample.shorthand.tsx
```diff
@@ -0,0 +1,40 @@
...
+  },
+  { controlType: Button, control: { content: 'Submit' }, key: 'submit' },
+]
+
+const FormExample = () => (
+  <Form
+    onSubmit={() => {
+      alert('Form submitted')
+    }}
+    fields={fields}
```
kuzhelov, line 34: nit: when this example is rendered, it seems that there is no any gap provided between checkbox and text field:

image

Could we address this problem in styles?

docs/src/examples/components/Form/Types/FormExample.tsx
```diff
@@ -0,0 +1,34 @@
...
+  <Form
+    onSubmit={() => {
+      alert('Form submitted')
+    }}
+  >
+    <Form.Field
+      label="First name"
+      controlType={Input}
```
kuzhelov: this seems to have `as` semantics - maybe we should name this prop like `controlAs`?

kuzhelov: also, thinking that for the first stages maybe it would be fine to eliminate this prop altogether - and just use `control` slot for that:

```jsx
 <Form.Field
      label="Last name"
      name="lastName"
      id="last-name"
      required={true}
      control={{ as: 'input' }}
```

A benefit of this approach would be that we will eliminate any possibilities for collisions of different prop values come from different places. What do you think?

docs/src/examples/components/Form/Types/FormExample.tsx
```diff
@@ -0,0 +1,34 @@
...
+      id="last-name"
+      required={true}
+    />
+    <Form.Field
+      label="I agree to the Terms and Conditions"
+      controlType="input"
+      type="checkbox"
+      id="conditions"
+    />
+    <Form.Field controlType={Button} control={{ content: 'Submit' }} />
```
kuzhelov: proceeding with the thought mentioned above - maybe this could be considered as an alternative for the first stages?
```jsx
<Form.Field control={{ as: Button, content: 'Submit' }} />
```

docs/src/examples/components/Form/Types/FormExample.tsx
mnajdova: I added this as a specific property because we need it for actually creating the slot used here:
```jsx
    const factoryMethod = createSlotFactory(controlType, name => ({ name }))
    const controlElement = factoryMethod(control || {}, {
      defaultProps: { required, id, name, type, styles: styles.control },
      render: renderControl,
    })
```
We are using the Slot as this can be some HTML element, or component. If we use this form the 'as' property, then it is becoming kind a tricky because this can be object, component, some primitive type etc. What do you think? I am okay with renaming it, but yeah, let's decide whether it should exists as a prop first.

docs/src/examples/components/Form/Types/FormExample.tsx
mnajdova: Please take a look into the previous comment: https://github.com/stardust-ui/react/pull/353#discussion_r226324506

docs/src/examples/components/Form/Types/FormExample.shorthand.tsx
mnajdova: Will address this of course.

docs/src/examples/components/Form/Types/FormExample.tsx
mnajdova: We might right away use the Slot.create and propagate the control shorthand there, but this was the proposed way for using the Slot create method. I am open for other suggestions.

docs/src/examples/components/Form/Types/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+const Types = () => (
+  <ExampleSection title="Types">
+    <ComponentExample
+      title="Default"
+      description="A default form."
+      examplePath="components/Form/Types/FormExample"
+    />
+    <ComponentExample
+      title="Inline"
+      description="The controls inside a form can appear next to the label instead of below it."
```
kuzhelov: nit: could be simplified as `The form controls ...`

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,128 @@
...
+import {
+  ComponentEventHandler,
+  Extendable,
+  ReactChildren,
+  ShorthandValue,
+  ShorthandRenderFunction,
+} from '../../../types/utils'
+import FormField from './FormField'
+
+export interface IFormProps {
```
kuzhelov: nit: just for the sake of consistency with other components, lets order these props alphabetically

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,128 @@
...
+  action?: string
```
kuzhelov, line 17: nit: just for the sake of consistency with other components, lets arrange these alphabetically

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,128 @@
...
+    content: customPropTypes.contentShorthand,
+
+    /** Shorthand array of props for the Form.Fields inside the Form. */
+    fields: customPropTypes.collectionShorthand,
+
+    /**
+     * A custom render iterator for rendering each of the Form fields.
+     * The default component, props, and children are available for each field.
+     *
+     * @param {React.ReactType} Component - The computed component for this slot.
```
kuzhelov, line 74: just to be sure, what are the scenarios where these JS Doc attributes (like @param) will augment IntelliSense experience? Maybe it is safe for us to just avoid them, _if_ there is no much value in them?

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,128 @@
...
+      </ElementType>
+    )
+  }
+
+  private handleSubmit = (e, ...args) => {
+    const { action } = this.props
+
+    // Heads up! Third party libs can pass own data as first argument, we need to check that it has preventDefault()
+    // method.
+    if (typeof action !== 'string') _.invoke(e, 'preventDefault')
```
kuzhelov: could you, please, elaborate on the case we are trying to address here? How it is read by me is:
- invoke `onSubmit` props callback for the proper set of arguments being provided when form is submitted
- but if `action` is not a string, call `preventDefault` on the event object

Frankly, a bit confused with the second condition. What we are trying to handle with that? And, suppose that I am passing some garbish but `string` value as action, like `aaa` - shouldn't we call `preventDefault` in that case as well?

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,128 @@
...
+    _.invoke(this.props, 'onSubmit', e, this.props, ...args)
```
kuzhelov, line 115: could you, please, suggest the case where these `...args` will be utilized?

docs/src/examples/components/Form/Types/FormExample.shorthand.tsx
mnajdova: Fixed for the checkbox and radio type inputs

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,207 @@
...
+  Extendable,
+  ReactChildren,
+  ShorthandValue,
+  ShorthandRenderFunction,
+} from '../../../types/utils'
+import Text from '../Text'
+import { createSlotFactory } from '../Slot/Slot'
+
+export interface IFormFieldProps {
+  as?: any
```
kuzhelov, line 17: nit: just to be consistent with other components, lets arrange these alphabetically

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,207 @@
...
+  public static className = 'ui-form__field'
+
+  static create: Function
+
+  public static propTypes = {
+    /** An element type to render as (string or function). */
+    as: customPropTypes.as,
+
+    /**
+     *  Button content for childrenApi
```
kuzhelov: oops :)

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,128 @@
...
+  public static displayName = 'Form'
+
+  public static className = 'ui-form'
+
+  public static propTypes = {
+    /** An element type to render as (string or function). */
+    as: customPropTypes.as,
+
+    /**
+     *  Form content for childrenApi
```
kuzhelov: lets end it with dot, for the sake of consistency

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  public renderComponent({
+    ElementType,
+    classes,
+    accessibility,
+    variables,
+    styles,
+    rest,
+  }): React.ReactNode {
+    const {
```
kuzhelov, line 129: just sharing my thoughts on that - this seem to be the list of items (from this quite big set) one does really care about:
```jsx
const {
  children,
      control,
      label,
      id,
      name,
      type,
      required,
      message,
}  = this.props
```
This looks much cleaner

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,207 @@
...
+      required,
+      message,
+      renderMessage,
+      controlType,
+    } = this.props
+
+    const labelElement = Text.create(label, {
+      defaultProps: {
+        as: 'label',
+        htmlFor: id,
```
kuzhelov, line 146: 👍

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,207 @@
...
+        styles: styles.label,
+      },
+      render: renderLabel,
+    })
+
+    const messageElement = Text.create(message, {
+      defaultProps: {
+        styles: styles.message,
+      },
+      render: renderMessage,
```
kuzhelov, line 156: just to be sure (but am pretty confident that this case is thought of) - it is safe to pass `null | undefined` value here, it will be just ignored, right?

mnajdova: Yes, this is the case, I was testing this without sending the render method

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,207 @@
...
+    })
+
+    const factoryMethod = createSlotFactory(controlType, name => ({ name }))
+    const controlElement = factoryMethod(control || {}, {
+      defaultProps: { required, id, name, type, styles: styles.control },
+      render: renderControl,
+    })
+
+    const content = (
```
kuzhelov, line 164: now it might seem to be a matter of preference, but with more complex elements structure it will introduce unnecessary code duplication (suppose that there are three elements, instead of two `controlElement` and `labelElement` that we have now). We might consider to remove this source for code duplication in the following way:

```jsx
<>
  {isControlElementFirst && controlElement}
  {labelElement}
  {!isControlElementFirst && controlElement}

  {messageElement}
</>
```

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,207 @@
...
+    )
+
+    return (
+      <ElementType className={classes.root} {...rest}>
+        {childrenExist(children) ? children : content}
+      </ElementType>
+    )
+  }
+
+  private shouldControlAppearFirst = () => {
```
kuzhelov, line 180: the name and intent is very readable 👍

src/components/Form/Form.tsx
mnajdova: The changes regarding accessibility were made from @kolaps33 Let's ping him.

src/components/Form/Form.tsx
mnajdova: This was taken from Semantic UI. Essentially, the onSubmit should be invoke only if the action is valid, in other cases the users usually use the onClick on some button inside the Form, that's why we are preventing this.

mnajdova: This is taken from Semantic UI. The explanation for this is in the comment above.

src/components/Form/Form.tsx
mnajdova: If there are some additional arguments sent here, for example if the form is going to be used with some additional library (as for validation for example) we must propagate this arguments further. Also taken from Semantic UI, made sense to me.

src/components/Form/FormField.tsx
mnajdova: Yep, getting rid of the renderXXX methods would reduce this much. Also, getting to experimenting to get rid of the control type :)

src/components/Form/FormField.tsx
mnajdova: Agreed looks much more cleaner.

src/components/Form/FormField.tsx
mnajdova: Got rid of the controlType

docs/src/examples/components/Form/Types/FormExample.shorthand.tsx
kuzhelov: Marija, this the rendering result that I had when tried the examples: 

image

Please, ping me if you'll need help to address that - probably this is result of recent changes made related to `controlType`, will be more than happy to help

mnajdova: This is regression from the latest changes of the Input component. After merging the latest master, using
```jsx
<Slot as={Input} />
```
renders the html:
```html
<div class="ui-slot ui-input ui-slot">div</div>
```
It works as expected with the RadioGroup, input, Button etc. as well as with the previous implementation of the Input component. Will need to investigate this with Alex tomorrow.

src/components/Form/Form.tsx
bmdalex: @mnajdova not addressed

src/components/Form/Form.tsx
bmdalex: @mnajdova not addressed

src/components/Form/Form.tsx
bmdalex: @mnajdova not addressed

bmdalex (changes requested):

src/components/Form/FormField.tsx
bmdalex: NA anymore

src/components/Form/Form.tsx
bmdalex: @mnajdova please disregard; Github is not that good at showing the correct changes..it was actually addressed

src/components/Form/Form.tsx
bmdalex: @mnajdova please disregard; Github is not that good at showing the correct changes..it was actually addressed

src/components/Form/Form.tsx
bmdalex: @mnajdova please disregard; Github is not that good at showing the correct changes..it was actually addressed

bmdalex (approved):
looks good, thx for addressing

bmdalex (changes requested):
blocking for now just to provide better fix for the `<Slot as={Input} />` issue

src/components/Input/Input.tsx
```diff
@@ -150,47 +134,36 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, IInputState
...
-              render: renderIcon,
-            })}
-          </>
-        ),
-        styles: styles.root,
-        ...rest,
-      },
-      render: renderWrapper,
-    })
+    return (
```
bmdalex: let's not make this huge amount of changes just for the fact that we cannot use <Slot as={Input} />
the issue was that I forgot to add `children` as handled prop to the `Input` so they are passed as undefined to the `Slot.create` method; better fix is to just add:

```tsx
/**
 *  Used to set content when using childrenApi - internal only
 *  @docSiteIgnore
 */
children: PropTypes.node,
```
to the `propTypes` object in this file and
```tsx
children?: ReactChildren
```
to the `IInputProps` interface

@mnajdova @kuzhelov

mnajdova: Yep, this solved the issue. Thanks!

bmdalex (approved):
looks good, let's merge 👍

docs/src/examples/components/Form/Types/FormExample.shorthand.tsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react'
+import { Form, Input, Button } from '@stardust-ui/react'
+
+const fields = [
+  {
+    label: 'First name',
+    control: { as: Input },
```
kuzhelov: could we provide either `Input` or `input` as a default value - for the sake of it not being necessary for the client to always specify the underlying type?

src/components/Form/Form.tsx
kuzhelov: yes, I've got it - but still cannot get the idea from the explanation, could you, please, help me with that? Specifically,  what am confused about is the following:

> a bit confused with the second condition. What we are trying to handle with that? And, suppose that I am passing some garbish but string value as action, like aaa - shouldn't we call preventDefault in that case as well?

test/specs/components/Form/FormField-test.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+const formFieldImplementsShorthandProp = implementsShorthandProp(FormField)
+
+describe('FormField', () => {
+  isConformant(FormField)
+  formFieldImplementsShorthandProp('label', Text)
+  formFieldImplementsShorthandProp('message', Text)
+  formFieldImplementsShorthandProp('control', Slot)
+
+  it('renders the control provided in the controlType prop', () => {
```
kuzhelov: now this could be renamed :) Same applies to the next test

docs/src/examples/components/Form/Types/FormExample.shorthand.tsx
mnajdova: We can provide the control to be by default `{as: Input}` but if the user wants to override it, would have to specify the as again, otherwise we would need to handle all different types of control (object, component etc) Would this be enough?

src/components/Form/Form.tsx
mnajdova: The onSubmit should be invoked only if the action is set to valid value. In other cases, the users should use onClick on some button inside the Form if they want to submit the form.

test/specs/components/Form/FormField-test.tsx
mnajdova: This will be renamed, no need for the second now that the control is a regular shorthand :)

src/components/Form/Form.tsx
mnajdova: They are differently displayed in the props table in the docs 
image

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,188 @@
...
+    /** The HTML input type. */
+    type: PropTypes.string,
+
+    /** Override for theme site variables to allow modifications of component styling via themes. */
+    variables: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
+  }
+
+  public static defaultProps = {
+    as: 'div',
+    control: { as: Input },
```
kuzhelov, line 118: 👍

src/components/Input/Input.tsx
```diff
@@ -22,6 +23,7 @@ import Ref from '../Ref'
 
 export interface IInputProps {
   as?: any
+  children?: ReactChildren
```
kuzhelov, line 26: actually, this is not correct - please, consider this comment that explains why

src/components/Form/FormField.tsx
```diff
@@ -0,0 +1,188 @@
...
+  ShorthandValue,
+  ShorthandRenderFunction,
+} from '../../../types/utils'
+import Text from '../Text'
+import { default as Slot } from '../Slot/Slot'
+import Input from '../Input'
+
+export interface IFormFieldProps {
+  as?: any
+  children?: ReactChildren
```
kuzhelov, line 18: this is not correct type - here is the explanation of that: https://github.com/stardust-ui/react/pull/328#discussion_r223225907

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,128 @@
...
+  ReactChildren,
+  ShorthandValue,
+  ShorthandRenderFunction,
+} from '../../../types/utils'
+import FormField from './FormField'
+
+export interface IFormProps {
+  action?: string
+  as?: any
+  children?: ReactChildren
```
kuzhelov, line 19: lets fix this type (https://github.com/stardust-ui/react/pull/328#discussion_r223225907)

src/index.ts
```diff
@@ -27,6 +27,8 @@ export { default as Status } from './components/Status'
 export { default as TabBehavior } from './lib/accessibility/Behaviors/Tab/tabBehavior'
 export { default as TabListBehavior } from './lib/accessibility/Behaviors/Tab/tabListBehavior'
 export { default as Text } from './components/Text'
+export { default as Form } from './components/Form'
```
kuzhelov: nit: those two could be grouped together `export { default as Form, FormField } from './components/Form'`

kuzhelov (approved):
couple of minor comments - please, feel free to reason whether they should be addressed

src/components/Input/Input.tsx
mnajdova: This is our custom ReactChildren so it is actually the correct one.

src/components/Form/FormField.tsx
mnajdova: Same as above

src/components/Form/Form.tsx
mnajdova: Same as above

PR #356, opened 2018-10-12
feat(Popup): add 'target' prop

kuzhelov:
# TODO
- [x] update CHANGELOG

### Target 
Accepts DOM element that will be used as popup's target (instead of `trigger` element that is used by default). This prop will also allow to support 'bridge' scenarios, where it is necessary for client to display a popup for some DOM element that is not managed by React.

```jsx
/* popup will be triggered by button, but displayed for 'domElement' */
<Popup 
  target={domElement} 
  trigger={<Button ... />} 
/>
```

This is an optional prop - if omitted (by default) `Popup` will still use `trigger` element as a target.
```jsx
/* popup will be triggered and displayed for button */
<Popup 
  trigger={<Button ... />} 
/>
```

### DIFF
--- a/src/components/Ref/Ref.tsx
+++ b/src/components/Ref/Ref.tsx
@@ -34,6 +34,6 @@ export default class Ref extends Component<IRefProps> {
   }
 
   render() {
-    return Children.only(this.props.children)
+    return this.props.children && Children.only(this.props.children)
   }
 }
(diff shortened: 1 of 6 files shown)


src/components/Ref/Ref.tsx
```diff
@@ -34,6 +34,6 @@ export default class Ref extends Component<IRefProps> {
   }
 
   render() {
-    return Children.only(this.props.children)
+    return this.props.children && Children.only(this.props.children)
```
bmdalex, line 37: why is this necessary?

bmdalex (approved):
looks good

src/components/Ref/Ref.tsx
kuzhelov: otherwise, in case if `children` is not provided for the component, this will throw the error and fails rendering process for the whole tree - while, in contrast, if we will handle it this way, then for `innerRef` callback `null` will be validly provided as captured DOM element (as there are no elements rendered)

mnajdova (approved):

PR #126, opened 2018-08-23
Add possibility to render submenu in MenuItem

sophieH29:
# MenuItem
This PR introduces possibility to add vertical submenu rendered in MenuItem.

### TODO

- [X] Conformance test
- [X] Minimal doc site example
- [ ] Stardust base theme
- [ ] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [ ] Confirm RTL usage
- [ ] [W3 accessibility](https://www.w3.org/standards/webdesign/accessibility) check
- [ ] [Stardust accessibility](https://github.com/stardust-ui/accessibility) check
- [ ] Update glossary props table
- [ ] Update the CHANGELOG.md

# API Proposal

## Submenu

submenu

```js
const submenuData = {
     items: [
           { key: 'new', content: 'New' },
           { key: 'open', content: 'Open' },
     ]
}
```

```jsx
<MenuItem submenu={submenuData} content="File" />
```

```html
<li>
  <a> File </a>
  <ul>
    <li> <a> New </a> </a>
    <li> <a> Open </a> </a>
 </ul>
</li>
```

PR #357, opened 2018-10-12
feat(menuItem: add menu prop)

gopalgoel19:
# menuItem ```menu``` prop

Add a menu prop which has the capability to render submenus.
screenshot 58

### TODO

- [ ] Conformance test
- [ ] Stardust base theme
- [ ] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [ ] Confirm RTL usage
- [ ] Update the CHANGELOG.md

### DIFF
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -133,14 +133,12 @@ class Menu extends AutoControlledComponent<Extendable<IMenuProps>, any> {
 
       this.trySetState({ activeIndex: index })
 
-      if (menu) {
-        this.setState(prev => {
-          if (prev.menuItemOpenKey === index) {
-            return { popupOpen: !prev.popupOpen }
-          }
-          return { menuItemOpenKey: index, popupOpen: true }
-        })
-      }
+      this.setState(prev => {
+        if (prev.menuItemOpenKey === index) {
+          return { popupOpen: !prev.popupOpen }
+        }
+        return { menuItemOpenKey: index, popupOpen: true }
+      })
 
       _.invoke(predefinedProps, 'onClick', e, itemProps)
     },
--- a/src/components/Menu/Menu.tsx
+++ b/src/components/Menu/Menu.tsx
@@ -121,11 +121,26 @@ class Menu extends AutoControlledComponent<Extendable<IMenuProps>, any> {
 
   static Item = MenuItem
 
+  state = {
+    menuItemOpenKey: 0,
+    popupOpen: false,
+    activeIndex: '0',
+  }
+
   handleItemOverrides = predefinedProps => ({
     onClick: (e, itemProps) => {
-      const { index } = itemProps
+      const { index, menu } = itemProps
 
       this.trySetState({ activeIndex: index })
+
+      if (menu) {
+        this.setState(prev => {
+          if (prev.menuItemOpenKey === index) {
+            return { popupOpen: !prev.popupOpen }
+          }
+          return { menuItemOpenKey: index, popupOpen: true }
+        })
+      }
 
       _.invoke(predefinedProps, 'onClick', e, itemProps)
     },
@@ -147,6 +162,9 @@ class Menu extends AutoControlledComponent<Extendable<IMenuProps>, any> {
           vertical,
           index,
           active: parseInt(activeIndex, 10) === index,
+          popupOpen: this.state.popupOpen,
+          menuItemOpenKey: this.state.menuItemOpenKey,
+          // togglePopup: this.togglePopup,
         },
         overrideProps: this.handleItemOverrides,
         render: renderItem,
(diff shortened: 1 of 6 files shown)


src/components/Menu/MenuItem.tsx
```diff
@@ -147,27 +157,48 @@ class MenuItem extends UIComponent<Extendable<IMenuItemProps>, IMenuItemState> {
...
+            align={vertical ? 'top' : 'start'}
+            position={vertical ? 'after' : 'below'}
+            content={{
+              content: <Menu items={menu.items} vertical type={type} />,
+              styles: {
+                padding: '0px',
+                border: '',
+              },
+            }}
+            // content={<Menu items={menu.items} type={type} vertical/>}
```
gopalgoel19: Can someone investigate why using this syntax is creating trouble?

src/components/Menu/Menu.tsx
```diff
@@ -121,10 +121,22 @@ class Menu extends AutoControlledComponent<Extendable<IMenuProps>, any> {
...
+  state = {
+    submenuOpen: false,
+    activeIndex: '',
+  }
+
   handleItemOverrides = predefinedProps => ({
     onClick: (e, itemProps) => {
       const { index } = itemProps
 
+      this.setState(prev => {
```
gopalgoel19, line 133: When should we use ```setState``` vs. ```trySetState```? Can someone please explain in brief the purpose of ```trySetState```.

kuzhelov: lets schedule a short sync for tomorrow, I will explain the concept to you. Before that, please, consider to read about Controlled and Uncontrolled Components, as well as try to make experiments with `input` component being used in these different modes - note how the value updates for each, when `onChange` handler is triggered, how value updates in controlled mode.

- https://reactjs.org/docs/uncontrolled-components.html
- https://reactjs.org/docs/forms.html

src/components/Menu/MenuItem.tsx
kuzhelov: which trouble it creates, specifically?

Please, note that when you are using 
```js
content={<Menu items={menu.items} type={type} vertical/>}
```
you don't introduce any wrapper for the Popup's content - and it could happen that the content that you are providing unwrapped (by the expression mentioned above) could have `z-index` value that would prevent it to be shown **on top** of the page's content. Consider to review this example for the Popup for more details: https://stardust-ui.github.io/react/components/popup#types-popup-content-wrapper-example

src/components/Menu/Menu.tsx
gopalgoel19: Will do. Thanks for the help. And, yes, a sync would be very helpful. 👍

src/components/Menu/MenuItem.tsx
gopalgoel19: Ah, totally missed the ```z-index```. Thanks for pointing it out.

jurokapsiar:
From ARIA point of view, menu and submenu needs to conform to [menu design pattern](https://www.w3.org/TR/wai-aria-practices/#menu) for both DOM structure and keyboard handling. Please go through the design pattern to understand the requirements. This design pattern is done as a dropdown - I suppose you can achieve similar functionality with popups, but that would require testing. Focus trap zone (https://github.com/stardust-ui/react/pull/239) should help you with focusing the items on open. But in general, you should look at https://github.com/stardust-ui/react/pull/126 which implements submenu. Was there a reason why you were not able to follow this approach?

gopalgoel19:
@jurokapsiar I did look at #126 before opening this PR. I went ahead with `Popup` as it provides alignment and positioning capabilities out of the box. Let me explore what Focus trap zone offers. I'll try to fix accessibility next. 
Also, if you strongly suggest to implement it as done in #126, let me know. I will then move ahead in that direction.

jurokapsiar:
@gopalgoel19 it is hard to say before all requirements are fully implemented - keyboard navigation is the first step, then once feature complete, we would need to test with screen readers. For positioning, I believe you can use popper.js and computePopupPlacement utility function the same way how popup uses it. Based on previous experience, following ARIA patterns saves us a lot of time. For #126 we even did validation, which you would need to do from scratch when using popup.

src/components/Menu/Menu.tsx
```diff
@@ -147,6 +159,7 @@ class Menu extends AutoControlledComponent<Extendable<IMenuProps>, any> {
           vertical,
           index,
           active: parseInt(activeIndex, 10) === index,
+          submenuOpen: this.state.activeIndex === index ? this.state.submenuOpen : false,
```
levithomason: Always destructure state and pros, see line 150:

```jsx
```

Here you would add the keys you want to use.  `activeIndex` is already destructured in scope, so you can use it directly.  `submenuOpen` needs to be added to the destructured values.  Destructuring keeps code clean, but more importantly it also breaks object references.

levithomason:
- [ ] Use nested ul instead of Popup, popup is placed at the root of the DOM opposed to the correct nested location
- [ ] Add icons indicating submenu
- [ ] Submenu needs to focus on open
- [ ] Keyboard navigation for submenu


The requirements and examples for these issues can be found in @jurokapsiar's links above.

PR #361, opened 2018-10-12
fix: updating input colors for dark theme

notandrew:


levithomason (approved):
don't merge, just  testing PR hooks :)

src/themes/teams/components/Input/inputStyles.ts
```diff
@@ -14,22 +15,35 @@ const inputStyles: ComponentSlotStylesInput<InputProps, InputVariables> = {
...
+      color: v.fontColor,
+    },
     ':focus': {
-      borderColor: v.inputFocusBorderColor,
+      borderBottomColor: v.inputFocusBorderColor,
+    },
+    '&::after': {
       borderRadius: v.inputFocusBorderRadius,
+      borderBottom: `${pxToRem(2)} solid transparent`,
+      content: '',
```
levithomason: In CSS, the value required for `content` is quoted.  However, in JS, we need to quote the CSS values 😝 

```suggestion
      content: '""',
```

We should really consider implementing a warning for this in dev mode.  Just add a fela plugin that is wrapped in a NODE_ENV check for dead code elimination.  That plugin will log a warning if the `content` value is not quoted.

@alinais / @kuzhelov / @miroslavstastny

notandrew: fixed

src/themes/teams/components/Input/inputVariables.ts
```diff
@@ -1,33 +1,42 @@
...
-  inputFocusBorderColor: string
+  inputFocusBorderBottomColor: string
   inputFocusBorderRadius: string
 }
 
-const [_2px_asRem, _3px_asRem, _6px_asRem, _12px_asRem, _24px_asRem] = [2, 3, 6, 12, 24].map(v =>
-  pxToRem(v),
-)
+const [_1px_asRem, _2px_asRem, _3px_asRem, _6px_asRem, _12px_asRem, _24px_asRem] = [
+  1,
```
notandrew: this gets expanded during commit -- not sure that we want this level of prettiness?

levithomason: Let's delete this and just use `pxToRem(1)` for example.

>"Simple is better than complex."
>
> &mdash; https://en.wikipedia.org/wiki/Zen_of_Python

src/themes/teams-high-contrast/componentVariables.ts
```diff
@@ -1,3 +1,5 @@
+export { default as Input } from './components/Input/inputVariables'
+
```
levithomason: nit: missed alphabetized sort here

PR #364, opened 2018-10-16
poc(Button): display svg icon path on button hover

miroslavstastny:
This is just an experiment to learn how `Button` with `Icon` is supposed to be used in order to implement button which contains an icon which is 'filled' on hover.

### Requirements (@200%):

  | normal | :hover
-- | -- | --
light |  image  |  image
dark | image | image
hc | image  | image

### Proposed implementation concerns

- Proposed implementation works fine for light and dark theme, I do not know how to implement that for HC theme.
- It uses static class name (`ui-icon__filled`)


src/themes/teams/components/Icon/svg/icons/bold.tsx
```diff
@@ -0,0 +1,43 @@
...
+	           c0.5,0.2,0.9,0.5,1.2,0.9c0.7,0.7,1,1.6,1,2.6c0,1.4-0.5,2.5-1.6,3.5C19.4,23.1,18,23.5,16.4,23.5z M14.5,20.2h1.2
+	           c0.6,0,1.1-0.1,1.4-0.4c0.3-0.2,0.4-0.7,0.4-1c0-0.4-0.1-0.7-0.4-0.9c-0.3-0.3-0.9-0.4-1.4-0.4h-1.2V20.2z M14.5,14.1h0.9
+	           c0.5,0,0.9-0.1,1.1-0.4c0.3-0.2,0.5-0.6,0.5-0.9c0-0.3,0-1-1.8-1h-0.7V14.1z"
+        />
+      </g>
+    </svg>
+  ),
+  styles: {
+    svg: () => ({
+      color: 'inherit',
```
miroslavstastny, line 36: This works better with button hover, see circular button example with this icon vs call icon

src/themes/teams/components/Icon/svg/icons/bold.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react'
+import * as cx from 'classnames'
+import { TeamsSvgIconSpec } from '../types'
+
+export default {
+  icon: ({ classes }) => (
+    <svg viewBox="0 0 32 32" role="presentation" className={classes.svg}>
+      <g className="icons-default-fill">
```
miroslavstastny, line 8: Not sure if we need this class for anything.

src/themes/teams/components/Icon/iconVariables.ts
```diff
@@ -1,14 +1,15 @@
...
-  [key: string]: string | number | undefined
+  [key: string]: string | number | boolean | undefined
 
   color?: string
   backgroundColor?: string
   borderColor?: string
   horizontalSpace: string
   margin: string
   secondaryColor: string
+  filled?: boolean
```
kuzhelov, line 13: probably, we should think about making this part of the Icon's interface

docs/src/examples/components/Button/Usage/index.tsx
```diff
@@ -9,6 +9,11 @@ const Usage = () => (
       description="A button used in cards is a &quot;tinted&quot; version of a default button."
       examplePath="components/Button/Usage/ButtonUsageExample"
     />
+    <ComponentExample
+      title="Toolbar button"
+      description="Icon paths change on hover"
```
kuzhelov, line 14: nit: just for the sake of consistency, there should be dot at the end ;)

kuzhelov (approved):
do like the approach - the only problem, probably, an apparent one, is that there are quite a lot of indirections involved here. At the same time, worth to notice that this PR reduces, arguably, their amount to the best possible one. Also, it seems that there are some ways to further encapsulate idea of this approach under more convenient generic API exposed by utility methods

miroslavstastny:
@codepretty - could you please review this to verify it is in line with your changes/plans on SVG icons?

docs/src/examples/components/Button/Usage/ButtonExampleToolbarButton.shorthand.tsx
```diff
@@ -0,0 +1,24 @@
...
+const ButtonExampleToolbarButton = () => (
+  <Button
+    icon={{
+      name: 'bold',
+      size: 'big',
+    }}
+    iconOnly
+    text
+    styles={{
+      '&:hover .ui-icon__filled': {
```
codepretty, line 13: Wouldn't you also need to hide the unfilled version?

src/themes/teams/components/Icon/svg/icons/bold.tsx
codepretty: We don't need this for icons in Stardust. I am going to be importing all the icons without this class

src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -122,6 +122,9 @@ const iconStyles: ComponentSlotStylesInput<IconProps, any> = {
   path: getSvgStyle('path'),
 
   secondaryPath: getSvgStyle('secondaryPath'),
+
+  unfilled: getSvgStyle('unfilled'),
```
codepretty, line 126: I am implementing this in my PR where we won't need to have these styles defined in each SVG. I also named them filled & outline. Do you think filled & unfilled works better?

miroslavstastny: I think `filled` & `outline` are fine, `unfilled` was used in legacy client.

miroslavstastny:
Closing this for now, will create a new PR once #478 is merged.

PR #371, opened 2018-10-18
fix: refactor approach for px-to-rem conversions

### LINKED ISSUE #394: pxToRem() doesn't support negative values
### Steps
pass a negative value to the pxToRem function

### Expected Result
I expect a negative Rem value to be returned

### Actual Result
An error is thrown "Invalid value of: …"

### Code:
`export const pxToRem = (value: number = 0): string => {
 if (!_htmlFontSizeInPx) {
 _htmlFontSizeInPx = getComputedFontSize()
 }
 **if (value < 0) {
 throw new Error(`Invalid value of: '${value}'.`)
 }**
 }
 const convertedValueInRems = value / _htmlFontSizeInPx
 return `${_.round(convertedValueInRems, 4)}rem`
}`

kuzhelov:
### Blocked by
- [x] decision on functionality we see to be introduced by pxToRem

--------------- 

This PR suggests an approach to `pxToRem` that would solve the following problems we are currently struggling with:
- ensure that there is no 'race condition' when global font size styles are applied (styles that change default font size for the page)
- allow each theme to define the value that should be used as a base one for `pxToRem` calculations - as each theme might be opinionated about this value (e.g. Semantic is built with base value of `10px` being considered, Teams - with value of `14px`)
- eliminate necessity for theme to change global base font size value - as this loaded theme might have its own "opinion" about what the base size should be for these conversions, and with two or more themes trying to fight for this global font size value the unexpected rendering results might happen (like the ones that we see now for Semantic and Teams themes being loaded on the same page)

Fixes #394


mnajdova (approved):
I like the changes. I have one additional question. Are we exporting from stardust the `pxToRem` service so that it can be used when defining the theme? This is very important for the consumers.

src/themes/teams/utils/index.ts
```diff
@@ -0,0 +1,3 @@
+import { pxToRem } from '../../../lib'
+
+export const teamsPxToRem = (sizeInPx: number) => pxToRem(sizeInPx, 14)
```
bmdalex: shouldn't `14` magic number be a property of the theme object, maybe `defaultFontSize`?

src/themes/teams/siteVariables.ts
```diff
@@ -1,9 +1,4 @@
-import { pxToRem } from '../../lib'
-
-//
-// VARIABLES
-//
-export const htmlFontSize = '10px' // what 1rem represents
```
bmdalex: shouldn't we make this `14` and use it in the new `teamsPxToRem` method?

bmdalex (approved):
couple of comments but agree with general approach

docs/src/app.tsx
```diff
@@ -6,7 +6,7 @@ import { semanticCssOverrides } from './Style'
 import { ThemeContext } from './context/theme-context'
 import Router from './routes'
 
-const semanticStyleOverrides = {
+const semanticStylesOverrideTheme = {
```
levithomason, line 9: 👍

src/themes/teams/utils/index.ts
levithomason: ### Util

Since this util is scoped to the teams theme, I propose not prefixing it with `teams`.  In fact, since the base util will need to use a _required_ key from the theme (html font size), it seems this util should remain a base util provided by Stardust itself so all consumers can rely on the correct font size being used.

### Theme based html font size

Yep, bump.  Reminder that this is required as the value needs to be configurable from the Provider.  This enables the same component to be used in apps with different html font sizes.

I would propose using `htmlFontSize` as the top level key, however.  Since, this is explicitly what the value is.  It is also used by the same name for the same purpose in [Material's theme config](https://material-ui.com/customization/themes/#typography---html-font-size).

levithomason (dismissed):

src/lib/fontSizeUtility.ts
```diff
@@ -1,73 +1,29 @@
...
+ * pxToRem(16, 8)
  * @returns {string} The value converted to the rem.
  */
-export const pxToRem = (value: number = 0): string => {
-  if (!_htmlFontSizeInPx) {
-    _htmlFontSizeInPx = getComputedFontSize()
-  }
-
+export const pxToRem = (value: number, baseSize?: number): string => {
   if (process.env.NODE_ENV !== 'production') {
```
hughreeling: As per 371 - as a general utility, we shouldn't be restricting to positive values.  There are cases for negative values in calculations.

hughreeling (changes requested):

hughreeling (approved):

layershifter:
I've checked `Button` docs page, we run `pxToRem()` there 1313 times. We can reach up to 5x/27x boost with memoization, a little perf test that confirms this: https://jsperf.com/stardust-ui-pxtorem

This is out of scope of this PR, just a note.

kuzhelov:
@layershifter, absolutely agree - actually, this topic was discussed long ago, essentially the same strategy has been suggested (memoise data that was used as part of styles calculation - there is definitely a good opportunity opportunity to utilize (styles, theme)-based caching). The reason it wasn't implemented is that we've agreed on the following plan:
- reach correct and flexible implementation first (which should be proved by real app's usage)
- obtain perf metrics (scaffold infrastructure for that)
- and after that optimise :)

kuzhelov:
## Understand pxToRem() semantics

This section, probably, is the most important one, as it lays down foundation for all the following conclusions made in this post.

### Definition

pxToRem(`pxValue`) => `remValue`,

where `remValue`, when rendered, is equal to `pxValue` **if `htmlFontSize` equals to theme's `fontSize`**.

### Simple case: `htmlFontSize` equals to theme's font size

- theme font size (defined by theme spec) is 14px
- button width is 28px
- pxToRem(28px) => 2rem
  - rendered as 28px if `htmlFontSize` = `themeSpecFontSize` = 14px

### Advanced case: `htmlFontSize` is different from theme's default

Then we should 'tell' this fact to theme. Consider that for aforementioned example we have `htmlFontSize` equal to 18px, value that is different from theme's default (14px). In that case we would expect the following:
- let theme know that page's default `htmlFontSize` is 18px
  - then pxToRem(28px) => `someRemValue`
- rendered: `someRemValue` * `htmlFontSize` = `someRemValue` * 18px = 28px

## API proposal
```jsx
<Provider theme={{
  ...
  fontFaces: ...,

  // optional - 16px is used if not specified. 
  // Should correspond to initial <html>'s font size used on page.
  htmlFontSize: 10
}}>
```

### Transform formula
The following formula should be used to do all the necessary conversions (trivial to infer from examples above):

pxToRem(`pxValue`) = `themeSpecFontSize (14px)` * `pxValue` / `actualHtmlFontSize (18px)`

### Applied to example's case:
- pxToRem(28px) =>  `themeSpecFontSize (14px)` * `pxValue (28px)` / `actualHtmlFontSize (18px)` => `someRemValue (1.55rem)`
- rendered: `someRemValue (1.55rem)` * `htmlFontSize (18px)` = 28px

## Theme nesting case

```jsx
// theme is adjusted to htmlFontSize=10px
// note, value will be 16px if not explicitly specified
// - normally should be defined for topmost Provider only, 
// - as its value SHOULD correspond to page's htmlFontSize.
<Provider theme={{ htmlFontSize: 10, ... }}>
   ... value from topmost provider is passed via context ..

    { /* will use context-defined value (10px) */ }
   <Provider theme={{ ... }}>
      ...
   </Provider>
  
   { /* this Provider's subtree will be rendered with rem sizes as if
    14px would be a value of htmlFontSize.
    - not sure when normally it might be necessary,
    - just maybe in case if relative scales of two themes need to be adjusted. */ }
   <Provider theme={{ htmlFontSize: 14 }}>
     ...
   </Provider>
</Provider>
```

## Thoughts

I really think that biggest problem of this functionality is **how not intuitive its effect might be for a client**. 

Simple example: if I will **increase** font size value for `Provider's` `htmlFontSize`, then, as effect, I will see my components rendered at **smaller** size - this behavior is reverse of what we expect from `HTML`'s font size. This has lead to couple of issues for Material UI folks who have exactly the same functionality introduced, most loud one: https://github.com/mui-org/material-ui/issues/10579

We should really do our best by:
- use better name for `Provider's` prop - pageFontSize, some other?
- properly deliver semantics of `pxToRem` via docs (in one sentence)
- providing good example page (where elements will change size according to theme's one).

## Why not consume `htmlFontSize` from DOM?

Note that **we shouldn't read HTML font size from DOM** as part of our logic (and let user specify this prop manually instead) - this is necessary because otherwise we will either:
- introduce race condition if `htmlFontSize` is rendered from DOM once topmost `Provider` is rendered: it could happen after user will manually scale page's font size (for accessibility purposes)
- remove scaling ability from page - in case if `htmlFontSize` will always be read from DOM on render.

Apparently, Material UI folks have same thoughts on this matter: https://github.com/mui-org/material-ui/pull/10687:

> I would encourage people to use the typography.fontSize option as applying a font size on the <html> element prevents the accessibility feature of the browser to kick in.

src/themes/teams/utils/index.ts
```diff
@@ -0,0 +1,3 @@
+import { pxToRem as basePxToRem } from '../../../lib'
+
+export const pxToRem = (sizeInPx: number) => basePxToRem(sizeInPx, 14)
```
layershifter: Can we avoid there a magic number and define `14` as a separate `const`? 😺 

```suggestion
/* Teams use 14px as a base font size. */
const baseFontSize = 14

export const pxToRem = (sizeInPx: number) => basePxToRem(sizeInPx, baseFontSize)
```

layershifter (approved):
Looks like a great simplification, I definitely like it, nice job 👍

layershifter:
@kuzhelov seems we have a visual regressions, please check them before merge. And we should do something with naming there, because now we have two `pxToRem()`.

docs/src/app.tsx
```diff
@@ -38,15 +38,7 @@ class App extends React.Component<any, AppState> {
     const { themeName } = this.state
     return (
       <ThemeContext.Provider value={this.state}>
-        <Provider
-          theme={mergeThemes(semanticStyleOverrides, themes[themeName], {
-            // adjust Teams' theme to Semantic UI's font size scheme
-            siteVariables: {
-              htmlFontSize: '14px',
```
kuzhelov, line 45: this now should be defined as `<html>` font size in docs' top-level `index.ejs` - to eliminate screen test regressions

PR #373, opened 2018-10-18
fix(Avatar): Adding dark and contrast theme files for Avatar and Status Indicator.

bcalvery:
Modified Avatar to support Dark and Contrast themes

Before:
image
image

After:
image
image

image


src/components/Status/Status.tsx
```diff
@@ -69,14 +69,16 @@ class Status extends UIComponent<Extendable<StatusProps>, any> {
...
     const { icon, renderIcon } = this.props as StatusPropsWithDefaults
     return (
       <ElementType {...rest} className={classes.root}>
         {Icon.create(icon, {
           defaultProps: {
             size: 'tiny',
-            variables: { color: 'white' }, // This is temporary. There is a ToDo to use icon's text/fill color for box-shadow, currently it uses color
+            styles: styles.icon,
+            variables: variables.icon,
+            className: classes.icon,
```
kuzhelov: these `classes.icon` shouldn't be passed, as `classes.icon` contain already processed styles that are transformed to classes (`styles.icon` -> `classes.icon`). We should just pass original `styles.icon` and let `Icon` component process them by itself. This is critical to ensure that these `styles.icon` will be correctly merged with those that could be provided to `icon` shorthand in a following way:

```jsx
<Status icon={{ styles: ... }} />
```

docs/src/examples/components/Avatar/Usage/AvatarUsageExample.shorthand.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react'
+import { Avatar } from '@stardust-ui/react'
+
+const status = { color: 'green', icon: 'check', title: 'Available' }
+
+const AvatarUsageExampleShorthand = () => (
```
kuzhelov, line 6: just a note for future: for all the components we have this section where examples that are rather theme-specific are provided. It is a clear indicator that we should think about mechanisms for theme developers to define custom docs section where theme-specific examples will be provided.

src/themes/teams-dark/components/Avatar/avatarStyles.ts
```diff
@@ -0,0 +1,40 @@
+import { pxToRem } from '../../../../lib'
```
kuzhelov: note that these styles are one-to-one the styles that are used for the default theme. To eliminate code repetition we could just reuse the ones that are used for default theme.

```js
const darkThemeOverrides = {
  componentVariables: {
    ...
    Avatar: { /* .. overriden variables for Avatar component .. */ }
  }
}
..

const darkTheme = mergeThemes(defaultTheme, darkThemeOverrides)
```

src/themes/teams-dark/components/Avatar/avatarVariables.ts
```diff
@@ -0,0 +1,9 @@
+export interface IAvatarVariables {
```
kuzhelov: the type declared for default theme should be reused - this would guarantee consistency of variables across different theme variants

src/themes/teams-dark/components/Status/statusStyles.ts
```diff
@@ -0,0 +1,63 @@
+import { pxToRem } from '../../../../lib'
```
kuzhelov: same story here - we shouldn't replicate the same styles, but rather introduce overrides that are specific to dark theme

kuzhelov (changes requested):
there is lots of code duplication at the moment, the ones that are caused by literal copying the styles and variables of default theme into dark and high-contrast one. Lets address that by using the following approach:

```js
const darkThemeOverrides = {
  componentVariables: {
    ...
    Avatar: { /* .. overriden variables for Avatar component .. */ }
  }
}
..

const darkTheme = mergeThemes(defaultTheme, darkThemeOverrides)
```

kuzhelov (approved):
please, just update CHANGELOG before merging 👍

Issue #345, opened 2018-10-10
[RFC] ContextMenu Component


gopalgoel19:
### Problem description
Building a new component ContextMenu. 
![getimage](https://user-images.githubusercontent.com/11490517/46723885-80087a00-cc96-11e8-87cf-49e37309c4aa.png)

### Proposed solution
As of now, I am planning to build this component using ```Menu```, ```Popup```, ```Avatar``` etc. 
Check progress at #340 

Requesting For Comments on:

1. What is the best approach to build this component?
2. What all functionality should we extend to ```Menu```, if any, to support ```ContextMenu```?

levithomason:
I think this component is great and we should include it.  The props I would imagine would be essentially the Menu's props initially.  This way you can pass any `items` the context menu.  As the Menu API grows to support more features, the Context Menu would also support those features.

gopalgoel19:
Should we have a `menu` prop in `Menu.Item` to render a sub-menu? If yes, what is the difference between a `ContextMenu` component and a `Menu` component with sub-menus?

jurokapsiar:
I would like to discuss starting work on this component - it would solve some very common use cases. We have Menu Button prototype available, but that is far too much code for the consumers to copy. ARIA has a pattern for menu button, other libraries have it as well (https://material-ui.com/components/menus/) and we recently got a requirement for it internally as well. We will also use it heavily in combination with list item and tree item.

The API should include:
- items shorthand (similar to Menu)
- on ['click' | 'hover+click' | 'context']

jurokapsiar:
The alternative would be `<Button menu={{ ... }} />`, let's consider that as well. The important scenario is:
- list item has a context menu button
- by default, when navigating in the list, the focus goes to the list items
- if the list item has context menu button, that would not be focusable, we would however allow users to press the context menu button (secondary action in screen readers) to open the menu
- with that, the oncontext event handler would need to be on the focused list item and the menu would need to be controlled from outside

If the scenario above is doable using Button with menu prop, I am ok with it. If not, we would need a separate Context Menu component.

PR #375, opened 2018-10-19
feat(accessibility):  Adding attachment behavior to handle space/enter key

kolaps33:
# attachmentBehavior

There is requirement that Attachment component is clickable and when user navigate on, then by space/enter key invoke the onclick action. 

What I changed:
Adding attachmentBehavior to specify space/enter key. 
Adding onClick handling to Attachment component.


src/components/Attachment/Attachment.tsx
```diff
@@ -131,7 +142,12 @@ class Attachment extends UIComponent<Extendable<AttachmentProps>, any> {
     } = this.props
 
     return (
-      <ElementType {...rest} className={classes.root}>
+      <ElementType
+        {...rest}
```
jurokapsiar: rest should go last so that it can override defaults

src/components/Attachment/Attachment.tsx
```diff
@@ -169,6 +185,21 @@ class Attachment extends UIComponent<Extendable<AttachmentProps>, any> {
       </ElementType>
     )
   }
+
+  protected actionHandlers: AccessibilityActionHandlers = {
+    performClick: event => this.handleKeyboardClick(event),
+  }
+
+  private handleKeyboardClick = e => {
+    if (e.keyCode === keyboardKey.Spacebar) {
```
jurokapsiar: why is preventDefault called only for spacebar? components should not have any knowledge about the key pressed

jurokapsiar: why is preventDefault called only for spacebar? components should not have any knowledge about the key pressed

jurokapsiar: why is preventDefault called only for spacebar? components should not have any knowledge about the key pressed

jurokapsiar: why is preventDefault called only for spacebar? components should not have any knowledge about the key pressed

sophieH29: We can consider here 2 options
- let users add ```e.preventDefault``` in the handler for ```onClick```
- or 
```js
 performClick: event => (event) = {

   event.preventDefault()
   this.handleClick(event)
},
```
I think we should be safe with a second option as I can't come up with any case that it will break smth (don't think users will render attachment as ```<a>``` for e.g. )

src/components/Attachment/Attachment.tsx
```diff
@@ -169,6 +184,36 @@ class Attachment extends UIComponent<Extendable<AttachmentProps>, any> {
       </ElementType>
     )
   }
+
+  protected actionHandlers: AccessibilityActionHandlers = {
+    performClick: event => this.handleKeyboardClick(event),
+    openContextMenu: event => this.handleOpenContextMenu(event),
```
levithomason: Attachment's don't necessarily have a context menu.  They have an action slot.  It is up to the developer implementing the Attachment what the action is or does.

We can work with the Attachment component API in terms of an `action` but there should be no concept or mention of a `contextMenu` at this point.

stardust-ui/specifications#1

src/components/Attachment/Attachment.tsx
```diff
@@ -169,6 +184,36 @@ class Attachment extends UIComponent<Extendable<AttachmentProps>, any> {
...
+  }
+
+  private handleOpenContextMenu = e => {
+    if (this.props.action && this.props.action['onClick']) {
+      this.props.action['onClick']()
```
levithomason: Not sure if this method will remain after the above comment, however, here are some comments:

1. The `e` should be passed along when calling onClick.
1. Always destructure props.
2. Prefer dot notation access over bracket notation.
3. Using `_.invoke()` to call nested methods can cleanup code.

In this case, since we're invoking a method at a nested path, we can just use invoke directly:

```jsx
_.invoke(this.props, 'action.onClick', e)
```

src/lib/accessibility/Behaviors/Attachment/attachmentBehavior.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Performs click action with 'Enter' and 'Spacebar' on 'root'.
+ */
+
+const attachmentBehavior: Accessibility = (props: any) => ({
+  keyActions: {
+    root: {
+      performClick: {
+        keyCombinations: [{ keyCode: keyboardKey.Enter }, { keyCode: keyboardKey.Spacebar }],
+      },
+      openContextMenu: {
```
levithomason: As noted above, an Attachment doesn't have a context menu but an action slot only.

https://github.com/stardust-ui/specifications/issues/1

levithomason (changes requested):
I am unsure whether or not this PR is applicable, If I am understanding the goal here.  It seems the goal is to forward keyboard invoked clicks (space/enter) on the root down the action slot.  My concern are:

1. this seems to be the need of only one application and not representative of all implementations
1. there is an assumption that the action slot contains a single action that is also a context menu
1. it does not seem intuitive that root events should be forwarded to a child in this way
1. the user's action.onClick is called only when using object shorthand but not for elements nor when using the render callback

It may be that the consumer in this case can render as a button and get these behaviors for free.  They would then handle the click to toggle a Popup.  In fact, they may even be able to use the Attachment as the trigger to a Popup.  I think we should investigate alternative approaches to this before merging.

I've left some comments in-line as well.

src/components/Attachment/Attachment.tsx
jurokapsiar: would something like `performAction` be a better name?

jurokapsiar:
@levithomason we are totally open for ideas on this.

Yes, this might be just one use case, but that is exactly what the concept of behaviors allows us to do - we can provide multiple behaviors for the users to choose from. Question is what should be the default one.

Shift+F10 is a standard for conext menu on Windows, mac does not have a standard shortcut for this so people usually go for the same shortcut independently of the platform. As an alternative, you can make the Action trigger element focusable, but especially in cases when the attachments are organized in a grid and you allow arrow navigation to move focus between them, it becomes hard to focus something that is inside of the focusable element. A specific keyboard shortcut is easier to use and is also more standard (but not an ARIA-standard as this is not defined there).

This solves one specific case when the action is a context menu. We do not have a good solution for other actions yet.

levithomason (changes requested):
Let's update the Attachment component to have an `actions` prop and a `contextMenu` prop.  The Context Menu component is not yet done, #345.  For now, let's just use the behavior against the `action` prop.  Once we have a Context Menu component, we'll update the Attachment to take a list of actions and a context menu, then update the behavior.

levithomason (approved):
Deferring review / approvals to @jurokapsiar

src/lib/accessibility/Behaviors/Attachment/attachmentBehavior.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * @specification
+ * Adds attribute 'tabIndex=0' to 'root' component's part.
+ * Performs click action with 'Enter' and 'Spacebar' on 'root'.
+ */
+
+const attachmentBehavior: Accessibility = (props: any) => ({
+  attributes: {
+    root: {
+      tabIndex: '0',
```
kolaps33: wouldn't be more safer to set rather it for tabIndex=-1 
Now probably focusZone handle it in some way... otherwise tabIndex=0 means it will be in browser Tab order, always..

src/components/Attachment/Attachment.tsx
```diff
@@ -94,6 +136,34 @@ class Attachment extends UIComponent<Extendable<AttachmentProps>, any> {
...
   }
+
+  protected actionHandlers: AccessibilityActionHandlers = {
+    performClick: event => this.handleKeyboardClick(event),
+  }
+
+  private handleKeyboardClick = e => {
+    e.stopPropagation()
+    this.handleClick(e)
+  }
```
sophieH29, line 147: I would not move this to separate method, just leave it as

```js
performClick: event => {
   event.stopPropagation()
   this.handleClick()
}
```
I think ```handleKeyboardClick``` can be confusable + too many things like ```handleSmth``` already here

docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
```diff
@@ -72,23 +73,57 @@ function createMessageContent(message: MessageData): ShorthandValue {
...
       accessibility={popupFocusTrapBehavior}
       trigger={
-        <Button aria-label="More attachment options" iconOnly circular icon="ellipsis horizontal" />
+        <Button
+          aria-label="More attachment options"
+          iconOnly
+          circular
+          icon="ellipsis horizontal"
+          onClick={e => e.stopPropagation()}
+          onKeyDown={stopEnterSpacePropagation}
```
jurokapsiar: @layershifter can you pls look at this? All the wierdnes with stopPropagation needs to be done just because there is a Attachment `<div>` element and a `<Button>` inside of it. Both are clickable, but when you click on the button, it should not trigger the click on the parent `<div>`. Similarly, when user presses enter or space on the button, it should only click on that button, not on `<div>`. However, TAB key needs to be propagated because FocusZone has handlers that react on it. This requires a lot of knowledge on the consumer side. I'm wondering if there is a generic solution for these cases. I was also thinking about performing the action depending on the event target, but I think that is equally strange. What do you think?

fyi @sophieH29

src/components/Attachment/Attachment.tsx
```diff
@@ -94,6 +136,34 @@ class Attachment extends UIComponent<Extendable<AttachmentProps>, any> {
...
+  private handleClick = (e: React.SyntheticEvent) => {
+    const { onClick, disabled } = this.props
+
+    if (disabled) {
+      e.preventDefault()
+      return
+    }
+
+    if (onClick) {
+      onClick(e, this.props)
```
layershifter: Should be called via `_.invoke()` and will be able to `if {}`

docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
layershifter: I previously did something like this, it's a little more obvious:
```jsx
const stopPropagationOnKeys = (keys: number[]) => (e: Event) => {
   if (keys.includes(keyboardKey.getCode(e))) {
     e.stopPropagation()
   }
}

//

onKeyDown={stopPropagationOnKeys([keyboardKey.Enter, keyboardKey.Spacebar])}
```

src/lib/accessibility/Behaviors/Attachment/attachmentBehavior.ts
```diff
@@ -0,0 +1,25 @@
+import { Accessibility } from '../../types'
+import * as keyboardKey from 'keyboard-key'
+
+/**
+ * @specification
+ * Adds attribute 'tabIndex=0' to 'root' component's part.
+ * Performs click action with 'Enter' and 'Spacebar' on 'root'.
+ */
+
```
layershifter: Useless empty line?

docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
```diff
@@ -72,23 +73,57 @@ function createMessageContent(message: MessageData): ShorthandValue {
...
+        },
       ]}
       vertical
       pills
     />
   )
 
+  const stopEnterSpacePropagation = (e: Event) => {
+    const code = keyboardKey.getCode(e)
+    if (code === keyboardKey.Enter || code === keyboardKey.Spacbar) {
```
layershifter: Please fix a typo (`SpacBar`)

PR #376, opened 2018-10-19
feat(factories): createShorthandFactory API update and removing HTML factories

bmdalex:
# feat(factories): `createShorthandFactory` API update and removing HTML factories

This PR focuses on:
- removing regular HTML factories from `lib/factories.tsx`
- updating `createShorthandFactory` API to use `mappedProp` as string instead of `mapValueToProps` function
- adding `Slot.create` and `Slot.createHTMLElement` functions
- updating usage
- adding UTs

### TODO

- [x] Conformance test
- [x] Update the CHANGELOG.md

# API Proposal

```tsx
import { createHTMLInput } from 'src/lib'

// usage
createHTMLInput(
  'text',
  { defaultProps: { ... }, overrideProps: { ... } }
)
```

### New

```tsx
import Slot from 'src/components/Slot'

// usage
Slot.createHTMLElement(
  shorthandProp,
  { defaultProps: { as: 'input', ... }, overrideProps: { ... } }
)
```

The advantages of the new API is that it will create a Stardust `Slot` component and render it as an `input` element (see `as` property in the usage), which means it will have all Stardust features for:
- behavior
- accessibility
- styling


mnajdova:
Just additional thought. It would be really beneficial if the user can use the Slot as any other stardust component, because there will be out of the box support for styles, variables, rtl etc. Meaning in any place, where at this moment we are using div in the prototypes, we would use the Slot component.

bmdalex:
@mnajdova 
totally agree on that but I need some evidence (through user scenarios) that there is a real need for it; is there anything preventing the user to do that now? or is it just the fact that we don't export `Slot` in `index.tsx`?

mnajdova:
The benefit is that **there will be out of the box support for styles, variables, rtl etc,** which is very important. Otherwise, using divs, labels and any other HTML elements doesn't support this. I already used the Slot in some of the prototypes and I consider it really valuable from user's perspective. Not sure that now the user can use the Slot from the library as we don't export it as a regular component.

src/components/Slot/Slot.tsx
```diff
@@ -56,7 +56,16 @@ class Slot extends UIComponent<Extendable<ISlotProps>, any> {
...
+
+  // ============================================================
+  // HTML Factories
+  // ============================================================
+  static createHTMLDivision = createSlotFactory('div', children => ({ children }))
+  static createHTMLIframe = createSlotFactory('iframe', src => ({ src }))
+  static createHTMLImage = createSlotFactory('img', src => ({ src }))
   static createHTMLInput = createSlotFactory('input', type => ({ type }))
+  static createHTMLLabel = createSlotFactory('label', children => ({ children }))
+  static createHTMLParagraph = createSlotFactory('p', children => ({ children }))
```
levithomason: Per last sync, let's not create statics on Slot.  We'll also not have dedicated factories for each HTML element. 
 This is because we can use the `as` prop instead for all these cases.

The proper `mapValueToProps` function will be chosen by the factory according to the final computed element type.  This will be done in the `createShorthand()` method in `factories.tsx`.

levithomason (changes requested):

src/components/Slot/Slot.tsx
bmdalex: changes were made @levithomason can u please another look?

bmdalex:
In reply to @mnajdova 's comment, do we want to expose the `Slot` component as a component to users?

@levithomason @kuzhelov

bmdalex:
@mnajdova @kuzhelov @levithomason 
bump for reviewing this

src/index.ts
```diff
@@ -88,6 +88,8 @@ export {
 
 export { default as Segment, SegmentProps } from './components/Segment/Segment'
 
+export { default as Slot } from './components/Slot/Slot'
```
miroslavstastny: `SlotProps` must be exported as well.
not sure about `createShorthandFactory`

src/components/Slot/Slot.tsx
```diff
@@ -18,12 +18,29 @@ export interface SlotProps {
...
+  ['div', 'children'],
+  ['iframe', 'src'],
+  ['img', 'src'],
+  ['input', 'type'],
+  ['label', 'children'],
+  ['p', 'children'],
+] as [HTMLTag, ShorthandProp][])
+
+const getMapValueFn = (propType: HTMLTag) => {
+  const mappedProp = htmlFactoryMapping.get(propType) || 'content'
```
levithomason: Since this is for HTML elements, it seems it should default to `children` instead of `content`, correct?  Almost all HTML support `children` but none support `content`.

src/index.ts
bmdalex: in the end we decided not to export anything; this will be done at a later time in a dedicated PR

src/components/Attachment/Attachment.tsx
```diff
@@ -162,7 +163,7 @@ class Attachment extends UIComponent<Extendable<AttachmentProps>, any> {
           </div>
         )}
         {!_.isNil(progress) &&
-          createHTMLDivision('', {
+          createSlotFactory('div')(progress, {
```
levithomason: I would not expect every call of this to create a new factory.  Seems the factory should internally handle mapping the appropriate mapValToProps function to the element type instead of creating new factory functions on every invocation.

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `iconPosition` property to `Input` component @mnajdova ([#442](https://github.com/stardust-ui/react/pull/442))
 - Add `color`, `inverted` and `renderContent` props and `content` slot to `Segment` component @Bugaa92 ([#389](https://github.com/stardust-ui/react/pull/389))
 - Add focus trap behavior to `Popup` @kuzhelov ([#457](https://github.com/stardust-ui/react/pull/457))
+- Change `createShorthandFactory` API, added `as` prop functionality in order to deprecate HTML factories @Bugaa92 ([#376](https://github.com/stardust-ui/react/pull/376))
```
levithomason: Since this change is not public facing, we don't need to add a changelog entry.

bmdalex: removed, thanks

levithomason (approved):
Merge after removing `Slot.createHTMLElement()` 👍

PR #377, opened 2018-10-19
feat(Input): expose Input's <input> element by inputRef

silviuaavram:
Exposed the <input> element through inputRef property of the Input component.


bmdalex (approved):
looks good to me

miroslavstastny (approved):

src/components/Input/Input.tsx
```diff
@@ -201,6 +205,12 @@ class Input extends AutoControlledComponent<Extendable<IInputProps>, IInputState
     })
   }
 
+  private handleInputRef = (inputNode: HTMLElement) => {
+    this.inputDomElement = inputNode as HTMLInputElement
```
levithomason, line 209: Let's also keep the naming convention of `this.ref` for the root and `this.*ref` for other slots.  This then would be `this.inputRef`.

levithomason (changes requested):

src/components/Input/Input.tsx
bmdalex: looks like we're in some sort of contradiction here..

this used to be called `this.inputRef` and was changed to `this.inputDomElement` by me in #326 at @kuzhelov 's [suggestion](https://github.com/stardust-ui/react/pull/326#discussion_r226376726):
screen shot 2018-10-22 at 18 17 48

@kuzhelov @levithomason 
can we get consensus and unblock @silviuavram ?

silviuaavram: I was following the way it was being handled in Portal.tsx . There, the actual HTMLElement is called triggerNode/portalNode. So at the very best, this may be renamed to `this.inputNode` . @Bugaa92 @kuzhelov what do you think?

kuzhelov: totally agree with @silviuavram's point - this variable definitely shouldn't be named as `*Ref`, as it is _not_ a ref. Consider the difference that is introduced by `ref` semantics (taking React ref as an example):

```
const ref = ref
const element = ref ------ current -------> DOM element
```
vs
```
const element = DOM element
```

While `ref`, as implied by its name, provides additional level of indirection for accessing DOM element, the object that is stored in the `inputDomElement` variable is DOM element itself - thus, it definitely shouldn't be named as `*ref` - more than that, the name that is currently used is quite precise in terms of reflecting what the value of this variable stays for.

kuzhelov (approved):

PR #378, opened 2018-10-19
fix(provider): fix for rendering components outside provider

### LINKED ISSUE #372: Rendering a Component outside of Provider should fail with meaningful error
### Steps
Create a new application, create a component (`<Button />`) without wrapping that by `<Provider>`.

### Expected Result
Component rendering fails with meaningful error (or warning) 

### Actual Result
Component dom renders, no styles are applied 💣

### Version
0.9.1

bmdalex:
# Provider

This PR is a fix proposal that closes #372.

### TODO

- [x] Update the CHANGELOG.md

## Proposed solution

Meaningful console error with link pointing to documentation where user can find examples of correct usage:
screen shot 2018-10-23 at 12 47 56


src/lib/renderComponent.tsx
```diff
@@ -136,6 +152,10 @@ const renderComponent = <P extends {}>(
   return (
     <FelaTheme
       render={(theme: IThemePrepared) => {
+        if (_.isEmpty(theme) || !theme.provided) {
```
kuzhelov: lets focus on the case where there is some non-empty `theme` object provided and `theme.provided` is false. When this could happen? Currently we have default team being **non-empty** object (that automatically passes `_.isEmpty` check) - and there is no need to additionally check for `provided`, as this is an equivalent of `emptyTheme` being used.

I would propose to remove `provided` prop and use just `_.isEmpty` check - this should be enough. Please, feel free to suggest the cases if I am wrong with this assumption.

src/lib/mergeThemes.ts
```diff
@@ -184,6 +184,7 @@ const mergeThemes = (...themes: IThemeInput[]): IThemePrepared => {
     fontFaces: [],
     staticStyles: [],
     icons: {},
+    provided: true,
```
kuzhelov: seems that we can avoid this property to be introduced - it would be enough just to check that `theme` object is not empty for theme context consumer

src/lib/renderComponent.tsx
```diff
@@ -118,6 +119,21 @@ const renderWithFocusZone = (render, focusZoneDefinition, config, focusZoneRef):
...
+const getProviderNotUsedWarningElement = (displayName: string) => {
+  console.error(`Incorrect usage. Your component needs to be wrapped using the Provider component:
+
+import * as React from 'react';
+import { ${displayName}, Provider, themes } from '@stardust-ui/react';
+
+<Provider theme={themes.teams}>
+  <${displayName} />
+</Provider>
+`)
```
levithomason: The error is not so much that the component is not wrapped, but that the Provider is missing.  Let's point that out and link them to the quick start docs.

src/lib/renderComponent.tsx
```diff
@@ -118,6 +119,21 @@ const renderWithFocusZone = (render, focusZoneDefinition, config, focusZoneRef):
   return render(config)
 }
 
+const getProviderNotUsedWarningElement = (displayName: string) => {
+  console.error(`Incorrect usage. Your component needs to be wrapped using the Provider component:
```
levithomason: We should probably also implement a cache check so we don't spam the console for every component on the page.

levithomason:
@Bugaa92 Note, if you slightly adjust the PR description then this PR will automatically close the associated issue once it is merged.  See here:

https://help.github.com/articles/closing-issues-using-keywords/

src/lib/renderComponent.tsx
bmdalex: removed `provided` flag

src/lib/mergeThemes.ts
bmdalex: removed `provided` flag

src/lib/renderComponent.tsx
bmdalex: done

src/lib/renderComponent.tsx
bmdalex: done

src/lib/constants.ts
```diff
@@ -0,0 +1,7 @@
+const docSiteUrl = 'https://stardust-ui.github.io/react'
+const quickStartUrl = `${docSiteUrl}/quick-start`
```
levithomason: Let's not dupe constants that already exist in the repo.  Let's pull these from `/docs` into `/src`.  The docs are allowed to import src files.

src/lib/providerMissingHandler.ts
```diff
@@ -0,0 +1,15 @@
+import ct from './constants'
```
levithomason: nit: no abbrev please :)

src/lib/constants.ts
bmdalex: done, pulled all constants together

src/lib/providerMissingHandler.ts
bmdalex: fixed

bmdalex:
@levithomason comments were addressed

kuzhelov (approved):

PR #381, opened 2018-10-19
fix(Button): Implementing Contrast Theme and Updating Dark Theme

bcalvery:
# Adding Contrast Theme styles for Button

Contrast:
image
image

Dark:
image
image


alinais (approved):

mnajdova (approved):

kuzhelov (approved):
👍

src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -216,34 +256,7 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps & ButtonState, any> = {
...
-        padding: 0,
-        borderRadius: circularRadius,
-      }),
-
-      ...(fluid && {
-        width: '100%',
-        maxWidth: '100%',
-      }),
-
+      // Overrides for "disabled" buttons
```
levithomason, line 258: nit: next line "says" the same thing.

levithomason (approved):
see one nit

src/themes/teams/components/Button/buttonStyles.ts
alinais: We have these comments in all the file, not only here. We can consider it as part of some cleanup task.

PR #384, opened 2018-10-22
fix(Icon): Color & docs update for Teams theme

codepretty:
Updating the color for icon Teams theme.. An icon will inherit the current color of the font unless a new color is explicitly passed to icon.

Added new **Color** section to the doc site under **Variations**
image


src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -87,43 +76,51 @@ const iconStyles: ComponentSlotStylesInput<IconProps, any> = {
   }): ICSSInJSStyle => {
     const iconSpec = theme.icons[name]
     const isFontBased = !iconSpec || !iconSpec.isSvg
+    const iconColor = v.color || 'currentColor'
```
kuzhelov, line 79: 👍

kuzhelov (approved):

docs/src/examples/components/Icon/Usage/IconButtonExample.shorthand.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+import { Icon, Button, Text } from '@stardust-ui/react'
+
+const IconExampleButton = () => (
+  <div>
+    <Button type="primary" icon iconPosition="before">
```
kuzhelov: actually, this is not a _shorthand_ example. Speaking of the scenario considered - seems that we should rather introduce it to the `Button Usage` section. And final note - there is no much need in `icon` and `iconPosition` props being introduced for `Button`, they affect just some layout aspects in this context. Seems that the same effect is better to be achieved by using `Button` props only.

kuzhelov (commented):
please, see comments. Thanks!

docs/src/examples/components/Icon/Usage/IconButtonExample.shorthand.tsx
codepretty: I agree button with icon should be part of Button docs. My goal with this is to show how icons can be used and inherit color (via currentColor) from their context (body text or containing component). Do you have another suggestion on what ways I could show that? Or do you not believe it is necessary?

kuzhelov: let me, please, try to quickly address it on my local env - will back right after..

kuzhelov: let me, please, bring the results back. 

#### TL;DR
While this example could seem to be an appealing one, at the same time it is **Teams theme specific**. Imagine that there is some other theme applied (with theme switcher in docs), the one where there is no `color` variable for `Button` and no `currentColor` provided for `Icon`'s styles (those are the aspects of Teams theme) - in this case this example will break. 

-----

First thing to note is that it would be possible to achieve the same goal this example provides by using the following code:

```jsx
const IconExampleButton = () => (
  <div>
    <Button type="primary" icon={{ name: "call", size: 'large' }} content="Call now" />
    <Button type="secondary" icon="home" content="Phone home" />
    <Button type="primary" disabled icon="home" content="Disabled button with icon" />
  </div>
)
```

Most critical thing to note is that _if there are children being explicitly provided to the component, then its props won't affect anyhow the rendered tree of this component_ - `children` are always rendered 'as is'. Speaking of the `Button` example we originally had

```jsx
 <Button type="secondary" icon iconPosition="before">
      <Icon name="home" xSpacing="after" />
      <Text content="Phone home" />
    </Button>
``` 

Please, note that as a consequence of this principle is that both `icon` and `iconPosition` _are simply ignored and make any effect to the rendered tree of a `Button`_, as long as there are children (`<Icon>` and `<Text>`).

------

## Maybe a better example

Please, consider the following example that, probably, better reflects the idea you would like to present:

```jsx
// note that 'color' which is specified for Button component
// is inherited by Icon
<>
    <Button variables={{ color: 'red' }} icon='home' content='Phone home' />
    <Button variables={{ color: 'orange' }} icon='home' content='Phone home' />
    <Button variables={{ color: 'green' }} icon='home' content='Phone home' />
    <Button variables={{ color: 'lightblue' }} icon='home' content='Phone home' />
</>
```

image

-----

## Final thoughts
While this example could seem to be an appealing one, at the same time it is Teams theme specific. Imagine that there is some other theme applied, the one where there is no `color` variable for `Button` and no `currentColor` provided for `Icon`'s styles - in this case this example will break. Given that we currently don't support any way to introduce theme-specific examples, I would rather suggest to omit introducing this at least for now.

Please, @codepretty, @levithomason, let me know about your thoughts on that.

kuzhelov (commented):
lets, please, agree on the example provided before merging these changes - other than that am totally for merging it 👍

docs/src/examples/components/Icon/Usage/IconButtonExample.shorthand.tsx
codepretty: Thanks for your detailed feedback. Just talked with @levithomason and have decided to pull this example out of this component and do a separate PR for the Button component for different colors and to have some of those buttons show using an icon.

kuzhelov (approved):
please, update the CHANGELOG before merge - thanks!

PR #389, opened 2018-10-23
feat(segment): add inverted prop and improved styles

bmdalex:
# Segment

This PR introduces following changes for `Segment` component:
- `inverted` prop;
- `color` variable;
- `content` slot and `renderContent` method;
- new styles. 

### TODO

- [ ] Conformance test
- [ ] Stardust base theme
- [x] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [x] Confirm RTL usage
- [x] Update the CHANGELOG.md

# API Proposal

## Default

```jsx
<Segment
  content="The elevator to success is out of order. You’ll have to use the stairs."
/>
```
renders:

screen shot 2018-10-23 at 11 38 14

## `color` variable

A segment can different colors. Use `color` variable to see the effect.

```jsx
<Segment
  content="Colored segment."
  variables={siteVars => ({ color: siteVars.brand })}
/>
```
renders
screen shot 2018-10-23 at 11 38 20

## `inverted` with `color` variable

A segment can have its colors inverted for contrast. Use `color` variable to see the effect.

```jsx
<Segment
  inverted
  content={<Text content="Colored inverted segment." styles={{ color: 'white' }} />}
  variables={siteVars => ({ color: siteVars.brand })}
/>
```
renders
screen shot 2018-10-23 at 11 38 25

### DIFF
--- a/docs/src/examples/components/Segment/Variations/SegmentExampleInverted.shorthand.tsx
+++ b/docs/src/examples/components/Segment/Variations/SegmentExampleInverted.shorthand.tsx
@@ -1,0 +1,16 @@
+import * as React from 'react'
+import { Segment, Text } from '@stardust-ui/react'
+
+const SegmentExampleInvertedShorthand = () => (
+  <div>
+    <Segment content="Colored segment." color="purple" />
+    <br />
+    <Segment
+      inverted
+      content={<Text content="Colored inverted segment." styles={{ color: 'white' }} />}
+      color="purple"
+    />
+  </div>
+)
+
+export default SegmentExampleInvertedShorthand
--- a/docs/src/examples/components/Segment/Variations/SegmentExampleInverted.shorthand.tsx
+++ b/docs/src/examples/components/Segment/Variations/SegmentExampleInverted.shorthand.tsx
@@ -1,15 +1,11 @@
 import * as React from 'react'
-import { Segment, Text } from '@stardust-ui/react'
+import { Segment } from '@stardust-ui/react'
 
 const SegmentExampleInvertedShorthand = () => (
   <div>
     <Segment content="Colored segment." color="purple" />
     <br />
-    <Segment
-      inverted
-      content={<Text content="Colored inverted segment." styles={{ color: 'white' }} />}
-      color="purple"
-    />
+    <Segment inverted content="Colored inverted segment" color="purple" />
   </div>
 )
 
--- a/docs/src/examples/components/Segment/Variations/SegmentExampleInverted.tsx
+++ b/docs/src/examples/components/Segment/Variations/SegmentExampleInverted.tsx
@@ -1,0 +1,14 @@
+import * as React from 'react'
+import { Segment, Text } from '@stardust-ui/react'
+
+const SegmentExampleInvertedShorthand = () => (
+  <div>
+    <Segment color="purple">Colored segment.</Segment>
+    <br />
+    <Segment inverted color="purple">
+      <Text styles={{ color: 'white' }}>Colored inverted segment.</Text>
+    </Segment>
+  </div>
+)
+
+export default SegmentExampleInvertedShorthand
--- a/docs/src/examples/components/Segment/Variations/SegmentExampleInverted.tsx
+++ b/docs/src/examples/components/Segment/Variations/SegmentExampleInverted.tsx
@@ -1,12 +1,12 @@
 import * as React from 'react'
-import { Segment, Text } from '@stardust-ui/react'
+import { Segment } from '@stardust-ui/react'
 
 const SegmentExampleInvertedShorthand = () => (
   <div>
     <Segment color="purple">Colored segment.</Segment>
     <br />
     <Segment inverted color="purple">
-      <Text styles={{ color: 'white' }}>Colored inverted segment.</Text>
+      Colored inverted segment
     </Segment>
   </div>
 )
--- a/src/components/Segment/Segment.tsx
+++ b/src/components/Segment/Segment.tsx
@@ -1,13 +1,17 @@
 import * as React from 'react'
 import * as PropTypes from 'prop-types'
 import { customPropTypes, UIComponent, childrenExist } from '../../lib'
-import { Extendable } from '../../../types/utils'
+import { Extendable, ShorthandValue, ShorthandRenderFunction } from '../../../types/utils'
 import { ComponentVariablesInput, ComponentSlotStyle } from '../../themes/types'
+import Slot from '../Slot/Slot'
 
 export interface SegmentProps {
   as?: any
   className?: string
-  content?: any
+  color?: string
+  content?: ShorthandValue
+  inverted?: boolean
+  renderContent?: ShorthandRenderFunction
   styles?: ComponentSlotStyle<SegmentProps, any>
   variables?: ComponentVariablesInput
 }
@@ -27,8 +31,22 @@ class Segment extends UIComponent<Extendable<SegmentProps>, any> {
     /** Additional CSS class name(s) to apply.  */
     className: PropTypes.string,
 
+    /** A segment can have different colors */
+    color: PropTypes.string,
+
     /** Shorthand for primary content. */
-    content: PropTypes.any,
+    content: PropTypes.contentShorthand,
+
+    /** A segment can have its colors inverted for contrast. */
+    inverted: PropTypes.bool,
+
+    /**
+     * A custom render function the content slot.
+     * @param {React.ReactType} Component - The computed component for this slot.
+     * @param {object} props - The computed props for this slot.
+     * @param {ReactNode|ReactNodeArray} children - The computed children for this slot.
+     */
+    renderContent: PropTypes.func,
 
     /** Additional CSS styles to apply to the component instance.  */
     styles: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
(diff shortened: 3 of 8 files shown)


docs/src/examples/components/Segment/Variations/SegmentExampleInverted.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react'
+import { Segment, Text } from '@stardust-ui/react'
+
+const SegmentExampleInvertedShorthand = () => (
+  <div>
+    <Segment variables={siteVars => ({ color: siteVars.brand })}>Colored segment.</Segment>
```
kuzhelov: would suggest to at least provide fallback color in case if `brand` won't be defined in `siteVars` for some reason

docs/src/examples/components/Segment/Variations/SegmentExampleInverted.shorthand.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { Segment, Text } from '@stardust-ui/react'
+
+const SegmentExampleInvertedShorthand = () => (
+  <div>
+    <Segment content="Colored segment." variables={siteVars => ({ color: siteVars.brand })} />
+    <br />
+    <Segment
+      inverted
+      content={<Text content="Colored inverted segment." styles={{ color: 'white' }} />}
+      variables={siteVars => ({ color: siteVars.brand })}
```
kuzhelov: would suggest to use some vivid fallback color here

kuzhelov (approved):
please, consider to use fallback options for `siteVars.brand` for those cases where those are not provided. I would even suggest to use some hardcoded colors for now, to make it easier for the client of these examples to get the idea

docs/src/examples/components/Segment/Variations/SegmentExampleInverted.shorthand.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react'
+import { Segment, Text } from '@stardust-ui/react'
+
+const SegmentExampleInvertedShorthand = () => (
+  <div>
+    <Segment content="Colored segment." variables={siteVars => ({ color: siteVars.brand })} />
+    <br />
+    <Segment
+      inverted
+      content={<Text content="Colored inverted segment." styles={{ color: 'white' }} />}
```
kuzhelov: what if `brand` color would be `white` for some theme as well? would better suggest just to hardcode colors for now

docs/src/examples/components/Segment/Variations/SegmentExampleInverted.tsx
bmdalex: thx, done

docs/src/examples/components/Segment/Variations/SegmentExampleInverted.shorthand.tsx
bmdalex: thx, done

docs/src/examples/components/Segment/Variations/SegmentExampleInverted.shorthand.tsx
bmdalex: I do not fully agree with hardcoding; we should use `siteVars` in order to leverage switching themes capabilities; I think we should create an issue on using only predefined colors in examples after we have a default theme; what do you think? @kuzhelov @levithomason 
for now I'll hardcode colors

levithomason: agree on both comments here.  we should hard code for now only because we lack a sufficient site variables interface.  once we have something for background and foreground in the site variables interface, we should consistently use only site variables.

docs/src/examples/components/Segment/Variations/SegmentExampleInverted.shorthand.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react'
+import { Segment, Text } from '@stardust-ui/react'
+
+const SegmentExampleInvertedShorthand = () => (
+  <div>
+    <Segment content="Colored segment." variables={{ color: 'purple' }} />
```
levithomason: Color in this case is a primary and common concern for the Segment.  It should then be available as a prop:

```jsx
<Segment color='purple' />
```

levithomason (changes requested):

docs/src/examples/components/Segment/Variations/SegmentExampleInverted.shorthand.tsx
bmdalex: done

docs/src/examples/components/Segment/Variations/SegmentExampleInverted.shorthand.tsx
bmdalex: this would be new API, we don't have `color` as prop in the API of any component...
@levithomason @kuzhelov

src/themes/teams/components/Segment/segmentStyles.ts
```diff
@@ -1,14 +1,23 @@
...
-      border: '1px solid rgba(34,36,38,.15)',
-      borderRadius: variables.borderRadius,
-      boxShadow: '0 1px 2px 0 rgba(34,36,38,.15)',
+      padding: v.padding,
+      background: v.background,
+      borderRadius: v.borderRadius,
+      boxShadow: '0 1px 1px 1px rgba(34,36,38,.15)',
+      ...(inverted
+        ? {
+            background: v.color,
```
mnajdova: Is it okay that the inverted segment doesn't have border color? Should be consider having two variables for colors (color and borderColor) and switch there when the inverted prop is applied?

src/components/Segment/Segment.tsx
```diff
@@ -27,6 +30,18 @@ class Segment extends UIComponent<Extendable<SegmentProps>, any> {
...
+    inverted: PropTypes.bool,
+
+    /**
+     * A custom render function the content slot.
+     *
+     * @param {React.ReactType} Component - The computed component for this slot.
+     * @param {object} props - The computed props for this slot.
+     * @param {ReactNode|ReactNodeArray} children - The computed children for this slot.
+     */
+    renderContent: PropTypes.func,
```
mnajdova, line 49: +1

src/themes/teams/components/Segment/segmentStyles.ts
bmdalex: discussed with @mnajdova  this is according to spec

bmdalex:
@mnajdova @kuzhelov @levithomason 
bump for reviewing this

mnajdova (approved):

docs/src/examples/components/Segment/Variations/SegmentExampleInverted.shorthand.tsx
Bugaa92: going to resolve for now we can reopen if we realize we didn't follow the right approach

mnajdova: I would leave it as a variable until we figure out if we want to expose the color as a prop to different components.

levithomason: Per our convo:

I don't this as adding new API, but simply one of the many missing design language terms in Stardust's component APIs:

https://react.semantic-ui.com/elements/button/#variations-colored
https://react.semantic-ui.com/elements/header/#variations-colored
https://react.semantic-ui.com/elements/label/#variations-colored
https://react.semantic-ui.com/elements/segment/#variations-colored
https://react.semantic-ui.com/collections/menu/#variations-colored
https://react.semantic-ui.com/collections/message/#variations-color
https://react.semantic-ui.com/collections/table/#variations-colors
https://react.semantic-ui.com/views/card/#variations-colored
https://react.semantic-ui.com/views/statistic/#variations-colored
https://react.semantic-ui.com/modules/progress/#variations-color

Our props are intended to be a design language based on natural language principles.  Design concepts that are common for a component will be props to that component, such as the "color" of a component.

Variables can differ between themes since themes can add their own variables.  They are not necessarily part of the common design language.  Having a prop means all themes can rely on the design language term being available and create a theme style for it.

bmdalex: added `color` prop per discussion, @levithomason can you review again?

docs/src/examples/components/Segment/Variations/SegmentExampleInverted.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react'
+import { Segment, Text } from '@stardust-ui/react'
+
+const SegmentExampleInvertedShorthand = () => (
+  <div>
+    <Segment color="purple">Colored segment.</Segment>
+    <br />
+    <Segment inverted color="purple">
+      <Text styles={{ color: 'white' }}>Colored inverted segment.</Text>
```
levithomason: For all examples, remove Text component and use plain text.
Update Segment styles to set the appropriate font color.

The examples will then work similar to SUIR:
https://react.semantic-ui.com/elements/segment/#types-segment

We don't yet have "inverted" concepts in our palette for text colors and other colors.  So for now, just hard code it to '#eee' to avoid passing "bodyBackground" to a font color.  We'll resolve these problems in #423.

levithomason (approved):
Please merge after addressing comments!

docs/src/examples/components/Segment/Variations/SegmentExampleInverted.tsx
bmdalex: done, will merge

PR #390, opened 2018-10-23
fix(button): removed dotted outline in firefox

bmdalex:
# Button

This PR removes the dotted outline in Firefox for `Button` component. Repro:
- go to doc site in `Firefox`;
- navigate to the buttons using keyboard navigation;
- observe focus outline.

### Before
screen shot 2018-10-22 at 23 29 13

### After
screen shot 2018-10-23 at 11 55 18


src/themes/teams/staticStyles/normalizeCSS.ts
```diff
@@ -207,17 +207,6 @@ button::-moz-focus-inner,
...
 
-/**
- * Restore the focus styles unset by the previous style.
- */
-
-button:-moz-focusring,
-[type="button"]:-moz-focusring,
-[type="reset"]:-moz-focusring,
-[type="submit"]:-moz-focusring {
-  outline: 1px dotted ButtonText;
```
bmdalex: @levithomason is this safe?

src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -81,6 +81,11 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps & ButtonState, any> = {
       verticalAlign: 'middle',
       cursor: 'pointer',
 
+      // @Bugaa92: temporary fix until semantic-ui-css is not referenced anymore
```
bmdalex: @levithomason 
@kuzhelov 
is there a plan to rewrite the doc site in stardust so we can remove `semantic-ui-css`?

bmdalex (commented):
@levithomason 
@kuzhelov 
pls take a look at my comments

src/themes/teams/components/Button/buttonStyles.ts
kuzhelov: is the same one currently used in Teams?

src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -81,6 +81,11 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps & ButtonState, any> = {
...
+      ':-moz-focusring': {
+        outline: 0,
```
kuzhelov: is the same one currently used in Teams to achieve the same effect?

src/themes/teams/components/Button/buttonStyles.ts
bmdalex: N.A. anymore, removed this change

src/themes/teams/components/Button/buttonStyles.ts
bmdalex: yes, checked

kuzhelov (approved):

notandrew (approved):

notandrew (changes requested):
I don't think the changes should be made within normalize -- if we update to a new version of normalize, this change will be lost.  Should it go into CustomCSS or siteVariables?

kuzhelov:
great point, thanks @notandrew - that's definitely would be a better approach 👍

bmdalex:
@kuzhelov 
addressed @notandrew 's comments and moved the change in the `semanticCssOverrides` overrides file as per our discussion; ready to merge?

docs/src/Style.ts
```diff
@@ -37,4 +37,11 @@ export const semanticCssOverrides = `
     color: rgba(255,255,255,.7)!important;
     cursor: not-allowed!important;
   }
+
+  button:-moz-focusring,
+  [type="button"]:-moz-focusring,
+  [type="reset"]:-moz-focusring,
+  [type="submit"]:-moz-focusring {
+    outline: 0;
+  }
```
levithomason, line 46: This will only work in the doc site, is that the intent?  If this should be the style for all usages of the Teams theme, including outside of the doc site by consumers, then this should be a `staticStyle` in the Teams theme.  This way, everywhere the Teams theme is used, there is no outline in Firefox.

kuzhelov: for now yes, the intent was just to address doc site issue and propose a question of whether the same change should be applied for Teams theme.

PR #398, opened 2018-10-24
feat(Accessibility): Make Grid keyboard navigable by adding Grid Behavior

sophieH29:
This PR adds a possibility to navigate Grid items with arrow buttons.
What was added:
- extended Grid's API, so now it can process ```accessibility``` prop. (though, a default value is ```defaultBehavior```)
- added ```gridBehavior``` with ```FocusZone``` circular bi-directional navigation
- added example of keyboard navigable Grid component, which uses ```gridBehavior``` for that purpose.

TODO:
- update CHANGELOG.md


docs/src/examples/components/Grid/Variations/GridExampleKeyboardNavigable.shorthand.tsx
```diff
@@ -0,0 +1,108 @@
+import React from 'react'
+import { Grid, Image, Button, gridBehavior } from '@stardust-ui/react'
+import _ from 'lodash'
+
+const images = [
```
bmdalex: can you make this an array to save space?

docs/src/examples/components/Grid/Variations/GridExampleKeyboardNavigable.tsx
```diff
@@ -0,0 +1,107 @@
+import React from 'react'
+import { Grid, Image, Button, gridBehavior } from '@stardust-ui/react'
+import _ from 'lodash'
+
+const images = [
```
bmdalex: can you make this an array to save space?

bmdalex (approved):
looks good generally, please see comments

docs/src/examples/components/Grid/Variations/GridExampleKeyboardNavigable.shorthand.tsx
sophieH29: yep, sure, that does make sense. Thank you!

jurokapsiar (approved):

bmdalex:
I was trying this in our react conf prototype and it does not behave correctly when the grid items contain focusable items themselves:

screen shot 2018-10-26 at 13 58 16

As you can see in the screenshot above, the grid items contain a button ("FULL BIO") so navigation works incorrectly:

**actual**: arrow keys navigate to grid items then to buttons
**expected**: two level of keyboard navigation:
- arrow keys navigate between grid items;
- enter key navigates inside the grid item (level 2 key nav) that will trap focus;
- esc key navigates back to the grid item (level 1 key nav)

fyi @sophieH29 @jurokapsiar

sophieH29:
@Bugaa92  You are totally right about having nested navigation. But it is not what expected in current PR. Current PR provides navigation with arrow keys on grid items itself, whether they are focusable elements or have set ```data-is-focusable=true```

hughreeling (approved):

alinais (approved):

bmdalex:
in the example I provided above, each item contains a button but also `data-is-focusable=true` at the top item element which causes weird navigation:

```jsx
<Grid
  <div className="grid item" data-is-focusable="true" >
    <button>FULL BIO</button>
  </div>
  <div className="grid item" data-is-focusable="true" >
    <button>FULL BIO</button>
  </div>
  ...
  <div className="grid item" data-is-focusable="true" >
    <button>FULL BIO</button>
  </div>
/>
```

let's say the focus is in the `Grid`; user scenario:
- focus is on first grid item;
- press `right arrow key`; focus goes to button inside first grid item;
- press `right arrow key`; focus goes to second grid item;
- press `right arrow key`; focus goes to button inside second grid item;
- etc.

This looks wrong to me, do we want merge it like this?

**LATER EDIT:**

Discussed with @sophieH29 and all these complex cases will be handled in future work

miroslavstastny (approved):

PR #406, opened 2018-10-26
fix: rename "styles" prop to "css"

kohlikohl:
This renames the `styles` prop for passing in css styles to a component to `css`.

### Why was this change made?
`styles` is too closely related to `style` which would add the passed in styles as inline styles to an element, instead of funneling it through Stardust's css-in-js processor.

### Example
**Before:**
```tsx
 <Button content="Click here" styles={{backgroundColor: "red"}} />
```

**Now:**
```tsx
 <Button content="Click here" css={{backgroundColor: "red"}} />
```


docs/src/views/ThemingExamples.tsx
```diff
@@ -81,14 +81,14 @@ export default () => (
...
-    <Header as="h4" content="Changing component styles" />
+    <Header as="h4" content="Changing component css" />
 
     <ExampleSnippet
       value={[
         `import React from 'react'`,
         `import { Button } from '@stardust-ui/react'`,
         ``,
-        `const styles = {`,
+        `const css = {`,
```
kuzhelov, line 91: 👍

kuzhelov (approved):
agree with the changes proposed, as well as share the reasons for them - but lets wait for other thoughts before merging this. @levithomason, @miroslavstastny, @mnajdova, @Bugaa92, @alinais

miroslavstastny (commented):
What's done is technically correct.

However, I am not sure that this is enough. Do we want to just rename the prop to `css` and still talk about it as *styles* throughout the code? 
Random places which were not renamed - `mergeComponentStyles` function, all `themes/**/*Styles.ts` files and variables in them.

`css` diverges from Fabric which uses `styles`.

I would like to hear @levithomason opinions on these.

kuzhelov:
agree, lets wait for other folks expressing their ideas on that. Speaking of the points expressed
> However, I am not sure that this is enough. Do we want to just rename the prop to css and still talk about it as styles throughout the code?
  - we might come to some other naming convention, but, frankly, doesn't seem to be that critical - due to this 'styles' term won't be ambiguous within the Stardust context. Don't get me wrong: if there are reasons to rename these files, lets rename them - just saying that it shouldn't be the factor that will stop us from making this move, as it seems to be not that critical.
> Random places which were not renamed - mergeComponentStyles function, all themes/**/*Styles.ts files and variables in them
  - quite similar thoughts on this. To me it seems that if we would pedantically follow naming convention that would resemble prop name _just for the sake of following it_ (here I might be mistaken, but currently it won't be ambiguous to me if we'll talk about _styles applied via `css` prop_).
> css diverges from Fabric which uses styles.
  - here the question we might ask ourselves is 'whether we are fine to leave `styles` prop name as it is' - here I'd definitely vote for changing it, even given that we will diverge from Fabric's names, as it introduces quite serious chance to confuse client.

So, to conclude - absolutely support the move introduced in the PR, and see it as a very necessary thing to do. See points raised by @miroslavstastny as important ones to agree upon, and make all the necessary steps to address and overcome them to make this `styles` rename change happen :)

levithomason:
A few thoughts.  If we do this, some more renaming is likely in order.  I also want to give current thoughts on why we may not want to rename but remove the `styles` prop.

### Style file renaming

If we rename the prop, I don't think filenames are an issue at present.  We can / should rename the files in a separate PR.

Non-public filename contracts can iterate in another PR if necessary.  They don't affect users and we can safely rename them with `git mv ...`.  That command works best if there are no changes to the files when moving them, otherwise, git sometimes gets confused and thinks you deleted the old file and made a new one opposed to simply moving it with changes.

### Context / Theme interface

This is more of a concern to me.  The user will still do:

```jsx
<Provider theme={{ componentStyles: { Button: { ... } } }} />

<Button css={{ ... }} />
```

This seems like it should be called `theme.componentCSS` to align with the `css` prop.  However, see next point.

### Public API Consistency

As we expose any public utilities or typings, they should be consistent in naming.  If we go with `css`, then I think the public API also needs to follow suit.  I'd rename exported typings from `*Styles` to `*CSS` and utilities from `*Styles()` to `*CSS()`.

### Discouraging style overrides

It is currently too easy to override styles, just pass a `styles` prop.  If we make this prop even shorter, it will only increase its appeal (however minimal the increase might be).

Overriding styles should not be part of the normal usage of Stardust, since the user should use the APIs and allow the component to correctly handle styles.  Evidence from early first usages is that the first thing the user tends to do is reach for `styles` to make things look right.  This destroys the hope of allow the theme to control the UI.  Any theme updates are going to conflict with whatever the user has overridden with an inline style.  Users should be pushed hard toward using the correct component with correct APIs instead of hacking styles all over the place.

Stardust style overrides might be compared to React's `dangerouslySetInnerHTML` prop.  Yes, in React you can pass an HTML string, however, it is extremely dangerous and you are skipping the whole point of React by doing so.  Yes, in Stardust you can pass your own CSS overrides, but you are skipping one of the main points of using Stardust by doing so.   Similar to React making it obtuse and ugly to set HTML directly, **we are considering ways to make it more obtuse and ugly to override styles**.  This way the user feels deeply that they are doing something bad, because they are.

One idea was to remove the `styles` prop entirely.  The user could still override styles, but they need to do so by importing the Provider and passing a new theme on context.  Reasons to consider this:
1. This would certainly discourage the user from overriding several styles on several components in a view.
1. The user would rather try to use the component APIs to achieve the goal (which is what we want).
1. The user would push styles overrides higher up in the tree since they need to be in a theme.  Putting them in a theme also forces you to organize them better and use theme concepts, like variables.

***

Feedback?

kuzhelov:
## TL;DR
Would rather suggest to stay with renaming, note removing - at least as a first step (with all the necessary rename refactoring being done for that). Simple explanation is that removal is much more serious change, and while `styles` -> `css` has clear evolutionary benefit of avoiding client's confusion, removal of this prop, in contrast, is rather a contradictory decision. Personally feel strict tension against removing this prop altogether, because _this move won't solve the goal it is intended for_.

-------------

> Stardust style overrides might be compared to React's `dangerouslySetInnerHTML` prop.  Yes, in React you can pass an HTML string, however, it is extremely dangerous and you are skipping the whole point of React by doing so.  Yes, in Stardust you can pass your own CSS overrides, but you are skipping one of the main points of using Stardust by doing so.   Similar to React making it obtuse and ugly to set HTML directly, **we are considering ways to make it more obtuse and ugly to override styles**.  This way the user feels deeply that they are doing something bad, because they are.

Cannot completely agree with this, and here are the reasons for that. First thing to note is that **_any_ major component library, and even Web itself (via `style` attribute) allows 'inline' style customizations.** Stardust isn't aimed to introduce anything new, but just resembles already emerged patterns. So, here we will break this promise, and I might say for certain that this move will be generally very discouraging to the client.

#### was
```jsx
<Button styles={...} .. />
```

#### now
```jsx
// this provider is introduced just to wrap a Button 
// note how many things client should know about Stardust just to simply style the Button:
// - Provider concept, 
// - 'componentStyles', 
// - the fact that selectors work by component display name
// - 'root'
<Provider theme={ componentStyles: { Button: { root: ... }  }} >
    <Button ... />
</Provider>
```

Here are few things to note
- what we'll end up with is that **`Provider` will be used for specific component only** (the first thought client would have in terms of how situation with styling should be addressed), and it won't fix the problem - essentially, we will left at the same point as it was with `styles` props, but _with much more efforts required from client to style component_
- the second point aims to address potential objection like 'ok, that seems to be exactly the thing we were striving for - to make it harder to the client to not use Provider'. Yes - but we are not solving the problem we would really like to solve - to direct client towards organising styles into themes with that. **_If there is no mental organisation of styles in client's head yet, irregardless of which tools we would suggest to style, there won't be any structure and consistency in client styles at the end_**. `Provider` will be used just as an annoying hack, and after couple of uses the very next thought of the client would be to wrap it like:
```jsx
const withStyles = (Element, styles) => <Provider theme={ ... }>{Element}</Provider>

// and latter
render() {
  return withStyles(<Button ... />,  { border: ... })
```
After that client would left being mad on us that we haven't introduced this way for customising component styles (and had wasted her time on learning all these concepts), **and will just use this shortcut everywhere in the code.**
- once again to reiterate previous point - **theming need is something that naturally emerges in client head, and it emerges at the moment when there is _clear vision on how styles should be organised_**. Note that **this is a natural flow that is introduced by tools that are provided by Web (`style` attribute vs CSS styles) and _any_ component library**. If we remove this capability, I am absolutely sure that we would rather raise client's confusion and even desire to find a better component library that is not that restrictive and revolutionary in this _commonly adopted aspect_.

-----------------

So, to conclude - I would suggest to stay with renaming, at least as a first step (with all the necessary rename refactoring being done for that). Feel strict tension against removing this prop altogether, because _this move wouldn't solve the goal it is intended for_.

kohlikohl:
Agree with @kuzhelov's evaluation and don't believe making it hard for developers to override styles is the right approach here. Why?
* variables are great, but they by far don't cover all the permutations on how a component can be styled. If we want maximum adoption, should we not also opt for flexibility?
* Moving every style override into the theme is not practical as that assumes that the change applies to all components of a type. How is a local permutation achieved that is not supported by stardust yet?

I recognize the need to have a mechanism to prioritize `variables` over `css` overrides. Maybe this can be approach through education in form of `console.warn` when a developer makes a change through custom styles that could be made through variables?

layershifter:
As newcomer I may be will have a more clear eye there.

-----

For customising we have:
- themes as files
- themes as object, _I know that is the same thing, but from my side it looks differently_
- `styles`

Too many options. May be I wrong, but neither `emotion` nor `styles-components` doesn't allow you to override styles dynamically with a `css` prop, https://codesandbox.io/s/k2mv409ox3

-----

image

```jsx
<Button content="Click here" styles={{ color: 'red' }} />
<Button content="Click here" style={{ color: 'red' }} />
```

Can someone explain me what is going in the code above? The renaming for `styles` to `css` doesn't solve this problem at all.

So, I have an unpopular opinion and will agree with Levi, let's drop `styles` at all 😈

hughreeling:
I think we should ANNOUNCE that we're going to drop styles, add comments to the docs to mark as depreciated, and update docs.  Then we can actually drop it after we've had feedback on alternatives.

alinais:
Given that the general suggestions are to drop entirely the ```styles``` prop, let's close this PR as it will not bring much value long term.

PR #408, opened 2018-10-29
fix(Segment): add background color

levithomason:
See #407, the Segment does not have a background color currently.

This PR adds a background color that matches the body background.  A few other style values were also hoisted to the variables position.

### DIFF
--- a/src/themes/teams/components/Segment/segmentVariables.ts
+++ b/src/themes/teams/components/Segment/segmentVariables.ts
@@ -1,0 +1,18 @@
+import { pxToRem } from '../../../../lib'
+import { ComponentVariablesPrepared } from '@stardust-ui/react'
+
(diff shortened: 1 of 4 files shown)


src/themes/teams/components/Segment/segmentVariables.ts
```diff
@@ -0,0 +1,18 @@
+import { pxToRem } from '../../../../lib'
+import { ComponentVariablesPrepared } from '@stardust-ui/react'
+
+export interface SegmentVariables {
+  padding: string
+  color: string
```
miroslavstastny: `color` and `fontSize` are not used anywhere in segmentStyles

levithomason: thanks, fixed

PR #410, opened 2018-10-31
feat(Popup): close popup on click

kuzhelov:
## TODO
- [x] update CHANGELOG

## Close Popup on mouse click

This set of changes introduces functionality that will close `Popup` on click. Currently `Portal` is not reused, as there is some refactoring planned for this component before reusing it in `Popup`. At the same time, provided changes introduce full set of functionality client would expect, so this remains to be just a question of further refactoring (`Portal`) and generalization (reuse `Portal`).

## EventStack subscription API adjustments

Event stack subscriptions were reworked, to make the API for subscribing and unsubscribing to be safer. Please, consider to take a look on the following comment for more details on that: https://github.com/stardust-ui/react/pull/391#issuecomment-432422207


src/components/Popup/Popup.tsx
```diff
@@ -126,16 +127,45 @@ export default class Popup extends AutoControlledComponent<Extendable<PopupProps
...
+  private closeAndFocusTrigger = e => {
+    if (this.state.open) {
+      this.trySetOpen(false, e, true)
+      _.invoke(this.triggerDomElement, 'focus')
+    }
+  }
+
+  private closeAndFocusTriggerOnClickIfOpen() {
+    if (this.state.open) {
+      setTimeout(() => this.clickSubscription.replaceWith('click', this.closeAndFocusTrigger))
```
joheredi: With the current implementation, a click on the Popup content will close it. Is this the intended behavior? Portal component checks if the click was outside of the content before closing. Do we want to do the same thing here? Or maybe allow a prop to choose if the popup should be closed onClick or onClickOutside?

src/components/Popup/Popup.tsx
```diff
@@ -126,16 +127,45 @@ export default class Popup extends AutoControlledComponent<Extendable<PopupProps
 
   private static isBrowserContext = isBrowser()
 
+  private clickSubscription = EventStackSubscription.empty()
```
joheredi: It may be a good idea to add a prop to choose if the popup should be closed onClick. We may want to have it true by default

src/components/Popup/Popup.tsx
```diff
@@ -126,16 +127,45 @@ export default class Popup extends AutoControlledComponent<Extendable<PopupProps
...
   }
 
   public state = { target: undefined, open: false }
 
+  public componentDidMount() {
+    this.closeAndFocusTriggerOnClickIfOpen()
+  }
+
+  public componentDidUpdate() {
+    this.closeAndFocusTriggerOnClickIfOpen()
```
sophieH29: what was a reason to have subscription logic on```componentDidUpdate```?  Isn't it enough to have it on the edge lifecycles callbacks, like ```componentDidMount``` and ```componentWillUnmount```?

src/lib/eventStack/eventStack.tsx
```diff
@@ -3,13 +3,13 @@ import EventTarget from './EventTarget'
...
 class EventStack {
-  _targets = new Map()
+  private readonly _targets = new Map()
 
   // ------------------------------------
   // Target utils
   // ------------------------------------
 
-  _find = (target, autoCreate = true) => {
+  private _find = (target, autoCreate = true) => {
```
sophieH29, line 12: can we add also typings for methods' parameters, in regards to refactoring of event stack?

sophieH29: btw, you might consider it as separate PR

src/lib/eventStack/eventStack.tsx
```diff
@@ -55,6 +54,32 @@ class EventStack {
...
+    this._unsubscribe = temp
+  }
+
+  private constructor(private _unsubscribe: Function, public readonly isEmpty: boolean = false) {}
+
+  public stop(): void {
+    this._unsubscribe()
+  }
+
+  public replaceWith(name, handlers, options: any = {}): void {
```
sophieH29: I understand why you did use namings like ```replaceWith``` and ```stop``` here in Subscription logic.
But if look from a usage point of view, me as a user of Event Stack, was really confusing what does it mean and how to use it when I saw it in the components above.
I would really like if stick with namings like ```subscribe``` and ```unsubscribe```, this is what expected from Event Stack API.
So I suggest exposing these 2 methods which will wrap your ```stop``` and ```replaceWith``` logic.

Let me know what you think

miroslavstastny: I had exactly the same problem - `EventStackSubscription` method names clearly describe what the methods do.
However, from consumers' POV, I have no idea why/when I should call them.

src/components/Popup/Popup.tsx
kuzhelov: there are two subtle problems that might occur with the alternative approach (if we will subscribe only once).
- first is that we will have subscription for all the popups on the page, even if those are not opened. In other words, say, we will have 100 components on the page that are 'poppable', and this will result in 100 subscriptions _just because these components exist_. This introduces some perf problems, but, what is more important, changes to global app state that are hard to debug - when you are looking on the event listeners, you would expect at each moment of time that there are only those that necessary, and now, in contrast, you will have full fledge of subscriptions that exist for no good reason (if popup is closed - due to those are responsible for Popup behavior in open state)
- another one is maybe even more serious. If you'll try to implement these changes on your local machine, you will immediately see that popup is never opened on click. The problem is that _if client code forgets to call `e.preventDefault()`, event will 
  - set Popup state to `open`
  - then propagate to `document` element and execute the logic of handler, which says that `for open state close the Popup`

Thus, there won't be any changes visible to the client. Is it a problem of the client? I would say that certainly not, as _this event subscription should only occur when popup is opened_ - and this is what code provided is doing

src/lib/eventStack/eventStack.tsx
kuzhelov: would rather focus on functionality in this PR - those haven't existed before, would like to devote PR to introduce them

src/lib/eventStack/eventStack.tsx
kuzhelov: ok, lets agree on what is the proposal here. Should we use the following names:

```js
const subscription = EventStackSubscription.subscribe('click', ...)  // ? rather confusing
// or 
const subscription = EventStackSubscription.new('click', ...)
// or
const subscription = EventStack.subscribe('click', ...)

// and then
subscription.unsubscribe()
// or 
subscription.clear() // note, this resembles 'clearTimeout()' naming
// or something else

// for 'replaceWith'
// - eliminate it for now
// - introduce better alternative, as the following usage pattern is quite common, 
//    consider Stardust code as a proof:
subscription.stop()
subscription = EventStackSuscription.create('click', ...)
```

Please, let me know about your thoughts on this. Thanks! @sophieH29, @miroslavstastny

hughreeling: For me : const subscription = EventStack.subscribe('click', ...)
subscription.unsubscribe()

miroslavstastny: Ignoring implementation details, this works best for me:

```jsx
const subscription = EventStack.subscribe('click', ...)
...
subscription.unsubscribe()
```

And get rid of `replaceWith`, let the developer unsubscribe before reusing the variable.

sophieH29: I like the above one, mentioned by Hugh, let's be more straightforward

src/components/Popup/Popup.tsx
sophieH29: got your point, thank you very much for a detailed explanation

src/components/Popup/Popup.tsx
sophieH29: in your previous refactoring ```setTimeout``` was part of ```EventStack``` and, as I understood, a motivation of it was performance reason (please, correct me, if I am wrong). So the question is, should we leave it as part of the library so everyone who uses it can benefit from it? It can be as a separate method ```subscribeAsync```.

kuzhelov: lets start with the simplest thing - specifically, the one you are suggesting. Will implement this check, so that Popup won't be closed on click inside the popup area

src/components/Popup/Popup.tsx
kuzhelov: just as a counterargument, we might argue that as long as Popup supports controlled mode, there are all the necessary tools to customize Popup's behavior.

Please, don't get me wrong - personally, see this suggestion as being absolutely reasonable. At the same time, would like to keep this PR simple and targeted to specific thing that it is aimed to address, so that we would be able to decide on this proposed prop later and, if necessary, introduce separate PR for that.

src/lib/eventStack/eventStack.tsx
kuzhelov: ok, let me do this change then, thanks!

src/components/Popup/Popup.tsx
kuzhelov: this is a very good point. Actually, for now have decided to move this aspect to the client's code, as have discovered few additional subtle problems that may arise with eventStack's logic (and the fact that subscription handler could be called at the same JS frame). 

The simplest example - suppose that there is a following client code:

```js
foo() {
  // note, with 'setTimeout' this subscription will be made async
  const subscription = EventStack.subscribe('click', ...)
   
  // so there is, in fact, no active subscription at this point
  subscription.unsubscribe()
}
```
Note that if subscription logic is async, this will result in subscription still being active at the end of the method - as 'unsubscribe' will be invoked at the moment when there is no subscription made yet.

One simplest solution would be to make all the necessary methods `async`, but this will complicate client's code. What I am thinking about, however, is a bit more elaborative approach where all these bits will be handled properly by eventStack, at the same time **without switch to async model** requested from the client (as async methods are quite error and race condition prone to be used in React event handlers). Now am working on the PR where would like to address this problem.

src/components/Popup/Popup.tsx
jmanuelh: Makes sense!

jmanuelh (approved):

src/components/Popup/Popup.tsx
joheredi: Makes sense!

joheredi (approved):

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -269,14 +272,16 @@ export class FocusTrapZone extends React.Component<FocusTrapZoneProps, {}> {
   private _subscribeToEvents = () => {
     const { forceFocusInsideTrap, isClickableOutsideFocusTrap } = this.props
     if (forceFocusInsideTrap) {
-      eventStack.sub('focus', this._handleOutsideFocus, {
+      this._focusSubscription.unsubscribe()
```
sophieH29, line 275: What is the main point to call ```unsubscribe``` before actual subscription? Can't we assure that we are unsubscribed from any events at the initialization ``` EventStack.noSubscription```?

bmdalex: +100 it doesn't look like a good pattern, we don't want to attempt teaching developers to do this, they will forget; I find the eventStack util difficult to use now

src/lib/eventStack/eventStack.tsx
bmdalex: @kuzhelov - I think whenever we claim we'll add something in a follow-up PR we should create tickets or issues to actually track the work so that it won't slip through the cracks and other devs have access to it and can pick it up

src/lib/eventStack/eventStack.tsx
```diff
@@ -31,18 +31,17 @@ class EventStack {
...
+    const eventTarget = this._find(target)
 
-      eventTarget.sub(name, handlers, pool, useCapture)
-    })
+    eventTarget.sub(name, handlers, pool, useCapture)
+    return () => this.unsub(name, handlers, options)
   }
 
-  unsub = (name, handlers, options: any = {}) => {
+  public unsub = (name, handlers, options: any = {}) => {
```
bmdalex, line 44: why is this still public with the current approach?

src/lib/eventStack/eventStack.tsx
```diff
@@ -55,6 +54,25 @@ class EventStack {
...
+
+  public constructor(private _unsubscribe: Function, public readonly isEmpty: boolean = false) {}
 
-export default instance
+  public unsubscribe(): void {
+    this._unsubscribe()
+  }
+}
+
+export class EventStack {
```
bmdalex, line 71: this approach feels complicated... why don't we just do something like:

```typescript
public sub = (name, handlers, options: any = {}): Function => {
  if (!isBrowser()) return () => {}

  const { target = document, pool = 'default', useCapture = false } = options
  const eventTarget = this._find(target)
  const safeHandlers = (...args) => handlers(...args)

  eventTarget.sub(name, safeHandlers, pool, useCapture)
  return () => this.unsub(name, safeHandlers, options)
}

// make `unsub` private

const eventStack = new EventStackInner()
export default eventStack.sub
```

usage will become:
```typescript

import { eventSub } from 'src/lib'

class MyClassUsingEventStack {
  private unsub: Function // handler for cleaning eventStack subscription

  init() {
    this.unsub = eventSub('click', this.handleDocumentClick)
  }

  cleanup() {
    this.unsub && this.unsub()
  }
```

what do you think, @kuzhelov ? am I missing something?

src/components/Popup/Popup.tsx
```diff
@@ -126,16 +127,53 @@ export default class Popup extends AutoControlledComponent<Extendable<PopupProps
...
+
+  private closeAndFocusTrigger = e => {
+    if (this.state.open) {
+      this.trySetOpen(false, e, true)
+      _.invoke(this.triggerDomElement, 'focus')
+    }
+  }
+
+  private closeAndFocusTriggerOnClickIfOpen() {
+    this.outsideClickSubscription.unsubscribe()
```
bmdalex, line 150: shouldn't we check if we check if we're unsubscribed already? something like:
```typescript
if (this.outsideClickSubscription) {
  this.outsideClickSubscription.unsubscribe()
  this.outsideClickSubscription = null
}
```

bmdalex (changes requested):
pls take a look at comments

src/components/Popup/Popup.tsx
kuzhelov: it is safe to call unsubscribe multiple times,  this call acts as a declarative instruction - thus, no checks are neceessary

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
kuzhelov: please, be specific about 'do what', thanks! Because see the following as a counterargument to the statement 'hard to use now':

#### now
```js
// safe subscription
const subscription = EventStack.subscribe('onclick', () => { .. /* processing logic here */ })

// - this object has IntelliSense support
// - safe unsubscribe instruction
subscription.unsubscribe() 
```

#### before
```js
// ooops
eventStack.sub('onclick', () => { ... })

// ooops
eventStack.sub('onclick', this.handleClick.bind(this))

eventStack.sub('onclick', this.handleClickWithBoundThis, { useCapture: true })
// oooops 
// - what were the arguments I've subscribed with? 
// - some of them are critical for proper unsubscription
eventStack.unsub('onclick', this.handleClickWithBoundThis)
```

Sorry, but now cannot agree with the point mentioned

src/lib/eventStack/eventStack.tsx
kuzhelov: I am strongly against introducing any code that will lead to null checks on the client side - it should be safe to call `unsub` in the following way:

```js
class MyClassUsingEventStack {
  ...
  cleanup() {
    this.unsub()
  }
```
--------------
**Second thing** is that with just function being returned from subscription method (stored in `unsub` variable), it would be much harder to introduce changes to unsubscription logic - subscription's API won't be extensible, and each potential change to it will result in breaking API changes.

--------------

>Expose subscription function only from `eventStack` module

Agree, this is a good move, thanks!

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: I agree that it looks much better now, that it was. The only my concern was about calling ```unsubscribe``` before actual ```subscribe```, as it seems to me not very intuitive.

kuzhelov: > The only my concern was about calling unsubscribe before actual subscribe, as it seems to me not very intuitive.

Here I would note that this is not something that is directly `eventStack`-related, as what we are doing is the following:
- we would like to subscribe and store subscription in the object's field
- there is (maybe) a pending subscription stored in this field - thus, we need to unsubscribe for it
  - as it is safe to call 'unsubscribe' on 'noSubscription', we are not making any null checks for it. If `null` value will be used as initialiser, we would write something like `subscription && subscription.unsubscribe()`
- then we are making new subscription, storing reference on it

---------------
*Note that initially the following API was suggested for it, but we've agreed to cancel/postpone it:
```js
// terminates previous subscription and replaces with new one
this.clickSubscription.replaceWith('click', ...)
```

src/lib/eventStack/eventStack.tsx
bmdalex: discussed with Roman and decided to stick with the approach he's introducing and revisit the EventStack API at a later time if needed

bmdalex (approved):

mnajdova (approved):
I briefly took a loot into this, but seems that other developers already review it. Approving as all comments are resolved.

PR #411, opened 2018-10-31
feat(proto): Form validation prototype page

mnajdova:
# Form validation prototype

This PR aims to add form validation prototype for showing how the Form component can be used together with formsy-react for validating the fields on change, or on submit.

Here is how the both form look:
image

Inside the prototype I created some wrapper components for the Input and RadioGroup components, that are using the Formsy* components


docs/src/prototypes/formValidation/FormValidationOnChange.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const { buttonDisabled } = this.state
+    const fields = [
+      {
+        control: { as: Input },
+        name: 'full-name-on-change',
+        id: 'full-name-on-change',
+        key: 'full-name',
+        required: true,
+        label: 'Full Name',
+        validationErrors: {
```
hughreeling, line 37: Is there duplication between prop required and validation rule IsDefaultRequired?  Or should there be warning if required is true and isDefaultRequired not set?  - Not sure these are good ideas - but wanted to think about it.

package.json
```diff
@@ -87,6 +87,7 @@
     "fela-plugin-placeholder-prefixer": "^5.0.18",
     "fela-plugin-prefixer": "^5.0.18",
     "fela-plugin-rtl": "^1.0.6",
+    "formsy-react": "^1.1.5",
```
hughreeling, line 73: Do we need to start TPS process for this?

hughreeling (approved):

docs/src/prototypes/formValidation/FormValidationOnChange.tsx
mnajdova: isDefaultRequired inside the validationErrors props is just the error message that should be shown, and the required is the actual rule.

package.json
mnajdova: If we decide to go with this library, then yes.

yarn.lock
```diff
@@ -5,14 +5,12 @@
 "@babel/code-frame@^7.0.0-beta.35":
   version "7.0.0-beta.52"
   resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.0.0-beta.52.tgz#192483bfa0d1e467c101571c21029ccb74af2801"
-  integrity sha1-GSSDv6DR5GfBAVccIQKcy3SvKAE=
```
bmdalex: @mnajdova can you please update your `yarn` to a version higher than `1.10.0`; we don't want to remove these lines; see #437 for details

bmdalex (dismissed):
pls update `yarn` then update `yarn.lock`

yarn.lock
mnajdova: Done, thanks!

docs/src/prototypes/formValidation/FormValidationOnChange.tsx
```diff
@@ -0,0 +1,98 @@
+import * as React from 'react'
+import { Form, Button, Input, RadioGroup } from '@stardust-ui/react'
+import Formsy from 'formsy-react'
+import FormsyFormField from './FormsyFormField'
+
+class FormValidationOnChange extends React.Component<any, any> {
```
layershifter: Let's avoid any as much as possible.

docs/src/prototypes/formValidation/index.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react'
+import { Header } from '@stardust-ui/react'
+import FormValidationOnChange from './FormValidationOnChange'
+import FormValidationOnSubmit from './FormValidationOnSubmit'
+
+class FormValidationPrototype extends React.Component<any, any> {
```
layershifter: This component can be an SFC I think.

docs/src/prototypes/formValidation/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+  render() {
+    return (
+      <div style={{ margin: '10px' }}>
+        <Header content={'Form validation on submit'} />
+        <FormValidationOnSubmit />
+        <Header content={'Form validation on change'} />
```
layershifter:
```suggestion
        <Header content='Form validation on change' />
```

docs/src/prototypes/formValidation/index.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react'
+import { Header } from '@stardust-ui/react'
+import FormValidationOnChange from './FormValidationOnChange'
+import FormValidationOnSubmit from './FormValidationOnSubmit'
+
+class FormValidationPrototype extends React.Component<any, any> {
+  render() {
+    return (
+      <div style={{ margin: '10px' }}>
+        <Header content={'Form validation on submit'} />
```
layershifter:
```suggestion
        <Header content='Form validation on submit' />
```

docs/src/prototypes/formValidation/FormValidationOnChange.tsx
```diff
@@ -0,0 +1,98 @@
...
+              valueProp="checkedValue"
+              onChangeProp="checkedValueChanged"
+              eventTargetAsValue={false}
+            />
+          )
+        }
+      }
+      return field
+    })
+  }
```
layershifter, line 99: We should definitely refactor this somehow, however I don't there a working idea there.

A plain JSX is more readable in this case:

```jsx
<FormsyInput name='email-on-change' label='Email' />
<FormsyRadioGroup name='email-on-change' options={[
  { label: 'Male', value: '1' },
  { label: 'Female', value: '2' },
]} />
```

The other question how to deal with validation rules there.

@mnajdova why did you choose objects? As I see official docs use JSX, too, https://github.com/formsy/formsy-react/#2-use-your-formsy-element

BTW, there are wrappers for SUIR and I really like their API: https://github.com/zabute/formsy-semantic-ui-react#usage

layershifter (changes requested):
I left a proposal how we can try to make forms more cleaner, @mnajdova feel free to discuss it 👍

docs/src/prototypes/formValidation/FormValidationOnChange.tsx
mnajdova: I had two different wrapperts for the Input and RadioGroup in the beginning, but I was thinking that i am over-complicating things... Seeing your point, I think yeah, if we will create wrappers for all stardust component (similar to this: https://github.com/zabute/formsy-semantic-ui-react#usage) the API would be much much better. I chose objects, just so that I can use the shorthand API, no other reason. Do you propose that we create several wrappers for our components (like for the Form, FormField, Input, RadioGroup using formsy-react) and them show these in the prototype?

layershifter: Yeah, I think that it will the best option.

When we will create a monorepo we can just move them to a separate package and instead of copying component's files users will be able to use a package with working bindings 👍

layershifter (approved):
I still have an opinion that the proposed API looks much more cleaner.

However, let's go with current implementation and then iterate on it.

mnajdova:
Created issue https://github.com/stardust-ui/react/issues/533 as a follow up to this PR.

layershifter:
This thing is obviously dead, closing for housekeeping.

Issue #370, opened 2018-10-17
Multiple Selectable lists cause scrolling issues due to auto-focus


musingh1:
### Steps

- Navigate to https://stardust-ui.github.io/react/components/list
- Try adding selection to the last list example
![image](https://user-images.githubusercontent.com/22622723/47118217-80b38a00-d21b-11e8-8847-49ee34b2c49a.png)
- Try scrolling the page

### Expected Result
The page should work fine if there are multiple selectable lists on it.

### Actual Result
The scrolling does not work. It seems the list is trying to auto set the focus on the first item when scrolling is happening.

### Version
v0.10

levithomason:
Confirmed, this is an odd bug.

The core issue seems to be that the first item in a `selection` list is auto focused on scroll.  Having two on the page causes both to be focused and the page tries to scroll them into view... while you're already scrolling.

### Smaller Reproduction
Note the behavior of the List's [Selection Example](https://stardust-ui.github.io/react/components/list#types-selection) when scrolling.

1. Open the [Selection Example](https://stardust-ui.github.io/react/components/list#types-selection)
1. Click on the page background to unfocus all list items
1. Scroll the page a tiny bit
1. 🐞The first item is now focused

![http://g.recordit.co/pJQUpdb2xV.gif](http://g.recordit.co/pJQUpdb2xV.gif)

levithomason:
OK, looks like an obvious List bug:

```jsx
  componentDidUpdate() {
    this.focusableItem.tryFocus(ReactDOM.findDOMNode(this.itemRef.current) as HTMLElement)
  }
```

Anytime the list updates it's going to try to focus an item.  There are a number of issues with this bit of code that we need to fix:

1. There are no tests for this List behavior.
1. `findDOMNode` should be avoided at all costs, see the [React Docs](https://reactjs.org/docs/react-dom.html#finddomnode).  It is also deprecated in `StrictComponent`.
1. The List Item has an instance member `this.focusableItem()` which is created by passing `this` to a constructor.  Passing instances around is code smell.
1. There is now a prop called `focusableItemProps` that doesn't make sense to consumers and should probably be removed.
1. The behavior of focusing an item is only required when the list is a `selection` list, but the `this.focusableItem()` is applied to every instance of the list.
1. There probably shouldn't be any `componentDidUpdate` code in the List, especially since it is stateless.

The source PR is #256.  It appears we need this for accessibility reasons, however, I propose we reconsider the implementation.

[...]

PR #413, opened 2018-11-01
fix(List): focus behavior

kuzhelov:
## TODO
- [x] update CHANGELOG

------------------

This PR aims to address List focus issues, as well as simplify generic logic used for collection's focus behavior. The problems that were mentioned in the following post (https://github.com/stardust-ui/react/issues/370#issuecomment-432874011) are addressed, specifically:

- **_Anytime the list updates it's going to try to focus an item._** - well, not anymore 👍 
- **_`findDOMNode` should be avoided at all costs._** - avoided 👍 
- **_The List Item has an instance member `this.focusableItem()` which is created by passing `this` to a constructor.  Passing instances around is code smell_** - this smell has been eliminated, only safe immutable data is passed to constructor now 👍 
- **_There is now a prop called `focusableItemProps` that doesn't make sense to consumers and should probably be removed._** - removed, and, generally speaking, provided focus behavior introduces **no** changes to public component's API now 👍 
- **_The behavior of focusing an item is only required when the list is a `selection` list, but the `this.focusableItem()` is applied to every instance of the list_** - this differentiation is made 👍 
- **_There probably shouldn't be any `componentDidUpdate` code in the List, especially since it is stateless._** - there is no any now 👍


src/components/List/List.tsx
```diff
@@ -120,12 +119,40 @@ class List extends UIComponent<Extendable<ListProps>, FocusContainerState> {
...
     const { items, renderItem } = this.props
-    const itemProps = _.pick(this.props, List.itemProps)
 
+    this.itemRefs = []
     return _.map(items, (item, idx) => {
-      itemProps.focusableItemProps = this.focusContainer.createItemProps(idx, items.length)
+      let maybeSelectableItemProps = {}
+
+      if (this.props.selection) {
+        const _ref = React.createRef()
```
sophieH29: some time ago we had a discussion in other PR when underscore was used for a local variable as here and we agreed that it should be removed :)

src/components/List/List.tsx
```diff
@@ -120,12 +119,40 @@ class List extends UIComponent<Extendable<ListProps>, FocusContainerState> {
...
+  }
+
   renderItems() {
     const { items, renderItem } = this.props
-    const itemProps = _.pick(this.props, List.itemProps)
 
+    this.itemRefs = []
     return _.map(items, (item, idx) => {
-      itemProps.focusableItemProps = this.focusContainer.createItemProps(idx, items.length)
+      let maybeSelectableItemProps = {}
```
sophieH29: nit, you may initialize it as

```js
const maybeSelectableItemProps = {} 
```

and then on line 145
```js
maybeSelectableItemProps.tabIndex = 0
maybeSelectableItemProps.ref = _ref
maybeSelectableItemProps.onFocus = () => this.focusHandler.syncFocusedItemIndex(idx)
```
So no rebinding will happen for the object

src/lib/getKeyDownHandlers.ts
```diff
@@ -38,7 +38,11 @@ const getKeyDownHandlers = (
             componentActionHandlers[actionName],
             componentPartKeyAction[actionName].keyCombinations,
           )
-          eventHandler && eventHandler(event)
+
+          if (eventHandler) {
+            event.preventDefault()
```
sophieH29: having concerns about this line. In FocusZone ```onKeyDown``` handler checks if the event is ```preventDefaulted```, and if yes, it will do nothing. Just to imagine, let's say the parent container of the component is wrapped with FocusZone, and the same key event is propagated to FocusZone, it won't work as expected.

Indeed, in some cases we need to preventDefault.  If you look at ```chatMessageBehavior```, we have done it using keyActions

```js
keyActions: {
    root: {
      // prevents default FocusZone behavior, in this case, prevents using arrow keys as navigation (we only want a Tab key to navigate)
      preventDefault: {
        keyCombinations: [{ keyCode: keyboardKey.ArrowUp }, { keyCode: keyboardKey.ArrowDown }],
      },
    },
  },
```

So to wrap up my thought, it's not really safe to ```preventDefault``` for all the cases.
We consider making it optional, by passing extra param or using key actions as described above.

Please, let me know what you think.

src/lib/accessibility/FocusHandling/FocusContainer.ts
```diff
@@ -1,92 +1,66 @@
...
-    if (this.getState().focusItemOnIdx <= 0) {
+    if (this.getItemsCount() === 0) {
       return
     }
 
-    this.setState(prev => {
-      return { focusItemOnIdx: prev.focusItemOnIdx - 1 }
-    })
+    this._focusedItemIndex -= 1
+    this.coerceFocusedItemIndex()
```
sophieH29: just a suggestion, maybe we can change ```coerce``` --> ```constrain```, as it seems more often used for describing such cases.

P.S. I learned one more English word, thank you :)

src/lib/accessibility/FocusHandling/FocusContainer.ts
```diff
@@ -1,92 +1,66 @@
...
+    this._focusedItemIndex = 0
+    this.setFocusAt(this._focusedItemIndex)
   }
 
   public moveLast(): void {
-    if (
-      !this.getProps().items ||
-      this.getState().focusItemOnIdx === this.getProps().items.length - 1
-    ) {
+    if (this.getItemsCount() === 0) {
```
sophieH29: I see 4 places using this check. What do you think if move this to one fucntion as
```js
private _isListEmpty() {
    return this.getItemsCount() === 0
}
```
So it could even bring more meaning in other functions

```js
public moveFirst(): void {
    if (this._isListEmpty()) return

   this._focusedItemIndex = 0
    this.setFocusAt(this._focusedItemIndex)
  }	 
```

src/components/List/ListItem.tsx
```diff
@@ -93,9 +91,12 @@ class ListItem extends UIComponent<Extendable<ListItemProps>, ListItemState> {
 
   private itemRef = React.createRef<HTMLElement>()
 
-  componentDidUpdate() {
-    // This needs to be as part of issue https://github.com/stardust-ui/react/issues/370
-    // this.focusableItem.tryFocus(ReactDOM.findDOMNode(this.itemRef.current) as HTMLElement)
+  handleMouseEnter = () => {
```
bmdalex: pls make private

src/components/List/ListItem.tsx
```diff
@@ -93,9 +91,12 @@ class ListItem extends UIComponent<Extendable<ListItemProps>, ListItemState> {
...
+    this.setState({ isHovering: true })
+  }
+
+  handleMouseLeave = () => {
```
bmdalex: pls make private

src/lib/accessibility/FocusHandling/FocusContainer.ts
```diff
@@ -1,92 +1,66 @@
...
-}
+export class ContainerFocusHandler {
+  private _focusedItemIndex = 0
 
-export interface FocusContainerState {
-  focusItemOnIdx: number
-}
+  constructor(private getItemsCount: () => number, private readonly setFocusAt: (number) => void) {}
+
+  private coerceFocusedItemIndex() {
```
bmdalex: nit: can we use a simpler vocabulary for this method name?

src/lib/accessibility/FocusHandling/FocusContainer.ts
bmdalex: +1

src/lib/accessibility/FocusHandling/FocusContainer.ts
```diff
@@ -1,92 +1,66 @@
 import * as _ from 'lodash'
-import { FocusableItemProps, SetStateDelegate } from './FocusableItem'
 
-export interface FocusContainerProps<T> {
-  items?: T[]
-}
+export class ContainerFocusHandler {
+  private _focusedItemIndex = 0
```
bmdalex: do we have a new convention with `_` before private members?

bmdalex (changes requested):
pls address comments

src/lib/accessibility/FocusHandling/FocusContainer.ts
kuzhelov: please, introduce your suggestions for that

src/components/List/List.tsx
kuzhelov: ok, will remove

src/lib/getKeyDownHandlers.ts
kuzhelov: this change is introduced to prevent scrolling of the page when one would click up-down keys for selectable List (please, take a look on the corresponding example on the Stardust page we have now to see this problem) - each keypress results not just in switching list item, but also in scrolling the page. This is not a desired effect for sure.

Solutions? While we might consider to make `preventDefault()` calls for each action handler in the List component, it seems to be redundant. The logic where we have check that client has provided handler for the event seems to be a clear suggestion that this event will be handled by client. So, **by default I would definitely expect that event with handler defined will stop bubbling at the point where it is handled**.

> Just to imagine, let's say the parent container of the component is wrapped with FocusZone, and the same key event is propagated to FocusZone, it won't work as expected.

Sorry, could you, please, suggest what the case is - cannot understand what this is about. Currently have the following picture from these words, but not sure that have got an idea. Maybe you could suggest code example, or further elaborate on the case youare talking about?

```jsx
<FocusZone>
  <Container>
     <Element />  <--- here is where event raised
  </Container>
</FocusZone>
```

src/lib/accessibility/FocusHandling/FocusContainer.ts
sophieH29: ```constrainFocusedItemIndex``` ?

src/lib/getKeyDownHandlers.ts
sophieH29: Let me try to explain from a different side.
I've tried to make this change in the code and tested the chat prototype. Chat is no more navigable with arrow keys.

The reason for that:

Chat has defined ```keyActions``` in ```chatBehavior```
```js
 focus: {
        keyCombinations: [{ keyCode: keyboardKey.ArrowLeft }],
      },
```
When we press Arrow up or down on a chat, next happens:
- ```onKeyDown``` handler is triggered in ```getKeyDownHandler.ts```
- after mapping action names to key actions, we've got our handler, which contains filtering logic with keys inside it
- ```event.preventDefault``` is called for any keyboard event and any key
- the event goes to FocusZone where it checks next
```js
if (ev.isDefaultPrevented()) {
      return undefined
    }
``` 
- so navigation won't work in FocusZone

If we want to stick with ```event.preventDefault```, we have to move it to ```keyboardHandlerFilter.ts```, line 19, so it will look like
```js
if (isHandled) {
    event.preventDefault()
    handler(event)
  }
```

So then  ```event.preventDefault``` will be called only for specific keys, when the keyboard handler is triggered.

But still, I am afraid if we miss something and it can break some other use case. I might vote for having it as an optional param.

 @kuzhelov  Let me know what you think and if you get the idea or we can talk about it

jurokapsiar: preventDefault should not be there by default. It should be part of each event handler that requires it. I understand that it requires users to write a bit more code and have this understanding, but it also gives them flexibility in how they want to write their handlers and where do they want to catch events.

src/lib/accessibility/FocusHandling/FocusContainer.ts
kuzhelov: yes, this is exact the rename made 👍

src/lib/getKeyDownHandlers.ts
kuzhelov: agree, let me address that

bmdalex (approved):

sophieH29 (approved):

mnajdova (approved):
I reviewed this locally and make sense, although I was expecting to be able to circularly navigate throughout the list,  but not sure if we want to support this (not saying that we should).

kuzhelov:
@mnajdova 
> I reviewed this locally and make sense, although I was expecting to be able to circularly navigate throughout the list, but not sure if we want to support this (not saying that we should).

Yes, we should, but this wasn't supported yet. The main goal of this PR was to address the problem of the **existing functionality** (https://github.com/stardust-ui/react/issues/370#issuecomment-432874011), and this circular traversing behavior wasn't supported there as well. See this functionality as something that will be introduced by means of a dedicated PR.

jurokapsiar:
@kuzhelov I did not investigate further, but with this PR you can use TAB to go throgh the selectable list item. That should not be the case, you should only be able to use the arrow keys.

jurokapsiar (changes requested):
as mentioned in the comment, with this PR user can go through the list items using TAB.

kuzhelov:
@jurokapsiar, going to merge these changes, as original ones haven't introduced _any_ keyboard handling capabilities (have checked for master branch). As we are not making things any worse in this aspect, as well as fixing the problems outlined here (https://github.com/stardust-ui/react/issues/370#issuecomment-432874011), it seems to be reasonable to introduce fix as a first step. For TAB support - will file an issue, open to have a discussion in regards to approach that is better to be taken for that.

jurokapsiar (approved):

PR #414, opened 2018-11-01
feat(theme): support keyframes through animation property

mnajdova:
# Keyframes

This PR aims to add possibility for defining and using keyframes in the theme. 

# API Proposal

We should add keyframes prop in the theme, that will define different keyframes in the following manner:

```
<Provider theme={{
    // ..some theme props
    keyframes: {
      spinner: () => ({
        from: {
          transform: 'rotate(0deg)',
        },
        to: {
          transform: 'rotate(360deg)',
        },
      }),
    }
  }}
> <Component /> </Provider>
```
Then, inside the components the users can access this keyframes inside the styles, using the theme property. Something like:

```
<Icon name="umbrella" size="big"
      styles={({theme}) => (theme.keyframes.spinner ? {animation: 
        `${theme.renderer.renderKeyframe(theme.keyframes.spinner, {})} 1s infinite`} : {})}
    />
```

Another approach that I have tried is to evaluate the keyframes inside the Provider component once and then use them like this: 
```
<Icon 
  name="umbrella" 
  size="big" 
  styles={({theme, props}) => ({
    animation: `${theme.keyframes.spinner} 1s infinite`})
  } 
/>
```

Altough this API is cleaner and easier for using, we are not able to provide any props to the keyframe, which is something that is not acceptable.

Please share your thoughts on this, so that we can conclude the final API.

# Update
After discussion, we decided to add animation property to the components, that would encapsulate all things regarding the keyframes and will provide easier usage. The definition of the keyframes in the theme will remain the same, but the usage will be provided in the following manner:
## using keyframe already defined in the theme
```
<Button
  circular
  content="C"
  animation={{
    name: 'spinner',
    duration: '2s',
    iterationCount: 'infinite',
  }}
/>
```
## using custom keyframe

```
const keyframe = ({ startColor, endColor }) => ({
  from: { color: startColor },
  to: { color: endColor },
})

<Icon
  name="umbrella"
  size="big"
  animation={{
    keyframe,
    keyframeParams: { startColor: 'red', endColor: 'yellow' },
    duration: '5s',
    delay: '2s',
    iterationCount: 'infinite',
    direction: 'alternate',
    fillMode: 'backwards',
    timingFunction: 'ease',
  }}
/>
```

The animation property is actually defining all animation-related css properties. The interface looks like this:

```
export type Animation = {
  keyframe?: any // custom keyframe function
  name?: string // theme defined keyframe
  keyframeParams?: object // params for the keyframe function
  delay?: string // animation-delay CSS property
  direction?: string // animation-direction CSS property
  duration?: string // animation-duration CSS property
  fillMode?: string // animation-fill-mode CSS property
  iterationCount?: string // animation-iteration-count CSS property
  playState?: string // animation-play-state CSS property
  timingFunction?: string // animation-timing-function CSS property
}
```


levithomason:
let's keep: **using keyframe already defined in the theme**

remove: **using custom keyframe**

Only the Provider will allow defining keyframes, for now.

We can iterate from there 👍

mnajdova:
# Update
The theme can define various animations through the animations property in the following manner:
```
<Provider theme={{
  animations: {
    colorChanger: {
      keyframe: ({ startColor, endColor }) => ({
        from: { backgroundColor: startColor },
        to: { backgroundColor: endColor },
      } as IStyle),
      keyframeParams: { startColor: 'red', endColor: 'yellow' },
      duration: '3s',
      iterationCount: 'infinite',
    },
 },
}} /> ... </Provider>
```

The usage of this animation then can be define in two manners:
## by specifying the name of the animation
```
<Icon name='umbrella' animation='colorChanger' />
```
## by specifying an animation object
```
<Icon 
  name='umbrella' 
  animation={{
    name: 'colorChanger', 
    duration: '1s', 
    keyframeParams: {
      startColor: 'blue', 
      endColor: 'green'
    }
  }} 
/>
```
With the second approach the user can if necessary change some properties of the animation, without having to redefine it in the Provider's theme.

mnajdova:
# Second approach
Using the new `Transition` component. The `Transition` component is a component that provides the mechanism for executing animation, and it simply wraps the element that needs to be animated. Some example usage:
```
<Transition
  animationName='spinner'>
    <Icon name='umbrella' circular />
</Transition>
```

## Pros
- no need for adding the animation property to all components
- can be used with components outside of the Stardust library

## Cons
- it adds additional div around the animated component
- as the animation is added on the outer div, if the child component has some margins different on some sides, the animation can behave strangely

src/components/Transition/Transition.tsx
```diff
@@ -0,0 +1,109 @@
...
+      duration: this.props.duration,
+      delay: this.props.delay,
+      iterationCount: this.props.iterationCount,
+      direction: this.props.direction,
+      fillMode: this.props.fillMode,
+      playState: this.props.playState,
+      timingFunction: this.props.timingFunction,
+    }
+
+    const animationStyle = createAnimationStyles(animation, theme)
```
levithomason, line 111: To prevent potentially rendering keyframes on prop changes, let's remove keyframe params.  This will require all keyframes to be defined in the Provider and rendered ahead of time.

src/components/Transition/Transition.tsx
```diff
@@ -0,0 +1,109 @@
...
+    /** An element type to render as. */
+    as: customPropTypes.as,
+
+    /**
+     *  Button content for childrenApi
+     *  @docSiteIgnore
+     */
+    children: PropTypes.node,
+
+    /** Delay of the animation. */
```
levithomason: Let's remove these descriptions or update to what fela has.  Ideally, these would give the user more detail about why/how they affect the transition.

src/components/Icon/Icon.tsx
```diff
@@ -44,6 +50,22 @@ class Icon extends UIComponent<Extendable<IconProps>, any> {
   static displayName = 'Icon'
 
   static propTypes = {
+    /** Generic animation property for adding CSS animation to the component's root element */
+    animation: PropTypes.oneOfType([
```
levithomason: Let's pull this to a common place.  All UI components will support this prop as well.

docs/src/examples/components/Icon/Variations/IconAnimated.shorthand.tsx
```diff
@@ -0,0 +1,47 @@
...
+  componentDidMount() {
+    // after 10s pause the animation
+    this.pauseTimeout = setTimeout(() => {
+      this.setState({ playState: 'paused' })
+    }, 10000)
+    // after 20s run the animation again
+    this.runningTimeout = setTimeout(() => {
+      this.setState({ playState: 'running' })
+    }, 20000)
+  }
```
levithomason: Let's simplify this example to not use timeouts.  Also, move it to the Transition examples for now.  We won't show individual component transitions just yet.

docs/src/examples/components/Button/Variations/index.tsx
```diff
@@ -24,6 +24,11 @@ const Variations = () => (
       description="A button can be shown in form of a text to indicate some less-pronounced actions."
       examplePath="components/Button/Variations/ButtonExampleText"
     />
+    <ComponentExample
+      title="Animated"
+      description="A button can have some animation applied on it."
+      examplePath="components/Button/Variations/ButtonAnimated"
+    />
```
levithomason: remove

src/themes/teams/index.ts
```diff
@@ -34,11 +34,36 @@ Object.keys(fontIcons).forEach(iconName => {
   icons[iconName] = declareFontBased(fontIcons[iconName])
 })
 
+const animations = {
+  spinner: {
+    keyframe: () => ({
```
levithomason: For now, we won't accept params to keyframes (to prevent rendering new animations on prop changes).   Instead, just accept an object here and use `callable()` on it internally.

levithomason (approved):
Merge after changes are addressed 👍

PR #418, opened 2018-11-02
Add cypress e2e testing to Stardust

silviuaavram:
Adding cypress to Stardust. Work in progress.

Will add cypress library for e2e testing, and maybe another helper library that will add more capabilities for cypress. Still working on how to add typings for that as well.

Will also create some few tests as part of this PR.

Probably the running of the tests as part of the CI process will not be added here. Remains to be decided.


silviuaavram:
close as we have another PR for this https://github.com/stardust-ui/react/pull/1146

PR #419, opened 2018-11-02
breaking(Button): replace `type` prop with `secondary` and `primary`

### LINKED ISSUE #374: Button: `type` is working incorrecly
## Bug Report

Hi guys, I'm building now some stuff based on Stardust specifications and I found the design problem.

image

image

However, the `type` prop is reserved in `button`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

### Steps

```jsx
<Button content="Click here" type="reset" />
```

### Expected Result

```html
<button class="ui-button" aria-disabled="false" type="reset">
  <span class="ui-slot">Click here</span>
</button>
```

### Actual Result

```html
  <span class="ui-slot">Click here</span>
</button>
```

### Version
0.9.1

layershifter:
Fixes #374, please refer there to get more details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### BREAKING
+- Replace the `type` prop with `secondary` and `primary` for `Button` @layershifter ([#419](https://github.com/stardust-ui/react/pull/419))
+
 ### Fixes
 - Fix endMedia to not be removed from DOM on mouseleave for `ListItem` @musingh1 ([#278](https://github.com/stardust-ui/react/pull/278))
 
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -35,7 +35,6 @@ export interface ButtonProps {
   primary?: boolean
   renderIcon?: ShorthandRenderFunction
   text?: boolean
-  type?: 'type' | 'submit' | 'reset'
   secondary?: boolean
   styles?: ComponentSlotStyle
   variables?: ComponentVariablesInput
@@ -113,9 +112,6 @@ class Button extends UIComponent<Extendable<ButtonProps>, ButtonState> {
     /** A button can be formatted to show only text in order to indicate some less-pronounced actions. */
     text: PropTypes.bool,
 
-    /** The type prop specifies the type of button. */
-    type: PropTypes.oneOf(['button', 'submit', 'reset']),
-
     /** A button can be formatted to show different levels of emphasis. */
     secondary: PropTypes.bool,
 
@@ -155,7 +151,7 @@ class Button extends UIComponent<Extendable<ButtonProps>, ButtonState> {
     styles,
     rest,
   }): React.ReactNode {
-    const { children, content, disabled, iconPosition, type } = this.props
+    const { children, content, disabled, iconPosition } = this.props
     const hasChildren = childrenExist(children)
 
     return (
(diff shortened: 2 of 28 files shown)


src/components/Button/Button.tsx
```diff
@@ -32,9 +32,11 @@ export interface ButtonProps {
   iconPosition?: 'before' | 'after'
   onClick?: ComponentEventHandler<ButtonProps>
   onFocus?: ComponentEventHandler<ButtonProps>
+  primary?: boolean
   renderIcon?: ShorthandRenderFunction
   text?: boolean
-  type?: 'primary' | 'secondary'
+  type?: 'type' | 'submit' | 'reset'
```
bmdalex: can we have 
```jsx
<button type="type">click</button>
```
?

src/components/Button/Button.tsx
```diff
@@ -105,11 +107,17 @@ class Button extends UIComponent<Extendable<ButtonProps>, ButtonState> {
...
     onFocus: PropTypes.func,
 
+    /** A button can be formatted to show different levels of emphasis. */
+    primary: PropTypes.bool,
+
     /** A button can be formatted to show only text in order to indicate some less-pronounced actions. */
     text: PropTypes.bool,
 
+    /** The type prop specifies the type of button. */
+    type: PropTypes.oneOf(['button', 'submit', 'reset']),
```
bmdalex: We don't seem to handle `type` so I think we should remove it from `propTypes` and the interface

src/components/Button/Button.tsx
```diff
@@ -156,6 +164,7 @@ class Button extends UIComponent<Extendable<ButtonProps>, ButtonState> {
         disabled={disabled}
         onClick={this.handleClick}
         onFocus={this.handleFocus}
+        type={type}
```
bmdalex: if we decide to remove type from handled props don't forget to remove it from here as well

bmdalex (changes requested):
pls take a look at comments

src/components/Button/Button.tsx
layershifter: Nope, it was a typo

src/components/Button/Button.tsx
layershifter: Fully agree, I will remove the `type` prop at all

src/components/Button/Button.tsx
```diff
@@ -183,16 +187,14 @@ class Button extends UIComponent<Extendable<ButtonProps>, ButtonState> {
...
 
     if (disabled) {
       e.preventDefault()
       return
     }
 
-    if (onClick) {
-      onClick(e, this.props)
-    }
+    _.invoke(this.props, 'onClick', e, this.props)
```
layershifter, line 197: Simplify with `invoke()`

miroslavstastny (commented):
The added test is failing.

If we decided to remove `type` for `Button` we should do the same for `Menu` and `Divider` in the same release.

miroslavstastny (approved):
Don't forget to update `CHANGELOG.md`

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixes
 - Fix endMedia to not be removed from DOM on mouseleave for `ListItem` @musingh1 ([#278](https://github.com/stardust-ui/react/pull/278))
+- Replace the `type` prop with `secondary` and `primary` for `Button` @layershifter ([#419](https://github.com/stardust-ui/react/pull/419))
```
miroslavstastny: This should  go to BREAKING CHANGES!

miroslavstastny (changes requested):

bmdalex (approved):

CHANGELOG.md
layershifter: Nice catch, done

bmdalex:
agreed with @miroslavstastny
@layershifter  
do we have an issue/ticket to track the other `type` props (for `Menu` and `Divider `)?
we should have PRs for these soon, i can help with that
fyi @alinais

miroslavstastny (approved):

Issue #388, opened 2018-10-23
Dropdown component


miroslavstastny:
Add basic implementation for Dropdown component.

![image](https://user-images.githubusercontent.com/9615899/47333859-ce4d4e00-d684-11e8-8db6-c804ea4f0fbd.png)

## API
For now, use an API similar to SUIR ([example](https://react.semantic-ui.com/modules/dropdown/#types-multiple-search-selection)):
```jsx
<Dropdown multiple search items={[ ... ]} />
```

Based on prototype in #334:

Since we only support "search" and "multiple", make these required for now.
Dropdown will only include the Input and associated List (menu), no buttons, field label, etc.

Separate `active` items from `selected`:
- `active` means the item is in input box as an active value
- `selected` means the item is highlighted in the menu

More keyboard handlers:
 - click to open
 - esc to close

### className
Additional classes.

### fluid
Boolean to make it stretch as much as container allows.

### items
Shorthand to provide items in the dropdown. Similar to providing ListItems as shorthand inside list.
<img width="429" alt="screen shot 2018-10-25 at 11 16 27" src="https://user-images.githubusercontent.com/11275392/47490325-c9d19280-d848-11e8-8629-7a82db9a8d0c.png">

### label
Shorthand to provide a label before the dropdown.
<img width="456" alt="screen shot 2018-10-25 at 11 16 48" src="https://user-images.githubusercontent.com/11275392/47490302-c0482a80-d848-11e8-929b-84b855ba3589.png">

### multiple
Boolean to provide multi-selection functionality. Will add the active shorthands as Labels before the Input, each with 'X' icon for removal.
<img width="480" alt="screen shot 2018-10-25 at 11 17 33" src="https://user-images.githubusercontent.com/11275392/47490343-d48c2780-d848-11e8-8c83-f563ea19f61a.png">

### onChange
Callback provided to allow user to handle change. Invoked with the props and also with the active option(s), which can be either the shorthand for the active item, or the shorthand collection for the active items.

### onInput
Callback provided to allow user to handle input change in the edit text. Only makes sense if dropdown is a search. Invoked with both the props and the new value of the text in the input.

### placeholder
Text to act as placeholder, optional, in case the dropdown is a search (with text field). Makes sense only if dropdown is a search.
<img width="427" alt="screen shot 2018-10-25 at 11 17 40" src="https://user-images.githubusercontent.com/11275392/47490362-dc4bcc00-d848-11e8-9f99-968d513eecca.png">

[...]

PR #422, opened 2018-11-02
Feat/multi dropdown component

silviuaavram:
Focus on creating the multiple pick dropdown.

Issue: https://github.com/stardust-ui/react/issues/388


src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,348 @@
...
+
+    /** A dropdown can have a search field. */
+    search: PropTypes.bool,
+
+    /** Additional CSS styles to apply to the component instance.  */
+    styles: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
+  }
+
+  state: DropdownState = {
+    active: this.props.multiple ? [] : null,
```
sophieH29: ```active``` is more suitable for boolean values.  I think ```activeItems``` is more descriptive for this case.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,348 @@
...
+            getMenuProps,
+            isOpen,
+            highlightedIndex,
+            selectItemAtIndex,
+            getRootProps,
+          }) => {
+            return (
+              <React.Fragment>
+                <div style={styles.containerDiv} onClick={this.onContainerClick.bind(this, isOpen)}>
+                  <span aria-live="assertive" style={styles.ariaLiveSpan}>
```
sophieH29: ```aria-*``` attributes as part of accessibility behaviors?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,362 @@
...
+  onDropdownChange = element => {
+    if (this.props.multiple) {
+      const newActive = [...this.state.active, element]
+
+      _.invoke(this.props, 'onChange', newActive)
+
+      this.setState({
+        active: newActive,
+        inputValue: '',
+        message: `${element.header} has been selected.`,
```
jurokapsiar: as we will need to be able to translate this text, we should probably pass a callback and evaluate it here, something like
```
message: props.translateSelectedMessage ? props.translateSelectedMessage(element) : `${element.header} has been selected.`
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,362 @@
...
+      active = active.filter(currentElement => currentElement !== element)
+    } else {
+      poppedElement = active.pop()
+    }
+
+    _.invoke(this.props, 'onChange', active)
+
+    this.setState({
+      active,
+      message: `${poppedElement.header} has been removed.`,
```
jurokapsiar: as we will need to be able to translate this text, we should probably pass a callback and evaluate it here, something like
```
message: props.translateRemovedMessage ? props.translateRemovedMessage(poppedElement) : `${element.header} has been removed.`
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,362 @@
...
+            styles={styles.activeListLabel}
+            circular
+            key={`active-item-${index}`}
+            content={item.header}
+            image={{ src: item.image, avatar: true }}
+            icon={{
+              name: 'close',
+              onClick: this.onCloseIconClick.bind(this, item),
+              onKeyDown: this.onCloseIconKeyDown.bind(this, item),
+              'aria-label': `Remove ${item.header} from selection.`,
```
jurokapsiar: translation comments apply here as well

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,362 @@
...
+
+    /** A message to serve as placeholder. */
+    placeholder: PropTypes.string,
+
+    /** A dropdown can have a search field. */
+    search: PropTypes.bool,
+
+    /** Additional CSS styles to apply to the component instance.  */
+    styles: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
+  }
```
jurokapsiar, line 181: we will also need to make downshift's `getA11StatusMessage` property available as our `Dropdown` property. This will allow consumers of the component to translate the message to desired language. See https://github.com/paypal/downshift#geta11ystatusmessage

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,362 @@
...
+    const selectedPropIfMultiple = { ...(multiple && { selectedItem: null }) }
+
+    return (
+      <ElementType>
+        <Downshift
+          onChange={this.onDropdownChange}
+          inputValue={inputValue}
+          {...selectedPropIfMultiple}
+          stateReducer={this.stateReducer}
+          itemToString={(item: DropdownListItem) => (item ? item.header : '')}
```
jurokapsiar: If we let consumers to pass `items` we should probably also allow them to pass `itemToString` (and maybe fallback to default if not provided). My worry is that if `item` is shorthand, it might be a string, component or functional component as well. @mnajdova we will need your advice here.

package.json
```diff
@@ -113,6 +113,7 @@
     "cross-env": "^5.1.4",
     "doctoc": "^1.3.0",
     "doctrine": "^2.0.0",
+    "downshift": "^3.0.0",
```
miroslavstastny: -> dependencies

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,369 @@
...
+      />
+    )
+  }
+
+  private renderToggleButton(getToggleButtonProps, styles, isOpen) {
+    return (
+      <Icon
+        name={`chevron ${isOpen ? 'up' : 'down'}`}
+        as="button"
+        tabindex="-1"
```
miroslavstastny: -> tabIndex

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,369 @@
...
+            getItemProps,
+            getMenuProps,
+            getRootProps,
+            getToggleButtonProps,
+            isOpen,
+            highlightedIndex,
+            selectItemAtIndex,
+          }) => {
+            return (
+              <div style={styles.containerDiv} onClick={this.onContainerClick.bind(this, isOpen)}>
```
miroslavstastny: You cannot apply styles using `style` prop as it does not work for RTL.
To see the issue, check `<Dropdown toggleButton />` in RTL (`paddingRight: '2rem'` breaks the layout):
image

Use `className={classes.containerDiv}`

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,369 @@
...
+  multiple?: boolean
+  onChange?: (active: DropdownListItem[]) => any
+  placeholder?: string
+  search?: boolean
+  styles?: ComponentSlotStyle<DropdownProps, DropdownState>
+  toggleButton?: boolean
+}
+
+export interface DropdownState {
+  active: DropdownListItem[]
```
miroslavstastny: `active` and `inputValue` should be autocontrolled props.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,369 @@
...
+    placeholder: PropTypes.string,
+
+    /** A dropdown can have a search field. */
+    search: PropTypes.bool,
+
+    /** Additional CSS styles to apply to the component instance.  */
+    styles: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
+
+    /** A dropdown can have a toggle button. */
+    toggleButton: PropTypes.bool,
```
miroslavstastny, line 160: `toggleButton` does not work for me. Clicking on it when dropdown is open closes it but then it is immediately reopened.

miroslavstastny:
It also drops down under code editor in docsite:
image

src/components/Dropdown/Dropdown.tsx
silviuaavram: I have renamed it with `value` and inputValue with `searchQuery` . Took the names from semantic UI. `activeItems` is a very good suggestion, it's just that sometimes it will be only one item.

package.json
silviuaavram: thanks, good catch!

src/components/Dropdown/Dropdown.tsx
silviuaavram: forgot to push, fixed it, thanks!

src/components/Dropdown/Dropdown.tsx
silviuaavram: it has an issue right now, when it is open, it will close and then open it again. I am still thinking on a proper fix.

Downshift checks if the button is part of the `combobox` div and if not, it will open the list. But I had to remove both the list and the toggle button from the `combobox` div because of screen reader navigation issue. And that's how the bug appeared.

Still thinking on how to fix this.

src/components/Dropdown/Dropdown.tsx
silviuaavram: added the following props:
`  getA11yStatusMessage?: (options: A11yStatusMessageOptions<DropdownListItem>) => string
  getA11ySelectedMessage?: (item: DropdownListItem) => string
  getA11yRemovedMessage?: (item: DropdownListItem) => string
  getA11yRemoveItemMessage?: (item: DropdownListItem) => string
  getNoResultsMessage?: () => string
`

src/components/Dropdown/Dropdown.tsx
silviuaavram: made the whole component autocontrolled and made necessary changes

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,460 @@
...
+import _ from 'lodash'
+import Icon from '../Icon/Icon'
+
+export interface DropdownProps {
+  className?: string
+  fluid?: boolean
+  getA11yStatusMessage?: (options: A11yStatusMessageOptions<DropdownListItem>) => string
+  getA11ySelectedMessage?: (item: DropdownListItem) => string
+  getA11yRemovedMessage?: (item: DropdownListItem) => string
+  getA11yRemoveItemMessage?: (item: DropdownListItem) => string
```
sophieH29: I think we should aim to keep all accessibility staff in behaviors. Already discussed with @silviuavram, as this PR is big enough, we have to consider to make it in the future PR.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,460 @@
...
+    defaultValue: PropTypes.oneOfType([PropTypes.object, PropTypes.array]),
+
+    /** An input can take the width of its container. */
+    fluid: PropTypes.bool,
+
+    /** Array of props for generating dropdown items and selected item labels if multiple selection. */
+    items: PropTypes.collectionShorthand,
+
+    /** A function that creates custom accessability message for dropdown status.
+     * @param {Object} messageGenerationProps - Object with properties to generate message from. See getA11yStatusMessage from Downshift reoi,
```
sophieH29: typo Downshift ```reoi```?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,460 @@
...
+
+  onContainerClick = isOpen => {
+    !isOpen && this.inputRef.focus()
+  }
+
+  onInputKeyUpIfMultiple = event => {
+    const { searchQuery, value, backspaceDelete } = this.state
+
+    switch (keyboardKey.getCode(event)) {
+      case keyboardKey.Backspace:
```
sophieH29: also could be moved to ```actionHandlers``` as for future

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,460 @@
...
+    return (
+      <List
+        {...getMenuProps({ refKey: 'listRef' })}
+        styles={styles.list}
+        aria-hidden={!isOpen}
+        items={isOpen ? this.renderItems(variables, getItemProps, highlightedIndex) : []}
+      />
+    )
+  }
+
```
sophieH29, line 330: The thing to consider: this is a lot of code for one component, which brings complexity for maintaining it in the future. 
What about if we can break this component into small parts, the same way as we have ```Menu``` and ```MenuItems``` for example.
We can have next set of drop-down components, like ```DropdownList```, ```DropdownInput``` etc.
So it can be used during rendering like

```js
 <Downshift
    ...

   { search && 
      <DropdownInput styles={styles} getRootProps={getRootProps} getInputProps={getInputProps}...> }
     <DropdownList styles={styles} getMenuProps={getMenuProps} getItemProps={getItemProps} isOpen= . 
      {isOpen}...>
    ...
<Downshift>

```

It's very easy to get lost now, so it can be even harder in the future.
Let me know what you think.

src/components/Dropdown/Dropdown.tsx
silviuaavram: created tracking issue for this and will address it after this PR gets in https://github.com/stardust-ui/react/issues/468

src/components/Dropdown/Dropdown.tsx
silviuaavram: added as optional prop

src/components/Dropdown/Dropdown.tsx
silviuaavram: repo. thanks :)

src/components/Dropdown/Dropdown.tsx
silviuaavram: it is a point to consider, as I know that Semantic does something similar.

Right now the code is grouped into separate functions, so it's not that bad for reading it. We will follow up on this, thanks!

@jurokapsiar

src/components/Dropdown/Dropdown.tsx
silviuaavram: got it, also looked at a similar use in List.tsx. Thank you for pointing this out!

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,671 @@
...
+                  : `Remove ${item.header} from selection.`,
+                'aria-hidden': false,
+                role: 'button',
+              }}
+            />
+          )
+        })
+  }
+
+  Wrapper = ({ children }) => (
```
sophieH29: What do you think if to rename ```Wrapper``` -> ```RefWrapper```? I saw first a usage of it ```<this.Wrapper>``` and was a bit confused what is it about

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,671 @@
...
+  }
+
+  static defaultProps = {
+    as: 'div',
+  }
+
+  static autoControlledProps = ['searchQuery', 'value']
+
+  state: DropdownState = {
+    backspaceDelete: this.props.multiple ? true : undefined,
```
sophieH29: is there any reason to set ```backspaceDelete``` to ```undefined```? I meant, why not just ```backspaceDelete: this.props.multiple```?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,671 @@
...
+    styles,
+    variables,
+    rest,
+  }: RenderResultConfig<DropdownProps>) {
+    const { search, multiple, toggleButton, getA11yStatusMessage, itemToString } = this.props
+    const { searchQuery } = this.state
+    const optionalDownshiftProps = {
+      // in multiple dropdown, we hold active values in the array, and default active is null.
+      ...(multiple && { selectedItem: null }),
+    }
```
sophieH29: I think we can make it simpler, without ```...``` operation
```js
const optionalDownshiftProps = multiple && { selectedItem: null } || {}
```

Let me know what you think

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,671 @@
...
+
+    return (
+      <ElementType {...rest} className={classes.root}>
+        <Downshift
+          onChange={this.handleChange}
+          inputValue={searchQuery}
+          stateReducer={this.stateReducer}
+          getA11yStatusMessage={getA11yStatusMessage || this.getA11yStatusMessage}
+          itemToString={
+            itemToString ? itemToString : (item: DropdownListItem) => (item ? item.header : '')
```
sophieH29: can be simplified a bit
```js
itemToString || (item: DropdownListItem) => (item ? item.header : '')
```

src/themes/teams/components/List/listItemVariables.ts
```diff
@@ -1,6 +1,7 @@
 export default siteVariables => ({
   // Header
   // TODO: prod app uses 17.5px here, it should be 16px per the design guide!
+  headerColor: siteVariables.gray02,
```
mnajdova: Can you move the comment, so it is above the line that is actually about.

src/themes/teams/components/List/listItemStyles.ts
```diff
@@ -49,6 +49,7 @@ const listItemStyles: ComponentSlotStylesInput<ListItemProps, any> = {
     }
   },
   header: ({ variables }): ICSSInJSStyle => ({
+    color: variables.headerColor,
```
mnajdova: Should this be changed for all list items?

src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -0,0 +1,80 @@
...
+    position: 'absolute',
+    zIndex: 1000,
+    maxHeight: listMaxHeight,
+    overflowY: 'auto',
+    width: fluid ? '100%' : width,
+    top: 'calc(100% + 2px)', // leave room for container + its border
+  }),
+
+  listItemRoot: ({ variables: { listItemBackgroundColor } }): ICSSInJSStyle => ({
+    backgroundColor: listItemBackgroundColor,
```
mnajdova, line 54: Any specific reason why we are using the wording root here? We don't have it on any other places when defining the slot's styles

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,671 @@
...
+    ]
+  }
+
+  private renderActiveValues(styles: ComponentSlotStylesInput) {
+    const value = this.state.value as DropdownListItem[]
+    const { getA11yRemoveItemMessage } = this.props
+
+    return value.length === 0
+      ? null
+      : value.map((item, index) => {
```
sophieH29: ```index``` is not used

src/components/Slot/Slot.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import * as _ from 'lodash'
```
mnajdova: Please remove this import as it is not used.

src/components/List/List.tsx
```diff
@@ -64,6 +71,7 @@ class List extends UIComponent<Extendable<ListProps>, ListState> {
     accessibility: PropTypes.func,
     debug: PropTypes.bool,
     items: customPropTypes.collectionShorthand,
+    innerRef: PropTypes.func,
```
mnajdova: Didn't we decide that we will use the Ref component instead of adding this? Are you updated to latest master?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,671 @@
...
+        },
+      },
+    ]
+  }
+
+  private renderActiveValues(styles: ComponentSlotStylesInput) {
+    const value = this.state.value as DropdownListItem[]
+    const { getA11yRemoveItemMessage } = this.props
+
+    return value.length === 0
```
sophieH29: nit: just for better reading

```js
if(value.length === 0) return null

return value.map(item => {
  ...
})
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,671 @@
...
+import Icon from '../Icon/Icon'
+import { commonUIComponentPropTypes } from '../../lib/commonPropTypes'
+import Ref from '../Ref/Ref'
+
+// To be replaced when Downshift will add highlightedItem in their interface.
+export interface DownshiftA11yStatusMessageOptions<Item> extends A11yStatusMessageOptions<Item> {
+  highlightedItem: Item
+}
+
+export interface DropdownProps {
```
mnajdova: You should extends the props with the UIComponentProps. I don't see here any of the common props (styles, variables, as etc.)

src/components/List/List.tsx
silviuaavram: We need that prop in the component already, so downshift callback prop will get the actual html node and add its stuff to it, here: `{...getMenuProps({ refKey: 'innerRef' })}` . If I don't specify it in the component, it won't work. I am doing the <Ref> magic in Dropdown as an HOC, but I still need this to be available as a handled prop. @kuzhelov @mnajdova

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,671 @@
...
+  }
+
+  private handleContainerClick = (isOpen: boolean, e: React.SyntheticEvent) => {
+    !isOpen && this.inputRef.focus()
+
+    _.invoke(this.props, 'onContainerClick', e, { ...this.props, isOpen })
+  }
+
+  private handleBackspaceDelete = (e: React.SyntheticEvent) => {
+    switch (keyboardKey.getCode(e)) {
```
sophieH29: I think there is no need for switch as we are only handling one case

```js
if(keyboardKey.getCode(e) !==  keyboardKey.Backspace) return

// rest of the code
...
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,671 @@
...
+  private handleInputKeyDown = (
+    highlightedIndex: number,
+    selectItemAtIndex: (
+      index: number,
+      otherStateToSet?: Partial<StateChangeOptions<any>>,
+      cb?: () => void,
+    ) => void,
+    e: React.SyntheticEvent,
+  ) => {
+    switch (keyboardKey.getCode(e)) {
```
sophieH29: I think there is no need for switch as we are only handling one case

```js
if(keyboardKey.getCode(e) === keyboardKey.Tab) {
 if (highlightedIndex !== undefined) {
          selectItemAtIndex(highlightedIndex)
 }
else {
 _.invoke(this.props, 'onInputKeyDown', e, {
          ...this.props,
          highlightedIndex,
          selectItemAtIndex,
        })
   }
}
```

src/components/Slot/Slot.tsx
sophieH29: as I see lodash is used for ```_.invoke``` multiple times

src/components/Dropdown/Dropdown.tsx
silviuaavram: renamed to RefProvidingWrapper :)

src/themes/teams/components/List/listItemVariables.ts
silviuaavram: oups :D

src/themes/teams/components/List/listItemStyles.ts
silviuaavram: The reason I added this is to be able to change it from Dropdown, on item highlight. And the default color was the same before this change anyway. Do you think we have a better alternative?

src/themes/teams/components/Dropdown/dropdownStyles.ts
silviuaavram: probably a leftover from previous styling. Thanks for pointing it out, will remove root.

src/components/Slot/Slot.tsx
silviuaavram: you are both right. in the meantime though I removed the code from Slot that used `invoke` (used that for the Ref prop callback). Will remove lodash.

src/components/Dropdown/Dropdown.tsx
silviuaavram: `export interface DropdownProps extends UIComponentProps<any, any> {`

src/components/Dropdown/Dropdown.tsx
silviuaavram: changed, thank you. also will add a small comment for what I am using this boolean.

src/components/Dropdown/Dropdown.tsx
silviuaavram: Probably a leftover. I had more than one param at some point. Will refactor, thank you :)

src/components/Dropdown/Dropdown.tsx
silviuaavram: `itemToString={itemToString || ((item: DropdownListItem) => (item ? item.header : ''))}`

src/components/Dropdown/Dropdown.tsx
silviuaavram: cleaned, thanks!

src/components/Dropdown/Dropdown.tsx
silviuaavram: changed, thank you

src/components/Dropdown/Dropdown.tsx
silviuaavram: refactored:
     

    if (keyboardKey.getCode(e) === keyboardKey.Tab && highlightedIndex !== undefined) {
      selectItemAtIndex(highlightedIndex)
    }

    _.invoke(this.props, 'onInputKeyDown', e, {
      ...this.props,
      highlightedIndex,
      selectItemAtIndex,
    })`

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,664 @@
...
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props and proposed value.
+   */
+  onBackspaceDelete?: ComponentEventHandler<DropdownProps>
+
+  /**
+   * Callback for change in dropdown active value(s).
+   * @param {DropdownListItem|DropdownListItem[]} value - Dropdown active value(s).
+   */
+  onDropdownChange?: (value: DropdownListItem | DropdownListItem[]) => any
```
kuzhelov: this prop should be split on two - one for each autocontrolled prop:
- `onChange` - for value change (no need to repeat `dropdown` word)
- `onSearchQueryChange` - for `searchQuery` changes

Note that the naming convention is:
- 'onChange' - for `value` prop changes
- 'onXChange' - for `X` prop changes

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,664 @@
...
+   */
+  onContainerClick?: ComponentEventHandler<DropdownProps>
+
+  /**
+   * Called on edit text blur.
+   *
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props and proposed value.
+   */
+  onInputBlur?: ComponentEventHandler<DropdownProps>
```
kuzhelov: `input` should be exposed as a slot of `Dropdown`, so that client will be able to pass all the necessary event subscriptions to it directly. This will prevent us from polluting Dropdown's API, as well as allow to reuse already existing Stardust merits:

```jsx
<Dropdown input={{ onFocus: (e) => ..., onKeyDown: (e) => ... }}>
```

Same story generally applies to `label` and `close icon`

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,664 @@
...
+   */
+  onDropdownChange?: (value: DropdownListItem | DropdownListItem[]) => any
+
+  /**
+   * Called on clicking the 'X' icon corresponding to an active value.
+   *
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props and proposed value.
+   */
+  onCloseIconClick?: ComponentEventHandler<DropdownProps>
```
kuzhelov: these `onCloseIconXXX` props could be omitted from Dropdown's API - instead of that `close icon` should be exposed as a slot, so that client will be able to utilize slot's benefits directly:

```jsx
<Dropdown closeIcon={{ onClick: (e) => ..., onKeyDown: (e) => ... }}>
```

With this approach we won't pollute Dropdown's API.

Also, currently thinking on whether `closeIcon` is a good name for this slot (doubts about `icon` word used - this is rather UI-related aspect, not just an abstract term).

src/components/Dropdown/Dropdown.tsx
kuzhelov: still, even now we can better shape the API that is provided by Dropdown, by combining all these related props and expose them within a single dedicated prop (`a11y`) to client:

```jsx
<Dropdown a11y={
    getStatusMessage: () => ...,
    getSelectedMessage: () => ....,
  } />
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,664 @@
...
+  onCloseIconKeyDown?: ComponentEventHandler<DropdownProps>
+
+  /**
+   * Called on container click. Especially useful for multiple search case, when user
+   * can click the container outside the edit text or active values.
+   *
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props and proposed value.
+   */
+  onContainerClick?: ComponentEventHandler<DropdownProps>
```
kuzhelov: @silviuavram, could you, please, provide code example with real use case when it might be necessary (maybe you could refer line from people picker prototype?). Thank you!

kuzhelov (changes requested):
lets start with cleaning up the exposed API a bit - currently there are too much props provided to the client

src/components/Dropdown/Dropdown.tsx
silviuaavram: I cannot, as I will need to provide an Event in the callback if I will name it 'onChange', and Downshift is not providing me one. :( :( Any idea? Oh yes, and there is already a `onSearchQueryChange` but that one does not have an Event either.

src/components/Dropdown/Dropdown.tsx
silviuaavram: unfortunately, I cannot, so it may be better not to expose this to the user. I will remove the prop, and keep the private handler, that will focus the input each time a click is received in the dropdown container.

docs/src/examples/components/Dropdown/Types/DropdownExampleMultipleSearch.shorthand.tsx
```diff
@@ -0,0 +1,74 @@
...
+          this.setState({
+            items: inputItems.filter(item => value.indexOf(item) === -1),
+          })
+        }}
+      />
+    )
+  }
+}
+
+const inputItems: DropdownListItem[] = [
```
bmdalex: nit: we usually declare data at the top before using it

docs/src/examples/components/Dropdown/Variations/DropdownExampleMultipleSearchFluid.shorthand.tsx
```diff
@@ -0,0 +1,75 @@
...
+          this.setState({
+            items: inputItems.filter(item => value.indexOf(item) === -1),
+          })
+        }}
+      />
+    )
+  }
+}
+
+const inputItems: DropdownListItem[] = [
```
bmdalex: nit: we usually declare data at the top before using it

docs/src/examples/components/Dropdown/Variations/DropdownExampleMultipleSearchTextItems.shorthand.tsx
```diff
@@ -0,0 +1,55 @@
...
+          this.setState({
+            items: inputItems.filter(item => value.indexOf(item) === -1),
+          })
+        }}
+      />
+    )
+  }
+}
+
+const inputItems: DropdownListItem[] = [
```
bmdalex: nit: we usually declare data at the top before using it

src/components/Slot/Slot.tsx
```diff
@@ -11,11 +11,19 @@ import {
...
   extends UIComponentProps<SlotProps, any>,
     ContentComponentProps,
-    ChildrenComponentProps {}
+    ChildrenComponentProps {
+  /**
+   * Ref callback with the input wrapper DOM node.
+   *
+   * @param {JSX.Element} node - input wrapper DOM node.
+   */
+  innerRef?: (node: HTMLElement) => void
```
bmdalex: how does this work? there's no functionality added; if it's something coming from `UIComponent` shouldn't this prop be part of the common props API? fyi @mnajdova

src/themes/teams/components/Dropdown/dropdownVariables.ts
```diff
@@ -0,0 +1,40 @@
...
+  editTextFlexBasis: string
+  listItemBackgroundColor: string
+  listItemHighlightedBackgroundColor: string
+  listItemTextColor: string
+  listMaxHeight: string
+  toggleButtonSize: string
+  width: string
+}
+
+const [_2px_asRem, _3px_asRem] = [2, 3].map(v => pxToRem(v))
```
bmdalex, line 19: 👍

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,664 @@
...
+  defaultSearchQuery?: string
+
+  /** The default value for the dropdown. */
+  defaultValue?: DropdownListItem | DropdownListItem[]
+
+  /** A dropdown can take the full width of its container. */
+  fluid?: boolean
+
+  /**
+   * A function that creates custom accessability message for dropdown status.
```
bmdalex: nit: ~~`accessability`~~ -> `accessibility`

src/components/Dropdown/Dropdown.tsx
bmdalex: +1 much cleaner API

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,664 @@
...
+              : `Remove ${item.header} from selection.`,
+            'aria-hidden': false,
+            role: 'button',
+          }}
+        />
+      )
+    })
+  }
+
+  RefProvidingWrapper = ({ children }) => (
```
bmdalex: this functional component has no connection with this component whatsoever, I think it should be extracted as a global in this file or imported from some helper; also, a better name could be used; suggestion:
```jsx
import { ReactChildren } from 'types/utils'

const RefWrapper = ({ children }: { children: ReactChildren }) => (
  <Ref innerRef={domNode => _.invoke(children.props, 'innerRef', domNode)}>{children}</Ref>
)
```
usage

```jsx
<RefWrapper>
  <Input ... />
</RefWrapper>
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,664 @@
...
+
+    _.invoke(this.props, 'onContainerClick', e, { ...this.props, isOpen })
+  }
+
+  private handleBackspaceDelete = (e: React.SyntheticEvent) => {
+    if (keyboardKey.getCode(e) === keyboardKey.Backspace) {
+      const { searchQuery, value, backspaceDelete } = this.state
+
+      if (searchQuery === '' && (value as DropdownListItem[]).length > 0) {
+        if (!backspaceDelete) {
```
bmdalex: nit: if / else block easier to read with truthy asertion first
```typescript
if (backspaceDelete) {
  // backspaceDelete true
} else {
  // backspaceDelete false
}
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,664 @@
...
+}
+
+export interface DropdownState {
+  value: DropdownListItem | DropdownListItem[]
+  backspaceDelete: boolean
+  focused: boolean
+  searchQuery?: string
+}
+
+export interface DropdownListItem {
```
bmdalex: nit: please move above `DropdownProps` definition to keep order of usage and logic of file:
```typescript
// External Imports
// Local Imports
// Local and external interface
// Component Props interface
// Component State interface
// Component Implementation
```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,664 @@
...
+    } else {
+      poppedItem = value.pop()
+    }
+
+    this.trySetState({
+      value,
+    })
+    this.setA11yStatus(
+      getA11yRemovedMessage
+        ? getA11yRemovedMessage(item)
```
bmdalex: shouldn't this be `getA11yRemovedMessage(poppedItem)` ?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,664 @@
...
+    if (!highlightedItem && resultCount !== previousResultCount) {
+      return `${resultCount} result${
+        resultCount === 1 ? 'is' : 's are'
+      } available, use up and down arrow keys to navigate. Press Enter key to select.`
+    }
+    return ''
+  }
+
+  private setA11yStatus = (statusMessage: string) => {
+    const elementId = 'stardust-dropdown-a11y-status'
```
bmdalex, line 442: this whole method does not look good at all... any way we can use React approach? I'm sure you wanted that, so I'm asking what's the limitation?

docs/src/examples/components/Dropdown/Types/DropdownExampleMultipleSearch.shorthand.tsx
```diff
@@ -0,0 +1,74 @@
...
+    items: inputItems,
+  }
+  render() {
+    return (
+      <Dropdown
+        multiple
+        search
+        placeholder="Start typing a name"
+        items={this.state.items}
+        onDropdownChange={(value: DropdownListItem[]) => {
```
bmdalex: can we avoid TS code in this example?

docs/src/examples/components/Dropdown/Variations/DropdownExampleMultipleSearchFluid.shorthand.tsx
bmdalex: also pls avoid TS code in example, hence don't cast this to `DropdownListItem[]`

bmdalex (dismissed):
I apologize for the late review but I think there are quite a few of issues that need addressing. Please take a look at the comments and the following bugs I found from manual testing:

#### 1. filtering does not work in the examples

I think we should implement some default filtering method for the dropdown items. TBD if this PR or next one, but please create an issue with follow-up items if we choose to go with the latter.

broken-filter

#### 2. dropdown without `search` prop provided collapses the input if no option selected

no-search

#### 3. clicking on `^` button does not close the dropdown

no-close

#### 4. hovering on the dropdown list items changes the currently selected item for keyboard navigation

hover-broken

#### 5. dropdown options need a border

screenshot 2018-11-23 at 14 33 45

#### 6. scrollbar is translucent

screenshot 2018-11-23 at 14 50 29

#### 7. weird small curly borders on the input

screenshot 2018-11-23 at 14 50 01
screenshot 2018-11-23 at 14 59 09

src/components/Dropdown/DropdownItem.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+export interface DropdownItemProps extends UIComponentProps<any, any> {
+  /** Text content for the item description. */
+  content?: string
+
+  /** The main text displayed for the item. */
+  header?: string
+
+  /** The item is highlighted inside the list. */
+  highlighted?: boolean
```
kuzhelov: currently we are either using `selection` prop (`ListItem`) or `active` (`MenuItem`) for that - lets use `active` here as well, just to constrain the entropy.

src/components/Dropdown/DropdownItem.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+  public renderComponent({ classes, variables, rest }: RenderResultConfig<DropdownItemProps>) {
+    const { content, header, highlighted, image } = this.props
+    return (
+      <ListItem
+        className={classes.root}
+        header={header}
+        onClick={this.handleClick}
+        variables={{
+          backgroundColor: highlighted
```
kuzhelov: all these calculations should be part of `DropdownItem` styles - there shouldn't be any `variables` in the component's code, as this is a theme-related aspect

src/components/Dropdown/DropdownItem.tsx
```diff
@@ -0,0 +1,100 @@
...
+            ? variables.listItemHighlightedBackgroundColor
+            : variables.listItemBackgroundColor,
+          ...(highlighted && {
+            headerColor: variables.listItemTextColor,
+            contentColor: variables.listItemTextColor,
+          }),
+        }}
+        {...{
+          media: image && <Image src={image} avatar />,
```
kuzhelov: would suggest to name this prop as `imageSrc`, to better indicate semantics of its value.

src/components/Dropdown/DropdownItem.tsx
```diff
@@ -0,0 +1,100 @@
...
+  content?: string
+
+  /** The main text displayed for the item. */
+  header?: string
+
+  /** The item is highlighted inside the list. */
+  highlighted?: boolean
+
+  /** The item can have an image avatar. */
+  image?: ShorthandValue
```
kuzhelov, line 28: not sure that `Dropdown`'s API that has `image` prop is generic enough - I would rather expect simple Text to be a default component for the dropdown's item. In contrast, now client might think that `image` is something that is absolutely generic for dropdown's item, which is not

src/components/Dropdown/DropdownLabel.tsx
```diff
@@ -0,0 +1,140 @@
...
+   *
+   * @param {React.ReactType} Component - The computed component for this slot.
+   * @param {object} props - The computed props for this slot.
+   * @param {ReactNode|ReactNodeArray} children - The computed children for this slot.
+   */
+  renderIcon?: ShorthandRenderFunction
+}
+
+/**
+ * A DropdownLabel is a sub-component if the Dropdown.
```
kuzhelov: typo: 

> **of** the Dropdown

src/components/Slot/Slot.tsx
```diff
@@ -31,6 +39,7 @@ const Slot: CreateComponentReturnType<SlotProps> & {
     ...commonUIComponentPropTypes,
     ...contentComponentPropsTypes,
     ...childrenComponentPropTypes,
+    innerRef: PropTypes.func,
```
kuzhelov: do we really need it now? why not to use `Ref` component?

src/components/List/List.tsx
kuzhelov: let me help you with that - we definitely shouldn't expose this prop for now for this component only, as we have workarounds at the moment. Refs for components, if they would be exposed to the API, should be exposed for all in as single step, for the sake of preserving API consistency

src/components/Dropdown/DropdownItem.tsx
```diff
@@ -0,0 +1,100 @@
...
+   * @param {object} props - The computed props for this slot.
+   * @param {ReactNode|ReactNodeArray} children - The computed children for this slot.
+   */
+  renderImage?: ShorthandRenderFunction
+}
+
+/**
+ * A DropdownItem is a sub-component if the Dropdown.
+ */
+class DropdownItem extends UIComponent<Extendable<DropdownItemProps>, any> {
```
mnajdova, line 43: Add these sub-components as a static props in the Dropdown, so that they can be used in the following manner: `Dropdown.Item`, `Dropdown.Label` etc. Also, would encourage you to add some more descriptive message for what the component actually is, rather then it is a sub-component of the Dropdown.

docs/src/examples/components/Dropdown/Types/DropdownExampleMultipleSearch.shorthand.tsx
silviuaavram: I know, it's just that so far we have 3,4 array items, and I chose to have 8-9. I thought it will be weird for the user to pop-up the editor and scroll to the very bottom to see the actual usage.

If you think it's still better than the current approach I will address it.

docs/src/examples/components/Dropdown/Variations/DropdownExampleMultipleSearchFluid.shorthand.tsx
silviuaavram: removing the typing in the examples

src/components/Slot/Slot.tsx
silviuaavram: just have to declare it so I don't get the handled Prop warning. I am getting the ref with the Ref component, using an HOC over the List and Input. See RefProvidingWrapper in this PR. need to update that as well, to have it only in one place and do another check.

src/components/Dropdown/Dropdown.tsx
silviuaavram: changed, thanks!

src/components/Dropdown/Dropdown.tsx
silviuaavram: removed this already.

src/components/Dropdown/Dropdown.tsx
silviuaavram: oups! refactoring.

src/components/Dropdown/Dropdown.tsx
silviuaavram: I did not, to be honest, but I was suggested to use Portal at some point. I could not, as I had to conditionally render its content.
I had to start with a false condition var on state, then change it to true on componentHasMounted. But then I ended up with a <div> for each Dropdown on the page. When they got rendered, no Portal knew there are others rendering as well. And we only need one. So ... here we are.

docs/src/examples/components/Dropdown/Types/DropdownExampleMultipleSearch.shorthand.tsx
silviuaavram: removed TS.

src/components/Dropdown/DropdownItem.tsx
silviuaavram: in both those cases we did not have any 'active' values. Here we have values (items), active values (value) and highlighted values (highlighted). I feel that changing the highlight to active or selection will conflict, at a naming level, with the actual value(s) in the dropdown selection. What do you think?

src/components/Dropdown/DropdownItem.tsx
silviuaavram: We use this `image` in the shorthand to generate both this item and the label subcomponent. and in the label, `image` is the same as here. I agree it is a bit confusing, but maybe be consistent for now? Oh, and also Avatar is the same as well.

src/components/Dropdown/DropdownLabel.tsx
silviuaavram: refactored this completely.

src/components/Slot/Slot.tsx
silviuaavram: we debated this many times, and we agreed for now to have it as handled prop to avoid console warning. It is populated with the Ref (via the HOC you suggested) but it still needs to be handled for that warning reason.

src/components/Dropdown/DropdownItem.tsx
silviuaavram: thank you, addressed it!

src/components/Label/Label.tsx
```diff
@@ -35,7 +35,7 @@ export interface LabelProps
   /** An icon label can format an Icon to appear before or after the text in the label */
   iconPosition?: 'start' | 'end'
 
-  /** Label can have an icon. */
+  /** Label can have an image. */
```
mnajdova, line 38: 👍

src/components/Dropdown/DropdownItem.tsx
silviuaavram: will change to active inside the DropdownItem

src/components/Dropdown/DropdownItem.tsx
silviuaavram: addressed this to accept shorthand. closing.

src/components/Dropdown/Dropdown.tsx
silviuaavram: thank you!

src/components/Dropdown/DropdownItem.tsx
silviuaavram: refactored the styles. thanks!

src/components/Dropdown/DropdownItem.tsx
silviuaavram: Will add the first example as the one with only text items

src/components/Dropdown/Dropdown.tsx
silviuaavram: removed this now.

docs/src/examples/components/Dropdown/Types/DropdownExampleMultipleSearch.shorthand.tsx
bmdalex: no, you're right, you can leave as is but don't revert if you already made the change

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,624 @@
...
+
+  static defaultProps = {
+    as: 'div',
+  }
+
+  static autoControlledProps = ['searchQuery', 'value']
+
+  static Item = DropdownItem
+  static Label = DropdownLabel
+  static SearchInput = DropdownLabel
```
bmdalex: I think you mean `static SearchInput = DropdownSearchInput` ?

src/components/Dropdown/Dropdown.tsx
bmdalex: I'm reopening this so that others are aware that we're introducing this code. 

fyi @kuzhelov @mnajdova @miroslavstastny 

I don't think it's a good idea to have this approach but also don't want to block you.

@silviuavram Since it doesn't seem critical functionality, what if we remove this feature for now (`setA11yStatus`) and add it as a TODO to #560 until we find a better way to create this message?

src/components/Dropdown/DropdownItem.tsx
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  public renderComponent({ classes, rest }: RenderResultConfig<DropdownItemProps>) {
+    const { content, header, image, accessibilityItemProps } = this.props
+    return (
+      <ListItem
+        className={classes.root}
+        header={header}
+        onClick={this.handleClick}
+        {...{
```
bmdalex: this seems redundant:
```suggestion
media={Image.create(image, {
  defaultProps: {
    avatar: true,
  },
})}
```

src/components/Dropdown/DropdownLabel.tsx
```diff
@@ -0,0 +1,124 @@
...
+        content={header}
+        icon={Icon.create(icon || 'close', {
+          defaultProps: {
+            'aria-label': `Remove ${header} from selection.`, // TODO: Extract this in a behaviour.
+            'aria-hidden': false,
+            role: 'button',
+          },
+          overrideProps: this.handleIconOverrides,
+        })}
+        {...optionalImage}
```
bmdalex: the shorthand factory handles undefined arguments; just do:
```suggestion
image={Image.create(image, {
  defaultProps: {
    avatar: true,
  },
})
```

src/components/Dropdown/DropdownLabel.tsx
```diff
@@ -0,0 +1,124 @@
...
+    onRemove: PropTypes.func,
+  }
+
+  private handleClick = (e: React.SyntheticEvent) => {
+    _.invoke(this.props, 'onClick', e, this.props)
+  }
+
+  public renderComponent({ classes, styles, rest }: RenderResultConfig<DropdownLabelProps>) {
+    const { header, icon, image } = this.props
+    const optionalImage = !image
```
bmdalex: no need for this, the shorthand factory handles undefined arguments; see my suggestion here: https://github.com/stardust-ui/react/pull/422/files#r239556574

src/themes/teams/components/Dropdown/dropdownItemStyles.ts
```diff
@@ -0,0 +1,19 @@
+import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
+import { DropdownVariables } from './dropdownVariables'
+import { DropdownItemProps } from '../../../../components/Dropdown/DropdownItem'
+import ListItem from '../../../../components/List/ListItem'
+
+const dropdownItemStyles: ComponentSlotStylesInput<DropdownItemProps, DropdownVariables> = {
+  root: ({ variables: v, props: { active } }): ICSSInJSStyle => ({
+    [`&.${ListItem.className}`]: { backgroundColor: v.listItemBackgroundColor },
```
bmdalex, line 8: 👍

bmdalex:
any idea why this is happening in the screener tests?

https://screener.io/v2/states/rkGqtccHX.stardust-ui-react/feat%2Fmulti-dropdown-component/1024x768/Chrome/inputexampleinputslotshorthandtsx-0

docs/src/examples/components/Dropdown/Types/DropdownExampleMultipleSearch.shorthand.tsx
```diff
@@ -0,0 +1,60 @@
...
+  'Alfred Pennyworth',
+  `Scarlett O'Hara`,
+  'Imperator Furiosa',
+  'Bruce Banner',
+  'Peter Parker',
+  'Selina Kyle',
+]
+
+class DropdownExample extends React.Component {
+  state = {
```
kuzhelov: why `state` is necessary here?

docs/src/examples/components/Dropdown/Variations/DropdownExampleMultipleSearchFluid.shorthand.tsx
```diff
@@ -0,0 +1,61 @@
...
+  `Scarlett O'Hara`,
+  'Imperator Furiosa',
+  'Bruce Banner',
+  'Peter Parker',
+  'Selina Kyle',
+]
+
+class DropdownExample extends React.Component {
+  state = {
+    items: inputItems,
```
kuzhelov: same issue, `state` is not necessary to use here

kuzhelov: same applies to all examples, won't spam - but all should be fixed

src/components/Dropdown/DropdownItem.tsx
silviuaavram: refactored, thanks!

src/components/Dropdown/DropdownLabel.tsx
silviuaavram: refactored

kuzhelov (approved):

kuzhelov:
@Bugaa92 

> any idea why this is happening in the screener tests https://screener.io/v2/states/rkGqtccHX.stardust-ui-react/feat%2Fmulti-dropdown-component/1024x768/Chrome/inputexampleinputslotshorthandtsx-0

This happens because of the

PR #427, opened 2018-11-02
fix(Header): Adding dark and contrast styles for Header

bcalvery:
# Adding Dark and Contrast theme variables for Header

image
image
image


bcalvery:
@mnajdova, you worked on the Theme Switcher, could you possibly shed some light on this?
image

mnajdova:
@bcalvery when the theme is changed, it is changed for all page. That means, every stardust component we are using on the page, is going to change it's appearance. The headers here, seems to be implemented using the Stardust's Header components, so we will need on changing the theme, to maybe change the docs's background color (the same is already done for the examples (you can see that on dark theme, the background is dark). Another approach would be to wrap these two Header components with Provider and always provide the team's default theme, so it would be the last theme applied. Ping me if you need some help with this.

codepretty:
Looks good. Is there an issue creating to fix the theme switcher?

src/themes/teams/components/Header/headerVariables.ts
```diff
@@ -1,4 +1,11 @@
...
-  descriptionColor: undefined,
-})
+export interface HeaderVariables {
+  color: string
+  descriptionColor: string
+}
+
+export default (siteVars: any): HeaderVariables => {
+  return {
+    color: 'black',
```
codepretty: Should this be `siteVars.black`?

mnajdova:
@bcalvery I commited the changes required for the docs' Header components to always be black. Ping me if there is anything else.

image

mnajdova (approved):

Issue #426, opened 2018-11-02
Implement color prop for each component that accepts colors


alinais:
### Problem description
Some components need the color prop for better customizing. Use the color palette interface from siteVariables to achieve that.

### Proposed solution
- Implement `color` prop
- create color usage examples

for each component that accepts colors:

- [x] Divider #451 
- [x] Text #597
- [x] Header #628 
- [x] Segment #632 
- [x] Icon #651 
- [x] Label #647
- [ ] Menu
- [ ] Button
- [x] Status

alinais:
depends on #423

layershifter:
This issue is unblocked now because #423 was resolved 👍

levithomason:
Considering this done for now.  We can add Menu/Button colors on request.

Issue #421, opened 2018-11-02
Align `Menu` and `Divider` components to use the `type` property values as boolean


alinais:
### Problem description
Align `Menu` and `Divider` components to use the `type` property values as boolean. 

### Proposed solution
 Check the `Button` update, where  `primary` and `secondary` become boolean props while the `type` remains as a reserved prop.

[Fork, update, and replace this pen if you can show the proposed solution]:
https://codesandbox.io/s/2l3n74j9y

layershifter:
We have already PR for `Divider`: #327

layershifter:
Picking this.

levithomason:
#327 was closed since primary/secondary doesn't make sense in the context of a Divider.  Instead, they will have `color`s.  See #451 for the color palette and divider color work.

PR #436, opened 2018-11-06
[RFC] feat(Divider): add `color` prop

### LINKED ISSUE #423: Create siteVariables palette interface with color names
### Problem description
Each component should have the ability to change its color through the `color` prop. The color value should be available as part of a predefined palette.

### Proposed solution
Create siteVariables palette interface with color names. 

layershifter:
Fixes #423.
Refs #421, #426.

This PR adds the `color` prop for the `Divider` component.

### Blocked By
- [ ] #423.

### TODO

- [ ] Cover `getColor()` with tests
- [ ] Update the CHANGELOG.md
- [ ] Define correct colors in the palette

# API Proposal

### `ColorPalette`

I took the SUI palette as reference:

```ts
interface ColorPalette {
  primary: string
  secondary: string

  black: string
  brown: string
  blue: string
  green: string
  gray: string
  olive: string
  orange: string
  pink: string
  purple: string
  teal: string
  red: string
  violet: string
  white: string
  yellow: string
}
```

I have doubts about `primary` and `secondary` colors, the code below looks very confusing:

```jsx
<Button color='primary' primary />
```

### `ComponentColors`

A type that we will be used in props, defines all possible names of colors and includes `string`.

### `color`

The `color` prop accepts colors from the palette and custom colors.

image

```jsx
  <Divider color='black' content="Black" />
  <Divider color='#000' content="Black" /> // still black!
  <Divider color='rgb(192, 192, 192)' content="Gray" /> // and even RGB
```

However, I'm also not sure about custom colors. It plays well semantically, however we have variables for this:

image

```jsx
  <Divider color='pink' content="Pink" variables={{ colors: { pink: 'pink' }}} />
```


mnajdova:
I like the implementing, and vote against using the primary as secondary as a color, as those are abstraction for something more then a color, that's why we have them as a properties.

miroslavstastny (commented):
I think color palette (list of color names) should not be part of the API, but should be defined by a theme. Different themes should be able to define different list of colors.

If that's the case, colors cannot be enumerated in PropTypes, examples should read the list from current theme (see `TextSizesExampleShorthand` as an example).

Other opinions? @levithomason

layershifter:
> Different themes should be able to define different list of colors.

We can define only base colors (i.e. black, white, blue, gray, etc) and leave `colors` extendable, so end users will be able to define own colors.

layershifter:
Blocked by #423.

layershifter:
Will be done as part of #451 to confirm that palette ideas are working.

PR #438, opened 2018-11-06
fix(lodash): add types and fix errors

### LINKED ISSUE #420: Fix Lodash typing errors after enforcing @types/lodash
## Bug Report
Fix Lodash typing errors after enforcing @types/lodash inside yarn.lock

### Steps
Add @types/lodash in yarn.lock or install a package that requires @types/lodash, such as Cypress.
`yarn build`

### Expected Result
Should build without errors.

### Actual Result
Approximately 20 errors are thrown at build from the non type-enforced use of lodash library in the code.

bmdalex:
Fix for adding `lodash` types. Closes #420

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentProp/ComponentPropExtra.tsx
+++ b/docs/src/components/ComponentDoc/ComponentProp/ComponentPropExtra.tsx
@@ -1,5 +1,12 @@
 import PropTypes from 'prop-types'
 import * as React from 'react'
+import { Extendable } from 'types/utils'
+
+export interface ComponentPropExtraProps {
+  children?: JSX.Element[]
+  title?: React.ReactNode
+  inline?: boolean
+}
 
 const descriptionStyle = {
   color: '#666',
@@ -16,7 +23,12 @@ const contentInlineStyle = {
   display: 'inline',
 }
 
-const ComponentPropExtra: any = ({ children, inline, title, ...rest }) => (
+const ComponentPropExtra = ({
+  children,
+  inline,
+  title,
+  ...rest
+}: Extendable<ComponentPropExtraProps>) => (
   <div {...rest} style={descriptionStyle}>
     <strong>{title}</strong>
     <div style={inline ? contentInlineStyle : contentBlockStyle}>{children}</div>
--- a/docs/src/components/ComponentDoc/ComponentProp/ComponentPropExtra.tsx
+++ b/docs/src/components/ComponentDoc/ComponentProp/ComponentPropExtra.tsx
@@ -1,4 +1,3 @@
-import PropTypes from 'prop-types'
 import * as React from 'react'
 import { Extendable } from 'types/utils'
 
@@ -34,11 +33,5 @@ const ComponentPropExtra = ({
     <div style={inline ? contentInlineStyle : contentBlockStyle}>{children}</div>
   </div>
 )
-
-ComponentPropExtra.propTypes = {
-  children: PropTypes.node,
-  inline: PropTypes.bool,
-  title: PropTypes.node,
-}
 
 export default ComponentPropExtra
--- a/docs/src/components/ComponentDoc/ComponentProp/ComponentPropFunctionSignature.tsx
+++ b/docs/src/components/ComponentDoc/ComponentProp/ComponentPropFunctionSignature.tsx
@@ -3,7 +3,16 @@ import PropTypes from 'prop-types'
 import * as React from 'react'
 
 import { neverUpdate } from 'docs/src/hoc'
-import ComponentPropExtra from './ComponentPropExtra'
+import ComponentPropExtra, { ComponentPropExtraProps } from './ComponentPropExtra'
+
+interface ComponentPropFunctionSignature extends ComponentPropExtraProps {
+  name?: string
+  tags?: {
+    name?: string
+    description?: string
+    title?: string
+  }[]
+}
 
 const descriptionStyle = {
   flex: '5 5 0',
@@ -22,7 +31,7 @@ const rowStyle: any = {
 
 const getTagType = tag => (tag.type ? (tag.type.type === 'AllLiteral' ? 'any' : tag.type.name) : '')
 
-const ComponentPropFunctionSignature: any = ({ name, tags }) => {
+const ComponentPropFunctionSignature = ({ name, tags }: ComponentPropFunctionSignature) => {
   const params = _.filter(tags, { title: 'param' })
   const returns = _.find(tags, { title: 'returns' })
 
--- a/docs/src/components/ComponentDoc/ComponentProp/ComponentPropFunctionSignature.tsx
+++ b/docs/src/components/ComponentDoc/ComponentProp/ComponentPropFunctionSignature.tsx
@@ -5,7 +5,7 @@ import * as React from 'react'
 import { neverUpdate } from 'docs/src/hoc'
 import ComponentPropExtra, { ComponentPropExtraProps } from './ComponentPropExtra'
 
-interface ComponentPropFunctionArg extends ComponentPropExtraProps {
+interface ComponentPropFunctionProps extends ComponentPropExtraProps {
   name?: string
   tags?: {
     name?: string
@@ -31,7 +31,7 @@ const rowStyle: any = {
 
 const getTagType = tag => (tag.type ? (tag.type.type === 'AllLiteral' ? 'any' : tag.type.name) : '')
 
-const ComponentPropFunctionSignature = ({ name, tags }: ComponentPropFunctionArg) => {
+const ComponentPropFunctionSignature: React.SFC<ComponentPropFunctionProps> = ({ name, tags }) => {
   const params = _.filter(tags, { title: 'param' })
   const returns = _.find(tags, { title: 'returns' })
 
--- a/docs/src/components/ComponentDoc/ComponentProp/ComponentPropFunctionSignature.tsx
+++ b/docs/src/components/ComponentDoc/ComponentProp/ComponentPropFunctionSignature.tsx
@@ -5,7 +5,7 @@ import * as React from 'react'
 import { neverUpdate } from 'docs/src/hoc'
 import ComponentPropExtra, { ComponentPropExtraProps } from './ComponentPropExtra'
 
-interface ComponentPropFunctionSignature extends ComponentPropExtraProps {
+interface ComponentPropFunctionArg extends ComponentPropExtraProps {
   name?: string
   tags?: {
     name?: string
@@ -31,7 +31,7 @@ const rowStyle: any = {
 
 const getTagType = tag => (tag.type ? (tag.type.type === 'AllLiteral' ? 'any' : tag.type.name) : '')
 
-const ComponentPropFunctionSignature = ({ name, tags }: ComponentPropFunctionSignature) => {
+const ComponentPropFunctionSignature = ({ name, tags }: ComponentPropFunctionArg) => {
   const params = _.filter(tags, { title: 'param' })
   const returns = _.find(tags, { title: 'returns' })
 
--- a/docs/src/components/Sidebar/Sidebar.tsx
+++ b/docs/src/components/Sidebar/Sidebar.tsx
@@ -13,13 +13,14 @@ import { themes } from '@stardust-ui/react'
 import { ThemeContext } from '../../context/theme-context'
 import { constants } from 'src/lib'
 
+type ComponentMenuItem = { displayName: string; type: string }
+
 const pkg = require('../../../../package.json')
-const componentMenu = require('docs/src/componentMenu')
-const behaviorMenu = require('docs/src/behaviorMenu')
+const componentMenu: ComponentMenuItem[] = require('docs/src/componentMenu')
+const behaviorMenu: ComponentMenuItem[] = require('docs/src/behaviorMenu')
 
 const selectedItemLabelStyle: any = { color: '#35bdb2', float: 'right' }
 const selectedItemLabel = <span style={selectedItemLabelStyle}>Press Enter</span>
-type ComponentMenuItem = { displayName: string; type: string }
 
 class Sidebar extends React.Component<any, any> {
   static propTypes = {
@@ -51,7 +52,7 @@ class Sidebar extends React.Component<any, any> {
     this._searchInput = (findDOMNode(this) as any).querySelector('.ui.input input')
   }
 
-  handleDocumentKeyDown = e => {
+  private handleDocumentKeyDown = e => {
     const code = keyboardKey.getCode(e)
     const isAZ = code >= 65 && code <= 90
     const hasModifier = e.altKey || e.ctrlKey || e.metaKey
@@ -60,20 +61,20 @@ class Sidebar extends React.Component<any, any> {
     if (!hasModifier && isAZ && bodyHasFocus) this._searchInput.focus()
   }
 
-  handleItemClick = () => {
+  private handleItemClick = () => {
     const { query } = this.state
 
     if (query) this.setState({ query: '' })
     if (document.activeElement === this._searchInput) this._searchInput.blur()
   }
 
-  handleSearchChange = e =>
+  private handleSearchChange = e =>
     this.setState({
       selectedItemIndex: 0,
       query: e.target.value,
     })
 
-  handleSearchKeyDown = e => {
+  private handleSearchKeyDown = e => {
     const { history } = this.props
     const { selectedItemIndex } = this.state
     const code = keyboardKey.getCode(e)
@@ -101,9 +102,9 @@ class Sidebar extends React.Component<any, any> {
     }
   }
 
-  menuItemsByType = _.map(nextType => {
+  private menuItemsByType = _.map(nextType => {
     const items = _.flow(
-      _.filter(({ type }) => type === nextType),
+      _.filter<ComponentMenuItem>(({ type }) => type === nextType),
       _.map(info => (
         <Menu.Item
           key={info.displayName}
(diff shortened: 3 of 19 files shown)


docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -46,12 +47,12 @@ class Sidebar extends React.Component<any, any> {
     document.removeEventListener('keydown', this.handleDocumentKeyDown)
   }
 
-  setSearchInput() {
+  private setSearchInput() {
```
levithomason: Class methods are all private in the React paradigm.  I don't think we should be marking every method `private` on in every class.  It is unnecessary, especially for internal only code like docs.

src/lib/customPropTypes.tsx
```diff
@@ -189,7 +193,7 @@ export const some = validators => (props, propName, componentName, ...rest) => {
   }
 
   const errors = _.compact(
-    _.map(validators, validator => {
+    map(validators, validator => {
```
levithomason: Let's not import `map` alone being that we're already importing all of `lodash/fp` and this file only uses `lodash/fp`. We should update this call use `lodash/fp` again and swap the iterator with the data:

```jsx
_.map(validator => {}, validators)
```

One of the main differences with `fp` is that it follows the "data last" functional pattern, so the iterator function is the first argument and the data (array in this case) is the last argument.

src/lib/customPropTypes.tsx
```diff
@@ -202,7 +206,7 @@ export const some = validators => (props, propName, componentName, ...rest) => {
   // fail only if all validators failed
   if (errors.length === validators.length) {
     const error = new Error('One of these validators must pass:')
-    error.message += `\n${_.map(errors, (err, i) => `[${i + 1}]: ${err.message}`).join('\n')}`
+    error.message += `\n${map(errors, (err, i) => `[${i + 1}]: ${err.message}`).join('\n')}`
```
levithomason: Let's also retain `lodash/fp` here.  In this case, we need to drop the `i` (index) since `lodash/fp`'s map iterator is a unary function:

```diff
-map(errors, (err, i) => `[${i + 1}]: ${err.message}
+_.map(errors, err => `- ${err.message}`)
```

We don't need the fancy number in brackets ;)

src/lib/eventStack/EventTarget.tsx
```diff
@@ -1,9 +1,10 @@
 import * as _ from 'lodash'
+import { OneOrArray } from 'types/utils'
 
 export default class EventTarget {
-  _handlers = {}
-  _pools = {}
-  target: any
+  private _handlers: { [key: string]: EventListener } = {}
```
levithomason, line 5: Heads up, the EventStack will be ported from https://github.com/layershifter/event-stack.  Stardust changes are not likely to survive.  The other module is more up to date and includes bug fixes.

src/lib/mergeThemes.ts
```diff
@@ -64,9 +65,10 @@ export const mergeComponentVariables = (
     return (...args) => {
       const accumulatedVariables = acc(...args)
       const computedComponentVariables = callable(next)(...args)
+      const mergedVariables: ObjectOf<any> = {}
 
-      const mergedVariables = {}
-      _.mapKeys(computedComponentVariables, (variableToMerge, variableName) => {
+      Object.keys(computedComponentVariables).forEach(variableName => {
```
levithomason: `Object.keys` will throw `Uncaught TypeError: Cannot convert undefined or null to object` if called with `null` or `undefined`.  Suggest sticking with lodash methods as they gracefully handle these cases.

src/lib/mergeThemes.ts
```diff
@@ -64,9 +65,10 @@ export const mergeComponentVariables = (
...
+        const variableToMerge = computedComponentVariables[variableName]
```
levithomason: This assumes `computedComponentVariables` exists and is an object.  If it is undefined or null, this will also throw.  Suggest sticking with more lodash here to safe guard ourselves.

Note, `_.forEach` will iterate `(val, key) => {}` just like `_.mapKeys` but doesn't have a return value so it is arguably more appropriate in this case.  Essentially doing the same as `Object.keys().forEach()` except in a safe manner.

test/specs/components/Popup/Popup-test.tsx
```diff
@@ -1,6 +1,5 @@
-import _ from 'lodash'
-import computePopupPlacement from 'src/components/Popup/positioningHelper'
-import { Position, Alignment } from 'src/components/Popup/Popup'
```
levithomason, line 3: Crazy, these are not even valid exports 😕 I wonder how they were even compiling...

bmdalex: Crazier thing I just found out today that explains this:

**We are not including `test/` folder in our typescript compiler options so nothing under tests is compiled by TypeScript.**

I did some research and tracked it down to `build/gulp/tasks/dist.ts`. The `build` task relies on the `paths` object defined in `config.ts` and there is nothing related to `test`.

I made the necessary changes to enable typescript compilation for `test` folder and we have 50+ errors, among them:

> test/specs/components/Popup/Popup-test.tsx(3,10): error TS2305: Module '"/Users/alex/work/react/src/components/Popup/Popup"' has no exported member 'Position'.
test/specs/components/Popup/Popup-test.tsx(3,20): error TS2305: Module '"/Users/alex/work/react/src/components/Popup/Popup"' has no exported member 'Alignment'.

**I extracted this to a separate RFC: #445**

docs/src/components/Sidebar/Sidebar.tsx
bmdalex: > Class methods are all private in the React paradigm

Agreed, but that's the JavaScript paradigm, where anything is publicly accessible anyway. But then comes the TypeScript story / paradigm / however we want to call it.

Since custom React components extend `React.Component`, TypeScript expects component lifecycle classes to actually be public even though they are private from React perspective since user cannot call them):
```
componentDidMount
componentWillUnmount
shouldComponentUpdate
// etc..
```
If we configure TypeScript compiler to make class methods private by default we will need to specify `public` keyword in front of every lifecycle method we define (even though they are private from React perspective).

> I don't think we should be marking every method private on in every class.

Why not? It's the correct thing to do from TypeScript perspective and it will help a lot with detecting dead code (unused methods). I know it's not a valid use-case, but these methods appear as public in case the user gets a reference to a component instance (maybe possible when working with refs?)
screenshot 2018-11-08 at 17 00 15

src/lib/customPropTypes.tsx
bmdalex: done, thanks

src/lib/customPropTypes.tsx
bmdalex: done, thanks

src/lib/mergeThemes.ts
bmdalex: done, thanks

src/lib/mergeThemes.ts
bmdalex: done, thanks

src/lib/eventStack/EventTarget.tsx
bmdalex: I'm aware, thanks, still needed to fix type errors

docs/src/examples/components/Text/Types/TextSizesExample.tsx
```diff
@@ -7,7 +7,7 @@ const TextSizesExample = () => (
     render={({ siteVariables }) => {
       return _.map(siteVariables.fontSizes, (value, key) => (
         <div key={key}>
-          <Text size={key}>This is size="{key}" size font.</Text>
+          <Text size={key as any}>This is size="{key}" size font.</Text>
```
silviuaavram, line 10: creating task for having a unified Interface for both the `siteVariables.fontSizes` and their use in the components, which is now `'small' | 'big' ...`

silviuaavram (approved):

docs/src/components/ComponentDoc/ComponentProp/ComponentPropFunctionSignature.tsx
```diff
@@ -22,7 +31,7 @@ const rowStyle: any = {
 
 const getTagType = tag => (tag.type.type === 'AllLiteral' ? 'any' : tag.type.name)
 
-const ComponentPropFunctionSignature: any = ({ name, tags }) => {
+const ComponentPropFunctionSignature = ({ name, tags }: ComponentPropFunctionSignature) => {
```
kuzhelov: maybe `ComponentPropFunctionArg(s)` would be better?

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -13,13 +13,14 @@ import { themes } from '@stardust-ui/react'
 import { ThemeContext } from '../../context/theme-context'
 import { constants } from 'src/lib'
 
+type ComponentMenuItem = { displayName: string; type: string }
+
 const pkg = require('../../../../package.json')
-const componentMenu = require('docs/src/componentMenu')
-const behaviorMenu = require('docs/src/behaviorMenu')
+const componentMenu: ComponentMenuItem[] = require('docs/src/componentMenu')
+const behaviorMenu: ComponentMenuItem[] = require('docs/src/behaviorMenu')
```
layershifter, line 20: May be we move `ComponentMenuItem` to `global.d.ts`?

```ts
const MenuItem = { displayName: string; type: string }

declare module 'componentMenu.json' {
  const value: ComponentMenuItem[]

  export default value
}
```

This will also allow us use `import` instead of `require`, I think.

docs/src/prototypes/chatPane/services/dataMock.ts
```diff
@@ -69,7 +69,7 @@ class ChatMock {
       }
 
       return message
-    }).sort((a, b) => a.date - b.date)
+    }).sort((a, b) => a.date.getMilliseconds() - b.date.getMilliseconds())
```
kuzhelov: this might introduce unexpected sort result, as you are now comparing just ms part of the dates collection (so that 17:00:0.456 > 17:01:0.123, as 456 > 123). Is it intentional?

src/lib/mergeThemes.ts
```diff
@@ -64,9 +65,9 @@ export const mergeComponentVariables = (
     return (...args) => {
       const accumulatedVariables = acc(...args)
       const computedComponentVariables = callable(next)(...args)
+      const mergedVariables: ObjectOf<any> = {}
 
-      const mergedVariables = {}
-      _.mapKeys(computedComponentVariables, (variableToMerge, variableName) => {
+      _.forEach(computedComponentVariables, (variableToMerge, variableName) => {
```
kuzhelov, line 71: 👍

src/lib/customPropTypes.tsx
```diff
@@ -365,7 +377,12 @@ export const collectionShorthand = every([disallow(['children']), PropTypes.arra
  * @param {string} help A help message to display with the deprecation warning.
  * @param {function} [validator] A propType function.
  */
-export const deprecate = (help, validator) => (props, propName, componentName, ...args) => {
+export const deprecate = (help: string, validator: Function) => (
+  props,
```
kuzhelov: types for these args are not introduced - seems that they should be like the ones that are used for the previous expression. The same applies to the `demand` function. Just to be sure, is it intentional?

src/lib/customPropTypes.tsx
```diff
@@ -28,15 +31,15 @@ export const domNode = (props, propName) => {
...
   if (!Array.isArray(suggestions)) {
     throw new Error('Invalid argument supplied to suggest, expected an instance of array.')
   }
 
-  const findBestSuggestions = _.memoize(str => {
+  const findBestSuggestions = _.memoize((str: string) => {
     const propValueWords = str.split(' ')
 
-    return _.flow(
+    return _.flow<string[], SuggestProps[], SuggestProps[], SuggestProps[]>(
```
kuzhelov: the same type correctness effect could be achieved a bit easier - it is possible to omit type specs for `flow<...>`, and introduce type for the `map` function only:

```ts
...
 _.map((suggestion: string) => {
...
```

The benefit is that we won't need to introduce types to `flow` for each new function added. This would be the result code:

```ts
_.flow(
      _.map((suggestion: string) => {
        const suggestionWords = suggestion.split(' ')

        const propValueScore = _.flow(
          _.map(x => _.map(y => leven(x, y), suggestionWords)),
          _.map(_.min),
          _.sum,
        )(propValueWords)

        const suggestionScore = _.flow(
          _.map(x => _.map(y => leven(x, y), propValueWords)),
          _.map(_.min),
          _.sum,
        )(suggestionWords)

        return { suggestion, score: propValueScore + suggestionScore }
      }),
      _.sortBy<SuggestProps>(['score', 'suggestion']),
      _.take(3),
    )(suggestions)
```

kuzhelov (commented):
have left several minor comments - overall looks very good to me 👍 Please, consider to review changes made for the sorting procedure for `dataMock` - have some doubts there.

One thing that really raises concerns is that we have to use `key as ..` for docs examples - while the goal is to avoid any TS artifacts from these snippets. At the same time, should emphasize that this is definitely not the problem of this PR, but rather is a broader issue.

Thanks!

docs/src/components/Sidebar/Sidebar.tsx
bmdalex: not sure we want `ComponentMenuItem` in global scope, it's not that relevant;

also, declaring the module like that would require lot more changes to fix the type errors that come out of it:
screenshot 2018-11-12 at 13 38 43

since this is not in the scope of the PR anyways, I decided not to fix, hope you're ok with that @layershifter

docs/src/components/ComponentDoc/ComponentProp/ComponentPropFunctionSignature.tsx
bmdalex: done, thanks

docs/src/prototypes/chatPane/services/dataMock.ts
bmdalex: meant to compare the dates, used `getTime` instead

src/lib/customPropTypes.tsx
bmdalex: I was just fixing the types needed for `lodash` to compile correctly, I added types where I could

src/lib/customPropTypes.tsx
bmdalex: very nice, thanks! applied your suggestion

bmdalex:
@levithomason @layershifter @kuzhelov 
ready again, can you take another look?

src/lib/customPropTypes.tsx
kuzhelov: now everything is fine and consistent, thanks!

kuzhelov (approved):

docs/src/components/ComponentDoc/ComponentProp/ComponentPropExtra.tsx
```diff
@@ -1,5 +1,12 @@
 import PropTypes from 'prop-types'
 import * as React from 'react'
+import { Extendable } from 'types/utils'
+
+export interface ComponentPropExtraProps {
+  children?: JSX.Element[]
+  title?: React.ReactNode
+  inline?: boolean
+}
```
layershifter, line 8: Now we can safely remove `propTypes` in this file.

docs/src/components/ComponentDoc/ComponentProp/ComponentPropFunctionSignature.tsx
```diff
@@ -22,7 +31,7 @@ const rowStyle: any = {
 
 const getTagType = tag => (tag.type.type === 'AllLiteral' ? 'any' : tag.type.name)
 
-const ComponentPropFunctionSignature: any = ({ name, tags }) => {
+const ComponentPropFunctionSignature = ({ name, tags }: ComponentPropFunctionArg) => {
```
layershifter: `ComponentPropFunctionSignature ` is `React.SFC`:

```jsx
const ComponentPropFunctionSignature: React.SFC<ComponentPropFunctionProps> = ({ name, tags }) => {
```

docs/src/components/ComponentDoc/ComponentProp/ComponentPropFunctionSignature.tsx
```diff
@@ -3,7 +3,16 @@ import PropTypes from 'prop-types'
 import * as React from 'react'
 
 import { neverUpdate } from 'docs/src/hoc'
-import ComponentPropExtra from './ComponentPropExtra'
+import ComponentPropExtra, { ComponentPropExtraProps } from './ComponentPropExtra'
+
+interface ComponentPropFunctionArg extends ComponentPropExtraProps {
```
layershifter: This definitely should be `ComponentPropFunctionProps`

docs/src/components/ComponentDoc/ComponentProp/ComponentPropExtra.tsx
bmdalex: done, thanks

docs/src/components/ComponentDoc/ComponentProp/ComponentPropFunctionSignature.tsx
bmdalex: done, thanks

docs/src/components/ComponentDoc/ComponentProp/ComponentPropFunctionSignature.tsx
bmdalex: done, thanks

docs/src/components/Sidebar/Sidebar.tsx
bmdalex: this is pending for a while; we can revisit later if necessary

PR #441, opened 2018-11-06
feat(Icon): Redlining icons

codepretty:
### Icon style updates for Teams theme
The style updates for Teams icons is mostly around the size of the icons. The normal (default) Icon size matches the Teams spec of being a 16px x 16px space. Other size variations of Icons will be addressed in a future PR.

**Before**
image

**After**
This shows the size update for both SVGs and Icon fonts.
image

### Filled and Outline versions of Teams Icon
Part of this PR includes updates to show either a _filled_ or _outline_ version of the icon.
image

**API update**
```
<Icon name="call-video" />
<Icon name="call-video" variables={{ outline: true }} />
```

### Changing viewBox for Teams SVGs
We're changing the viewBox of all (or most?) Teams icons. Default icons are built with built-in padding, a viewBox of 32x32px and a 16x16px icon size. We changed the viewBox to remove the bounding box and just have the icon, which remains the same 16x16px size.

The reason for this is because if you want an icon to appear to not have any space you would need to do some negative margin hacks.

**Before**
image

**Now**
image

---

**Question:** Are the icons examples too tied to Teams theme? In a default theme, what would the expectation be if there is no icon provider shown? Today, it fails silently by not showing anything which feels right to me, but wanted to get others opinions..


miroslavstastny:
Some icons do not support outline/filled and are not visible, some others ignore the variable:
image
I really like the changes here, once merged, I will update #364 to match the changes.

It would be great if @kuzhelov could review the changes as well.

docs/src/examples/components/Icon/Variations/IconExampleSpace.shorthand.tsx
```diff
@@ -5,27 +5,27 @@ const IconExampleSpace = () => (
   <div>
     <div style={{ marginBottom: '1.2rem' }}>
       <Label content="Default" />
-      <Icon name="umbrella" size="big" />
+      <Icon name="call-video" />
       <Label content="Default" />
     </div>
     <div style={{ marginBottom: '1.2rem' }}>
       <Label content="Before" />
-      <Icon name="umbrella" xSpacing="before" size="big" />
```
kuzhelov, line 13: 👍

docs/src/examples/components/Popup/Types/PopupCustomTargetExample.shorthand.tsx
```diff
@@ -16,7 +16,11 @@ class PopupExample extends React.Component {
         {/* CUSTOM DOM ELEMENT is used as target for Popup */}
         <Popup
           target={this.state.popupTarget}
-          trigger={<Icon name="question" circular bordered styles={{ cursor: 'pointer' }} />}
+          trigger={
+            <div>
```
kuzhelov: just to be sure - could you, please, suggest why `<div>` is necessary here? Is it because of some layout concerns?

src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -4,38 +4,43 @@ import { fittedStyle } from '../../../../styles/customCSS'
...
     fontFamily,
-    textAlign: 'center',
+    fontSize: getSize(size),
     lineHeight: 1,
-    width: '1.18em',
+    textAlign: 'center',
 
     '::before': {
       content,
+      display: 'block',
```
kuzhelov, line 41: 👍

src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -53,21 +58,23 @@ const getXSpacingStyles = (xSpacing: IconXSpacing, horizontalSpace: string): ICS
   }
 }
 
-const getBorderedStyles = (isFontBased, circular, borderColor, color): ICSSInJSStyle => {
+const getBorderedStyles = (circular, borderColor, color, boxShadowFallbackColor): ICSSInJSStyle => {
```
kuzhelov: seems that both `borderColor`, `color` and `boxShadowFallbackColor` are the arguments that should provide `boxShadow` color to the calculated styles object - this is the only responsibility for each of them. Thus, I would suggest to simplify a contract of this function:

```js
const getBorderedStyles = (circular,  boxShadowColor) => {
  ... 
  return {
    ...
    boxShadow: `... ${boxShadowColor}`
  }
}
```

and do all the necessary reductions on the client side:

```js
getBorderedStyles(
  circular, 
  v.borderColor || v.color || theme.siteVariables.bodyColor
)
```

This will make code more expressive, as well as a contract for this function to be much cleaner (and we won't need to introduce additional parameters if another fallback value will come into play :).

docs/src/examples/components/Icon/Usage/IconSetExample.shorthand.tsx
```diff
@@ -1,16 +1,27 @@
...
 
+const cellStyles = {
+  margin: '10px 0',
+}
+
 const IconSetExampleShorthand = () => (
   <Provider.Consumer
     render={theme => (
-      <Grid columns={3} style={{ textAlign: 'center' }}>
+      <Grid columns={4} style={{ textAlign: 'center' }}>
```
kuzhelov: nit: just for the sake of introducing better separation of these groups, maybe the following markup could be considered (feel free to judge by yourself :) 

```jsx
const IconSetExampleShorthand = () => (
  <Provider.Consumer
    render={theme => (
      <>
        <div>
          <Divider>
            <Header as="h3" content="Regular"/>
          </Divider>
          <Grid columns={4} style={{ textAlign: 'center' }}>
            {Object.keys(theme.icons).map(name => (
              <div key={name} style={cellStyles}>
                <Icon name={name} />
                <br />
                <code>{name}</code>
              </div>
            ))}
          </Grid>
        </div>
        
        <div>
          <Divider>
            <Header as="h3" content="Outline"/>
          </Divider>
          <Grid columns={4} style={{ textAlign: 'center' }}>
            {Object.keys(theme.icons).map(name => (
              <div key={name + '-outline'} style={cellStyles}>
                <Icon name={name} variables={{ outline: true }} />
                <br />
                <code>{name} outline</code>
              </div>
            ))}
          </Grid>
        </div>
      </>
    )}
  />
)
```

image

src/themes/teams/components/Icon/svg/icons/call.tsx
```diff
@@ -21,11 +23,5 @@ export default {
...
     </svg>
   ),
-  styles: {
-    svg: ({ variables }) => {
-      return {
-        fill: variables.color,
-      }
-    },
-  },
+  styles: {},
```
kuzhelov: if there are no icon-specific styles (the ones that could be applied to `call` icon exclusively), this prop could be just omitted

kuzhelov, line 26: note to myself - if there are no icon-specific styles (like the ones that will be applied to `call` icon exclusively), this `styles` prop should be safe to delete from this object

kuzhelov (approved):
do like the approach taken - have left several minor comments, please, feel free to decide on whether they should be addressed 👍Please, just make CHANGELOG entry before merging these changes.

Also have made a not to myself in regards to `styles` prop of the icon with empty object value.

kuzhelov:
@codepretty 

>Question: Are the icons examples too tied to Teams theme? In a default theme, what would the expectation be if there is no icon provider shown? Today, it fails silently by not showing anything which feels right to me, but wanted to get others opinions..

The Icon examples are indeed to tied to Teams theme now, but see this as a normal thing for now at least, and speaking of the last section on the Icon's docs page (where all the icons are shown) - this close tie is inevitable there. In either case, it seems that we are in a pretty good position to address this "close tie" problem for icons if necessary - as these changes don't make the problem any worse.

Speaking of the behavior in case if icon is absent in theme's collection - I would expect that there will be a warning/error message logged to the dev console, so this fact won't be unnoticed; speaking of rendering process - yes, absolutely agree that this fact shouldn't result in exceptions being thrown: whole components' tree should be just rendered without this missing icon.

docs/src/examples/components/Popup/Types/PopupCustomTargetExample.shorthand.tsx
codepretty: Yes, it's layout. I made this change because of a failing screen reader test.

src/themes/teams/components/Icon/iconStyles.ts
codepretty: Makes sense. I will make this update.

docs/src/examples/components/Icon/Usage/IconSetExample.shorthand.tsx
codepretty: Yes, I think this works better for now. Perhaps in the future would be good to show filled/outline versions side-by-side, but i suspect when enough icons go in we'll want to display in a different way anyways.

docs/src/examples/components/Popup/Types/PopupCustomTargetExample.shorthand.tsx
codepretty: However, after pulling in the latest looks like the trigger was changed to a button which doesn't require a wrapper div

kohlikohl:
@codepretty There are a bunch of additional icons that we would need support for. Would they be added as part of this PR?

**Icons needed**
* Like
* send message
* bookmark
* trash
* redo
* pencil (edit)
* mark as unread
* translate
* immersive reader
* add participants
* alert (triangle with exclamation mark)
* add participant (add one participant)
* remove participant
* @ for at-mention
* exclamation mark for important messages
* add to calendar
* member joined (control message in conversation)
* member left (control message in conversation)

codepretty:
@kohlikohl importing icons is a separate PR. It is almost ready, just want to get this checked in first.

PR #442, opened 2018-11-07
feat(Input): add iconPosition property

mnajdova:
This PR aims to add iconPosition property on the Input component.

# API

## iconPosition: 'start' | 'end'
new prop for positioning the icon consistent with the other components. By default it's value is 'end'.

# Usage
## Example code
```
<Input icon="search" placeholder="Search..." iconPosition="start" />
```
## Rendered component
image

## Rendered HTML
```html
<div class="ui-slot ui-input">
  <input type="text" placeholder="Search..." value="" class="ui-slot" />
</div>
```


silviuaavram (approved):

docs/src/examples/components/Input/Variations/index.tsx
```diff
@@ -9,6 +9,11 @@ const Variations = () => (
       description="An input can have an icon."
       examplePath="components/Input/Variations/InputExampleIcon"
     />
+    <ComponentExample
+      title="Icon position"
+      description="The icon inside the input can be positioned in the start of the input."
```
silviuaavram: *at the start

src/themes/teams/components/Input/inputVariables.ts
```diff
@@ -7,6 +8,9 @@ export interface InputVariables {
   fontSize: string
   iconPosition: string
   iconRight: string
+  iconLeft: string
+  inputPaddingWithIconOnStart: string
```
kuzhelov: nit: 'at' would be a better fit here:  `inputPaddingWithIconAtStart`

src/themes/teams/components/Input/inputStyles.ts
```diff
@@ -26,11 +26,23 @@ const inputStyles: ComponentSlotStylesInput<InputProps, InputVariables> = {
       borderColor: v.inputFocusBorderColor,
       borderRadius: v.inputFocusBorderRadius,
     },
+    ...(p.iconPosition === 'start' &&
```
kuzhelov: nit: we could simplify and decouple conditions in the following way:

```js
...
...(p.clearable && { padding: v.inputPaddingWithIconAtEnd, }),
...(p.icon && { padding: p.iconPosition === 'start'
  ? v.inputPaddingWithIconAtStart
  : v.inputPaddingWithIconAtEnd })
```

Most important benefit of this approach would be decoupling the effects of `clearable` and `iconPosition` props for styles, so that if, say, one of these props will be eliminated in future, it would be just a matter of removing the corresponding paragraph

kuzhelov (approved):

src/themes/teams/components/Input/inputStyles.ts
mnajdova: Rearranged styles and changed variables names. Thanks!

PR #443, opened 2018-11-07
feat(Chat): Set default chatBehavior which uses Enter/Esc keys

sophieH29:
Currently, we have implemented 2 chat behaviors:
- ```chatBehavior```-  uses ```ArrowRight``` and ```ArrowLeft``` to go in and out of the message
- ```chatEnterEscBehavior``` - uses ```Enter``` and ```Esc``` to go in and out of the message 

After a couple of discussions, we agreed that default behavior should be the one that uses ```Enter``` and ```Esc```.
Chat behavior which uses ```ArrowRight``` and ```ArrowLeft``` currently not needed, so it was removed from the repo.


jurokapsiar (approved):

miroslavstastny (approved):
Don't forget to update changelog

PR #450, opened 2018-11-08
fix(FocusZone): Fix "defaultTabbableElement" prop

sophieH29:
```defaultTabbableElement``` previously was defined in ```FocusZone``` as a string selector which brought a couple of the issues.

Structure of the dom is changing and can be complex - it's quite hard to select the exact element with one string selector. Often it's needed to make some dom manipulations.

```defaultTabbableElement``` API was changed to function

```js
  defaultTabbableElement?: (root: HTMLElement) => HTMLElement
```

```FocusZone``` calls this method to get an initial active element and passing itself as a root element, so the dom quering is constrained by the FocusZone area.

### CODE REVIEW

File: src/lib/accessibility/Behaviors/Chat/chatBehavior.ts [hunk 1]
```diff
@@ -34,4 +34,13 @@ const ChatBehavior: Accessibility = (props: any) => ({
   },
 })
 
+const getLastTabbableElement = (root: HTMLElement): HTMLElement => {
+  const chatItemsElements = root.querySelectorAll(
+    `[${CHAT_FOCUSZONE_ATTRIBUTE}] .ui-chat__item > [${IS_FOCUSABLE_ATTRIBUTE}]`,
+  )
+  return (
```
kuzhelov: note that this will return `false` in case if condition is not satisfied. For the sake of preserving type of return value, it is better to explicitly return `null` instead:

```js
(chatItemsElements.length > 0 
  ? chatItemsElements[chatItemsElements.length - 1]
  : null)  as HTMLElement
```

sophieH29: agree, thank you

src/lib/accessibility/Behaviors/Chat/chatBehavior.ts
```diff
@@ -34,4 +34,13 @@ const ChatBehavior: Accessibility = (props: any) => ({
   },
 })
 
+const getLastTabbableElement = (root: HTMLElement): HTMLElement => {
+  const chatItemsElements = root.querySelectorAll(
+    `[${CHAT_FOCUSZONE_ATTRIBUTE}] .ui-chat__item > [${IS_FOCUSABLE_ATTRIBUTE}]`,
+  )
+  return (chatItemsElements.length > 0
```
silviuaavram: maybe 
`return chatItemsElements.length > 0 ? (chatItemsElements[chatItemsElements.length - 1] as HTMLElement) : null`

silviuaavram (approved):

src/lib/accessibility/Behaviors/Chat/chatBehavior.ts
sophieH29: yeah-yeah, fixed together with the above comment. Thank you

jurokapsiar (approved):

src/lib/accessibility/Behaviors/Chat/chatBehavior.ts
kuzhelov: yes, this would be better - the reason I've suggested to enclose everything in braces was to guarantee to preserve type for the evaluated expression. This moment will become critical in case if `strictNullCheck` flag will be enabled for TS compiler. At the same time, this type is explicitly defined by function return type, and, thus, we don't need to brace everything, thanks!

kuzhelov (approved):

PR #451, opened 2018-11-08
feat(Theming): add color palette

### LINKED ISSUE #423: Create siteVariables palette interface with color names

layershifter:
Fixes #423.


layershifter:
This PR will be never merged. Using it to prove some ideas.

docs/src/views/ColorPalette.tsx
```diff
@@ -0,0 +1,169 @@
...
+import * as React from 'react'
+import * as color from 'tinycolor2'
+
+import Slot from 'src/components/Slot/Slot'
+import {
+  baseColors,
+  ColorPalette,
+  naturalPalettes,
+  statePalettes,
+  StatePalettes,
```
hughreeling: Should palettes be plural - or like ColorPalette

layershifter: Sure. However, I don't think that will call it `Palette` at all, too many palettes there. Naming problem :)

docs/src/views/ColorPalette.tsx
```diff
@@ -0,0 +1,92 @@
...
+              <ColorBox key={color} name={color} rounded size="big" value={colors[color][500]} />
+            ))}
+          </Grid>
+
+          <Header as="h2">Text colors</Header>
+          <p>TODO: Describe</p>
+
+          <Segment>
+            {_.map(colors.text, color => (
+              <Text as="p" key={color} style={{ color }}>
```
miroslavstastny: `Text` has `display: inline-block` - that results in multiple colors displayed on a single line which can be difficult to understand.
image

layershifter: I changed `Text` to a simple `p` 👍

docs/src/views/ColorPalette.tsx
```diff
@@ -0,0 +1,113 @@
...
+      render={({ siteVariables: { colors, contextualColors, emphasisColors, naturalColors } }) => (
+        <DocPage title="Color Palette">
+          <Header as="h2">Introduction</Header>
+          <p>
+            The color palette for a theme has many requirements and constraints. There is a need to
+            be intentional and functional with color use. We analyzed existing frameworks and picked
+            the best ideas from them.
+          </p>
+          <p>
+            An each theme should match our color palette types fully. This will allow you to use our
```
hughreeling: An each?  Remove An?

docs/src/views/ColorPalette.tsx
```diff
@@ -0,0 +1,113 @@
...
+            theming features completely and keep your palette structured.
+          </p>
+
+          <Header as="h2">Primitive colors</Header>
+          <p>
+            This part of the palette includes only <i>black</i> and <i>white</i> colors, we decided
+            to separate by semantical ideas. There are nothing more blacker than black and nothing
```
hughreeling: There is nothing blacker than black and nothing whiter than white

docs/src/views/ColorPalette.tsx
```diff
@@ -0,0 +1,113 @@
...
+              <Grid.Column key={color}>
+                <ColorBox name={color} rounded size="big" value={colors[color]} />
+              </Grid.Column>
+            ))}
+          </Grid>
+
+          <Header as="h2">Natural colors</Header>
+          <p>
+            This part of palette includes nine colors that are the most frequently used among
+            popular frameworks. Each color includes ten gradients, this allows to satisfy all common
```
hughreeling: allows us to satisfy most common

docs/src/views/ColorPalette.tsx
```diff
@@ -0,0 +1,113 @@
...
+            needs. This decisition expirienced from Material UI and allows to define more variants
```
hughreeling: this differs from

hughreeling (approved):

docs/src/views/ColorPalette.tsx
layershifter: Fixed to:
> This decision is experienced from Material UI

docs/src/views/ColorPalette.tsx
layershifter: 😭

CHANGELOG.md
```diff
@@ -25,6 +29,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add all default size Teams icons processed & ready to be consumed by Stardust as needed @codepretty ([#478](https://github.com/stardust-ui/react/pull/478))
 - Add `Tree` Component @priyankar205 ([#479]
 (https://github.com/stardust-ui/react/pull/479))
+- Add `color` prop to `Divider` component @layershifter ([#451](https://github.com/stardust-ui/react/pull/451))
```
kuzhelov: 👍

docs/src/components/CopyToClipboard.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  componentWillUnmount() {
+    clearTimeout(this.timeoutId)
+  }
+
+  handleClick = () => {
+    const { timeout, value } = this.props
+
+    this.setState({ active: true })
+    this.timeoutId = setTimeout(() => {
```
kuzhelov, line 35: note that race conditions are introduced here, as previous timeout callback might not be fired when new is activated (i.e. `clearTimeout` is not called for previous value of `timeoutId`)

docs/src/views/ColorPalette.tsx
```diff
@@ -0,0 +1,113 @@
...
+            {_.map(naturalColors, (variants, color) => (
+              <Grid.Column key={color}>
+                <ColorBox name={color} rounded value={colors[color][500]} />
+              </Grid.Column>
+            ))}
+          </Grid>
+
+          <Header as="h2">Contextual colors</Header>
+          <p>
+            This part of the palette includes primary and secondary (i.e. brand) colors and
```
kuzhelov: would rather see `primary` and `secondary` colors being provided in a separate category (`Brand`, `Base` colors) - as those have completely different semantics: those, in contrast to `Contextual` ones, serve as palette pillars, base colors that largely influence and define the rest of color palette. Many `Contextual` ones, at the same time, are largely influenced by these base colors as well

docs/src/views/ColorPalette.tsx
```diff
@@ -0,0 +1,113 @@
...
+            {_.map(['black', 'white'], color => (
+              <Grid.Column key={color}>
+                <ColorBox name={color} rounded size="big" value={colors[color]} />
+              </Grid.Column>
+            ))}
+          </Grid>
+
+          <Header as="h2">Natural colors</Header>
+          <p>
+            This part of palette includes nine colors that are the most frequently used among
```
kuzhelov: don't see nine of them (just six are provided in this section, which ones are implied as the rest three?)

docs/src/views/ColorPalette.tsx
```diff
@@ -0,0 +1,113 @@
...
+            {_.map({ ...emphasisColors, ...contextualColors }, (variants, color) => (
+              <Grid.Column key={color}>
+                <ColorBox name={color} rounded size="big" value={colors[color][500]} />
+              </Grid.Column>
+            ))}
+          </Grid>
+
+          <Header as="h2">Text colors</Header>
+          <p>
+            Text variants are also separated as state color because in the most cases it's not
```
kuzhelov: I am reading this as a client, and have a reasonable question: text color is largely influenced by the backgeround where text is used, but thjis paragraph suggests that there is only single text color suggested (yes, maybe with several tints, but still just a single color) - and this contradicts most approaches to color palette where text color is largely defined by the background - and, thus, `on` variants for text color (generally - content color, such as text, icons, etc) are suggested: `onPrimary`, `onSecondary`, etc. More than that, depending on the tint of the background different text color might be used, so that content will appear with enough visual contrast.

What is our suggestion here?

kuzhelov: also, might suggest to rename it to something that is more generic than `text`, as this color generally applies not only to text content

docs/src/components/CopyToClipboard.tsx
layershifter: Fixed 👍

kuzhelov:
one thing that I still missing, even after reading the guide, is what are the steps I should follow as a client to introduce color palette to my theme? What are the tools that will aid me in that? If it is not the purpose of this PR - please, let me know what are our steps planned to address that?

docs/src/views/ColorPalette.tsx
layershifter: Agree, they are placed even in the separate type now, will move them to other section.

docs/src/views/ColorPalette.tsx
```diff
@@ -0,0 +1,113 @@
...
+            correct to use grey color for text.
+          </p>
+
+          <Segment>
+            {_.map(colors.text, color => (
+              <p key={color} style={{ color }}>
```
kuzhelov: note, first paragraph is almost unreadable because of the low visual contrast. As a client I would rather not expect to see this example on the 'guide' page where examples of suggested practices are provided. Just thinking if we could present it better somehow

image

docs/src/views/ColorPalette.tsx
layershifter: We decided to go with optional colors, so yes, Teams theme contains six colors while default contains all nine:

image

docs/src/views/ColorPalette.tsx
layershifter: I reworked Text block, now it looks much readable:

image

docs/src/views/ColorPalette.tsx
layershifter: image

Now they are separated 👍

docs/src/views/ColorPalette.tsx
kuzhelov: then, just as a suggestion, we might include note in text like `This part of palette includes nine colors (note, this number might be different for non-default theme) ...`

docs/src/views/ColorPalette.tsx
```diff
@@ -0,0 +1,122 @@
...
+          <Grid columns={2}>
+            {_.map(naturalColors, (variants, color) => (
+              <Grid.Column key={color}>
+                <ColorBox name={color} rounded value={colors[color][500]} />
+              </Grid.Column>
+            ))}
+          </Grid>
+
+          <Header as="h2">Emphasis colors</Header>
+          <p>This part of the palette includes primary and secondary (i.e. brand) colors.</p>
```
kuzhelov: so, emphasis has the same semantics as brand ones? As I might recall, we've decided that semantics of these two groups should be different - while, at the same time, agree that it is better to reduce complexity (and, thus, number of categories introduced) at this stage

docs/src/views/ColorPalette.tsx
```diff
@@ -0,0 +1,122 @@
...
+            {_.map(contextualColors, (variants, color) => (
+              <Grid.Column key={color}>
+                <ColorBox name={color} rounded size="big" value={colors[color][500]} />
+              </Grid.Column>
+            ))}
+          </Grid>
+
+          <Header as="h2">Text colors</Header>
+          <p>
+            Text variants are also separated as state color because in the most cases it's not
```
kuzhelov: > are also separated as state color

could you, please, suggest what is the meaning intended to be provided by that? Is it about something like:

> Text variants are also provided as a separate group

src/themes/types.ts
```diff
@@ -10,6 +10,82 @@ import { Extendable, ObjectOf, ObjectOrFunc } from '../../types/utils'
...
+  700: string
+  800: string
+  900: string
+}
+
+/**
+ * A type for a predefined natural colors.
+ */
+type NaturalColorsStrict = Partial<{
+  blue: ColorVariants
```
kuzhelov, line 37: frankly, still have question whether we should predefine this set - essentially, this would imply that we will _request_ this set to be defined by the client, even in case if only two colors are intended to be used by the app. We should think about making this process as smooth as possible for the client, as color palette may be far from being the main concern of the one who has decided to use Stardust (and, potentially, other important features of theming)

src/themes/types.ts
```diff
@@ -10,6 +10,82 @@ import { Extendable, ObjectOf, ObjectOrFunc } from '../../types/utils'
...
+  yellow: ColorVariants
+}>
+
+export type NaturalColors = Extendable<NaturalColorsStrict, ColorVariants>
+
+/**
+ * A type for a predefined state colors.
+ */
+export type ContextualColorsStrict = Partial<{
+  text: ColorVariants
```
kuzhelov, line 54: emm, not sure whether this `text` key delivers any context

src/themes/types.ts
```diff
@@ -30,11 +106,19 @@ export type State = ObjectOf<any>
 // ========================================================
 
 export interface SiteVariablesInput extends ObjectOf<any> {
+  colors?: ColorPalette
+  contextualColors?: ContextualColors
+  emphasisColors?: EmphasisColors
+  naturalColors?: NaturalColorsStrict
   brand?: string
```
kuzhelov, line 113: what this key stands for? is it theme's name?

kuzhelov (approved):

src/themes/types.ts
layershifter: Defines brand color in our themes, should be definitely removed in next PRs.

src/themes/types.ts
layershifter: Yes, should be moved to other section somehow. We don't reached any decision there.

src/themes/types.ts
layershifter: This type is not exported, we export `NaturalColors` which is extendable. This allows you to get autocomplete on frequent colors and define yours.

docs/src/views/ColorPalette.tsx
layershifter: Nice, will add 👍

docs/src/views/ColorPalette.tsx
layershifter: I removed `i.e. brand to avoid confusions there 👍

docs/src/views/ColorPalette.tsx
layershifter: Corrected to:
> Text variants are also provided as a separate color because in the most cases it's not correct to use grey color for text.

👍

PR #445, opened 2018-11-07
feat(tests): make ts-jest compile tests

bmdalex:
# feat(tests): make `ts-jest` compile tests

## Problem:

**We are not including `test/` folder in our typescript compiler options so nothing under tests is compiled by TypeScript.**

@layershifter did some research and found out the following:

We're using a Gulp's task to run `jest` and our watchers for JSON info files, [#task](https://github.com/stardust-ui/react/blob/master/build/gulp/tasks/test-unit.ts#L34). In the Jest's config we will see, that we're using [`ts-jest`](https://github.com/stardust-ui/react/blob/master/package.json#L74). So the main question why `ts-jest` compiles and runs the code silently? 👎 

## Solution:

The version bump was enough to get it working correctly:

```json
"jest": "^23.6.0",
"jest-axe": "^3.1.0",
"ts-jest": "^23.10.4",
```

image

## Proposal:

I talked to @smykhailov about this as well and he suggested that **we should not rely on all these custom build scripts and tools (such as `gulp` / `grunt`) for our build**. Reasons:
- we end up creating build scripts everywhere that are hard to understand and maintain.
- custom build scripts can introduce lots of errors and bugs as this one;
- custom build scripts contain duplicated work (`webpack` accomplishes out of the script functionality and we only need to configure it).

We should leverage more the `webpack` functionality for setting up our build. This might be a good time to do this since we are just kicking off an effort for creating a Stardust monorepo #433 

**Do we have any issue related to our build system approach and tooling? I can create an issue and copy these findings there.**

What is your take on this, guys?
@levithomason 
@kuzhelov 
@miroslavstastny 
@layershifter 
@mnajdova 
@alinais 
@smykhailov

PR #453, opened 2018-11-08
chore(tests): fail tests on console output

kuzhelov:
### Blocked by
- [x] address listRef issue of `List` component: #489 
- [x] address `Input` component issue (discovered by tests) - fix introduced as part of this PR
image 

---------------

This PR aims to introduce CI gate checks that will verify that there is no console output logged during unit tests run.

### Was

This feature was originally eliminated by the following PR (https://github.com/stardust-ui/react/pull/359) - please, consider to review description where reasoning for this step is expressed. However, this move has resulted in the following problem sneaked into the code and being accidentally committed (please, note that test state is 'PASS'):

image
-------------

### Now

With this PR the feature is restored while, at the same time, **all the benefits of clear error output are preserved for local test runs**. As a result, **for local test runs there will be exactly the same behavior as it was before** (with `yarn test` or `jest` being used), while for CI build `--strict` flag will be applied: `yarn test --strict`. 

image

-----------------

PS. Please, note that this PR will result in failed CI builds till the moment when aforementioned `List` issue will be addressed.


miroslavstastny (approved):

test/setup.strict.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Setup (Strict)
+ * This is the bootstrap code that is run before any tests, utils, mocks in 'strict' mode.
+ */
+require('./setup.common')
+
+jest.spyOn(console, 'log')
```
bmdalex, line 7: 👍 very nicely done

test/jest.config.common.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
```
bmdalex, line 1: @kuzhelov not sure how much these things are related, but is this the place where we specify how the files in `test` directory are compiled? `ts-jest` vs `typescript` compiler? more details in #445

bmdalex (approved):

layershifter:
> The outcome is that actual console errors are improperly reported by jest since they are caught afterAll.

As I remember, I had similar issues in https://github.com/Semantic-Org/Semantic-UI-React/pull/2971. I think that we should wait for @levithomason there before merge.

test/jest.config.common.js
kuzhelov: actually, you are partially right, but let me provide full story behind that.

Currently `ts-jest` tool is responsible for transpiling jest tests before they will run - and this fact is stated by value of `transform` option of Jest config file. Thus, to run the tests we don't need to do any transpilation ourselves for test files **as long as this tool is used for making necessary transform**. Another thing to note is that `ts-jest` doesn't produce any (visible) transpilation artifacts - in contrast to the process that is used to compile sources.

For source files, in contrast, **`tsconfig.json` file is in charge of defining how and which files should be transpiled**.

Answering on your question - while, yes, `transform` option serves to suggest Jest that `ts-jest` should be used for making transform, there is no was to specify directly there that 'raw' TS compiler (without being wrapped as a plugin) should be used. If we would like to use raw TS compiler, we would need to make test process consisting of two steps:
- transpile test files and **store them**
- run tests for transpiled (JS) test files

To my mind this move would definitely complicate things, but let me first take a look on the issue to obtain more details on why this move might be necessary

test/setup.strict.ts
kuzhelov: I would like to confess that those changes are mine, but this is not the case :) actually, the approach of using spies for console methods was already introduced before and just have been removed from sources for a while

silviuaavram:
Will address the error so we can go ahead with the merge. I will not create any additional bug, as this PR should serve for tracking.

kuzhelov:
@layershifter 

> As I remember, I had similar issues in Semantic-Org/Semantic-UI-React#2971. I think that we should wait for @levithomason there before merge.

we may wait, ok, but if it will take too long, it seems that we'd better address the problem that we clearly have now, rather than just let similar problems sneak through

mnajdova:
The issue with the Input test lays in the `implementsWrapperProp` test. We have test there `wraps the component with a custom element` which tests whether the wrapper can be implemented as a custom element - `span` in this case:
```
    it('wraps the component with a custom element', () => {
      wrapperTests(mount(<Component wrapper={<span />} />).find('span'))
    })
```
I know @Bugaa92 is working currently on a PR that will prevent spreading props in this case, so it may be fixed in his PR. For unblocking this one, I would suggest removing this test, and addressing it later if we decide that this is something we want to support. @kuzhelov @miroslavstastny what do you think?

mnajdova:
Btw, the usage `<Input wrapper={<span />} />` doesn't even work in the docs...

mnajdova (approved):

kuzhelov:
@mnajdova, @Bugaa92, have introduced fix as part of this PR, thus will merge now

PR #455, opened 2018-11-09
test(behaviors): make behavior tests and specification mandatory

kolaps33:
In this PR implemented:
Test:
- if there is no @specification tag in the behavior file, then test will be generated and failed
- if some line under @specification tag is not matching any regex(any test), then test will be generated and  failed

Documentation:
- in the documentation both tags @specification and @description will be displayed. 
- if @specification and @description are missing, then test will be generated and failed. This should ensure that there is NO empty description in documentation. 

Reamining work for future:
https://github.com/stardust-ui/react/issues/476
https://github.com/stardust-ui/react/issues/500


test/specs/behaviors/testHelper.tsx
```diff
@@ -17,6 +17,13 @@ export interface TestDefinition {
   testMethod: (TestMethod) => void
 }
 
+const excludedFiles = [
+  'chatBehavior.ts', // ticket 410026
```
sophieH29: what is that ticket number?

build/gulp/plugins/gulp-component-menu-behaviors.ts
```diff
@@ -39,17 +40,17 @@ export default () => {
...
+      // getting object that describes '@description' and '@specification' part of the comment's text
       if (!_.isEmpty(blockComments)) {
         const commentTokens = doctrine.parse(blockComments[0].raw, { unwrap: true }).tags
         const descriptionToken = commentTokens.find(token => token.title === 'description')
-        description = descriptionToken ? descriptionToken.description : emptyDescriptionText
-      } else {
-        description = emptyDescriptionText
+        const specificationToken = commentTokens.find(token => token.title === 'specification')
+        specificationText = specificationToken ? specificationToken.description : ''
+        description = descriptionToken ? descriptionToken.description : ''
```
sophieH29: nit: can be done a small improvement:

```js
const getTextFromCommentToken = (commentTokens, tokenTitle): string {
    const token = commentTokens.find(token => token.title === tokenTitle)
    return token ? token.description : ''
}

description = getTextFromCommentToken(commentTokens, 'description')
specificationText = getTextFromCommentToken(commentTokens, 'specification')

```

test/specs/behaviors/testHelper.tsx
```diff
@@ -96,4 +117,20 @@ export class TestHelper {
     }
     return stringToConvert
   }
+
+  private verifySpecificationTag(behaviorFileName: string) {
+    if (!excludedFiles.find(item => item === behaviorFileName)) {
+      test(`${behaviorFileName} : File is missing specification tag.`, () => {
+        fail(
+          `File should have specification tag. If tests are written in separate file then add behavior file name into 'const excludedFiles'.`,
+        )
```
sophieH29: nit: without ```const```, just `add behavior file name to 'excludedFiles' `

sophieH29 (approved):

build/gulp/plugins/gulp-component-menu-behaviors.ts
```diff
@@ -39,17 +45,18 @@ export default () => {
...
       if (!_.isEmpty(blockComments)) {
         const commentTokens = doctrine.parse(blockComments[0].raw, { unwrap: true }).tags
-        const descriptionToken = commentTokens.find(token => token.title === 'description')
-        description = descriptionToken ? descriptionToken.description : emptyDescriptionText
-      } else {
-        description = emptyDescriptionText
+        const descriptionTokenDesc = getTextFromCommentToken(commentTokens, 'description')
+        const specificationTokenDesc = getTextFromCommentToken(commentTokens, 'specification')
+
+        specificationText = specificationTokenDesc ? specificationTokenDesc : ''
```
sophieH29: no need to do that check, as it is already done in ```getTextFromCommentToken```, so it will return either value or empty string

src/lib/accessibility/Behaviors/Button/buttonBehavior.ts
```diff
@@ -2,9 +2,9 @@ import { Accessibility } from '../../types'
 import * as _ from 'lodash'
 
 /**
- * @description
- *  Adds role='button' if element type is other than 'button'. This allows screen readers to handle the component as a button.
- *  Adds attribute 'aria-disabled=true' based on the property 'disabled'. This can be overriden by providing 'aria-disabled' property directly to the component.
+ * @specification
```
kuzhelov, line 5: 👍

src/lib/accessibility/Behaviors/Icon/iconBehavior.ts
```diff
@@ -2,8 +2,9 @@ import { Accessibility } from '../../types'
 
 /**
  * @description
- * Adds attribute 'aria-hidden=true' to icon.
  * Icon is usually only visual representation and therefore is hidden from screen readers.
+ * @specification
```
kuzhelov, line 7: nit: could we, please, just for the sake of it being visually easier to read, split these sections by empty line?

src/lib/accessibility/Behaviors/Icon/iconBehavior.ts
```diff
@@ -2,8 +2,9 @@ import { Accessibility } from '../../types'
 
 /**
  * @description
```
kuzhelov, line 4: it seems that we can omit @description tag altogether, as it is redundant. It will look like:

```js
/**
 * Adds attribute 'aria-hidden=true', if there is no 'alt' property provided.	
 * Image is usually only visual representation and therefore is hidden from screen readers.	
 *  
 * @specification
 * Adds attribute 'aria-hidden=true', if there is no 'alt' property provided.
 */
```

A benefit of this move is that we will be able to say that **only specially tagged sections of comments will undergo special processing** - this is, actually, common thing that developer might expect.

src/lib/accessibility/Behaviors/List/selectableListBehavior.ts
```diff
@@ -3,8 +3,9 @@ import { Accessibility } from '../../types'
 
 /**
  * @description
- * Adds role='listbox'.
  * The listbox role is used to identify an element that creates a list from which a user may select one or more items.
+ * @specification
```
kuzhelov, line 8: like so much this formal-vs-informal split ❤️

test/specs/behaviors/testHelper.tsx
```diff
@@ -60,25 +67,39 @@ export class TestHelper {
...
+          variant.specification.split('\n').forEach(singleLineText => {
+            this.iterateRegexDefinitions(singleLineText, variant.name)
+          })
+        }
       })
     })
   }
 
   public iterateRegexDefinitions(singleLineText: string, behaviorName: string) {
+    let regMatched = false
```
kuzhelov: lets omit shortcuts - `regexMatched` would be much more easier to read :)

test/specs/behaviors/testHelper.tsx
```diff
@@ -60,25 +67,39 @@ export class TestHelper {
...
         this.filteredDescriptionWithAssignedTestMethod.push({
           behaviorName,
           testMethod: testDefinition.testMethod,
           params: result,
         })
       }
     })
+    if (!regMatched) {
+      test(`${behaviorName} \n LINE: ${singleLineText} `, () => {
+        fail(`Line mentioned in test name doesn't match any regex expressions.`)
```
kuzhelov: I would suggest something that is more descriptive, like: `Line mentioned in **behavior specification** doesn't match any regex expression of validation tests.`

test/specs/behaviors/testHelper.tsx
```diff
@@ -60,25 +67,39 @@ export class TestHelper {
...
   }
 
   public iterateRegexDefinitions(singleLineText: string, behaviorName: string) {
+    let regMatched = false
     this.testDefinitions.forEach(testDefinition => {
       const regex = new RegExp(testDefinition.regexp)
       const result = regex.exec(singleLineText)
       if (result) {
+        regMatched = true
         this.filteredDescriptionWithAssignedTestMethod.push({
```
kuzhelov: this should now be renamed to something like `filteredSpecificationWithAssignedTestMethod` - this also attributes to related types like `FilteredDescription -> FilteredSpecification`. Otherwise the logic will be read in misleading way

test/specs/behaviors/testHelper.tsx
```diff
@@ -96,4 +117,20 @@ export class TestHelper {
...
+    if (!excludedFiles.find(item => item === behaviorFileName)) {
+      test(`${behaviorFileName} : File is missing specification tag.`, () => {
+        fail(
+          `File should have specification tag. If tests are written in separate file then add behavior file name into 'excludedFiles'.`,
+        )
+      })
+    }
+  }
+
+  private createMissingDescriptionTest(behaviorFileName: string) {
```
kuzhelov: lets use more expressive name here, as there is nothing that actually _tests_ missing description in the suggested logic - the test is apriori failed (thus, it is better to use 'failed' word in name):

`createMissingDescriptionTest` -> `failDescriptionPresenceTest`

This move will be aligned with other names that you are using for these conditional expressions: https://github.com/stardust-ui/react/pull/455/files#diff-2a7cb3f0b6bbaad1f964056827a99caeR70

test/specs/behaviors/testHelper.tsx
```diff
@@ -96,4 +117,20 @@ export class TestHelper {
...
+    test(`${behaviorFileName} : File is missing description.`, () => {
```
kuzhelov: maybe would be better to use `Accessibility behavior is missing description` (as this, in fact, what we do care about and is more precise description for failure reason). Lets compare:

- `buttonBehavior.tsx: File is missing description.` - not sure, which description we are talking about? for _file_?? 
- `buttonBehavior.tsx: Accessibility behavior is missing description.` - now it is clear

kuzhelov: same applies to `fail('File should have description.')` (no need to use interpolation quotes, btw - regular ones '' will be just enough)

test/specs/behaviors/testHelper.tsx
```diff
@@ -96,4 +117,20 @@ export class TestHelper {
     }
     return stringToConvert
   }
+
+  private verifySpecificationTag(behaviorFileName: string) {
+    if (!excludedFiles.find(item => item === behaviorFileName)) {
+      test(`${behaviorFileName} : File is missing specification tag.`, () => {
+        fail(
+          `File should have specification tag. If tests are written in separate file then add behavior file name into 'excludedFiles'.`,
```
kuzhelov: actually, it is not `file` but `accessibility behavior` that should have a `specification` tag :) This text seems to be a bit misleading

test/specs/behaviors/testHelper.tsx
```diff
@@ -60,25 +67,39 @@ export class TestHelper {
   public findRegexAndAssingCorrespondingInfoToArray(behaviorMenuItems: any) {
```
kuzhelov, line 67: why `behaviorMenuItems`? seems to be name that has sneaked in for historical reasons - or here we are talking about menu items that will be displayed in docs?

build/gulp/plugins/gulp-component-menu-behaviors.ts
```diff
@@ -58,6 +62,7 @@ export default () => {
         variations: {
           name: behaviorVariantName,
           text: description,
```
kuzhelov: lets rename `text` => `description`, as general `text` is very hard to reason about while reading the following lines: https://github.com/stardust-ui/react/pull/455/files#diff-105d2d2a4ca631ebeb4f7bdafcc0542bR74 . Then, with all suggested renames, as a result we will have:

```js
// or maybe behaviorVariations
variations = {
  name: behaviorVariantName,
  description,
  specification
}
```

To me it seems to be much more expressive and unambiguous - what do you think?

build/gulp/plugins/gulp-component-menu-behaviors.ts
```diff
@@ -39,17 +45,15 @@ export default () => {
...
-      // getting object that describes '@description' part of the comment's text
+      // getting object that describes '@description' and '@specification' part of the comment's text
       if (!_.isEmpty(blockComments)) {
         const commentTokens = doctrine.parse(blockComments[0].raw, { unwrap: true }).tags
-        const descriptionToken = commentTokens.find(token => token.title === 'description')
-        description = descriptionToken ? descriptionToken.description : emptyDescriptionText
-      } else {
-        description = emptyDescriptionText
+        description = getTextFromCommentToken(commentTokens, 'description')
+        specificationText = getTextFromCommentToken(commentTokens, 'specification')
```
kuzhelov: to be consistent with naming approach taken, `specification` should be just enough here

test/specs/behaviors/testHelper.tsx
```diff
@@ -60,25 +67,39 @@ export class TestHelper {
...
     behaviorMenuItems.forEach(behavior => {
       behavior.variations.forEach(variant => {
-        variant.text.split('\n').forEach(singleLineText => {
-          this.iterateRegexDefinitions(singleLineText, variant.name)
-        })
+        if (!variant.specification && !variant.text) {
```
kuzhelov: it is very hard to reason what `variation.text` stands for. Lets rename it to `variation.description`, as this, in fact, what it is

test/specs/behaviors/testHelper.tsx
```diff
@@ -60,25 +67,39 @@ export class TestHelper {
...
+        } else {
+          variant.specification.split('\n').forEach(singleLineText => {
+            this.iterateRegexDefinitions(singleLineText, variant.name)
+          })
+        }
       })
     })
   }
 
   public iterateRegexDefinitions(singleLineText: string, behaviorName: string) {
```
kuzhelov: singleLineText -> singleSpecLine, this would better reflect semantics of this parameter

test/specs/behaviors/testHelper.tsx
```diff
@@ -17,6 +17,13 @@ export interface TestDefinition {
   testMethod: (arg: TestMethod) => void
 }
 
+const excludedFiles = [
```
kuzhelov: it is not very clear what this files are excluded from. Maybe we could name it as something like `skipSpecChecksForFiles`, so it would be more expressive - because, note, this files still will undergo description presence tests, and this fact is not clearly seen now

test/specs/behaviors/testHelper.tsx
```diff
@@ -60,25 +67,39 @@ export class TestHelper {
   public findRegexAndAssingCorrespondingInfoToArray(behaviorMenuItems: any) {
     behaviorMenuItems.forEach(behavior => {
       behavior.variations.forEach(variant => {
-        variant.text.split('\n').forEach(singleLineText => {
-          this.iterateRegexDefinitions(singleLineText, variant.name)
-        })
+        if (!variant.specification && !variant.text) {
+          this.createMissingDescriptionTest(variant.name)
+        }
+        if (!variant.specification) {
```
kuzhelov: I would suggest to include `skippedFiles` filter right here, so it will be like:

```js
if (!variant.specification && skipSpecCheckForFiles.contains(variant.name)) {
  ...
}
```

kuzhelov (commented):
most of my comments are about renaming - lets try to polish the code a bit so it will be more expressive :)

src/lib/accessibility/Behaviors/Icon/iconBehavior.ts
kolaps33: new line added :)

src/lib/accessibility/Behaviors/List/selectableListBehavior.ts
kolaps33: thank you :) 👍

test/specs/behaviors/testHelper.tsx
kolaps33: yes this is from historical reasons, we named it like it...
and this is basically whole object of loaded file "behaviorMenu.json"

src/lib/accessibility/Behaviors/Icon/iconBehavior.ts
kolaps33: created for this new ticket:
https://github.com/stardust-ui/react/issues/501

build/gulp/plugins/gulp-component-menu-behaviors.ts
kolaps33: I change to "specificationText", because if I used just "specification" it does lint error on the line 64 below...

build/gulp/plugins/gulp-component-menu-behaviors.ts
```diff
@@ -38,26 +44,25 @@ export default () => {
...
 
-      let description
+      let descriptionText
+      let specificationText
       const fileContent = fs.readFileSync(file.path).toString()
       const blockComments = extract(fileContent).filter(comment => comment.type === 'BlockComment') // filtering only block comments
-      const emptyDescriptionText = 'Behavior file has no description.'
 
-      // getting object that describes '@description' part of the comment's text
+      // getting specification of the comment's text
```
kuzhelov: nit: actually, it is about `getting 'description' AND 'specification' from the comment's text

build/gulp/plugins/gulp-component-menu-behaviors.ts
```diff
@@ -38,26 +44,25 @@ export default () => {
       const behaviorVariantName = file.basename
       const behaviorName = path.basename(dir)
 
-      let description
+      let descriptionText
```
kuzhelov: nit: these two could be declared closer to the place where those are initialized (`if` statement) - i.e. on line 52.

Actually, you could use more expressive code here, with no `let`:

```js
  const fileContent = ..
  const blockComments = ...

  const variation = {
    name: behaviorVariantName
  }
  
  if (...) {
     ...
    variation.description = getTextFromCommentToken(commentTokens, 'description')
    variation.specification = getTextFromCommentToken(commentTokens, 'specification')
    ...
  }

File: build/gulp/plugins/gulp-component-menu-behaviors.ts (hunk 17 above)
Comment (kuzhelov) [reply build/gulp/plugins/gulp-component-menu-behaviors.ts]: please, take a look on this comment - it should resolve this problem as well :)

https://github.com/stardust-ui/react/pull/455/files#r235030213

File: docs/src/components/DocsBehaviorRoot.tsx [hunk 24]
```diff
@@ -71,14 +71,24 @@ class DocsBehaviorRoot extends React.Component<any, any> {
...
                   <br />
-                  {variation.text.split('\n').map((splittedText, keyValue) => {
-                    return (
-                      <span key={keyValue}>
-                        {splittedText}
-                        <br />
-                      </span>
-                    )
-                  })}
+                  {variation.text &&
```
kuzhelov: oops, `text` now should be `description` here

test/specs/behaviors/testHelper.tsx
```diff
@@ -1,7 +1,7 @@
 import * as _ from 'lodash'
 import { Accessibility } from '../../../src/lib/accessibility/types'
 
-interface FilteredDescription {
+interface FilteredSpecification {
```
kuzhelov, line 4: 👍

test/specs/behaviors/testHelper.tsx
```diff
@@ -17,11 +17,18 @@ export interface TestDefinition {
   testMethod: (arg: TestMethod) => void
 }
 
+const skipSpecChecksForFiles = [
```
kuzhelov, line 20: 👍 the intent is much more clear now, thank you!

test/specs/behaviors/testHelper.tsx
kuzhelov: got it, thank you!

test/specs/behaviors/testHelper.tsx
```diff
@@ -60,32 +67,50 @@ export class TestHelper {
...
-        variant.text.split('\n').forEach(singleLineText => {
-          this.iterateRegexDefinitions(singleLineText, variant.name)
-        })
+        if (!variant.specification && !variant.description) {
+          this.failDescriptionPresenceTest(variant.name)
+        }
+        if (!variant.specification && !skipSpecChecksForFiles.find(item => item === variant.name)) {
+          this.failSpecificationPresenceTest(variant.name)
+        } else {
+          variant.specification.split('\n').forEach(singleSpecLine => {
```
kuzhelov: sorry, I do know that it was my proposal, but now it is seen that name of the arg could be further simplified: `singleSpecLine` -> `specLine`. 

Please, feel free to reason by yourself whether it is worth to introduce, as it is quite cosmetic change :)

test/specs/behaviors/testHelper.tsx
```diff
@@ -60,32 +67,50 @@ export class TestHelper {
...
+              this.iterateRegexDefinitions(singleSpecLine, variant.name)
+            }
+          })
+        }
       })
     })
   }
 
-  public iterateRegexDefinitions(singleLineText: string, behaviorName: string) {
+  public iterateRegexDefinitions(singleSpecLine: string, behaviorName: string) {
```
kuzhelov: nit: the same here - `singleSpecLine` -> `specLine` might be better fit here. Once again, please, decide by yourself what you'd prefer, just left suggestion :)

test/specs/behaviors/testHelper.tsx
```diff
@@ -60,32 +67,50 @@ export class TestHelper {
...
           behaviorName,
           testMethod: testDefinition.testMethod,
           params: result,
         })
       }
     })
+    if (!regexMatched) {
+      test(`${behaviorName} \n LINE: ${singleSpecLine} `, () => {
+        fail(
+          `Line mentioned in **behavior specification** doesn't match any regex expression of validation tests.`,
```
kuzhelov, line 103: 👍 👍 very explanatory!

kuzhelov (approved):

PR #456, opened 2018-11-09
chore(CI): add a check for CHANGELOG by dangerjs

miroslavstastny:
This PR adds [Danger JS](https://danger.systems/js/).

Danger JS runs tests as part of CI and writes results message back to the PR:
image

It can even fail the CI build in case of an error:
image

Currently there are no meaningful tests in our [`dangerfile.ts`](https://github.com/stardust-ui/react/pull/456/files#diff-6fa27d45e8ab8eb03f57cafdc300982e) but we plan to use the workflow for bundle size checks (#97).

### Caveats
By design danger can run only on CI builds related to a PR and refuses to run on a branch build with no associated PR:
image

This is not compatible with our current CI configuration:
image


layershifter:
Wow-wow, really cool ❤️

dangerfile.ts
```diff
@@ -0,0 +1,8 @@
+import { danger, warn } from 'danger'
+
+// Check for a CHANGELOG entry
+const hasChangelog = danger.git.modified_files.some(f => f === 'CHANGELOG.md')
```
kuzhelov, line 4: note, `FocusZone` has changelog file as well - probably, it is still better to improve this filter before merging, for the sake of making the logic of warning to be consistent

image

miroslavstastny: Ahh, good catch, I haven't realised that.

miroslavstastny: Actually, `modified_files` contains paths from repo root so the `hasChangelog` is true only if `CHANGELOG.md` in root is changed.

miroslavstastny:
CircleCI reconfigured to build PRs only. Ready for review.

layershifter (approved):
Cool job 👍 

Only one thing, please update a PR's title, something like: `chore(CI): add a check for CHANGELOG by dangerjs`

.circleci/config.yml
```diff
@@ -55,6 +55,12 @@ jobs:
       - run:
           name: Project Tests
           command: yarn test:projects:cra-ts
+      - run:
+          name: Danger JS
```
kuzhelov, line 59: would suggest some other name that will provide the intent of this step, like 'Conventions Tests'

dangerfile.ts
```diff
@@ -0,0 +1,8 @@
...
+
+if (!hasChangelog) {
+  warn('Please add a changelog entry for your changes.')
```
kuzhelov: maybe something like `Ensure that CHANGELOG is not updated intentionally.` or `There are no updates provided to CHANGELOG. Ensure there are no publicly visible changes introduced by this PR.` The benefit is that we won't blindly suggest to update CHANGELOG for those cases where it might not be necessary

kuzhelov (approved):

.circleci/config.yml
miroslavstastny: I would rather leave it as it is (for now). I intentionally used that name to make it clear that the step is driven by `dangerfile.ts`.

kuzhelov: lets discuss offline - cannot agree here, as there are other steps that are driven by dedicated tools (like Screener), but so far we have used semantics-based approach for naming.

PR #457, opened 2018-11-09
feat(Popup): add focus trap

kuzhelov:
This PR introduces focus trap functionality to `Popup` - this will allow to focus popup's content right after popup is opened.

### TODO
- [x] introduce focus trap examples
- [x] update CHANGELOG

### API

```jsx
// focusable popup content will be trapped by default
<Popup trigger={<Button .. />} content={<Input />} />

// focusTrap={false} will disable default focus trap
<Popup focusTrap={false} trigger={<Button .. />} content={<Input />} />

// it is possible to customize focus trap behavior
<Popup focusTrap={{ isClickableOutsideFocusTrap: false, elementToFocusOnDismiss: null }} 
  trigger={<Button .. />} 
  content={<Input />} 
/>
```

Should be clear about that - the final goal of all the refactoring process around `Popup` component is to end up at the point where there won't be any code duplication - and this implies that finally **`Popup` will reuse `Portal`'s functionality**. At the same it is intentionally step-by-step process taken into consideration, for the sake of ensuring that only the necessary parts are introduced to both components, as well as there is clear separation of concerns between them.

>Where we are at the moment with `Popup`?

This PR will solidify final bits of the functionality expected from the `Popup` on the client side - after that the next plans are to do inner refactoring by reusing `Portal` component as part of `Popup`'s implementation - **with the interface and behavior exposed to the client for `Popup` won't be affected anyhow**.


jurokapsiar:
can you add unit test that cover the changes? Especially if we want to do internal refactoring.

kuzhelov:
@jurokapsiar 

yes, but could we, please, start with reviewing the functionality itself - whether it satisfies the criterias. When we agree on that, I will add the tests. Also, given that this is a highly anticipated feature, I would rather split the PRs to not introduce the unnecessary delay (note that there were no tests for that before).

docs/src/examples/components/Popup/Types/PopupNoFocusTrapExample.shorthand.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import { Button, Input, Popup } from '@stardust-ui/react'
+
+const PopupNoFocusTrapExample = () => (
+  <>
+    <Popup
```
hughreeling: Having example here first with focus trap enabled is super confusing.  Yes, it's all correct, but I didn't read carefully and thought it wasn't working.  I would get rid of the 'default' option with it on.

docs/src/examples/components/Popup/Types/PopupFocusTrapExample.shorthand.tsx
```diff
@@ -0,0 +1,50 @@
...
+const PopupFocusTrapExample = () => (
+  <>
+    <Popup
+      trigger={<Button icon="expand" content="Default trap behavior" />}
+      content={{ content: <Input icon="search" placeholder="Search..." /> }}
+    />
+
+    <Popup
+      trigger={<Button icon="expand" content="Lets trap your focus even more ツ" />}
+      focusTrap={{
```
sophieH29: we kind of duplicating logic which is done in the Portal, for the exposing ```focusTrap``` prop.
So the original idea is
- Popup is not focus trapped by default
- FocusTrapZone props are coming from a behavior same way as for FocusZone.
Of course it's part of this PR, but what we need is to read FocusTrapZone props from ```accessibility.focusTrapZone = {} ```

After your PR merged, I will create PR that ads FocusTrapZone props to PopupBehavior.

docs/src/examples/components/Popup/Types/PopupNoFocusTrapExample.shorthand.tsx
kuzhelov: will swap them, thanks!

docs/src/examples/components/Popup/Types/PopupFocusTrapExample.shorthand.tsx
kuzhelov: have discussion with @sophieH29, agreed on the following plan:
- prepare branch where focus trap properties will be read from behavior by `Popup`
- merge necessary changes to introduce this behavior into this PR
- remove `focusTrap` prop from the API of `Popup`

docs/src/examples/components/Popup/Types/PopupNoFocusTrapExample.shorthand.tsx
kuzhelov: addressed

jurokapsiar (approved):

sophieH29 (approved):

PR #458, opened 2018-11-09
 docs(ComponentExample): fix types

layershifter:
This PR fixes small types issues.

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
+++ b/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
@@ -21,8 +21,8 @@ import { mergeThemeVariables } from '../../../../../src/lib/mergeThemes'
 import { ThemeContext } from '../../../context/theme-context'
 
 export interface ComponentExampleProps extends RouteComponentProps<any, any> {
-  title: string
-  description: string
+  title: React.ReactNode
+  description?: React.ReactNode
   examplePath: string
   themeName?: string
 }
@@ -724,10 +724,10 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
   }
 }
 
-const ComponentExampleWithTheme = React.forwardRef((props: ComponentExampleProps) => (
+const ComponentExampleWithTheme = props => (
   <ThemeContext.Consumer>
     {({ themeName }) => <ComponentExample {...props} themeName={themeName} />}
   </ThemeContext.Consumer>
-))
+)
 
 export default withRouter(ComponentExampleWithTheme)
--- a/docs/src/components/ComponentDoc/ComponentExample/ComponentExampleTitle.tsx
+++ b/docs/src/components/ComponentDoc/ComponentExample/ComponentExampleTitle.tsx
@@ -1,4 +1,3 @@
-import PropTypes from 'prop-types'
 import * as React from 'react'
 import { Header } from 'semantic-ui-react'
 
@@ -6,12 +5,12 @@ const titleStyle = {
   margin: 0,
 }
 
-export default class ComponentExampleTitle extends React.PureComponent<any, any> {
-  static propTypes = {
-    description: PropTypes.node,
-    title: PropTypes.node,
-  }
+interface ComponentExampleTitleProps {
+  description?: React.ReactNode
+  title: React.ReactNode
+}
 
+export default class ComponentExampleTitle extends React.PureComponent<ComponentExampleTitleProps> {
   render() {
     const { description, title } = this.props
     return (
--- a/docs/src/components/DocsBehaviorRoot.tsx
+++ b/docs/src/components/DocsBehaviorRoot.tsx
@@ -46,7 +46,12 @@ class DocsBehaviorRoot extends React.Component<any, any> {
           {behaviorMenuItems
             .find(behavior => behavior.displayName === _.capitalize(match.params.name))
             .variations.map((variation, keyValue) => (
-              <Grid.Row key={keyValue} className="docs-example" style={exampleStyle}>
+              <Grid.Row
+                key={keyValue}
+                className="docs-example"
+                id={_.kebabCase(variation.name)}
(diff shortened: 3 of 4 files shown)


docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -21,8 +21,8 @@ import { mergeThemeVariables } from '../../../../../src/lib/mergeThemes'
 import { ThemeContext } from '../../../context/theme-context'
 
 export interface ComponentExampleProps extends RouteComponentProps<any, any> {
-  title: string
-  description: string
+  title: React.ReactNode
+  description?: React.ReactNode
```
layershifter, line 25: `title` and `description` are simply passed to `ComponentExampleTitle` that accepts `ReactNode`, too.

docs/src/components/ComponentDoc/ComponentExample/ComponentExampleTitle.tsx
```diff
@@ -1,17 +1,16 @@
...
 
-export default class ComponentExampleTitle extends React.PureComponent<any, any> {
-  static propTypes = {
-    description: PropTypes.node,
-    title: PropTypes.node,
-  }
+interface ComponentExampleTitleProps {
+  description?: React.ReactNode
+  title: React.ReactNode
+}
```
layershifter, line 11: Replaced there `propTypes` with valid types.

docs/src/components/DocsBehaviorRoot.tsx
```diff
@@ -54,7 +54,6 @@ class DocsBehaviorRoot extends React.Component<any, any> {
                   <div style={{ display: 'flex' }}>
                     <div style={{ flex: '1', marginBottom: '1em' }}>
                       <ComponentExampleTitle
-                        id={_.kebabCase(variation.name)}
```
layershifter, line 57: This stuff is not working because `ComponentExampleTitle` doesn't spread rest props and accepts only `title` and `description`.

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -724,10 +724,10 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
   }
 }
 
-const ComponentExampleWithTheme = React.forwardRef((props: ComponentExampleProps) => (
+const ComponentExampleWithTheme = props => (
```
layershifter, line 727: `React.forwardRef` doesn't make a sense because the second `ref` argument is not used.

kuzhelov: 👍

kuzhelov: 👍

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
kuzhelov: not sure about that. We should express correct semantics to the client of this component, and if we would search all the places where `<ComponentExample />` is used, all of them suggest that `string` values should be provided for these two

docs/src/components/ComponentDoc/ComponentExample/ComponentExampleTitle.tsx
kuzhelov: 👍

docs/src/components/DocsBehaviorRoot.tsx
kuzhelov: this will prevent us from referencing specific example - previously it was possible to share links like the following one: https://stardust-ui.github.io/react/components/popup#types-popup-content-wrapper-example

Now I might expect that it won't work. Seems that this is not a right approach to fix this types-related issue - am I missing something?

layershifter: image

This change is related to behaviours and it's is not working there.

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
layershifter: Why component's description should be a `string`? I think that we should be able to pass an additional formatting there as it was in SUIR:

image

kuzhelov: got it - so, essentially proposal is to augment current capabilities, where `string` was previously used everywhere?

docs/src/components/DocsBehaviorRoot.tsx
kuzhelov: I would expect that it rather should be there, if we would talk about perspectives, especially given that it would be consistent with component's examples. It seems that with this change we are making step in opposite direction - while now these behavior pages lack functionality of grabbing reference to section (and this is, in fact, the functionality that better to be introduced), we are trying to remove existing bits altogether

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
layershifter: In fact, yes. We have the same thing in SUIR docs, so it will work nicely there.

I think that the `string` type was assigned by mistake and this change just corrects the behaviour to correct.

docs/src/components/DocsBehaviorRoot.tsx
layershifter: I will check this thing and will get it working.

layershifter: I restored the `id` on `Grid.Row`, it's the same thing as in `ComponentExample`

kuzhelov: great, thank you!

kuzhelov (approved):

PR #460, opened 2018-11-09
feat(utils): replace react-docgen with typescript

mnajdova:
# RFC 
This PR aims to replace react-docgen with alternative that will generate the info files for the components using the typescript interfaces. The PR introduces the react-docgen-typescript library, that out of the interfaces for the props of the component, generates json in the same format as the react-docgen library. I added one interface for the UIComponentProps with some common props that can be extended in the other component's props.

## Example
For example, here is how the Button component prop section looks now, in comparison with the previous:
### NOW
image
### PREV
image

You can see that the types of the props are much descriptive with the new approach, but there are still some things that needs to be taken into account.
There is some bug I guess, because the types of the @params used inside the JSDoc are not picked up. I created issue for this on their github repo. In addition, the default value of the props are taken just if the value is some primitive type. However for the second, there is an alternative way of defining, by adding @default in the comment of the property.

Please share your feedback, so that we can make progress with this on the other components as well. @alinais @Bugaa92 @levithomason @kuzhelov @miroslavstastny @layershifter 

# Updated components
- [x] Accordion
- [x] Attachment
- [x] Avatar
- [x] Button
- [x] Chat
- [x] Divider
- [x] Form
- [x] Grid
- [x] Header
- [x] Icon
- [x] Image
- [x] Input
- [x] ItemLayout
- [x] Label
- [x] Layout
- [x] List
- [x] Menu
- [x] Popup
- [x] Portal
- [x] Provider
- [x] Text
- [x] Radio Group
- [x] Segment
- [x] Status
- [x] Slot


layershifter:
It seems that we can drop `recast` from dependencies, too.

layershifter:
We will also need to correct some types `"before" | "after"`. However, it looks promising, nice job 👍

package.json
```diff
@@ -91,6 +91,7 @@
     "lodash": "^4.17.10",
     "prop-types": "^15.6.1",
     "react-custom-scrollbars": "^4.2.1",
+    "react-docgen-typescript": "^1.12.0",
```
layershifter: Oops, this should be in `devDependencies`

mnajdova: Thanks for the catch! Moved.

mnajdova:
I also noticed that some Component are missing some common properties:
AccordionContent
- styles, variables

AccordionTitle
- styles, variables

Attachment
- className

Popup (maybe these shouldn't even be added, but just in case...)
- as
- styles
- variables

Portal (maybe these shouldn't even be added, but just in case...)
- as
- className
- styles
- variabels

Segment
- children (it's not added in props, but it is actually used)

Slot
- children (it's not added in props, but it is actually used)

Text
- children (it's not added in props, but it is actually used)

Using some common interface for these would help us avoiding some of these mistake, plus we would have consistent description for the common properties without having to copy paste the same thing, or change it everywhere if we decide for better description.

levithomason (approved):

mnajdova:
Future work:
- Had to add the accessibility defaultProps everywhere with a @default - we have to find alternative to this
- Avatar – getInitials default value is not shown
- ItemLayout render*** (some bugs even on prod) - default value not shown
- Label accessibility was just added without description and default (didn't change anything there)
- Portal context default is not shown
- Popup content type has to be checked again
- Variables type is any - there is interface for this but is not shown.

PR #462, opened 2018-11-11
chore: Upgrade to Webpack 4

miroslavstastny:
Upgrade Webpack to version 4
https://webpack.js.org/migrate/4/

### DIFF
--- a/build/webpack.config.umd.ts
+++ b/build/webpack.config.umd.ts
@@ -1,4 +1,4 @@
-import * as webpack from 'webpack'
+// import * as webpack from 'webpack'
 
 import config from '../config'
 import webpackConfig from '../webpack.config'
@@ -8,6 +8,7 @@ const { paths } = config
 
 const webpackUMDConfig = {
   target: 'web',
+  mode: 'production',
   devtool: false,
   entry: {
     [pkg.name]: paths.src('umd.ts'),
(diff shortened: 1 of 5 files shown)


build/webpack.config.umd.ts
```diff
@@ -8,6 +8,7 @@ const { paths } = config
 
 const webpackUMDConfig = {
   target: 'web',
+  mode: 'production',
```
miroslavstastny, line 9: required in WebPack 4

build/webpack.config.umd.ts
```diff
@@ -24,20 +25,23 @@ const webpackUMDConfig = {
...
-  plugins: [
-    new webpack.optimize.UglifyJsPlugin({
-      compress: {
-        unused: true,
-        dead_code: true,
-        warnings: false,
-      },
-    }),
-  ],
+  // plugins: [
```
miroslavstastny: Default in production mode.
Should we configure it somehow?

webpack.config.ts
```diff
@@ -38,7 +39,7 @@ const webpackConfig: any = {
     readline: 'empty',
   },
   module: {
-    noParse: [/\.json$/, /anchor-js/],
+    noParse: [/anchor-js/],
```
miroslavstastny, line 42: Without this change, resulting package does not contain files like `package.json`, `Accordion.info.json`...

build/webpack.config.umd.ts
layershifter: Yes, but via `optimization` section of config. We should ensure that we're using `terser` instead of `uglify-es`:
https://twitter.com/dan_abramov/status/1060216917493981184

layershifter:
It seems that we can also do this:
```diff
- "satisfied": "satisfied --skip-invalid --ignore \"^typescript$\"",
+ "satisfied": "satisfied --skip-invalid",
```

build/webpack.config.umd.ts
miroslavstastny: Confirmed with @layershifter that this is the default config.

layershifter:
> It seems that we can also do this:

We should update `ts-jest` before, nevermind.

layershifter (approved):

levithomason (approved):

PR #465, opened 2018-11-12
Add focus trap props to Popup behavior

sophieH29:


docs/src/examples/components/Popup/Types/PopupNoFocusTrapExample.shorthand.tsx
```diff
@@ -1,36 +1,20 @@
...
-            <Input icon="search" placeholder="Search..." />
-          </>
-        ),
-      }}
-      trigger={<Button icon="expand" content="Popup that will trap focus" />}
-    />
-  </>
+  <Popup
+    /** 'popupBehavior' does not use focus trap */
+    accessibility={popupBehavior}
```
jurokapsiar: let's maybe call it `popupNoFocusTrapBehavior` to make it more clear

sophieH29: I am afraid it can be more confusing having 2 behaviors:
```popupNoFocusTrapBehavior``` and ```popupFocusTrapBehavior```, as the names could seem the same.
I think better to agree with one of these 2 options:

1. ```popupFocusTrapBehavior``` - default, with focus trap. ```popupBehavior``` - optional, without focus trap

2. ```popupBehavior``` - default, with focus trap. ```popupNoFocusTrapBehavior``` - optional, without focus trap

Currently the 1st option is implemented in the PR, but we can change it

Comment (jurokapsiar) [reply docs/src/examples/components/Popup/Types/PopupNoFocusTrapExample.shorthand.tsx]: what about following 2 names:
- popupBehavior (without focus trap, default)
- modalPopupBehavior (with focus trap)

Comment (sophieH29) [reply docs/src/examples/components/Popup/Types/PopupNoFocusTrapExample.shorthand.tsx]: From my own experience, I could never differentiate popup and modal :) And as a user, I probably won't remember which of them has focus trap. 
I would stick to more semantically descriptive name
Let me try once again:
- popupBehavior (without focus trap, default)
- popupFocusTrapBehavior (with focus trap)

So actually it's the same as initial but default would be without focus trap

Comment (jurokapsiar) [reply docs/src/examples/components/Popup/Types/PopupNoFocusTrapExample.shorthand.tsx]: agree

File: src/components/Popup/Popup.tsx [hunk 2]
```diff
@@ -271,7 +267,7 @@ export default class Popup extends AutoControlledComponent<Extendable<PopupProps
     }
 
     const focusTrapProps = {
-      ...(typeof focusTrap === 'boolean' ? {} : focusTrap),
+      ...(typeof accessibility.focusTrapZone === 'boolean' ? {} : accessibility.focusTrapZone),
```
kuzhelov: can we name it as focus trap? This will be more intuitive for the client

src/lib/accessibility/Behaviors/Popup/popupFocusTrapBehavior.ts
```diff
@@ -0,0 +1,17 @@
...
+import * as _ from 'lodash'
+
+/**
+ * @description
+ * Adds role='button' to 'trigger' component's part, if it is not focusable element and no role attribute provided.
+ * Adds tabIndex='0' to 'trigger' component's part, if it is not tabbable element and no tabIndex attribute provided.
+ * Adds attribute 'aria-disabled=true' to 'trigger' component's part based on the property 'disabled'.
+ * Wraps component in FocusTrapZone, with default prop 'isClickableOutsideFocusTrap' set to 'true'.
+ */
+const popupFocusTrapBehavior: Accessibility = (props: any) => ({
```
kuzhelov, line 12: just to be sure - shouldn't focus trap be a default behavior for the `Popup`?

src/components/Popup/Popup.tsx
sophieH29: yes, I'll rename it

docs/src/examples/components/Popup/Types/index.tsx
```diff
@@ -22,14 +22,9 @@ const Types = () => (
     />
     <ComponentExample
       title="Focus Trap"
-      description="Popup content traps focus on appearance by default - at the same time, this focus trap behavior can be customized."
+      description="Popup content traps focus on appearance by using accessibility behavior 'popupFocusTrapBehavior'. At the same time, this focus trap properties can be customized, by creating your own behavior (see Behaviors section for Popup component)."
```
kuzhelov: we should omit variable names in descriptions

src/lib/accessibility/Behaviors/Popup/popupFocusTrapBehavior.ts
sophieH29: Yep, firstly we decided that the default one would be with focus trap. But after saw my initial commit, it had confusion in namings and understanding between behavior with focus trap and without. We agreed with @jurokapsiar on these options
- popupBehavior (without focus trap, default)
- popupFocusTrapBehavior (with focus trap)

So to bring both semantically descriptive name and more intuitive logic for the clients.
Everyone still can have focus trap, because we expose ```popupFocusTrapBehavior``` to the Stardust lib and show a miningful example on how to use it.

@kuzhelov if you have any other thoughts on it, please, feel free, to share.

docs/src/examples/components/Popup/Types/index.tsx
sophieH29: Did you mean ```popupFocusTrapBehavior```?

src/lib/accessibility/Behaviors/Popup/popupFocusTrapBehavior.ts
```diff
@@ -0,0 +1,17 @@
+import { Accessibility } from '../../types'
+import popupBehavior from './popupBehavior'
+import * as _ from 'lodash'
+
+/**
+ * @description
+ * Adds role='button' to 'trigger' component's part, if it is not focusable element and no role attribute provided.
+ * Adds tabIndex='0' to 'trigger' component's part, if it is not tabbable element and no tabIndex attribute provided.
+ * Adds attribute 'aria-disabled=true' to 'trigger' component's part based on the property 'disabled'.
+ * Wraps component in FocusTrapZone, with default prop 'isClickableOutsideFocusTrap' set to 'true'.
```
kuzhelov: this is very confusing to the client who will read docs. While client is aware of `aria-` and other well-known accessibility aspects, `FocusTrapZone` is an implementation detail. I would suggest to either
- introduce other general description
- omit this statement for now

src/lib/accessibility/Behaviors/Popup/popupFocusTrapBehavior.ts
```diff
@@ -0,0 +1,17 @@
+import { Accessibility } from '../../types'
+import popupBehavior from './popupBehavior'
+import * as _ from 'lodash'
+
+/**
+ * @description
+ * Adds role='button' to 'trigger' component's part, if it is not focusable element and no role attribute provided.
+ * Adds tabIndex='0' to 'trigger' component's part, if it is not tabbable element and no tabIndex attribute provided.
+ * Adds attribute 'aria-disabled=true' to 'trigger' component's part based on the property 'disabled'.
+ * Traps focus inside component and by default allowing click outside this component.
```
kuzhelov: seems that `allowing click outside this component` is not checked by tests. maybe we should omit it for now - and leave only the statement that is verified by tests?

kuzhelov (approved):

src/lib/accessibility/Behaviors/Popup/popupFocusTrapBehavior.ts
sophieH29: Yes, it is not tested, and as we discussed it's done not only for tests. This is also what actually the users will read. So just a bit extended description, specifying the behavior. 
But ok, will remove it for now, as we don't have any places using different configurations with focus trap.

PR #466, opened 2018-11-13
State Manager approach suggested

kuzhelov:
**NOTE: THIS PR WILL NEVER BE MERGED AS IS**

This PR aims to improve original approach provided in this branch: https://github.com/stardust-ui/react/tree/feat/state. 

## Framework-based examples

Following examples are aimed to prove framework-agnosticism of state management logic - for each examples only the logic that 'connects' state to framework is changing. Note that for _all examples content of `stateManager` and `popupStateManager` was just copy-pasted from this PR_.

- Plain vanilla HTML project: https://codesandbox.io/s/7m8vw1pro0
- VueJS project: https://codesandbox.io/s/ll27104k8z

---------------

### Why

This PR introduces alternative **implementation** approach to the idea expressed in original branch devoted to state manager functionality (https://github.com/stardust-ui/react/tree/feat/state) - while being inspired by the same set of principles and ideas.

### Benefits

Please, note that intent of this PR is not to introduce a clean code, but rather to deliver an idea that should introduce the following benefits over original implementation approach:
- there is a **single source of truth for component's state** provided by state manager's state <-> in the original approach state may be consumed from two sources: either from React's state, or from stateManager's state
  - thus, with single source of state data, there is no need to struggle with syncing state manager's state - it is always guaranteed to be up-to-date
- **less verbose**, **less error-prone** and **more intuitive** manager's integration logic with React - the one that client will be responsible for
  - no `(?) => (??) => ...` functions exposed to client's API
- clear **separation of framework-agnostic code and React-specific one**

### API example (with React bindings only)

```js
class Popup extends UIComponent {
 ...
  private stateManager = popupStateManager.withBindings({
    getState: () => {
        // AUTOCONTROLLED state logic
        return { ...this.state, ...this.props } 
    },

    setState: ({ stateDiff, prevState, newState, actionName, userArgs }) => {
      const { eventArgs } = userArgs

      this.setState(
        stateDiff, 
        () =>  _.invoke(this.props, 'onOpenChange', eventArgs, newState))
    },
  })
}
```

### API example (full)
```js
const stateManager = StateManager.create({
    // ----------------------------------
    // FRAMEWORK-AGNOSTIC actions
    // ----------------------------------
    actions: {
      open: it => {
        it.setState({ open: true })
      },
      ...
      toggle: it => {
        const state = it.getState()

        if (state.open) {
          it.close()
        } else {
          it.open()
        }
      },
    },

    // ----------------------------------
    // BINDINGS TO REACT
    // ----------------------------------
    bindings: {
      /* returns component's state */
      getState: () => {
        return { ...this.state, ...this.props }    // AUTOCONTROLLED state logic
      },

      setState: ({ stateDiff, prevState, newState, actionName, userArgs }) => {
        const { eventArgs } = userArgs
        ...
      },
    },
  })
```

### Extensibility points
Please, note that both action's **middleware**, as well as **hook**/**sideEffects** (`willOpen`, `didOpen`) concepts could be introduced as part of suggested implementation in case if necessary. Currently `willSetState` filter method (that may act like a filter for updated state) is introduced as an example.

Please, feel free to test `Popup` component that will run on top of the proposed approach - while some edge cases are not handled (like setting `null` or `undefined` prop values explicitly), the examples still should be able to serve as proof of concept.

### Disclaimer
Please, note that this code is not polished and may have sloppy implementations at non-critical places (for example, there is no code that will map component's prop names to state manager's ones). However, this code will never get to the Stardust repo in the state it is now - its main goal is just to deliver an essence of the idea. Please, use pragmatic judgement while estimating exact implementation strategy chosen - at some places where it is not critical I've deliberately chosen sloppy approaches. At the same time, in case if some code snippet is seen as a critical issue, please, feel free to let me know about that, to open related discussion. 

Thank you :) 

https://codesandbox.io/s/7m8vw1pro0


src/components/Popup/Popup.tsx
```diff
@@ -66,6 +68,36 @@ export default class Popup extends AutoControlledComponent<Extendable<PopupProps
...
+    // willSetState: ({ prevState, stateDiff, args, actionName }) => {
+    //   return excludeControlledProps(stateDiff, this.props)
+    // },
+
+    setState: ({ stateDiff, prevState, newState, actionName, userArgs }) => {
+      const { eventArgs } = userArgs
+
+      // although state doesn't precisely reflect autocontrolled 'state',
+      // the state of stateManager is always properly defined
+      this.setState(stateDiff, () => _.invoke(this.props, 'onOpenChange', eventArgs, newState))
```
kuzhelov: `actionName` could be tested as well, to guarantee proper event will be triggered

kuzhelov:
needs couple of changes to simplify abstractions exposed to client - at the same time, the idea will stay the same

layershifter (commented):
I want to make a step back there.

---

As I remember we're trying to follow Redux's ideas. And if we will check then we will find that:

> React bindings for Redux separate presentational components from container components. This approach can make your app easier to understand and allow you to more easily reuse components. 

In the same time we're trying to **couple** our components with state. My opinion that a component should not know anything about a state and should follow the idea of _Presentational Components_.

I see there three options there:
- HOC (`react-redux`)
- render props ([`constate`](https://github.com/diegohaz/constate))
- hooks

_We can't use the last until they will have the stable API, however, they look the most suitable._

---

It doesn't matter what solution we will choose, but in the React world we **should** follow one way data-flow and have a single SSOT. It should be _props/state from our manager_ or _React's state_. We can't use both:

```js
// We have this
this.stateManager.getState().open
// But I CAN use this because it exists on this level
this.state.open
```

We also should follow community ideas:

```jsx
// Looks clear and native for React users
<Container initialState={initialState} actions={actions} />

// Doesn't look neither clear nor common
class Component {
  private stateManager = popupStateManager.withBindings({
    getState: () => {},
    setState: ({ stateDiff, prevState, newState, actionName, userArgs }) => {}
  }
}
```

The center points of our idea are `actions` and `state` and only them should be visible to our users. I don't have a better idea now, but in _my opinion_ we need to create a React's friendly binding or just use existing solutions:

https://github.com/FormidableLabs/freactal
https://github.com/MicheleBertoli/react-automata

I think that may be we need try to implement one of our components with these libraries and just pick the best.

kuzhelov:
@layershifter 

Thanks a lot for expressing your thoughts on that. Let me, please, address your comments below.

---------

> React bindings for Redux separate presentational components from container components. This approach can make your app easier to understand and allow you to more easily reuse components.

While I would agree that we would like to get best from Redux ideas, at the same time it is very important to start with fundamental building blocks first - and this fundamental block is not the connectors. Key thing that we would like to achieve is to **introduce an API that would be framework agnostic** - so that there will be a **guarantee that it could be consumed from an app based on any JS-based framework/library**.

After that, as a next step, we will be able to introduce bindings in any way that would be most suitable for specific library - if it is React, for example, surely we might opt to HOC (or maybe even Hooks) - but this comes as a next, higher level of abstraction.

Thus, I don't see any problems and contradictions of proposed approach with those points you've expressed here (other than provide React connector in addition).

------

> It doesn't matter what solution we will choose, but in the React world we should follow one way data-flow and have a single SSOT.

I agree here - but if
- basic functionality will be introduced, with the API that will be framework agnostic
- React connector will be implemented on top of that (HOC, or I am rather inclined to Hooks-based) in a way that it will serve a single source of truth for component's state

then there won't be any concerns here.

------

To conclude, I see that **your comment is rather focused on what React connector should look like** - I totally agree with the points expressed there. At the same time, another (and, probably, main one) goal of the PR is to **introduce state management implementation that would be 'connectable' to any library that is JS-based**, so that client will be able to consume state management logic from different app's stack - and to me it clearly seems that provided low-level API is able to address all your concerns by just introducing another approach to React connector.

layershifter:
> To conclude, I see that **your comment is rather focused on what React connector should look like** - I totally agree with the points expressed there. At the same time, another (and, probably, main one) goal of the PR is to **introduce state management implementation that would be 'connectable' to any library that is JS-based**, so that client will be able to consume state management logic from different app's stack - and to me it clearly seems that provided low-level API is able to address all your concerns by just introducing another approach to React connector.

Yes, you're right. I'm claiming about React bindings, they should be implemented in another way and use React's patterns. The current implementation doesn't look well for me.

I'm afraid that in pursuit of the framework agnostic implementation we will not do a really good API neither for React. nor for Vue and others.

----

As I see, RFC for Hooks was merged https://github.com/reactjs/rfcs/pull/68, we can discuss this today on sync meeting 👍 

We also should check that hooks will work with our `createComponent()`:
> Only call Hooks at the top level. Don’t call Hooks inside loops, conditions, or nested functions.

layershifter (changes requested):
As this PR was marked as ready for review, I'm going to share my opinion.

---

```ts
  private stateManager = connect(getPopupStateManager()).toAutoControlled(this, {
    open: 'onOpenChange',
  })
  
  private isControlled() {
    return this.props.open !== undefined
  }

  // 

  this.updateOutsideClickSubscription(this.stateManager.getState().open)
```

This API still contains all issues that I described before, https://github.com/stardust-ui/react/pull/466#pullrequestreview-177690901:

It couples a component's state and a state of its manager, both of these states can be accessed by a user:

```js
this.stateManager.getState().open // vs
this.state.open
```

I don't understand how we can use these bindings inside functional component, i.e. `createComponent()`.

---

Yes, `react-redux` couples the state, too, in [Provider.js](https://github.com/reduxjs/react-redux/blob/master/src/components/Provider.js), but:
- Provider doesn't know anything about action handlers
- component that **uses** state doesn't know anything about it and _can_ be used without it

---

```ts
_.invoke(componentRef.props, changeHandlerName, eventArgs, newState)
```

What is purpose of calling component's handlers **in** the state manager? There is `mapDispatchToProps` that solves this in `react-redux`.

We can do something similar:

```js
mapActionsToProps: (actions) => {
  open: (e, someOtherProp) => (currentProps) => {
    _.invoke(props, 'onOpen', e, { ...currentProps, open: true })
    actions.open()
  }
}
```

kuzhelov:
@layershifter 

> It couples a component's state and a state of its manager, both of these states can be accessed by a user

Lets be specific here - its not `coupling`, but, essentially, making state manager being responsible for **_making all the necessary adjustment's to component's state when action is invoked_**. 

So, essentially, your objection is not about 'coupling', but about the fact that state could be read from different places by the client. However, here I would object with the following arguments:
- this necessity for reading from state manager (instead of component's `state` directly) is relevant for autocontrolled components only
  - similar to AutoControlled component, where it is also necessary for client to remember about `trySetState()` function (instead of `setState()`)
  - more than that, **with proposed approach it is easily possible to introduce connector for AutoControlled components, so that the following invariants will always hold** - and this will eliminate your concerns about 'where state should be consumed from', as it might be consumed from any source then:
 ```jsx
// code of Foo which extends AutoControlledComponent 
this.stateManager = connect(popupStateManager).toAutoControlled(this)

// and this will ensure the following conditions are always true:

// first
if (this.props.open !== undefined) {
  this.state.open === this.props.open
}

// second
this.stateManager.getState().open === this.state.open
```
  - the problem is that we are not going to rely on `AutoControlled` component, as it was previously discussed, because of its `componentWillUpdate` hook (which is used to keep `props` and `state` in sync) - and this was the reason I've not introduced it yet
-------

### Use HOC as a connector

If I would summarise your comments, those, in their essence, reduce to the fact that HOC-based approach is suggested for state management in React (or Hooks-based, will leave this for now) - and **given that there is no objections about framework-agnostic part of the PR, this suggests me that here we have an agreement there, right**?

Moving to HOC-based connector - I can introduce this as well, to provide an idea of what it will look like (essentially it is about hoisting the logic of current connector to this HOC component) - key thing is that I do not see _**any**_ limitations for that. **Actually, let me introduce that**, to conclude this topic.

------

### Hooks-based connector

Might introduce this as well. The reason I've not done that and have intentionally used older React API is because to me it seems like a **good test for introduced state management functionality, in terms of whether it is really agnostic of the web technology or not: _if it wouldn't be possible to introduce any approach for connecting it with 'older' React, then this test is failed_**.

Thus, I may decide to skip this moment for now, as if we will resolve ongoing concerns, we will be absolutely assured that Hooks-based connector logic also can be implemented.

levithomason:
Looks much simpler, let's merge this into `/state/...` in the repo.  Add HTML, customElement, Vue, and React examples.  We'll iterate on it there. 👍

layershifter:
Fixed by #1905.

PR #469, opened 2018-11-13
feat(FocusZone): Force focus to first element when container receives focus

sophieH29:
This PR extends FocusZone API and adds prop ```shouldFocusFirstElementWhenReceivedFocus``` which when ```true```, forces focus to a first focusable element when a container of the FocusZone receives focus.
There are a few cases when this feature is very important to bring an appropriate experience.

Let's consider next case:

```js
   <FocusZoneA>
        <input/>
        <button/>
           <FocusZoneB data-is-focusable="true">
             <input/>
             <button/>        
          </FocusZoneB>
   </FocusZoneA>
```

When we navigate with keys inside the ```<FocusZoneA>```, focus will go to ```<FocusZoneB>```, but instead of focusing whole FocusZoneB and show focus ring around it, we want the focus to be forced to the first focusable element inside ```<FocusZoneB>```.

Current FocusZone implementation constraints such need, so that's why the API is extended by an optional prop, which can improve the experience for specific cases.


src/lib/accessibility/FocusZone/FocusZone.types.ts
```diff
@@ -53,6 +53,11 @@ export interface FocusZoneProps extends React.HTMLAttributes<HTMLElement | Focus
    */
   shouldFocusOnMount?: boolean
 
+  /**
+   * If a first focusable element should be force focused when FocusZone container receives focus.
+   */
+  shouldFocusFirstElementWhenReceivedFocus?: boolean
```
hughreeling, line 59: What's default for this?  Seems like most times this would be desirable?

sophieH29: The default is false. It is needed in many times, especially when we use nested focus zones, but I wouldn't say the most times. The most times actually we don't need it, for e.g. when we have a ```Chat```, which is a ```FocusZone``` itself, and ```chatMesssages```, which are the nested FocusZones. But when we navigate with Up/Down arrow keys, we still expect the focus goes to the message (FocusZone container) unless we press, for e.g., Enter key to go inside message and focus first focusable element there.

jurokapsiar: I think the case of having focus zone in focus zone is rare enough that we can say consumers need to think about it. I'm fine with having it set to false.

jurokapsiar (approved):

test/specs/lib/FocusZone-test.tsx
```diff
@@ -1110,4 +1110,71 @@ describe('FocusZone', () => {
...
+      keyboardKey.getCode(e) === keyboardKey.Enter
+    const isFocusableProperty = { [IS_FOCUSABLE_ATTRIBUTE]: true }
+
+    const component = ReactTestUtils.renderIntoDocument(
+      <div {...{ onFocusCapture: onFocus }}>
+        <FocusZone
+          direction={FocusZoneDirection.horizontal}
+          shouldEnterInnerZone={isInnerZoneKeystroke}
+        >
+          <button className="a">a</button>
```
kuzhelov: nit: seems that using 'id' attribute would better reflect tests semantics, as for selectors there is **always** one element expected to be returned

test/specs/lib/FocusZone-test.tsx
```diff
@@ -1110,4 +1110,71 @@ describe('FocusZone', () => {
...
+      },
+    })
+
+    // Focus the first button.
+    ReactTestUtils.Simulate.focus(buttonA)
+    expect(lastFocusedElement).toBe(buttonA)
+
+    ReactTestUtils.Simulate.keyDown(focusZone, { which: keyboardKey.ArrowRight })
+    // Focus goes to FocusZone container, which forces focus to first focusable element - buttonB
+    expect(lastFocusedElement).toBe(buttonB)
```
kuzhelov, line 1174: to be sure, what would be the default behavior if this `focusFirstElement` prop won't be set? To me it seems that the following DOM tree is rendered:

```html
<div data='top focus zone container'>
  <button id='a' />

  <div data='focus zone container of b'>
     <button id='b' />
  </div>
</div>
```

It seems that even without this prop (`focusOnFirstElement` for `b` focus zone) behavior will be the same as test suggests - most probably missing something, could you clarify please? Is the difference in the fact that _by default_ focus won't propagate into `b` focus zone? 

It would be, actually, great if you could clarify focus behavior with this prop being unset, and with this one being set, so that it would be evident where the difference is, thanks!

bmdalex (approved):

test/specs/lib/FocusZone-test.tsx
sophieH29: fair enough, but I think it easier to read tests when it is written

```document.querySelector('.a')```

vs 

```document.querySelector('[id="a"]')```

(```getElementById``` does not exists on ```HTMLElement```)

I think we can be safe here as it is just a plain example, so I would leave it as it is.

Comment (kuzhelov) [reply test/specs/lib/FocusZone-test.tsx]: `document.querySelector('#a')` - this is a selector that should be used in that case. With this one there is no much difference for readability, but at the same time semantics will be expressed much cleaner.

> it is just a plain example, ..

Actually, this is the reason I am worrying about semantics - would it be test or other example, it should clearly suggest its semantics as it is _meant to be read_, as it documents expected behavior. If there is class selector used, I am trying to search for another buttons provided in the example's code - in contrast, with an id selector I am pretty sure that this button is _the only_ one that might be selected with related expression.

File: test/specs/lib/FocusZone-test.tsx (hunk 3 above)
Comment (sophieH29) [reply test/specs/lib/FocusZone-test.tsx:1174]: Current ```FocusZone``` implementation constraints propagating focuses across nested focus zones. If you need to go inside some nested focus zone, you have to use a prop ```shouldEnterInnerZone```, for e.g.
```js
shouldEnterInnerZone = (e:): boolean =>
      keyboardKey.getCode(e) === keyboardKey.Enter
```
so specify some different key which you'll use to go inside focus zone, so focus elements inside it.

The following tree which is rendered is like you mentioned, but the only difference that inner FocusZone has ```data-is-focusable="true"``` attribute.

```js
<div data='top focus zone container'>
  <button id='a' />

  <div data='focus zone container of b'
          data-is-focusable="true"
          shouldEnterInnerZone="Enter key">
     <button id='b' />
  </div>

   <button id='c' />
</div>
```

Here is a user case without ```focusFirstElement``` : you have a top focus zone container, across which you navigate with arrows Left / Right.
1. you press```->```
2. button ```b``` gets focus
3. you press ```->```
4. Sub focus zone ```focus zone container of b``` gets focus
    4.1 at this point you can press either ```Enter``` to go inside this FocusZone or ```->``` to focus the next element button ```c```
5. you press ```Enter``` 
6. Focus goes inside sub focuszone, so now you can navigate to button ```b```


Case with ```focusFirstElement``` set to ```true```
1. you press```->```
2. button ```b``` gets focus
3. you press ```->```
4. Sub focus zone ```focus zone container of b``` gets focus and forces focus to first element
5.  so button ```b``` gets focus
6. you press ```->``` and you go out of this sub focus zone (in case it doesn't handle navigation with arrow right/left) 
7. button ```c``` get focus

Please, let me know if it got more clear fo you know

kuzhelov: > 2. button `b` gets focus

you've meant "button `a`" in all the cases described, right? If so, the explanation is clear, thank you :)

kuzhelov (approved):

kuzhelov (approved):
would still like `class` selector being replaced with `id` selector for tests, reasons provided.

test/specs/lib/FocusZone-test.tsx
sophieH29: oh, yeah, I meant button ```a```, sorrryyy. Tried to explain but confused even more 😄

PR #470, opened 2018-11-13
fix(factories): styles defined as func are not working

### LINKED ISSUE #415: Styles defined as functions in shorthand are not working
### Steps
Define some styles as a function in any shorthand of the components. For example, in the Button, define the icon shorthand with some styles in form of a function.

```
<Button 
  type="primary" 
  icon={{
    name: "book", 
    styles: () => ({margin:'10px'}),
  }} 
  content="Click me before" 
  iconPosition="before" />
```

### Expected Result
The styles should be applied for the shorthand.

### Actual Result
They are not applied. The only way that those are applied is if the styles are defined as object 💣

### Version
0.11.0

mnajdova:
## TODO
- [x] update CHANGELOG

--------------

This PR fixes https://github.com/stardust-ui/react/issues/415

### DIFF
--- a/src/lib/factories.tsx
+++ b/src/lib/factories.tsx
@@ -8,7 +8,7 @@ import {
   ShorthandValue,
   Props,
 } from '../../types/utils'
-import callable from './callable'
+import { mergeStyles } from './mergeThemes'
 
 interface CreateShorthandOptions {
   /** Override the default render implementation. */
@@ -107,13 +107,7 @@ export function createShorthand(
 
   // Merge styles
   if (defaultProps.styles || overrideProps.styles || usersProps.styles) {
-    props.styles = (...args) => {
-      return _.merge(
-        callable(defaultProps.styles)(...args),
-        callable(usersProps.styles)(...args),
-        callable(overrideProps.styles)(...args),
-      )
-    }
+    props.styles = mergeStyles(defaultProps.styles, usersProps.styles, overrideProps.styles)
   }
 
   // ----------------------------------------
--- a/src/lib/factories.tsx
+++ b/src/lib/factories.tsx
@@ -8,6 +8,7 @@ import {
   ShorthandValue,
   Props,
 } from '../../types/utils'
+import callable from './callable'
 
 interface CreateShorthandOptions {
   /** Override the default render implementation. */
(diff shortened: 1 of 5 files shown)


test/specs/lib/factories-test.tsx
```diff
@@ -264,7 +265,72 @@ describe('factories', () => {
...
+              ':hover': {
+                position: 'keep',
+                color: 'blue',
+              },
+            })
+          },
+        })
+      })
+
+      test('deep merges styles prop as function onto defaultProps styles', () => {
```
kuzhelov, line 280: `into`? actually, was surprised to discover existence of `onto` word for myself :)

test/specs/lib/factories-test.tsx
```diff
@@ -264,7 +265,72 @@ describe('factories', () => {
...
+              ':hover': { color: 'blue' },
+            })
+          },
+        })
+      })
+
+      test('deep merges overrideProps styles as function onto styles prop', () => {
+        expect.assertions(1)
+
+        const overrideProps = {
```
kuzhelov, line 309: it might be a bit easier to reason about expected results if these two `props` objects will be swapped

src/lib/factories.tsx
```diff
@@ -106,7 +107,13 @@ export function createShorthand(
 
   // Merge styles
   if (defaultProps.styles || overrideProps.styles || usersProps.styles) {
-    props.styles = _.merge(defaultProps.styles, usersProps.styles, overrideProps.styles)
+    props.styles = (...args) => {
+      return _.merge(
```
kuzhelov: don't we want to have `mergeStyles` function that would encapsulate this logic? Similar semantics is expressed in the `mergeThemes` file:

```js
// mergeComponentStyles() {
...
return _.merge(callable(originalTarget)(styleParam), callable(originalSource)(styleParam))
```

maybe we could encapsulate and reuse this logic in the following way:

```ts
mergeStyles(target: ComponentSlotStyle, ...styles: ComponentSlotStyle[]) {
  ...
}
```

kuzhelov (approved):
it would be great if we'll be able to encapsulate this logic for merging styles, to prevent similar mistakes from happening in future

miroslavstastny (approved):

src/lib/factories.tsx
mnajdova: Agreed, will encapsulate this logic.

mnajdova: Introduced, please review again.

test/specs/lib/factories-test.tsx
mnajdova: I would rather leave this tests the same as the previous ones for consistency, as the only difference is changing the objects to functions.

PR #471, opened 2018-11-13
feat(prototype): connect custom components

kuzhelov:
This PR introduces alternative approach to factory-based method of integrating third-party components. 

Please, note that this PR is in prototype phase, its only goal is just to deliver the idea - and while not the same set of functionality is introduced _for now_ to this PR to be on pair with factory-based approach (only styles and accessibility aspects are captured now), all the necessary logic bits could be introduced if necessary. In other words, this approach is not worse in terms of 'which client cases may be addressed'.

### Benefits

Main benefit of the suggested approach - it **makes it very easy to connect existing components to Stardust, no rewrite/restructuring of existing component's logic is necessary (in contrast to factory-based one)**. Essentially, the only thing required from the component's author is to suggest where evaluated Stardust bits (like styles/classes, accessibility attributes, etc) should be consumed in component's logic. 'Connect' function follows the Redux approach to pass Redux data to component as props - this aids seamless integration process. 

### Cons

Speaking of cons - this approach **requires additional wrapper component in comparison to factory-based approach - 2 vs 1 wrappers** will be rendered for connected components (one is necessary for consuming theme context data). At the same time, given the benefits and the fact that React moves towards alleviate the issue related to wrapper components, to me it seems that benefits clearly outweighs the 'costs'.

### Example

To provide an example that might emphasize the issues of factory-based approach that client might struggle with, I've decided to search for some component that will be 
- stateful
- would have a fair amount of private logic (instance methods)
- would use React lifetime hooks

Have decided to take a DateTimePicker component for that from here: https://github.com/wojtekmaj/react-datetime-picker/blob/master/src/DateTimePicker.jsx. Suppose that we would like to introduce styles based on (props + **state**) of this component. 

Let me provide the only changes necessary to integrate Stardust styles to this component:
```jsx
/// map Stardust evaluated classes to some prop
export connect({
  mapToComponentProps: ({ renderStyles }) => ({ renderStardustStyles: renderStyles })
}, DateTimePicker)

// ....

class DateTimePicker extends React.Component {
   ...
   render() {
      // 'renderStardustStyles' is properly bound to right fela renderer
      const { className, renderStardustStyles } = this.props

      // the intent is to use state + props object to evaluate styles
      const stardustClasses = renderStardustStyles.root({ ...this.state, ...this.props })

      <div 
          className={mergeClassNames(
            baseClassName,
            `${baseClassName}--${isCalendarOpen || isClockOpen ? 'open' : 'closed'}`,
            `${baseClassName}--${disabled ? 'disabled' : 'enabled'}`,
            className,
            stardustClasses  // <------------ use classes here
        )}>
       ...
      </div>
   }
}
```

That's it - all the steps are absolutely intuitive. With factory-based approach it will require more work and, most probably, will be less intuitive to the client how the same goal could be achieved.

### DIFF
--- a/docs/src/examples/components/Button/Types/ButtonExample.shorthand.tsx
+++ b/docs/src/examples/components/Button/Types/ButtonExample.shorthand.tsx
@@ -24,6 +24,8 @@ const ButtonExample = () => (
     <>
       <div>
         <MyButton />
+
+        {/* 'accessibility' option for connect() could be introduced for that */}
         <MyAccessibleButton accessibility={buttonBehavior} disabled>
           I have aria-disabled=true
         </MyAccessibleButton>
--- a/docs/src/examples/components/Button/Types/ButtonExample.shorthand.tsx
+++ b/docs/src/examples/components/Button/Types/ButtonExample.shorthand.tsx
@@ -30,7 +30,8 @@ const ButtonExample = () => (
       <div>
         <MyButton />
 
-        {/* 'accessibility' option for connect() could be introduced for that */}
+        {/* 'accessibility' option for connect() could be introduced,
+        so that it won't be necessary to define attribute here. */}
         <MyAccessibleButton accessibility={buttonBehavior} disabled>
           I have aria-disabled=true
         </MyAccessibleButton>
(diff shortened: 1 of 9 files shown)


kuzhelov:
agreed to move with factory-based approach, it covers the same set of cases and might be superior in cases when Stardust component will be introduced from scratch.

PR #473, opened 2018-11-14
chore(Editor): make editor fully functional

layershifter:
This PR uses `react-source-render` and Babel to render examples from source.
This allows to resolve our existing problems with the live editing process.

----

### Before

_Blank page on fail._

before

### After 

after


docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -31,14 +31,10 @@ interface ComponentExampleState {
...
-  exampleElement?: JSX.Element
-  handleMouseLeave?: () => void
-  handleMouseMove?: () => void
+  handleMouseLeave: () => void
+  handleMouseMove: () => void
   sourceCode: string
-  markup: string
-  error?: string
   showCode: boolean
-  showHTML: boolean
```
layershifter, line 41: It's very confusing, this variable wasn't used at all

docs/src/components/Playground/renderConfig.ts
```diff
@@ -0,0 +1,25 @@
+import * as _ from 'lodash'
+import * as React from 'react'
+import * as Stardust from '@stardust-ui/react'
+
+export const babelConfig = {
+  plugins: ['proposal-class-properties', ['transform-typescript', { isTSX: true }]],
```
layershifter, line 6: It's strange, but `typescript`'s preset is registered, but not working 🤔 
Will fill an issue to Babel's repo.

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -125,23 +117,22 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
     }
     const { themeName } = nextProps
     if (this.state.themeName !== themeName) {
-      this.setState({ themeName }, this.renderSourceCode)
```
kuzhelov, line 128: 👍

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -203,26 +194,12 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
...
-
   private handleShowRtlClick = e => {
     e.preventDefault()
 
     const { showRtl } = this.state
 
-    this.setState({ showRtl: !showRtl }, () => {
-      this.renderSourceCode()
-    })
+    this.setState({ showRtl: !showRtl })
```
kuzhelov: would suggest to do it safer, by using state callback arg

layershifter: Agree. I'm planning to refactor this stuff to use a dedicated context, this will simplify `ComponentExample` and `ComponentControls`.

For this PR I want to focus on getting working editor and make changes small as possible :)

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -297,40 +289,13 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
...
-
   private handleKnobChange = knobs => {
-    this.setState(
-      prevState => ({
-        knobs: {
-          ...prevState.knobs,
-          ...knobs,
-        },
-      }),
-      this.renderSourceCode,
```
kuzhelov, line 332: 👍

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
mnajdova: I believe this is some left over, as we are showing the HTML alongside with the code editing... My proposal is to just remove it..

layershifter: Yes, I have already done this 😁

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -688,35 +636,43 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
...
+                      color: siteVariables.bodyColor,
+                      backgroundColor: siteVariables.bodyBackground,
+                      ...(showTransparent && {
+                        backgroundImage:
+                          'url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs+9AAAAKUlEQVQoU2NkYGAwZkAD////RxdiYBwKCv///4/hGUZGkNNRAeMQUAgAtxof+nLDzyUAAAAASUVORK5CYII=")',
+                        backgroundRepeat: 'repeat',
+                      }),
+                    }}
+                  >
+                    {this.renderExampleFromCode()}
```
mnajdova, line 662: 👍 much cleaner!

docs/src/components/Playground/renderConfig.ts
mnajdova: So, are we evaluating now the examples as js code?

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -494,51 +441,58 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
...
-            onOutsideClick={this.handleShowHTMLInactive}
-            value={beautifiedHTML}
-          />
-        </div>
-      </div>
+      <SourceRender.Consumer>
+        {({ markup }) => {
+          // add new lines between almost all adjacent elements
+          // moves inline elements to their own line
+          const preFormattedHTML = markup.replace(/><(?!\/i|\/label|\/span|option)/g, '>\n<')
```
kuzhelov: `\/option` (also with slash?) Or this is intended to ensure that `<option>` elements will be rendered on the same line?

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -494,51 +441,58 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
...
   private renderError = () => {
-    const { error } = this.state
-
-    if (!error) return null
-
     return (
-      <Segment size="small" color="red" basic inverted padded secondary>
-        <pre>{error}</pre>
-      </Segment>
+      <SourceRender.Consumer>
```
kuzhelov, line 443: not sure whether context consumer model is the one that fits this case, because as an alternative we might think about storing rendered result in a state and render by consuming stored value. Not sure if this context-driven indirection is necessary here

kuzhelov (commented):
changes look absolutely good to me, just left couple of questions to refine

docs/src/components/Playground/renderConfig.ts
kuzhelov: yes, at least till the moment when this issue will be fixed :)

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
layershifter: Context allows to avoid the storage of render related stuff in the state of `ComponentExample` and avoid cycling updates like this: `render ComponentExample` => `render source` =>`update state` => `render ComponentExample` (again).

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
layershifter: Not sure, this code was moved from other place on this file.

We should to switch to Prettier for formatting, #535

kuzhelov: ok, but should we consider to fix this thing then? Currently meaning of this expression is very confusing for `<option>` element, and, thus, I would rather suggest to clarify that and either introduce it to the list of self-closed elements to format, or delete from this regex

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
kuzhelov: this will be a small step - just refactor it to the callback and ensure that, before announced changes will be introduced, it will still work correctly. Frankly, don't understand why we cannot eliminate this problem now and improve on mechanism later

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -297,40 +289,13 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
     )
   }
 
-  private renderSourceCode = _.debounce(() => {
```
kuzhelov, line 300: no debouncing anymore?

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
kuzhelov: sorry, cannot completely understand how it is different with the context.

Lets, for the sake fo simplicity, consider the following case: we have _code pane_ and _render pane_, where code is mapped to rendered DOM.

Also, suppose that there is `ComponentExample` component who manages these two.

Initially code and rendered DOM are rendered at the corresponding panes. After that we start editing code:

### State-based approach

The following will happen if `ComponentExample`'s `state` will be used:

```
`RENDER PANE`
-----------------
`CODE PANE` -> content changed ->`code` evaluated to `state.renderedTree`, RERENDER
```

### Context-based approach

So, one rerendering for state (two if this process will be debounced) - no extra unexpected cycles. How it is different from context-based way? What we will have is the following (under assumption that debouncing is removed, as we have it now):

```
`RENDER PANE`
-----------------
`CODE PANE` -> content changed -> evaluated to `state.renderedTree`, RERENDER with updated context provider
```

So, still one update, but why do we need context-based indirection?

-----------

Please, help me settle these things in my head. Thank you!

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
layershifter: I removed `option` from regex

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
layershifter: I refactored it to use `prevState` 👍

docs/src/components/Playground/renderConfig.ts
layershifter: Nope, all TS transforms are applied. I mentioned that use there the plugin:

```
plugins: [['transform-typescript', { isTSX: true }]],
```

Instead of the preset:

```
presets: ['typescript'],
```

However, **everything** is works.

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
layershifter: Yes, we don't need now it fact. Performance is good enough.

kuzhelov (approved):

docs/src/utils/evalTypeScript.ts
```diff
@@ -1,111 +0,0 @@
...
-/**
- * Evaluates TypeScript source code by transpiling it to JavaScript and then
- * executing it within our sandbox, which provides a suite of global variables.
- */
-const evalTypeScript = (sourceCode): any => {
-  const transpiledCode = transpileTypeScript(sourceCode)
-  return execute(transpiledCode, {
-    REACT: React,
-    REACT_DOM: ReactDOM,
-    STARDUST_UI_REACT: Stardust,
```
mnajdova, line 106: I can't seem to find where are we handling this now...

docs/src/components/Playground/renderConfig.ts
```diff
@@ -0,0 +1,25 @@
...
+      },
+    ],
+  ],
+}
+
+const imports = {
+  '@stardust-ui/react': Stardust,
+  lodash: _,
+  react: React,
+}
```
mnajdova, line 23: We should add react-dom here

PR #478, opened 2018-11-15
feat(Icon): Process Teams icons for use in Stardust

codepretty:
Converted the majoirty of svg icons we have in Teams today to be consumed in Stardust. I created a separate folder for these icons **ProcessedIcons**. As new icons are needed in Stardust, we can move them from ProcessedIcons to icons. This way we will only add the icons we need to the Teams theme.

In order to find the icon you want to use, I created a new page for viewing all the processed icons.
image

There are 2 scripts I used for converting Icons:
1. stardust-icons.sh - this runs through a directory of Angular Teams SVGs and converts to .tsx files ready to be consumed by Stardust.
2. create-index.sh - creates a new index.ts file for the icons based on the files in its directory.

**Question:**
1. In the processed icon viewer, teams theme icons are included in that view. How would I only show the processed icons? Thanks @mnajdova! I fixed this using your suggestion below.


src/themes/teams/components/Icon/svg/ProcessedIcons/icons-accept.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react'
+import { TeamsSvgIconSpec } from '../types'
+
+export default {
+  icon: ({ classes }) => (
+    <svg role="presentation" focusable="false" viewBox="8 8 16 16" className={classes.svg}>
+      <g>
+        <path
+          className={classes.outlinePart}
```
miroslavstastny, line 9: In order to support changing visible paths on parent element's hover (#364), I will need to add static class names to both outline and filled paths. I plan to change this to something like:
```jsx
className={iconOutlinePart(classes)}
```
Please let me know if that works for you, I will do that in a separate PR once this one is merged.

mnajdova:
@codepretty regarding your question. As in the prototypes we are merging the providers, you will always end up with all icons there (stardust + processed). My suggestion would be, when you are exporting the processed, add some prefix in the icons' names (processed or something like that), and then we you are looping over all icons, you can display just the processed ones. Let me know if this makes sense to you, or if you need help with implementing it.

src/themes/teams/components/Icon/svg/ProcessedIcons/icons-accept.tsx
codepretty: As long as the filledPart & outlinePart in IconStyles still work, then we should be good.

src/themes/teams/components/Icon/svg/ProcessedIcons/icons-info.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+import { TeamsSvgIconSpec } from '../types'
+
+export default {
+  icon: ({ classes }) => (
+    <svg role="presentation" focusable="false" viewBox="8 8 16 16" className={classes.svg}>
```
alinais, line 6: What should this class do? Should we define it in the styles so that the svg like in other examples?
```
  styles: { 
    svg: ({ variables }) => { 
      return { 
        fill: variables.color, 
      } 
    }, 
  },
```
I see it in all the other svg files. I think we'll need to have the same patterns for all the svgs.

codepretty: It's a good question. It was already there when I started importing icons, but I don't think we have a use case for it yet/anymore. In the styles, I have this 
```
  svg: ({ props: { size, disabled }, variables: v }): ICSSInJSStyle => {
    return {
      fill: getIconColor(v.color),

      ...(disabled && {
        fill: v.disabledColor,
      }),

      ...getSvgStyle('svg'),
    }
  },
```

So, I think we'd only use it if the icon itself is overriding it's fill color. However, the cases I've seen for overriding would happen on paths, not the entire svg. Do you see a reason for it or should I just remove it from all the svg files?

kuzhelov: @codepretty, yes, you are absolutely right - there is no need to change anything, as styles for this part are defined in general place, `iconStyles.ts` - and those are currently reused by all the icons. So, provided implementation (and approach taken) looks absolutely fine to me 👍

docs/src/prototypes/IconViewer/index.tsx
```diff
@@ -0,0 +1,67 @@
+import * as React from 'react'
+import { Provider, Grid, Divider, Header, Icon } from '@stardust-ui/react'
+import { processedIconsTheme } from 'src/themes/teams/iconTest'
```
kuzhelov: would suggest to rename `iconTest` to something that would be more expressive, like `withProcessedIcons` (and it would work nice, as path for this file will provide all the necessary context: `src/themes/teams/withProcessedIcons`). As a next step, we might consider to use default export in `withProcessedIcons.ts` to export corresponding theme, so that final solution would be:

```js
// in withProcessedIcons.ts
const theme = { ... }
export default theme

.....

// here, in this file
import themeWithProcessedIcons from 'src/themes/teams/withProcessedIcons'
```

src/themes/teams/components/Icon/svg/ProcessedIcons/icons-onenote-section.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+import { TeamsSvgIconSpec } from '../types'
+
+export default {
+  icon: ({ classes }) => (
+    <svg role="presentation" focusable="false" viewBox="8 8 16 16">
+      <path fill="none" d="M4 4h24v24H4z" />
+      <path
+        d="M19 25h-1v-1a.92.92 0 0 0-1-1h-2.5a2.57 2.57 0 0 1-2.5-2.5v-9A2.57 2.57 0 0 1 14.5 9H17a.92.92 0 0 0 1-1V7h1z"
+        fill-rule="evenodd"
```
kuzhelov: note, this should be defined as `fillRule`, no dash is allowed - similar to `border-color` vs `borderColor`. Currently this results in runtime error.

src/themes/teams/components/Icon/svg/ProcessedIcons/icons-settings.tsx
```diff
@@ -0,0 +1,19 @@
...
+      />
+      <path
+        className={classes.filledPart}
+        fill-rule="evenodd"
```
kuzhelov: note, this should be defined as `fillRule`, no dash is allowed - similar to `border-color` vs `borderColor`. Currently this results in runtime error

docs/src/prototypes/IconViewer/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+          <Provider.Consumer
+            render={theme => (
+              <div>
+                <div>
+                  <Divider>
+                    <Header as="h3" content="Regular" />
+                  </Divider>
+                  <Grid columns={4} style={{ textAlign: 'center' }}>
+                    {Object.keys(theme.icons)
+                      .filter(name => name.indexOf('processedIcons_') !== -1)
```
kuzhelov: would suggest to declare constant, like

```js
const processedIconsNamePrefix = 'processedIcons_'
```

and reuse it, just to reduce duplication of this string literal

kuzhelov: it would be great to ensure that icons are sorted alphabetically, like

```js
const processedIconNamePrefix = 'processedIcons_'

Object.keys(theme.icons)
  .filter(name => name.startsWith(processedIconNamePrefix))   // <-- 'startsWith' would be more expressive 
  .map(name => name.replace(processedIconNamePrefix, ''))
  .sort()
  .map(name => <div>...</div>)

File: src/themes/teams/components/Icon/svg/ProcessedIcons/create-index.sh [hunk 7]
```diff
@@ -0,0 +1,30 @@
...
+
+  Name=${UnprocessedName//-/}
+  echo "Name is: $Name"
+
+  # REPLACE INSTANCES IN IMPORT & EXPORT TEMPLATES
+  ImportTest=$(echo "$importTemplate" | sed "s/ICONNAME/$Name/" | sed "s/ICONPATH/$Path/")
+  ExportTest=$(echo "$exportTemplate" | sed "s/ICONNAME/$Name/")
+
+  # CREATE IMPORT
+  sed -i "/IMPORT/a  $ImportTest" $FileName
```
kuzhelov, line 26: note that with the following `prepend` processing semantics the order of icons will be _reversed_ alphabetical. So, would suggest either
- ensure that icons are sorted properly at the point when their name is mapped to react elements (would prefer this option)
- or use `-r` flag for `ls` command on this line: https://github.com/stardust-ui/react/pull/478/files#diff-2c8401364afce39e9a7dc9c807267b5aR8

docs/src/prototypes/IconViewer/index.tsx
codepretty: I really like this. I updated to use `startsWith` and `sort()`. I didn't do the `map` for name here because I would need to then prepend the const to the name for the Icon lookup, so I'm not sure it's worth it.

docs/src/prototypes/IconViewer/index.tsx
```diff
@@ -6,6 +6,8 @@ const cellStyles = {
   margin: '10px 0',
 }
 
+const processedIconsNamePrefix = 'processedIcons_'
```
kuzhelov, line 9: 👍

kuzhelov (approved):
awesome work 👍 💯

PR #479, opened 2018-11-15
feat(Tree): add Component

priyankar205:
tree

- Implemented the Tree component including the Teams theme
- Work Left: Support for other themes. Implement accessibility props.


levithomason:
@layershifter could you please take a look at this component structure and API?

docs/src/examples/components/Tree/Types/TreeExample.shorthand.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react'
+import { Tree } from '@stardust-ui/react'
+const treeData = [
```
layershifter: An empty line before `const treeData`, please
```suggestion

const treeData = [
```

docs/src/examples/components/Tree/Types/TreeExample.shorthand.tsx
```diff
@@ -0,0 +1,29 @@
...
+    title: ' two',
+    subtree: [
+      {
+        title: 'two one',
+      },
+    ],
+  },
+]
+
+const TreeExampleShorthand = () => <Tree treedata={treeData} />
```
layershifter: Did you speak with Levi about this? I think that we should use `items` instead of `treeData` there.

```suggestion
const TreeExampleShorthand = () => <Tree items={items} />
```

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,93 @@
...
+}
+
+class Tree extends UIComponent<TreeProps, any> {
+  static create: Function
+
+  static className = 'tree-list'
+
+  static displayName = 'Tree'
+
+  // static handledProps = ['as', 'children', 'content', 'styles', 'variables']
```
layershifter: Oops, useless comment.

```suggestion
```

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    /** Custom variables to be applied to the component. */
+    variables: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
+
+    /** Shorthand array of props for Tree. */
+    treedata: PropTypes.array,
+
+    /** Whether the tree is a subtree. */
+    isSubTree: PropTypes.boolean,
+  }
```
layershifter, line 50: We recently merged PR for usage `react-docgen-typescript`, https://github.com/stardust-ui/react/pull/460

All comments for props should be moved to `TreeProps`

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  public static defaultProps = {
+    as: 'ul',
+    accessibility: treeBehavior as Accessibility,
+  }
+
+  renderContent(styles, variables) {
+    const { treedata } = this.props
+    if (!treedata) return []
+    return treedata.map(obj => {
```
layershifter: We can use there map from `lodash`, it will be useless condition above:

```suggestion
    return _.map(items, item => {
```

src/themes/teams/components/Tree/treeTitleVariables.ts
```diff
@@ -0,0 +1,11 @@
+// import { pxToRem } from '../../../../lib'
```
layershifter: Oops, useless comment.

```suggestion
```

src/themes/teams/components/Tree/treeTitleStyles.ts
```diff
@@ -0,0 +1,11 @@
+import { ICSSInJSStyle } from '../../../types'
+
+const treeTitleStyles = {
+  root: ({ variables }): ICSSInJSStyle => ({
+    padding: '.1rem 0',
```
layershifter: I'm not sure, but may be we should use there `pxToRem()` util?

/cc @miroslavstastny @kuzhelov

src/lib/accessibility/Behaviors/Tree/treeTitleBehavior.ts
```diff
@@ -0,0 +1,18 @@
+import { Accessibility } from '../../types'
+import * as _ from 'lodash'
+
+/**
+ * @description
+ *  Adds attribute 'aria-expanded=true' based on whether the folowing subtree is expanded or not.
+ */
+
```
layershifter: We don't need an empty line there.

```suggestion
```

src/lib/accessibility/Behaviors/Tree/treeBehavior.ts
```diff
@@ -0,0 +1,17 @@
+import { Accessibility } from '../../types'
+import * as _ from 'lodash'
+
+/**
+ * @description
+ *  Adds role='tree' to the tree element and role='group' to the subtree elements.
+ */
+
```
layershifter: We don't need an empty line there.

```suggestion
```

src/components/Tree/TreeTitle.tsx
```diff
@@ -0,0 +1,79 @@
...
+  styles?: ComponentSlotStyle
+  variables?: ComponentVariablesInput
+  active?: boolean
+  hasSubtree?: boolean
+}
+
+class TreeTitle extends UIComponent<TreeTitleProps, any> {
+  static create: Function
+
+  static className = 'tree-title'
```
layershifter: As I remember, we're following BEM for `className`

```suggestion
  static className = 'ui-tree__title'
```

src/components/Tree/TreeTitle.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+  static displayName = 'TreeTitle'
+
+  // static handledProps = ['as', 'children', 'content', 'styles', 'variables']
```
layershifter: Oops, useless comment

```suggestion
```

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,93 @@
...
+      const subtree = obj.subtree
+      return TreeListItem.create(obj.title, {
+        defaultProps: {
+          styles: styles.listItem,
+          subtree,
+          variables: variables.listItem,
+          titleStyles: styles.title,
+          titleVariables: variables.title,
+          onItemClick: obj.onItemClick,
```
layershifter: We don't need to manually map props there, it will be done by factory.

```suggestion
```

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  renderContent(styles, variables) {
+    const { treedata } = this.props
+    if (!treedata) return []
+    return treedata.map(obj => {
+      const subtree = obj.subtree
+      return TreeListItem.create(obj.title, {
+        defaultProps: {
+          styles: styles.listItem,
+          subtree,
```
layershifter: We don't need to manually map props there, it will be done by factory.

```suggestion
```

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,93 @@
...
+          variables: variables.listItem,
+          titleStyles: styles.title,
+          titleVariables: variables.title,
```
layershifter: We also should not apply `styles` and `variables` there, the better idea will be to create a separate component `TreeTitle`. And we have it, BTW.

Please see my final review comment.

layershifter (changes requested):
As we have already discussed with @levithomason we decided to introduce the `kind` property for complex markups.

## API

This means that API should be aligned. Each `Tree` consists of `items`:

```jsx
<Tree items={items} />
```

### TreeItem

Each item can be an `item`. Renders only item, no any magic with items/subitems.
_I'm not sure why it's called there `title`_, can you please clarify?

```js
const items = {
  'string item', // just pass a simple item, is equal to { content: 'string item', key: 'string item' }
  { content: 'an object item' },
  { kind: 'item', content: 'an object item' }, // is equal to line above
}
```

We will name this component as `TreeItem`.

### TreeTree

`items` can contain other trees:

```js
const items = {
  {
     kind: 'tree', // we need always to define `kind` for non-default item!
     title: 'A simple title' // is a shorthand for `TreeItem`,
     items: [
       'sub item 1',
       { content: 'sub item 2', key: '2' },
     ],
  }
}
```

I'm not sure about naming, let's go with `TreeTree` for now. After the final discussion with @levithomason  we will be able to change it.

All your logic with item selections will be in this component. My initial view of this:

```js
// somewhere on top of file

const factoryMapping = {
  item: TreeItem.create,
  tree: TreeTree.create,
}

const callFactory = (item) => {
  // This will cover all usages
  // _.get('something', 'kind', 'item');
  // _.get({ content: 'something' }, 'kind', 'item'); 
  // _.get({ content: 'something', key: '1' }, 'kind', 'item'); 
  const kind = _.get(item, 'kind', 'item')
  const factory = factoryMapping[kind]

    if (!factory) {
      throw new Error(
        'The unsupported factory there, you can use only ' + _.keys(factoryMapping).join(', ') + 'as `kind`',
      )
    }
  }

  factory(item)
}

// render() in TreeTree

return _.map(items, (item) => callFactory(item))
```

Let's also omit `activeContent` in the first implementation. This API also makes `hasSubtree` useless because it will be always `false` for `TreeItem` and always `true` for `TreeTree`.

The `Tree` component simply renders `TreeTree` component.

docs/src/examples/components/Tree/Types/TreeExample.shorthand.tsx
priyankar205: Yes Levi had a look that day. @levithomason What are your inputs?

priyankar205:
@layershifter : We will need activeContent for the use cases in Teams. We can support the 'kind' attribute in future iterations. I don't see any other component using it nor did I have any discussion on that with Levi (@levithomason)

src/components/Tree/Tree.tsx
priyankar205: The whole object is not getting passed. So we will need it.

src/components/Tree/Tree.tsx
priyankar205: The whole object is not getting passed. So we will need it.

docs/src/examples/components/Tree/Types/TreeExample.shorthand.tsx
layershifter: I'm not Levi, but I am 100% sure that we will not use `treedata` as a prop name.

src/components/Tree/TreeTitle.tsx
```diff
@@ -0,0 +1,72 @@
...
+  active?: boolean
+
+  /** Whether or not the item has a subtree. */
+  hasSubtree?: boolean
+}
+
+class TreeTitle extends UIComponent<TreeTitleProps, any> {
+  static create: Function
+
+  static className = 'tree-list__title'
```
layershifter: Please prefix with `ui-`:

```suggestion
  static className = 'ui-tree-list__title'
```

Other our components do the same:
https://github.com/stardust-ui/react/blob/master/src/components/Button/Button.tsx#L106
https://github.com/stardust-ui/react/blob/master/src/components/Header/HeaderDescription.tsx#L27

docs/src/examples/components/Tree/Types/TreeExclusiveExample.shorthand.tsx
```diff
@@ -0,0 +1,73 @@
...
+          <span>two</span>
+        </div>
+      ),
+      activeContent: (
+        <div>
+          <Icon name="arrow down" />
+          <span>two</span>
+        </div>
+      ),
+    },
```
layershifter: As I said before, we don't there an `activeContent` prop and it should be removed.

We have a `renderShorthand` pattern which allows us to modify the rendered output, here is an example: https://github.com/stardust-ui/react/blob/master/src/components/Header/Header.tsx#L88

So, this stuff should be done with `renderTitle`.

I've made an example how this can be made: https://codesandbox.io/s/l2593zrzwm

layershifter:
## Final changes

- [x] Rename `subtree` to `items`
- [x] Move `renderTitle` to the top level of `Tree`
- [x] rename `ui-tree-list` className to `ui-tree` on `Tree`
- [x] items on `Tree` should have type `ShorthandValue`
- [x] add `accessibility` props
- [x] drop `activeContent` from props

layershifter (approved):
I'm going to approve this thing, but I have a strong opinion that we need to hear there @levithomason 

### 1. What is `TreeListItem`?

For me, the of this component is very unclear. I think that we need somehow move all logic to `Tree`.

### 2. `preventDefault()`, `href=#` in `defaultProps`

I'm not a big fan of this method, `href=#` is very confusing for me, too. What if we will pass `as='div'`?

src/components/Tree/Tree.tsx
```diff
@@ -69,9 +69,7 @@ class Tree extends UIComponent<TreeProps, any> {
     return _.map(items, obj => {
       return TreeListItem.create(obj.title, {
         defaultProps: {
-          styles: styles.listItem,
```
priyankar205: We will need these variables for theming.

src/components/Tree/Tree.tsx
```diff
@@ -72,8 +72,6 @@ class Tree extends UIComponent<TreeProps, any> {
           styles: styles.listItem,
           items: obj.items,
           variables: variables.listItem,
-          titleStyles: styles.title,
```
priyankar205: We will need these variables for theming.

src/components/Tree/TreeListItem.tsx
```diff
@@ -112,18 +112,19 @@ class TreeListItem extends UIComponent<TreeListItemProps> {
             defaultProps: {
               items,
               nested: true,
+              renderTitle,
```
priyankar205: render title needs to be in the TreeListItem level. Different titles may need different renderers. Please see my commits and examples

priyankar205:
@layershifter : TreeListItem can be renamed to TreeItem. The structure for the tree was already agreed upon after an accessibility review so the 'a' tag is necessary

src/components/Tree/TreeListItem.tsx
layershifter: It's a default prop, so it you will pass another renderer in `items` it will work:

```jsx
const items = [
{
    key: '2',
    title: ' two',
    items: [
      {
        key: '1',
        title: 'two one',
        renderTitle: OTHER_RENDERER,
      },
    ],
  },
]

const TreeExclusiveExample = () => <Tree items={items} renderTitle={DEFAULT_RENDERER} />
```

src/components/Tree/Tree.tsx
layershifter: Can you please provide an example?

https://github.com/stardust-ui/react/blob/master/src/components/Header/Header.tsx#L85
I'm almost sure that we don't need to pass down all variables by tree.

priyankar205:
@layershifter Should we move href to the defaultProps of TreeTitle rether than passing it in.

layershifter:
> @layershifter Should we move href to the defaultProps of TreeTitle rether than passing it in.

Yes, this can be an option, let's do this.


However, the problem is a little deeper:

```jsx
<TreeTitle /> => // <a href='#' /> - valid markup :)
<TreeTitle as='div' /> => // <div href='#' /> - invalid markup :(
```

src/components/Tree/Tree.tsx
priyankar205: I was using the pattern used here in button https://github.com/stardust-ui/react/blob/28c8307fa24dc8e5adbdb882f1c0c2d8cfbeeec2/src/components/Button/Button.tsx#L171

priyankar205:
@layershifter We are not exposing TreeTitle to users. We will only use it in context of a Tree. So no one should be able to use ```<TreeTitle as='div' /> ```

layershifter:
> @layershifter We are not exposing TreeTitle to users. We will only use it in context of a Tree. So no one should be able to use

```jsx
const items = [{
  as: 'div',
  title: ' two',
}]
```

It's possible with shorthand API.

priyankar205:
> > @layershifter We are not exposing TreeTitle to users. We will only use it in context of a Tree. So no one should be able to use
> 
> ```js-jsx
> const items = [{
>   as: 'div',
>   title: ' two',
> }]
> ```

This will change the 'li' to div

layershifter:
> This will change the 'li' to div

```jsx
const items = [{
  title: { as: 'div', content: two' }},
}]
// <div class="ui-tree__title" role="treeitem" href="#">two</div>
```

priyankar205:
> > This will change the 'li' to div
> 
> ```js-jsx
> const items = [{
>   title: { as: 'div', content: two' }},
> }]
> // <div class="ui-tree__title" role="treeitem" href="#">two</div>
> ```

Yes in this case we will have to explicitly override it.

jurokapsiar:
@priyankar205 the tree does not have keyboard navigation yet. In general, it should follow https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html except of using `tree` and `treeitem` roles since they do not work with  JAWS correctly. Do you plan to include it in this PR or will you create a separate PR for that?

In general, you can use `focusZone` prop in the behavior similar to `menuBehavior` and add custom action handlers for left and right arrow keys for opening/closing the nodes and/or going to the parent or child nodes. For an example with action handlers, you can take a look at `RadioGroup` component.

src/lib/accessibility/Behaviors/Tree/treeBehavior.ts
```diff
@@ -0,0 +1,15 @@
+import { Accessibility } from '../../types'
+
+/**
+ * @specification
+ * Adds role='tree' to the tree element and role='group' to the subtree elements.
+ */
+const treeBehavior: Accessibility = (props: any) => ({
+  attributes: {
+    root: {
+      role: props.nested ? 'group' : 'tree',
```
jurokapsiar: please do not use group and tree roles as they do not work correctly with some of the screen readers. Instead, follow the HTML structure that was sent before

src/components/Tree/TreeListItem.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+export interface TreeListItemProps
+  extends UIComponentProps<any, any>,
+    ChildrenComponentProps,
+    ContentComponentProps {
+  /** Initial activeIndex value. */
+  defaultOpen?: boolean
+
+  /** Shorthand array of props for sub tree. */
+  items?: any[]
```
jurokapsiar: can you use `ShorthandValue` instead of `any`?

priyankar205:
> @priyankar205 the tree does not have keyboard navigation yet. In general, it should follow https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2a.html except of using `tree` and `treeitem` roles since they do not work with JAWS correctly. Do you plan to include it in this PR or will you create a separate PR for that?

This is in accordance with the HTML that was shared with me. We can add the keyboard interactions in a separate PR (either me or someone else can pick up)

src/lib/accessibility/Behaviors/Tree/treeBehavior.ts
priyankar205: HTML structure is correct and the tabs & click work as per the shared HTML. I will remove these roles

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,84 @@
...
+    ...commonUIComponentPropTypes,
+    ...childrenComponentPropTypes,
+    accessibility: PropTypes.func,
+    items: customPropTypes.collectionShorthand,
+    nested: PropTypes.bool,
+    renderTitle: PropTypes.func,
+  }
+
+  public static defaultProps = {
+    as: 'ul',
```
jurokapsiar, line 53: please add
```
```

src/components/Tree/TreeTitle.tsx
```diff
@@ -0,0 +1,80 @@
...
+  ContentComponentProps,
+} from '../../lib/commonPropInterfaces'
+
+export interface TreeTitleProps
+  extends UIComponentProps<any, any>,
+    ChildrenComponentProps,
+    ContentComponentProps {
+  /**
+   * Accessibility behavior if overridden by the user.
+   * @default treeTitleBehavior
```
jurokapsiar: defaultBehavior

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { ShorthandValue, ShorthandRenderFunction } from '../../../types/utils'
+import { Accessibility } from '../../lib/accessibility/types'
+import { commonUIComponentPropTypes, childrenComponentPropTypes } from '../../lib/commonPropTypes'
+import { UIComponentProps, ChildrenComponentProps } from '../../lib/commonPropInterfaces'
+import * as customPropTypes from '../../lib/customPropTypes'
+
+export interface TreeProps extends UIComponentProps<any, any>, ChildrenComponentProps {
+  /**
+   * Accessibility behavior if overridden by the user.
+   * @default treeBehavior
```
jurokapsiar: defaultBehavior

src/lib/accessibility/Behaviors/Tree/treeTitleBehavior.ts
```diff
@@ -0,0 +1,15 @@
+import { Accessibility } from '../../types'
+
+/**
+ * @specification
+ * Adds attribute 'aria-expanded=true' based on whether the folowing subtree is expanded or not.
```
jurokapsiar: ```Adds attribute 'aria-expanded=true' based on the properties 'hasSubtree' and 'open'.```

src/components/Tree/TreeTitle.tsx
```diff
@@ -0,0 +1,80 @@
...
+    ...childrenComponentPropTypes,
+    ...contentComponentPropsTypes,
+    accessibility: PropTypes.func,
+    open: PropTypes.bool,
+    hasSubtree: PropTypes.bool,
+  }
+
+  public static defaultProps = {
+    as: 'a',
+    href: '#',
```
jurokapsiar, line 54: agree with @layershifter's comment on `href` default value. In the future PR for #511 can you please remove this default value and add keyboard handlers for `Enter` and `Space`?

src/lib/accessibility/Behaviors/Tree/treeTitleBehavior.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+const treeTitleBehavior: Accessibility = (props: any) => ({
+  attributes: {
+    root: {
+      ...(props.hasSubtree && { 'aria-expanded': props.open ? 'true' : 'false' }),
```
jurokapsiar, line 10: for some reason this does not seem to work correcttly. In the screenshot below, first opened node is highlighted, but it has `aria-expanded='false'`:
image

priyankar205: Fixed

kolaps33:
I took look on tee and figure out following:
"Default" variation:
- on the list leaf (last list elements) we shouldn't put aria-expanded
- on items on first level (one, two) please put aria-level=1 and then on second level(one one, two one) add aria-level=2 

Second list "exclusive":
image
should we show up on the leaf element the arrow image if there is no collapse/expand action available?

priyankar205:
> I took look on tee and figure out following:
> "Default" variation:
> 
> * on the list leaf (last list elements) we shouldn't put aria-expanded
> * on items on first level (one, two) please put aria-level=1 and then on second level(one one, two one) add aria-level=2
> 
> Second list "exclusive":
> image
> should we show up on the leaf element the arrow image if there is no collapse/expand action available?

@kolaps33  aria-expanded is not applied to a leaf node. I did not apply the aria-level because the dom hierarchy is correct here.

The example was modified by @layershifter. I will fix that

level

jurokapsiar (approved):

docs/src/examples/components/Tree/Types/TreeExclusiveExample.shorthand.tsx
```diff
@@ -0,0 +1,42 @@
...
+]
+
+const titleRenderer = (Component, { content, open, hasSubtree, ...rest }) => (
+  <Component open={open} {...rest}>
+    {hasSubtree && <Icon name={open ? 'arrow down' : 'arrow right'} />}
+    <span>{content}</span>
+  </Component>
+)
+
+const TreeExclusiveExample = () => <Tree items={items} renderTitle={titleRenderer} />
```
kuzhelov: could you, pease, suggest what does `exclusive` implies in this example's name? If it is all about custom rendered title, we could consider the following name for this example `TreeTitleCustomizationExample` - this would follow naming conventions we are already using for docs

docs/src/examples/components/Tree/Types/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const Types = () => (
+  <ExampleSection title="Types">
+    <ComponentExample
+      title="Default"
+      description="A default Tree."
+      examplePath="components/Tree/Types/TreeExample"
+    />
+    <ComponentExample
+      title="Exclusive"
```
kuzhelov: would suggest something like `Custom Title`, it will be more expressive

src/themes/teams/components/Tree/treeTitleStyles.ts
kuzhelov: yes, absolutely valid point - @priyankar205, please, consider to use `pxToRem` utility for all sizes that should scale with page's font size being changed, similar how it is done for button styles: https://github.com/stardust-ui/react/blob/master/src/themes/teams/components/Button/buttonStyles.ts#L83

src/themes/teams-high-contrast/components/Tree/treeTitleVariables.ts
```diff
@@ -0,0 +1,9 @@
+export interface TreeTitleVariables {
```
kuzhelov: these interface definitions shouldn't be duplicated for `dark` and `high-contrast` themes - please, consider the approach taken for `buttonVariables`, where interfaces for these theme variants are based on the default one: https://github.com/stardust-ui/react/blob/master/src/themes/teams-dark/components/Button/buttonVariables.ts#L4

src/themes/teams-dark/components/Tree/treeTitleVariables.ts
```diff
@@ -0,0 +1,9 @@
+export interface TreeTitleVariables {
```
kuzhelov: please, don't duplicate interface definitions - consider to use the one defined for a default theme, similar to how it is done for `buttonVariables`: https://github.com/stardust-ui/react/blob/master/src/themes/teams-dark/components/Button/buttonVariables.ts#L4

src/themes/teams/componentStyles.ts
```diff
@@ -52,4 +52,7 @@ export { default as Status } from './components/Status/statusStyles'
 
 export { default as Text } from './components/Text/textStyles'
 
+export { default as Tree } from './components/Tree/treeStyles'
+export { default as TreeListItem } from './components/Tree/treeListItemStyles'
+export { default as TreeTitle } from './components/Tree/treeTitleStyles'
 export { default as Transition } from './components/Transition/transitionStyles'
```
kuzhelov: nit: would suggest to put empty line after these three, to aid visual separation of components from different domains

src/themes/teams/componentVariables.ts
```diff
@@ -47,4 +47,5 @@ export { default as Status } from './components/Status/statusVariables'
 
 export { default as Text } from './components/Text/textVariables'
 
+export { default as TreeTitle } from './components/Tree/treeTitleVariables'
 export { default as Transition } from './components/Transition/transitionVariables'
```
kuzhelov: nit: please, put empty line after `TreeTitle` to drive visual separation

src/themes/teams/components/Tree/treeListItemStyles.ts
```diff
@@ -0,0 +1,10 @@
+import { ICSSInJSStyle } from '../../../types'
+
+const treeListItemStyles = {
+  root: (): ICSSInJSStyle => ({
+    listStyleType: 'none',
+    padding: '0 0 0 .5rem',
```
kuzhelov: please, consider to use `pxToRem` here, as this value, apparently, should scale with page's font size being changed (this we have now) - but more than that, **there should be a possibility to relatively scale theme's components relative to the _rest_ content on the same page**, and `pxToRem` is the utility that will guarantee it

src/themes/teams/components/Tree/treeStyles.ts
```diff
@@ -0,0 +1,10 @@
+import { ICSSInJSStyle } from '../../../types'
+
+const treeStyles = {
+  root: (): ICSSInJSStyle => ({
+    display: 'block',
+    paddingLeft: '.5rem',
```
kuzhelov: please, use pxToRem here

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,86 @@
...
+  accessibility?: Accessibility
+
+  /** Shorthand array of props for Tree. */
+  items: ShorthandValue[]
+
+  /** Whether the tree is a subtree. */
+  nested?: boolean
+
+  /**
+   * A custom render function the title slot.
```
kuzhelov: nit: `for`/`of` is apparently missed: 'A custom render function **for/of** the **item's** title slot.'

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,86 @@
...
+   *
+   * @param {React.ReactType} Component - The computed component for this slot.
+   * @param {object} props - The computed props for this slot.
+   * @param {ReactNode|ReactNodeArray} children - The computed children for this slot.
+   */
+  renderTitle?: ShorthandRenderFunction
+}
+
+/**
+ * A tree allows users to display nested trees.
```
kuzhelov: :) actually, this definition is aimed to define term based on this (yet undefined) term itself. We might consider to improve by something like: `Allows users to display data organised in tree-hierarchy.`

src/components/Tree/TreeListItem.tsx
```diff
@@ -0,0 +1,156 @@
...
+   * A custom render iterator for rendering each Accordion panel title.
+   * The default component, props, and children are available for each panel title.
+   *
+   * @param {React.ReactType} Component - The computed component for this slot.
+   * @param {object} props - The computed props for this slot.
+   * @param {ReactNode|ReactNodeArray} children - The computed children for this slot.
+   */
+  renderTitle?: ShorthandRenderFunction
+
+  /** Shorthand for TreeTitle. */
```
kuzhelov: lets omit type names usage in the descriptions

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,86 @@
...
+  nested?: boolean
+
+  /**
+   * A custom render function the title slot.
+   *
+   * @param {React.ReactType} Component - The computed component for this slot.
+   * @param {object} props - The computed props for this slot.
+   * @param {ReactNode|ReactNodeArray} children - The computed children for this slot.
+   */
+  renderTitle?: ShorthandRenderFunction
```
kuzhelov: would suggest to rename to `renderItemTitle`, as it will better reflect semantics of this prop

src/components/Tree/TreeListItem.tsx
```diff
@@ -0,0 +1,156 @@
...
+      <ElementType className={classes.root} {...accessibility.attributes.root} {...rest}>
+        {childrenExist(children) ? children : this.renderContent()}
+      </ElementType>
+    )
+  }
+}
+
+TreeListItem.create = createShorthandFactory(TreeListItem, 'content')
+
+export default TreeListItem
```
kuzhelov: maybe we could simplify this name (and be consistent with general Stardust naming convention) by renaming it to just `TreeItem`? Please, let me know if you have some thoughts that go against this renaming move.

src/components/Tree/TreeListItem.tsx
```diff
@@ -0,0 +1,156 @@
...
+    renderTitle: PropTypes.func,
+    title: customPropTypes.itemShorthand,
+  }
+
+  public static defaultProps = {
+    as: 'li',
+    accessibility: defaultBehavior,
+  }
+
+  handleItemOverrides = predefinedProps => ({
```
kuzhelov: `handleItemOverrides` => `handleTitleOverrides`?

src/components/Tree/TreeListItem.tsx
```diff
@@ -0,0 +1,156 @@
...
+  /** Shorthand array of props for sub tree. */
+  items?: ShorthandValue[]
+
+  /**
+   * Called when a tree title is clicked.
+   *
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All title props.
+   */
+  onItemClick?: ComponentEventHandler<TreeListItemProps>
```
kuzhelov: this one can be safely removed, as client will be able to utilize `title` shorthand's API for that

src/components/Tree/TreeListItem.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+  handleItemOverrides = predefinedProps => ({
+    onClick: (e, titleProps) => {
+      e.preventDefault()
+      this.trySetState({
+        open: !this.state.open,
+      })
+
+      _.invoke(predefinedProps, 'onClick', e, titleProps)
+      _.invoke(this.props, 'onItemClick', e, titleProps)
```
kuzhelov: this `onItemClick` prop can be safely removed (actually, it should be rather called `onTitleClick`) - please, suggest those cases where it might be necessary otherwise

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+    return (
+      <ElementType className={classes.root} {...accessibility.attributes.root} {...rest}>
+        {childrenExist(children) ? children : this.renderContent()}
+      </ElementType>
+    )
+  }
+}
+
+Tree.create = createShorthandFactory(Tree, 'content')
```
kuzhelov: there is no `content` property defined for this component

src/components/Tree/TreeListItem.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+    return (
+      <ElementType className={classes.root} {...accessibility.attributes.root} {...rest}>
+        {childrenExist(children) ? children : this.renderContent()}
+      </ElementType>
+    )
+  }
+}
+
+TreeListItem.create = createShorthandFactory(TreeListItem, 'content')
```
kuzhelov: note, `content` prop is not used _anywhere_ in this implementation. We should introduce some meaningful prop (that would be consumed as part of the implementation) here

test/specs/behaviors/testDefinitions.ts
```diff
@@ -162,6 +162,27 @@ definitions.push({
...
+      propertyDependingOnFirst,
+      propertyDependingOnSecond,
+    ] = [...parameters.props]
+    const property = {}
+    property[propertyDependingOnFirst] = attributeExpectedValue
+    property[propertyDependingOnSecond] = true
+
+    const expectedResult = parameters.behavior(property).attributes.root[attributeToBeAdded]
+    expect(testHelper.convertToBooleanIfApplicable(expectedResult)).toEqual(
+      testHelper.convertToBooleanIfApplicable(attributeExpectedValue),
```
kolaps33, line 183: This is good test 👍 

but i think we could cover more path of the expression which we have in behavior, because with this test we cover only 1/3 of the statement. 

I was trying myself and we could extend like to more statement to be tested:

  const propertySecondPropUndefined = {}
    propertySecondPropUndefined[propertyDependingOnFirst] = attributeExpectedValue
    propertySecondPropUndefined[propertyDependingOnSecond] = undefined
    const expectedResultSecondPropUndefined = parameters.behavior(propertySecondPropUndefined).attributes.root[attributeToBeAdded]
    expect(testHelper.convertToBooleanIfApplicable(expectedResultSecondPropUndefined)).toEqual(
      testHelper.convertToBooleanIfApplicable(undefined),
    )

    const propertyFirstPropNegate = {}
    propertyFirstPropNegate[propertyDependingOnFirst] = !attributeExpectedValue
    propertyFirstPropNegate[propertyDependingOnSecond] = true
    const expectedResultFirstPropertyNegate = parameters.behavior(propertyFirstPropNegate).attributes.root[attributeToBeAdded]
    expect(testHelper.convertToBooleanIfApplicable(expectedResultFirstPropertyNegate)).toEqual(
      testHelper.convertToBooleanIfApplicable(!attributeExpectedValue),
    )

src/lib/accessibility/Behaviors/Tree/treeTitleBehavior.ts
```diff
@@ -0,0 +1,15 @@
+import { Accessibility } from '../../types'
+
+/**
+ * @specification
+ * Adds attribute 'aria-expanded=true' based on the property 'open' and 'hasSubtree'
```
kolaps33: Please what do you think if we change the line like:
Adds attribute 'aria-expanded=true' based on the property 'open' if the component has 'hasSubtree' property.

then it would invoke/consider more "expect" verification in the unit test for this behavior... please see my comment in the unit tests for this behavior...

If you change this line, please don't forget to amend the regex expression as well...
Thank you :)

src/components/Tree/TreeListItem.tsx
priyankar205: Please suggest one

src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,141 @@
...
+  extends UIComponentProps<any, any>,
+    ChildrenComponentProps,
+    ContentComponentProps {
+  /**
+   * Accessibility behavior if overridden by the user.
+   * @default defaultBehavior
+   */
+  accessibility?: Accessibility
+
+  /** Initial activeIndex value. */
```
sophieH29: wrong comment -> ```Initial open value```

src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,141 @@
...
+  renderContent() {
+    const { items, title, renderItemTitle } = this.props
+    const { open } = this.state
+
+    return (
+      <>
+        {TreeTitle.create(title, {
+          defaultProps: {
+            open,
+            hasSubtree: !!(items && items.length),
```
sophieH29: Why do you need to use double inverted boolean? I think this should be enough ```items && items.length```.

src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,141 @@
...
+          },
+          render: renderItemTitle,
+          overrideProps: this.handleTitleOverrides,
+        })}
+        {items &&
```
sophieH29: Should we still render a subtree if ```items``` is an empty array?
So maybe check ```items && items.length``` as you checked on the line 110. 
Event better to move this to a variable -> ```const hasSubtree = !!(items && items.length)```
and use it on
- line 110 - ```hasSubtree: hasSubtree```
- line 115 - ```{hasSubtree && open ....} ```

src/components/Tree/TreeItem.tsx
layershifter: Nice catch, my fault 🐱

src/components/Tree/TreeTitle.tsx
```diff
@@ -0,0 +1,80 @@
...
+    accessibility: treeTitleBehavior as Accessibility,
```
sophieH29: instead of casting, to ensure proper typings
```js
public static defaultProps: TreeTitleProps = {
  ...
}
```

test/specs/behaviors/testDefinitions.ts
```diff
@@ -162,6 +162,49 @@ definitions.push({
...
+      attributeExpectedValue,
+      propertyDependingOnFirst,
+      propertyDependingOnSecond,
+    ] = [...parameters.props]
+
+    const property = {}
+
+    property[propertyDependingOnFirst] = attributeExpectedValue
+    property[propertyDependingOnSecond] = true
+    const expectedResult = parameters.behavior(property).attributes.root[attributeToBeAdded]
```
sophieH29: I think it should be ```actualResult``` instead of ```expected```

test/specs/behaviors/testDefinitions.ts
```diff
@@ -162,6 +162,49 @@ definitions.push({
...
+    expect(testHelper.convertToBooleanIfApplicable(expectedResult)).toEqual(
+      testHelper.convertToBooleanIfApplicable(attributeExpectedValue),
+    )
+
+    const propertyFirstPropNegate = {}
+    propertyFirstPropNegate[propertyDependingOnFirst] = !testHelper.convertToBooleanIfApplicable(
+      attributeExpectedValue,
+    )
+    propertyFirstPropNegate[propertyDependingOnSecond] = true
+    const expectedResultFirstPropertyNegate = parameters.behavior(propertyFirstPropNegate)
```
sophieH29: Again, a little improvement to naming, to reflect the meaning ```expectedResultFirstPropertyNegate``` - >  ```actualResultFirstPropertyNegate```, because this is what actually the behavior has set.  So then you verify ```actual``` and ```expected```, not ```expected``` and ```expected```.

test/specs/behaviors/testDefinitions.ts
```diff
@@ -162,6 +162,49 @@ definitions.push({
...
+      .attributes.root[attributeToBeAdded]
+    expect(testHelper.convertToBooleanIfApplicable(expectedResultFirstPropertyNegate)).toEqual(
+      !testHelper.convertToBooleanIfApplicable(attributeExpectedValue),
+    )
+
+    const propertyFirstPropUndefined = {}
+    propertyFirstPropUndefined[propertyDependingOnFirst] = true
+    propertyFirstPropUndefined[propertyDependingOnSecond] = undefined
+    const expectedResultFirstPropertyNegateUndefined = parameters.behavior(
```
sophieH29: Same comment above applies here

docs/src/examples/components/Tree/Types/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+const Types = () => (
+  <ExampleSection title="Types">
+    <ComponentExample
+      title="Default"
+      description="A default Tree."
+      examplePath="components/Tree/Types/TreeExample"
+    />
+    <ComponentExample
+      title="Custom Title"
+      description="A Tree with components as title."
```
kuzhelov: nit: `A Tree with customized title rendering` ?

src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,143 @@
...
+    ChildrenComponentProps,
+    ContentComponentProps {
+  /**
+   * Accessibility behavior if overridden by the user.
+   * @default defaultBehavior
+   */
+  accessibility?: Accessibility
+
+  /** Initial open value. */
+  defaultOpen?: boolean
```
kuzhelov: I am not sure about exact use case where it might be necessary. If it is about specifying default value for `open` prop, then I would suggest to use `defaultProps` for that. 

At the same time, it seems that this prop is not used _anywhere_ now - would suggest to remove it then.

src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,143 @@
...
+    title: customPropTypes.itemShorthand,
+  }
+
+  public static defaultProps = {
+    as: 'li',
+    accessibility: defaultBehavior,
+  }
+
+  handleTitleOverrides = predefinedProps => ({
+    onClick: (e, titleProps) => {
```
kuzhelov, line 87: note that while this is a behavior that is generally expected, there are examples of tree components where either:
- this action shouldn't be supported by click on the title
- another action should serve as a trigger (like double click)

Probably, we should think about customizing this moment - probably, we should file an issue for that.

src/components/Tree/TreeTitle.tsx
```diff
@@ -0,0 +1,80 @@
...
+    ...commonUIComponentPropTypes,
+    ...childrenComponentPropTypes,
+    ...contentComponentPropsTypes,
+    accessibility: PropTypes.func,
+    open: PropTypes.bool,
+    hasSubtree: PropTypes.bool,
+  }
+
+  public static defaultProps = {
+    as: 'a',
```
kuzhelov, line 53: why `<a />` element is used for item's title (and, say, not regular `<span />`)?

kuzhelov (commented):
have couple of questions that I would like to clarify:
- could you, please, suggest how it would be possible to pass some arbitrary content as tree's child (other than just plain `string`), like in the following case:

```
items: [
  <span dir='auto'>Hello from component!</span>,
  <span>for more info, click <button>here</button>
]
```

The only way how it could be achieved now seems to be the following:

```js
items: [
  {
    title: <span dir='auto'>Hello from component!</span>
  },
  {
     title: <span>for more info, click <button>here</button>
  }
```

This syntax is quite verbose for achieving this simple task. Note that this problem might be alleviated by passing `title` property as a default one for `TreeItem` component: 

```js
TreeItem.create = createShorthandFactory(TreeItem, 'title')
```

------

- second question is about this line (and related comment): https://github.com/stardust-ui/react/pull/479/files#diff-560cfd0141d58538f8a5e7845f4ab0aeR84 (about the fact that there is no `content` prop defined for `Tree` component, but this one is used in a shorthand). Could you, please, suggest what is a resolution suggested for it?

Thank you!

src/components/Tree/TreeItem.tsx
priyankar205: Here is the link to the discussion [and example that we are following]. There is another issue open for accessibility enhancements.
https://github.com/stardust-ui/react/pull/479#issuecomment-440795071

src/components/Tree/TreeTitle.tsx
priyankar205: We are following a dom structure that was approved from the accessibility team. @jurokapsiar has context.

jurokapsiar: It should be an actionable element - `button` or `a`. It then depends on concrete usage if `a` is preferable to `button`. I would leave that as a separate discussion, outside of this PR.

priyankar205:
@kuzhelov : I agree to your first point. I will update the code to `TreeItem.create = createShorthandFactory(TreeItem, 'title')` 
For the second question: As the tree is represented by an 'ul' element which cannot have any child except 'li' no property can be mapped. I will update the code for that.

CHANGELOG.md
```diff
@@ -39,6 +39,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Apply `dir=auto` attribute to string content of `Text` @kuzhelov  ([#5](https://github.com/stardust-ui/react/pull/5))
 - Improve `Menu` accessibility behaviors @sophieH29 ([#523](https://github.com/stardust-ui/react/pull/523))
 - Add ability to style every slot of `Chat.Message` and remove dependency on `Layout` component @Bugaa92 ([#518](https://github.com/stardust-ui/react/pull/518))
+- Add `Tree` Component @priyankar205 ([#479]
```
kuzhelov: please, ensure that this entry will be provided in 'Unreleased' section - yes, this is an unfortunate consequence of conflicts resolution

src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,136 @@
...
+  static className = 'ui-tree__item'
+
+  static displayName = 'TreeItem'
+
+  static autoControlledProps = ['open']
+
+  static propTypes = {
+    ...commonUIComponentPropTypes,
+    ...childrenComponentPropTypes,
+    ...contentComponentPropsTypes,
```
kuzhelov: as `content` prop is not used anywhere, lets remove it from API - both from propTypes and TS interface

src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,84 @@
...
+  static className = 'ui-tree'
+
+  static displayName = 'Tree'
+
+  static propTypes = {
+    ...commonUIComponentPropTypes,
+    ...childrenComponentPropTypes,
+    accessibility: PropTypes.func,
+    items: customPropTypes.collectionShorthand,
+    nested: PropTypes.bool,
```
kuzhelov: this prop seems to be not used anywhere - am I missing something (please, just reference the line in code in that case)? Thank you

kuzhelov:
@priyankar205 
thanks a lot for hour response. One moment that I still would like to clarify, though - the one that is related to the question of using content other than plain string for tree leafs.

While I totally agree that `<ul>` should contain only `<li>` elements, at the same time it is still valid that we could, at least in theory, provide some content that is more complex than plain string to this `<li>` element, right? Like in the following example:

```html
<ul>
  <li> <span>I am not just a string</span> <li>
  <li> Me neither, click here if you disagree: <button>Disagree!</button> </li>
</ul>
```

So, to me it seems that current implementation is unable to handle these cases - in a sense, as I've already mentioned before, client will need to provide this content as a value for `title` prop, that is completely non-obvious (why `header` if semantics of this is a `content`). Please, just verify or object my assumptions, so that we would be able to have mutual understanding of where we are and what are the next steps that should be planned.

Thank you!

src/components/Tree/Tree.tsx
priyankar205: I had added a class. Got deleted in refactoring perhaps. Also it might be needed for accessibility props.

src/components/Tree/TreeItem.tsx
priyankar205: Done

priyankar205:
@kuzhelov We can handle these cases with the renderItemTitle function for now as in the second example. Also I see that other components like List also fail to handle these cases directly.

src/components/Tree/Tree.tsx
kuzhelov: so, if it is not used anywhere now, lets delete it - even if it will be used for accessibility later, we will add it. Lets limit the API to only those things that are necessary

src/components/Tree/Tree.tsx
```diff
@@ -71,7 +71,11 @@ class Tree extends UIComponent<TreeProps> {
   }
 
   renderComponent({ ElementType, classes, accessibility, rest, styles, variables }) {
-    const { children } = this.props
+    const { children, nested } = this.props
+
+    if (nested) {
+      classes.root = `ui-subtree ${classes.root}`
```
kuzhelov: first point is that we should use `cx` function for that. Another is more about semantics of this branching - I don't agree that we should differentiate class names depending on whether it is a nested tree or not (classes are used for CSS purposes, so that `ui-tree ui-tree smt { .. }` selector would work just fine for subtree). Lets remove that and always use `ui-tree` class name, as it was before

kuzhelov (approved):

src/components/Tree/Tree.tsx
priyankar205: Done

PR #480, opened 2018-11-15
docs(Prototype): Create prototype for Emoji / Sticker popups

sophieH29:
**Components used:**
- Button,
- Image,
- Popup,
- Grid,
- Input,
- Menu,

**Accessibility behaviors used:**
- tabBehavior,
- tabListBehavior,
- popupFocusTrapBehavior,
- gridBehavior

**Emoji popup:**

image

**Sticker popup:**
image


jurokapsiar (approved):

docs/src/prototypes/popups/emojiPopup.tsx
```diff
@@ -0,0 +1,47 @@
...
+  <Popup
+    accessibility={popupFocusTrapBehavior}
+    position="below"
+    trigger={<Button icon="smile" aria-label="Choose an emoji." />}
+    content={{
+      content: (
+        <div
+          aria-label="Choose an emoji. Press Enter key to insert emoji."
+          role="dialog"
+          aria-modal="true"
```
jurokapsiar: let's see if we can create a Dialog (?) component and behavior that describes it, sets the role and aria-modal. The component can live within the prototypes/popup directory

docs/src/prototypes/popups/stickerPopup.tsx
```diff
@@ -0,0 +1,135 @@
...
+                {
+                  <Input
+                    styles={{ marginBottom: '10px' }}
+                    fluid
+                    icon="search"
+                    placeholder="Search..."
+                  />
+                }
+                {
+                  <Grid
```
jurokapsiar: let's create a GridImagePicker (?) component from this.

docs/src/prototypes/popups/helper.tsx
```diff
@@ -0,0 +1,297 @@
...
+const handleSelection = e => {
+  if (!e.target) return
+  const img = e.target.nodeName !== 'IMG' ? e.target.querySelector('img') : e.target
+  const selectedItem = img && img.getAttribute('aria-label')
+  if (!selectedItem) return
+}
+
+export const renderImages = (images, ariaLabelPart) => {
+  return _.map(images, (image, index) => (
+    <li key={image.key}>
```
jurokapsiar: let's create a GridImagePickerItem (?) component

jurokapsiar:
I was thinking about this. It looks really good but the accessibility part is baked into the example and not externalized. We should follow the component+behavior principle. Let's create these components within the prototypes/portal directory, we will see later if the 'internal' components will be generalized and baked into normal stardust components.

docs/src/prototypes/popups/GridImagePicker/GridImagePicker.tsx
```diff
@@ -0,0 +1,52 @@
...
+  inputStyles = {
+    marginBottom: '10px',
+  }
+
+  render() {
+    const gridTag = this.props.as || 'ul'
+
+    return (
+      <>
+        {<Input styles={this.inputStyles} fluid icon="search" placeholder="Search..." />}
```
layershifter: `{` & `}` are useless there

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+    background: '#fff',
+  }
+
+  handleClick = e => {
+    _.invoke(this.props, 'onClick', e, this.props)
+  }
+
+  render() {
+    const { title, imageSrc } = this.props
+    const Tag = this.props.as || 'li'
```
layershifter: Please, use `defaultProps` there

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+  imageStyle = {
+    width: '100%',
+  }
+
+  imageButtonStyles = {
+    minWidth: '56px',
+    height: '56px',
+    padding: '0',
+    background: '#fff',
+  }
```
layershifter: Styles can be defined as `const` outside the `GridImageItemPicker` class. Please also align names `imageStyle` => `imageStyles`

docs/src/prototypes/popups/GridImagePicker/GridImagePicker.tsx
```diff
@@ -0,0 +1,50 @@
...
+    margin: '0',
+    gridRowGap: '10px',
+  }
+
+  inputStyles = {
+    marginBottom: '10px',
+  }
+
+  render() {
+    const gridTag = this.props.as || 'ul'
```
layershifter: Please, use `defaultProps` there

docs/src/prototypes/popups/GridImagePicker/GridImagePicker.tsx
```diff
@@ -0,0 +1,50 @@
...
+          style={this.gridStyles}
+          content={this.renderGridItems()}
+        />
+      </>
+    )
+  }
+
+  renderGridItems() {
+    return _.map(this.props.items, (item, idx) => (
+      <GridImagePickerItem imageSrc={item.imageSrc} title={item.title} />
```
layershifter: `idx` is unused variable

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
```diff
@@ -0,0 +1,43 @@
+import * as React from 'react'
+
+import { Image, Button } from '@stardust-ui/react'
+
+import _ from 'lodash'
```
layershifter: Too many empty lines 😸 

```suggestion
import { Image, Button } from '@stardust-ui/react'
import _ from 'lodash'
import * as React from 'react'
```

sophieH29: I put label ```wip``` to this PR hours ago, and removed ```ready for review``` 😄 
Sorry for the confusion, working on it after comments from @jurokapsiar. Anyway, thank you for the initial code review

docs/src/prototypes/popups/GridImagePicker/GridImagePicker.tsx
```diff
@@ -0,0 +1,53 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+
+import { Grid, Input, gridBehavior } from '@stardust-ui/react'
+import GridImagePickerItem, { GridPickerItemProps } from './GridImagePickerItem'
```
layershifter, line 6: I'm going to introduce a tslint rule for imports, but for now, let's just arrange them

```suggestion
import { Grid, Input, gridBehavior } from '@stardust-ui/react' // stardust is on top because it's namespaced package
import * as _ from 'lodash' // other goes by module name
import * as React from 'react'

import GridImagePickerItem, { GridPickerItemProps } from './GridImagePickerItem'
```

docs/src/prototypes/popups/GridImagePicker/GridImagePicker.tsx
```diff
@@ -0,0 +1,53 @@
...
+class GridImagePicker extends React.Component<GridPickerProps, any> {
+  static defaultProps = {
+    as: 'ul',
+    gridColumns: 5,
+  }
+
+  render() {
+    return (
+      <>
+        <Input styles={inputStyles} fluid icon="search" placeholder="Search..." />
```
layershifter: May be we need allow to customise `placeholder` via prop?

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
```diff
@@ -0,0 +1,51 @@
+import * as React from 'react'
+import { Image, Button } from '@stardust-ui/react'
+import * as _ from 'lodash'
```
layershifter, line 4: Imports again 😸

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export interface GridPickerItemProps {
+  as?: keyof React.ReactHTML
+  title?: string
+  imageSrc: string
+  onClick?: (e) => void
```
layershifter: `e` should have type definition,

```suggestion
  onClick?: (e: React.SyntethicEvent) => void
```

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    _.invoke(this.props, 'onClick', e, this.props)
+  }
+
+  render() {
+    const { title, imageSrc } = this.props
+
+    return (
+      <this.props.as>
```
layershifter: It's a working solution, but it looks a little unfriendly to me, make we can destructure it:

```suggestion
const { as: ElementType } = this.props

      <ElementType>
```

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
layershifter: It also doesn't match definition, the correct signature is:

```suggestion
  onClick?: (e: React.SyntethicEvent, props: GridPickerItemProps) => void
```

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+}
+
+const imageButtonStyles = {
+  minWidth: '56px',
+  height: '56px',
+  padding: '0',
+  background: '#fff',
+}
+
+class GridImagePickerItem extends React.Component<GridPickerItemProps, any> {
```
layershifter: Please don't use `any` in `Component` generic, it can cause issue in the future

```suggestion
class GridImagePickerItem extends React.Component<GridPickerItemProps> {
```

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+  static defaultProps = {
+    as: 'li',
+  }
+
+  handleClick = e => {
+    if (e.target) {
+      const btn = e.target.nodeName !== 'IMG' ? e.target : e.target.parentElement
+      const selectedItem = btn && btn.title
+      selectedItem && alert(`Selected item is: ${selectedItem}`)
+    }
```
layershifter: _May be we can pass something as `data` attribute?_

**NOT VALID**

```jsx
handleClick = (e, props) => {
  const title = props['data-title']
  // Other goes there
}

render() {
  return <Button data-title='COOL TITLE' />
}
```

docs/src/prototypes/popups/helper.ts
```diff
@@ -0,0 +1,73 @@
+import * as _ from 'lodash'
+
+export const getItemsData = (images, ariaLabelPart) => {
```
layershifter:
```suggestion
export const getItemsData = (images: string[], ariaLabelPart: string) => {
```

Can we add types?

docs/src/prototypes/popups/stickerPicker.tsx
```diff
@@ -0,0 +1,83 @@
...
+      ? this.state.activeMenuIndex
+      : 0
+    return getItemsData(arrayOfStickerImagesNames[index], 'sticker of')
+  }
+
+  onMenuItemClick = (e, props) => {
+    this.setState({ activeMenuIndex: props.index }, () => {
+      const gridPickerElement = ReactDOM.findDOMNode(this.gridPickerRef.current) as HTMLElement
+      const input = gridPickerElement && gridPickerElement.querySelector('input')
+      input && input.focus()
```
layershifter: Sorry, but it's not cool. If we want to focus `input` we can make it in class method:

```jsx
class GridImagePicker {
  focusInput() {}
}
```

```suggestion
      this.gridPickerRef.current.focusInput()
```

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
sophieH29: ```title``` attribute is needed because that is read by screen reader. And ```handleSelection``` it is just to show that something happened when you click or press Enter / Space on an image.

layershifter: Nevermind, please. It was stupid idea at all. Another proposal, why we can't just do this?

```diff
-  handleClick = e => {
-    if (e.target) {
-      const btn = e.target.nodeName !== 'IMG' ? e.target : e.target.parentElement
-      const selectedItem = btn && btn.title
-      selectedItem && alert(`Selected item is: ${selectedItem}`)
-    }
+ handleClick = (e, props) => {
+   const { title } = props
+
+   alert(`Selected item is: ${title}`)
+ }
```

We are passing a `title` prop to `Button`, so we will get it in `onClick`.

docs/src/prototypes/popups/GridImagePicker/GridImagePicker.tsx
sophieH29: is it some best practice of how to arrange imports?

layershifter: Yep, imports in SUIR are arranged in the same manner

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
sophieH29: ha, you're totally right, I copied that handler from ```helper.ts``` and didn't figure it out. Thank you!

layershifter (approved):

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const imageStyles = {
+  width: '100%',
+}
+
+const imageButtonStyles = {
+  minWidth: '56px',
+  height: '56px',
+  padding: '0',
+  background: '#fff',
```
codepretty, line 17: How do you handle themeing (dark / contrast) for values like this? Or is that not something we worry about in prototypes?

sophieH29: yes, exactly, this prototype just shows how things can be combined together from stardust components to build a new functional and accessible experience, so I think we can skip on theming here.

docs/src/prototypes/popups/GridImagePicker/GridImagePicker.tsx
```diff
@@ -0,0 +1,71 @@
+import { Grid, Input, gridBehavior } from '@stardust-ui/react'
+import * as React from 'react'
+import * as _ from 'lodash'
+
+import { ShorthandValue } from '../../../../../types/utils'
+import GridImagePickerItem, { GridPickerItemProps } from './GridImagePickerItem'
+
+export interface GridPickerProps {
+  as?: keyof React.ReactHTML
```
mnajdova, line 9: Is there a strong reason why we support here just the HTML react elements?

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
```diff
@@ -0,0 +1,46 @@
+import { Image, Button } from '@stardust-ui/react'
+
+import * as React from 'react'
+import * as _ from 'lodash'
+
+export interface GridPickerItemProps {
+  as?: keyof React.ReactHTML
```
mnajdova, line 7: Same question as above.

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
```diff
@@ -0,0 +1,46 @@
...
+  handleClick = e => {
+    _.invoke(this.props, 'onClick', e, this.props)
+  }
+
+  render() {
+    const { title, imageSrc, as: ElementType } = this.props
+
+    return (
+      <ElementType>
+        <Button styles={imageButtonStyles} onClick={this.handleClick} title={title} role="listitem">
```
mnajdova: `const { onClick } = this.props`
Then on the Button, you can just use `onClick={onClick}`
No need for additional handler if you are just invoking the onClick from the props there.

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export interface GridPickerItemProps {
+  as?: keyof React.ReactHTML
+  title?: string
+  imageSrc: string
+  onClick?: (e: React.SyntheticEvent, props: GridPickerItemProps) => void
+}
+
+const imageStyles = {
+  width: '100%',
```
mnajdova: Have you tried using the fluid prop for the image, instead of these styles? I think it should do the same thing.

docs/src/prototypes/popups/helper.ts
```diff
@@ -0,0 +1,77 @@
+import * as _ from 'lodash'
```
mnajdova: Would rename this to dataMocks, or anything else more related to `data`. The name helpers associate more to some utils function, at least in my opinion. What do you think?

docs/src/prototypes/popups/stickerPicker.tsx
```diff
@@ -0,0 +1,80 @@
...
+]
+
+class StickerPicker extends React.Component {
+  state = { activeMenuIndex: 0 }
+  gridPickerRef = React.createRef<GridImagePicker>()
+
+  getStickersData = () => {
+    const index = arrayOfStickerImagesNames[this.state.activeMenuIndex]
+      ? this.state.activeMenuIndex
+      : 0
```
mnajdova: The state is initialized with the value of 0, so no need for adding this check.

mnajdova (commented):
Left some comments. In general, it's close to be merged.

docs/src/prototypes/popups/helper.ts
sophieH29: agree

docs/src/prototypes/popups/GridImagePicker/GridImagePicker.tsx
sophieH29: There is no strong reason for that. For now, it is just prototype + I wanted to constrain it to type, not to make ```as?: any``` as we have in Stardust components.

docs/src/prototypes/popups/GridImagePicker/GridImagePickerItem.tsx
sophieH29: I used ```fluid```, you're right, this one is redundant

docs/src/prototypes/popups/stickerPicker.tsx
sophieH29: agree

docs/src/prototypes/popups/GridImagePicker/GridImagePicker.tsx
mnajdova: Okay, let's leave it like this for now. When we will have all createComponent functionality available on the public API we will be able to reuse this from the Stardust library.

mnajdova (approved):

PR #481, opened 2018-11-15
fix(FocusTrapZone): Move comments from propTypes to props

sophieH29:


mnajdova (approved):
Approved, thanks!

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -35,61 +35,15 @@ export class FocusTrapZone extends React.Component<FocusTrapZoneProps, {}> {
   private createRef = elem => (this._root.current = ReactDOM.findDOMNode(elem) as HTMLElement)
 
   static propTypes = {
-    /**
-     * Element type the root element will use. Default is "div".
-     */
     as: customPropTypes.as,
```
mnajdova, line 38: Check the existing interfaces for the props and the common prop types if you can reuse some of those, before merging

sophieH29: there is only ```as``` in common, so I merge it as it is

PR #482, opened 2018-11-15
[RFC] Move props to components' types file

sophieH29:
**Divide and conquer!**

This RFC on how we can re-organize the structure of our component and make it easier to read.
Our component's props are growing and also with comments around, it can get harder to grasp quickly the main logic and read the code.

I found for myself it's much easier when props are in separate ```*.types.ts``` file than when having everything in one file, kind of mix of many things.

I think it is more personal, depending on how you used to work. But I would love to hear your thought on it.
Currently, I made changes for ```Button``` and ```ButtonGroup```, and if you like it, I can make this change for every component.


layershifter:
I'm going to downvote this change because it will complicate our docgen tasks:
- we will need to watch for two files (`Component.tsx`, `Component.types.ts`) instead of one
- it will require more work from `react-docgen-typescript` that also need to resolve these files, it will beat our performance
- it's not a common practice

### `build:docs:docgen`

| Framework | Time |
| - | - |
| Semantic UI React | 12 s |
| Stardust UI | 36 s |

And will take more time if we will separate definitions. 

---

However, I think that we can move types to bottom of component files.

levithomason:
Agree on following common practice:

- keep typings in the component file
- do not put them at the bottom 🤓 

Editors can collapse blocks.  Devs should use that for navigating around.  There are other tooling options we can talk about offline.  Let's do that.

Note that for consumers separate `*.d.ts` files are also generated, so this is not a consumer facing issue either. As always, open to future changes around this if needed (such as sharing types for the specifications repo to allow themes/etc to use them independent of the react package).

PR #485, opened 2018-11-16
build(docs): use caching for `build:docs:docgen`

layershifter:
I've implemented a simple caching based on MD5 checksum for the `build:docs:docgen` task.

If a component's source is not changed and definitions exist we just skip the generation process for a file. An example of cached run:

```
[15:12:11] Requiring external module ts-node/register
[15:12:14] Using gulpfile ~\WebstormProjects\react\gulpfile.ts
[15:12:14] Starting 'build:docs:docgen'...
[15:12:14] Finished 'build:docs:docgen' after 149 ms
Done in 5.03s.
```

### DIFF
--- a/build/gulp/plugins/gulp-react-docgen.ts
+++ b/build/gulp/plugins/gulp-react-docgen.ts
@@ -19,7 +19,7 @@ export default () =>
       return
     }
 
-    const infoFilename = file.basename.replace(/tsx$/, 'info.json')
+    const infoFilename = file.basename.replace(/\.tsx$/, '.info.json')
     const nextChecksum = getBufferChecksum(file.contents)
 
     if (getInfoChecksum(infoFilename) === nextChecksum) {
--- a/build/gulp/plugins/gulp-react-docgen.ts
+++ b/build/gulp/plugins/gulp-react-docgen.ts
@@ -3,7 +3,7 @@ import * as path from 'path'
 import * as through2 from 'through2'
 import * as Vinyl from 'vinyl'
 
-import { getComponentInfo } from './util'
+import { getComponentInfo, getBufferChecksum, getInfoChecksum } from './util'
 
 const pluginName = 'gulp-react-docgen'
 
@@ -19,11 +19,18 @@ export default () =>
       return
     }
 
-    try {
-      const contents = getComponentInfo(file.path)
+    const infoFilename = file.basename.replace(/tsx$/, 'info.json')
+    const nextChecksum = getBufferChecksum(file.contents)
 
+    if (getInfoChecksum(infoFilename) === nextChecksum) {
+      cb(null)
+      return
+    }
+
+    try {
+      const contents = getComponentInfo(file.path, nextChecksum)
       const infoFile = new Vinyl({
-        path: `./${file.basename.replace(/tsx$/, 'info.json')}`,
+        path: `./${infoFilename}`,
         contents: Buffer.from(JSON.stringify(contents, null, 2)),
       })
 
--- a/build/gulp/plugins/util/checksumUtils.ts
+++ b/build/gulp/plugins/util/checksumUtils.ts
@@ -1,0 +1,21 @@
+import * as fs from 'fs'
+import * as crypto from 'crypto'
+
+import config from '../../../../config'
+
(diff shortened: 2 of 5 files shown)


build/gulp/tasks/docs.ts
```diff
@@ -53,7 +49,6 @@ task('clean:docs:example-menus', cb => {
 task(
   'clean:docs',
   parallel(
-    'clean:docs:component-info',
```
layershifter, line 56: We shouldn't clear directory for caching purposes

build/gulp/plugins/gulp-react-docgen.ts
```diff
@@ -19,11 +19,18 @@ export default () =>
       return
     }
 
-    try {
-      const contents = getComponentInfo(file.path)
+    const infoFilename = file.basename.replace(/tsx$/, 'info.json')
```
kuzhelov: would suggest to make expression more explicit by using dot at start:

```js
 const infoFilename = file.basename.replace(/\.tsx$/, '.info.json')
```

build/gulp/plugins/gulp-react-docgen.ts
```diff
@@ -19,11 +19,18 @@ export default () =>
...
+    const nextChecksum = getBufferChecksum(file.contents)
 
+    if (getInfoChecksum(infoFilename) === nextChecksum) {
```
kuzhelov, line 25: here (or as part of `getInfoChecksum `) it would be great if we'd introduce simple check to prevent reads from files that we know for certain are not component info files, i.e. ones that end with `.info.json` - that will reduce amount of disk IO operations

build/gulp/plugins/util/checksumUtils.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import config from '../../../../config'
+
+export const getInfoChecksum = (filename: string): string | null => {
+  try {
+    const buffer = fs.readFileSync(config.paths.docsSrc('componentInfo', filename))
+    const componentInfo = JSON.parse(buffer.toString())
+
+    return componentInfo.checksum
+  } catch (e) {}
```
kuzhelov, line 12: just to be sure - could you, please, suggest known cases this `catch` targets? Is it possible that some file other than component's info will be passed to this method in our workflow?

layershifter: `readFileSync` throws an error if file not exists, however it's a valid case for the first run

build/gulp/plugins/gulp-react-docgen.ts
layershifter: I don't understand what you want to do there. `getInfoChecksum()` will be always called for `*.info.json`, how are you planning to reduce reads?

build/gulp/plugins/gulp-react-docgen.ts
layershifter: Agree, will be done tomorrow

build/gulp/tasks/docs.ts
miroslavstastny: So there is no way how to clear the cache (other than deleting the files manually)? :-(

layershifter: Yep. We can make a separate script (in `package.json`) for this, delete DLL, doc definitions, clear dist), something like `yarn cleanup`. What do you think?

miroslavstastny: ...and we can do that in a separate PR

miroslavstastny (approved):

build/gulp/plugins/gulp-react-docgen.ts
kuzhelov: the reason I've asked is because seen `try/catch` block in `checksumUtils.ts` - have clarified which cases we are capturing there. If this is only about `non-existing file`, then we are fine :)

build/gulp/plugins/util/checksumUtils.ts
kuzhelov: got it, thanks!

kuzhelov (approved):

PR #487, opened 2018-11-19
fix(handleRef): handle `null` ref correctly

layershifter:
A small fix for the `handleRef` ref utility.

```js
typeof null // "object"
```

There is a similar condition in React's code: https://github.com/facebook/react/blob/e58ecda9a2381735f2c326ee99a1ffa6486321ab/packages/react-reconciler/src/ReactFiberCommitWork.js#L625

### DIFF
--- a/src/lib/handleRef.ts
+++ b/src/lib/handleRef.ts
@@ -20,7 +20,7 @@ const handleRef = <N>(ref: React.Ref<N>, node: N) => {
     return
   }
 
-  if (typeof ref === 'object') {
+  if (ref !== null && typeof ref === 'object') {
     // @ts-ignore The `current` property is defined as readonly, however it's a valid way because
     // `ref` is a mutable object
     ref.current = node
--- a/test/specs/lib/handleRef-test.ts
+++ b/test/specs/lib/handleRef-test.ts
@@ -18,6 +18,14 @@ describe('handleRef', () => {
     expect(ref).toBeCalledWith(node)
   })
 
+  it('does not do anything when "ref" is null', () => {
+    const node = document.createElement('div')
+
+    expect(() => {
+      handleRef(null, node)
+    }).not.toThrowError()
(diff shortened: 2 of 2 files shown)


layershifter:
Should I add a record to `CHANGELOG.md` for this?

src/lib/handleRef.ts
```diff
@@ -20,7 +20,7 @@ const handleRef = <N>(ref: React.Ref<N>, node: N) => {
     return
   }
 
-  if (typeof ref === 'object') {
+  if (ref !== null && typeof ref === 'object') {
```
kuzhelov: could you, please, suggest when this condition is critical for us to check? What is the case that could lead to that? Is it some sort of 'race condition' during component's unmount that we would like to prevent with that?

Thing that I am worrying about is that it could lead to memory leaks. Suppose that there is a parent component that will capture a dom element of its child component:

```jsx
<Parent>
  <Child innerRef={domNode => this.childDomNode = domNode} />
</Parent>
```

In that case `Parent` might never update `this.childDomNode` to `null`, even if `Child` will be unmounted (without this check, previously, `this.childDomNode` will be updated to `null`).

kuzhelov (approved):

layershifter (commented):
The easiest case if when we will try use `cloneElement` and try to catch a component's ref:

```jsx
const htmlInput = <input />

// ----

cloneElement(htmlInput, {
  ref: (node) => {
    // 1. Calling existing handler
    // A. We need to keep user's behaviour working, for example for 3rd party libs
    // B. `ref` is not a prop in React, so when `ref` is not passed it's `null` by default 
    //     and here it will fail without this fix
    handleRef(child.ref, node)
    // 2. Calling our handler
    this.handleInputRef(node)
  }
})
```

So, this fix is for a `handler` ref, not for a `value`.

PR #491, opened 2018-11-19
feat(Ref): support of `forwardRef()` API

layershifter:
This PR is a next stage for #417.

---

`findDOMNode()` is deprecated in `StrictMode`, [ref](https://reactjs.org/docs/react-dom.html#finddomnode). and we can't rely on it, it also forces us to use `react-dom` that means that it will be impossible to use it with React Native.

This PR is redesign of `Ref` component to use `forwardRef` API by default and use `findDOMNode()` only as a fallback. If React's team will decide to remove `findDOMNode()` it will be not issue for us.


docs/src/examples/components/Ref/Types/RefExampleForwardRef.tsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react'
+import { Grid, Ref, Segment } from '@stardust-ui/react'
+
+const ExampleButton = React.forwardRef<HTMLButtonElement>((props, ref) => (
```
layershifter: The usage of generic in docs :( It's impossible to omit it there.

package.json
```diff
@@ -90,6 +90,7 @@
     "lodash": "^4.17.10",
     "prop-types": "^15.6.1",
     "react-fela": "^7.2.0",
+    "react-is": "^16.6.3",
```
layershifter, line 72: **(!)** A new dependency.

src/components/Ref/Ref.tsx
```diff
@@ -30,15 +41,39 @@ export default class Ref extends React.Component<RefProps> {
...
 
+  static getDerivedStateFromProps(props: RefProps) {
+    const child = React.Children.only(props.children)
+
+    return {
+      child,
+      isForward: isForwardRef(child),
+    }
+  }
+
```
layershifter, line 23: I'm not happy that we have the `state` there, however without it we will need to duplicate checks in `componentDidMount` and `componentWillUnmount`.

I'm opened for better ideas 🙀

package.json
```diff
@@ -113,8 +115,8 @@
     "cross-env": "^5.1.4",
     "doctoc": "^1.3.0",
     "doctrine": "^2.0.0",
-    "enzyme": "^3.1.0",
-    "enzyme-adapter-react-16": "^1.0.1",
+    "enzyme": "^3.7.0",
```
mnajdova: I guess these changes should be part of https://github.com/stardust-ui/react/pull/493

layershifter: Of #492 👍 I will rebase this PR after it will be merged, it's in TODO 😺

src/components/Ref/Ref.tsx
bmdalex: extract duplicate code in a private method and call that method from both `componentDidMount` and `componentWillUnmount`?

layershifter: We need check for `isForwardRef(child)` in render, too.

layershifter:
We can separate this logic obviously to two subcomponents `RefForwardRef` and `RefFindDOMNode`, it will be very clean. But it will add one more level of nesting to React's tree.

/cc @miroslavstastny

package.json
kuzhelov: 👍

docs/src/examples/components/Ref/Types/RefExampleForwardRef.tsx
kuzhelov: yes, and this is definitely not the problem of this PR :) seems that we should use JS engine to process this code, as docs are aimed to present **JS** variant of examples to the client (this is the case for now, at least)

layershifter: I'm planning to left this thing as is and push https://github.com/stardust-ui/react/issues/495 that will resolve all our issues.

docs/src/examples/components/Ref/Types/RefExampleForwardRef.tsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react'
```
kuzhelov: maybe name like `RefForwardingExample` would be more readable and easier to grasp

src/components/Ref/Ref.tsx
```diff
@@ -30,15 +41,39 @@ export default class Ref extends React.Component<RefProps> {
...
   }
 
   componentWillUnmount() {
-    handleRef(this.props.innerRef, null)
+    if (!this.state.isForward) {
+      handleRef(this.props.innerRef, null)
+    }
+  }
+
+  private handleRefOverride = (node: HTMLElement) => {
```
kuzhelov: according to ref's docs: https://reactjs.org/docs/refs-and-the-dom.html

> The function receives the **React component instance** or HTML DOM element as its argument, which can be stored and accessed elsewhere.

The parameter type used in this function leads to false assumptions.

docs/src/examples/components/Ref/Types/index.tsx
```diff
@@ -15,6 +15,15 @@ const RefTypesExamples = () => (
       }
       examplePath="components/Ref/Types/RefExampleRef"
```
kuzhelov: the name of this file also can be simplified - `RefExample` should be just enough

docs/src/examples/components/Ref/Types/RefExampleForwardRef.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { Grid, Ref, Segment } from '@stardust-ui/react'
+
+const ExampleButton = React.forwardRef<HTMLButtonElement>((props, ref) => (
+  <div>
+    <button {...props} ref={ref} />
+  </div>
+))
+
+class RefExampleForwardRef extends React.Component {
```
kuzhelov: actually, very confused by trying to understand how `Ref` component is helpful in case of ref forwarding (and this, as stated in the description, is its main purpose: `This PR is redesign of Ref component to use forwardRef API by default and use findDOMNode() only as a fallback.`)

Let me explain resaons for this by following the example. Suppose that I am a Stardust client and have the following component: 

```jsx
const MyButton = React.forwardRef((props, ref) => (
  <div>
    <button {...props} ref={ref} />
  </div>
))
```

With this component I already expect the following to be true:

```jsx
// when mounted, this.ref.current will refer to <button />
<MyButton ref={this.ref} />
```

Then comes Stardust and proposes to use `Ref` - with no clear benefits to me, as now it will be either:
- I will use `Ref` as **_another_** wrapper - but what are the benefits of that :/ ?
```jsx
const WithRef = (props) => (
  <Ref innerRef={props.forwardedRef}>
    { React.forwardRef((props, ref) => <MyButton ref={ref} />
  </Ref>
)
```
- or what is the scenario where I might need it? 

----------------

**`Ref`, in case of ref forwarding case, acts as a simple `ref` translator** - so, what are the reasons I wouldn't use built-in React ref forwarding capabilities and would prefer `Ref` component abstraction instead?

kuzhelov (dismissed):
Have expressed points of general concerns/questions related to the approach.

Also, to me it seems wrong that we are exposing different semantics for `innerRef` for these cases:
- forward ref case
```jsx
// element can be React Component reference (this) or DOM element
<Ref innerRef={ref => .. }>
  ...
</Ref>
```
- findDOMNode() case
```jsx
// DOM element is always guaranteed to be returned (and never React component ref)
// semantics is absolutely different from react ref
<Ref innerRef={domNode => ..}>
  ...
</Ref>
```

Consider the following potential implementation for render method of some component - you would never able to tell which logic will be fired:

```jsx
<Ref innerRef={ (?) => ... }>
  <SomeComponent />
</Ref>
```

Have expressed these concerns earlier here: https://github.com/stardust-ui/react/pull/459#discussion_r234181558. Also, React docs have warnings even about **just** introducing ref forwarding (https://reactjs.org/docs/forwarding-refs.html) - this is clearly might be a breaking change; and with this implicit logic switch we will make problem even worse:

image

docs/src/examples/components/Ref/Types/index.tsx
layershifter: Yep, I've seen that Stardust uses different naming for examples.

src/components/Ref/Ref.tsx
layershifter: See my comment below, we will enforce user to pass correct refs.

layershifter:
@kuzhelov you're absolutely right there, but!

We should avoid `findDOMNode`:
- it is deprecated in `StrictMode` and can be removed
- causes performance degrades

---

Yes, `Ref` component will become a proxy that checks that `forwardRef()` is used.

Why we need it at all? This will allow us to avoid this logic in every component. There are no guaranty that uses will pass us `forwardRef()` in `as` prop.

I understand that a forwarded ref can match to a component instance, so my proposal there: is throw warning in DEV env.

```js
  if (isDev) {
    if (node is not HTMLElement) {
      console.warn('HEY, your forwarded ref does not match to DOM node') 
    } 
  }

 handleRef(this.state.child.ref, node)
 handleRef(this.props.innerRef, node)
```

---

About naming, may be we can try `NodeResolver`, there is a similar package, https://github.com/jossmac/react-node-resolver.

kuzhelov:
@layershifter, let me, please, first elaborate on each of the statements you've made in this post: https://github.com/stardust-ui/react/pull/491#issuecomment-440334597 - in order of their appearance, and then conclude with summarising my thoughts.

-------------

> We should avoid findDOMNode:
> - it is deprecated in StrictMode and can be removed
> - causes performance degrades

Agree, but lets consider the problem that previously was solved by `Ref` in Stardust (this is very important, because alternative solution will need to match these criterias as well)
- `Ref` has allowed to abstract away slot's content from parent component, so that parent component might reliably set, say, ARIA attributes to slot's content _irregardless of what this content is represented with - whether it is stateless/stateful component, string (HTML) component, etc

Frankly, I don't see how we could achieve the same functionality by using ref/ref forwarding in case we won't, as before, introduce any serious restrictions on slot's content like the following ones:
- it shouldn't be a stateless component (as refs won't work there)
- it should be either
  - stateful component
  - component that is created by React.forwardRef
- an all the cases **ref should guarantee to be eventually forwarded to the nearest DOM node**

The later one is absolutely crazy term in terms of how strict it is. Suppose that you have written you library of components **where have already established some policies around ref forwarding** - and now, if you would like to integrate them to Stardust, you will need to _completely_ rewrite this ref forwarding logic! While, in contrast, with `Ref` based on `findDOMNode()` it was completely seamless process before.

--------

Speaking of your comment related to `Ref` vs using ref forwarding directly:

> Why we need it at all? This will allow us to avoid this logic in every component. There are no guaranty that uses will pass us forwardRef() in as prop.

**Could you, please, provide an example of that**, just for me to be sure that we are talking about the same case. Also, am very curious **how we are going to solve the problem** of abstracting away slot's content from parent Stardust component **with ref forwarding**, so that any of the following will be safely possible (general case for a problem):

```jsx
// suppose that we need to subscribe slot's DOM element on events, 
// as well as provide ARIA attributes to it.
//
// now, with Ref (and findDOMNode()) this is possible to achieve
<StardustComponent slot={'div'} />

// can be stateful/stateless, no matter
<StardustComponent slot={<SomeArbitraryComponent />} />
```

because to me it seems that, in case if slot's component doesn't provide any ref forwarding logic as part of its implementation (where ref is _required_ to be ultimately forwarded to the DOM element), then we won't be able to abstract `StardustComponent`'s implementation from slot's content.

---------

> I understand that a forwarded ref can match to a component instance, so my proposal there: is throw warning in DEV env.

Unfortunately, warning are quite prone to not be noticed - especially given that in our library we are often using Ref to achieve the following goals:
- provide ARIA attributes to DOM node
- provide event handler subscriptions

These actions, in general, do not result in some consequences that are immediately visible.

More than that - even if we will ensure that this rule is honored for Stardust components, given our 'slot'-feature where client is able to provide _any_ component to the slot, we won't be able to guarantee that _in client's code_ this rule will be satisfied, and we are returning back to the same problem of warnings being unnoticed on the client's side.

We might consider to throw errors then - but don't see it as a solution as well, as it could result in app that will be suddenly broken for the client after some slot's component update (like HOC being introduced on top of previously used plain 'div').

------

## To conclude,
I have reviewed related discussions around this topic - started, obviously, from this one:
https://github.com/yannickcr/eslint-plugin-react/issues/678, where the same concerns expressed by the following comments (note, those are quite similar concerns and those are not addressed by Dan):
- https://github.com/yannickcr/eslint-plugin-react/issues/678#issuecomment-232228192
- https://github.com/yannickcr/eslint-plugin-react/issues/678#issuecomment-232244116
- https://github.com/yannickcr/eslint-plugin-react/issues/678#issuecomment-419403757

Additional related concerns raise during another discussion:
- https://github.com/reagent-project/reagent/issues/251#issuecomment-286428744
- https://github.com/reagent-project/reagent/issues/251#issuecomment-286467147

Should say explicitly that I totally support the intent to remove `findDOMNode()` from our code - at the same time, to me it is not clear now how alternative approach will be able to solve the same set of tasks (of critical importance) for us as it was previously done by `findDOMNode()` approach. 

@layershifter, would like to hear your comments on that - and it would be absolutely great if they will come along with examples :)

layershifter:
@kuzhelov it seems that you don't hear me :) Let's start from the beginning.

### Axiom 1: We need `ref`s to DOM nodes

The `Ref` component in the current state allows us to get DOM nodes and manipulate them.

### Axiom 2: `findDOMNode` is deprecated

> Warning: findDOMNode is deprecated in StrictMode.
https://codesandbox.io/s/w2jm4j4n37

We can't rely on deprecated API even it's only in `StrictMode`. You can say that it's not serious warning, but `StrictMode` also claims about [unsafe lifecycles](https://reactjs.org/docs/strict-mode.html#identifying-unsafe-lifecycles) and _my opinion that these lifecycles will be removed from React 17_.

For us this means that we can suddenly receive nothing while in the same time we heavily rely on `findDOMNode`.

### Axiom 3: `findDOMNode` is irreplaceable

We don't have any replacement for this API with the same behaviour.

---

Do you agree? 😃

kuzhelov:
@layershifter,

agree on all the axioms 👍, just don't agree on the following
>  it seems that you don't hear me

Would be great if you'd provide specific example in the discussion where this was the case, and let me apologise for that then - as always tried to use either your code or your statements as basis for the discussion.

layershifter:
Cool, let's continue 😺 

### Axiom 4: `forwardRef` is the single solution for refs control

There is only one solution that allows us to do something with refs, it's `forwardRef()` API.

> You can instead make this explicit by pass a ref to your custom component and pass that along to the DOM using ref forwarding.

https://reactjs.org/docs/strict-mode.html#warning-about-deprecated-finddomnode-usage

----

That was the last axiom, the below items are suggestions.

----

### Input 1: we will use `forwardRef()`

PR #509 brings support of `forwardRef()` to all our components. This means that all `ref`s of our components will point to DOM nodes, _except, some edge cases._

```jsx
<Button ref={node => console.log(node)} /> // will point to `button` element
<Input ref={node => console.log(node)} /> // will point to `input` element
```

So, this means that code below will work, too:

```jsx
<Ref innerRef={node => console.log(node)}>// will point to `button` element
  <Button />
</Ref>
```

### Input 2: we will still use `findDOMNode()`

If a component is not wrapped by `forwardRef()` we will use `findDOMNode()` as **fallback** (`forwardRef()` _will be a default option_):

```jsx
const SFC = () => <div />
class Classy extends React.Component {
  render() {
    return <div />
  }
}

// ---

<Ref innerRef={node => console.log(node)}>// will point to `div` element
  <SFC />
</Ref>
<Ref innerRef={node => console.log(node)}>// will point to `div` element
  <Classy />
</Ref>
```

layershifter:
The inputs above are very cool and everything is awesome, but let's go to cons.

----

### Cons 1: `forwardRef()` is not a guaranty

There no any guarantee that `forwardRef()` will forward us to a valid DOM element, it can move us to a functional or class component.

Sandbox: https://codesandbox.io/s/9z826l61xr

#### Case 1: Functional

```jsx
const ForwardSFC = React.forwardRef((props, ref) => <OtherSFC ref={ref} />)

// ---

<Ref innerRef={node => console.log(node)}>// will be `null` and produce React's warning
  <ForwardSFC />
</Ref>
```

_I should note that a warning will be produced by React, before a `Ref` component will be rendered._

In fact, this case is not valid at all because you will never forward refs to a functional component, this doesn't make any sense.

#### Case 2: Class

```jsx
const ForwardClassy = React.forwardRef((props, ref) => <Classy ref={ref} />)

// ---

<Ref innerRef={node => console.log(node)}>// will be an instance of `Classy` component!
  <ForwardClassy />
</Ref>
```

It's a possible case, our user can do this and it will be an edge case for us because we will receive an instance of `Classy` instead of DOM node.

But, let's stop a for a second. Just try to form a question: _Who will use the API whose task is to send refs to DOM nodes?_ I understand that it's a possible case, but is it a valid case? Especially when you will try to pass such component to the library.

layershifter:
### Pros 1: More control

The usage of `forwardRef()` will allow us to give more control to our users:

```jsx
const ExternalButton = () => (
  <div>
    <button>Some cool text</button>
  </div>
)
const ForwardedButton = React.forwardRef((props, ref) => (
  <div>
    <button ref={ref}>Some cool text</button>
  </div>
))

//

<Ref innerRef={node => console.log(node)}>// will point to `div` element, WAT???
  <ExternalButton />
</Ref>
<Ref innerRef={node => console.log(node)}>// will point to `button` element, WOW cool!
  <ForwardedButton />
</Ref>
```

layershifter:
### Final proposal

I want to go with a following behaviour:

```jsx
// Option 1. Functional component
// Use: `findDOMNode()`
// Result: DOM node
<Ref>
  <SFC />
</Ref>

// Option 2. Class component
// Use: `findDOMNode()`
// Result: DOM node
<Ref>
  <ClassComponent />
</Ref>

// Option 3. Forward ref to a DOM node
// Use: `forwardRef()`
// Result: DOM node
<Ref>
  <ForwardedSFCComponent />
</Ref>

// Option 4. Forward ref to a class component
// Use: `forwardRef()`
// Result: Class instance
<Ref>
  <ForwardedClassComponent />
</Ref>
```

Sandbox: https://codesandbox.io/s/2ovp15zrj0

As I see Downshift also [allows](https://github.com/paypal/downshift/pull/612/files#diff-e5a2a6cc790a2db69edd4cddc1043d4dR1160) to use `forwardRef()`.

----

@kuzhelov I hope that my idea is more clear now. Yes, we have a case `ForwardedClassComponent`, but in other cases everything will go well.

I want to hear @levithomason opinion, may be we don't need this at all and it's an overcomplicated thing.

levithomason:
Don't log "Result: null and warning Please forward your ref to a valid DOM node" just return the class.  Let's merge and iterate.

kuzhelov (commented):
Still, am disagree with the mixed approach taken, as we are 
- not alleviating the issue with `findDOM`, as we are still relying on that for certain cases
- introduce additional entropy to the set of types that might be returned in the `innerRef`, say, callback - now it is not guaranteed to be a DOM node

However, let me, please, reset my vote for unblocking this PR, as it was decided yesterday.

src/components/Ref/Ref.tsx
```diff
@@ -30,15 +41,39 @@ export default class Ref extends React.Component<RefProps> {
...
+
+    return {
+      child,
+      isForward: isForwardRef(child),
+    }
+  }
+
   componentDidMount() {
-    handleRef(this.props.innerRef, findDOMNode(this))
+    if (!this.state.isForward) {
```
miroslavstastny: I am afraid the conditional login here can surprise user of the component. I would rather prefer not to be smart here and introduce two separate `Ref` components - one for forward ref, the other one with `findDOMNode`.

layershifter: Cool, this is my initial idea, https://github.com/stardust-ui/react/pull/491#issuecomment-440194159

Will do

kuzhelov:
### TL;DR

By making checks against the most popular component libraries have found out that the satement we've made

> Ref forwarding to DOM node should be supported by third-party components. If this is not the case, then it is a problem of third-party component.

**doesn't stay for any of them (Material UI, React Bootstrap, Ant Design)**.

Ref forwarding mechanism is not supported by **_any_** of them for the provided components - in other words, if we will remove `findDOMNode` from our API, **there won't be any chance for us to use third-party components in our component slots, given our ref-forwarding approach taken**. More than that, currently even the following cases (_which previously were working_) will result in broken behavior with updated `Ref`'s logic:

```jsx
// note, this is the general principle that Wrapper component may implement 
// with the intent to auto forward ref
const StyledButton = React.forwardRef((props, ref) => <MaterialUIButton ref={ref} style={...} {...props} />

<Popup
  trigger={<StyledButton />}    // !!! NO EVENT HANDLERS BOUND - but previously worked fine !!!
  ..
/>
```

## Detailed story

Final thoughts that I'd like to share, those relate to the following point we've made to argue necessity to introduce ref forwarding to `Ref` component:

> 'Proper' Ref forwarding should be supported by third-party components, so that refs provided would always resolve to the DOM element. If this is not the case, then it is a problem of third-party component.

For the sake fo argument I've considered the following component libraries to make checks on - was selecting by 'quite popular ones' principle:
- Material UI
- React Bootstrap
- Ant Design

For each of them I've tried to see what `ref` will be resolved to for the components they expose - because, **as we were claiming, if `ref` is resolved to DOM node, then it is not our problem, it is a problem of component (or third-party component library in our case)**.

Results are provided below.

### Material UI (GH likes: 8,4K) ###

```jsx
import Button from '@material-ui/core/Button';
....
<Button ref={node => console.warn(node)}>Hey</Button>
```

**RESULT: COMPONENT (not a DOM node)**
image




### React Bootstrap (GH likes: 14,2K) ###

```jsx
import Alert from 'react-bootstrap/lib/Alert'
....
<Alert ref={node => console.warn(node)} />
```

**RESULT: COMPONENT (not a DOM node)**
image



### Ant Design (GH likes: 36,4K) ###

```jsx
import { DatePicker } from 'antd'
....
<DatePicker ref={node => console.warn(node)} />
```
image


**RESULT: COMPONENT (not a DOM node)**

-------------

## Conclusion

With these results, I still see this step
- introducing much more _complexity_
- introducing _breaking change for `Ref` component_, as type of returned value in the callback is not defined now
- _doesn't introducing any clear value_ (I mean, what are the cases we would like to support? Ref forwarding with returning DOM nodes - but the we should provide an implementation that would still **_guarantee_** that DOM node is provided as `innerRef` callback arg, then I will understand that)
- based on this review - would _fail seamless integration with third-party components_

given all these points I really believe that we are making premature mistaken step here.

levithomason:
I think this analysis is too surface for two reasons.  GitHub stars are not a good indicator of the latest best practices or the best ideas.  The current state of a repo is not a good indicator of the current direction or views of the maintainers and contributors.

### Material UI

Note that Material UI provides a RootRef component that always returns a DOM node:
https://material-ui.com/api/root-ref.  This was done per the [guidance given by the React team](https://github.com/facebook/react/issues/11401#issuecomment-340543801).  

It is also worth noting that the primary voices in the ref discussion are the React Bootstrap maintainers. You can see there that their intent is that the Ref component pattern should always return a DOM node, not a class instance.  Note the use of `findDOMNode()` in the utility.

Also, if you read Material UI's [issues and PRs about Refs](https://github.com/mui-org/material-ui/issues?utf8=✓&q=ref+), you'll see they are struggling with the same problems we are.  Specifically, you can see the community issues complaining that their refs don't support forwardRef.  You can also see their PR where they've updated their refs to do so https://github.com/mui-org/material-ui/pull/11757.

Note, they check for an existing ref, whether the ref has a ref (ie it is a class), handle functional refs explicitly, and also handle forwardRef (ref.current) explicitly:

```jsx
handleRefInput = node => {
  this.input = node;

  let ref;

  if (this.props.inputRef) {
    ref = this.props.inputRef;
  } else if (this.props.inputProps && this.props.inputProps.ref) {
    ref = this.props.inputProps.ref;
  }

  if (ref) {
    if (typeof ref === 'function') {
      ref(node);
    } else {
      ref.current = node;
    }
  }
};
```

### React Bootstrap

Note that React Bootstrap also solve this problem by exposing props like [`inputRef`](https://react-bootstrap.github.io/components/forms/#forms-props-form-control) which also return the DOM nodes.  These are the refs you would need in these cases, DOM nodes that have APIs you sometimes need, not component instances.

***

The point I'm making is that on the surface your argument seems correct, however, if you dig deeper and allow more time and real-world usage to pass you will find that the useful thing about refs is getting DOM nodes.  User's don't want class instances and, as much as possible, they don't want special cases.

layershifter:
```jsx
const StyledButton = React.forwardRef((props, ref) => <MaterialUIButton ref={ref} style={...} {...props} />
```

@kuzhelov the provided case is fully invalid, you're forwarding ref to component that doesn't support this feature. It's a shot in the foot 💣 

---

Users want to have ref forwarding:
https://github.com/mui-org/material-ui/issues/13492#issuecomment-435449958
https://github.com/mui-org/material-ui/issues/10825#issuecomment-383318055

https://github.com/mui-org/material-ui/issues/13221 (issue marked as important)

And the same problems with deprecation:
https://github.com/mui-org/material-ui/issues/13394

----

https://github.com/facebook/react/pull/13841#issuecomment-429444800

kuzhelov:
@levithomason 

yes, but the key difference from our `Ref` is that these components are **guaranteeing** to return DOM node by their API - this is what I was talking about through all the comments to this PR: **our `Ref` now is not guaranteed to return DOM node anymore**

-------
Speaking of the analysis taken:

> This analysis is too surface

I was aimed to challenge the following statement: 

> _any_ third-party component should resolve `ref` to DOM node, and if it is not the case, then it is a component's problem

So, that was the motivation for the approach to analysis I've taken, and to my mind its results loudly prove that statement above doesn't reflect current state of things.

--------

@layershifter 

> the provided case is fully invalid, you're forwarding ref to component that doesn't support this feature. It's a shot in the foot

when this code is written by the same person (and, thus, entirely visible) - yes, it is. But consider the case when 
- wrapped component is client's one (`Foo`). She has used this component as Stardust's `Popup` trigger, and everything was working fine
- next day some `wrappers` third-party lib was considered by client to wrap her `Foo` component. This third-party lib is following ref-forwarding ideology, based on the assumption that `wrappers should forward refs and let wrapper components resolve them to DOM nodes`.

```jsx
// this what client's code now is:
import { wrapped } from 'wrappers'

const Foo = () => ...

export wrapped(Foo)
```

and, suddenly, after that everything has stopped to work. Is it clear for the client that she has done something that might shoot her leg? Don't think so

levithomason:
@kuzhelov it is not quite that solid.  The Material UI `inputRef` attempts to resolve nested refs by checking for `this.props.inputRef.ref`.  That would be an `inputRef` passed by the user which resulted in a class, that also had a `ref`.  However, Material has no control over whether or not the final `ref` obtained will resolve to a DOM node.  All they can do is call `ref()` if it is a function or handle `ref.current` if it is forwarded.  What the result of that will be is not in the control of the library since it is passed in by the user at runtime.  This is an infinite loop, there is no way to ensure that all components from all users will always use refs that resolve to DOM nodes.  Therefore, it is impossible to create a ref API that guarantees all refs will return DOM nodes every time.

***

I believe we are all saying the same thing in regard to the problems that are present.  Let's turn to solutions at this point.  What are the optimal solutions to these issues?

I would start with the statement that returning a class is not a solution, but a problem as well.  Our goals are:

- return the DOM node as often as possible, every time if possible
- support every form of ref API

Let's shift our focus to APIs and patterns for meeting these goals.

miroslavstastny (approved):
I vote for merging this. Let's iterate when we hit the wall.

kuzhelov:
@levithomason 

Speaking of Material's case provided here: https://github.com/stardust-ui/react/pull/491#issuecomment-444247750 - please, note that this is about `Input` component only, as well the fact that **Material doesn't support slot functionality, where any arbitrary client's component may be passed - they are handling ref forwarding for their own Input component**. Thus, they are doing pretty safe thing:
- they know what is the implementation of their `Input` component, and they _do know for certain_ what ref will be resolved to (it is always guaranteed to be the DOM node, as this is how this component is designed)
- thus, there are no exceptions in the promises of the docs where Material states that **DOM node is guaranteed to be always returned**

Note, this is completely different from the Stardust's case where there is possibility for **any** client component to be used as a slot content.

----------------

>  This is an infinite loop, there is no way to ensure that all components from all users will always use refs that resolve to DOM nodes.

There is, actually - and this will be about following forward refs chain and do `findDOMNode` at the final stage. 

Note that all the implementations we are considering from other libraries, as well as the one that is proposed by this PR - those still largely rely on the `findDOMNode` support, as there is no other reliable mechanism to fetch it. Thus, with that being acknowledged, the only thing that I am, in fact, proposing (and arguing about) is the API consistency - if we agree that `findDOMNode` cannot be removed from `Ref`'s implementation, we can utilize it then to provide this API consistency.

levithomason:
Thanks for the correction on my misreading of materials' input.  I think I've become a bit hasty in trying to close this issue.  Let's move on.

PR #492, opened 2018-11-19
chore(package): update `enzyme`

layershifter:
This PR updates `enzyme` and `enzyme-adapter-react-16`, they have correct support of `Fragment` and `forwardRef` APIs.


test/specs/components/Icon/Icon-test.tsx
```diff
@@ -33,7 +33,7 @@ describe('Icon', () => {
         const renderedComponent = mountWithProviderAndGetComponent(
           Icon,
           <Icon name="fontIcon" />,
-          null,
+          undefined,
```
layershifter, line 36: This argument is `options`, `null` and `undefined` aren't equal for default args:

```js
function getValue(value = { foo: 'bar' }) {
  return value
}

console.log(getValue(null)) // => null
console.log(getValue(undefined)) // {foo: "bar"}
```

kuzhelov (approved):

test/specs/components/Icon/Icon-test.tsx
kuzhelov: great catch!

PR #493, opened 2018-11-19
fix(docs): fix basename logic for accessibility behaviors root component

kuzhelov:
Provided fix makes the logic of stripping out `ts` extension to be more strict. 

Previously for files like `petsBehavior.ts` (yes, contrieved:) we would have 'Pe Behavior Ts' displayed on the docs page (instead of expected 'Pets Behavior').

### DIFF
--- a/docs/src/components/DocsBehaviorRoot.tsx
+++ b/docs/src/components/DocsBehaviorRoot.tsx
@@ -18,7 +18,7 @@ class DocsBehaviorRoot extends React.Component<any, any> {
   }
 
   baseName(fileName: string) {
-    const divided = _.startCase(fileName.replace('ts', ''))
+    const divided = _.startCase(fileName.replace(/\.ts$/, ''))
     return _.upperFirst(_.lowerCase(divided))
   }
 


docs/src/components/DocsBehaviorRoot.tsx
```diff
@@ -18,7 +18,7 @@ class DocsBehaviorRoot extends React.Component<any, any> {
   }
 
   baseName(fileName: string) {
-    const divided = _.startCase(fileName.replace('ts', ''))
+    const divided = _.startCase(fileName.replace(/\.ts$/, ''))
     return _.upperFirst(_.lowerCase(divided))
```
layershifter, line 22: May be we can move this to `gulp-component-menu-behaviors` and call this prop as `title`?

kuzhelov: actually, would like just to introduce fix first - but generally absolutely agree, let me consider this move for the follow up PR :)

layershifter (approved):

mnajdova (approved):
Good catch!

PR #496, opened 2018-11-19
fix(factories): shorthand fix for applying props to react element

bmdalex:
### Description

This PR tackles the first argument type of the function returned by `createShorthandFactory` factory, more precisely when passing a `React.Element` as  argument.

This comes into play in examples such as:
```jsx
<Button
  icon={<span name='coffee' style={{ background: 'red' }}>span content</span>}
  content="Click Me"
/>
<Button
  icon={<Icon name='coffee' style={{ background: 'red' }}>icon content</Icon>}
  content="Click Me"
/>
```

that translate into calling:

```typescript
createShorthandFactory(Icon, 'name').create(icon, {
  defaultProps: {
    styles: styles.icon,
    xSpacing: !content ? 'none' : iconPosition === 'after' ? 'before' : 'after',
    variables: variables.icon,
  }
})

// with icon being
<span name='coffee' style={{ background: 'red' }}>span content</span>

// and respectively:
<Icon name='coffee' style={{ background: 'red' }}>icon content</Icon>
```

and rendering:

screenshot 2018-11-19 at 21 07 16

The `icon` argument can be any `React.Element` so we cannot make too many presumptions on what to do with the props send as the 2nd argument (`defaultProps` and `overrideProps`).

### Current Behavior:

We blindly send all props to the `React.cloneElement` call (see `createShorthand` call in `factories.tsx`).
For the example above, this can lead to passing invalid props to certain components and we currently catch this wrong usage in one of our unit tests:
screenshot 2018-11-19 at 19 34 10

### Solution:

If the first argument of the function returned by `createShorthandFactory` factory is a `React.Element` **just return the element early without cloning and applying extra props**.


test/specs/lib/factories-test.tsx
```diff
@@ -138,6 +142,48 @@ const itOverridesDefaultPropsWithFalseyProps = (propsSource, shorthandConfig) =>
   })
 }
 
+const itHandlesTheElementBasedOnComponentType = () => {
```
layershifter: What is purpose of `itHandlesTheElementBasedOnComponentType`? As I see this code is not reusable, let's just use it in place.

src/lib/isElementTypeOfComponent.ts
```diff
@@ -0,0 +1,28 @@
...
+): boolean => {
+  const componentNameOfElement = getComponentName(element.type)
+  const componentName = getComponentName(component)
+
+  return (
+    componentNameOfElement !== null &&
+    componentName !== null &&
+    componentNameOfElement === componentName
+  )
+}
```
layershifter: Can you clarify why we need this function at all?

```jsx
const SFC = (props) => React.createElement('div', props)
class Classy extends React.Component {
  render() {
    return null
  }
}

// 

const sfc = React.createElement(SFC)
const classy = React.createElement(Classy)

//

classy.type === Classy // true
sfc.type === SFC // true
```

The main issue that this will not work for HOCs, however your solution will fail, too :)

docs/src/examples/components/Input/Variations/InputExampleWrapperSlot.shorthand.tsx
```diff
@@ -23,22 +23,6 @@ const InputExampleWrapperSlot = () => (
         styles: { padding: '5px', backgroundColor: 'red' },
       }}
     />
-
-    <Text content="Wrapped Input with existing component:" />
```
kuzhelov, line 27: could you, please, suggest a reason for them to be removed?

src/lib/isElementTypeOfComponent.ts
```diff
@@ -0,0 +1,28 @@
+import { Props } from '../../types/utils'
+
+const getComponentName = (component: React.ReactType) => {
```
kuzhelov: ideally this should be exported as well from lib, as needs for this are quite common

src/lib/isElementTypeOfComponent.ts
```diff
@@ -0,0 +1,28 @@
...
+    case 'function':
+      return component.displayName || component.name || null // unknown
+    case 'string':
+      return component
+    default:
+      return null // unknown
+  }
+}
+
+const isElementTypeOfComponent = (
```
kuzhelov: the name of the function is very misleading - actually the only thing it is doing is comparing names of component types.

I would see it being done like:

```jsx
const areTypeNamesEqual = (first: React.ReactType, second: React.ReactType) => {
  ...
}
```

and let it to be client's responsibility to fetch type (i.e. this do `element.type` fetch):

```js
if (
    valIsReactElement &&
    !areTypeNamesEqual(value.type, Component)
) {
```

src/lib/factories.tsx
```diff
@@ -56,6 +57,14 @@ export function createShorthand(
   const valIsPropsObject = _.isPlainObject(value)
   const valIsReactElement = React.isValidElement(value)
 
+  // just return ReactElements that have different type than Component
```
kuzhelov: plural form is a bit misleading - maybe something like 

```js
// return value as is if its type is different than Component
```

kuzhelov: actually, have some doubts about it: this change implies that the following is true (`Button`'s 'icon' slot as an example):

```jsx
<Button icon={{ name: '...', bordered: 'true', circular: 'true' }} />

// this is processed in the same way
<Button icon={<Icon name='...' bordered='true' circular='true' />} />

// and this will be processed DIFFERENTLY from second example!
<Button icon={<MyIcon name='...' bordered='true' circular='true' />} />
```

I am not sure that **type+format** of shorthand property value should drive the decision on how it will be processed - seems that relying on **format only** (i.e. passing React element <...> vs passing JS object {...}) should be just enough to drive this difference. This will introduce easier principle that we could explain to client, as well as much easier and predictable debugging experience for the client afterwards.

src/lib/isElementTypeOfComponent.ts
bmdalex: @layershifter 
Obviously I tried function comparison first. I have a sandbox (https://codesandbox.io/s/oqwj254k16) where both solution work.
Unfortunately function comparison doesn't work in our codebase. It was surprising for me too :( that's why I needed this method..
screenshot 2018-11-22 at 18 29 39

As of HOC argument, we don't really care about that, it's not a relevant use case; if users pass HOCs to the factory it's their problem.

In case I'm missing something, please provided valid user scenarios to counter this.

docs/src/examples/components/Input/Variations/InputExampleWrapperSlot.shorthand.tsx
bmdalex: because that API is not supported anymore; user needs to pass `string`, `object` or `Slot` element to `wrapper` slot

src/lib/factories.tsx
bmdalex: @kuzhelov 
```jsx
// and this will be processed DIFFERENTLY from second example!
<Button icon={<MyIcon name='...' bordered='true' circular='true' />} />
```
**the whole purpose of this change is not to support examples like this; the arguments are:**
- `MyIcon` can have implementation, props and structure that are very different than `Icon` so the `Icon` component factory (aka `Icon.create`) won't know how to process this element; please refer to the examples in the description for more details or we can have a sync tomorrow;
- this API (passing random elements to known component slots) is almost never used (discussed with @levithomason );
- user will still have the ability to achieve this scenario using the `renderIcon` prop **so we're not eliminating existing functionality**
```jsx
// and this will be processed DIFFERENTLY from second example!
<Button renderIcon={(Icon, iconProps) => <MyIcon name='...' bordered='true' circular='true' />} />
```

test/specs/lib/factories-test.tsx
bmdalex: done, thanks

src/lib/isElementTypeOfComponent.ts
bmdalex: done, thanks

src/lib/isElementTypeOfComponent.ts
bmdalex: thanks, applied suggestion

docs/src/examples/components/Input/Variations/InputExampleWrapperSlot.shorthand.tsx
kuzhelov: and this is a policy that applies to `wrapper` only, right? So, say, the following is still seen as possible?

```jsx
<Button icon={<span .. >▼</span>} ... />
```

bmdalex: exactly, this is still possible, yes

src/lib/reactComponentUtils.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export const areTypeNamesEqual = (first: React.ReactType, second: React.ReactType): boolean => {
+  const componentNameOfElement = getComponentName(first)
+  const componentName = getComponentName(second)
+
+  return (
+    componentNameOfElement !== null &&
+    componentName !== null &&
+    componentNameOfElement === componentName
```
kuzhelov: so, if both will be `undefined`, `true` will be returned. Is it intentional?

src/lib/reactComponentUtils.ts
```diff
@@ -0,0 +1,21 @@
+export const getComponentName = (component: React.ReactType) => {
+  switch (typeof component) {
```
kuzhelov: nit: this could be simplified a bit to a single flow:

```jsx
return (
    Component.displayName ||
    Component.name ||
    (typeof Component === 'string' && Component.length > 0
      ? Component
      : null // unknown)
```

test/utils/createMockComponent.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
+import { Props } from 'types/utils'
+
+const createMockComponent = <P extends Props = Props>(displayName: string): React.SFC<P> => {
```
kuzhelov: this is used in single file only - would suggest to inline as there is no much code introduced here

test/specs/lib/reactComponentUtils-test.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+    it('returns false for different types', () => {
+      expect(areTypeNamesEqual(React.createElement(MyComponent).type, OtherComponent)).toBeFalsy()
+      expect(areTypeNamesEqual(React.createElement(MyComponent).type, 'OtherComponent')).toBeFalsy()
+    })
+
+    it('returns true for the same type', () => {
+      expect(areTypeNamesEqual(React.createElement(MyComponent).type, MyComponent)).toBeTruthy()
+      expect(
+        areTypeNamesEqual(React.createElement(OtherComponent).type, 'OtherComponent'),
```
kuzhelov: condition checked for `OtherComponent` is not what the test prescribes - actually, the name of string component is compared with name of `OtherComponent`. For `MyComponent` everything is fine in these regards

test/specs/lib/factories-test.tsx
```diff
@@ -530,6 +534,46 @@ describe('factories', () => {
...
+            Component: MyComponent,
+            value: MyComponent({ common: 'user', user: true }),
+            defaultProps: { common: 'default', default: true },
+            overrideProps: { common: 'override', override: true },
+          },
+          { common: 'override', user: true, default: true, override: true },
+        )
+      })
+
+      test('applies only the element props and forwards the element as is if element type is different than the primitive component', () => {
```
kuzhelov: actually, this description could be simplified, as `element` implies type _and_ props. So, we might say here something like:

> forwards original element 'as is' if element type is different 

In fact, this is the condition we would like to state and check.

Then, if we would like to review different cases, SFC-based element vs functional element, I would suggest either
- not  do it, as this is not the statement we would like to check :)
- create subtests in the following way:

```js
describe("forwards original element 'as is' if element type is different ", () => {
  test('applies to functional component', ...)
  test('applies to string component', ...)
})
```

But, once again, I would rather not make this differentiation, as the statement is absolutely clear - if name is different, then component should be passed 'as is'. When there are many tests for the same condition (and those have no common describe block with statement mentioned there), it is much harder to infer that, in fact, both these tests check the same thing.

test/specs/lib/reactComponentUtils-test.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { createMockComponent } from 'test/utils'
+
+describe('getComponentName', () => {
+  it('returns null for invalid arguments', () => {
+    const invalidArgs = [undefined, null, '', {}]
+    invalidArgs.map(invalidArg => expect(getComponentName(invalidArg as any)).toBeFalsy())
+  })
+
+  it('returns the string component itself', () => {
+    const stringArgs = ['div', 'span', 'MyComponent']
```
kuzhelov: nit: would suggest to remove `MyComponent` from list, as is it a bit misleading: React doesn't support custom string components. Would be easier to read test cases if those will correspond to domain we are primarily interested in (while I do understand that you've tried to make tests generic):

```js
describe('getComponentName', () => {
  it('returns displayName for class component types', ...)
  it('returns function name for functional component types', ...)
  it('returns type itself for string component types', ...)

  it('returns null for invalid arguments', ...)
})
```

test/specs/lib/reactComponentUtils-test.tsx
```diff
@@ -0,0 +1,63 @@
...
+})
+
+describe('areTypeNamesEqual', () => {
+  it('returns false for invalid arguments', () => {
+    expect(areTypeNamesEqual(null, undefined)).toBeFalsy()
+    expect(areTypeNamesEqual(undefined, '')).toBeFalsy()
+    expect(areTypeNamesEqual('', '')).toBeFalsy()
+  })
+
+  describe('for primitive components', () => {
```
kuzhelov: rather `for primitive component types`, not just `components` (which implies instances)

test/specs/lib/reactComponentUtils-test.tsx
```diff
@@ -0,0 +1,63 @@
...
+    it('returns false for different types', () => {
+      expect(areTypeNamesEqual(React.createElement('p').type, 'div')).toBeFalsy()
```
kuzhelov: not sure why it is necessary to test two semantically absolutely equal cases. Would rather suggest to merge tests under describe blocks and simplify tests hierarchy (as these tests are fairly simple):

```js
describe('areTypeNamesEqual', () => {
  test('properly handles string component types', ...)
  test('properly handles functional component types', ...)
})
```
For me as a client it would absolutely obvious what does this **properly handes* implies by default given the name of the function - so, there is no need to introduce additional test cases to further describe this behavior for each set of component types.

src/lib/isElementTypeOfComponent.ts
layershifter: I didn't try, but I think that issue is caused by `react-hot-loader`, BTW they have own util:
https://github.com/gaearon/react-hot-loader/blob/master/README.md#arecomponentsequalcomponent1-component2

Can we check it? I'm not sure that the comparison by names is the best idea.

layershifter (commented):
Can we check my solution for `areTypeNamesEqual`? I think that the check of `displayName` is not the best solution.

src/lib/isElementTypeOfComponent.ts
bmdalex: awesome, thank you! I'll take a look for sure if we keep current implementation idea

src/lib/factories.tsx
bmdalex: had a conversation with @kuzhelov and we'd like to open the following question:

for `createShorthand(Component, mappedProp, value, options)` if `value` is of type `ReactElement` do we:
#### 1. clone the element and apply the props sent as parameter if the type of `value` is the same as the type of `Component` (currently implemented);
#### 2. just pass the element and return early? (previously discussed with @levithomason )
#### 3. remove this API altogether since the user will still have the ability to achieve this scenario using the `renderIcon` prop **so we're not eliminating existing functionality**
```jsx
// and this will be processed DIFFERENTLY from second example!
<Button renderIcon={(Icon, iconProps) => <MyIcon name='...' bordered='true' circular='true' />} />
```

fyi @kuzhelov @levithomason @layershifter

bmdalex:
marking blocked until having a resolution on:

https://github.com/stardust-ui/react/pull/496#discussion_r236325417

src/lib/factories.tsx
bmdalex: for `createShorthand(Component, mappedProp, value, options)` if `value` is of type ReactElement we decided to just return `value` early (solution 2).

src/lib/isElementTypeOfComponent.ts
bmdalex: N.A. anymore since the implementation was drastically simplified following decision:
https://github.com/stardust-ui/react/pull/496#discussion_r238708722

test/utils/createMockComponent.tsx
bmdalex: N.A. anymore since the implementation was drastically simplified following decision:
https://github.com/stardust-ui/react/pull/496#discussion_r238708722

src/lib/reactComponentUtils.ts
bmdalex: N.A. anymore since the implementation was drastically simplified following decision:
https://github.com/stardust-ui/react/pull/496#discussion_r238708722

src/lib/reactComponentUtils.ts
bmdalex: N.A. anymore since the implementation was drastically simplified following decision:
https://github.com/stardust-ui/react/pull/496#discussion_r238708722

test/specs/lib/reactComponentUtils-test.tsx
bmdalex: N.A. anymore since the implementation was drastically simplified following decision:
https://github.com/stardust-ui/react/pull/496#discussion_r238708722

test/specs/lib/factories-test.tsx
bmdalex: N.A. anymore since the implementation was drastically simplified following decision:
https://github.com/stardust-ui/react/pull/496#discussion_r238708722

test/specs/lib/reactComponentUtils-test.tsx
bmdalex: N.A. anymore since the implementation was drastically simplified following decision:
https://github.com/stardust-ui/react/pull/496#discussion_r238708722

test/specs/lib/reactComponentUtils-test.tsx
bmdalex: N.A. anymore since the implementation was drastically simplified following decision:
https://github.com/stardust-ui/react/pull/496#discussion_r238708722

test/specs/lib/reactComponentUtils-test.tsx
bmdalex: N.A. anymore since the implementation was drastically simplified following decision:
https://github.com/stardust-ui/react/pull/496#discussion_r238708722

bmdalex:
As per our sync conclusion, for `createShorthand(Component, mappedProp, value, options)` if `value` is of type ReactElement we decided to just return `value` early.

docs/src/examples/components/RadioGroup/Item/RadioGroupItemExampleChecked.shorthand.tsx
```diff
@@ -4,7 +4,7 @@ import { RadioGroup } from '@stardust-ui/react'
 const RadioGroupItemExampleCheckedShorthand = () => (
   <RadioGroup
     defaultCheckedValue="1"
-    items={[<RadioGroup.Item key="1" label="This radio comes pre-checked" value="1" />]}
+    items={[{ key: '1', label: 'This radio comes pre-checked', value: '1' }]}
```
layershifter: Can we add a second unchecked element there? It's hard to read `items={[{`

docs/src/examples/components/RadioGroup/Types/RadioGroupColorPickerExample.shorthand.tsx
```diff
@@ -22,37 +46,6 @@ class RadioGroupColorPickerExample extends React.Component {
...
-        <RadioGroup
-          defaultCheckedValue="pink"
-          items={this.items()}
-          checkedValueChanged={this.handleChange}
-        />
-      </div>
-    )
-  }
 }
+
```
layershifter, line 50: This example should be definitely refactored in the future.

test/specs/components/RadioGroup/RadioGroup-test.tsx
```diff
@@ -60,7 +60,13 @@ describe('RadioGroup', () => {
...
   })
 
-  const itemsTest = getItems => {
+  const itemsTest = ({
+    getItems,
+    isChildrenApiTest = false,
+  }: {
+    getItems: Function
+    isChildrenApiTest?: boolean
+  }) => {
```
layershifter: For now let's leave the two arguments instead of object:

```suggestion
const itemsTest = (getItems: Fuction, isChildrenApiTest: boolean = false) => {
```

One line is better than seven 😸

test/specs/components/RadioGroup/RadioGroup-test.tsx
```diff
@@ -102,29 +108,30 @@ describe('RadioGroup', () => {
     })
 
     describe('click event handler', () => {
-      it('should set the value when item is clicked', () => {
-        const checkedValueChanged = jest.fn()
-        const wrapper = mountWithProvider(
-          <RadioGroup items={getItems()} checkedValueChanged={checkedValueChanged} />,
-        )
-        const radioGroupItems = wrapper.find('RadioGroupItem')
+      !isChildrenApiTest &&
```
layershifter: Please, use `if` in such cases it will be more clear. Also, I saw that you're using in both cases `!isChildrenApiTest` may be it should `isShorthandApiTest`?

bmdalex: I used that parameter name in order to emphasize the tests are not supported by children API, i thought it's more readable. I can still make the change if you're feeling strongly for it @layershifter

layershifter: Yes, let's change it. Because in both cases you're using negative condition.

docs/src/examples/components/RadioGroup/Types/RadioGroupColorPickerExample.shorthand.tsx
bmdalex: if you have a good suggestion, I don't mind adding it now

test/specs/components/RadioGroup/RadioGroup-test.tsx
bmdalex: done

docs/src/examples/components/RadioGroup/Item/RadioGroupItemExampleChecked.shorthand.tsx
bmdalex: added

test/specs/components/RadioGroup/RadioGroup-test.tsx
bmdalex: done as suggested

layershifter (approved):
We discussed these changes before, the final result LGTM 👍

docs/src/examples/components/RadioGroup/Item/RadioGroupItemExample.shorthand.tsx
```diff
@@ -8,8 +8,8 @@ const handleChange = () => {
 const RadioGroupItemExample = () => (
   <RadioGroup
     items={[
-      <RadioGroup.Item key="1" label="Make your choice" value="1" checkedChanged={handleChange} />,
-      <RadioGroup.Item key="2" label="Another option" value="2" checkedChanged={handleChange} />,
+      { key: '1', label: 'Make your choice', value: '1', checkedChanged: handleChange },
```
kuzhelov, line 11: 👍

src/lib/factories.tsx
```diff
@@ -105,18 +105,16 @@ function createShorthandFromValue(
   const valIsNoop = _.isNil(value) || typeof value === 'boolean'
   if (valIsNoop && !options.render) return null
 
+  // return value as is if it a ReactElement
```
kuzhelov: nit: this line is hard to be read from the first time :)

Would suggest small change:

> return value 'as is' if it is a ReactElement

src/lib/factories.tsx
```diff
@@ -105,18 +105,16 @@ function createShorthandFromValue(
...
 
   // unhandled type warning
-  if (
-    process.env.NODE_ENV !== 'production' &&
-    !valIsPrimitive &&
-    !valIsPropsObject &&
-    !valIsReactElement &&
-    !valIsNoop
-  ) {
+  if (process.env.NODE_ENV !== 'production' && !valIsPrimitive && !valIsPropsObject && !valIsNoop) {
```
kuzhelov: suggest this check being left as it was before, and put all this section to the top of the method. 

This one **serves as method's precondition**, and thus:
- it is aimed to clearly indicates method's expectations about input data (`value`)
  - currently this expectation condition has changed in code, while, in fact, method's precondition hasn't changed anyhow
- it should run before all the method's logic

test/specs/components/RadioGroup/RadioGroup-test.tsx
```diff
@@ -101,31 +101,33 @@ describe('RadioGroup', () => {
...
-    describe('click event handler', () => {
-      it('should set the value when item is clicked', () => {
-        const checkedValueChanged = jest.fn()
-        const wrapper = mountWithProvider(
-          <RadioGroup items={getItems()} checkedValueChanged={checkedValueChanged} />,
-        )
-        const radioGroupItems = wrapper.find('RadioGroupItem')
+    if (isShorthandApiTest) {
+      describe('click event handler', () => {
+        it('should set the value when item is clicked', () => {
```
kuzhelov: "should set `checked` when item is clicked"? At least this is what I see from test (other than check that event handler is called with proper value for `value`)

test/specs/components/RadioGroup/RadioGroup-test.tsx
```diff
@@ -147,21 +149,23 @@ describe('RadioGroup', () => {
       expect(checkedValueChanged).not.toHaveBeenCalled()
     })
 
-    it('should not set the value when disabled item is clicked', () => {
-      const wrapper = mountWithProvider(<RadioGroup items={getItems({ disabledItem: 1 })} />)
-      const radioGroupItems = wrapper.find('RadioGroupItem')
+    if (isShorthandApiTest) {
+      it('should not set the value when disabled item is clicked', () => {
```
kuzhelov: would suggest to replace `value` with `checked` - otherwise description is a bit misleading

test/specs/lib/factories-test.tsx
```diff
@@ -524,18 +518,19 @@ describe('factories', () => {
...
-        { value: <div {...{ overridden: true } as any} /> },
-      )
-      itOverridesDefaultPropsWithFalseyProps('element', {
-        value: <div {...{ undef: undefined, nil: null, zero: 0, empty: '' } as any} />,
+
+      test('forwards original element "as is"', () => {
+        testCreateShorthand(
+          {
+            Component: 'p',
+            value: <span {...{ common: 'user', user: true } as any} />,
```
kuzhelov: may suggest `originalElement` to be used as value for `common` - then test (and, especially, its assertion) would be more descriptive

kuzhelov (approved):

test/specs/lib/factories-test.tsx
bmdalex: that one is the name of the prop, I think `originalElement` as prop name is confusing; I made a few changes to make the test easier to understand

src/lib/factories.tsx
bmdalex: agreed, done

test/specs/components/RadioGroup/RadioGroup-test.tsx
bmdalex: done

test/specs/components/RadioGroup/RadioGroup-test.tsx
bmdalex: done

src/lib/factories.tsx
bmdalex: it was saving few unnecessary checks but valid points, done

PR #499, opened 2018-11-20
fix(AutoControlledComponent): fix behaviour when undefined it passed

layershifter:
We discussed this changes during meetings on the previous week, we need to change the behaviour of `AutoControlledComponent` to match React's.

An example that shows issue: https://codesandbox.io/s/l20mjn74zz

---

After this PR we can safely migrate to `getDerivedStateFromProps()` 👍

### DIFF
--- a/src/lib/AutoControlledComponent.tsx
+++ b/src/lib/AutoControlledComponent.tsx
@@ -171,13 +171,10 @@ export default class AutoControlledComponent<P = {}, S = {}> extends UIComponent
 
     // Solve the next state for autoControlledProps
     const newState = autoControlledProps.reduce((acc, prop) => {
-      const isNextUndefined = _.isUndefined(nextProps[prop])
-      const propWasRemoved = !_.isUndefined(this.props[prop]) && isNextUndefined
+      const isNextDefined = !_.isUndefined(nextProps[prop])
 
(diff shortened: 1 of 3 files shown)


src/lib/AutoControlledComponent.tsx
```diff
@@ -171,13 +171,10 @@ export default class AutoControlledComponent<P = {}, S = {}> extends UIComponent
...
     const newState = autoControlledProps.reduce((acc, prop) => {
-      const isNextUndefined = _.isUndefined(nextProps[prop])
-      const propWasRemoved = !_.isUndefined(this.props[prop]) && isNextUndefined
+      const isNextDefined = !_.isUndefined(nextProps[prop])
 
       // if next is defined then use its value
-      if (!isNextUndefined) acc[prop] = nextProps[prop]
-      // reinitialize state for props just removed / set undefined
-      else if (propWasRemoved) acc[prop] = getAutoControlledStateValue(prop, nextProps)
+      if (isNextDefined) acc[prop] = nextProps[prop]
```
layershifter, line 177: In fact, we don't need to do something with `undefined` at all. We should update the state only for defined values from props

bmdalex (approved):

mnajdova (approved):

PR #503, opened 2018-11-20
feat(utils): add wrapper for the createComponent function 

### LINKED ISSUE #498: Expose a wrapper for `createComponent`
### Problem description
We need to enable users of the library to leverage Stardust capabilities, styling especially, in their own components or 3rd party components.

### Proposed solution
- create a thing wrapper function with the same name on top of `createComponent`;
- export it in index file;
- create relevant doc site example / prototype; thoughts?
- the signature of the function should be:
```typescript
import createComponentInternal from './createComponent'

export interface CreateComponentWrapperConfig<P> {
  displayName: string
  render: (props: P & { styles: ComponentSlotStylesPrepared }) => React.ReactNode
}

const createComponent = <P extends {} = {}>({
  displayName,
  render,
}: CreateComponentWrapperConfig<P>): React.SFC<P> => createComponentInternal({ displayName, render })

export default createComponent
```

mnajdova:
# Introducing createComponent functionality as part of the public API in Stardust
This PR aims to fix https://github.com/stardust-ui/react/issues/498 by adding a wrapper around the internal createComponent function. The API for the function looks like this:
```
export interface RenderStardustResultConfig {
  classes: ComponentSlotClasses
  rtl: boolean
}

export interface CreateStardustComponentConfig<P> {
  displayName: string
  render: (props: P & { stardust: RenderStardustResultConfig }) => React.ReactNode
}

const createComponent = <P extends {} = {}, S extends {} = {}>({
  displayName = 'StardustComponent',
  render,
}: CreateStardustComponentConfig<P>) => React.SFC<P> 
```
Only the bits related to the styles are added in the config of the component.

### Definition of the custom component
The createComponent function should be used for defining custom components in the following way:
```
export interface StyledButtonProps {
  styles?: ComponentSlotStyle
  children?: React.ReactNodeArray | React.ReactNode
}

const StyledButton: React.SFC<StyledButtonProps> = createComponent<StyledButtonProps, {}>({
  displayName: 'MyStyledButton',
  render({stardust, ...props}) {
    const { classes } = stardust
    
    return (<button className={cx(props.className, classes.root)} {...props} />)
  },
})
```
### Example usage of the custom component
Some example usages of the `StyledButton` component would be:
#### Using styles and accessing the theme from the Provider
```
<StyledButton styles={({ theme: {siteVariables} }) => ({ backgroundColor: siteVariables.brand })}>
  My styled button
</StyledButton>
```
#### Using styles and accessing the props as they would use with some Stardust component 
```
<StyledButton isRed styles={({props}) => ({backgroundColor: props.isRed ? 'red' : 'blue'})}>
  My red styled
</StyledButton>
```
## Theming support
```
<Provider theme={{
  componentStyles: {
    StyledButton: {
      root: ({ props, variables, theme: {siteVariables} }) => ({
        backgroundColor: props.isRed ? (variables as any).red : siteVariables.brand,
      }),
    },
  },
  componentVariables: {
    StyledButton: {
      red: '#fc798c',
    },
  },
}}>
  <StyledButton isRed>My styled button on Provider</StyledButton>
</Provider>
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -126,3 +126,4 @@ export { default as dialogBehavior } from './lib/accessibility/Behaviors/Dialog/
 //
 export { default as mergeThemes } from './lib/mergeThemes'
 export { createComponent } from './lib'
+export { RenderStardustResultConfig, CreateStardustComponentConfig } from '../types/utils'
--- a/src/index.ts
+++ b/src/index.ts
@@ -125,4 +125,4 @@ export { default as dialogBehavior } from './lib/accessibility/Behaviors/Dialog/
 // Utilities
 //
 export { default as mergeThemes } from './lib/mergeThemes'
-export { createStardustComponent } from './lib'
+export { createComponent } from './lib'
--- a/src/index.ts
+++ b/src/index.ts
@@ -125,3 +125,4 @@ export { default as dialogBehavior } from './lib/accessibility/Behaviors/Dialog/
 // Utilities
 //
 export { default as mergeThemes } from './lib/mergeThemes'
+export { createStardustComponent } from './lib'
--- a/src/index.ts
+++ b/src/index.ts
@@ -126,4 +126,7 @@ export { default as dialogBehavior } from './lib/accessibility/Behaviors/Dialog/
 //
 export { default as mergeThemes } from './lib/mergeThemes'
 export { createComponent } from './lib'
-export { RenderStardustResultConfig, CreateStardustComponentConfig } from '../types/utils'
+export {
+  RenderStardustResultConfig,
+  CreateStardustComponentConfig,
+} from './lib/createStardustComponent'
--- a/src/lib/createStardustComponent.tsx
+++ b/src/lib/createStardustComponent.tsx
@@ -30,17 +30,15 @@ const createComponent = <P extends {} = {}, S extends {} = {}>({
   displayName = 'StardustComponent',
   render,
 }: CreateStardustComponentConfig<P>): React.SFC<P> => {
-  const StardustComponent: React.SFC<P> = createComponentInternal({
+  return createComponentInternal({
     displayName,
     render(config, props) {
       // TODO add here everything that the client may expect
-      const restrictedConfig = _.pick(config, ['classes', 'variables', 'styles', 'rtl', 'theme'])
+      const filteredConfig = _.pick(config, ['classes', 'variables', 'styles', 'rtl', 'theme'])
 
-      return render(restrictedConfig, props)
+      return render(filteredConfig, props)
     },
   })
-  StardustComponent.displayName = displayName
-  return StardustComponent
 }
 
 export default createComponent
--- a/src/lib/createStardustComponent.tsx
+++ b/src/lib/createStardustComponent.tsx
@@ -10,18 +10,18 @@ export interface RenderStardustResultConfig {
 
 export interface CreateStardustComponentConfig<P> {
   displayName?: string
-  render: (config: RenderStardustResultConfig, props: P) => React.ReactNode
+  render: (props: P & { stardust: RenderStardustResultConfig }) => React.ReactNode
 }
 
 const createComponent = <P extends {} = {}, S extends {} = {}>({
   displayName = 'StardustComponent',
   render,
 }: CreateStardustComponentConfig<P>): React.SFC<P> => {
-  return createComponentInternal({
+  return createComponentInternal<P, S>({
     displayName,
-    render(config, props) {
+    render: (config, props) => {
       const filteredConfig = pick(config, ['classes', 'rtl'])
-      return render(filteredConfig, props)
+      return render(Object.assign({ stardust: filteredConfig }, props))
     },
   })
 }
--- a/src/lib/createStardustComponent.tsx
+++ b/src/lib/createStardustComponent.tsx
@@ -1,19 +1,11 @@
 import createComponentInternal from './createComponent'
 import * as React from 'react'
 import * as _ from 'lodash'
-import {
-  ComponentSlotClasses,
-  ThemePrepared,
-  ComponentSlotStylesPrepared,
-  ComponentVariablesObject,
-} from '../themes/types'
+import { ComponentSlotClasses } from '../themes/types'
 
 export interface RenderStardustResultConfig {
   classes: ComponentSlotClasses
-  variables: ComponentVariablesObject
-  styles: ComponentSlotStylesPrepared
   rtl: boolean
-  theme: ThemePrepared
 }
 
 export type RenderStardustComponentCallback<P> = (
@@ -34,7 +26,7 @@ const createComponent = <P extends {} = {}, S extends {} = {}>({
     displayName,
     render(config, props) {
       // TODO add here everything that the client may expect
-      const filteredConfig = _.pick(config, ['classes', 'variables', 'styles', 'rtl', 'theme'])
+      const filteredConfig = _.pick(config, ['classes', 'rtl'])
 
       return render(filteredConfig, props)
     },
--- a/src/lib/createStardustComponent.tsx
+++ b/src/lib/createStardustComponent.tsx
@@ -1,7 +1,17 @@
 import createComponentInternal from './createComponent'
 import * as React from 'react'
-import * as _ from 'lodash'
-import { CreateStardustComponentConfig } from '../../types/utils'
+import { pick } from 'lodash'
+import { ComponentSlotClasses } from '@stardust-ui/react'
+
+export interface RenderStardustResultConfig {
+  classes: ComponentSlotClasses
+  rtl: boolean
+}
+
+export interface CreateStardustComponentConfig<P> {
+  displayName?: string
+  render: (config: RenderStardustResultConfig, props: P) => React.ReactNode
+}
 
 const createComponent = <P extends {} = {}, S extends {} = {}>({
   displayName = 'StardustComponent',
@@ -10,7 +20,7 @@ const createComponent = <P extends {} = {}, S extends {} = {}>({
   return createComponentInternal({
     displayName,
     render(config, props) {
-      const filteredConfig = _.pick(config, ['classes', 'rtl'])
+      const filteredConfig = pick(config, ['classes', 'rtl'])
       return render(filteredConfig, props)
     },
   })
--- a/src/lib/createStardustComponent.tsx
+++ b/src/lib/createStardustComponent.tsx
@@ -1,22 +1,7 @@
 import createComponentInternal from './createComponent'
 import * as React from 'react'
 import * as _ from 'lodash'
-import { ComponentSlotClasses } from '../themes/types'
-
-export interface RenderStardustResultConfig {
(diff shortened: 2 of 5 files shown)


kuzhelov:
@mnajdova, please, note that example's code provided in description could be simplified:

```jsx
const StyledButton = createComponent({
  displayName: 'MyStyledButton',
  render(stardust, props) {
    const { classes } = stardust
    
    // note that we should respect client's classes passed as well
    return (<button className={cx(props.className, classes.root)} {...props} />)
  },
})
```

src/lib/createStardustComponent.tsx
```diff
@@ -0,0 +1,46 @@
...
+import {
+  ComponentSlotClasses,
+  ThemePrepared,
+  ComponentSlotStylesPrepared,
+  ComponentVariablesObject,
+} from '../themes/types'
+
+export interface RenderStardustResultConfig {
+  classes: ComponentSlotClasses
+  variables: ComponentVariablesObject
```
kuzhelov: not sure if we should expose `variables` and `styles` for the first iteration - it seems that the only bits client is interested in are _evaluated_ styles, so it is all about `classes`. The same generally applies to `theme` object as well - to me it seems to be safer not expose it if there is no need to.

So, to conclude, I would suggest to leave `classes` and maybe `rtl` only for the first run.

src/lib/createStardustComponent.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+const createComponent = <P extends {} = {}, S extends {} = {}>({
+  displayName = 'StardustComponent',
+  render,
+}: CreateStardustComponentConfig<P>): React.SFC<P> => {
+  const StardustComponent: React.SFC<P> = createComponentInternal({
+    displayName,
+    render(config, props) {
+      // TODO add here everything that the client may expect
+      const restrictedConfig = _.pick(config, ['classes', 'variables', 'styles', 'rtl', 'theme'])
```
kuzhelov: nit: 'restricted' -> 'filtered', seems to be more expressive. Feel free to decide by yourself :)

src/lib/createStardustComponent.tsx
```diff
@@ -0,0 +1,46 @@
...
+  displayName?: string
+  render: RenderStardustComponentCallback<P>
+}
+
+const createComponent = <P extends {} = {}, S extends {} = {}>({
+  displayName = 'StardustComponent',
+  render,
+}: CreateStardustComponentConfig<P>): React.SFC<P> => {
+  const StardustComponent: React.SFC<P> = createComponentInternal({
+    displayName,
```
kuzhelov, line 21: is there any reason we are using this value both here, as well as on the following line: https://github.com/stardust-ui/react/pull/503/files#diff-3cd0aeac45681cbb710926892d4cf28bR42 ? Maybe I am missing something, just would expect that there will be the only place where this is used

mnajdova: We don't need it on the second usage, just leftover from the renderComponent. Will remove it.

src/lib/createStardustComponent.tsx
mnajdova: I had just the classes in the beginning but was not sure whether we want this additional bits. Will leave only the classes and rtl for now. Thanks!

src/lib/createStardustComponent.tsx
```diff
@@ -0,0 +1,36 @@
+import createComponentInternal from './createComponent'
+import * as React from 'react'
+import * as _ from 'lodash'
+import { ComponentSlotClasses } from '../themes/types'
+
+export interface RenderStardustResultConfig {
```
bmdalex, line 6: no `styles`?

src/lib/createStardustComponent.tsx
```diff
@@ -0,0 +1,36 @@
...
+  classes: ComponentSlotClasses
+  rtl: boolean
+}
+
+export type RenderStardustComponentCallback<P> = (
+  config: RenderStardustResultConfig,
+  props: P,
+) => any
```
bmdalex: `React.ReactNode` as result type?

bmdalex: also, didn't we agree to have only only one argument for render?
```typescript
render: (props: P & RenderStardustResultConfig) => React.ReactNode
```

src/lib/createStardustComponent.tsx
```diff
@@ -0,0 +1,36 @@
...
+import * as React from 'react'
+import * as _ from 'lodash'
+import { ComponentSlotClasses } from '../themes/types'
+
+export interface RenderStardustResultConfig {
+  classes: ComponentSlotClasses
+  rtl: boolean
+}
+
+export type RenderStardustComponentCallback<P> = (
```
bmdalex: I would just inline this type to the render method, it's lightweight and it'd make things more readable

src/index.ts
```diff
@@ -125,3 +125,4 @@ export { default as dialogBehavior } from './lib/accessibility/Behaviors/Dialog/
 // Utilities
 //
 export { default as mergeThemes } from './lib/mergeThemes'
+export { createComponent } from './lib'
```
bmdalex, line 128: I'm not sure if we should we export some types as well?

bmdalex (commented):
left comments

src/lib/createStardustComponent.tsx
mnajdova: We are providing the classes to user, which they can safely use in the className props. The styles should be just input for the createComponent that will create the classes for the user.

src/lib/createStardustComponent.tsx
mnajdova: Not sure if we agreed on this, but from user perspective, if I want to pass all props to the thing I am rendering I would have to manually extract the things from this one object which are part of the props. Having two different properties, allows me to do this:
`<button {...props} />` I would say we can return object which contains {config, props}, but this seems cleaner to me. I am open for changes if we decide that the other proposal should be taken.

src/index.ts
mnajdova: Not sure on this either, maybe we should add them..

mnajdova: I exported them

src/lib/createStardustComponent.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { ComponentSlotClasses } from '@stardust-ui/react'
+
+export interface RenderStardustResultConfig {
+  classes: ComponentSlotClasses
+  rtl: boolean
+}
+
+export interface CreateStardustComponentConfig<P> {
+  displayName?: string
+  render: (config: RenderStardustResultConfig, props: P) => React.ReactNode
```
layershifter: As a I remember, we decided  to go with one argument:

```diff
- render: (config: RenderStardustResultConfig, props: P) => React.ReactNode
+ render: (props: P & { stardust: RenderStardustResultConfig }) => React.ReactNode
```

mnajdova: I left this comment regarding that: https://github.com/stardust-ui/react/pull/503#discussion_r235078288 but I am open to change the implementation if there is a reason why we want that.

layershifter: We have a discussion about this with Levi. The signature with one argument matches an `React.SFC` signature:

```jsx
export const A: React.SFC = (props) => <div />

export const B = createComponent({
  render: A,
})
```

A little crazy, but it will work. Note sure that this will not break rendering, but SFC is a same render function.

mnajdova: Got it, okay, I will change it. Thanks for the clarification.

src/lib/createStardustComponent.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+const createComponent = <P extends {} = {}, S extends {} = {}>({
+  displayName = 'StardustComponent',
+  render,
+}: CreateStardustComponentConfig<P>): React.SFC<P> => {
+  return createComponentInternal<P, S>({
+    displayName,
+    render: (config, props) => {
+      const filteredConfig = pick(config, ['classes', 'rtl'])
+      return render(Object.assign({ stardust: filteredConfig }, props))
```
mnajdova, line 24: I had to use here Object.assing as spread is not supported for generics :\ https://github.com/Microsoft/TypeScript/issues/10727

src/lib/createStardustComponent.tsx
mnajdova: Changed!

src/index.ts
levithomason: Always export all typings that are for the public API.

src/lib/createStardustComponent.tsx
```diff
@@ -0,0 +1,29 @@
+import createComponentInternal from './createComponent'
+import * as React from 'react'
+import { pick } from 'lodash'
```
levithomason: Always keep code consistent. It makes maintenance and contributions easier. import * as here. We'll optimize this in our move to a babel build.

levithomason (approved):
Minor changes requested, no blockers from me

src/lib/createStardustComponent.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { ComponentSlotClasses } from '@stardust-ui/react'
```
layershifter: This thing should be imported from `/lib` or `/component` I think.

src/lib/createStardustComponent.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export interface RenderStardustResultConfig {
+  classes: ComponentSlotClasses
+  rtl: boolean
+}
+
+export interface CreateStardustComponentConfig<P> {
+  displayName?: string
```
layershifter: May be `displayName` is required?

src/lib/createStardustComponent.tsx
layershifter: Yep, it's a known issue

levithomason:
Next up we'll publish some docs and examples showing how to use this. There should be a new Guide for "Custom Components".

src/lib/createStardustComponent.tsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+const createComponent = <P extends {} = {}, S extends {} = {}>({
+  displayName = 'StardustComponent',
+  render,
+}: CreateStardustComponentConfig<P>): React.SFC<P> => {
+  return createComponentInternal<P, S>({
+    displayName,
+    render: (config, props) => {
+      const filteredConfig = pick(config, ['classes', 'rtl'])
```
kuzhelov: agree with https://github.com/stardust-ui/react/pull/503/files#r235287115 - now it is quite hard to immediately reason about what `pick(..)` is - in contrast to `_.pick(..)` suggested

kuzhelov (approved):

src/lib/createStardustComponent.tsx
mnajdova: Changed, imported from themes/types

alinais (approved):

brianespinosa:
Sweet!

PR #505, opened 2018-11-20
BREAKING(Transition): renamed Transition to Animation and added docs for the animations

mnajdova:
This PR is adding docs about how the animations should be defined in the theme, and used together with the Transition component, as well as the animation property.
# Update 
As part of this PR, I am going to also rename the `Transition` component to `Animation`, and the `animationName` property to `name`, following the review comments.


docs/src/views/Theming.tsx
```diff
@@ -208,6 +208,82 @@ export default () => (
...
+      define the animations in a very similar way as you would define them using css, by providing a
+      keyframe, as well as the other animation properties: duration, delay, timingFunction etc.
+    </p>
+
+    <p>
+      You can define the <code>animations</code> in a part of your render tree using the{' '}
+      <NavLink to="components/provider">Provider</NavLink>.
+    </p>
+    <p>
+      This is done with the Provider's <code>theme</code> prop. The animations are then are applied
```
miroslavstastny: > The animations are then *are* applied
remove the second 'are'

docs/src/views/Theming.tsx
```diff
@@ -208,6 +208,82 @@ export default () => (
...
+      based on their name by using the <NavLink to="components/transition">Transition</NavLink>{' '}
+      component, or the <code>animation</code> property available on all UI component. Here's how we
```
miroslavstastny: nit: *all UI component* - I would rather write something like *all Stardust components*

miroslavstastny (commented):
The first question which comes to my mind is *when I should use `Transition` and when `animation`? In the original animation PR, there was a clear description and comparison of the two. Would it make sense to add similar description here as well?

levithomason:
Note, PR title would be "docs(theming):..."

docs/src/views/Theming.tsx
```diff
@@ -208,6 +208,82 @@ export default () => (
...
+        `>`,
+        `  <div>`,
+        `    <Transition animationName="spinner"><Icon name="user" circular /></Transition>`,
+        `    <Icon name="book" animation="spinner" circular/>`,
+        `  </div>`,
+        `</Provider>`,
+      ].join('\n')}
+      render={() => (
+        <div>
+          <Transition animationName="spinner">
```
levithomason: This strikes me as odd, "Transition animationName=..." Should we rename this "Transition name=..." or "Animation name=..."?

docs/src/views/Theming.tsx
```diff
@@ -208,6 +208,82 @@ export default () => (
       <NavLink to="components/provider">Provider</NavLink> at the root of your app.
     </p>
 
+    <Header as="h2" content="Animations" />
+    <p>
+      Another important part of the theming in Stardust are the <code>animations</code>. You can
```
levithomason: >Another important part of the theming in Stardust are the <code>animations</code>.

This line doesn't add any value to the user's goal. Suggest removing it.

docs/src/views/Theming.tsx
```diff
@@ -208,6 +208,82 @@ export default () => (
...
+      define the animations in a very similar way as you would define them using css, by providing a
```
levithomason: Suggest:

>You define animations in Stardust in a very similar way to CSS, by providing keyframes and animation properties.

Then show a minal provider code snippet immediately following this.

docs/src/views/Theming.tsx
mnajdova: Thanks for the suggestion, I will rename the `Transition` component to `Animation`, as that is the word we use for describing this behavior everywhere else. `Transition` was just picked up because of SUIR. Then we can safely rename the `animationName` prop to `name`. I will add this changes to this PR too.

docs/src/examples/components/Animation/Types/AnimationExample.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
+import { Animation, Icon } from '@stardust-ui/react'
+
+const AnimationExample = () => (
+  <Animation name="spinner">
```
miroslavstastny: All the `Animation` examples are theme dependent. Would it make more sense to add `Provider` to the examples and explicitly define the animations there?
Not sure about that, would like to hear other opinions.

docs/src/views/Theming.tsx
```diff
@@ -208,6 +208,106 @@ export default () => (
...
+    <p>
+      You can also override some of the defined <code>animation</code> properties, by providing
+      additional properties to the <code>Animation</code> component, or the <code>animation</code>{' '}
+      prop.
+    </p>
+
+    <ExampleSnippet
+      value={[
+        `<Animation name="spinner" delay="2s" duration="1s"><Icon name="user" circular /></Animation>`,
+        `<Icon name="book" animation="spinner" delay="5s" duration="2s" circular/>`,
```
miroslavstastny: `animation` here should be an object:
```jsx
<Icon name="book" animation={{name: "spinner", delay:"5s", duration:"2s"}} circular/>
```

docs/src/views/Theming.tsx
```diff
@@ -208,6 +208,106 @@ export default () => (
...
+          <Animation name="spinner">
+            <Icon name="user" circular />
+          </Animation>
+          <Icon name="book" animation="spinner" circular />
+        </div>
+      )}
+    />
+
+    <p>
+      You can also override some of the defined <code>animation</code> properties, by providing
```
miroslavstastny, line 296: I would explicitly mention here that `keyframe` cannot be overriden and can be only defined in the `Provider`.

docs/src/views/Theming.tsx
```diff
@@ -208,6 +208,106 @@ export default () => (
...
+      value={[
+        `<Animation name="spinner" delay="2s" duration="1s"><Icon name="user" circular /></Animation>`,
+        `<Icon name="book" animation="spinner" delay="5s" duration="2s" circular/>`,
+      ].join('\n')}
+      render={() => (
+        <div>
+          <Animation name="spinner" delay="2s" duration="1s">
+            <Icon name="user" circular />
+          </Animation>
+          <Icon name="book" animation="spinner" delay="5s" duration="2s" circular />
```
miroslavstastny: `animation` -> object

docs/src/views/Theming.tsx
```diff
@@ -208,6 +208,106 @@ export default () => (
...
+        </div>
+      )}
+    />
+
+    <p>
+      The difference between using the Animation component versus the animation property is that,
+      the Animation component can be safely use for applying animations on{' '}
```
miroslavstastny: use -> used

docs/src/examples/components/Animation/Types/AnimationExample.tsx
mnajdova: I understand your point, but I don't think this is good idea, because we will have much code in the examples, for something that is documented in the Theming guide. How the animation are created will be document as part of the theming guidem and the default theme will have few animations that will be shown in the examples. Let's hear the opinion from the other folks too. @kuzhelov @Bugaa92 @alinais @levithomason @layershifter

docs/src/views/Theming.tsx
mnajdova: Right, thanks for the catch!

docs/src/views/Theming.tsx
mnajdova: Good point! Will be sure to mention it.

docs/src/examples/components/Animation/Types/AnimationExampleDelay.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react'
+import { Animation, Icon } from '@stardust-ui/react'
+
+const AnimationExampleDelay = () => (
+  <div>
+    {'This animation will start after 5 seconds'}
```
layershifter:
```suggestion
    This animation will start after 5 seconds
```

Are brackets necessary?

docs/src/examples/components/Animation/Types/AnimationExample.tsx
layershifter: I have a similar problem in #451, we definitely need to show theme specific stuff because examples can be broken when you will try to switch to another theme.

> I understand your point, but I don't think this is good idea, because we will have much code in the examples, for something that is documented in the Theming guide. 

This is a valid point, too.

docs/src/examples/components/Animation/Types/AnimationExampleDelay.tsx
mnajdova: Removed.

docs/src/examples/components/Animation/Types/AnimationExample.tsx
mnajdova: After offline discussion we decided to add the Provider with the theme. All examples as well as the Theming.tsx guide page, now contains a Provider where the animations are defined.

miroslavstastny (approved):

PR #509, opened 2018-11-21
[WIP] BREAKING(refs): add support of automatic refs

layershifter:
Refs #417.

--- 

This PR will add automatic refs to all our components. The basic idea of this is quite simple, I tried to document it in comments however I also want to describe it separately in this PR.

---

### 1. Wrap all our components with `forwardRef()`

`forwardRef()` is a single API that allows you to pass down refs to elements. In fact, `forwardRefFactory()` is only a wrapper around it. We will wrap all components with it.

This means that we will receive a `ref` in our factory and catch it:

```jsx
<Button ref={/* this ref */} />
```

### 2. Pass down intercepted refs

1. Any intercepted ref will be assigned to the `forwardRefSymbol` prop.
2. Because `forwardRefSymbol` is a simple prop we will have it in unhandled props
3. As all unhandled props are assigned to an `ElementType` element it will work without any code changes

image


docs/src/examples/components/Avatar/Variations/AvatarExampleStatusCustomization.shorthand.tsx
```diff
@@ -54,12 +54,12 @@ const AvatarExampleStatusCustomizationShorthand = () => (
...
     <Avatar
       image={{ src: 'public/images/avatar/small/matt.jpg', alt: 'Profile picture of Matt' }}
-      size={Avatar.defaultProps.size * 1.5}
+      size={48}
       status={{
         color: 'green',
         icon: 'check',
         title: 'Available',
-        size: Status.defaultProps.size * 1.5,
+        size: 48,
```
layershifter, line 62: I will move this to a separate PR.

src/components/Status/Status.tsx
```diff
@@ -75,5 +72,9 @@ class Status extends UIComponent<Extendable<StatusProps>, any> {
 
 Status.create = createShorthandFactory(Status, 'state')
 
-export default Status
+/**
+ * A status graphically represents someone's or something's state.
+ */
+export default forwardRefFactory(Status)
```
layershifter, line 75: One of side effects, a component's description should be on `export`.

src/lib/forwardRefFactory/componentUtils.ts
```diff
@@ -0,0 +1,30 @@
...
+ * We can't use there `isForwardRef()` from the `react-is` package because it asserts an element, not a component.
+ * @see https://github.com/facebook/react/blob/v16.5.2/packages/react/src/forwardRef.js#L42
+ * @see https://github.com/facebook/react/blob/v16.6.3/packages/shared/isValidElementType.js
+ * @see https://codesandbox.io/s/kwkw3pq1k3
+ * @param Component A component to check
+ */
+export const isForwardRefComponent = (Component: React.ReactType): boolean =>
+  // Heads up!
+  // Any is used there because type's definitions do not contain this internal property.
+  typeof Component === 'object' && (Component as any).$$typeof === ForwardRef
```
kuzhelov, line 16: is it part of public API?

layershifter: Nope, and it's a problem. I don't know any other way to perform this check, I have even made a codesandbox that shows my problem: https://codesandbox.io/s/kwkw3pq1k3

I'm open for better proposals there.

layershifter:
https://github.com/stardust-ui/react/pull/587#issuecomment-445905932

PR #510, opened 2018-11-21
docs(Accessibility): Display available behaviors for component

jurokapsiar:
Display available behaviors for component.

Available behaviors can be specified using `@available` tag in the `accessibility` prop comment:
```
  /**
   * Accessibility behavior if overridden by the user.
   * @default menuBehavior
   * @available toolbarBehavior, tabListBehavior
   * */
  accessibility?: Accessibility
```

image

The source of truth in this case is by purpose the component, as the plan forward is to separate behaviors to a separate repo. Then, behaviors will not have any knowledge about the components - components will only use the behaviors.

Once the issue with getting the default value using docgen directly will be resolved, we can also change this to use a compiler-checked reference.


layershifter (changes requested):
_My opinion:_ this is a cool feature, but we should keep these things in `docgen` task because our docs components will become overcomplicated. We have [doctrine](https://github.com/eslint/doctrine), I almost sure that it can do the same more clearly.

This also can cause performance degrades.

build/gulp/plugins/util/getComponentInfo.ts
```diff
@@ -111,7 +111,29 @@ const getComponentInfo = (filepath: string, checksum?: string) => {
   // sort props
   info.props = _.sortBy(info.props, 'name')
 
+  // available behaviors
+  info.behaviors = getAvailableBehaviors(info.props.accessibility)
+
   return info
 }
 
+const getAvailableBehaviors = accessibilityProp => {
```
layershifter:
```suggestion
const getAvailableBehaviors = (accessibilityProp): string[] => {
```

Can we add an input and result type there? It's our internal code, but it we more understandable.

build/gulp/plugins/util/getComponentInfo.ts
```diff
@@ -111,7 +111,29 @@ const getComponentInfo = (filepath: string, checksum?: string) => {
...
+  const availableBehaviorNames =
+    (accessibilityProp &&
+      _.get(_.find(accessibilityProp.tags, { title: 'available' }), 'description')) ||
```
layershifter: Nesting is hard to read, may be the code below will be more readable?

```js
const docTags = accessibilityProp.tags

const availableTag= _.find(docTags, { title: 'available' }))
const availableBehaviorNames = _.get(availableTag, 'description', '')
```

As I understand it will also allow us to drop a condition below:

```diff
 return (
--    availableBehaviorNames &&
    availableBehaviorNames
      .replace(/\s/g, '')
      .split(',')
      .map(name => ({
        name,
        displayName: _.upperFirst(name.replace('Behavior', '')),
        category: _.upperFirst(name.split(/(?=[A-Z])/)[0]),
      }))
  )
```

build/gulp/plugins/util/getComponentInfo.ts
jurokapsiar: added an interface for output, input is part of `any`, so I just added `{ tags: [] }`

layershifter (approved):

sophieH29 (approved):

PR #514, opened 2018-11-22
docs(ChatPane): Add attachments and link to chat pane messages in prototype

sophieH29:
This PR aims to improve chat pane behavior by adding Attachments and links to chat message so provide the proper accessibility experience for it.

1. Navigate with UP / DOWN arrow keys between messages
2. Click Enter on focused message
3. The link will get focus first
4. Press Tab
5. First attachment will get focus
6. Press Tab
7. Action button on the attachment will get focus
8. Press Enter
9. Popup will open with context menu and trap the focus
10. Press Esc 
11. Focus goes to action trigger
12. Press Tab and go to the next attachment OR press Esc and focus message again, so you'll be able to navigate between messages with UP/DOWN arrow keys

image

Opened context menu:
image

### DIFF
--- a/docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
+++ b/docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
@@ -34,7 +34,7 @@ const statusMap: Map<UserStatus, StatusPropsExtendable> = new Map([
 function generateChatMsgProps(msg: MessageData, fromUser: UserData): ChatMessage {
   const { content, mine } = msg
   const msgProps: ChatMessage = {
-    content: getMessageContent(content),
+    content: createMessageContent(content),
     mine,
     tabIndex: 0,
     timestamp: { content: msg.timestamp, title: msg.timestampLong },
@@ -46,7 +46,7 @@ function generateChatMsgProps(msg: MessageData, fromUser: UserData): ChatMessage
   return msgProps
 }
 
-function getMessageContent(content: string) {
+function createMessageContent(content: string) {
   const contextMenu = (
     <Menu
       items={[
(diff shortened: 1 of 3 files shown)


jurokapsiar (approved):

docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
```diff
@@ -44,6 +46,53 @@ function generateChatMsgProps(msg: MessageData, fromUser: UserData): ChatMessage
...
+
+  return (
+    <>
+      <span>
+        {content} <a href="/"> Some link </a>
+      </span>
+      <div style={{ marginTop: '20px' }}>
+        {_.map(['MeetingNotes.pptx', 'Document.docx'], (fileName, index) => {
+          return (
+            <Attachment
```
layershifter: We don't need there `return`:

```diff
- {_.map(['MeetingNotes.pptx', 'Document.docx'], (fileName, index) => {
-          return (
-            <Attachment
+{_.map(['MeetingNotes.pptx', 'Document.docx'], (fileName, index) => (
+  <Attachment
```

This will allow to reduce the nesting level 👍

docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
```diff
@@ -44,6 +46,53 @@ function generateChatMsgProps(msg: MessageData, fromUser: UserData): ChatMessage
   return msgProps
 }
 
+function getMessageContent(content: string) {
```
layershifter: In fact, this is a content factory, may be other name?

```suggestion
function createMessageContent(content: string) {
```

layershifter (approved):
Left one small note, LGTM

mnajdova (commented):
The changes looks good, have tested it locally and it is behaving as in the description. However, it is a bit overwhelming for all messages to have attachments. Can you at least add the attachments on every 5th message let's say? Will approve after this change. No other changes required from my side.

mnajdova (approved):

PR #517, opened 2018-11-22
docs(custom-components): adding initial guide for creating custom components

mnajdova:
This PR adds guide for how the custom components can be created using the renderComponent functionality in Stardust. 
To do
- [x] update the changelog

### DIFF
--- a/docs/src/views/CustomComponents.tsx
+++ b/docs/src/views/CustomComponents.tsx
@@ -1,0 +1,235 @@
+import * as React from 'react'
+import * as cx from 'classnames'
+import * as _ from 'lodash'
+import { NavLink } from 'react-router-dom'
+import { Header } from 'semantic-ui-react'
+import {
+  Button,
+  Divider,
+  Provider,
+  createComponent,
+  ComponentSlotStyle,
+  ComponentVariablesInput,
+} from '@stardust-ui/react'
+
+import DocPage from '../components/DocPage/DocPage'
+import ExampleSnippet from '../components/ExampleSnippet/ExampleSnippet'
+
+interface StyledButtonProps {
+  className?: string
+  styles?: ComponentSlotStyle
+  variables?: ComponentVariablesInput
+  children?: React.ReactNodeArray | React.ReactNode
+}
+
+const StyledButton: React.SFC<StyledButtonProps> = createComponent<StyledButtonProps, {}>({
+  displayName: 'StyledButton',
+  render({ stardust, ...props }) {
+    const { classes } = stardust
+    const componentProps = _.pick(props, 'children')
+    return <button className={cx(props.className, classes.root)} {...componentProps} />
+  },
+})
+
+export default () => (
+  <DocPage title="Custom Components">
+    <Header as="h2" content="Overview" />
+    <p>
+      You can use your own components as part of the Stardust's theming mechanism. In order for all
+      theming aspects to be available to your custom components, you should use the{' '}
+      <code>createComponent</code>
+      function, provided by the Stardust library.
+    </p>
+
+    <Header as="h2" content="Create custom component" />
+    <p>
+      Let's take a look into one simple example of using the <code>createComponent</code> function
+      for adapting your custom component to the Stardust's theming mechanism.
+    </p>
+
+    <ExampleSnippet
+      value={[
+        `import { createComponent } from '@stardust-ui/react'`,
+        ``,
+        `export interface StyledButtonProps {`,
+        `  className?: string`,
+        `  styles?: ComponentSlotStyle`,
+        `  variables?: ComponentVariablesInput`,
+        `  children?: React.ReactNodeArray | React.ReactNode`,
+        `}`,
+        ``,
+        `const StyledButton: React.SFC<StyledButtonProps> = createComponent<StyledButtonProps, {}>({`,
+        `  displayName: 'StyledButton',`,
+        `  render: ({stardust, ...props}) => {`,
+        `    const { classes } = stardust`,
+        `    const componentProps = _.pick(props, 'children')`,
+        `    return (<button className={cx(props.className, classes.root)} {...componentProps} />)`,
+        `  }`,
+        `})`,
+      ].join('\n')}
+    />
+
+    <p>Let's go step by step throughout all bits from the renderComponent method.</p>
+
+    <p>
+      The first argument to the <code>createComponent</code> config's param is the displayName. This
+      name is very important, because that's the key that will be used in the Provider's theme if
+      you want to define some styles or variables for this custom component. Let's see one example:
+    </p>
+
+    <ExampleSnippet
+      value={[
+        `<Provider`,
+        `  theme={{`,
+        `    componentStyles: {`,
+        `      StyledButton: {`,
+        `        root: ({ props, variables, theme: {siteVariables} }) => ({`,
+        `          backgroundColor: siteVariables.brand,`,
+        `          color: (variables as any).color,`,
+        `        }),`,
+        `      },`,
+        `    },`,
+        `    componentVariables: {`,
+        `      StyledButton: {`,
+        `        color: '#F2F2F2',`,
+        `      },`,
+        `    },`,
+        `  }}`,
+        `>`,
+        `  <StyledButton> Provider styled button </StyledButton>`,
+        '</Provider>',
+      ].join('\n')}
+      render={() => (
+        <Provider
+          theme={{
+            componentStyles: {
+              StyledButton: {
+                root: ({ props, variables, theme: { siteVariables } }) => ({
+                  backgroundColor: siteVariables.brand,
+                  color: (variables as any).color,
+                }),
+              },
+            },
+            componentVariables: {
+              StyledButton: {
+                color: '#F2F2F2',
+              },
+            },
+          }}
+        >
+          <StyledButton>Provider styled button</StyledButton>
+        </Provider>
+      )}
+    />
+
+    <p>
+      The second argument of the <code>createComponent</code> config param is the render method.
+      This is the place where you link the Stardust bits with your custom component. This method is
+      invoked with the following parameters:
+    </p>
+    <ul>
+      <li>
+        <code>stardust</code> - this is the object containing the evaluated theming props (classes
+        and rtl)
+      </li>
+      <li>
+        <code>...props</code> - all other props provided by the user
+      </li>
+    </ul>
+    <p>
+      The previous implementation of the render method inside the <code>createComponent</code>{' '}
+      function for the
+      <code>StyledButton</code>, allows the user, to use the styles and variables properties on the
+      custom component in the same way they would be used for any Stardust component.
+    </p>
+
+    <Header as="h2" content="Using the custom components" />
+
+    <p>
+      We already saw how the <code>Provider</code> can define some stylings and variables for the
+      custom components. Next, we will take a look into few examples of how the user can further
+      customize the styling and variables of the custom components, the same way they would do with
+      the Stardust components.
+    </p>
+
+    <Header as="h3" content="Example 1. Using the variables property" />
+
+    <ExampleSnippet
+      value={[
+        `<StyledButton variables={(siteVars) => ({color: siteVars.black})}>`,
+        `  Inline styled button`,
+        `</StyledButton>`,
+      ].join('\n')}
+      render={() => (
+        <Provider
+          theme={{
+            componentStyles: {
+              StyledButton: {
+                root: ({ props, variables, theme: { siteVariables } }) => ({
+                  backgroundColor: siteVariables.brand,
+                  color: (variables as any).color,
+                }),
+              },
+            },
+            componentVariables: {
+              StyledButton: {
+                color: '#F2F2F2',
+              },
+            },
+          }}
+        >
+          <StyledButton variables={siteVariables => ({ color: siteVariables.black })}>
+            inline styled button
+          </StyledButton>
+        </Provider>
+      )}
+    />
+
+    <Header as="h3" content="Example 2. Using the styles property" />
+
+    <ExampleSnippet
+      value={[
+        `<StyledButton styles={({theme: {siteVariables}}) => ({backgroundColor: siteVariables.black})}>`,
+        `  Inline styled button`,
+        `</StyledButton>`,
+      ].join('\n')}
+      render={() => (
+        <Provider
+          theme={{
+            componentStyles: {
+              StyledButton: {
+                root: ({ props, variables, theme: { siteVariables } }) => ({
+                  backgroundColor: siteVariables.brand,
+                  color: (variables as any).color,
+                }),
+              },
+            },
+            componentVariables: {
+              StyledButton: {
+                color: '#F2F2F2',
+              },
+            },
+          }}
+        >
+          <StyledButton
+            styles={({ theme: { siteVariables } }) => ({ backgroundColor: siteVariables.black })}
+          >
+            Inline styled button
+          </StyledButton>
+        </Provider>
+      )}
+    />
+
+    <br />
+    <Divider size={1} />
+    <br />
+    <Button
+      as={NavLink}
+      content="Theming Examples"
+      icon="arrow left"
+      iconPosition="before"
+      primary
+      to="/theming-examples"
+    />
+  </DocPage>
+)
--- a/docs/src/views/CustomComponents.tsx
+++ b/docs/src/views/CustomComponents.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react'
 import * as cx from 'classnames'
-import * as _ from 'lodash'
 import { NavLink } from 'react-router-dom'
 import { Header } from 'semantic-ui-react'
 import {
@@ -24,10 +23,9 @@ interface StyledButtonProps {
 
 const StyledButton: React.SFC<StyledButtonProps> = createComponent<StyledButtonProps, {}>({
   displayName: 'StyledButton',
-  render({ stardust, ...props }) {
+  render({ stardust, className, children }) {
     const { classes } = stardust
-    const componentProps = _.pick(props, 'children')
-    return <button className={cx(props.className, classes.root)} {...componentProps} />
+    return <button className={cx(className, classes.root)}>{children}</button>
   },
 })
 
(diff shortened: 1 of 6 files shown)


docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,235 @@
...
+        `  render: ({stardust, ...props}) => {`,
+        `    const { classes } = stardust`,
+        `    const componentProps = _.pick(props, 'children')`,
+        `    return (<button className={cx(props.className, classes.root)} {...componentProps} />)`,
+        `  }`,
+        `})`,
+      ].join('\n')}
+    />
+
+    <p>Let's go step by step throughout all bits from the renderComponent method.</p>
```
bmdalex: ~~`renderComponent`~~ ->  `createComponent`
I also changed this in the CHANGELOG.md file that seems to mention `renderComponent` as well
https://github.com/stardust-ui/react/pull/518/files#r235773446

bmdalex (approved):

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,237 @@
...
+}
+
+const StyledButton: React.SFC<StyledButtonProps> = createComponent<StyledButtonProps, {}>({
+  displayName: 'StyledButton',
+  render({ stardust, ...props }) {
+    const { classes } = stardust
+    const componentProps = _.pick(props, 'children')
+    return <button className={cx(props.className, classes.root)} {...componentProps} />
+  },
+})
```
layershifter: I think that we should omit the usage of `lodash` in such basic examples:

```js
const StyledButton: React.SFC<StyledButtonProps> = createComponent<StyledButtonProps, {}>({
  displayName: 'StyledButton',
  render({ className, stardust, ...props }) {
    const { classes } = stardust

    return <button {...rest} className={cx(className, classes.root)} />
  },
})
```

As I see, my code does the same thing or no?

mnajdova: We are applying in this case the styles and variables to the button, but I get your point, will change it not to use lodash.

mnajdova: Something like this?
```jsx
const StyledButton: React.SFC<StyledButtonProps> = createComponent<StyledButtonProps, {}>({
  displayName: 'StyledButton',
  render({ stardust, className, children }) {
    const { classes } = stardust
    const componentProps = { children }

    return <button className={cx(className, classes.root)} {...componentProps} />
  },
})
```

layershifter:
```diff
-return <button className={cx(className, classes.root)} {...componentProps} />
+return <button className={cx(className, classes.root)}>{children}</button>
```

👍

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+const StyledButton: React.SFC<StyledButtonProps> = createComponent<StyledButtonProps, {}>({
+  displayName: 'StyledButton',
+  render({ stardust, className, children }) {
+    const { classes } = stardust
+    return <button className={cx(className, classes.root)}>{children}</button>
+  },
+})
+
+export default () => (
+  <DocPage title="Custom Components">
```
kuzhelov: maybe something like "Integrate Custom Components" - because currently this is not absolutely clear from header whether we are suggesting to **develop** custom components or **integrate** existing ones

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+    <Header as="h2" content="Overview" />
+    <p>
+      You can use your own components as part of the Stardust's theming mechanism. In order for all
```
kuzhelov: 'styling and theming'

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+    <p>
+      Let's take a look into one simple example of using the <code>createComponent</code> function
+      for adapting your custom component to the Stardust's theming mechanism.
+    </p>
+
+    <ExampleSnippet
+      value={[
+        `import { createComponent } from '@stardust-ui/react'`,
+        ``,
+        `export interface StyledButtonProps {`,
```
kuzhelov: I would suggest to remove TS from this example - main reason is that it introduces quite a lot of 'noise', and additional one (and the one that applies to broader community) is that our docs should be readable for folks that have no prior knowledge of TS

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+        `})`,
+      ].join('\n')}
+    />
+
+    <p>
+      Let's go step by step throughout all bits from the <code>createComponent</code> method.
+    </p>
+
+    <p>
+      The first argument to the <code>createComponent</code> config's param is the displayName. This
```
kuzhelov: would suggest to also wrap `displayName` and `Provider` in `<code/>`

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+      name is very important, because that's the key that will be used in the Provider's theme if
```
kuzhelov: we could simplify this sentence a bit:

> is the displayName, which value might be used as key to define component's styles and variables in theme, exactly the same way how it might be done for any first-class Stardust component.

I think that this will provide enough thrill to the reader, so one will fully understand the importance of this prop ;)

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+      you want to define some styles or variables for this custom component. Let's see one example:
+    </p>
+
+    <ExampleSnippet
+      value={[
+        `<Provider`,
+        `  theme={{`,
+        `    componentStyles: {`,
+        `      StyledButton: {`,
```
kuzhelov: would suggest to swap `variables` and `styles` sections, as the later one uses data defined in former.

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+            },
+          }}
+        >
+          <StyledButton>Provider styled button</StyledButton>
+        </Provider>
+      )}
+    />
+
+    <p>
+      The second argument of the <code>createComponent</code> config param is the render method.
```
kuzhelov: lets wrap `render` in `<code>` as well

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+      This is the place where you link the Stardust bits with your custom component. This method is
```
kuzhelov: nit-nit:

> ... where you **might** link Stardust bits with your custom component - **e.g. by simply passing them as props**. This **`render`** method **will be** invoked with the following parameters:

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+      invoked with the following parameters:
+    </p>
+    <ul>
+      <li>
+        <code>stardust</code> - this is the object containing the evaluated theming props (classes
```
kuzhelov: nit: 'this is' is not necessary, can be safely omitted.

`classes` and `rtl` should be wrapped in `<code />`

kuzhelov: lets end each item's text with dot

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+        and rtl)
+      </li>
+      <li>
+        <code>...props</code> - all other props provided by the user
+      </li>
+    </ul>
+    <p>
+      The previous implementation of the render method inside the <code>createComponent</code>{' '}
```
kuzhelov: sorry, not sure that can completely understand what this paragraph is about - why `previous` and, also, it seems that there are too many commas there

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+      function for the
+      <code>StyledButton</code>, allows the user, to use the styles and variables properties on the
+      custom component in the same way they would be used for any Stardust component.
+    </p>
+
+    <Header as="h2" content="Using the custom components" />
+
+    <p>
+      We already saw how the <code>Provider</code> can define some stylings and variables for the
+      custom components. Next, we will take a look into few examples of how the user can further
```
kuzhelov: 'few' -> 'several/couple' (as 'few' has rather 'not enough' semantics :)

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+      customize the styling and variables of the custom components, the same way they would do with
```
kuzhelov: nit: maybe simplified:
customize the styling and variables -> customize styles and variables of **these** components

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+      the Stardust components.
+    </p>
+
+    <Header as="h3" content="Example 1. Using the variables property" />
```
kuzhelov: would omit `the` and use `<code>`:

```jsx
<Header as="h3" content={<>Example 1. Using <code>variables</code> property</>} />
```

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+    <ExampleSnippet
+      value={[
+        `<StyledButton variables={(siteVars) => ({color: siteVars.black})}>`,
+        `  Inline styled button`,
+        `</StyledButton>`,
+      ].join('\n')}
+      render={() => (
+        <Provider
+          theme={{
+            componentStyles: {
```
kuzhelov: same here - would suggest to swap places of `componentVariables` and `componentStyles` sections

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+          }}
+        >
+          <StyledButton variables={siteVariables => ({ color: siteVariables.black })}>
+            inline styled button
+          </StyledButton>
+        </Provider>
+      )}
+    />
+
+    <Header as="h3" content="Example 2. Using the styles property" />
```
kuzhelov: just 'using <code>styles</code> property' should be enough

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+    <ExampleSnippet
+      value={[
+        `<StyledButton styles={({theme: {siteVariables}}) => ({backgroundColor: siteVariables.black})}>`,
```
kuzhelov: I would rather strive for simplifying examples as much as possible. We should just show that it is _possible_ to use styles prop, as well as suggest the fact that it could be used the same way as for Stardust components. Then I see us providing simple example that might prove that, and after this example make a sentence that **all other advanced scenarios are supported as well, please, take a look on the corresponding [docs](#)**. 

This will allow us
-  to simplify example's code (crucially important thing)
- not introduce additional maintainability burden to us - say, in case if implementation of `styles` prop change, we won't need to fix this example's code

docs/src/views/CustomComponents.tsx
kuzhelov: also, would rather expect to see section devoted to `styles` prop first, as this is a first thing customer would learn about Stardust - with `variables` being rather more advanced concept

docs/src/views/CustomComponents.tsx
```diff
@@ -0,0 +1,234 @@
...
+      custom components. Next, we will take a look into few examples of how the user can further
+      customize the styling and variables of the custom components, the same way they would do with
+      the Stardust components.
+    </p>
+
+    <Header as="h3" content="Example 1. Using the variables property" />
+
+    <ExampleSnippet
+      value={[
+        `<StyledButton variables={(siteVars) => ({color: siteVars.black})}>`,
```
kuzhelov: lets try to simplify it a bit - while I do understand that this example is aimed to cover all things that are possible to do with this prop being introduced, at the same time, while being in the client's shoes, I would rather expect something pretty simple that would just deliver necessary idea to me. Otherwise, just imagine that I am not yet familiar with `siteVariables` and what the beast it is - should I learn that or, maybe, it is fine for me to use `variables` without this knowledge? I would say that the later is just fine

docs/src/views/CustomComponents.tsx
mnajdova: This paragraph is not really necessary as we already covered all important bits in the paragraph before the list items :)

mnajdova:
@kuzhelov thanks much for the review! All comments resolved :)

docs/src/views/IntegrateCustomComponents.tsx
```diff
@@ -0,0 +1,238 @@
+import * as React from 'react'
+import * as cx from 'classnames'
+import { NavLink } from 'react-router-dom'
+import { Header } from 'semantic-ui-react'
```
layershifter, line 4: I think that we can use own `Header` component there 😃

docs/src/views/IntegrateCustomComponents.tsx
```diff
@@ -0,0 +1,238 @@
...
+} from '@stardust-ui/react'
+
+import DocPage from '../components/DocPage/DocPage'
+import ExampleSnippet from '../components/ExampleSnippet/ExampleSnippet'
+
+interface StyledButtonProps {
+  className?: string
+  styles?: ComponentSlotStyle
+  variables?: ComponentVariablesInput
+  children?: React.ReactNodeArray | React.ReactNode
```
layershifter: We have own type for this, `ReactChildren`

docs/src/views/IntegrateCustomComponents.tsx
```diff
@@ -0,0 +1,238 @@
...
+}
+
+const StyledButton: React.SFC<StyledButtonProps> = createComponent<StyledButtonProps, {}>({
```
layershifter: The state is optional.

```suggestion
const StyledButton: React.SFC<StyledButtonProps> = createComponent<StyledButtonProps>({
```

docs/src/views/IntegrateCustomComponents.tsx
```diff
@@ -0,0 +1,238 @@
...
+        `    componentVariables: {`,
+        `      StyledButton: {`,
+        `        color: '#F2F2F2',`,
+        `      },`,
+        `    },`,
+        `    componentStyles: {`,
+        `      StyledButton: {`,
+        `        root: ({ props, variables, theme: {siteVariables} }) => ({`,
+        `          backgroundColor: siteVariables.brand,`,
+        `          color: (variables as any).color,`,
```
layershifter:
```suggestion
        `          color: variables.color,`,
```

Can we omit `any` this in example's code?

docs/src/views/IntegrateCustomComponents.tsx
```diff
@@ -0,0 +1,238 @@
...
+            componentVariables: {
+              StyledButton: {
+                color: '#F2F2F2',
+              },
+            },
+            componentStyles: {
+              StyledButton: {
+                root: ({ props, variables, theme: { siteVariables } }) => ({
+                  backgroundColor: siteVariables.brand,
+                  color: (variables as any).color,
```
layershifter: The same thing, ^

docs/src/views/IntegrateCustomComponents.tsx
mnajdova: Switching the theme to dark will make the Header component white, and we don't have the other themes implemented for all page (like making the background dark in dark-mode). I would not like to introduce unnecessary providers just for handling things like that. In all docs pages, the Headers are from semantic-ui, so when the dark mode will be implemented for all page, we can safely replace them all in one iteration.

docs/src/views/IntegrateCustomComponents.tsx
mnajdova: We cannot omit it here, but it is removed from the string displayed as a code in the docs page.

docs/src/views/IntegrateCustomComponents.tsx
```diff
@@ -0,0 +1,239 @@
...
+        `  render: ({stardust, className, children}) => {`,
+        `    const { classes } = stardust`,
+        `    return <button className={cx(className, classes.root)}>{children}</button>`,
+        `  }`,
+        `})`,
+      ].join('\n')}
+    />
+
+    <p>
+      Let's go step by step throughout all bits from the <code>createComponent</code> method.
```
kuzhelov: nit: '.. from the createComponent ..' => '... of the createComponent ...'

docs/src/views/IntegrateCustomComponents.tsx
```diff
@@ -0,0 +1,239 @@
...
+    </p>
+    <ul>
+      <li>
+        <code>stardust</code> - the object containing the evaluated theming props (<code>
+          classes
+        </code>
+        and <code>rtl</code>).
+      </li>
+      <li>
+        <code>...props</code> - all other props provided by the user.
```
kuzhelov: nit: maybe `user props` would be enough

docs/src/views/IntegrateCustomComponents.tsx
```diff
@@ -0,0 +1,239 @@
...
+      content={
+        <>
+          Example 2. Using <code>variables</code> property
+        </>
+      }
+    />
+
+    <ExampleSnippet
+      value={[
+        `<StyledButton variables={{color: "black" }}>`,
```
kuzhelov: just to be sure - could you, please, suggest where is the code that suggests to the user where this variable is used (i.e. styles code)?

docs/src/views/IntegrateCustomComponents.tsx
```diff
@@ -0,0 +1,239 @@
...
+            },
+          }}
+        >
+          <StyledButton variables={{ color: 'black' }}>inline styled button</StyledButton>
+        </Provider>
+      )}
+    />
+
+    <p>
+      All other advanced scenarios for applying styles and variables are supported as well. For more
```
kuzhelov: Would suggest to write it this way: 

- for each example (Styles and Variables) introduce the concluding paragraph that would tell the following (essentially, the though expressed in this paragraph: 'Note that all other scenarios for applying styles/variables are supported as well - please, take a look on the corresponding ... section for more details.')

- introduce additional example where theming support will be mentioned (primitive example is necessary):

```jsx
const customTheme = {
  // component's displayName arg is used as a selector
  StyledButton: {
   ...
  }
}

<Provider theme={customTheme}>
  <StyledButton>
</Provider>
```

- and, the same way, conclude with a similar paragraph saying 'for more advanced theming scenarios, please, take a look on the corresponding Theming section'

kuzhelov (approved):

docs/src/views/IntegrateCustomComponents.tsx
mnajdova: Added paragraph with Provider and example componentStyles that would be applied.

PR #518, opened 2018-11-22
feat(chat): remove layout component and allow styling slots

bmdalex:
# feat(chat): remove layout component and allow styling slots

## Description

This PR includes the following changes:

#### 1. Refactors the render function for `Chat.Message` component to not use `Layout` component
`Layout` component does not allow the users to achieve proper styling of slots and it introduces a lot of unnecessary DOM nodes; we are using `Slot` instead.

#### 2. Reduces the size of the tree for `Chat.Message` by 50% (from 6 to 3):

- before:
screenshot 2018-11-22 at 16 19 34
- after:
screenshot 2018-11-22 at 16 18 38

#### 3. Adds functionality for targeting all subcomponents of `Chat.Message`:
- `avatar`;
- `messageBody` than contains `author`, `timestamp` and `content`;

#### 4. Docs example:
For showing the users the `Chat.Message` anatomy and how they can style every slot:
screenshot 2018-11-22 at 16 16 24

#### 5. Styles (see short demo below):
  - for focusing the correct area (`messageBody` instead of the whole `Chat.Message`)
  - Teams theme styles for focus outline for `messageBody`;
  - Teams theme styles for focus underline for links.
key-nav

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## [Unreleased]
 
 ### Features
-- Add `renderComponent` function in the public API @mnajdova ([#503](https://github.com/stardust-ui/react/pull/503))
+- Add `createComponent` function in the public API @mnajdova ([#503](https://github.com/stardust-ui/react/pull/503))
 - Apply `dir=auto` attribute to string content of `Text` @kuzhelov  ([#5](https://github.com/stardust-ui/react/pull/5))
 - Improve `Menu` accessibility behaviors @sophieH29 ([#523](https://github.com/stardust-ui/react/pull/523))
+- Add ability to style every slot of `Chat.Message` and remove dependency on `Layout` component @Bugaa92 ([#518](https://github.com/stardust-ui/react/pull/518))
 
 ### Fixes
 - Fix the behaviour of `AutoControlledComponent` when `undefined` is passed as a prop value @layershifter ([#499](https://github.com/stardust-ui/react/pull/499))
--- a/docs/src/components/ComponentDoc/ComponentSidebar/ComponentSidebar.tsx
+++ b/docs/src/components/ComponentDoc/ComponentSidebar/ComponentSidebar.tsx
@@ -44,11 +44,11 @@ class ComponentSidebar extends React.Component<any, any> {
     return (
       <Sticky context={examplesRef} offset={15}>
         <Menu as={Accordion} fluid style={sidebarStyle} text vertical>
-          {_.map(sections, ({ examples, sectionName }) => (
+          {_.map(sections, ({ examples, sectionName }, index) => (
             <ComponentSidebarSection
               activePath={activePath}
               examples={examples}
-              key={sectionName}
+              key={`${sectionName}-${index}`}
               sectionName={sectionName}
               onItemClick={onItemClick}
             />
--- a/docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx
+++ b/docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx
@@ -6,41 +6,62 @@ const janeAvatar = {
   status: { color: 'green', icon: 'check' },
 }
 
-const content = (
-  <div>
-    Sure! Let's try one of these places:<br />
-    <a href="#">www.goodFood1.com</a>,<br />
-    <a href="#">www.goodFood2.com</a> or<br />
-    <a href="#">www.goodFood3.com</a>
-  </div>
-)
-
 const items = [
-  <Chat.Message content="Hello" author="John Doe" timestamp="Yesterday, 10:15 PM" mine />,
-  <Chat.Message
-    content="Hi"
-    author="Jane Doe"
-    timestamp="Yesterday, 10:15 PM"
-    avatar={janeAvatar}
-  />,
-  <Chat.Message
-    content="Would you like to grab a lunch?"
-    author="John Doe"
-    timestamp="Yesterday, 10:16 PM"
-    mine
-  />,
-  <Chat.Message
-    content={{ content }}
-    author="Jane Doe"
-    timestamp="Yesterday, 10:15 PM"
-    avatar={janeAvatar}
-  />,
-  <Divider content="Today" type="primary" important />,
-  <Chat.Message content="Let's have a call" author="John Doe" timestamp="Today, 11:15 PM" mine />,
-].map((chatItemContent, index) => ({
-  key: `chat-item-${index}`,
-  content: chatItemContent,
-}))
+  {
+    content: (
+      <Chat.Message content="Hello" author="John Doe" timestamp="Yesterday, 10:15 PM" mine />
+    ),
+    key: 'message-id-1',
+  },
+  {
+    content: (
+      <Chat.Message
+        content="Hi"
+        author="Jane Doe"
+        timestamp="Yesterday, 10:15 PM"
+        avatar={janeAvatar}
+      />
+    ),
+    key: 'message-id-2',
+  },
+  {
+    content: (
+      <Chat.Message
+        content="Would you like to grab a lunch?"
+        author="John Doe"
+        timestamp="Yesterday, 10:16 PM"
+        mine
+      />
+    ),
+    key: 'message-id-3',
+  },
+  {
+    content: (
+      <Chat.Message
+        content="Sure! Let's try the new place downtown"
+        author="Jane Doe"
+        timestamp="Yesterday, 10:15 PM"
+        avatar={janeAvatar}
+      />
+    ),
+    key: 'message-id-4',
+  },
+  {
+    content: <Divider content="Today" type="primary" important />,
+    key: 'message-id-5',
+  },
+  {
+    content: (
+      <Chat.Message
+        content="Let's have a call"
+        author="John Doe"
+        timestamp="Today, 11:15 PM"
+        mine
+      />
+    ),
+    key: 'message-id-6',
+  },
+]
 
 const ChatExample = () => <Chat items={items} />
 
--- a/docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx
+++ b/docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx
@@ -6,56 +6,41 @@ const janeAvatar = {
   status: { color: 'green', icon: 'check' },
 }
 
+const content = (
+  <div>
+    Sure! Let's try one of these places:<br />
+    <a href="#">www.goodFood1.com</a>,<br />
+    <a href="#">www.goodFood2.com</a> or<br />
+    <a href="#">www.goodFood3.com</a>
+  </div>
+)
+
 const items = [
-  {
-    content: (
-      <Chat.Message content="Hello" author="John Doe" timestamp="Yesterday, 10:15 PM" mine />
-    ),
-  },
-  {
-    content: (
-      <Chat.Message
-        content="Hi"
-        author="Jane Doe"
-        timestamp="Yesterday, 10:15 PM"
-        avatar={janeAvatar}
-      />
-    ),
-  },
-  {
-    content: (
-      <Chat.Message
-        content="Would you like to grab a lunch?"
-        author="John Doe"
-        timestamp="Yesterday, 10:16 PM"
-        mine
-      />
-    ),
-  },
-  {
-    content: (
-      <Chat.Message
-        content="Sure! Let's try the new place downtown"
-        author="Jane Doe"
-        timestamp="Yesterday, 10:15 PM"
-        avatar={janeAvatar}
-      />
-    ),
-  },
-  {
-    content: <Divider content="Today" type="primary" important />,
-  },
-  {
-    content: (
-      <Chat.Message
-        content="Let's have a call"
-        author="John Doe"
-        timestamp="Today, 11:15 PM"
-        mine
-      />
-    ),
-  },
-]
+  <Chat.Message content="Hello" author="John Doe" timestamp="Yesterday, 10:15 PM" mine />,
+  <Chat.Message
+    content="Hi"
+    author="Jane Doe"
+    timestamp="Yesterday, 10:15 PM"
+    avatar={janeAvatar}
+  />,
+  <Chat.Message
+    content="Would you like to grab a lunch?"
+    author="John Doe"
+    timestamp="Yesterday, 10:16 PM"
+    mine
+  />,
+  <Chat.Message
+    content={{ content }}
+    author="Jane Doe"
+    timestamp="Yesterday, 10:15 PM"
+    avatar={janeAvatar}
+  />,
+  <Divider content="Today" type="primary" important />,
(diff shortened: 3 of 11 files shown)


docs/src/components/ComponentDoc/ComponentSidebar/ComponentSidebar.tsx
```diff
@@ -44,11 +44,11 @@ class ComponentSidebar extends React.Component<any, any> {
     return (
       <Sticky context={examplesRef} offset={15}>
         <Menu as={Accordion} fluid style={sidebarStyle} text vertical>
-          {_.map(sections, ({ examples, sectionName }) => (
+          {_.map(sections, ({ examples, sectionName }, index) => (
             <ComponentSidebarSection
               activePath={activePath}
               examples={examples}
-              key={sectionName}
+              key={`${sectionName}-${index}`}
```
layershifter, line 51: Can we move this to a separate PR? It will be hard to find why this change was done in future.

docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx
```diff
@@ -6,56 +6,41 @@ const janeAvatar = {
...
+  <Chat.Message
+    content={{ content }}
+    author="Jane Doe"
+    timestamp="Yesterday, 10:15 PM"
+    avatar={janeAvatar}
+  />,
+  <Divider content="Today" type="primary" important />,
+  <Chat.Message content="Let's have a call" author="John Doe" timestamp="Today, 11:15 PM" mine />,
+].map((chatItemContent, index) => ({
+  key: `chat-item-${index}`,
```
layershifter: We should not use antipatterns in examples, [Index as a key is an anti-pattern](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318)

src/components/Chat/ChatMessage.tsx
```diff
@@ -198,29 +194,26 @@ class ChatMessage extends UIComponent<Extendable<ChatMessageProps>, any> {
...
-            }
-            main={contentElement}
-          />
-        }
-        end={mine && avatarElement}
-      />
+      <>
+        {!mine && avatarElement}
+        <Slot
+          key="chat-message-body"
```
layershifter: As I remember, we don't need keys inside fragments

CHANGELOG.md
```diff
@@ -18,7 +18,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ## [Unreleased]
 
 ### Features
-- Add `renderComponent` function in the public API @mnajdova ([#503](https://github.com/stardust-ui/react/pull/503))
+- Add `createComponent` function in the public API @mnajdova ([#503](https://github.com/stardust-ui/react/pull/503))
```
bmdalex, line 21: @mnajdova fyi

docs/src/components/ComponentDoc/ComponentSidebar/ComponentSidebar.tsx
bmdalex: @layershifter  there was a console error that about rendering elements with the same key using array/iterator

screenshot 2018-11-22 at 17 06 51

I cannot open another PR as I have 2 already.. wouldn't this comment here be enough? I can amend the commit comment as well if necessary

docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx
bmdalex: @layershifter  proposal?
- we cannot import other libraries in examples in order to generate keys
- we could introduce the keys manually for each item but that'd be sad
- we could have a global counter (to this file) to hold the seed for our key generation and just increment it.

src/components/Chat/ChatMessage.tsx
bmdalex: indeed, this was a leftover from when I was fixing the console errors, I thought it's coming from here; thanks, good catch

src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -16,17 +17,24 @@ const chatMessageStyles: ComponentSlotStylesInput<ChatMessageProps, ChatMessageV
...
+    flex: 'none',
     display: p.mine ? 'none' : undefined,
     marginTop: px10asRem,
     marginBottom: px10asRem,
     marginLeft: p.mine ? px10asRem : 0,
     marginRight: p.mine ? 0 : px10asRem,
   }),
 
-  content: ({ props: p, variables: v }): ICSSInJSStyle => ({
+  messageBody: ({ props: p, variables: v }): ICSSInJSStyle => ({
```
bmdalex, line 37: when we were using `Layout`, we had `display: 'grid'` here; now we have `display: block` by default since this is a div; **this adds extra 5px to the height, even though the sum of the height of the children is the same**, and all paddings and margins are the same; any idea on what could cause this?

it's like this:
```html
<div class="container"> <!-- height: 35px (15px+20px) if display: grid, 40px otherwise?? -->
  <div class="child">some content</div> <!-- height: 15px -->
  <div class="child">some content</div> <!-- height: 20px -->
</div>

File: docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx (hunk 2 above)
Comment (jurokapsiar) [reply docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx]: in reality, messages retrieved from a backend will have some kind if ID. can you just manually add an ID to the  example data?

Comment (layershifter) [reply docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx]: > we could introduce the keys manually for each item but that'd be sad

It's a normal way, my opinion that we should keep examples maximally simple.

```jsx
{
    content: (
      <Chat.Message
        content="Let's have a call"
        author="John Doe"
        timestamp="Today, 11:15 PM"
        mine
      />
    ),
    key: 'message-today-11-15',
  },
```

May be it will be not so short, but it will be more obvious.

CHANGELOG.md
mnajdova: Thanks, I made this change already in my PR, but let's merge it soon after whichever of these two PRs gets in first.

src/themes/teams/components/Chat/chatMessageStyles.ts
mnajdova: I'll take a look on this now..

src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -16,17 +17,24 @@ const chatMessageStyles: ComponentSlotStylesInput<ChatMessageProps, ChatMessageV
     maxWidth: v.messageWidth,
     wordBreak: 'break-word',
     wordWrap: 'break-word',
+    ':focus': {
+      outline: 'none',
+      '& .ui-chat__message__messageBody': {
+        outline: `.2rem solid ${siteVariables.brand}`,
+      },
```
sophieH29, line 24: to avoid writing classes manually, can we do smth like?

```js
':focus': {
   outline: 'none',

   messageBody: {
      outline: `.2rem solid ${siteVariables.brand}`,
   }
}
```

src/themes/teams/components/Chat/chatMessageStyles.ts
mnajdova: Plain HTML is not behaving like this, so it is definitely something in our styles. I tried adding just content to the message, and the height was as expected - the same with the child. However when I tried adding the just the timestamp, the height was bigger then the child. I would suggest inspecting the styles for the timestamp (lineHeight, margin.. anything that could affect the height). In the end, we can try setting some additional styling to the div to fit the content...

docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx
```diff
@@ -6,56 +6,41 @@ const janeAvatar = {
...
-      <Chat.Message
-        content="Let's have a call"
-        author="John Doe"
-        timestamp="Today, 11:15 PM"
-        mine
-      />
-    ),
-  },
-]
+  <Chat.Message content="Hello" author="John Doe" timestamp="Yesterday, 10:15 PM" mine />,
```
mnajdova: Any particular reason why are we changing the existing examples? It would be hard to notice if there are any styling regression in screener this way. I know you already found one, but it would be much easier to catch all of theme on screener. You can add additional examples, if you want to advertise some different usage.

jurokapsiar: as far as I can see the data is the same, it's just the API change

src/themes/teams/components/Chat/chatMessageStyles.ts
bmdalex: why would this work, there is no style / DOM element named `messageBody`. am I missing something, @sophieH29 ?

docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx
bmdalex: reverted and added key to each item, thanks

src/themes/teams/components/Chat/chatMessageStyles.ts
bmdalex: I'll play with these, thanks @mnajdova I'll keep u posted

docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx
bmdalex: done, thanks guys

docs/src/components/ComponentDoc/ComponentSidebar/ComponentSidebar.tsx
bmdalex: @layershifter any resolution or can I resolve this?

src/themes/teams/components/Chat/chatMessageStyles.ts
bmdalex: I looked in the redlines and current behavior looks closer to that; the redlines are incomplete anyways so I'll leave it as is

layershifter (approved):
I don't have any other comments, LGTM 👍

src/themes/teams/components/Chat/chatMessageStyles.ts
sophieH29: I meant to use already defined class variable ```messageBody```  that you have here ```className={cx('ui-chat__message__messageBody', classes.messageBody)}```
Wasn't sure if it is possible, that's why I asked :)

docs/src/examples/components/Chat/Types/ChatMessageExampleStyled.shorthand.tsx
```diff
@@ -0,0 +1,92 @@
...
+    theme={{
+      componentStyles: {
+        ChatMessage: {
+          root: { ...slotLabelStyles('root'), backgroundColor: '#2E8B57' },
+          avatar: {
+            ...slotLabelStyles('avatar', { bottom: '-11px' }),
+            backgroundColor: '#FF00FF',
+            padding: 0,
+          },
+          messageBody: { ...slotLabelStyles('messageBody'), backgroundColor: '#87CEFA' },
```
kuzhelov, line 48: would be great to generalize this functionality and export under 'debug' prop to the client (or via any other generalized approach), in the same way it is done for Layout - that will reduce amount of code necessary in examples

src/components/Chat/ChatMessage.tsx
```diff
@@ -104,10 +99,10 @@ class ChatMessage extends UIComponent<Extendable<ChatMessageProps>, any> {
   static propTypes = {
     ...commonUIComponentPropTypes,
     ...childrenComponentPropTypes,
-    ...contentComponentPropsTypes,
     accessibility: PropTypes.oneOfType([PropTypes.object, PropTypes.func]),
     author: customPropTypes.itemShorthand,
     avatar: customPropTypes.itemShorthand,
+    content: customPropTypes.itemShorthand,
```
kuzhelov, line 105: just to be sure - this is done because of `content` prop type (`contentShorthand`) that doesn't fit here?

src/components/Chat/ChatMessage.tsx
```diff
@@ -198,29 +193,23 @@ class ChatMessage extends UIComponent<Extendable<ChatMessageProps>, any> {
     })
 
     const contentElement = Slot.create(content, {
-      styles: styles.content,
-      variables: variables.content,
+      defaultProps: {
```
kuzhelov: 👍

src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -16,17 +17,24 @@ const chatMessageStyles: ComponentSlotStylesInput<ChatMessageProps, ChatMessageV
     maxWidth: v.messageWidth,
     wordBreak: 'break-word',
     wordWrap: 'break-word',
+    ':focus': {
+      outline: 'none',
+      messageBody: {
```
kuzhelov: this won't work - `messageBody` styles block won't be applied on root focus

kuzhelov (changes requested):
please, take a look on the comment related to message body's styles - other than that everything looks fine to me

docs/src/components/ComponentDoc/ComponentSidebar/ComponentSidebar.tsx
bmdalex: will keep the change in this PR

src/themes/teams/components/Chat/chatMessageStyles.ts
bmdalex: we cannot, CSS in JS styles cannot be  aware of CSS classes **they** generate;
`messageBody` class is generated from the styles in `messageBody` lot in `src/themes/teams/components/Chat/chatMessageStyles.ts` so it cannot be referenced yet; 

moreover, we don't have access to class names anyway, you we're referring to `messageBody ` as the value of the class name itself in `src/themes/teams/components/Chat/chatMessageStyles.ts` but it's just the name of the variable storing the class name (the value is something generated by Fella).

hope it makes sense

docs/src/examples/components/Chat/Types/ChatMessageExampleStyled.shorthand.tsx
bmdalex: good proposals, let's discuss

src/components/Chat/ChatMessage.tsx
bmdalex: yes

src/themes/teams/components/Chat/chatMessageStyles.ts
bmdalex: I committed that by mistake, reverted but you already saw it haha

docs/src/examples/components/Chat/Types/ChatMessageExampleStyled.shorthand.tsx
bmdalex: discussed with @kuzhelov and decided to leave this in examples so that users can understand how we achieved the custom styling

kuzhelov (approved):

Issue #355, opened 2018-10-11
Alternative to 'renderX' set of props for slots


kuzhelov:
Let this issue serve as a discussion hub for [alternative approach](#proposed-solution) to `renderXXX` one that is currently used for Stardust components. 

**TL;DR** Here is the list of main reasons for that:
- downsides of `renderXXX`
  - necessity to [bloat API surface](#api-bloat) with `renderXXX` methods for each slot of each component
  - general incorrectness of shorthand evaluation for [async scenarios](#problem-async)
  - it is easy to for the client to [accidentally use this API improperly](#problem-do-it-yourself)
  - [points of confusion](#problem-confusing-api) related to API
- suggested alternative approach
  - is able to cover all the cases covered by the original one
  - introduces [safe API](#problem-do-it-yourself) that could be [easily explained to user](#explain-concept-to-consumer)
  - is able to [correctly handle async shorthand rendering cases](#problem-async)

Please, note that scenario-based [comparison examples](#use-case-examples) for both approaches are also provided.

--------

# Contents

- [Motivation](#motivation)
- [Proposed Solution Overview](#proposed-solution)
  - [Idea](#idea)
  - [Example](#example)
  - [Benefits](#benefits)
- [Scenario-based comparison](#use-case-examples)
  - [Explain Concept](#explain-concept-to-consumer)
  - [Async Rendering](#async-rendering)
  - [Custom Tree Rendering](#custom-tree-rendering)
  - [Selectively Apply Custom Rendering](#selectively-apply-custom-rendering)

<h1 id='motivation'>Motivation</h1>
There are several problems with current `renderItem` prop - let me mention them.

<h2 id="problem-do-it-yourself"> 1. "Do It Yourself" pattern provided by `renderXXX` to compose component</h2>

Lets consider the API we now have for `renderItem`:
```jsx
/**
* Its client responsibility now to not forget about:
* - consuming props as a second argument
* - spread these props to the component
* 
* Failure to do these steps will result in the problems 
* that could be hard to immediately detect
* - broken/incorrect styles
* - broken/incorrect component behavior/accessibility behavior
**/
<Menu items={items} renderItem={(Component, props) => {  // <== don't forget 'props' !
    ... 
     return <Component {...props} />    // <===== oh, don't forget to expand 'props'!
}/>
```
As we might see, there are quite a lot of things that client should worry about now - thus, this approach is quite error-prone.

[...]

Issue #512, opened 2018-11-22
RFC: Shorthand polymorphism using `kind` property 


miroslavstastny:
### Problem description
Shorthand is a powerful way how you can define component slots:
```jsx
<Button icon={{name: "user", size: "micro"}} content="Click me!" />
```
In the example above, both `icon` and `content` are shorthands. It allows you to define as few properties as possible while still giving you full control over the slot props. And in all cases shorthand still manages state, styling and accessibility for you.

Shorthand becomes extremely useful when defining collection of children, for example in `ButtonGroup`:
```jsx
<Button.Group
  buttons={[
    { key: 'book', icon: 'book', iconOnly: true },
    { key: 'coffee', icon: 'coffee', iconOnly: true },
    { key: 'play', icon: 'play', iconOnly: true },
  ]}
/>
```

But what if you would like to add different component to the collection of buttons? What if you want to add a divider?

### `kind`
For that reason we plan to add `kind` property to the shorthand definition. With that, you can choose which component will be created using the shorthand:
```jsx
<Button.Group
  buttons={[
    { key: 'book', icon: 'book', iconOnly: true },
    { key: 'coffee', icon: 'coffee', iconOnly: true },
    { kind: 'divider' },
    { key: 'play', icon: 'play', iconOnly: true },
  ]}
/>
```

### Difference between `as` and `kind`
All Startdust components already support `as` property which allows you to override element type in the rendered result. So what is the reason for introducing `kind`? What is the difference between the two?  

Stardust component defines resulting DOM structure, state, styles and accessibility. With `as`, you are overriding `ElementType` - that is usually the root element in the resulting DOM structure.
But the original component is still created - with the structure, state, styles and accessibility of the original component.

On the other hand when `kind` is used, different component is created - with its own structure, state, styles and accessibility.

**To summarize** - `kind` defines what component is created. `as` defines `ElementType` of the created component.

It is technically possible to combine the two:
```jsx
<Button.Group
  buttons={[
    ...
    { kind: 'divider', as: MyCustomDivider },
    ...
  ]}
/>
```

### Component implementation
There is no magic happening in factory, it is component's responsibility to handle `kind` attribute.

Component should allow the `kind` to be only one of predefined options (`ButtonGroup` should allow only `'button'` or `'divider'` and default to `'button'` if no kind is specified and nothing else).

### Possible use cases considered:
- ButtonGroup - (button), divider
- Form - field
- FormGroup - field
[...]

PR #519, opened 2018-11-23
feat: 'render' callback for shorthand props

kuzhelov:
Implements #355. 

## TODO
- [x] add/adjust unit tests
- [x] remove render methods from component's code
- [x] update CHANGELOG

-------------------

Provided changes introduce additional ways to specify shorthand's value using `render` callback. Following examples are aimed to deliver key points.

### Base case

```jsx
<Button icon="book" />
<Button icon={{ name: 'book' }} />
```
### Render primitive value
```jsx
<Button icon={render => render('book')} />
```

### Render object value
```jsx
<Button icon={render => render({ name: 'book' }) } />
```

### Render React element (no evaluated props consumed)
```jsx
<Button icon={render => render(<span>▼▲</span>) } />
```

### Render custom tree with **evaluated** shorthand props
Note, this is a functionality we've previously had with `render` methods.

```jsx
<Button icon={render => render(
   'book',

  // where 'props' contains all evaluated ones from { name: 'book' }
  (Component, props) => <Component {...props} />)} 
/>
```

OR

```jsx

<Button icon={render => render(
   { name: 'book' },

   // where 'props' contains all evaluated ones from { name: 'book' }
   (Component, props) => <Component {...props} />) } 
/>
```


types/utils.d.ts
```diff
@@ -30,7 +30,21 @@ export type ComponentEventHandler<TProps> = (event: React.SyntheticEvent, data:
...
+  value: ShorthandValue | ShorthandRenderTreeFunc,
+  renderTree?: ShorthandRenderTreeFunc,
+) => React.ReactElement<any>
+
+export type ShorthandRenderer = (render: ShorthandRenderCallback) => React.ReactElement<any>
+
+export type ShorthandValue = React.ReactNode | Props | ShorthandRenderer
+
+// OBSOLETE
 export type ShorthandRenderFunction = (
```
kuzhelov: should be removed

src/lib/factories.tsx
```diff
@@ -45,6 +52,34 @@ export function createShorthand(
...
 ): React.ReactElement<Props> | null | undefined {
+  const valIsRenderFunction = typeof value === 'function' && !React.isValidElement(value)
+  if (valIsRenderFunction) {
+    const render: ShorthandRenderCallback = (shorthandValueOrRenderTree, renderTreeArg) => {
+      const shorthandArgType = {
+        isReactElement: React.isValidElement(shorthandValueOrRenderTree),
+        isRenderTreeFunc: typeof shorthandValueOrRenderTree === 'function',
+      }
+
+      const ShorthandElement = createShorthand(
```
mnajdova: Would it be a good idea to extract the rest of the logic in createShorthand in a separate method that we would called here, and maybe after this `if` statement is finished, just so that we don't make recursion here? If somebody for some reason change the condition inside the if statement we may introduce infinite recursion...

src/lib/factories.tsx
```diff
@@ -45,6 +52,34 @@ export function createShorthand(
...
+        Component,
+        mappedProp,
+        shorthandArgType.isRenderTreeFunc ? {} : (shorthandValueOrRenderTree as ShorthandValue),
+        options,
+      )
+
+      const renderTree =
+        (shorthandArgType.isRenderTreeFunc &&
+          (shorthandValueOrRenderTree as ShorthandRenderTreeFunc)) ||
+        renderTreeArg
```
mnajdova: In my opinion is difficult to reason what will the value of `renderTree` be in this case. Maybe we can initialize it with `renderTreeArg` and then conditionally change it to `shorthandValueOrRenderTree`? Not saying this is must but it would be helpful at least for me :)

mnajdova (commented):
I tested it locally and it works as expected. Left two comments for consideration. I vote that we start applying the changes to all components.

src/lib/factories.tsx
kuzhelov: I do understand your concerns here - let me try to play with that a bit, just the only criteria I am striving for is that final solution would be readable - and, actually, assured, it would be :) Thank you!

src/lib/factories.tsx
kuzhelov: agree, now it is not-easy-to-grasp-the-intent expression, will rework, thanks!

src/lib/factories.tsx
kuzhelov: have played with that - won't like to lose `const` in declaration (suggested option, unfortunately, will inevitably lead to `let` being used), but it seems that reasonable compromise is introduced :)

src/lib/factories.tsx
kuzhelov: have tried that - actually, won't like to lose const in the variable declaration (unfortunately, suggested approach will inevitably lead to that, as `let` will be used), but it seems that reasonable compromise is suggested now :)

mnajdova: After refactoring is much better actually. No need for additional changes in my opinion.

layershifter:
I want to mention that we have a case where `render*` is passed down to other components and this can be really useful for nested items like submenus, #512.

It's a real case:

image

```jsx
<Tree items={items} renderItemTitle={titleRenderer} />
```

----

How we will get the same behaviour after this PR?

kuzhelov:
@layershifter,

we are not losing anything with this change - for those components where this `render` functionality will be useful - here I am getting `Tree` case you've mentioned as an example, we will be able to use this value passed to `renderItemTitle` prop and pass it to the corresponding shorthand. Key difference is that we will have these `render` props only for those components' API where it is absolutely necessary (little percent of the whole number), and won't pollute the API surface for the others.

src/components/Accordion/Accordion.tsx
```diff
@@ -40,26 +35,6 @@ export interface AccordionProps extends UIComponentProps<any, any>, ChildrenComp
...
 
-  /**
-   * A custom render iterator for rendering each Accordion panel content.
-   * The default component, props, and children are available for each panel content.
-   *
-   * @param {React.ReactType} Component - The computed component for this slot.
-   * @param {object} props - The computed props for this slot.
-   * @param {ReactNode|ReactNodeArray} children - The computed children for this slot.
-   */
-  renderContent?: ShorthandRenderFunction
```
mnajdova, line 56: In this case, I find this methods useful. They are not callback render for a shorthand, but for each item of array. Do we want to get rid of this as well? It would be nice from user perspective for the array shorthands, to be able to define this logic via an API property. What are your thoughts on this @kuzhelov ?

src/components/Attachment/Attachment.tsx
```diff
@@ -28,51 +28,6 @@ export interface AttachmentProps extends UIComponentProps<any, any>, ChildrenCom
...
-  renderIcon?: ShorthandRenderFunction
-
-  /**
-   * A custom render function the progress slot.
-   *
-   * @param {React.ReactType} Component - The computed component for this slot.
-   * @param {object} props - The computed props for this slot.
-   * @param {ReactNode|ReactNodeArray} children - The computed children for this slot.
-   */
-  renderProgress?: ShorthandRenderFunction
```
mnajdova, line 80: 👍 much cleaner API!

src/components/Menu/Menu.tsx
```diff
@@ -46,16 +46,6 @@ export interface MenuProps extends UIComponentProps<any, any>, ChildrenComponent
...
 
-  /**
-   * A custom render iterator for rendering each of the Menu items.
-   * The default component, props, and children are available for each item.
-   *
-   * @param {React.ReactType} Component - The computed component for this slot.
-   * @param {object} props - The computed props for this slot.
-   * @param {ReactNode|ReactNodeArray} children - The computed children for this slot.
-   */
-  renderItem?: ShorthandRenderFunction
```
mnajdova, line 62: Same as the Accordion, I know that the user can use the current API and mapping logic for introducing this, but notice that this was not added with the renderXXX methods PR, it was here before, because it introduces clearer API for rendering each item in the Menu.

mnajdova (commented):
Let's decide whether we want to keep in the API the render method for the arrays of shorthands available in some components like Accordion, Menu, Tree etc. Other than that, the changes looks good!

src/components/Accordion/Accordion.tsx
kuzhelov: #### Solution with no additional API changes

```jsx
// this could be a helper util
const customTreeShorthand = (value, renderTree) => (render => render(value, renderTree))

const renderCustomMenuItem = (Component, props) => <Component {...props} />

<Menu items={[ 
    { as: ..., ... },
    ...
  ]
 .map(shorthandValue => customTreeShorthand(shorthandValue, renderCustomMenuItem))
} 
.. />
```

#### Solution with `render` method introduced for limited set of components - specifically, collection and hierarchical ones: Menu, Tree, Accordion, List (?)

Needs changes, but we may talk about the final result even now:

```jsx
<Menu items={[ 
    { as: ..., ... },
    ...
  ]}
  renderItem={(Component, props) => ... }  { /* is treated a default rendering logic, if not overwritten by shorthand */ }
.. />
```

What I don't like here is that it is not absolutely clear for the client where `Component` and `props` have come from, and where the data that is provided as shorthand object (i.e. `{ as: ..., ...}` resides in those `Component` and `props`).

mnajdova: I am okay with both approaches, but when we have shorthand array, with two shorthands inside, as the panels are inside the Accordion, the user will have to define two maps for both shorthand, which may look awkward... Let's hear other folks' opinion before deciding.

docs/src/examples/components/Avatar/Variations/AvatarExampleImageCustomization.shorthand.tsx
```diff
@@ -9,9 +9,9 @@ const AvatarExampleImageCustomizationShorthand = () => (
     />
     &emsp;
     <Avatar
-      renderImage={(Image, props, children) => (
+      image={
         <Icon name="user" circular variables={{ color: 'blue' }} styles={{ padding: '8px' }} />
-      )}
+      }
```
mnajdova, line 14: 👍

mnajdova (approved):
After offline discussion, we decided to add explicit render method just for the Tree component, the other ones can easily be defined by the user. Let's merge this after that is added.

kuzhelov:
@mnajdova, this was already added, and I've not touched this method: https://github.com/stardust-ui/react/blob/master/src/components/Tree/Tree.tsx#L35 . It works as before :)

PR #521, opened 2018-11-23
Fix(Popup): Trigger's props are not passed in onClick handler

sophieH29:
This PR addresses fix for a small issued encountered in Popup

Let's say we have a Popup, with Button component as a trigger

```js
 <Popup
        // you expect on Button's click receive 2 arguments, but recieve only event object
       // props are undefined
        trigger={<Button onClick={(e, props) =>  {}} />}
        content={{ content: <Some content> }}
      />
```

For Button's ```onClick``` you expect to receive 2 arguments, ```event``` and ```props```, but in fact, receive only ```event``` object as popup's handler cut off this from click event pipeline.

Even bigger issue with that can be found, when use ```Popup``` as ```MenuItem```

```js
  <Menu>
     <Menu.Item>
     <Menu.Item>
     <Popup><Menu.Item></Popup>
  </Menu>
```

When click on ```MenuItem```, event firstly propagates to Popup which cut off essential props, then ```Menu``` handles that event and can't find props as arguments.


kuzhelov (approved):
👍 please, just update CHANGELOG before merging

jurokapsiar (approved):

PR #524, opened 2018-11-26
docs(Prototype): Chat messages with reactions popover prototype

sophieH29:
ezgif com-crop


docs/src/prototypes/chatMessageWithPopover/ChatMessageWithPopover.tsx
```diff
@@ -0,0 +1,56 @@
...
+const janeAvatar = {
+  image: 'public/images/avatar/small/ade.jpg',
+  status: { color: 'green', icon: 'check' },
+}
+
+class ChatMessageWithPopover extends React.Component {
+  render() {
+    return (
+      <Chat.Message
+        styles={{
```
layershifter: We should apply `styles` with `Provider`: https://github.com/stardust-ui/react/pull/406#issuecomment-441041119

docs/src/prototypes/chatMessageWithPopover/ChatMessageWithPopover.tsx
```diff
@@ -0,0 +1,56 @@
...
+          content: (
+            <div>
+              <Popover className="actions" />
+              <a href="/">Link</a> Hover me to see the actions <a href="/">Some Link</a>
+            </div>
+          ),
+        }}
+        avatar={janeAvatar}
+        onFocus={handleFocus}
+        onBlur={handleBlur}
```
layershifter: Why we're using event handlers from a separate file? It looks weird.

---

Can we use them there as class methods and just pass down to `Popover` if there is a real need?

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
```diff
@@ -0,0 +1,118 @@
...
+  const currentTarget = e.currentTarget
+  const relatedTarget = e.relatedTarget
+
+  if (currentTarget.contains(relatedTarget)) {
+    if (!currentTarget.classList.contains('focused')) {
+      currentTarget.classList.add('focused')
+    }
+  } else {
+    currentTarget.classList.remove('focused')
+  }
```
layershifter: We should not manipulate DOM directly, we can introduce a state variable that will define applied props or classes in `render()`

docs/src/prototypes/chatMessageWithPopover/ChatMessageWithPopover.tsx
sophieH29: I would rather use just ```style``` for that

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
sophieH29: We can treat these methods ```handleBlur``` and ```handleFocus``` as additional lib (can move them to separate file), which defines if element got focus or not, by adding an additional class to target. These methods work both equally for message container and popover itself.
Also  we avoid re-rendering component on each state change (focus/blur events can be triggered many-many times)

docs/src/prototypes/chatMessageWithPopover/ChatMessageWithPopover.tsx
bmdalex: @sophieH29 
`style` props:
- are just going to be inlined directly to the DOM;
- are impossible to override;
- they do not go through the Fella Renderer;
- they are not reusable through classes so abusing them would create huge styling objects (imagine if you have hundreds of `Chat.Message` components, each with their own ` style` object)
etc.

We should avoid using `style` as much as we can.

docs/src/prototypes/chatMessageWithPopover/ChatMessageWithPopover.tsx
```diff
@@ -0,0 +1,56 @@
...
+      />
+    )
+  }
+}
+
+const CustomChat = () => (
```
bmdalex: ~~`CustomChat`~~  -> `ChatWithPopover` ?

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  return (
+    <Popup
+      key={props.key}
+      position="below"
+      accessibility={popupFocusTrapBehavior}
+      trigger={<MenuItem {...props} />}
+      content={
+        <div
```
bmdalex: @kuzhelov @mnajdova @layershifter @miroslavstastny 
should we reopen the discussion for exporting a blank component for styling (our `Slot` -> `Component`)?

@sophieH29 could use `createComponent` but it's a bit odd to tell users to create a blank component just so they can style things or use it as wrapper / container for other components

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
bmdalex: +1
why not use `:focus` pseudoselector instead of this `focused` class?

docs/src/prototypes/chatMessageWithPopover/ChatMessageWithPopover.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+class ChatMessageWithPopover extends React.Component {
+  render() {
+    return (
+      <Chat.Message
+        styles={{
+          position: 'relative',
+
+          '&.focused .actions': {
```
bmdalex: why not use `:focus` instead?

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
```diff
@@ -0,0 +1,118 @@
...
+          background: '#fff',
+          boxShadow: '0px 2px 4px #ddd',
+          borderRadius: '.3rem',
+          opacity: 0,
+
+          '& .smile-emoji': {
+            display: 'none',
+          },
+
+          '&.focused .smile-emoji': {
```
bmdalex: why not use `:focus` instead?

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
```diff
@@ -0,0 +1,118 @@
...
+          />
+        </div>
+      }
+    />
+  )
+}
+
+export const handleFocus = e => {
+  const currentTarget = e.currentTarget
+  if (!currentTarget.classList.contains('focused')) {
```
bmdalex: why not use `:focus` instead?

bmdalex (changes requested):

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
layershifter: 1. We're going to drop `styles` prop at all, https://github.com/stardust-ui/react/pull/406#issuecomment-441041119
2. What is a benefit of usage `Slot` in this case?

docs/src/prototypes/chatMessageWithPopover/ChatMessageWithPopover.tsx
sophieH29: because if the focus is on the child element, like popover's menu item, we should still show it and keep this "focused" state for the whole container.

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
sophieH29: because if the focus is on the child element, like popover's menu item, we should still show it and keep this "focused" state.
Anyway, going to move it to state based on @layershifter comment

docs/src/prototypes/chatMessageWithPopover/ChatMessageWithPopover.tsx
sophieH29: ok, but please, @Bugaa92, give me an example of a proper way for reaching the same goal.
btw, if look at different prototypes, many of them are using different ways of styling, like ```style``` or ```styles``` or ```Provider and styles```, so what is a proper way?

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
sophieH29: sorry, but don't get it, why should I use ```createComponent```

docs/src/prototypes/chatMessageWithPopover/ChatMessageWithPopover.tsx
sophieH29: will change that

docs/src/prototypes/chatMessageWithPopover/ChatMessageWithPopover.tsx
sophieH29: UPD: used Provider, please check the last changes if it's ok

bmdalex: you're still using `styles` and you're also using `Provider.Consumer` not `Provider`; please take a look at this example for reference:
https://github.com/stardust-ui/react/blob/eb583a7548a173e27145b087ac156009606c99df/docs/src/examples/components/Chat/Types/ChatMessageExampleStyled.shorthand.tsx#L37

the idea was to add a provider above your custom `Chat` component:

```jsx
// this doesn't have to be a class
const ChatMessageWithPopover = () => (
  <Chat.Message
    author="Jane Doe"
    content={{
      content: (
        <div>
          <Popover className="actions" />
          <a href="/">Link</a> Hover me to see the actions <a href="/">Some Link</a>
        </div>
      )
    }}
    avatar={janeAvatar}
    onFocus={handleFocus}
    onBlur={handleBlur}
  />
)

const ChatMessageExampleStyled = () => (
  <Provider
    theme={{
      componentStyles: {
        ChatMessage: {
          root: { /* your styles go here */ },
        },
      },
    }}
  >
    <Chat
      items={[
        { key: 'a', content: <ChatMessageWithPopover /> },
        { key: 'b', content: <ChatMessageWithPopover /> },
        { key: 'c', content: <ChatMessageWithPopover /> },
      ]}
    />
  </Provider>
)
```

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
bmdalex: ah I read the code more carefully, got it; too bad IE and Edge don't have support for [`:focus-within`](https://css-tricks.com/almanac/selectors/f/focus-within/)

docs/src/prototypes/chatMessageWithPopover/ChatMessageWithPopover.tsx
bmdalex: ah I read the code more carefully, got it; too bad IE and Edge don't have support for [`:focus-within`](https://css-tricks.com/almanac/selectors/f/focus-within/)

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
bmdalex: > We're going to drop styles prop at all, #406 (comment)
oh, right..

> What is a benefit of usage Slot in this case?
I just realized we're going to drop styles so I am not sure right now.

@layershifter @sophieH29 disregard my comment, pls and leave code as is; I'll close this conversation and open a new comment to express my query

bmdalex: @sophieH29 needs to use a `div` primitive and inline `style` in order to wrap a `Menu` for the content of a `Popup`; from our perspective, **are we ok with that?**

Shouldn't our users have a way to custom one-off things like this with a (surpris,e surprise) custom component that they can style and will go through the fella renderer too in order to provide all Stardust capabilities (accessibility, rtl, computed styles, etc)?

I'm thinking `createComponent` should be used here, but how? I was trying to prototype this and doesn't look smooth at all...

This example should become something like:

```jsx
const MyMenuWithWrapper = createComponent({
  displayName: 'MenuWrapper',

  render({ stardust }) {
    const { classes, rtl } = stardust // how do we pass style object to custom component?
    return <div className={classes}><Menu .../></div> // we still use `div`?
  },
})
```
and then use `MyMenuWithWrapper` for `Popup` `content`

can you guys weigh in here?

@layershifter @kuzhelov @mnajdova @miroslavstastny

bmdalex (commented):
responded, pls take another look

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
mnajdova: If you add in the props for the `MyMenuWithWrapper` styles and variables, those should work exactly the same way as they work for any first-class Stardust component. You can target MenuWrapper in the provider (by the `displayName`) or use `styles={{/* your custom styles*/}}` and `variables={{/* your custom variables */}}` on component level. BTW I would recommend using the same name for the Component name and the `displayName` arg to `createComponent`. This is the PR that contains examples for this: https://github.com/stardust-ui/react/pull/517 and is already merged, so on master you should be able to take a look into the Integrate Custom components page. Let me know if you need more input on this.

sophieH29: @mnajdova, I see that there is a way to rewrite it using ```createComponent```.
I was making this prototype actually as a user of Stardust, I combined all the needed components and received the result which I expected. 
So as a user, I have no clue why should I rewrite it using ```createComponent``` :) 
And I see the rest of the prototypes are made in the same manner (without ```createComponent```), I am kind of confused. I read in your responses "what" and "how" to do it, but I didn't read "why"

bmdalex: > And I see the rest of the prototypes are made in the same manner (without createComponent), I am kind of confused

`createComponent` was just released yesterday, the rest of the prototypes were created before that;

> I read in your responses "what" and "how" to do it, but I didn't read "why"

The why is in my message above:
https://github.com/stardust-ui/react/pull/524#discussion_r236722003

_Shouldn't our users have a way to custom one-off things like this with a (surpris,e surprise) custom component that they can style and will go through the fella renderer too in order to provide all Stardust capabilities (accessibility, rtl, computed styles, etc)?_

Also, Imagine you have some components passing props to the `Menu` and you're adding a primitive `div` in between and break the chain.

These are not that valid for this specific usecase but prototypes should be our suggestions for users to design their code.

@sophieH29 We are not asking you to take this approach immediately; we don't expect that since we just released `createComponent`.

However, user feedback is crucial and as you said, you're one of our users. We're just trying to understand how they will feel about the new API so we can improve it.

sophieH29: frankly, this was a proper "why" answer for me _These are not that valid for this specific usecase but prototypes should be our suggestions for users to design their code_  :)
why the ```createComponent``` exists I understand, the thing was about the reason it should be applied for this specific use case.
Anyway thank you for your responses, guys, really appreciate that.
I have a few more things I want to ask/clarify/ and maybe need your assistance. But probably let's discuss it tomorrow offline, as don't wanna take more of your time on replying.
Again, thank you!

bmdalex (approved):
thanks for addressing

docs/src/prototypes/chatMessageWithPopover/ChatMessageWithPopover.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+import * as React from 'react'
+import Popover from './Popover'
+
+const janeAvatar = {
+  image: 'public/images/avatar/small/ade.jpg',
+  status: { color: 'green', icon: 'check' },
+}
+
+class ChatMessageWithPopover extends React.Component {
```
mnajdova: Split this into two files - `ChatMessageWithPopover` and `ChatWithPopover`. It's confusing now that the name of the file is ChatMessageWithPopuover and we are exporting as default ChatWithPopover. Let's keep consistency for having one class per file.

docs/src/prototypes/chatMessageWithPopover/ChatMessageWithPopover.tsx
```diff
@@ -0,0 +1,102 @@
...
+            <div>
+              <Popover className="actions" />
+              <a href="/">Link</a> Hover me to see the actions <a href="/">Some Link</a>
+            </div>
+          ),
+        }}
+        avatar={janeAvatar}
+        onFocus={this.handleFocus}
+        onBlur={this.handleBlur}
+        className={this.state.focused ? 'focused' : undefined}
```
mnajdova: Please add `className` property to the API of this component. With the current implementation we are not respecting the user's input for this property. More over, the user currently has no way of styling it at all, except by using the `style` prop, which is not something we want to advertise! That way you can refactor the the className value in the following way: `className={cx(this.props.className, this.state.focused ? 'focused' : '')}`. More over, this is also a good example for using the createComponent paradigm, as in that way you would allow the user to use the ChatMessageWithProvider exactly the same way as they would use the ChatMessage.

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
```diff
@@ -0,0 +1,183 @@
...
+  changeFocusState = (isFocused: boolean) => {
+    this.state.focused !== isFocused && this.setState({ focused: isFocused })
+  }
+
+  handleFocus = () => {
+    this.changeFocusState(true)
+  }
+
+  handleBlur = e => {
+    // if e.relatedTarget === null, so the click was outside this container
```
mnajdova:
```
//if e.relatedTarget === null it means that the click was outside this container
```

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
```diff
@@ -0,0 +1,183 @@
...
+      e.stopPropagation()
+    }
+  }
+
+  handleMenuClick = () => {
+    // close popup when other MenuItem clicked, but the event propagation was stopped
+    this.state.popupOpened && this.setState({ popupOpened: false })
+  }
+
+  menuStyles = ({ theme: { siteVariables } }) => ({
```
mnajdova: In my opinion all styles shuold be separate in different file then the definition of the component using them.

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
```diff
@@ -0,0 +1,183 @@
...
+    },
+  })
+
+  render() {
+    return (
+      <Menu
+        styles={this.menuStyles}
+        iconOnly
+        className={cx(this.props.className, this.state.focused ? 'focused' : '')}
+        items={[
```
mnajdova: 👍 I see now that you have already implemented this component to respect user's input on the className. See me previous comment for refactoring the other component in the same way.

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
```diff
@@ -0,0 +1,183 @@
...
+      display: 'flex',
+    },
+  })
+
+  render() {
+    return (
+      <Menu
+        styles={this.menuStyles}
+        iconOnly
+        className={cx(this.props.className, this.state.focused ? 'focused' : '')}
```
mnajdova, line 81: 👍 I see now that you have already implemented this component to respect user's input on the className. See my previous comment for refactoring the other component in the same manner.

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
```diff
@@ -0,0 +1,183 @@
...
+            'aria-label': 'thumbs up',
+          },
+          {
+            key: 'c',
+            icon: 'ellipsis horizontal',
+            accessibility: toolbarButtonBehavior,
+            'aria-label': 'more options',
+          },
+        ]}
+        renderItem={this.renderItemOrContextMenu}
```
mnajdova: Watch out, we merge the PR for removing the renderXXX methods from the component's API. This should be refactored. For more info, please see the examples on the PR https://github.com/stardust-ui/react/pull/519 and the Issue https://github.com/stardust-ui/react/issues/355

mnajdova (commented):
Please see the comments provided.

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
sophieH29: Yeah, I just found out it yesterday when Popup hasn't render. Will take a look today how to solve it

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
sophieH29: ok, will move that

sophieH29: agree that this might be better when having styles separately so the code will look cleaner. But I would leave exactly these styles where they are, so it be more understandable when the ```focused``` state is changed it has direct affect in styles

jurokapsiar (approved):

mnajdova (approved):
I am approving these changes, but before you merge, please take a look into one issue that I found (or at least I expected different behavior), while testing locally. This are the steps:
- hover on the message (the popover is open)
- hover on the popover (all 5 items were shown) 
- hover the three dots and click one of the items in the menu
The menu closed but the popup is still shown with all five things opened, until you click somewhere outside of the it.
Here is a gif showing this: https://i.imgur.com/u8PV7l3.gif

PR #527, opened 2018-11-27
chore(CI): add vulnerability tests (Snyk)

kuzhelov:
This PR aims to introduce vulnerability checks for package dependencies used in Stardust.

Recent 'event-stream' package's case has served as main source of inspiration: https://snyk.io/blog/malicious-code-found-in-npm-package-event-stream

Tests are added as part of CI build, step itself takes 1.8 secs to finish. 
image

## TODO

~~Note that currently token from `kuzhelov` organization is used there.~~
- [x] we should consider to replace this SYNC_TOKEN on the Circle CI side. However, this fact won't prevent/restrict test checks anyhow for now
- [ ] after that we will be able to introduce 'vulnerabilities' badge to our README.md - as at that time it will lead to org's page


.circleci/config.yml
```diff
@@ -61,6 +61,9 @@ jobs:
             if [ ! -z $CIRCLE_PULL_REQUEST ] && [ ! -z $DANGER_GITHUB_API_TOKEN ]; then
               yarn danger ci
             fi
+      - run:
```
miroslavstastny: Note: In #97, we will introduce bundle size stats which must be run as "last" step in the build (as commiting new stats will cancel the build). That means that we will need to move Vulnerability Tests before Danger JS.

kuzhelov: yes - actually, as it is very quick step, I am already moving it as a step that will be done before visual tests

miroslavstastny (approved):

PR #528, opened 2018-11-27
fix(content): fixing the type of the content prop in all components

### LINKED ISSUE #504: Replace contentShorthand with itemShorthand
### Steps
We have a bug in our components `propTypes` for the `content` slot. Most of our components use the `Slot` component to generate their `content` slot, meaning that the `propTypes` for `content` prop should be:

### Expected Result

```typescript
content: customPropTypes.itemShorthand
```

### Actual Result

```typescript
content: customPropTypes.contentShorthand
```

### Solution
- go to all components and verify their `propTypes` for the slots are correct;
- for all component slots created with `Slot` component, replace `customPropTypes.contentShorthand` propType for that slot with `customPropTypes.itemShorthand`
- check in docs examples that warnings such as the one below don't appear anymore (for instance when providing an object as a slot value):

```jsx
<Button content={{ content: "Click here" }} />
```
will generate the following warning:
screenshot 2018-11-20 at 16 59 50

mnajdova:
This PR introduces the correct typings in the props' interface as well as in the propTypes in all components regarding the content property. For this purpose two different interfaces and commonPropTypes definition are introduced:

# 1. The content is used as a shorthand
```jsx
export interface ContentShorthandComponentProps {
  content?: ShorthandValue
}

export const contentShorthandComponentPropsTypes = {
  content: customPropTypes.itemShorthand,
}
````

# 2. The content is used as a ReactNode
```jsx
export interface ContentNodeComponentProps {
  content?: React.ReactNode
}

export const contentNodeComponentPropsTypes = {
  content: customPropTypes.contentShorthand,
}
```

In addition, the customPropTypes.itemShorthand was changed in the following manner:
## Before
```jsx
export const itemShorthand = every([
  disallow(['children']),
  PropTypes.oneOfType([
    PropTypes.node,
    PropTypes.object,
    PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.node, PropTypes.object])),
  ]),
])
```
## After
```jsx
export const itemShorthand = every([
  disallow(['children']),
  PropTypes.oneOfType([
    PropTypes.node,
    PropTypes.object,
  ]),
])

```
With this change, it corresponds to the TS definition ShorthandValue:
```jsx
export type ShorthandValue = React.ReactNode | Props
```

Fixes #504


layershifter (commented):
## Improve propTypes

We don't use `propTypes` to generate `componentInfo`, so may be we can just create a factory for them?

```diff
-    ...commonPropTypes.commonUIComponentPropTypes,
-    ...commonPropTypes.childrenComponentPropTypes,
+ ...customPropTypes.createCommon({
+   content: false,
+ }}
```

```diff
-    ...commonPropTypes.commonUIComponentPropTypes,
-    ...commonPropTypes.childrenComponentPropTypes,
-    ...commonPropTypes.contentShorthandComponentPropsTypes,
+ ...customPropTypes.createCommon({
+   content: 'shorthand',
+ }}
```

```diff

-  propTypes: {
-    ...commonPropTypes.commonUIComponentPropTypes,
-    ...commonPropTypes.contentNodeComponentPropsTypes,
-    ...commonPropTypes.childrenComponentPropTypes,
-  },
+ propTypes: customPropTypes.createCommon(),
```

layershifter:
It will be awesome to have a similar generic for `UIComponentProps`, but I don't have a good idea there.

mnajdova:
Thanks @layershifter this is actually a great idea about the propTypes. Will add this functionality and will try to think about something similar for the props' interfaces.

src/components/Animation/Animation.tsx
```diff
@@ -1,16 +1,26 @@
...
-export interface AnimationProps extends StyledComponentProps<any, any>, ChildrenComponentProps {
+export interface AnimationProps extends StyledComponentProps {
   /** An element type to render as (string or function). */
   as?: any
 
+  /**
+   *  Content for childrenApi
+   *  @docSiteIgnore
+   */
+  children?: React.ReactChild
```
kuzhelov: as description is duplicated now, I would rather suggest to still encapsulate it at the single place (`ChildrenComponentProps`), and just introduce optional generic parameter to this type:

```ts
ChildrenComponentProps<TChildren=React.ReactChildren> {
  /**
   *  Content for childrenApi
   *  @docSiteIgnore
   */
  children?: TChildren
}
```

src/components/Ref/Ref.tsx
```diff
@@ -2,11 +2,19 @@ import * as PropTypes from 'prop-types'
...
-import { ReactChildren } from '../../../types/utils'
 import { handleRef } from '../../lib'
 
 export interface RefProps {
-  children?: ReactChildren
+  /**
+   *  Used to set content when using childrenApi - internal only
+   *  @docSiteIgnore
+   */
+  children?: React.ReactChild
```
kuzhelov: I would not this see to be internal only, as _in case if Ref will be exposed to the client from Stardust_, either of the following is absolutely legal:

```jsx
<Ref innerRef={...}>
  <MyCustomComponent>
</Ref>
```

or its equivalent:

```jsx
<Ref innerRef={...} children={<MyCustomComponent>} />
```

So, thus, would suggest to use already existing prop type here as well (`ChildrenComponentProps`), as its description for `children` will be an exact match for this component as well

src/lib/commonPropInterfaces.ts
```diff
@@ -14,19 +15,26 @@ export interface AnimatedComponentProps {
...
+    AnimatedComponentProps {
   /** An element type to render as (string or function). */
   as?: any
 
   /** Additional CSS class name(s) to apply.  */
   className?: string
 }
 
-export interface ContentComponentProps {
+export interface ContentShorthandComponentProps {
```
kuzhelov: would suggest to still use single type (`ContentComponentProps `) with generic parameter:

```jsx
ContentComponentProps<TContent=...> { ... }
```

src/lib/commonPropTypes.ts
```diff
@@ -17,7 +17,11 @@ export const commonUIComponentPropTypes = {
   className: PropTypes.string,
 }
 
-export const contentComponentPropsTypes = {
+export const contentShorthandComponentPropsTypes = {
+  content: customPropTypes.itemShorthand,
+}
+
+export const contentNodeComponentPropsTypes = {
   content: customPropTypes.contentShorthand,
```
kuzhelov: would suggest to rename this custom prop type - while underlying type is correct, its name is misleading. Something like `nodeContent` would be a better choice here: in either case, I would rather not expect `shorthand` word being used here

src/lib/customPropTypes.tsx
```diff
@@ -378,11 +378,7 @@ export const wrapperShorthand = PropTypes.oneOfType([
  */
 export const itemShorthand = every([
   disallow(['children']),
-  PropTypes.oneOfType([
-    PropTypes.node,
-    PropTypes.object,
-    PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.node, PropTypes.object])),
-  ]),
+  PropTypes.oneOfType([PropTypes.node, PropTypes.object]),
```
kuzhelov, line 381: 👍

src/lib/commonPropTypes.ts
```diff
@@ -1,26 +1,43 @@
...
+      as: customPropTypes.as,
+    }),
+    ...(children && {
+      children: children === true ? PropTypes.node : PropTypes.element,
+    }),
+    ...(className && {
+      className: PropTypes.string,
+    }),
+    ...(content && {
+      content: content === true ? customPropTypes.contentShorthand : customPropTypes.itemShorthand,
```
layershifter: I think that `children` and `content` should accept other values:

- `children: false` => `{}`
- `children: 'node'` => `PropTypes.node`
- `children: 'element'` => `PropTypes.element`

- `content: false` => `{}`
- `content: 'content'` => `customPropTypes.contentShorthand`
- `content: 'item'` => `customPropTypes.itemShorthand,`

layershifter: And use as defaults:
- `children: 'node'` => `PropTypes.node`
- `content: 'content'` => `customPropTypes.contentShorthand`

mnajdova: Changed, just note for the content, the values available are `boolean | 'node' | 'shorthand'`, as it is aligned more with the renamed `contentShorthand -> nodeContent` custom prop type. Please take a look.

src/components/Accordion/AccordionTitle.tsx
```diff
@@ -45,9 +43,7 @@ class AccordionTitle extends UIComponent<Extendable<AccordionTitleProps>, any> {
   static className = 'ui-accordion__title'
 
   static propTypes = {
-    ...commonUIComponentPropTypes,
-    ...childrenComponentPropTypes,
-    ...contentComponentPropsTypes,
+    ...commonPropTypes.createCommon(),
```
layershifter, line 46: ❤️

layershifter (approved):
Great cleanup 👍

kuzhelov (approved):

src/components/Grid/Grid.tsx
```diff
@@ -8,13 +8,17 @@ import {
...
+import { Extendable } from '../../../types/utils'
 import { Accessibility } from '../../lib/accessibility/types'
 import { defaultBehavior } from '../../lib/accessibility'
 import ReactNode = React.ReactNode
 
-export interface GridProps extends UIComponentProps, ChildrenComponentProps {
+export interface GridProps
+  extends UIComponentProps,
+    ChildrenComponentProps,
+    ContentComponentProps<React.ReactNode | React.ReactNode[]> {
```
kuzhelov, line 21: seems that the same logic could be applied to many other components, like `AccordionContent`, `AccordionTitle`, `ButtonGroup`, etc

kuzhelov: yes, actually, would agree with that - now see the reason. In contrast to those components I've mentioned, `Grid` actually, by design, expects that multiple elements may be passed as its content

PR #529, opened 2018-11-27
docs(Prototype): Accessibility enhancements for for chat pane prototype

sophieH29:
This PR aims to make chat prototype accessible, mainly by adding additional aria-* attributes


docs/src/prototypes/chatPane/chatPaneContent.tsx
```diff
@@ -1,21 +1,37 @@
 import * as React from 'react'
 import Scrollbars from 'react-custom-scrollbars'
-import { Chat, Divider } from '@stardust-ui/react'
-
+import { Chat, Divider, ChatMessage } from '@stardust-ui/react'
 import { ChatData, ChatItemTypes, generateChatProps } from './services'
 
+const ariaLiveStyle: React.CSSProperties = {
```
jurokapsiar: `screenreaderOnly`?

docs/src/prototypes/chatPane/chatPaneContent.tsx
```diff
@@ -1,21 +1,37 @@
...
+}
+
 export interface ChatPaneContainerProps {
   chat: ChatData
 }
 
 class ChatPaneContainer extends React.PureComponent<ChatPaneContainerProps> {
   public render() {
-    const items = this.generateChatItems(this.props.chat)
+    const chat = this.props.chat
```
kuzhelov: nit: `props` object could be decomposed here, just to be consistent with other places in code:

```js
const { chat } = this.props
```

docs/src/prototypes/chatPane/chatPaneContent.tsx
kuzhelov: what these styles are for?

docs/src/prototypes/chatPane/chatPaneContent.tsx
```diff
@@ -24,9 +40,15 @@ class ChatPaneContainer extends React.PureComponent<ChatPaneContainerProps> {
   private generateChatItems(chat: ChatData): JSX.Element[] {
     return generateChatProps(chat).map(({ itemType, ...props }, index) => {
       const ElementType = this.getElementType(itemType)
+      const newProps = { ...props, text: undefined }
       return (
         <Chat.Item key={`chat-item-${index}`}>
-          <ElementType {...props} />
+          {itemType === ChatItemTypes.message && (
+            <div style={ariaLiveStyle} role="heading" aria-level={4}>
```
kuzhelov: ok, now see where these styles are used - in that case `ariaLiveContainerStyle` or `screenReaderMessageContainer` (even more preferable, to my mind - as it is much more descriptive) would work better

docs/src/prototypes/chatPane/chatPaneContent.tsx
```diff
@@ -24,9 +40,15 @@ class ChatPaneContainer extends React.PureComponent<ChatPaneContainerProps> {
...
+              {this.getMessagePreviewForScreenReader(props)}
+            </div>
+          )}
+          <ElementType {...newProps} />
```
kuzhelov: what these `newProps` stand for? :) Name doesn't explain much. I would suggest to do it explicitly, as it is the only place where those `newProps` are used:

```jsx
 <ElementType {...props} text={undefined}  />
```

docs/src/prototypes/chatPane/chatPaneContent.tsx
```diff
@@ -46,6 +68,17 @@ class ChatPaneContainer extends React.PureComponent<ChatPaneContainerProps> {
       scrollRef.scrollToBottom()
     }
   }
+
+  private getMessagePreviewForScreenReader(props) {
+    // Show the first 60 characters from the message.
+    // If to show more, NVDA splits it into 2 lines
```
kuzhelov: nit: comment's text could be adjusted and simplified a bit, something like:

> Show the first 60 characters from the message, as NVDA splits it into 2 lines if more is shown

docs/src/prototypes/chatPane/chatPaneContent.tsx
```diff
@@ -46,6 +68,17 @@ class ChatPaneContainer extends React.PureComponent<ChatPaneContainerProps> {
...
+    const messageText = props.text || ''
+    if (messageText.length > 60) {
+      return `${messageText.slice(0, 60)} ..., by ${props.author}`
```
kuzhelov: as I see, two identical strings (in a sense of composing logic) are returned. In that case this could be simplified to the following code, no branching is necessary

```js
const messageText = props.text || ''
return `${messageText.slice(0, 60)} ..., by ${props.author}`
```

docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
```diff
@@ -34,20 +37,41 @@ const statusMap: Map<UserStatus, StatusPropsExtendable> = new Map([
 function generateChatMsgProps(msg: MessageData, fromUser: UserData): ChatMessage {
   const { content, mine } = msg
   const msgProps: ChatMessage = {
-    content: msg.withAttachment
-      ? { content: createMessageContentWithAttachments(content) }
-      : content,
+    // aria-labelledby will need to by generated based on the needs. Currently just hardcoded.
```
kuzhelov, line 39: this comment should be placed directly above `aria-labelledby`

docs/src/prototypes/chatPane/chatPaneContent.tsx
sophieH29: for an aria-live element, it is not visible but needed for a screen reader to narrate message preview

docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
```diff
@@ -34,20 +37,41 @@ const statusMap: Map<UserStatus, StatusPropsExtendable> = new Map([
...
   }
 
   return msgProps
 }
 
+function createMessageContent(msg: MessageData) {
+  return {
+    id: `content-${msg.id}`,
+    ...(msg.withAttachment
+      ? { content: createMessageContentWithAttachments(msg.content) }
```
kuzhelov: branching should be placed at the inner level, for `content`'s value:

```js
{ 
  ...
  content: msg.withAttachment ? ...
}
```

docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
```diff
@@ -34,20 +37,41 @@ const statusMap: Map<UserStatus, StatusPropsExtendable> = new Map([
...
+    },
     avatar: !msg.mine && { image: fromUser.avatar, status: statusMap.get(fromUser.status) },
     itemType: ChatItemTypes.message,
+    text: content,
   }
 
   return msgProps
 }
 
+function createMessageContent(msg: MessageData) {
```
kuzhelov: lets specify the type of data that is returned from this function, similar to how it is done for any function here

docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
```diff
@@ -34,20 +37,41 @@ const statusMap: Map<UserStatus, StatusPropsExtendable> = new Map([
 function generateChatMsgProps(msg: MessageData, fromUser: UserData): ChatMessage {
   const { content, mine } = msg
   const msgProps: ChatMessage = {
-    content: msg.withAttachment
-      ? { content: createMessageContentWithAttachments(content) }
-      : content,
+    // aria-labelledby will need to by generated based on the needs. Currently just hardcoded.
+    role: undefined,
```
kuzhelov: where and how this role is used?

sophieH29: @kolaps33 can you answer on this?

sophieH29: but I guess it's because by default we have set role="presentation" in behavior. Then probably it's needed to create an additional PR for improving ```messageBehavior``` to set role="undefined".
@jurokapsiar @kolaps33 need your input here

sophieH29: discussed with @kolaps33, will update behavior in this PR (just remove "role" attribute)

docs/src/prototypes/chatPane/chatPaneContent.tsx
```diff
@@ -1,21 +1,37 @@
 import * as React from 'react'
 import Scrollbars from 'react-custom-scrollbars'
 import { Chat, Divider } from '@stardust-ui/react'
-
 import { ChatData, ChatItemTypes, generateChatProps } from './services'
 
+const screenReaderMessageContainerStyles: React.CSSProperties = {
```
kuzhelov, line 6: 👍

docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
```diff
@@ -34,21 +36,42 @@ const statusMap: Map<UserStatus, StatusPropsExtendable> = new Map([
...
     itemType: ChatItemTypes.message,
+    text: content,
   }
 
   return msgProps
 }
 
-function createMessageContentWithAttachments(content: string) {
+function createMessageContent(msg: MessageData): ShorthandValue {
+  const msgId = `content-${msg.id}`
```
kuzhelov: lets avoid shortcuts, for the sake of consistency with the rest of the code

docs/src/prototypes/chatPane/chatPaneContent.tsx
```diff
@@ -46,6 +74,14 @@ class ChatPaneContainer extends React.PureComponent<ChatPaneContainerProps> {
       scrollRef.scrollToBottom()
     }
   }
+
+  private getMessagePreviewForScreenReader(props) {
```
kuzhelov, line 78: we should think about making this to be a part of Chat's API, so that these customization means will be available to the client (as logic of this method is item-specific)

docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
```diff
@@ -34,21 +36,42 @@ const statusMap: Map<UserStatus, StatusPropsExtendable> = new Map([
 function generateChatMsgProps(msg: MessageData, fromUser: UserData): ChatMessage {
   const { content, mine } = msg
   const msgProps: ChatMessage = {
-    content: msg.withAttachment
-      ? { content: createMessageContentWithAttachments(content) }
-      : content,
+    // aria-labelledby will need to by generated based on the needs. Currently just hardcoded.
+    'aria-labelledby': `sender-${msg.id} timestamp-${msg.id} content-${msg.id}`,
```
kuzhelov: lets avoid `msg` shortcuts, as there is no need for them (and, also, this would follow our general principle to avoid shortcuts)

kuzhelov (approved):
please, consider to take a look on minor comments before merge

docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
sophieH29: agree

PR #530, opened 2018-11-27
fix(Accessibility): Remove role="presentation" from chatMessageBehavior and FocusZone

sophieH29:
After testing chat with multiple screen readers, we figured out that no ```role``` should be assigned to Chat message. Since ```FocusZone``` sets ```role=presentation``` to container by default, I override it in ```chatMessageBehavior``` by setting ```role=undefined```


jurokapsiar (approved):

src/lib/accessibility/Behaviors/Chat/chatMessageBehavior.ts
```diff
@@ -5,15 +5,14 @@ import { FocusZoneTabbableElements, FocusZoneDirection } from '../../FocusZone'
...
- * Adds role 'presentation' until we come up with final roles for chat.
  * Sets the message to be a focusable element.
  * Adds a vertical circular focus zone navigation where a user navigates using a Tab key.
  * Adds a key action which prevents up and down arrow keys from navigating in FocusZone, we only want a Tab key to navigate.
  */
 const chatMessageBehavior: Accessibility = (props: any) => ({
   attributes: {
     root: {
-      role: 'presentation',
+      role: undefined,
```
kuzhelov: shouldn't it be removed altogether then?

sophieH29: that's a good question, I was also considering that. 
@jurokapsiar, do you think this ```role=presentation``` to be set by default in FocusZone? We always can add in behavior in cases where we need it

sophieH29: After discussion with @jurokapsiar, we decided to remove that attribute from FocusZone. Thank you Roman, for raising this question, it always good to have someone to look at the situation with a fresh eye :)
So this attribute is added in the original FocusZone in fabric-ui, and they have some reasons for that, but we don't see the need for it right now, as in most cases we control ```role``` in behaviors.
Will update the PR

kuzhelov (approved):

PR #539, opened 2018-11-29
feat(menu): Submenu

### LINKED ISSUE #542: [RFC] Submenu functionality in Menu component
### Problem description
Menu should allow defining submenus. Each menu item can have a submenu.
Submenu should follow ARIA specification and best practices:
https://www.w3.org/TR/wai-aria-practices/#menu
https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-1/menubar-1.html

especially with regards to the rendered DOM strucutre and keyboard navigation:
- Submenu should be rendered as `<ul>` inside of the MenuItem `wrapper` slot. Menu items of that submenu should render normally as `<li>` items.
- Submenu is always vertical
- Home and End keys should not be used as the menu implements circular navigation.

As opposed to the specification described in the links above, submenu should not open when the user hovers on the menu item with a submenu. It should open when the user clicks on such a menu item or if he uses keyboard navigation keys for subemnu open.

### API Proposal
- add `menu` prop to the `MenuItem` component. User can pass a `Menu` shorthand to it.

### MVP
There has already been two PRs implementing this. Both had good and bad sides and the good parts were merged into #539.

jurokapsiar:
Based on two previous menu/submenu PRs. Fixes https://github.com/stardust-ui/react/issues/542

Missing things:
- [x] document click / focus out should close submenu
- [x] action should close submenu
- [x] left/right key should move to the next/previous menu item according to ARIA practices
- [x] submenu indicator
- [x] commented out overflow, not sure why it was there, need to verify
- [x] verify wrapper render

Update by @mnajdova 
Some questions:
- I needed to add key to the wrapper element created, so in the defaultProps I added key that is generated by a new added dependency `uniqid`. Are we ok with this?
- The tests are failing, I guess because I added `Ref` wrapper around the `ElementType`. How can be tackle this? Is the wrapper maybe causing this?
- How are we handling event handlers when we have wrapper? There is issue with the onClick event, which is added on the wrapper, but on the anchor is added only if the wrapper is not added. Am I doing something wrong here?

The questions were resolved thanks to @layershifter

All accessibility requirements are added, let me know if you find some issue with it.


layershifter:
Should we use the `kind` prop to simplify the render logic, #512?

src/themes/teams/components/Menu/menuStyles.ts
```diff
@@ -28,7 +28,7 @@ export default {
             ...solidBorder(variables.primaryBorderColor),
           }),
           borderRadius: pxToRem(4),
-          overflow: 'hidden',
+          // overflow: 'hidden',
```
jurokapsiar: this needs to be verified - it breaks submenu when oveflow is hidden

src/themes/teams/components/Menu/menuVariables.ts
```diff
@@ -24,7 +24,7 @@ export interface MenuVariables {
 export default (siteVars: any): MenuVariables => {
   return {
     defaultColor: siteVars.gray02,
-    defaultBackgroundColor: 'transparent',
+    defaultBackgroundColor: '#FFF',
```
jurokapsiar: this should come from theme/site

src/components/Menu/MenuItem.tsx
```diff
@@ -176,6 +272,48 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
+
+  private handleSubmenuKeyDown = (e: React.SyntheticEvent, props: MenuProps) => {
+    if (
+      keyboardKey.getCode(e) === keyboardKey.Enter ||
+      keyboardKey.getCode(e) === keyboardKey.Spacebar
+    ) {
+      this.trySetState({ submenuOpen: false })
+    }
+    _.invoke(props, 'onKeyDown', props)
+  }
```
sophieH29: Keys should be assigned to actions in behaviors. I think we can have one action called ```performSubmenuClick``` which will call ```handleSubmenuClicked``` on Enter or Space. Same way as we have on MenuItem ```performClick``` line 231.

So basically in ```menuItemBehavior``` you add to key action

```js
keyActions: {
   ...
   
   submenu: {
      performSubmenuClick: {
      },
    },
   ...
}
```

In ```MenuItem``` comp, add action handler (line 231)

```js
actionHandlers = {
    ...
    performSubmenuClick: (event) => this.handleSubmenuClicked(event, null)

    ...
}
```

NOTE: I see that in your handlers you receive props as second param. In case of ```keydown``` event, client code is handled in```getKeyDownHandlers.ts```, by calling ```_.invoke(props, 'onKeyDown', event)```, so we can just pass NULL as 2nd param in action handler.

and then line 188 where you create submenu

```js
const maybeSubmenu =
       ...
        Menu.create(menu, {
            defaultProps: {
              accessibility: submenuBehavior, 
              
            },
            overrideProps: {
              onClick: this.handleSubmenuClicked,
             ...accessibility.keyHandlers.submenu   // adds onkeyDown handler which will call 'performSubmenuClick'
            },
          })
        : null
```

Please let me know what you think.

src/lib/accessibility/Behaviors/Menu/submenuBehavior.ts
```diff
@@ -0,0 +1,31 @@
+import { Accessibility, FocusZoneMode } from '../../types'
```
sophieH29: Don't see any reason for now to have ```submenuBehavior``` as it is the same as ```menuBehavior```

src/components/Menu/MenuItem.tsx
```diff
@@ -118,32 +152,74 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
+          {content}
+        </ElementType>
+      </Ref>
     )
+    const maybeSubmenu =
+      menu && active && submenuOpen ? (
+        <Ref
+          innerRef={domElement => {
+            this.submenuDomElement = domElement
+          }}
```
layershifter: Please use a `createRef()` API there: https://github.com/stardust-ui/react/pull/584#discussion_r241463373

mnajdova: Done, thanks!

src/themes/teams/components/Menu/menuStyles.ts
mnajdova: This is the PR that added it: https://github.com/stardust-ui/react/pull/360 Will sync with @Bugaa92 for finding alternative solution for the cropped borders, I have new ideas.

src/components/Menu/MenuItem.tsx
```diff
@@ -153,18 +225,69 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
+    openSubmenu: event => this.openSubmenu(event),
+    closeMenu: event => this.closeMenu(event),
+    closeSubmenu: event => this.closeSubmenu(event),
+  }
+
+  private updateOutsideClickSubscription() {
+    this.outsideClickSubscription.unsubscribe()
+
+    if (this.props.menu && this.state.submenuOpen) {
+      setTimeout(() => {
```
kuzhelov, line 255: note, this approach (presumably, taken from the Popup) has one subtle problem: although it solves the issue of item being opened and closed within the same frame, it also opens possibility for the following behavior:

- MenuItem has opened submenu, this opened prop value doesn't change anyhow for the next JS frame.

///------ JS FRAME --------////
- render method was called (but `submenuOpen ` wasn't changed changed anyhow, it is still `true`)
- updateOutsideClickSubscription
  -  this.outsideClickSubscription.unsubscribe(), no subscription on outside click (!!)
  - setTimeout(() => subscription will be made on the next frame)

- ... some other logic ..
   - note, there is no click subscription at this moment, but it should be (!!)

////---------------------------////

Would suggest to leave a fix for a separate PR, for both components.

src/components/Menu/MenuItem.tsx
```diff
@@ -110,6 +139,11 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
     underlined: PropTypes.bool,
     vertical: PropTypes.bool,
     wrapper: PropTypes.oneOfType([PropTypes.node, PropTypes.object]),
+    menu: customPropTypes.itemShorthand,
+    submenuOpen: PropTypes.bool,
```
kuzhelov: this one is easy to be confused with `openSubmenu` which we have in the same file. Verb, probably, would help to better differentiate - or past tense could be used. Lets see whether this is consistent with the rest of the code (and naming used), but I would rather like to eliminate this confusion.

Potential suggestions:
- `isSubmenuOpen`
- `submenuOpened`

src/components/Menu/MenuItem.tsx
```diff
@@ -118,32 +152,70 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
     wrapper: { as: 'li' },
   }
 
+  static autoControlledProps = ['submenuOpen']
+
   state = {
     isFromKeyboard: false,
+    submenuOpen: false,
```
kuzhelov: won't this override the effect of `submenuOpen={true}` prop being provided by the client? Have taken a look on the logic of `AutoControlledComponent` (`getAutoControlledStateValue`) and it seems to be pretty much the case

docs/src/examples/components/Menu/Types/MenuExampleWithSubmenu.shorthand.tsx
```diff
@@ -0,0 +1,57 @@
...
+  {
+    key: 'review',
+    content: 'Reviews',
+    menu: {
+      items: [
+        { key: '1', content: 'item1' },
+        {
+          key: '2',
+          content: 'item2',
+          menu: { items: [{ key: '1', content: 'item2.1' }, { key: '2', content: 'item2.2' }] },
```
kuzhelov, line 33: 👍 for naming consistency from level to level in the nested tree (no props like `submenu`, etc)

src/components/Menu/MenuItem.tsx
mnajdova: Yep, we should never have initial value for e state that is part of the Autocontrolled prop. This was leftover from the prev PRs merged here. Will remove it. Thanks for the catch!

src/components/Menu/MenuItem.tsx
mnajdova: We never use `is` as a prefix for the boolean props, and the similar auto-controlled prop in the Popup, has the name `open`. But, I agree with your concerns, would rather rename it to `submenuOpened`, but not sure as it won't be consistent. What do you think?

src/components/Menu/MenuItem.tsx
mnajdova: Will create issue for this, after the PR will be merged

src/components/Menu/MenuItem.tsx
```diff
@@ -153,18 +225,69 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
+              {menuItemInner}
+              {maybeSubmenu}
+            </>
+          ),
+          onClick: this.handleClick,
+          onBlur: this.handleWrapperBlur,
         }),
       })
     }
     return menuItemInner
```
kuzhelov, line 234: sorry, just to understand what current limitations are. What will happen if `wrapper` will be evaluated to falsy value?
There won't be any submenu functionality supported in this case, right?

src/components/Menu/MenuItem.tsx
kuzhelov: lets rely on consistency first - by that we will ensure that problem will be fixed for all components and in the same way. If this name is consistent with other components' prop names, lets leave it 👍

src/components/Menu/MenuItem.tsx
```diff
@@ -118,32 +152,65 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
-        {content}
-      </ElementType>
+      <Ref innerRef={this.itemRef}>
+        <ElementType
+          className={classes.root}
+          onBlur={this.handleBlur}
+          onFocus={this.handleFocus}
+          {...accessibility.attributes.anchor}
+          {...rest}
+          {...!wrapper && { onClick: this.handleClick }}
```
kuzhelov, line 189: are we really sure that we would like to override client's `onClick` logic in that case?

src/components/Menu/MenuItem.tsx
```diff
@@ -118,32 +152,65 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
+        </ElementType>
+      </Ref>
     )
+    const maybeSubmenu =
+      menu && active && submenuOpen ? (
+        <Ref innerRef={this.submenuRef}>
+          {Menu.create(menu, {
+            defaultProps: {
+              accessibility: submenuBehavior,
+              vertical: true,
```
kuzhelov, line 205: is there any possibility that it could be horizontal one? I mean, I might imagine it (horizontal menu and horizontal submenu) - this was the case for Word's toolbar, for example

src/components/Menu/MenuItem.tsx
mnajdova: Yes, because the submenu as an element is part of the wrapper (it cannot be part of the anchor tag, which is the element created if the wrapper is not provided). I agree that this is not obvious, maybe we should add that in the doc somewhere.

src/components/Menu/MenuItem.tsx
mnajdova: Actually, we are not overriding. The implementation of `handleClick` is the following:
```
  private handleClick = e => {
    this.performClick(e)
    _.invoke(this.props, 'onClick', e, this.props)
  }
```

src/components/Menu/MenuItem.tsx
mnajdova: According to the spec: https://github.com/stardust-ui/react/issues/542 the submenus should always be vertical. @jurokapsiar opinion on this?

docs/src/examples/components/Menu/Types/index.tsx
```diff
@@ -19,6 +19,16 @@ const Types = () => (
       description="A vertical menu displays elements vertically."
       examplePath="components/Menu/Types/MenuExampleVertical"
     />
+    <ComponentExample
+      title="Menu with Submenus"
+      description="A menu can have submenus."
+      examplePath="components/Menu/Types/MenuExampleWithSubmenu"
+    />
+    <ComponentExample
+      title="Vertical submenus"
```
sophieH29: I would name it ```Vertical menu with submenus```. Because both of the examples are vertical submenus :)

src/components/Menu/MenuItem.tsx
```diff
@@ -179,6 +297,44 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
 
     _.invoke(this.props, 'onFocus', e, this.props)
   }
+
+  private closeMenu = e => {
+    const { menu, inSubmenu } = this.props
+    const { submenuOpen } = this.state
+    if (menu && submenuOpen) {
+      this.setState({ submenuOpen: false })
+      if (!inSubmenu && (keyboardKey.getCode(e) === keyboardKey.Escape || this.props.vertical)) {
```
sophieH29: what is a use case to have an additional check on ```Esc``` button? We should avoid using keys in components, if you need a specific thing to happen on Esc, better create a separate action handler for that.

src/lib/accessibility/Behaviors/Menu/menuBehavior.ts
```diff
@@ -21,6 +22,8 @@ const menuBehavior: Accessibility = (props: any) => ({
       isCircularNavigation: true,
       preventDefaultWhenHandled: true,
       shouldFocusFirstElementWhenReceivedFocus: true,
+      // TODO: check if this should be applied to other behaviors as well
```
sophieH29: probably, we don't need this comment here :)

src/components/Menu/MenuItem.tsx
```diff
@@ -153,18 +220,69 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
+        this.outsideClickSubscription = EventStack.subscribe('click', this.outsideClickHandler)
+      })
+    }
+  }
+
+  private outsideClickHandler = e => {
+    if (
+      this.itemRef &&
+      (!this.itemRef.current || !this.itemRef.current.contains(e.target)) &&
+      (!this.submenuRef.current || !this.submenuRef.current.contains(e.target))
```
layershifter: We have a `doesNodeContainClick()` util.

src/components/Menu/MenuItem.tsx
mnajdova: Added new handler, thanks for the suggestion!

src/lib/accessibility/Behaviors/Menu/menuBehavior.ts
mnajdova: I left it intentionally for your guys, as I don't think we had final decision whether we wanted to have this in the other behaviors.

src/components/Menu/MenuItem.tsx
mnajdova: Changed, thanks!

docs/src/examples/components/Menu/Types/MenuExampleWithSubmenu.shorthand.tsx
```diff
@@ -0,0 +1,57 @@
...
+  { key: 'events', content: 'Upcoming Events' },
+]
+
+const MenuExampleWithSubMenu = () => (
+  <Provider
+    theme={{
+      componentStyles: {
+        Menu: {
+          root: {
+            zIndex: 1000,
```
kuzhelov: shouldn't it be part of the default styles? like

```
// menuItemStyles
'& .ui-menu {
  z-index: ..
}'
```

src/components/Menu/Menu.tsx
```diff
@@ -88,6 +96,7 @@ class Menu extends AutoControlledComponent<Extendable<MenuProps>, any> {
     secondary: customPropTypes.every([customPropTypes.disallow(['primary']), PropTypes.bool]),
     underlined: PropTypes.bool,
     vertical: PropTypes.bool,
+    submenu: PropTypes.bool,
```
kuzhelov, line 99: here is another clear example - lets just take a look on the following code:

```
<Menu submenu={...} />
```

given this code, what we might expect to be passed as a `submenu`'s value? actually, I would expect items collection. This is what we have for `MenuItem`'s `menu`, but here - this is a boolean flag :)

src/components/Menu/MenuItem.tsx
```diff
@@ -78,16 +91,32 @@ export interface MenuItemProps
 
   /** Shorthand for the wrapper component. */
   wrapper?: ShorthandValue
+
+  /** Shorthand for the submenu. */
+  menu?: ShorthandValue
+
+  /** Indicates if the submenu is open */
+  submenuOpen?: boolean
```
kuzhelov: don't like this names duality - on the one side we have `menu`, on the other one `submenuOpen` - while, in fact, the same exact menu is implied by both. Consider the following expression as an example:

```if (menu && submenuOpen) { ... }```

 Suggest to rename to `menuOpen`

File: src/components/Menu/MenuItem.tsx [hunk 20]
```diff
@@ -179,6 +297,55 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
+        focusAsync(this.itemRef.current)
+      }
+    }
+  }
+
+  private closeMenuAndFocusNextParentItem = e => {
+    const { menu, inSubmenu } = this.props
+    const { submenuOpen } = this.state
+    if (menu && submenuOpen) {
+      this.setState({ submenuOpen: false })
```
kuzhelov: how focus is passed in case if it is not a top-level menu? Are we implicitly relying on event bubbling here?

src/components/Menu/MenuItem.tsx
```diff
@@ -110,6 +139,11 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
     underlined: PropTypes.bool,
     vertical: PropTypes.bool,
     wrapper: PropTypes.oneOfType([PropTypes.node, PropTypes.object]),
+    menu: customPropTypes.itemShorthand,
+    submenuOpen: PropTypes.bool,
+    defaultSubmenuOpen: PropTypes.bool,
+    setActiveIndex: PropTypes.func,
```
kuzhelov: `index` is only used for making this item `active` by this callback. We can remove `index` altogether and rename this prop to something that is more specific (and better reflects intent of this method).

Actually, the thing we have here is event that is fired in controlled mode of `MenuItem` - and, thus, we can name it as `onActiveChanged` and call when `MenuItem` would like to 'activate' itself

kuzhelov (commented):
lets discuss mentioned points and try to simplify the approach a bit. 

Also, please, note that currently we have the following problem related to visual representation of focused item:

image

Could you say which one os focused now? :)

mnajdova:
@kuzhelov thanks for the review. Sure, let's discuss the points, and for the focus styling (it is something that has been already implemented. Not sure what are the spec for this, but we can change the color of course, and update it later if that is not the correct one).

src/components/Menu/MenuItem.tsx
mnajdova: Fow now yes, let's consider implementing this with callback mechanism in a follow up PR.

src/components/Menu/Menu.tsx
mnajdova: Totally agree, for consistency I will leave it as this, and will create issue for renaming all boolean props to have the `is` prefix. Thanks for pointing this out.

docs/src/examples/components/Menu/Types/MenuExampleWithSubmenu.shorthand.tsx
mnajdova: I thought this was added only because of the docs experience, but it turns out that it is common practice to add z-index on the submenus. Will add this styles in the MenuItem, thanks!

docs/src/examples/components/Menu/Types/index.tsx
```diff
@@ -19,6 +19,16 @@ const Types = () => (
...
     />
+    <ComponentExample
+      title="Menu with submenus"
+      description="A menu can have submenus."
+      examplePath="components/Menu/Types/MenuExampleWithSubmenu"
+    />
+    <ComponentExample
+      title="Vertical menu with submenus"
+      description="A vertical menu can have submenu."
+      examplePath="components/Menu/Types/MenuExampleVerticalWithSubmenu"
```
layershifter: This example should be present, but we can leverage the attention to it there.

```diff
<ComponentExample
-  title="Vertical menu with submenus"
-  description="A vertical menu can have submenu."
  examplePath="components/Menu/Types/MenuExampleVerticalWithSubmenu"
/>
```

<details>
  <summary>Preview</summary>

image
</details>

src/components/Menu/Menu.tsx
```diff
@@ -60,12 +61,19 @@ export interface MenuProps extends UIComponentProps, ChildrenComponentProps {
 
   /** A vertical menu displays elements vertically. */
   vertical?: boolean
+
+  /** Indicates whether the menu is submenu. */
+  submenu?: boolean
+}
+
+export interface MenuState {
```
layershifter, line 69: Should we export this interface?

src/components/Menu/Menu.tsx
```diff
@@ -119,11 +137,14 @@ class Menu extends AutoControlledComponent<Extendable<MenuProps>, any> {
...
       vertical,
+      submenu,
     } = this.props
     const { activeIndex } = this.state
 
-    return _.map(items, (item, index) =>
-      MenuItem.create(item, {
+    return _.map(items, (item, index) => {
+      const active =
+        typeof activeIndex === 'string' ? parseInt(activeIndex, 10) : activeIndex === index
```
layershifter: This condition is not valid, if we `activeIndex` is `string` it will be converted to `number`, in other cases it will be boolean:

```js
let activeIndex = 1
let index = 1
let active

active = typeof activeIndex === 'string' ? parseInt(activeIndex, 10) : activeIndex === index
console.log(typeof active) // boolean

activeIndex = '1'
active = typeof activeIndex === 'string' ? parseInt(activeIndex, 10) : activeIndex === index
console.log(typeof active) // number
```

src/components/Menu/Menu.tsx
kuzhelov: while understand the reason for this sentiment, it turns out that, actually, we should - as in the style functions we need to provide both state and props types, as one of the args of this style function is the result of `props` being merged to `state`:

```js
// this is the code that will be written by client
const buttonStyles = {
  // here props are of type (ButtonState & ButtonProps)
  root: ({ props, ... }) => { ... }
```

src/components/Menu/Menu.tsx
mnajdova: Yep, I am missing brackets
` typeof activeIndex === 'string' ? parseInt(activeIndex, 10) : activeIndex === index` => ` (typeof activeIndex === 'string' ? parseInt(activeIndex, 10) : activeIndex) === index
`

src/components/Menu/Menu.tsx
mnajdova: Agreed, will export it.

src/components/Menu/Menu.tsx
mnajdova: The active prop should always be boolean. Fixed.

mnajdova:
All comments are resolved. I changed the look of the hovered/focused vs active styles. Couple of examples of the variants:
image
image
image
image
image

src/components/Menu/Menu.tsx
mnajdova: Exported

src/lib/accessibility/Behaviors/Menu/menuBehavior.ts
mnajdova: @sophieH29 @jurokapsiar should we add the direction prop to the toolbarBehavior, tabListBehavior as well, cus we have circular navigation there as well?

sophieH29: yes, we probably will add it in the future when needed. But don't see any reason to have this comment here.

docs/src/examples/components/Menu/Variations/MenuExampleIconOnlyPrimaryInverted.shorthand.tsx
```diff
@@ -14,9 +14,9 @@ const MenuExampleIconOnlyPrimaryInverted = () => (
     items={items}
     primary
     variables={siteVars => ({
-      defaultColor: siteVars.gray06,
-      defaultBackgroundColor: siteVars.brand,
-      typePrimaryActiveBorderColor: siteVars.white,
+      color: siteVars.gray06,
```
kuzhelov, line 17: 👍

src/components/Menu/MenuItem.tsx
```diff
@@ -153,18 +220,69 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
+      })
+    }
+  }
+
+  private outsideClickHandler = e => {
+    if (
+      !doesNodeContainClick(this.itemRef.current, e) &&
+      !doesNodeContainClick(this.menuRef.current, e)
+    ) {
+      this.state.menuOpen && this.trySetState({ menuOpen: false })
```
kuzhelov: nit: we can make this check (`this.state.menuOpen`) as the first step, because (`doesNodeContainClick`) are less efficient ones:

```js
if (!this.state.menuOpen) {
  return 
}
...
```

src/components/Menu/MenuItem.tsx
```diff
@@ -153,18 +220,69 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
...
+    }
+  }
+
+  private performClick = e => {
+    const { active, menu } = this.props
+    if (menu) {
+      if (doesNodeContainClick(this.menuRef.current, e)) {
+        // submenu was clicked => close it and propagate
```
kuzhelov, line 275: thanks for comments, very helpful ones!

src/components/Menu/MenuItem.tsx
```diff
@@ -179,6 +297,55 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
 
     _.invoke(this.props, 'onFocus', e, this.props)
   }
+
+  private closeAllMenus = e => {
+    const { menu, inSubmenu } = this.props
+    const { menuOpen } = this.state
+    if (menu && menuOpen) {
+      this.setState({ menuOpen: false })
+      if (!inSubmenu) {
```
kuzhelov: for some cases we have this focus logic being run in the callback of the state change (which is correct), somewhere not. Lets make approach to that being consistent (or, if there is any reason why it is not, please, just let me know)

src/components/Menu/MenuItem.tsx
```diff
@@ -179,6 +297,55 @@ class MenuItem extends UIComponent<Extendable<MenuItemProps>, MenuItemState> {
 
     _.invoke(this.props, 'onFocus', e, this.props)
   }
+
+  private closeAllMenus = e => {
```
kuzhelov: it is hard to infer the difference of logic provided by this method vs subsequent `closeMenuAndFocusNextParentItem` one by using their names only - more than that, quite a lot of implementation code is similar. Maybe we could refactor it to be a single method with some parameter that will drive (and describe) this difference?

mnajdova: It was initially one method, but we have different handlings if the esc key was used, or right arrow, so that suggestions from behaviors point of view was that we should have different methods for these, as we don't want to use the check for the keys inside the component's logic.

src/components/Menu/MenuItem.tsx
mnajdova: Initially I had here check for which key was pressed, and I could not use the event in async manner there. That is not the case anymore, as we decided to use different handlers for different keys. We can move this safely in the callback now.

src/components/Menu/MenuItem.tsx
kuzhelov: as a proof for my point:

image

kuzhelov (approved):

src/components/Menu/MenuItem.tsx
mnajdova: Will be added in a follow up PR.

PR #545, opened 2018-11-29
[IGNORE] run visual tests in parallel

kuzhelov:


kuzhelov:
agreed to not introduce any moments that would require interaction to the build process - it should still be completely automated

PR #551, opened 2018-12-03
fix(Button): content shorthand as element was not truncating the text inside

### LINKED ISSUE #520: bug(Teams theme): Button should apply truncate styles to any content
## Bug Report

It is necessary to ensure that `Button`'s content other than string will be truncated for `Button`. Main case that drives this necessity is properly displaying `Button`'s text in RTL mode, which requires `Button`'s content to be wrapped in <span dir="auto">.

### Steps
Use the following code to repro the problem:

```jsx
<Button content={<span dir="auto">See how this very long text shows up on the button</span>} />
```

### Expected Result
Content that is wrapped in, say, <span /> element should be truncated as well. Note that <span> element might be necessary to properly support RTL mode by button context, where content text is necessary to be wrapped by <span dir="auto">.

image

### Actual Result
Currently truncate styles are applied to string content only.

image

### Version
0.12.0

mnajdova:
This PR fixed https://github.com/stardust-ui/react/issues/520 - truncating the content of the button, even if it is provided as a element. 

## Prev: 
content:
image

## Fixed:
content:
image

Not, sure if we want to, but I tried fixing the issue for the children API as well, by adding this styling on the root, but unfortunately `display: 'inline-flex'`, cannot be used with `textOverflow: 'ellipsis'`. I tried replacing `inline-flex` with `inline-block`, but there were lots of regressions, I tried fixing them but it is not something trivial (basically, I tried replacing `alignItems: 'center'` and `justifyContent: 'center',` with `verticalAlign: 'middle'`, then I had to add float for the icon on order for it to be correctly align with the content, but still there are lots of regressions... If there is some proposal, please let me know, and I will try them...) But yes, the most important thing is whether we want to support this for the children API, if we don't then the changes in this PR are enough for fixing the existing issue.


kuzhelov (approved):

kuzhelov:
I would leave `Children API` for future discussion - there are several reasons for why it seems to be right thing to do, the most appealing one, to my mind, is that
- we are not exposing any prop that will guarantee truncated content for the `Button` at the moment (like `truncateContent`). In that case, yes, we would need to worry about default styling approach that would guarantee this effect
- client is not limited to introduce truncation effect by herself - actually, children API suggests that content will be rendered as is, so, if necessary, truncate/no-overflow styles could be provided by client then

PR #555, opened 2018-12-04
fix(Editor): fix broken HTML preview when Provider.Consumer was used

layershifter:
This PR fixes broken HTML preview for examples that using `Provider.Consumer`, it happens because `renderToStaticMarkup()` was done without `Provider`.

image

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
+++ b/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
@@ -260,6 +260,10 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
       return this.renderMissingExample()
     }
 
+    return <SourceRender.Consumer>{({ element }) => element}</SourceRender.Consumer>
+  }
+
+  private renderElement = (element: React.ReactElement<any>) => {
     const { showRtl, componentVariables, themeName } = this.state
     const theme = themes[themeName]
 
@@ -270,11 +274,7 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
       rtl: showRtl,
     }
 
-    return (
-      <Provider theme={newTheme}>
-        <SourceRender.Consumer>{({ element }) => element}</SourceRender.Consumer>
-      </Provider>
-    )
+    return <Provider theme={newTheme}>{element}</Provider>
   }
 
   private renderMissingExample = (): JSX.Element => {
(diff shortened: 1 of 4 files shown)


docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -270,11 +274,7 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
       rtl: showRtl,
     }
 
-    return (
-      <Provider theme={newTheme}>
-        <SourceRender.Consumer>{({ element }) => element}</SourceRender.Consumer>
-      </Provider>
-    )
+    return <Provider theme={newTheme}>{element}</Provider>
```
kuzhelov, line 277: this seems to be another problem - seems that `theme` was intended to be used here

layershifter: Can you please clarify what you mean?

mnajdova, line 277: Is this always invoked with the result from `renderExampleWithCode()`? Not sure whether we want to remove the consumer from here.

package.json
```diff
@@ -132,7 +132,7 @@
     "react-hot-loader": "^4.1.3",
     "react-router": "^4.1.2",
     "react-router-dom": "^4.1.2",
-    "react-source-render": "^2.0.0-beta.3",
+    "react-source-render": "^2.0.0-beta.4",
```
mnajdova, line 135: Why we needed this change?

layershifter: Because the `render` prop was introduced in this version of the package.

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
layershifter: Nope, it will called with the result of rendering i.e. an element from `ButtonExample`. This result will be will wrapped with our `Provider` and then will be passed to `renderExampleFromCode()`. It's quite similar to `render*` props that we had before.

mnajdova: Got it, thanks for the clarification.

kuzhelov: my bad here, not noticed that some lines are hidden

mnajdova (approved):
Works as expected! 👍

kuzhelov (approved):

PR #558, opened 2018-12-05
refactor(Divider): remove `type` prop

### LINKED ISSUE #421: Align `Menu` and `Divider` components to use the `type` property values as boolean

layershifter:
Fixes #421.

# BREAKING CHANGES

We decided to drop `type` prop on this component because `type` is a reserved prop in HTML.

### Before

```jsx
<Divider type='primary' />
<Divider type='secondary' />
```

### After

```jsx
<Divider color='primary' />
<Divider color='secondary' />
```


docs/src/views/ThemingExamples.tsx
```diff
@@ -90,20 +83,21 @@ export default () => (
...
-        `      styles={styles} />`,
+        `  <Button`,
+        `    content="Send email"`,
+        `    icon={{`,
+        `      name: "at",`,
+        `      styles: { color: "brown" }`,
+        `    }}`,
+        `    secondary`,
+        `    styles={styles} />`,
+        `  />`,
```
layershifter: I aligned code in these examples.

<details>
  <summary>Before</summary>

  image
</details>
<details>
  <summary>After</summary>

image

</details>

layershifter:
I decided to remove `Divider` from theming examples because we using color palette there and examples will be overcomplicated (current and previous pattern).

After we will apply color palette to all our components we will return `Divider`.

docs/src/views/ThemingExamples.tsx
mnajdova: 👍

src/themes/teams-high-contrast/components/Divider/dividerVariables.ts
```diff
@@ -2,7 +2,9 @@ import { DividerVariables } from '../../../teams/components/Divider/dividerVaria
 import { Partial } from 'types/utils'
 
 export default (siteVars: any): Partial<DividerVariables> => ({
+  colors: {
+    primary: siteVars.white,
+  },
```
mnajdova, line 7: No `secondary` here?

layershifter: Yes, it was never supported in `Divider` :(

src/themes/teams/components/Divider/dividerVariables.ts
```diff
@@ -1,25 +1,25 @@
+import * as _ from 'lodash'
+
 import { EmphasisColors, NaturalColors } from '../../../types'
 import { pxToRem } from '../../../../lib'
 
 export interface DividerVariables {
-  colors: EmphasisColors & NaturalColors
+  colors: Record<keyof (EmphasisColors & NaturalColors), string>
```
mnajdova, line 7: 👍

src/themes/teams/components/Divider/dividerVariables.ts
```diff
@@ -1,25 +1,25 @@
...
   textLineHeight: string
-  primaryColor: string
   importantFontWeight: string
   dividerPadding: string
 }
 
 export default (siteVars: any): DividerVariables => {
   return {
-    colors: { ...siteVars.emphasisColors, ...siteVars.naturalColors },
+    colors: _.mapValues({ ...siteVars.emphasisColors, ...siteVars.naturalColors }, '500'),
```
mnajdova: What is the magic `'500'` doing here? :)

src/themes/teams-high-contrast/components/Divider/dividerVariables.ts
mnajdova: Got i! :\

src/themes/teams/components/Divider/dividerVariables.ts
layershifter: We have `colors` with following structure:
```js
const colors = {
  blue: { 500: 'blue' // other values },
  pink: { 500: 'pink'  },
}
```

I decided to reduce this structure with `mapValues` to as we using only single variant:
```js
const colors = {
  blue: 'blue',
  pink: 'pink',
}
```

This structure is more usable for overrides, too.
Yes, it looks like a magic number, but it is a simply color variant.

layershifter: We can do this:

```js
const colorVariant = '500'

return {
  colors: _.mapValues({ ...siteVars.emphasisColors, ...siteVars.naturalColors }, colorVariant),
}
```

May be, this change will improve readability.

docs/src/views/ThemingExamples.tsx
levithomason: @layershifter I just merged prettier integration #568 which auto formats CodeSnippets and ExampleSnippets.  It also removes the need to have the `value` string prop altogether by printing the `render` prop return value and prettifying it 👍 

Just fix conflicts here to account for that.

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### BREAKING CHANGES
+- Remove `type` in `Divider` component @layershifter ([#558](https://github.com/stardust-ui/react/pull/558))
```
levithomason: Perhaps note that `type` is "replaced" by the `color` prop, instead of just removed?

src/themes/teams/components/Divider/dividerVariables.ts
levithomason: TIL: `Record`

src/themes/teams/components/Divider/dividerVariables.ts
levithomason: We can save my comment here for another PR, but here are some thoughts.

***

Wondering if we should go this route, defining single color variable for every component.  Most components will implement the same variant of a color for the same purpose.  Example, a blue border will be the same shade in components that have blue borders.  Red text is similar where ever it appears.  The odds that a specific component needs to override a specific shade of color I think would be the exception and not the norm.

We might be able to produce a "scheme" for a component to implement.  Imagine picking colors along a color object like this:

```jsx
const primary = {
  50: '#F4F4FC',  // background
  100: '#E2E2F6', // shadow
  200: '#BDBDE6', // border
  300: '#8F90C1',
  400: '#6E70AE',
  500: '#6264A7', // text
  600: '#55578D',
  700: '#4A4C78',
  800: '#414265',
  900: '#33344A',
}
```
image

Or, an inverted version of the same component using different variants from the same object:

```jsx
const primary = {
  50: '#F4F4FC',  // text
  100: '#E2E2F6',
  200: '#BDBDE6', // shadow
  300: '#8F90C1',
  400: '#6E70AE',
  500: '#6264A7', // background
  600: '#55578D',
  700: '#4A4C78', // border
  800: '#414265',
  900: '#33344A', 
}
```
image

### Divider

Applied to the Divider, I can imagine variables for the `borderColor` and a `textColor` since these are the relevant parts of the component and can have different colors, such as here:

image

However, I think what is missing in this PR right now is that there is no indication of which variant the text color will use compared to the border variant.  We might want a way to choose a "scheme" for applying to a component or at least specify which variants will be used.

### Schemes

If we try the scheme route, we might look at many components in a UI and ask what the common color schemes are.  A scheme would includes colors for text, background, border, and shadow.

I can see defining a "primary" scheme or a "secondary" scheme, even "inverted" schemes in this way.  The scheme from the first Message looking component above would look something like this:

```jsx
const scheme = {
  foreground: primary[500],

  quiet: primary[400],         // range of "muted" foreground colors
  quieter primary[300],
  quietest: primary[200],

  background: primary[50],
  border: primary[200],
  shadow: primary[100]
}
```

These could be generated for any color, inversion, emphasis, and state like disabled.  They could also be adjusted for higher/lower contrast or brightness.

CHANGELOG.md
layershifter: Updated, thanks 👍

src/themes/teams/components/Divider/dividerVariables.ts
layershifter: Yes, it's a long discussion :) Not for this PR, that's why I removed `Divider` from examples of theming.

docs/src/views/ThemingExamples.tsx
mnajdova: Not sure how would this reflect the fact that we want to show the examples as Javascript code, but we are writing the actual rendered code in Typescript... I have something like this in the Integrate Custom Components guide. What is the proposal for addressing this?

src/themes/teams/components/Divider/dividerVariables.ts
mnajdova: @layershifter yep, please introduce the changes proposed in https://github.com/stardust-ui/react/pull/558#discussion_r it is much clear that way.

mnajdova (approved):
👍

src/themes/teams/components/Divider/dividerVariables.ts
```diff
@@ -1,25 +1,27 @@
+import * as _ from 'lodash'
+
 import { EmphasisColors, NaturalColors } from '../../../types'
 import { pxToRem } from '../../../../lib'
 
 export interface DividerVariables {
-  colors: EmphasisColors & NaturalColors
+  colors: Record<keyof (EmphasisColors & NaturalColors), string>
```
bmdalex, line 7: @layershifter  can we extract this interface
`Record<keyof (EmphasisColors & NaturalColors), string>`
in `src/themes/types.ts` to reuse it for other's components `color` prop?

src/themes/teams/components/Divider/dividerVariables.ts
```diff
@@ -1,25 +1,27 @@
...
   importantFontWeight: string
   dividerPadding: string
 }
 
 export default (siteVars: any): DividerVariables => {
+  const colorVariant = '500'
+
   return {
-    colors: { ...siteVars.emphasisColors, ...siteVars.naturalColors },
+    colors: _.mapValues({ ...siteVars.emphasisColors, ...siteVars.naturalColors }, colorVariant),
```
bmdalex, line 20: @layershifter  can we extract this method
`_.mapValues({ ...siteVars.emphasisColors, ...siteVars.naturalColors }, colorVariant),`
or even an object if the `colorVariant` is always 500
`export themeColors = _.mapValues({ ...siteVars.emphasisColors, ...siteVars.naturalColors }, 500)`
 to reuse it for other's components `color` prop?

PR #559, opened 2018-12-05
fix(Provider): staticStyles should be executed with the merged siteVariables

### LINKED ISSUE #526: Function to resolve static styles gets incomplete site variables
### Steps
          <UIProvider
            theme={{
              staticStyles: [
                     resolveThemedStyles,
                     …...
              ]
            }}
          >

function resolveThemedStyles(siteVariables) {
    // Site variables only contains font...
    // Where are all theme variables?
}

### Expected Result
Expect the siteVariable parameter to have correct site variables

### Actual Result

### Version
x.y.z

### Testcase

mnajdova:
This PR aims to fix https://github.com/stardust-ui/react/issues/526 
# Note
Important thing to note is that, per one Provider we will execute the staticStyles only once, when they are the first time applied.


src/components/Provider/Provider.tsx
```diff
@@ -23,7 +23,14 @@ export interface ProviderProps {
...
  */
-class Provider extends React.Component<ProviderProps, any> {
+class Provider extends React.Component<ProviderProps> {
+  staticStylesRendered: boolean
+
+  constructor(props: ProviderProps) {
+    super(props)
+    this.staticStylesRendered = false
+  }
+
```
layershifter, line 28:
```suggestion
staticStylesRendered: boolean = false
```

You can set a value without `constructor`, or I am missing something?

mnajdova: Yep, I will remove the constructor. I was trying different things, it is just leftover.

levithomason (approved):

kuzhelov (approved):

PR #563, opened 2018-12-05
feat(FocusZone): Handle keyDownCapture based on client needs

sophieH29:
This PR aims to extend FocusZone API, so the client can decide whether global ```onKeyDownCapture``` should be handled or the client wants to control tab indexes on his own.

The current use case for that  is when using Chat with VirtualList, chat items are loaded asynchronously so FocusZone doesn't know when items are in place so it can update tab indexes in incorrect order. It's only happens when we Tab to chat. As long as tab indexes are set by client in correct order (as they are aware of the whole state and ecosystem of Virtual list), FocusZone will handle the rest (navigation with keys, and updating tabindexes while navigating)


jurokapsiar (approved):

silviuaavram (approved):

src/lib/accessibility/FocusZone/FocusZone.types.ts
```diff
@@ -58,6 +58,11 @@ export interface FocusZoneProps extends React.HTMLAttributes<HTMLElement | Focus
    */
   shouldFocusFirstElementWhenReceivedFocus?: boolean
 
+  /**
+   * If global onKeyDownCapture should be handled and updating tab indexes
```
layershifter: NIT: Missing trailing dot 😸

layershifter (approved):
This change looks like a reasonable to me.

layershifter:
@sophieH29 We changed `chatBehaviour` there, I think that we need to update the main `CHANGELOG.md` , too.

kuzhelov (approved):

PR #566, opened 2018-12-05
fix(SelectableList): Items in list should be selectable

### LINKED ISSUE #464: [Selection list] Aria-selected is not changing the state, is always false
### Steps
Testing with reader require to open it in Edge/Narrator. 
1. Navigate to the selection list item
2. press down arrow key to move to another list item
3. press down arrow key to move to another list item
4. observe aria-selected attribute or listen Narrator. 

### Expected Result
aria-selected should change state with selected item

### Actual Result
aria-selected doesn't change state based on selected item
reader doesn't narrate proper selected state

### Version
latest master, before version is v0.11.0

### Testcase

### LINKED ISSUE #565: Selectable list focus handling
## Feature Request

Selectable list has a specific focus handling:
- The list itself needs to have tabindex=0 by default
- When the list is focused, its tabindex needs to be changed to -1, the tabindex of the currently navigable item needs to be set to 0. Consequently, that item needs to be focused.
- When the list and listitems loose focus, tabindex of the items should be set to -1 and tabindex of the list should be set to 0.

### MVP
See listbox examples in http://a11y-guidelines.orange.com/web_EN/exemples.html

sophieH29:
ezgif com-video-to-gif

The current implementation of a Selectable list just gives the ability to navigate within a list and update tabindexes appropriately, but still, a selection is not handled when clicking on items or use keys like Enter or Space.
This PR aims to improve list items selection as well as fix the defect https://github.com/stardust-ui/react/issues/464, https://github.com/stardust-ui/react/issues/565

With this PR, it is possible to select items by clicking or pressing Enter or Space. Also ```selectedItemIndex``` is an autocontrolled prop, so the client will be able to set ```defaulSelectedItemIndex``` to specify if some item should be selected in a list by default

Fixes #464 
Fixes #565


docs/src/examples/components/List/Types/ListExampleSelection.tsx
```diff
@@ -1,30 +1,28 @@
...
-  <List selection={selection(knobs)}>
+const ListExampleSelection = () => (
+  <List selection={true}>
     <List.Item
       media={<Image src="public/images/avatar/small/matt.jpg" avatar />}
       header="Irving Kuhic"
       headerMedia="7:26:56 AM"
       content="Program the sensor to the SAS alarm through the haptic SQL card!"
-      selection={selection(knobs)}
+      selection={true}
```
levithomason: nit: the `={true}` portion is not required.  A prop alone means `true`.  We'll add an auto format fix for this but let's remove it from here for now.

sophieH29: yep, right

src/components/List/List.tsx
```diff
@@ -33,6 +33,12 @@ export interface ListProps extends UIComponentProps, ChildrenComponentProps {
   /** A selection list formats list items as possible choices. */
   selection?: boolean
 
+  /** Index of the currently selected item. */
+  selectedItemIndex?: number
```
kuzhelov: just `selectedIndex`?

src/components/List/List.tsx
```diff
@@ -142,9 +161,11 @@ class List extends UIComponent<Extendable<ListProps>, ListState> {
         const ref = React.createRef()
         this.itemRefs[idx] = ref
 
-        maybeSelectableItemProps.tabIndex = idx === selectedItemIndex ? 0 : -1
         maybeSelectableItemProps.ref = ref
         maybeSelectableItemProps.onFocus = () => this.focusHandler.syncFocusedItemIndex(idx)
+        maybeSelectableItemProps.onClick = () => this.trySetState({ selectedItemIndex: idx })
```
kuzhelov: shouldn't it be part of the accessibility behavior?

src/components/List/List.tsx
```diff
@@ -119,7 +129,7 @@ class List extends UIComponent<Extendable<ListProps>, ListState> {
     this.focusHandler = new ContainerFocusHandler(
       () => this.props.items.length,
       index => {
-        this.setState({ selectedItemIndex: index }, () => {
+        this.setState({ focusedItemIndex: index }, () => {
           const targetComponent = this.itemRefs[index] && this.itemRefs[index].current
           const targetDomNode = ReactDOM.findDOMNode(targetComponent) as any
```
kuzhelov: shouldn't `selectedIndex` be updated (synced) as well?

src/components/List/ListItem.tsx
```diff
@@ -79,17 +79,20 @@ class ListItem extends UIComponent<Extendable<ListItemProps>, ListItemState> {
...
+  protected actionHandlers: AccessibilityActionHandlers = {
+    performClick: event => {
+      this.handleClick(event)
+      event.preventDefault()
+    },
   }
 
-  private itemRef = React.createRef<HTMLElement>()
+  handleClick = e => {
+    _.invoke(this.props, 'onClick', e, this.props)
```
kuzhelov, line 99: don't see this prop in the prop types and props interface - seems that it is missed

src/components/List/List.tsx
sophieH29: then ```focusedIndex``` to keep consistency. I am ok with both variants

src/components/List/List.tsx
sophieH29: ```onClick``` handlers are part of component as it's functionality. I have added keydown handlers to accessibility behavior, to trigger this click handler when pressing ```Enter``` or ```Space```

src/components/List/ListItem.tsx
sophieH29: you're right, missed it, thank you!

src/components/List/List.tsx
sophieH29: no need for that, ```ContainerFocusHandler``` is just responsible for moving focus during navigation. When you actually select an item by clicking, it also forces onFocus and updates ```focusedItemIndex```. When you select an item by pressing Enter or Space, it means the focus is already on the element that you're going to select.

kuzhelov (approved):

src/components/List/List.tsx
```diff
@@ -62,22 +69,25 @@ class List extends UIComponent<Extendable<ListProps>, ListState> {
...
   }
 
   static defaultProps = {
     as: 'ul',
     accessibility: listBehavior as Accessibility,
   }
 
+  static autoControlledProps = ['selectedItemIndex']
+  getInitialAutoControlledState() {
+    return { selectedItemIndex: -1, focusedItemIndex: 0 }
```
kuzhelov: currently it initializes more than just auto controlled state (`focusedItemIndex`). we can use the following to initialize these guys:

```jsx
 static defaultProps = {
    as: 'ul',
    selectedItemIndex: -1,
    accessibility: listBehavior as Accessibility,	    accessibility: listBehavior as Accessibility,
  }	  
```
and initialize `state.focusedItemIndex` in constructor

kuzhelov (changes requested):
sorry, one more thing - lets address the question about `getAutocontrolledState()` method before merge

src/components/List/List.tsx
```diff
@@ -131,7 +141,16 @@ class List extends UIComponent<Extendable<ListProps>, ListState> {
 
   renderItems() {
     const { items } = this.props
-    const { selectedItemIndex } = this.state
+    const { focusedItemIndex, selectedItemIndex } = this.state
+
+    const setTabIndex = (idx): number => {
```
kuzhelov: this method actually doesn't set anything, but have `get` semantics. Lets rename it to better reflect the purpose

src/components/List/ListItem.tsx
```diff
@@ -68,28 +73,34 @@ class ListItem extends UIComponent<Extendable<ListItemProps>, ListItemState> {
...
 
-  constructor(props: ListItemProps) {
-    super(props, null)
+  private itemRef = React.createRef<HTMLElement>()
 
-    this.state = {
-      isHovering: false,
-    }
+  protected actionHandlers: AccessibilityActionHandlers = {
+    performClick: event => {
```
kuzhelov, line 92: just `click`? we are not using `perform` anywhere in the actions descriptions so far - because, according to the logic of currently suggested code, we need to rename several actions to maintain consistency:
```jsx
focus -> performFocus
toggle -> performToggle
```

src/components/List/ListItem.tsx
```diff
@@ -32,18 +32,23 @@ export interface ListItemProps extends UIComponentProps, ContentComponentProps<a
 
   /** A list item can indicate that it can be selected. */
   selection?: boolean
```
kuzhelov: where and how this prop is used? it is very confusing to have both `selection` and `selected` prop for the same component, and especially given that both accepts `bool`

sophieH29: this props defines if list is "selectable", what also includes navigable and focusable

sophieH29: do you have any ideas on other naming? maybe just ```selectable``` can bring more meaning for that instead of ```selection```?

src/components/List/ListItem.tsx
sophieH29: we have the same pattern in ```MenuItem```. I think we had this discussion some time ago and agreed on ```performClick``` as it's action description. Agree that we need to keep consistency on it. One of the reasons also why to use ```performClick```, so it wouldn't feel confused like this thing is kind of "on click" handler when use just ```click```. We need to read it like ```this action will call click handler or perform click```

kuzhelov: got it - lets not change naming convention now then. However, in both these cases `click` name is reasonable and goes along with other names used for action handlers

src/components/List/List.tsx
```diff
@@ -131,20 +141,25 @@ class List extends UIComponent<Extendable<ListProps>, ListState> {
...
         const ref = React.createRef()
         this.itemRefs[idx] = ref
 
-        maybeSelectableItemProps.tabIndex = idx === selectedItemIndex ? 0 : -1
         maybeSelectableItemProps.ref = ref
-        maybeSelectableItemProps.onFocus = () => this.focusHandler.syncFocusedItemIndex(idx)
+        maybeSelectableItemProps.onFocus = () => this.focusHandler.syncFocusedIndex(idx)
+        maybeSelectableItemProps.onClick = () => {
+          this.trySetState({ selectedIndex: idx })
+          this.setState({ focusedIndex: idx })
```
kuzhelov: note that at this moment you cannot be sure whether `selectedIndex` had been set to `idx` or not. Consider controlled mode example: when `trySetState` is called, there should be corresponding event fired (`onSelectedIndexChange` - in fact, this is missed from the API at the moment) that will notify client that component would like to change this prop - but it will be up to the client to decide whether its value should be updated, where client may just decide to deny update.

src/components/List/List.tsx
```diff
@@ -59,24 +66,27 @@ class List extends UIComponent<Extendable<ListProps>, ListState> {
     accessibility: PropTypes.func,
     debug: PropTypes.bool,
     items: customPropTypes.collectionShorthand,
-    selection: PropTypes.bool,
+    selectable: PropTypes.bool,
     truncateContent: PropTypes.bool,
     truncateHeader: PropTypes.bool,
+    selectedIndex: PropTypes.number,
```
kuzhelov, line 79: if `selectedIndex` is autocontrolled prop, we also need `onSelectedIndexChange` event

sophieH29: don't see such pattern in```MenuItem```. Do you think it should be part of this PR?

src/components/List/List.tsx
sophieH29: Added ```onSelectedIndexChange```

docs/src/examples/components/List/Types/ListExampleSelectableControlled.shorthand.tsx
```diff
@@ -0,0 +1,48 @@
...
+  render() {
+    return (
+      <List
+        selectable
+        selectedIndex={this.state.selectedItemIndex}
+        onSelectedIndexChange={(e, newProps) => {
+          alert(
+            `List is requested to change its selectedIndex state to "${newProps.selectedIndex}"`,
+          )
+          this.setState({ selectedItemIndex: newProps.selectedIndex })
```
kuzhelov: lets make the name of the `state` property consistent: `selectedIndex`

docs/src/examples/components/List/Types/index.tsx
```diff
@@ -10,9 +10,14 @@ const Types = () => (
...
     <ComponentExample
-      title="Selection"
+      title="Selectable list"
       description="A list can be formatted to indicate that its items can be selected."
-      examplePath="components/List/Types/ListExampleSelection"
+      examplePath="components/List/Types/ListExampleSelectable"
+    />
+    <ComponentExample
+      title="Controlled selectable list"
+      description="A list's 'selectedIndex' state can be controlled by handling 'onSelectedIndexChange' event and passing 'selectedIndex' prop"
```
kuzhelov: as always, would suggest to omit exact prop names from the description:

> List can handle selected index in controlled mode.

src/lib/accessibility/FocusHandling/FocusContainer.ts
```diff
@@ -1,68 +1,68 @@
...
+  private focusedIndex = 0
 
   constructor(private getItemsCount: () => number, private readonly setFocusAt: (number) => void) {}
 
   private noItems = (): boolean => this.getItemsCount() === 0
 
-  private constrainFocusedItemIndex(): void {
-    if (this.focusedItemIndex < 0) {
-      this.focusedItemIndex = 0
+  private constrainfocusedIndex(): void {
```
kuzhelov: nit: camel case is not preserved

src/components/List/ListItem.tsx
```diff
@@ -31,19 +31,24 @@ export interface ListItemProps extends UIComponentProps, ContentComponentProps<a
   media?: any
 
   /** A list item can indicate that it can be selected. */
-  selection?: boolean
+  selectable?: boolean
+  /** Indicates if the current list item is selected. */
```
kuzhelov, line 36: lets introduce empty lines between each prop to improve readability.

src/components/List/ListItem.tsx
```diff
@@ -67,29 +72,35 @@ class ListItem extends UIComponent<Extendable<ListItemProps>, ListItemState> {
...
   }
 
   static defaultProps = {
     as: 'li',
     accessibility: listItemBehavior as Accessibility,
   }
 
-  constructor(props: ListItemProps) {
-    super(props, null)
+  private itemRef = React.createRef<HTMLElement>()
```
kuzhelov: where this guy is used? seems that we should delete it

src/lib/accessibility/FocusHandling/FocusContainer.ts
sophieH29: oh, good catch! thanks!

src/components/List/ListItem.tsx
sophieH29: I don't know this guy either, will delete it

src/components/List/List.tsx
```diff
@@ -131,20 +149,29 @@ class List extends UIComponent<Extendable<ListProps>, ListState> {
...
         maybeSelectableItemProps.ref = ref
-        maybeSelectableItemProps.onFocus = () => this.focusHandler.syncFocusedItemIndex(idx)
+        maybeSelectableItemProps.onFocus = () => this.focusHandler.syncFocusedIndex(idx)
+        maybeSelectableItemProps.onClick = e => {
+          this.trySetState({ selectedIndex: idx })
+          _.invoke(this.props, 'onSelectedIndexChange', e, {
+            ...this.props,
+            ...{ selectedIndex: idx },
+          })
+          this.setState({ focusedIndex: idx })
```
kuzhelov: here we should call `this.focusHandler.syncFocusedIndex(idx)` - this 'setState()' logic will be part of it. Otherwise the problem is that we won't sync focused state with the focusManager

kuzhelov (approved):

PR #567, opened 2018-12-06
fix(docs): fix problem with styles in RTL

kuzhelov:
## NOTE: ongoing work..

Working state is achieved now, with the code of playground quality

The cause of the problem is the fact that Fela was not able to define `rtl-*` classes in the `<head>`'s `<style>` element with previous rendering technique (while these `rtl-*` class names **were** applied to rendered DOM elements).

Currently have distilled everything to key bits of the rendering process - with the fix we should remove all the unnecessary indirection in the code rendering logic.


kuzhelov:
closing as there is a cleaner fix PR () introduced for this problem: #579

PR #569, opened 2018-12-06
fix(Icon): color and borderColor should inherit the color, or be changed via the variables

### LINKED ISSUE #506: Icon does not respect 'error' property, but Text does
### Steps
              <Icon
                name="exclamation"
                size="small"
                circular
                error
                styles={{ paddingTop: "0.6rem" }}
              />
              <Text error content={composeErrorMessage(error)} />

### Expected Result
Expect both icon and text to be red

### Actual Result
image

### Version
x.y.z

### Testcase

mnajdova:
This PR fixes https://github.com/stardust-ui/react/issues/506 following the instructions in the https://github.com/stardust-ui/react/issues/506#issuecomment-443291023
Here are the rules that are applied:
1. If no variables are provided for the `color` and `borderColor`, both are inherited
2. If the `color` variable is provided, then both the **color** of the icon as well as the **border** are respecting that value
3. If the `borderColor` variable is provided then the color of the **border** is respecting this value

In addition to this, the example for the color of the icon is updated, so that we can catch any regression for the behaviors of the colors. It looks like this now:
image


docs/src/examples/components/Icon/Variations/IconExampleColor.shorthand.tsx
```diff
@@ -3,10 +3,42 @@ import { Icon } from '@stardust-ui/react'
 
 const IconExampleColor = () => (
   <div>
-    <Icon name="home" />
-    <Icon name="home" variables={{ color: 'blue' }} />
-    <Icon name="home" variables={{ color: 'red' }} />
-    <Icon name="home" variables={{ color: 'orange' }} />
+    {/* The colors for the icon and the border are inherited. */}
+    <div style={{ color: 'violet', padding: '1rem', display: 'inline-block' }}>
```
layershifter: May be we can use our `Grid` component to avoid styling (`display`, `padding`)?

```jsx
  <Grid columns='repeat(4, auto)'>
    <div />
    <div />
    // others
  </Grid>
```

layershifter: We can also split this to 4 four examples and may be this will better option. For example, https://react.semantic-ui.com/collections/menu/#types-pointing

This section has two examples and we can do the same. This will allow to have description in the title instead of code.

mnajdova: Yep, it makes sense to have four different examples. Will refactor this. Thanks!

docs/src/examples/components/Icon/Variations/index.tsx
```diff
@@ -11,7 +11,7 @@ const Variations = () => (
     />
     <ComponentExample
       title="Color"
-      description="An icon can have a different color"
+      description="An icon is inheriting the color, but can have a different color if provided by the user."
```
kuzhelov: nit:
> ... is inheriting color _by default_, ...

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixes
+- The border color of the Icon is inherited if no value is provided for the `color` and `borderColor` variables @mnajdova ([#569](https://github.com/stardust-ui/react/pull/569))
```
kuzhelov, line 26: please, ensure that this entry will be in a proper section of `CHANGELOG.md` after merging the recent master

kuzhelov (approved):
👍 please, just address the layout issue @layershifter has mentioned before merging

docs/src/examples/components/Icon/Variations/IconExampleColor.shorthand.tsx
mnajdova: We decided offline, that we will use the Grid with description for each of the examples, as introducing another sections will introduce unnecessary additional scrolling on the page.

docs/src/examples/components/Icon/Variations/IconExampleColor.shorthand.tsx
```diff
@@ -1,13 +1,59 @@
...
+      <Icon name="call-video" bordered variables={{ outline: true }} />
+    </div>
+    <Text
+      content={
+        <span>
+          USING THE <code>color</code> VARIABLE:
+        </span>
+      }
+      weight="bold"
+    />
```
levithomason, line 25: In the future, try to minimize the number of lines of code in doc site examples:

```jsx
    <Text
      content={
        <span>
          USING THE <code>color</code> VARIABLE:
        </span>
      }
      weight="bold"
    />
```
```jsx
<Text weight="bold">USING THE <code>color</code> VARIABLE:</Text>
```

PR #550, opened 2018-12-03
fix(mixed): correct types to match propTypes

layershifter:
Fixes #548.

---

[TS sandbox with types](http://www.typescriptlang.org/play/#src=export%20type%20Extendable%3CT%3E%20%3D%20T%20%26%20%7B%0D%0A%20%20%5Bkey%3A%20string%5D%3A%20any%0D%0A%7D%0D%0A%0D%0Aexport%20type%20Nullable%3CT%3E%20%3D%20T%20%7C%20null%0D%0Aexport%20type%20NullableIfUndefined%3CT%3E%20%3D%20T%20extends%20undefined%20%3F%20Nullable%3CT%3E%20%3A%20T%0D%0A%0D%0A%0D%0Aexport%20type%20ReactPropsStrict%3CT%3E%20%3D%20%7B%20%5BK%20in%20keyof%20T%5D%3A%20NullableIfUndefined%3CT%5BK%5D%3E%20%7D%0D%0Aexport%20type%20ReactProps%3CT%3E%20%3D%20Extendable%3CReactPropsStrict%3CT%3E%3E%0D%0A%0D%0A%2F%2F%20---%0D%0A%0D%0Aexport%20const%20testA%3A%20ReactProps%3C%7B%20a%3F%3A%20string%2C%20b%3A%20string%20%7D%3E%20%3D%20%7B%0D%0A%20%20a%3A%20null%2C%0D%0A%20%20b%3A%20null%0D%0A%7D%0D%0A%0D%0Aexport%20const%20testB%3A%20ReactPropsStrict%3C%7B%20a%3F%3A%20string%20%7D%3E%20%3D%20%7B%0D%0A%20%20a%3A%20null%2C%0D%0A%20%20b%3A%20null%0D%0A%7D%0D%0A%0D%0Aexport%20const%20testC%3A%20ReactProps%3C%7B%20a%3F%3A%20string%20%7D%3E%20%3D%20%7B%0D%0A%20%20a%3A%20'string'%2C%0D%0A%20%20b%3A%20'string'%0D%0A%7D%0D%0A), don't forget to enable `strictNullChecks` in options 👍

PR #570, opened 2018-12-06
chore(mixed): fix imports and introduce bundler test

### LINKED ISSUE #547: Impossible to build project with Parcel or Rollup
## Bug Report

The same problem as in #497.

### Steps

1. Prepare project

```bash
git clone https://github.com/rollup/rollup-starter-app
cd rollup-starter-app
yarn
yarn add @stardust-ui/react react react-dom
yarn dev
```

2. Paste code

```js
import React from "react";
import ReactDOM from "react-dom";
import { Button, Provider, themes } from "@stardust-ui/react";

ReactDOM.render(
  React.createElement(
    Provider,
    { theme: themes.teams },
    React.createElement(Button, {
      content: "Theming",
      icon: "arrow right",
      iconPosition: "after",
      primary: true
    })
  ),
  document.getElementById("root")
);

```

### Expected Result

Everything works.

### Actual Result

image

### Version
0.13.0

layershifter:
This PR:
- adds a test for Rollup bundler
- cleanups imports to avoid circular issues
- fixes #547 

---

I also tested a build with Parcel manually, it works like a charm ⭐️


build/gulp/tasks/test-projects.tsx
```diff
@@ -15,12 +15,13 @@ const log = msg => {
...
+export const createPackageFilename = () => tmp.tmpNameSync({ prefix: 'stardust-', postfix: '.tgz' })
 
-const buildAndPackStardust = async (): Promise<string> => {
-  await sh('yarn build:dist')
+export const runIn = path => cmd => sh(`cd ${path} && ${cmd}`)
 
-  return (await sh(`npm pack`, true)).trim()
+export const buildAndPackStardust = async (packageFilename: string) => {
+  await sh('yarn build:dist')
+  await sh(`yarn pack --filename ${packageFilename}`)
```
layershifter, line 26: I decided to use `yarn pack` instead of NPM's. We also passing own name, this allows to generate random names and avoid caching locally.

types/classnames/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+  interface ClassArray extends Array<ClassValue> {}
+
+  interface ClassNamesFn {
+    (...classes: ClassValue[]): string
+  }
+
+  const classNames: ClassNamesFn
+
+  export default classNames
+}
```
layershifter, line 20: We decided to use custom typings for this module.

src/lib/colorUtils.ts
```diff
@@ -1,4 +1,5 @@
-import * as Color from 'color'
+// @ts-ignore
+import Color from 'color'
```
layershifter: This module has the same issue as `classnames` :(

We need to make a decision about this.

src/components/Avatar/Avatar.tsx
```diff
@@ -1,6 +1,8 @@
 import * as PropTypes from 'prop-types'
 import * as React from 'react'
-import { Image, Label, Status } from '../../'
+import Image from '../Image/Image'
+import Label from '../Label/Label'
+import Status from '../Status/Status'
```
layershifter, line 5: This change allows to avoid circular imports.

build/gulp/tasks/test-bundlers.tsx
```diff
@@ -0,0 +1,178 @@
...
+          "createRef",
+          "PureComponent",
+          "Fragment",
+          "Children",
+          "createElement",
+          "isValidElement"
+        ],
+        "node_modules/react-dom/index.js": ["createPortal", "findDOMNode"],
+        "node_modules/react-is/index.js": ["isForwardRef"],
+        "node_modules/what-input/dist/what-input.js": ["ask"]
```
layershifter: https://github.com/rollup/rollup-plugin-commonjs#custom-named-exports

---

Rollup works badly with CJS modules, this config looks very 💩

package.json
```diff
@@ -13,7 +13,7 @@
   "scripts": {
     "build": "gulp build",
     "build:docs": "gulp --series dll build:docs",
-    "build:dist": "gulp --series dll build:dist",
+    "build:dist": "gulp --series build:dist",
```
layershifter, line 16: We don't need `dll` in this task, this change will save some time for us 👍

package.json
```diff
@@ -113,7 +114,7 @@
     "gulp-rename": "^1.3.0",
     "gulp-replace": "^1.0.0",
     "gulp-transform": "^3.0.5",
-    "gulp-typescript": "^5.0.0-alpha.1",
+    "gulp-typescript": "^5.0.0",
```
layershifter, line 116: This is stable now 👍

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - The border color of the Icon is inherited if no value is provided for the `color` and `borderColor` variables @mnajdova ([#569](https://github.com/stardust-ui/react/pull/569))
 - Do not focus `Popup`'s trigger on outside click @sophieH29 ([#578](https://github.com/stardust-ui/react/pull/578))
 - Add `https` protocol to all urls used in the scripts and stylesheets in index.ejs @mnajdova ([#571](https://github.com/stardust-ui/react/pull/571))
+- Fix issue with bundling packagewith Rollup and Parcel @layershifter ([#570](https://github.com/stardust-ui/react/pull/570))
```
miroslavstastny: packagewith - missing space :-)

build/gulp/scaffold/rollup.config.js
```diff
@@ -0,0 +1,110 @@
+import replace from 'rollup-plugin-replace'
+import resolve from 'rollup-plugin-node-resolve'
+import commonjs from 'rollup-plugin-commonjs'
+
+const lodashExports = [
```
miroslavstastny:
```
(!) Missing exports
https://rollupjs.org/guide/en#error-name-is-not-exported-by-module-
node_modules/@stardust-ui/react/dist/es/components/Dropdown/Dropdown.js
pickBy is not exported by node_modules/lodash/lodash.js
```

How is this going to scale? Whenever a new function is used, do I need to remember I have to add it here?

build/gulp/tasks/test-projects.tsx
miroslavstastny: I vaguely remember @kuzhelov used `npm` for a reason...

build/gulp/tasks/test-bundlers.tsx
```diff
@@ -0,0 +1,47 @@
...
+    path.resolve(tmpDirectory, 'app.js'),
+  )
+  fs.copyFileSync(
+    path.resolve(__dirname, '..', 'scaffold', 'rollup.config.js'),
+    path.resolve(tmpDirectory, 'rollup.config.js'),
+  )
+  logger(`✔️Source and bundler's config were created`)
+
+  await runIn(tmpDirectory)(`yarn rollup -c`)
+  logger(`✔️Example project was successfully built: ${tmpDirectory}`)
```
miroslavstastny: It seems we are ignoring rollup warnings. Are we ok with it or would we like to use this as an opportunity to tighten our rules and require warning-free build in rollup? (no warnings are currently allowed in webpack build)

.circleci/config.yml
```diff
@@ -58,6 +58,9 @@ jobs:
 #      - run:
```
kuzhelov: when we are planning to return this? And, actually, what was specific reason to disable these tests - currently we have no _any_ checks that _any_ client is able to consume our library

layershifter: After #550 will be merged

build/gulp/tasks/test-projects.tsx
kuzhelov: the reason for that was just to use the standard tool and not its wrapper, as it could easily entice us to use some features that are `yarn`-specific. However, we may leave it as it is at the moment, no strong opinion here, as this describes set of actions made on our side, not the client's one

build/gulp/scaffold/rollup.config.js
layershifter: Yes, it's a real problem 👎  `lodash` is CJS module, so the there is only one option.
If we will use Babel and `babel-plugin-lodash` this problem will gone.

build/gulp/scaffold/rollup.config.js
```diff
@@ -0,0 +1,110 @@
+import replace from 'rollup-plugin-replace'
```
kuzhelov: this should be placed somewhere under `test-projects` at the very least (we should create corresponding dir) - so that this client-oriented tests functionality will be cohesively structured

.circleci/config.yml
```diff
@@ -58,6 +58,9 @@ jobs:
...
 #          name: Project Tests
 #          command: yarn test:projects:cra-ts
+      - run:
+          name: Bundler Tests
+          command: yarn test:bundlers:rollup
```
kuzhelov: would suggest `test:projects:rollup` - as this won't introduce additional entropy and will still suggest that stack that client is using for her project

build/gulp/scaffold/rollup.config.js
layershifter: It's already placed in the separate dir. I don't want to be so specific for now.

.circleci/config.yml
layershifter: Moved into single entry, `yarn test:projects` 👍

build/gulp/scaffold/rollup.config.js
layershifter: Resolved with `external` and `output.globals`, now is much cleaner

build/gulp/tasks/test-bundlers.tsx
layershifter: I reworked it, now we have a warning whitelist, other warnings will throw.

We can fix all left circular imports in a separate PR because it will require more work.

build/gulp/tasks/test-projects.tsx
layershifter: It looks more cleaner than `(await sh(`npm pack`, true)).trim()`, so I will left `yarn pack` there.

src/lib/colorUtils.ts
layershifter: Done, this package was removed from dependencies

build/gulp/scaffold/rollup.config.js
kuzhelov: yes, but the same reasoning would apply well in case it will be put in _any_ dir - thus, I am not sure that this is a valid argument. 

The reasoning behind my ask is the following: currently, with the dir introduced at the place where you are suggesting, we would have the following project structure:

```
/build/gulp
  /scaffold
  /tasks
       /test-projects
```

So, reasonable question is why `scaffold` directory is placed at this level of folder hierarchy? Is its content shared by any agents/tasks? The answer is "no" - the only place where it is used is `test-projects` module. Thus, I would suggest to put these bits to the module those belong to at the moment.

.circleci/config.yml
```diff
@@ -55,9 +55,9 @@ jobs:
       - run:
           name: Visual Tests
           command: yarn test:visual
-#      - run:
-#          name: Project Tests
-#          command: yarn test:projects:cra-ts
+      - run:
+          name: Project Tests
+          command: yarn test:projects
```
kuzhelov, line 60: 👍

build/gulp/tasks/test-projects.tsx
```diff
@@ -120,6 +125,54 @@ export default App;
...
+  )
+  logger(`✔️Source and bundler's config were created`)
+
+  await runIn(tmpDirectory)(`yarn rollup -c`)
+  logger(`✔️Example project was successfully built: ${tmpDirectory}`)
+})
+
+task(
+  'test:projects',
+  series(
```
kuzhelov, line 169: actually, can we run them in parallel? It seems that there is no shared state between these tasks

layershifter: I will left it as is for now. When this task will be enabled, we will able to test that it works in parallel. However, I almost sure that it will work.

build/gulp/scaffold/rollup.config.js
layershifter: > Is its content shared by any agents/tasks? The answer is "no" - the only place where it is used is test-projects module.

At now, yes. But in the future, I think that we will have other tasks that will should also have files and templates.

mnajdova: I must agree here that better directory structure would reflect exactly what we test even. Having structure, something like:
```
/build/gulp
  /tasks
       /test-projects
         /cra-ts
         /parcel
         ...
```
will be much more cleaner, and it will be obvious what scenarios we got covered so far...

layershifter: @kuzhelov @mnajdova you convinced me and I made these changes 👍

kuzhelov (approved):

PR #571, opened 2018-12-06
fix(docs): EDGE could not open the docs

mnajdova:
Removed crossOrigin='true' from the new added `script` tags in `index.ejs`, because edge has some issues with those, resulting in - **the docs can't be opened in edge**. If there is some solid reason why these were added, please let me know. @levithomason @miroslavstastny @kolaps33

### DIFF
--- a/docs/src/index.ejs
+++ b/docs/src/index.ejs
@@ -6,7 +6,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
   <link rel="shortcut icon" type="image/x-icon" href='<%= __BASENAME__ + 'logo.png' %>'/>
   <link rel="stylesheet"
-        href="//cdnjs.cloudflare.com/ajax/libs/semantic-ui/<%= htmlWebpackPlugin.options.versions.sui %>/semantic.min.css">
+        href="https://cdnjs.cloudflare.com/ajax/libs/semantic-ui/<%= htmlWebpackPlugin.options.versions.sui %>/semantic.min.css">
   <title>Stardust</title>
   <script>
     // Redirect to HTTPs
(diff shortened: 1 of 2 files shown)


layershifter:
We had  a similar issue in SUIR repo, https://github.com/Semantic-Org/Semantic-UI-React/issues/3163. I enforced `https` for scripts: https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/src/components/Document.js#L44

It there any reason why we need `http` on CDN links?

mnajdova:
Thanks for the suggestion @layershifter All script/styles resources urls are now https, and I confirmed that it works on Edge.

kuzhelov (approved):
please, ensure that `CHANGELOG.md` entry will be present before merging. Thank you!

docs/src/index.ejs
```diff
@@ -26,27 +26,24 @@
   </script>
 </head>
 <body>
-  <script src="//cdnjs.cloudflare.com/ajax/libs/anchor-js/3.2.1/anchor.min.js"></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/anchor-js/3.2.1/anchor.min.js"></script>
   <script
-    crossOrigin="true"
```
levithomason, line 31: The `crossorigin` attribute let's us see errors that are thrown in these scripts.  Without it, the error will not be logged in the console and debugging is impossible.

When copied from Semantic UI React's `Document.js` I failed to change the camel cased `crossOrigin` to `crossorigin`.  If this is fixed, is Edge happy?  We'd like to have these properties still.

docs/src/index.ejs
```diff
@@ -26,27 +26,24 @@
   </script>
 </head>
 <body>
-  <script src="//cdnjs.cloudflare.com/ajax/libs/anchor-js/3.2.1/anchor.min.js"></script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/anchor-js/3.2.1/anchor.min.js"></script>
```
levithomason, line 29: ### Why did I normalize all scripts to `//`?
Scripts without the protocol (http / https) allow the browser to request the script using the current protocol.  In dev mode, http, this would allow scripts to be fetched over http which is much faster than https.  This is valid HTML per the specifications as you can see [here](https://stackoverflow.com/questions/550038/is-it-valid-to-replace-http-with-in-a-script-src-http).

### Which should we use?

In fact, it seems I needed to update my best practices. You can see more from Paul Irish [here](https://www.paulirish.com/2010/the-protocol-relative-url/). The short answer is https is encouraged for all sources, it is no longer a perf concern, and protocol-less urls are considered an anti-pattern.

Your changes here are correct.

levithomason (approved):
Let's first try fixing the `crossorigin` casing and testing in Edge before merging.  We really want to be able to read errors raised in CDN scripts during development.

mnajdova:
@levithomason Edge is **not working** with neither of this four combinations:
`crossOrigin="true"`
`crossOrigin="anonymous"`
`crossorigin="true"`
`crossorigin="anonymous"`
The docs is opened only if this attr is removed. I don't see another alternative at this moment other then removing them, although I understand the benefits we would have from using it. My proposal is, until we find another alternative, to just remove them. Thoughts?

miroslavstastny:
@levithomason - I think the casing with capital O is correct - `crossOrigin` in ejs file results in `crossorigin` in html.
That being said, do you have any reference to what `crossorigin="true"` really does? All the [references](https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes) only talk about `anonymous` and `use-credentials` (with a note that *"An invalid keyword and an empty string will be handled as the anonymous keyword."*).
@mnajdova - If you change the urls to `https://` then `crossOrigin="true"` works in Edge. I think that would be the best solution.

mnajdova:
@miroslavstastny that was the only combination I missed yesterday, thanks for pointing that out. Confirmed, it works on edge. Agreed that we should use `anonymous` as it is one of the valid options: https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_settings_attributes

miroslavstastny (approved):

PR #573, opened 2018-12-06
fix(lib): correctly handle arrays in felaSanitizeCssPlugin

### LINKED ISSUE #572: CSS Fallback values do not work
### Steps
```jsx
import React from 'react'
import { Menu, Provider } from '@stardust-ui/react'

const items = [
  { key: 'editorials', content: 'Editorials' },
  { key: 'review', content: 'Reviews' },
]

const MenuExamplePrimary = () => (
  <Provider theme={{
    componentStyles: {
      MenuItem: {
        root: {
          border: '2px solid gray',
          borderColor: ['red', 'blue'],
        }
      }
    }
  }}>
    <Menu defaultActiveIndex={0} items={items} primary />
  </Provider>
)

export default MenuExamplePrimary
```
### Expected Result
`MenuItem::a` has following styles:
```css
border-color: red;
border-color: blue;
```

### Actual Result
image

### Version
0.14.0

miroslavstastny:
Fixes #572 

`color: ['red', 'blue']` is a valid style but `felaSanitizeCssPlugin` converted this to following object:

```js
color: {
  0: 'red',
  1: 'blue'
}
```
which is incorrect. Now the plugin handles arrays correctly.

### DIFF
--- a/src/lib/felaSanitizeCssPlugin.ts
+++ b/src/lib/felaSanitizeCssPlugin.ts
@@ -41,19 +41,19 @@ export default (config?: { skip?: string[] }) => {
   const sanitizeCssStyleObject = styles => {
     const processedStyles = Array.isArray(styles) ? [] : {}
 
-    Object.keys(styles).forEach(cssPropertyName => {
-      const cssPropertyValue = styles[cssPropertyName]
+    Object.keys(styles).forEach(cssPropertyNameOrIndex => {
+      const cssPropertyValue = styles[cssPropertyNameOrIndex]
 
       if (typeof cssPropertyValue === 'object') {
-        processedStyles[cssPropertyName] = sanitizeCssStyleObject(cssPropertyValue)
+        processedStyles[cssPropertyNameOrIndex] = sanitizeCssStyleObject(cssPropertyValue)
         return
       }
 
       const isPropertyToSkip = cssPropertiesToSkip.some(
-        propToExclude => propToExclude === cssPropertyName,
+        propToExclude => propToExclude === cssPropertyNameOrIndex,
       )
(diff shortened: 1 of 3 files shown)


src/lib/felaSanitizeCssPlugin.ts
```diff
@@ -39,7 +39,7 @@ export default (config?: { skip?: string[] }) => {
   const cssPropertiesToSkip = [...((config && config.skip) || [])]
 
   const sanitizeCssStyleObject = styles => {
-    const processedStyles = {}
+    const processedStyles = Array.isArray(styles) ? [] : {}
 
     Object.keys(styles).forEach(cssPropertyName => {
       const cssPropertyValue = styles[cssPropertyName]
```
kuzhelov: nit: `cssPropertyName` -> `cssPropertyNameOrIndex`? Then intend would be easier to capture a week after :)

test/specs/lib/felaSanitizeCssPlugin-test.ts
```diff
@@ -62,4 +62,13 @@ describe('felaSanitizeCssPlugin', () => {
 
     assertCssPropertyValue(`url('../../lib')`, true)
   })
+
+  describe('should pass arrays', () => {
```
kuzhelov: A bit confused with the test name, as it doesn't provide some necessary insights to me. Specifically, here:`should pass array` - pass untouched? Process somehow? If process, then what is the logic that will be applied?

Would expect tests to cover the following logic:
> _if array is passed, then for each item of an array CSS styles sanitization should be applied_

kuzhelov (approved):

PR #574, opened 2018-12-06
feat(Icons): mark Processed Icons moved to Teams theme

kuzhelov:
This PR introduces necessary mapping for Processed Icons that will be exported by Teams theme. This will help to address the following concerns:
- **separate this problem from the core Stardust logic**
- introduce the necessary tracking mechanism for Teams theme

## TODO
- [x] introduce filtering functionality for Processed Icons page
- [x] update changelog

## How it will look like

image

## With Filtering

image

## Review hint
Changes provided in `icons-*` files are only about introducing type correctness - and those are identical for each of these `icons-*`.

### DIFF
--- a/docs/src/prototypes/IconViewer/index.tsx
+++ b/docs/src/prototypes/IconViewer/index.tsx
@@ -3,7 +3,7 @@ import { Provider, Grid, Divider, Header, Icon } from '@stardust-ui/react'
 import themeWithProcessedIcons from 'src/themes/teams/withProcessedIcons'
 import { TeamsProcessedSvgIconSpec } from 'src/themes/teams/components/Icon/svg/types'
 
-import { Menu } from 'semantic-ui-react'
+import { Menu, Segment } from 'semantic-ui-react'
 
 const cellStyles = {
   margin: '10px 0',
@@ -40,7 +40,7 @@ class IconViewerExample extends React.Component<any, {}> {
 
   render() {
     return (
-      <div style={{ margin: '20px' }}>
+      <Segment styles={{ padding: '30px' }}>
         <Header
           as="h3"
           content="Teams Icons"
@@ -51,10 +51,11 @@ class IconViewerExample extends React.Component<any, {}> {
           }}
         />
 
-        <div>
-          <Menu pointing tabular>
+        <div style={{ marginTop: '15px' }}>
+          <Menu tabular style={{ margin: '15px 0' }}>
             {Object.keys(this.iconFilters).map(filterName => (
               <Menu.Item
+                key={filterName}
                 name={filterName}
                 active={this.state.filter === filterName}
                 onClick={() => this.setState({ filter: filterName })}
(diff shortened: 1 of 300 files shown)


docs/src/prototypes/IconViewer/index.tsx
```diff
@@ -21,48 +50,67 @@ class IconViewerExample extends React.Component<any, {}> {
...
-                          <Icon name={name} />
-                          <br />
-                          <code>{name.replace(processedIconsNamePrefix, '')}</code>
-                        </div>
-                      ))}
-                  </Grid>
-                </div>
+
+        <div>
+          <Menu pointing tabular>
```
levithomason: I don't think pointing adds anything to the tabular style.

docs/src/prototypes/IconViewer/index.tsx
```diff
@@ -1,14 +1,43 @@
...
+    )
+  )
+}
+
 class IconViewerExample extends React.Component<any, {}> {
+  private readonly iconFilters = {
+    All: () => true,
+    Exported: (icon: TeamsProcessedSvgIconSpec) => icon.exportedAs,
+    'Not Exported': (icon: TeamsProcessedSvgIconSpec) => !icon.exportedAs,
+  }
```
levithomason, line 30: This object doesn't need to be a class property, it doesn't reference `this` nor the class.

docs/src/prototypes/IconViewer/index.tsx
```diff
@@ -21,48 +50,67 @@ class IconViewerExample extends React.Component<any, {}> {
...
-                      .sort()
-                      .map(name => (
-                        <div key={`${name}-outline`} style={cellStyles}>
-                          <Icon name={name} variables={{ outline: true }} />
-                          <br />
-                          <code>{name.replace(processedIconsNamePrefix, '')} outline</code>
-                        </div>
-                      ))}
-                  </Grid>
+                  <div>
```
levithomason, line 70: Suggest making the root `div` a `Segment` so that the icons appear on a white background and the white menu tabs appear correctly.

docs/src/prototypes/IconViewer/index.tsx
```diff
@@ -21,48 +50,67 @@ class IconViewerExample extends React.Component<any, {}> {
...
+                    <Divider>
+                      <Header as="h3" content="Regular" />
+                    </Divider>
```
levithomason: For simplicity's sake, probably don't need the Divider wrapping the Header here.  The Header alone will cause a line break.

***

Side note, we are missing a "dividing" Header which would accomplish a similar look:
https://react.semantic-ui.com/elements/header/#variations-dividing

image

docs/src/prototypes/IconViewer/index.tsx
kuzhelov: I would see this object still being part of the class, as this is the only place where this one is necessary (note that key strings of the object are specific to this component's view only). Thus, to make the class itself and this object being cohesive, I would rather leave this object here, even if it is a `static` one by its semantics

levithomason: Since this is a subjective judgement call and only prototype code, I'll agree to move us forward.

If it were production code, I would make a case for why you should only implement required code and nothing more.

levithomason (approved):

Issue #576, opened 2018-12-06
Popup: do not focus trigger after outside click


sophieH29:
### Steps
- Open popup
- click outside

### Expected Result
Nothing should be focused. The focus should go to trigger only when Esc button press, or when toggling on a trigger.

### Actual Result
Focus goes to the trigger

PR #578, opened 2018-12-07
fix(Popup): Do not focus trigger on outside click

sophieH29:
Fix for https://github.com/stardust-ui/react/issues/576


mnajdova (approved):
👍 Update the changelog before merging :)

PR #579, opened 2018-12-07
fix(docs): component styles in RTL mode

### LINKED ISSUE #582: RTL mode breaks styles in docs
## Bug Report

When RTL mode is enabled for component example in docs, styles are not applied to rendered component. 

### Steps
- open component docs page of Stardust
- navigate to any component's page (say, `Segment`)
- enable RTL mode for any example

### Expected Result
Styles are properly rendered (preserved) for the rendered component.

### Actual Result
Styles completely disappear for rendered component.

image

image

### Version
0.13.1

kuzhelov:
Currently all the component styles are broken for docs in RTL mode. This PR fixes this problem.

## TODO
- [x] update changelog

## What was the issue

The root cause of that is the fact that Fela updates subscription to the DOM node where styles should be rendered to (i.e `<head> -> <styles>`) **only** on its Provider component being mounted.

It turns out that our rendering loigc hasn't guaranteed that new Fela's `Provider` will be mounted with RTL switch - and, this, the logic of <style> DOM node subscription wasn't fired for RTL renderer.

### Follow-up, general issue: #581

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
+++ b/docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
@@ -55,10 +55,6 @@ const codeTypeApiButtonLabels: { [key in SourceCodeType]: string } = {
 
 const disabledStyle = { opacity: 0.5, pointerEvents: 'none' }
 
-const getComponentExampleElementKey = (examplePath, isRtl) => {
-  return `${examplePath}${isRtl ? '-rtl' : ''}`
-}
-
 /**
  * Renders a `component` and the raw `code` that produced it.
  * Allows toggling the the raw `code` code block.
@@ -254,7 +250,9 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
   }
 
   renderElement = (element: React.ReactElement<any>) => {
+    const { examplePath } = this.props
     const { showRtl, componentVariables, themeName } = this.state
+
     const theme = themes[themeName]
 
     const newTheme: ThemeInput = {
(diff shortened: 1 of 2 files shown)


docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -260,7 +262,11 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
       rtl: showRtl,
     }
 
-    return <Provider theme={newTheme}>{element}</Provider>
+    return (
+      <Provider key={`${examplePath}${showRtl ? '-rtl' : ''}`} theme={newTheme}>
```
levithomason, line 266: Perhaps this should be done in the Provider itself?  It seems adding it here would means RTL will work in our doc examples, but not necessarily anywhere else the Provider is used.

kuzhelov: not sure, as there is no straightforward way to **generate** the key (as we won't like to generate a new one on each rerendering). Should we request client to provide it?

kuzhelov: or you would suggest to do it here?

```jsx
// Provider.ts
...
   <ProviderConsumer
      render={(incomingTheme: ThemePrepared) => {
      const outgoingTheme: ThemePrepared = mergeThemes(incomingTheme, theme)

      return (
          <RendererProvider renderer={outgoingTheme.renderer} {...{ rehydrate: false }}>
             <ThemeProvider key={outgoingTheme.isRtl ? 'rtl' : 'ltr'} theme={outgoingTheme}>{children}</ThemeProvider>
          </RendererProvider>
       )
    }}
   />
```

levithomason (approved):
Just add a TODO and link to this issue to fix it in the Provider

PR #583, opened 2018-12-10
 refactor(lib): remove `createColorVariants` and `setColorLightness` utils

layershifter:
# BREAKING CHANGES

This PR removes `createColorVariants()` and `setColorLightness()` utils. We didn't mention these utils somewhere in docs, but they were exported.

---

We decided to remove them because we want to omit the dependency on the `color` package, refs #570.
We will add a similar functionality to our docs soon.


package.json
```diff
@@ -91,6 +90,7 @@
     "@types/react-is": "^16.5.0",
     "@types/react-router": "^4.0.27",
     "awesome-typescript-loader": "^5.2.1",
+    "color": "^3.1.0",
```
layershifter, line 93: We still use this package in docs so I left it as dev dependency.

miroslavstastny (approved):

Issue #560, opened 2018-12-05
Address dropdown improvement requests


silviuaavram:
This summarizes all points mentioned in the comments untill 1/3/2019:

**Code:**

- [x] rename DropdownLabel to DropdownSelectedItem and make changes as necessary (#725)
- [ ] improve semantics/description of DropdownLabel onClick prop, as it applies only on image/text click.
- [x] refactor setA11yState to use Portal. Follow up with Roman on this.
- [ ] allow getA11ySelectionMessage, itemToString and items (possibly others) to accept React.Elements.
- [x] refactor toggleButton, change the name, make it accept Shorthand.
- [x] cleanup `getA11yStatusMessage` and `A11yStatusMessageOptions` once updated versions get released in Downshift.
- [x] improve the attribute handling of the toggleButton, from Downshift or overrinding attributes better in Stardust.

**Functionality:**

- [x] add clearable prop to replace the toggle icon and which will clear the selected value(s) (#885)
- [x] when Dropdown is closed, focused, has single selection and has selected value, Up/Down Arrows need to open the list at selection position -/+ 1.
- [x] when Dropdown is closed, focused, has single selection and has selected value, click/Spacebar/Enter need to open the list at selected position.
- [ ] when Dropdown is open and has single selection, hitting character key needs to jump highlighted option to the one starting with that key. In case of multiple options, a list with options needs to be generated and looped over.
- [x] when Dropdown is open, hitting End/Home should highlight last/first option.
- [x] when Dropdown is closed and has multiple selection, Left/Right need to move focus between the selected options (to be defined if it should go through 'X' icons, label titles (which will be focusable with user defined action), or both).
- [ ] expose DropdownLabel render callback prop.
- [x] have loader for async operations (#729)
- [x] have toggleButton (toggleIndicator) to appear by default in the examples (#729)
- [ ] have aria-expanded on the triggerButton, if Dropdown is not a search.

**Style:** (#786)

- [x] corner rounding, on combobox corners, when closed, and on combobox upper and listbox lower, when open.
- [x] border bottom from <input> or combobox div adds a strange rounding on the container border bottom, when it's focused.
- [x] style a border around the options list.
- [x] fix issue with list maxHeight. Follow up with Embedd team on this.

bmdalex:
TODO from my side:

### 1. Create better styles by defining and using more variables rather than hardcoding styles:

[...]

PR #584, opened 2018-12-10
feat(Dropdown): Single selection

silviuaavram:
Single Selection Dropdown. Created with a Button instead if an Input.
Behaviours:

1. Button:
- on click, enter, space, should open the list and move focus to it. If item is already selected, that should be the one highlighted after opening (last part not working atm, added as improvement request).
- on up/down, should open the list and move focus to it. If no item is already selected, should focus on first or last item. If item is already selected, focus should be on the selected item position +/- 1. (last part not working atm, added as improvement request).

2. List:
- up/down should move focus through the options.
- home/end should move focus to first/last option (not working atm, improvement request).
- enter should select option and move the focus to the button.
- tab and shift tab should select option and move focus to next/previous element (button).
- click should select option and move focus to the button.
- selecting an option should trigger aria-live with '<option> has been selected.'
- letter key should focus on items starting with that key (not working atm, improvement request).

Navigation with Screen Readers should work as well, with and without virtual cursor.
Button should be made shorthand, in a future task.

Added some improvement comments in https://github.com/stardust-ui/react/issues/560 while working on this.


src/components/Dropdown/Dropdown.tsx
```diff
@@ -247,23 +257,50 @@ export default class Dropdown extends AutoControlledComponent<
...
+
+  private renderTriggerButton(
+    getToggleButtonProps: (options?: GetToggleButtonPropsOptions) => any,
+    styles: ComponentSlotStylesInput,
+  ): JSX.Element {
+    const { placeholder, itemToString } = this.props
+    const { value } = this.state
+    return (
+      <Button
+        content={value ? itemToString(value) : placeholder}
```
mnajdova: I would expect the user to be able to provide some shorthand for this button, maybe even different shorthands for opened and closed toggle button. This, however can be the default button if nothing is provided (I liked the styling and the look of it 👍 )

src/components/Dropdown/Dropdown.tsx
```diff
@@ -205,39 +206,48 @@ export default class Dropdown extends AutoControlledComponent<
...
+                {search
+                  ? this.renderSearchInput(
+                      accessibilityRootPropsRest,
+                      getInputProps,
+                      highlightedIndex,
+                      selectItemAtIndex,
+                      variables,
+                    )
+                  : this.renderTriggerButton(getToggleButtonProps, styles)}
                 {toggleButton && this.renderToggleButton(getToggleButtonProps, styles, isOpen)}
```
mnajdova: If the toggle button is not provided, would expect some closing icon available instead of it (similar to the input clearable variant).

mnajdova (commented):
Some minor comments/opinion. The changes are going in the right direction :)

src/components/Dropdown/Dropdown.tsx
kuzhelov: agree that we would like to have this aspect of dropdown being customizable at the end (not concern of this PR, however)

src/components/Dropdown/Dropdown.tsx
silviuaavram: It is a good idea, and I see Semantic has it as well, as an optional prop. And it also pretty new in their library as well.
Will add it to the improvements list for now, so we can develop it separately from this PR.

src/components/Dropdown/Dropdown.tsx
silviuaavram: yes, I agree, I just wanted to have this as a place to start. Added it to https://github.com/stardust-ui/react/issues/560 for tracking.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -222,48 +229,93 @@ export default class Dropdown extends AutoControlledComponent<
...
-              <div
-                className={classes.containerDiv}
-                onClick={this.handleContainerClick.bind(this, isOpen)}
-              >
-                {multiple && this.renderSelectedItems(styles)}
-                {search &&
-                  this.renderSearchInput(
-                    getRootProps,
-                    getInputProps,
+              <Ref innerRef={innerRef}>
```
silviuaavram, line 248: will leave the dropdown root on this div by having it in this ref, and will get the rest of the accessibility props and put them on the searchInput (that's the combobox). This way Downshift will work as expected and also the combobox props (that Downshift provides by the same getRootProps) will be applied to the searchInput.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -308,19 +361,44 @@ export default class Dropdown extends AutoControlledComponent<
...
+    selectItemAtIndex: (index: number) => void,
+    getMenuProps: (options?: GetMenuPropsOptions, otherOptions?: GetPropsCommonOptions) => any,
+    getItemProps: (options: GetItemPropsOptions<ShorthandValue>) => any,
+    getInputProps: (options?: GetInputPropsOptions) => any,
   ) {
-    const accessibilityMenuProps = getMenuProps({ refKey: 'innerRef' })
+    const accessibilityMenuProps = {
+      ...getMenuProps({ refKey: 'innerRef' }, { suppressRefError: true }),
+    }
+    if (!this.props.search) {
```
silviuaavram: if we have a non-search, the keyboard handles, aria-activedescendant and some other stuff that was on the SearchInput now goes on the List. Also there were some custom handlers added on the List, in `this.handleListKeyDown`. Needed these custom handlers because Downshift does not support non-search by default.

src/components/Dropdown/DropdownSearchInput.tsx
```diff
@@ -98,29 +97,26 @@ class DropdownSearchInput extends UIComponent<Extendable<DropdownSearchInputProp
   }
 
   public renderComponent({ rest, styles }: RenderResultConfig<DropdownSearchInputProps>) {
-    const { accessibilityWrapperProps, accessibilityInputProps, placeholder } = this.props
-    const { innerRef, ...accessibilityWrapperPropsRest } = accessibilityWrapperProps
+    const { accessibilityComboboxProps, accessibilityInputProps, placeholder } = this.props
     return (
-      <Ref innerRef={innerRef}>
```
silviuaavram, line 104: Won't need ref for this anymore, since we will use the container `<div>`

src/components/Dropdown/Dropdown.tsx
```diff
@@ -222,48 +229,93 @@ export default class Dropdown extends AutoControlledComponent<
...
+  ): JSX.Element {
+    const { placeholder, itemToString, multiple } = this.props
+    const { value } = this.state
+    const content = multiple ? placeholder : value ? itemToString(value) : placeholder
+    return (
+      <Ref
+        innerRef={buttonNode => {
+          this.buttonNode = buttonNode
+        }}
+      >
```
layershifter: Please use  `createRef()` API: https://reactjs.org/docs/refs-and-the-dom.html#creating-refs
This is more simple and effective.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -222,48 +229,93 @@ export default class Dropdown extends AutoControlledComponent<
...
     )
   }
 
+  private renderTriggerButton(
+    styles: ComponentSlotStylesInput,
+    getToggleButtonProps: (options?: GetToggleButtonPropsOptions) => any,
+  ): JSX.Element {
+    const { placeholder, itemToString, multiple } = this.props
+    const { value } = this.state
+    const content = multiple ? placeholder : value ? itemToString(value) : placeholder
```
layershifter: 💣 

Can we simplify this somehow? May be to a separate function.

```jsx
const computeContentForButton = ({ placeholder, itemToString, multiple }: Props, { value }: State) => {
  if (multiple) return placeholder 
  return value ? itemToString(value) : placeholder
}
```

src/components/Dropdown/Dropdown.tsx
silviuaavram: Then I will get a ref object and it will point to my Button component, that cannot be focused.
If I need to do more magic to access the DOM button, I would rather leave it like this.

layershifter: Why? It does the same thing as callback.

```diff
-private buttonNode: HTMLElement
+private buttonRef = React.createRef<HTMLElement>()
```

```diff
-      <Ref
-        innerRef={buttonNode => {
-          this.buttonNode = buttonNode
-        }}
-      >
+<Ref innerRef={this.buttonRef}>
```

```diff
-this.buttonNode.focus()
+this.buttonNode.current.focus()
```

This is more clear and simple. And allows to avoid inline functions, BTW: https://github.com/palantir/tslint-react (`jsx-no-lambda`)

layershifter: BTW, we a have a decision to name refs to as ***Ref**: https://github.com/stardust-ui/react/pull/459#discussion_r233852203

silviuaavram: I was using it as in the React docs, `ref={this.buttonRef}`. Will refactor, thanks.

src/components/Dropdown/Dropdown.tsx
silviuaavram: is it hard to read? creating a function that will be used once? Inside this method? Destructuring again? is it really worth it?

layershifter: Yep, it's hard to read because it requires more attention. You can extract this to a method or into top level function.

https://eslint.org/docs/rules/no-nested-ternary
https://github.com/airbnb/javascript/blob/eslint-config-airbnb-v17.1.0/packages/eslint-config-airbnb-base/rules/style.js#L323

BTW, this rule will be enabled in #600.

src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -3,16 +3,18 @@ import { DropdownProps } from '../../../../components/Dropdown/Dropdown'
 import { DropdownVariables } from './dropdownVariables'
 
 const dropdownStyles: ComponentSlotStylesInput<DropdownProps, DropdownVariables> = {
-  containerDiv: ({
+  root: (): ICSSInJSStyle => ({}),
+
+  container: ({
```
miroslavstastny, line 9: Toggle button is not vertically centered:
image

src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -22,21 +24,44 @@ const dropdownStyles: ComponentSlotStylesInput<DropdownProps, DropdownVariables>
...
     }),
     ...(focused && {
-      borderColor: containerDivFocusBorderColor,
-      borderRadius: containerDivFocusBorderRadius,
+      borderColor: borderColorFocus,
+      borderRadius: borderRadiusFocus,
     }),
   }),
 
+  button: ({ variables: { comboboxPaddingButton } }): ICSSInJSStyle => {
```
miroslavstastny, line 39: Outline is displayed around triggerButton and the list even if mouse is used:
image
image

sophieH29:
Found an issue, not sure if it is an expected behavior:
- TAB to toggle button 
- Expand list
- navigate to some item with arrow key
- Press TAB
- Focus goes to next button
- List is collapsed

Actual: the item from which you TAB from is selected
Expected: selected item should not change when TAB (Shift + TAB) from list

Also, is there any possibility to select the initial value, in your example ```Select your hero?```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -222,48 +229,96 @@ export default class Dropdown extends AutoControlledComponent<
...
+  ): JSX.Element {
+    const { placeholder, itemToString, multiple } = this.props
+    const { value } = this.state
+    let content
+
+    if (multiple) {
+      content = placeholder
+    } else {
+      content = value ? itemToString(value) : placeholder
+    }
```
sophieH29: can be simplified a bit ```const content = value && !multiple ? itemToString(value) : placeholder```

src/components/Dropdown/Dropdown.tsx
```diff
@@ -222,48 +229,96 @@ export default class Dropdown extends AutoControlledComponent<
...
+          fluid
+          styles={styles.button}
+          {...getToggleButtonProps({
+            onFocus: () => {
+              this.setState({ focused: true })
+            },
+            onBlur: () => {
+              this.setState({ focused: false })
+            },
+            'aria-label': content, // TODO: add this to behaviour
```
sophieH29: It seems for me that ```aria-label``` is redundant here. We pass the same content to Button, screen readers can cope with that and read content inside a button element.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -298,7 +354,7 @@ export default class Dropdown extends AutoControlledComponent<
       <Icon
         name={`chevron ${isOpen ? 'up' : 'down'}`}
         as="button"
-        tabIndex="-1"
+        tabIndex="0"
```
sophieH29: no need to add ```tabindex``` here. Icon is rendered as ```<button>```, so it will get focusable behavior out of the box.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -308,19 +364,44 @@ export default class Dropdown extends AutoControlledComponent<
...
+        innerRef={(listNode: HTMLElement) => {
+          this.listNode = listNode
+          innerRef(listNode)
+        }}
+      >
         <List
           {...accessibilityMenuPropsRest}
           styles={styles.list}
+          tabIndex={-1}
           aria-hidden={!isOpen}
```
sophieH29, line 400: I am thinking how we can move such accessibility things to behaviors. At least for configuring attributes, also based on props

src/components/Dropdown/Dropdown.tsx
```diff
@@ -509,7 +592,7 @@ export default class Dropdown extends AutoControlledComponent<
       e: React.SyntheticEvent,
       searchInputProps: DropdownSearchInputProps,
     ) => {
-      if (keyboardKey.getCode(e) === keyboardKey.Tab && highlightedIndex !== undefined) {
+      if (keyboardKey.getCode(e) === keyboardKey.Tab && _.isNil(highlightedIndex)) {
```
sophieH29: shouldn't be inverted condition ```!_.isNil(highlightedIndex)```?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -298,7 +354,7 @@ export default class Dropdown extends AutoControlledComponent<
       <Icon
         name={`chevron ${isOpen ? 'up' : 'down'}`}
```
mnajdova: I am seeing this now... The available names of the icons are defined in the theme. If the theme using the Dropdown doesn't have these icons, nothing will be displayed. I would propose using some unicode characters for this (we already have something in the accordion for this). My bad that in the submenu I was using the '>' character instead of this, but I really didn't like that on windows the right arrow looks like this: 
image
The **SMALL TRIANGLE-*** characters should be sufficient for all combination that we have across the components (see the list here: https://www.w3schools.com/charsets/ref_utf_geometric.asp). Of course later, we will support the user to have some icon shorthand instead of those.

src/themes/teams/components/Dropdown/dropdownStyles.ts
silviuaavram: it is, the issue was with the heights. Will correct it by having 32px on toggle button and making the button 30px. The other 2px difference is the div container bottom border, that will give the underlined effect on focus.

src/components/Dropdown/Dropdown.tsx
mnajdova: Please see PR: https://github.com/stardust-ui/react/pull/673 for more information.

src/themes/teams/components/Dropdown/dropdownStyles.ts
silviuaavram: 1. will remove the toggle button from tab order and will make it clickable only when it's a search.
2. the outline style should be replaced at some point with a style from designers. The focus on the list will be kept, as screen readers need it to narrate options.

src/components/Dropdown/Dropdown.tsx
silviuaavram: let's not start this again, haha
screenshot 2019-01-02 at 17 24 08

src/components/Dropdown/Dropdown.tsx
silviuaavram: removing it, thanks.

src/components/Dropdown/Dropdown.tsx
silviuaavram: removing it, although I had some issues at some point that made me add tabIndex.

anyway, will need to refactor this, and probably will add only a click handler to the icon, as I looked over toggle buttons in dropdowns a bit more and none of them are tabbable, only clickable.

src/components/Dropdown/Dropdown.tsx
silviuaavram: we may need to have a meeting at some point, as this is one of many things accessibility related that we need to encapsulate.

src/components/Dropdown/Dropdown.tsx
silviuaavram: thanks good catch! that moment where I thought I will improve something by refactoring and break it instead.

src/components/Dropdown/Dropdown.tsx
sophieH29: 😄I see. However, I simplified it to one ternary condition 😉

src/components/Dropdown/Dropdown.tsx
silviuaavram: used unicode chars. thanks!

docs/src/examples/components/Dropdown/Types/DropdownExampleSingleSelection.shorthand.tsx
```diff
@@ -0,0 +1,52 @@
...
+        getA11yStatusMessage={getA11yStatusMessage}
+        getA11ySelectionMessage={{
+          onAdd: item => `${item} has been selected.`,
+        }}
+        placeholder="Select your hero"
+        items={inputItems}
+      />
+    )
+  }
+}
```
layershifter: Can be functional component as I see

src/components/Dropdown/Dropdown.tsx
```diff
@@ -224,50 +234,97 @@ export default class Dropdown extends AutoControlledComponent<
...
             isOpen,
+            toggleMenu,
             highlightedIndex,
             selectItemAtIndex,
           }) => {
+            const accessibilityRootProps = getRootProps(
+              { refKey: 'innerRef' },
+              { suppressRefError: true },
+            )
+            const { innerRef, ...accessibilityRootPropsRest } = accessibilityRootProps
```
layershifter:
```js
const { innerRef, ...accessibilityRootProps } = getRootProps(
  { refKey: 'innerRef' },
  { suppressRefError: true },
)
```

Can be just this as I see

src/components/Dropdown/Dropdown.tsx
```diff
@@ -285,44 +344,66 @@ export default class Dropdown extends AutoControlledComponent<
...
+    toggleMenu: () => void,
+    selectItemAtIndex: (index: number) => void,
+    getMenuProps: (options?: GetMenuPropsOptions, otherOptions?: GetPropsCommonOptions) => any,
+    getItemProps: (options: GetItemPropsOptions<ShorthandValue>) => any,
+    getInputProps: (options?: GetInputPropsOptions) => any,
   ) {
-    const accessibilityMenuProps = getMenuProps({ refKey: 'innerRef' })
+    const accessibilityMenuProps = {
+      ...getMenuProps({ refKey: 'innerRef' }, { suppressRefError: true }),
+    }
```
layershifter:
```suggestion
 const accessibilityMenuProps = getMenuProps({ refKey: 'innerRef' }, { suppressRefError: true })
```

Destruction doesn't make any sense there

layershifter:
### 1. `toggleButton` is selectable

image

I'm not sure that it's an incorrect behavior, but looks strange.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -414,30 +484,38 @@ export default class Dropdown extends AutoControlledComponent<
...
-      search
-        ? _.isFunction(search)
-          ? search(itemsMatchSearchQuery, searchQuery)
-          : itemToString(item)
-              .toLowerCase()
-              .indexOf(searchQuery.toLowerCase()) !== -1
-        : true,
-    )
+    if (multiple) {
+      filteredItems = filteredItems.filter(item => (value as ShorthandValue[]).indexOf(item) === -1)
```
layershifter:
```suggestion
      _.pullAll(filteredItems, value)
```

Feel free to use `lodash` 👍 
https://lodash.com/docs/4.17.11#pullAll

src/components/Dropdown/Dropdown.tsx
silviuaavram: agreed, thanks!

silviuaavram:
> ### 1. `toggleButton` is selectable
> image
> 
> I'm not sure that it's an incorrect behavior, but looks strange.

because of the unicode char. it happens in accordion arrows as well, and it happened since forever. it is strange, but no idea if we can do anything about that now.

src/components/Dropdown/Dropdown.tsx
silviuaavram: doing that will mutate `filteredItems`, and also `items`, which is not what I want to do. to use `pullAll` I need to first create a deep copy of the array and use `pullAll` on that copy. unless you think of something better :)

docs/src/examples/components/Dropdown/Types/DropdownExampleSingleSelection.shorthand.tsx
silviuaavram: I updated all the examples to functional components. thanks!

src/components/Dropdown/Dropdown.tsx
layershifter: You can use `_.difference()` in this case, it will return a new array

src/components/Dropdown/Dropdown.tsx
silviuaavram: it wasn't redundant, I used it to override the one from Downshift's `getToggleButtonProps`. will leave it there :)

src/components/Dropdown/Dropdown.tsx
```diff
@@ -224,50 +236,91 @@ export default class Dropdown extends AutoControlledComponent<
...
-              >
-                {multiple && this.renderSelectedItems(styles)}
-                {search &&
-                  this.renderSearchInput(
-                    getRootProps,
-                    getInputProps,
+              <Ref innerRef={innerRef}>
+                <div
+                  className={classes.container}
+                  onClick={multiple ? this.handleContainerClick.bind(this, isOpen) : undefined}
```
mnajdova, line 251: Why are we binding this? The `handleContainerClick` is already an arrow function. Am I missing something?

silviuaavram: probably a leftover, from a ancient time when it wasn't an arrow function. it skipped me, will change, thanks for pointing it out!

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `color` prop to `Icon` component @Bugaa92 ([#651](https://github.com/stardust-ui/react/pull/651))
 - Create a `base` theme with Text component example @almedint ([#618](https://github.com/stardust-ui/react/pull/618))
 - Adding attachment behavior to handle space/enter key @kolaps33 ([#375](https://github.com/stardust-ui/react/pull/375))
+- Add Dropdown Single Selection variant @silviuavram ([#584](https://github.com/stardust-ui/react/pull/584))
```
mnajdova: Watch out after merging the master! This part should be in the unreleased section.

docs/src/examples/components/Dropdown/Types/DropdownExampleMultipleSearch.shorthand.tsx
```diff
@@ -13,21 +13,17 @@ const inputItems = [
...
+  <Dropdown
+    multiple
+    search
+    getA11ySelectionMessage={getA11ySelectionMessage}
+    getA11yStatusMessage={getA11yStatusMessage}
+    noResultsMessage="We couldn't find any matches."
+    placeholder="Start typing a name"
+    items={inputItems}
+  />
+)
```
mnajdova, line 26: 👍

mnajdova (approved):
👍  all issues are tracked, so we are save to merge this one!

docs/src/examples/components/Dropdown/Types/index.tsx
```diff
@@ -4,6 +4,11 @@ import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
 
 const Types = () => (
   <ExampleSection title="Types">
+    <ComponentExample
+      title="Single Selection"
+      description="A dropdown with single selection."
+      examplePath="components/Dropdown/Types/DropdownExampleSingleSelection.shorthand"
```
layershifter:
```suggestion
      examplePath="components/Dropdown/Types/DropdownExampleSingleSelection"
```

We don't need `.shorthand` there

silviuaavram: fixed, thanks!

layershifter (approved):

PR #587, opened 2018-12-10
chore(docgen): remove `constructorName` property

layershifter:
Cut from #509.

---

This PR removes the `constructorName` property from our docgen task. This property is used only in our tests and blocks the usage of HOC, refs #509.

### Props
- we omit `require(absPath).default` in docgen task, this require is handled by `ts-node` so it will increase performance (even in CI 1:30 -> 1:12 for the `Unit Tests` task)
- unblocks the usage of HOCs in our components

### Cons
- `isConformant()` now has one more argument. However it's not a real issue. BTW, `componentName` will be compared with `displayName`, so there is a really small change to get a typo

### DIFF
--- a/build/gulp/plugins/util/getComponentInfo.ts
+++ b/build/gulp/plugins/util/getComponentInfo.ts
@@ -46,10 +46,6 @@ const getComponentInfo = (filepath: string, checksum?: string) => {
   // add checksum
   info.checksum = checksum
 
-  // add exported Component info
-  const Component = require(absPath).default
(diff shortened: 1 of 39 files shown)


levithomason:
For now, we'll close this PR.  We won't wrap all components with the forwardRef HOC just yet.  We're going to let the community figure out what the final solution is on refs before doing anything drastic.

layershifter:
But we want still to get rid of this `require()`, #589.

PR #590, opened 2018-12-10
fix(Dropdown): use pxToRem from Teams theme

kuzhelov:
This is a leftover from the dropdown PR - essentially, this is about right `pxToRem` tool being used used (i.e. the one that is aware of theme's font size).

Note that this source of confusion is temporary, as necessary changes to pxToRem mechanism are about to come.

### TODO
- [x] update CHANGELOG.md


silviuaavram (approved):
I also tested the dropdown with these changes while working for single selection, and confirm it fixed.

layershifter (approved):
It's a valid fix, however I should mention that we met the issue from my comment, https://github.com/stardust-ui/react/pull/371#issuecomment-444827263. We have two `pxToRem()` and it's hard to understand what util should be used.

kuzhelov:
@layershifter, agree - but, as have mentioned, this cause for this issue should go away this week

PR #592, opened 2018-12-11
fix(Popup): restore ability to render popup from React element

kuzhelov:
## TODO
- [x] update changelog

------

With recent changes being introduced to shorthand factories (specifically, because of the change that now dictates to pass React Element shorthand value 'as is'), the following way of defining `Popup` has become broken:

```jsx
<Popup content={<p>This is the content that will be rendered 'as is'</p>} ... />
```

The way it was broken is that now the content of the `<p>` element is, essentially, not visible (however, while still rendered in the DOM tree), because positioning styles are not applied.

----

This corresponds to the following example on the docs page:

image

----

This PR restores this functionality.

### DIFF
--- a/src/components/Popup/Popup.tsx
+++ b/src/components/Popup/Popup.tsx
@@ -280,7 +280,7 @@ export default class Popup extends AutoControlledComponent<Extendable<PopupProps
 
     const popupContent = React.isValidElement(content)
       ? React.cloneElement(content, {
-          ...content.props,
+React.cloneElement(content, popupContentAttributes)
           ...popupContentAttributes,
         })
       : Popup.Content.create(content, {
--- a/src/components/Popup/Popup.tsx
+++ b/src/components/Popup/Popup.tsx
@@ -12,6 +12,7 @@ import {
   isBrowser,
   ChildrenComponentProps,
   ContentComponentProps,
+  StyledComponentProps,
   commonPropTypes,
 } from '../../lib'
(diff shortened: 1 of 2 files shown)


src/components/Popup/Popup.tsx
```diff
@@ -266,16 +269,23 @@ export default class Popup extends AutoControlledComponent<Extendable<PopupProps
...
-    })
+    /**
+     * if there is no focus trap wrapper, we should apply
+     * HTML attributes and positioning to popup content directly
+     */
+    const popupContentAttributes = accessibility.focusTrap ? {} : popupWrapperAttributes
+
+    const popupContent = React.isValidElement(content)
+      ? React.cloneElement(content, {
+          ...content.props,
```
layershifter: Why we need a spread there? React will add missing properties: https://codesandbox.io/s/9ornj4qrko

```suggestion
React.cloneElement(content, popupContentAttributes)
```

layershifter (changes requested):
This issue shows that our idea from #496 may be not so good.

The factories change is not released yet, so I will block this PR for additional discussion on sync.

layershifter (approved):
Unblocked, let's go.

kuzhelov:
@layershifter 

As we've agreed with @layershifter while having discussion around that, moving with this now but leaving this topic for future discussion around shorthand factories, as this change might serve as a warning sign for the approach we've taken, especially if this pattern 

```
isValidElement() ? clone() : create()   // © @layershifter :)
```

will be repeated for other components in many places.

PR #595, opened 2018-12-11
fix(getKeyDownHandler): Pass props to client's "onKeyDown" handler

sophieH29:


src/lib/getKeyDownHandlers.ts
```diff
@@ -42,7 +42,7 @@ const getKeyDownHandlers = (
           eventHandler && eventHandler(event)
```
mnajdova, line 42: Do we want to add the props to the eventHandler as well? We can define then the actions in the following manner:
```
    performClick: (event, props) => this.handleClick(event, props),
  }
```
Not sure, if the previous definition with the event is necessary:
```
    performClick: event => this.handleClick(event),
  }
```
because the event is send in the getKeyDownHandler, I don't see why we cannot send the props too. Let me know if this works and make sense (it makes sense in my opinion, as we would have unified way of defining the handlers).

mnajdova (commented):
Please take a look into my comment. It would be great if we can add this too.

src/lib/getKeyDownHandlers.ts
mnajdova: Disregard my comment, I was mixing it with other problems. The changes for this PR are sufficient 👍

mnajdova (approved):

PR #596, opened 2018-12-11
fix(Menu): correctly handle `isFromKeyboard`

### LINKED ISSUE #434: "what-input" library doesn't work correctly when pressing key for first time
### Steps

1. Have a menu with three items,
2. click on the third item,
3. hit left arrow,
4. hit left arrow again.

### Expected Result
- The second item selected and highlighted after step 3.
- The first item selected and highlighted after step 4.

### Actual Result
After step 3, the second item is correctly selected, but it is not highlighted 💣

menu_keyboard

### Version
Current master

### The root cause
Scenario:
- ```onKeyDown``` handler is called on a ```FocusZone```
- inside this handler, there is a quite big logic, which at the end finds a particular child ```MenuItem``` and focuses it
- ```onFocus``` handler of ```MenuItem``` is called where it is checked if the ```whatInput.ask() === 'keyboard'```, we will get here ```false```
- and only afterward the ```window.addEventListener('keydown')``` from ```what-input``` lib is called and update a state to ```keyboard``` but it's too late.

#### With styles
```what-input``` library adds attribute ```data-whatinput``` to html with updated value of what was the last input (for e.g. keyboard, mouse, etc)
So in styles we use it as
```js
[data-whatintent='mouse'] *:focus {
  outline: none;
}
```

**Pros:**
- it we solve all possible issues with handling it with scripts and we'll get rid of re-rendering component when we focus ```MenuItem```

**Cons:**
- Currently Fela doesn't allow such global selectors used in particular component

#### Update state of ```what-input``` manually

Update what-input state by just calling its function, so in FocusZone it will look like:
```js
 private _onKeyDown = (ev: React.KeyboardEvent<HTMLElement>): boolean | undefined => {
    ...
    // call it beforehand and update state from 'mouse' to 'keyboard'
    whatInput.setInput(ev)
   ...
```

**Pros:**
- Full and transperent control of state

**Cons:**
- ```what-input``` currently does not expose function for triggering a state change manually. I'll create an issue in their repo, but we can't be assure they'll approve. Meanwhile, we can add this library to Stardust, as there is no a lot of code and also remove parts that we're not using.

layershifter:
Fixes #434.

---

I investigated the issue a bit more and it comes from our side. We have own event listener for `onFocus` and it has the higher priority than a handler from `what-input`. This means that our handlers will always take an unactual value from `whatInput.ask()`.

I also tried to play around with a wrapper component (via [Custom Callbacks](https://github.com/ten1seven/what-input#custom-callbacks)) that will handle this properly, but didn't reach a good and simple result. `isFromKeyboard` is respected only when component is focused, this means that we need separately handle focus and blur events and also events from `what-input`.

---

As were suggested, I made a cleanup there:
- copied library and removed unused functions
- add `useCapture` in `addEventListener`
- drop `what-input` dependency
- removed `isFromKeyboard.ts` as util
- `typescriptUtils.ts` also removed because they were used only by `isFromKeyboard.ts`

---

fix

# TODO

- [ ] update CHANGELOG


src/lib/whatInput.ts
```diff
@@ -0,0 +1,427 @@
+// Taken from https://github.com/ten1seven/what-input/blob/master/src/scripts/what-input.js
```
sophieH29: If we are copying the whole library, I think might be worth to optimize it a bit, and get rid of things that we are not using. For example, I am not sure about ```data-whatintent```, as we are reading value from ```data-whatinput```.  Also additional props like ```ignoreKeys``` and so on are not used and can be removed. 

Basically, what we need it's just to get info if the last input was 'keyboard'. So I suggest to leave everything that is related to it and remove anything else. 
``` touchstart``` and ```touchend``` are triggered only for mobile, so we might not need it also as input won't be changed to 'keyboard' on mobile

Comment (layershifter) [reply src/lib/whatInput.ts]: Agree, but this will mean that we want to adopt it. I'm not sure that actually need to do this.

Comment (layershifter) [reply src/lib/whatInput.ts]: We also don't need `fireFunctions` and many other stuff, but as I see: we need to fix our bug and just wait for this function in the module.

And may be, a more better solution that matches React's ideas. Because even with `setInput()` it will still be a monkey patch.

Comment (sophieH29) [reply src/lib/whatInput.ts]: we can start with removing unused things (because it's easy and fast :)) as waiting for this thing to be fixed in a module takes an unknown time.
Reinventing something new, more in React way, might be considered for some future RFC, but at least we need to get back from authors that they won't fix it.

Comment (layershifter) [reply src/lib/whatInput.ts]: Please understand my position there, we have two options.

*Option 1.* We want to get an enchantment in the package and we will temporary copy its code as is. We know that this code is stable and working, we should not touch it at all.

*Option 2.* We fork the package and adopt it for our needs. In this case I of course will remove all unused options.

Comment (levithomason) [reply src/lib/whatInput.ts]: Option 3: Let's try deleting as much of this as possible.  The `isFromKeyboard` abstraction currently has lots of code that is unnecessary. I'm also not sure if we need this utility or if it can be done in a more simple way.

Comment (layershifter) [reply src/lib/whatInput.ts]: I reworked this as you suggested guys.

- drop `what-input`
- removed `isFromKeyboard.ts` with util
- `typescriptUtils.ts` also removed
- removed unused functions and `intent` in `whatInput.ts`

For now, we have only a `isFromKeyboard()` function. Issue with events priority was resolved by `useCapture`.

---

However, I would be glad if matching changes were done in third party dependency and will be able to remove `whatInput.ts` from our code.

File: src/components/RadioGroup/RadioGroupItem.tsx [hunk 2]
```diff
@@ -137,7 +137,7 @@ class RadioGroupItem extends AutoControlledComponent<
     accessibility: radioGroupItemBehavior as Accessibility,
   }
 
-  static autoControlledProps = ['checked', isFromKeyboard.propertyName]
+  static autoControlledProps = ['checked', 'isFromKeyboard']
```
miroslavstastny: Does it make sense for `isFromKeyboard` to be autocontrolled?

src/components/RadioGroup/RadioGroupItem.tsx
```diff
@@ -180,12 +180,12 @@ class RadioGroupItem extends AutoControlledComponent<
   }
 
   private handleFocus = (e: React.SyntheticEvent) => {
-    this.setState(isFromKeyboard.state())
+    this.setState({ isFromKeyboard: isFromKeyboard() })
```
miroslavstastny, line 174: And if it is autocontrolled, you should use `trySetState()`

src/components/RadioGroup/RadioGroupItem.tsx
layershifter: This looks really strange to me, agree. Will remove this.

kuzhelov (approved):

PR #597, opened 2018-12-11
feat(text): color prop

bmdalex:
### Description

This PR:
- adds `color` prop to `Text` component;
- creates generic types for the theme `color` story;
- creates `Text` color examples;
- cleans up `Divider` style files after #558 

### API

```jsx
<Text color={'primary' | 'secondary' | 'blue' | 'green' | 'grey'
 | 'orange' | 'pink' | 'purple' | 'teal' | 'red' | 'yellow' | string} .../>
```
screenshot 2018-12-11 at 18 15 04

### DIFF
--- a/src/lib/colorUtils.ts
+++ b/src/lib/colorUtils.ts
@@ -3,7 +3,7 @@ import { SiteVariablesInput, ColorVariants, ColorValues } from '../themes/types'
 
 export const mapColorsToScheme = <T>(
   siteVars: SiteVariablesInput,
-  mapper: string | number | keyof ColorVariants | ((color: ColorVariants) => T),
+  mapper: keyof ColorVariants | ((color: ColorVariants) => T),
 ): ColorValues<T> =>
   _.mapValues(
     { ...siteVars.emphasisColors, ...siteVars.naturalColors },
--- a/src/lib/colorUtils.ts
+++ b/src/lib/colorUtils.ts
@@ -1,0 +1,11 @@
+import * as _ from 'lodash'
+import { SiteVariablesInput, ColorVariants, ColorValues } from '../themes/types'
+
+export const mapColorsToScheme = <T>(
+  siteVars: SiteVariablesInput,
+  mapper: string | number | keyof ColorVariants | ((color: ColorVariants) => T),
+): ColorValues<T> =>
+  _.mapValues(
+    { ...siteVars.emphasisColors, ...siteVars.naturalColors },
+    typeof mapper === 'number' ? String(mapper) : (mapper as any),
+  ) as ColorValues<T>
--- a/src/themes/teams/components/Divider/dividerStyles.ts
+++ b/src/themes/teams/components/Divider/dividerStyles.ts
@@ -6,18 +6,22 @@ import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
 import { DividerPropsWithDefaults } from '../../../../components/Divider/Divider'
 import { DividerVariables } from './dividerVariables'
 
-const beforeAndAfter = (color, size, variables: DividerVariables): ICSSInJSStyle => ({
+const beforeAndAfter = (
+  color: string,
+  size: number,
+  variables: DividerVariables,
+): ICSSInJSStyle => ({
   content: '""',
   flex: 1,
   height: `${size + 1}px`,
-  background: color ? _.get(variables.colors, color) : variables.dividerColor,
+  background: _.get(variables.colors, color, variables.dividerColor),
 })
 
 const dividerStyles: ComponentSlotStylesInput<DividerPropsWithDefaults, DividerVariables> = {
   root: ({ props, variables }): ICSSInJSStyle => {
     const { children, color, fitted, size, important, content } = props
     return {
-      color: color ? _.get(variables.colors, color) : variables.textColor,
+      color: _.get(variables.colors, color, variables.textColor),
       display: 'flex',
       alignItems: 'center',
       ...(!fitted && {
--- a/src/themes/teams/components/Divider/dividerStyles.ts
+++ b/src/themes/teams/components/Divider/dividerStyles.ts
@@ -2,36 +2,27 @@ import * as _ from 'lodash'
 
 import { childrenExist } from '../../../../lib'
 import { pxToRem } from '../../utils'
-import { ComponentSlotStylesInput, ICSSInJSStyle, ICSSPseudoElementStyle } from '../../../types'
+import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
 import { DividerPropsWithDefaults } from '../../../../components/Divider/Divider'
+import { DividerVariables } from './dividerVariables'
 
-const dividerBorderStyle = (size, color): ICSSInJSStyle => ({
-  height: `${size + 1}px`,
-  background: color,
-})
-
-const beforeAndAfter = (color, size, type, variables): ICSSPseudoElementStyle => ({
+const beforeAndAfter = (color, size, variables: DividerVariables): ICSSInJSStyle => ({
   content: '""',
   flex: 1,
-  ...dividerBorderStyle(size, variables.dividerColor),
-  ...(color && {
-    ...dividerBorderStyle(size, _.get(variables.colors, color)),
-  }),
+  height: `${size + 1}px`,
+  background: color ? _.get(variables.colors, color) : variables.dividerColor,
 })
 
-const dividerStyles: ComponentSlotStylesInput<DividerPropsWithDefaults, any> = {
+const dividerStyles: ComponentSlotStylesInput<DividerPropsWithDefaults, DividerVariables> = {
   root: ({ props, variables }): ICSSInJSStyle => {
-    const { children, color, fitted, size, type, important, content } = props
+    const { children, color, fitted, size, important, content } = props
     return {
-      color: variables.textColor,
+      color: color ? _.get(variables.colors, color) : variables.textColor,
       display: 'flex',
       alignItems: 'center',
       ...(!fitted && {
         paddingTop: variables.dividerPadding,
         paddingBottom: variables.dividerPadding,
-      }),
-      ...(color && {
-        color: _.get(variables.colors, color),
       }),
       ...(important && {
         fontWeight: variables.importantFontWeight,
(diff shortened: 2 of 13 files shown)


src/lib/colorUtils.ts
```diff
@@ -0,0 +1,11 @@
+import * as _ from 'lodash'
+import { SiteVariablesInput, ColorVariants, ColorValues } from '../themes/types'
+
+export const mapColorsToScheme = <T>(
+  siteVars: SiteVariablesInput,
+  mapper: string | number | keyof ColorVariants | ((color: ColorVariants) => T),
```
layershifter: It was wrong advice from my side to use `string`, in fact the valid signature is below:

```suggestion
  mapper: keyof ColorVariants | ((color: ColorVariants) => T),
```

src/themes/teams/components/Divider/dividerStyles.ts
```diff
@@ -2,37 +2,28 @@ import * as _ from 'lodash'
...
-    ...dividerBorderStyle(size, _.get(variables.colors, color)),
-  }),
+  height: `${size + 1}px`,
+  background: color ? _.get(variables.colors, color) : variables.dividerColor,
 })
 
-const dividerStyles: ComponentSlotStylesInput<DividerPropsWithDefaults, any> = {
+const dividerStyles: ComponentSlotStylesInput<DividerPropsWithDefaults, DividerVariables> = {
   root: ({ props, variables }): ICSSInJSStyle => {
-    const { children, color, fitted, size, type, important, content } = props
```
layershifter, line 24: Nice catch, `type` was removed before 👍

src/themes/types.ts
```diff
@@ -71,6 +71,16 @@ type EmphasisColorsStrict = Partial<{
 
 export type EmphasisColors = Extendable<EmphasisColorsStrict, ColorVariants>
 
+/**
+ * A type for extracting the color names.
+ */
+type ColorNames = keyof (EmphasisColorsStrict & NaturalColorsStrict)
```
layershifter, line 77:
```suggestion
type ColorNames = keyof (EmphasisColors & NaturalColors)
```

We should use extendable color types, otherwise it will block the usage of custom colors.

src/themes/teams/components/Divider/dividerStyles.ts
```diff
@@ -2,37 +2,28 @@ import * as _ from 'lodash'
...
+    const { children, color, fitted, size, important, content } = props
     return {
-      color: variables.textColor,
+      color: color ? _.get(variables.colors, color) : variables.textColor,
```
layershifter:
```suggestion
      color: _.get(variables.colors, color, variables.textColor),
```

This should work, too. However, I am not happy with this because `dividerColor` and `textColor` looks completely strange there 🤔 

---

It seems that we need to have a `default` color (may be `base`???) that will include this:
```diff
-type ColorNames = keyof (EmphasisColors & NaturalColors)
+type ColorNames = keyof (EmphasisColors & NaturalColors) & 'default'
```

And then we will able to do something like this.
```js
// color is `default` when is not defined

{
  color: _.get(variables.colors, [color, 'text'], 'inherit'), // inherit is a fallback value
  background: _.get(variables.colors, [color, 'background'], 'inherit'), // this thing can be our util in fact
}
```

Please let me know, what you think about this.

src/themes/types.ts
```diff
@@ -71,6 +71,16 @@ type EmphasisColorsStrict = Partial<{
...
+
+/**
+ * A type for extracting the color names and values.
```
layershifter: We should add a more correct description there. `Record`'s description:

> Construct a type with a set of properties K of type T

src/themes/teams/components/Text/textVariables.ts
```diff
@@ -29,7 +33,10 @@ export interface TextVariables {
 }
 
 export default (siteVariables): TextVariables => {
+  const colorVariant = '500'
+
   return {
+    colors: mapColorsToScheme(siteVariables, colorVariant),
```
layershifter, line 39: With current structure it seems that the change of `color` will not affect `disabled`:

```js
<Text color='red' disabled />
```

And it will have `siteVariables.gray06` as color 😨

layershifter (commented):
I left some comments, let me know what you think 🕵️

src/themes/types.ts
bmdalex: with your suggestions we don't get the correct keys exactly because `keyof` does not work as expected on extendable types:
screenshot 2018-12-12 at 19 15 15
with my change:
screenshot 2018-12-12 at 19 15 06

and to address your comment I changed

```tsx
export type ColorValues<T> = Partial<Record<ColorNames, T>>
```
to
```tsx
export type ColorValues<T> = Extendable<Partial<Record<ColorNames, T>>, T>
```
in order not to block the usage of custom colors. What do you think?

src/themes/teams/components/Text/textVariables.ts
bmdalex: @layershifter I know, that seems wrong, `disabled` prop should not change color, `disabled` prop should only impact functionality and opacity

@codepretty I can only see you as last author, can you weigh in on this?

Also tagging @notandrew  who implemented the same behavior (`disabled` changes color) for `Button` in #216

src/themes/types.ts
bmdalex: changed

src/lib/colorUtils.ts
bmdalex: done

src/themes/teams/components/Divider/dividerStyles.ts
bmdalex: @layershifter I made the change for

`color: _.get(variables.colors, color, variables.textColor)`

but for the default color I'm not even sure I get it, looks a bit like a premature optimization. Are you satisfied with current state?

src/lib/colorUtils.ts
```diff
@@ -0,0 +1,11 @@
+import * as _ from 'lodash'
+import { SiteVariablesInput, ColorVariants, ColorValues } from '../themes/types'
+
+export const mapColorsToScheme = <T>(
+  siteVars: SiteVariablesInput,
+  mapper: keyof ColorVariants | ((color: ColorVariants) => T),
+): ColorValues<T> =>
+  _.mapValues(
+    { ...siteVars.emphasisColors, ...siteVars.naturalColors },
```
bmdalex, line 9: @layershifter 
thinking one more time about this, rather than merging always these 2 objects, why don't we just use the
`siteVars.colors` that contains already this merge:

https://github.com/stardust-ui/react/blob/4a7eceae3b3f0571a667c8622eb87d7b0cdf00d5/src/themes/default/colors.ts#L138
```jsx
export const colors: ColorPalette = {
  ...contextualColors,
  ...emphasisColors,
  ...naturalColors,

  black: '#000',
  white: '#fff',
}
```
we'd need `contextualColors` anyways when deciding upon text colors for box components like `Button` or `Label`:
```jsx
Label color='red' ../>
```

will apply
```jsx
const bgColor = siteVars.colors['white'][500], // or other variant
const textColorVariant = Color(bgColor).isDark()
  ? 50 /* white text */
  : 900 /* black text (or other variants) */
{
  backgroundColor: bgColor,
  color: siteVars.colors.text[textColorVariant],
}
```

tl;dr
**What do you think about:**

```suggestion
    siteVars.colors,
```
and the other approach using `Color(bgColor).isDark()`?

src/themes/types.ts
layershifter: LGTM, there is only one thing that confuses me, it's a name of `ColorNames`. It doesn't represent the merge of `EmphasisColors` and `NaturalColors`. May be other folks have a better idea there 💡

src/themes/teams/components/Divider/dividerStyles.ts
layershifter: It seems that I wasn't very clear there. Take a look on `DividerVariables`, we have `colors`, `dividerColor` and `textColor` 🤔 

We can do this in a separate PR, but my proposal is to introduce a scheme there:
```diff
-type ColorNames = keyof (EmphasisColorsStrict & NaturalColorsStrict)
+type ColorNames = keyof (EmphasisColorsStrict & NaturalColorsStrict) & 'base'

//

interface DividerColorScheme {
  background: string
  text: string
}

export interface DividerVariables {
  colors: ColorValues<DividerColorScheme>
-  dividerColor: string
-  textColor: string
}
```

This will allow to drop `dividerColor` and `textColor`, they looks extremely confusing now.

src/themes/teams/components/Text/textVariables.ts
layershifter: > should only impact functionality and opacity

Yep, and it seems that it should be a part of a color scheme and will result something like this:
image

src/lib/colorUtils.ts
layershifter: > siteVars.colors that contains already this merge:

From one side I will agree with this, from other side `colors` contains colors from `contextualColors` like `text` and `danger`. We considered previously that `danger` is a state of a component, not a color, i.e:

```jsx
<Message danger /> // valid
<Message color='danger' /> // invalid
```

layershifter: We can use `Color()`, i.e the `color` package for docs needs. But we should omit it as potential dependency, #583.

src/themes/teams/components/Divider/dividerStyles.ts
bmdalex: I'll make a proposal to introduce this to a separate PR, maybe we find a pattern for creating a unified scheme

src/themes/types.ts
bmdalex: if anyone has better ideas for name, pls provide some recommendations:

@mnajdova @kuzhelov @miroslavstastny @levithomason

src/lib/colorUtils.ts
bmdalex: > from other side colors contains colors from contextualColors like text and danger

I'd like to avoid recreating this object every time the method is called. What about creating this `{ ...siteVars.emphasisColors, ...siteVars.naturalColors },` object in the theme object directly, the same way we did here: https://github.com/stardust-ui/react/blob/4a7eceae3b3f0571a667c8622eb87d7b0cdf00d5/src/themes/default/colors.ts#L138 for colors object?

bmdalex: > We can use Color(), i.e the color package for docs needs. But we should omit it as potential dependency, #583.

how else can we use it if we want to omit it as potential dependency? 😕

src/themes/teams/components/Text/textVariables.ts
codepretty: @layershifter I understand the sentiment and (maybe?) that is what you would want for some themes, but not for Teams.

The `disabled` prop should dictate color for Teams theme since opacity isn't the correct treatment for disabled states in Teams.

src/themes/teams/components/Divider/dividerStyles.ts
layershifter: Yep, we discussed it yesterday. Marked this item as resolved.

src/lib/colorUtils.ts
layershifter: I mean that we can use it as dev dependency, but `color` can not be used as a package dependency, i.e. it can be a part of lib or themes.

src/themes/teams/components/Text/textVariables.ts
layershifter: @codepretty got it. Thanks for the feedback.

layershifter (approved):
LGTM, let's merge it and iterate if we will see issues.

Issue #557, opened 2018-12-04
Popup not closing when in Menu


johannao76:
### Steps
When using Popup with Menu, clicking on a menu item when another menu item's popup is open, will not close the original item. Use this code snippet in the docs site:
```jsx
import React from 'react'
import { Menu, Popup} from '@stardust-ui/react'

const items = [
  { key: 'editorials', content: 'Editorials' },
  { key: 'review', content: 'Reviews' },
  { key: 'events', content: 'Upcoming Events' },
]

const MenuExample = () => <Menu defaultActiveIndex={0} items={items} renderItem={renderItem}/>

const renderItem = (MenuItem, props) => {
  return (
    <Popup 
      trigger={<MenuItem {...props} />}
      content={{content: "TestContent"}}
    />
  )
}

export default MenuExample
```

### Expected Result
When a menu item's popup is open, clicking on another menu item should close the popup and open the clicked item's popup.

### Actual Result
We end up with two open popups

### Version
0.13.1

miroslavstastny:
The problem is that when a `MenuItem` is clicked:
1. whole menu is updated,
2. this results in `Popup.componentDidUpdate()`,
3. which removes click listener.
4. and re-adds it in `setTimeout()`.

Document click event is called between steps 3 and 4 resulting in the event being missed by the `Popup`.

Click on the *Upcoming Events* opens popup:
```log
MenuItem.tsx:74 MenuItem::componentDidUpdate Upcoming Events
12:43:08.398 MenuItem.tsx:60 MenuItem::handleClick Upcoming Events
12:43:08.415 MenuItem.tsx:74 MenuItem::componentDidUpdate Editorials
12:43:08.416 Popup.tsx:104 Popup::updateSubcription Editorials
12:43:08.416 MenuItem.tsx:74 MenuItem::componentDidUpdate Reviews
12:43:08.417 Popup.tsx:104 Popup::updateSubcription Reviews
12:43:08.417 MenuItem.tsx:74 MenuItem::componentDidUpdate Upcoming Events
12:43:08.418 Popup.tsx:104 Popup::updateSubcription Upcoming Events
12:43:08.419 Menu.tsx:84 Menu::componentDidUpdate
12:43:08.425 EventTarget.tsx:16 _emit [ƒ]
12:43:08.425 Menu.tsx:80  Document.onClick
12:43:08.443 Popup.tsx:110 Popup - subscribe Upcoming Events
```

Click on the *Reviews*:
```log
[...]

PR #598, opened 2018-12-11
fix(Popup): ensure popup is closing on outside click

kuzhelov:
Fixes the issue is described here: #557 

## TODO
- [x] update changelog

## Root cause
The cause of the problem is the fact that `Menu` component "intercepts" click event, so that `Popup` is unaware of whether this event has happened or not.

## Fix
`useCapture` event interception logic is now used for `Popup` event subscriptions - this will guarantee that event won't pass unnoticed.


miroslavstastny (approved):
This fixes the issue with `MenuItem` being the trigger for popup.
It still does not work for nested popup (no regression) - https://codesandbox.io/s/wk5vvw56pw.

PR #600, opened 2018-12-11
chore(package): add ESLint

levithomason:
This PR replaces ts-lint with eslint. 

- reduced linting time from >13s to ~3s 🎉 
- there is far more support and better plugins in the eslint community
- we can now lint/fix JS files, which were getting ignored before by ts-lint

----

### Changes
- added ESLint with common plugins
- disabled all failing rules, with uncomment them in future PRs


package.json
```diff
@@ -19,6 +19,8 @@
     "deploy:docs": "gulp deploy:docs",
     "lint": "tslint -t stylish -p .",
```
miroslavstastny: this is now broken

package.json
```diff
@@ -19,6 +19,8 @@
...
     "lint:fix": "yarn lint --fix",
+    "eslint": "eslint .",
```
miroslavstastny: I would prefer to name this just `lint`. The fact that we use eslint for that is an implementation detail I don't care about.

levithomason: Just FYI, while the PR was in `WIP` state, I had both so I could run parallel tests of the two.  Certainly agree on `lint` 👍

levithomason:
This PR is blocked waiting on the latest release from `eslint-typescript-parser`.  Our update to `typescript@3.2.x` is actually a breaking change for `typescript-estree`:

```
WARNING: You are currently running a version of TypeScript which is not officially
supported by typescript-estree.

SUPPORTED TYPESCRIPT VERSIONS: ~3.1.1
YOUR TYPESCRIPT VERSION: 3.2.2
```

See either of these issues:
https://github.com/eslint/typescript-eslint-parser/pull/576
https://github.com/eslint/typescript-eslint-parser/pull/589

layershifter:
Updated dependencies and unblocked PR 👍

levithomason:
I fixed the yarn.lock, but this caused issues with ESLint resolving the eslint config.  Once fixing the resolution, there were over 3k lint errors.  Will have to come back to this one.

.eslintrc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": ["./packages/internal-tooling/eslint/index.js"],
```
layershifter, line 2: Paths are not great, should be fixed in ESLint 6: https://github.com/eslint/eslint/pull/11388

build/gulp/tasks/test-projects/.eslintrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+    "import/no-unresolved": "off"
```
layershifter, line 3: We have requires for Enzyme, Simulant and other stuff that it's not in direct dependencies. Currently disabled

packages/react/.eslintrc.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": ["../internal-tooling/eslint/index.js"],
+  "overrides": [
+    {
+      "files": "**/icons/*.tsx",
+      "rules": {
+        "react/prop-types": "off"
```
layershifter, line 7: ESLint with React plugin handles these icons as React components, it will be very strange to have there `propTypes`...

packages/internal-tooling/eslint/index.js
```diff
@@ -0,0 +1,122 @@
...
+    'no-use-before-define': 'off',
+    'operator-assignment': 'off',
+    'prefer-destructuring': 'off',
+    'spaced-comment': 'off',
+  },
+  overrides: [
+    {
+      files: '**/jest.config.js',
+      rules: {
+        'global-require': 'off',
```
layershifter, line 105: It's safe to use `require()` there

miroslavstastny (approved):

layershifter:
We decided to keep TSLint before at least important rules will work properly.

PR #585, opened 2018-12-10
[template] feat(icon): export `call-recording` icon

bmdalex:
# feat(icon): export `call-recording` icon

## Description
This PR exports `call-recording` icon.

It also acts as a `TEMPLATE PR` for exporting Teams icons; steps are:
#### 1. find your `icons-[NAME].tsx` file:
  - visually in the local doc site experience (need to build and run the project and go to http://localhost:8080/icon-viewer):
screenshot 2018-12-10 at 14 06 32
  - find the icon at path `src/themes/teams/components/Icon/svg/ProcessedIcons`: `src/themes/teams/components/Icon/svg/ProcessedIcons/icons-call-recording.tsx` in this case

#### 2. copy `icons-[NAME].tsx` file to `src/themes/teams/components/Icon/svg/icons` and rename it to the necessary name:
- **source**: `src/themes/teams/components/Icon/svg/ProcessedIcons/icons-call-recording.tsx`
- **dest**: `src/themes/teams/components/Icon/svg/icons/callRecording.tsx`

#### 3. import and export the new icon file under desired name USING KEBAB CASE:
```jsx
// ...
import callRecording from './callRecording'
// ...

export default {
  // ...
  'call-recording': callRecording,
  // ...
}
```

#### 4. go back to `src/themes/teams/components/Icon/svg/ProcessedIcons/icons-[NAME].tsx` and add `exportedAs` property with the value of name of the exported icon:

in `src/themes/teams/components/Icon/svg/ProcessedIcons/icons-call-recording.tsx`:
```jsx
export default {
  // ...
  exportedAs: 'call-recording',
}
```

## API

```jsx
<Icon name="call-recording" variables={siteVars => ({ color: siteVars.red })} />
```

screenshot 2018-12-10 at 13 53 33

PR #601, opened 2018-12-12
feat(Icons): Added icon for add-page and also added missing icon sizes

priyankar205:
1. Added add-page icon
2. Added sizes 'x-large' and 'xx-large' for icons.


src/themes/teams/components/Icon/svg/icons/addPage.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
```
kuzhelov: we need to mark corresponding icon as exported one on the Processed Icons page in the docs - it is needed to specify value for `exportedAs` field for the corresponding icon from processed set (`canvasaddnewpage`?). 

Please, take a look on this small PR that you could consider as an example for that: https://github.com/stardust-ui/react/pull/585

src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -14,6 +14,8 @@ const sizes = new Map([
   ['small', 14],
   ['normal', 16],
   ['large', 20],
+  ['x-large', 24],
```
kuzhelov: lets move it to icon's variables, specifically:

```jsx
<Icon size='large' variables={{ sizeModifier: 'xx' }}>
```

And adjust the logic of mapping `size` prop + `sizeModifier` variable => size value.

With that we will remove the following problems:
- we won't introduce any changes to the `Icon's` API - note, currently `xx`-prefixed values are not supported, according to the Icon's typings (and those are used for docs generation)
- won't introduce inconsistency in type for `size` prop - note, currently `xx` prefixes are only supported for large values, and for other ones `size` type was left unchanged.

I am sure this problem (with extendable enums) will be addressed in future by Stardust, but for now using variables seems to be less evil we can do to circumvent the problem.

priyankar205: @kuzhelov The better approach will be to let the user enter a number as a size. Something as sizeInNumber

priyankar205: Done

src/themes/teams/components/Icon/svg/icons/addPage.tsx
priyankar205: Done

src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -19,6 +19,8 @@ const sizes = new Map([
   ['massive', 64],
 ])
 
+const modifiedSizes = new Map([['x-large', 24], ['xx-large', 28]])
```
kuzhelov: this one is used in `getSize` method only - lets inline this mapping there (actually, there is no particular reason to use `Map` type for that). This will keep things cohesive

src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -71,7 +74,12 @@ const getPaddedStyle = (): ICSSInJSStyle => ({
   padding: pxToRem(4),
 })
 
-const getSize = size => pxToRem(sizes.get(size))
+const getSize = (size, sizeModifier) => {
+  if (sizeModifier && size === 'large') {
```
kuzhelov: would rather suggest to stick to the semantics of this method: 
- if there is no modifier provided, then just `size` should be returned
- otherwise `sizeModifier` should be processed
   - and currently it has specific logic for `large` size values only

```jsx
const getSize = (size, sizeModifier) => {
  if (!sizeModifier) {
     return size
  }

  const modifiedSizes = {
     'large': {
        'x': ...,
        'xx': ...
     } 
  } 

  return modifiedSizes[size] && modifiedSizes[size][sizeModifier]
}
```

This delivers much cleaner intent and will alleviate future refactorings.

src/themes/teams/components/Icon/svg/ProcessedIcons/icons-canvas-addpage.tsx
```diff
@@ -18,4 +18,5 @@ export default {
     </svg>
   ),
   styles: {},
+  exportedAs: 'canvas-add-page',
```
kuzhelov, line 21: 👍

src/themes/teams/components/Icon/svg/icons/index.ts
```diff
@@ -50,6 +51,7 @@ export default {
   'call-control-present-new': callControlPresentNew,
   'call-control-stop-presenting-new': callControlStopPresentingNew,
   'call-recording': callRecording,
+  'canvas-add-page': canvasAddPage,
```
kuzhelov, line 54: 👍

src/themes/teams/components/Icon/iconStyles.ts
priyankar205: Done

kuzhelov (approved):

src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -71,7 +72,19 @@ const getPaddedStyle = (): ICSSInJSStyle => ({
   padding: pxToRem(4),
 })
 
-const getSize = size => pxToRem(sizes.get(size))
+const getIconSize = (size, sizeModifier): number => {
+  if (!sizeModifier) {
+    return sizes.get(size)
+  }
+  const modifiedSizes = {
```
codepretty, line 79: What is the context this icon will be used? We will need a way to specify sizing, but I don't believe we want to do it in css. In Teams, the different sizes should actually be different icons (in most cases).

levithomason:
We need to undo the complexity added here with the `sizeModifier`.  The code base already has number sizes, Semantic UI sizes, plus an additional 'micro' size added.  This PR introduces a 4th concept to sizing.

We should instead unify our 3 sizing patterns to accomplish this feature.  Let's always reduce complexity with changes.

levithomason:
https://github.com/stardust-ui/react/issues/136#issuecomment-447152727

PR #602, opened 2018-12-12
fix(docs): fix key warnings

layershifter:
This PR fixes two React's warnings about missing keys in our docs.


miroslavstastny (approved):
changelog?

layershifter:
@miroslavstastny Sure, added an entry 👍

PR #603, opened 2018-12-12
perf(docgen): speed up `build:docs:component-menu`

### LINKED ISSUE #507: Docs: improve build time of component menu
_I've have already made caching for our `build:docs:docgen` task #485, we get a cool boost there._

However, our `build:docs:component-menu` is still slow :( I take a look on it and found ot the main problem. It uses that the uncached calls of `getComponentInfo()` which is quite slow.

My proposal is simple.

### 1. Update task pipeline

```diff
-parallel(
-  'build:docs:docgen',
-  'build:docs:component-menu',
-  // more tasks
-)
+parallel(
+  series('build:docs:docgen', 'build:docs:component-menu'),
+  // more tasks
+)
```

This will grant us that `build:docs:component-menu` will be executed after `build:docs:docgen`.

### 2. Use generated info

Instead of calls of `getComponentInfo()`, we can just read info from JSON files in `build:docs:component-menu`.

----

In result this will save us more than 30 seconds (1.5 minutes for me) for `yarn start`.

layershifter:
Fixes #507.

---

This PR removes duplicate uncached call of `getComponentInfo()` and just uses already existing info files.

### Before
```
[14:06:46] Starting 'build:docs:component-menu'...
[14:08:04] Finished 'build:docs:component-menu' after 1.3 min
```

### After
```
[14:10:30] Starting 'build:docs:component-menu'...
[14:10:31] Finished 'build:docs:component-menu' after 290 ms
```

### DIFF
--- a/build/gulp/plugins/gulp-component-menu.ts
+++ b/build/gulp/plugins/gulp-component-menu.ts
@@ -5,6 +5,7 @@ import * as through2 from 'through2'
 import * as Vinyl from 'vinyl'
 
 import config from '../../../config'
+import getComponentInfo from './util/getComponentInfo'
 
 const pluginName = 'gulp-component-menu'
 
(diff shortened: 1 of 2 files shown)


miroslavstastny (approved):
I don't like tasks with hidden dependencies on other tasks.
But that is just a theoretical problem, the numbers are real. Good job!

build/gulp/plugins/gulp-component-menu.ts
```diff
@@ -27,12 +28,15 @@ export default () => {
     }
 
     try {
-      const info = getComponentInfo(file.path)
+      const infoFilename = file.basename.replace(/\.tsx$/, '.info.json')
 
-      if (info.isParent) {
+      const jsonInfo = fs.readFileSync(config.paths.docsSrc('componentInfo', infoFilename))
```
kuzhelov: we could use a following semantics here, to prevent issue of tasks dependency:
- if file exists, use info from this file
- if not, use `getComponentInfo()` as fallback

layershifter: Nice idea, I added a condition there

kuzhelov (approved):

PR #605, opened 2018-12-12
fix(ButtonGroup): remove 'accessibility' prop and corresponding behavior

### LINKED ISSUE #604: ButtonGroup should not be navigable using arrow keys
## Feature Request

Button Group should not be navigable using arrow keys

### Problem description
Users do not have any information, that the keys are available. If the intention of the user is to have related buttons, Menu component with ToolbarBehavior should be used instead.

### Proposed solution
Remove ButtonGroupBehavior

kolaps33:
Removing buttonGroup behavior.

Fixes #604


src/components/Button/ButtonGroup.tsx
```diff
@@ -50,7 +48,6 @@ class ButtonGroup extends UIComponent<Extendable<ButtonGroupProps>, any> {
 
   public static defaultProps = {
     as: 'div',
-    accessibility: buttonGroupBehavior as Accessibility,
```
jurokapsiar, line 53: you can either change the default value to defaultBehavior or remove also the `accessibility` prop (line 27). I would leave the decision up to more experienced folks - removing the prop is a breaking change.

kuzhelov: agree with the concerns expressed - lets introduce `defaultBehavior` as default `accessibility` prop value and merge

kuzhelov: talked with @jurokapsiar  - as in foreseeable future there is no need to define accessibility behavior for this component, agreed to remove this prop (to better reflect current state of things)

kuzhelov (approved):

PR #549, opened 2018-11-30
chore(CI): temporary disable `Project Tests`

layershifter:
This PR temporary disables `Project Tests`, refs #548.

Will be reverted with typings fixes.

PR #608, opened 2018-12-13
chore(CI): reenable `test:projects:cra-ts`

layershifter:
This PR reenables previously disabled `test:projects:cra-ts`, #549.


build/gulp/tasks/test-projects.tsx
```diff
@@ -167,7 +124,7 @@ task('test:projects:rollup', async () => {
 task(
   'test:projects',
   series(
-    // 'test:projects:cra-ts', Temporary disabled
-    'test:projects:rollup',
+    series('build:dist', 'test:projects:pack'),
```
layershifter: `yarn gulp build:dist` and `yarn build:dist` do the same work, if we do it there as Gulp's we will avoid double transpilation with `ts-node` 🚀

build/gulp/tasks/test-projects.tsx
```diff
@@ -53,93 +42,61 @@ const createReactApp = async (atTempDirectory: string, appName: string): Promise
...
-          <Image src="//placehold.it" />
-          <Input placeholder="Type here" />
-        </div>
-      </Provider>
-    );
-  }
-}
-
-export default App;
-`
```
layershifter, line 92: We have now a separate folder for this 👍

build/gulp/tasks/test-projects.tsx
```diff
@@ -53,93 +42,61 @@ const createReactApp = async (atTempDirectory: string, appName: string): Promise
...
 
+task('test:projects:pack', async () => {
+  const logger = debug('bundle:pack')
+  logger.enabled = true
+
+  packageFilename = createPackageFilename()
+  await sh(`yarn pack --filename ${packageFilename}`)
+
+  logger(`Stardust package is published: ${packageFilename}`)
+})
```
layershifter: We will run `test:projects:pack` only once for both tests 🚀

build/gulp/tasks/test-projects.tsx
```diff
@@ -9,23 +9,12 @@ import config from '../../../config'
 import * as tmp from 'tmp'
 
 const { paths } = config
-
-const log = msg => {
```
kuzhelov, line 13: now the output is indistinguishable, it is very hard to understand what is going on and immediately find the step necessary:

image

It will significantly complicate debugging scenarios

build/gulp/tasks/test-projects.tsx
```diff
@@ -9,23 +9,12 @@ import config from '../../../config'
...
-  console.log()
-  console.log('='.repeat(80))
-  console.log('CRA TS:', msg)
-  console.log('='.repeat(80))
-}
+let packageFilename
```
kuzhelov: don't like this broken encapsulation between the tasks, especially given that all can be tried to use in isolation. This implicit composability assumptions would be really hard to track - thus, would prefer correctness there at the first place

layershifter: I'm not so familiar with Gulp, do you have a better approach there? How I can pass a name between tasks?

layershifter: Nevermind, `yarn pack` is quite fast so I moved it back to each step.

build/gulp/tasks/test-projects.tsx
layershifter: Agree, I restored your logger

build/gulp/tasks/test-projects.tsx
kuzhelov: @layershifter, thanks - unfortunately, there is built-in support in gulp to produce result from task (just in a form of artefacts, but this is definitely not what your question about:) As alternative, we might consider cache semantics, but absolutely agree that it is better to split, especially if this is a quick step 👍

kuzhelov (approved):

kuzhelov:
just note aside - could you, please, suggest if you've had a chance to research the cause (source) of these TS warnings:

image

I mean, I see that those are produced by rollup's env, but just to understand which problem this is. Thanks!

layershifter:
> I mean, I see that those are produced by rollup's env, but just to understand which problem this is. 

This problem caused by TS helpers in the transpiled code, more details are there: https://github.com/rollup/rollup/issues/794#issuecomment-262395554. I am not sure that we need to silence this warning, especially if we want to migrate to Babel.

PR #611, opened 2018-12-13
feat(icon): Add and export new compose icons

joheredi:
This PR adds and exports a set of new icons for compose
image


kuzhelov:
note, you are creating PR with the wrong target - it should be just `master` (not `stardust:master`, no colons):

image

joheredi:
> note, you are creating PR with the wrong target - it should be just `master` (not `stardust:master`, no colons):
> 
> image

@kuzhelov  Thanks for pointing this out. I'll create a new PR, is this the right way?
image

joheredi:
> > note, you are creating PR with the wrong target - it should be just `master` (not `stardust:master`, no colons):
> > image
> 
> @kuzhelov Thanks for pointing this out. I'll create a new PR, is this the right way?
> image

@kuzhelov I think it shows as `stardust-ui:master` because I'm merging from my Fork. I'm not an "official" contributor so I couldn't create a branch within the main fork.

joheredi:
Submitting a new Pull Request against master

PR #612, opened 2018-12-13
fix(Popup): allow to 'detach' from trigger and RTL adjustments

kuzhelov:
## TODO
- [x] introduce unit tests for RTL transform logic
- [x] update changelog

This PR fixes the following issues: https://github.com/stardust-ui/react/pull/606#issuecomment-447103020

## Allow to detach popup from trigger
Previously it was impossible - it was guaranteed by implementation logic that popup and trigger would always have at least single common point. Thus, even for the following example where user might intentionally would like to specify large offset value - even in that case rendered popup will be 'attached' to its trigger. 

```jsx
<Popup
  position='above'
  align='end'
  offset='-100%'
/>
```
image

### with an extremely huge offset result had stayed the same:
```jsx
<Popup
  position='above'
  align='end'
  offset='-10000%'
/>
```
image

Provided fix ensures that client's offset will be applied 'as is', without any additional heuristics being applied.

## Ensure correct RTL transform for complex `offset` expressions
Previous RTL transform logic had implicitly assumed that it could be only single value passed as `offset` - however, `offset` also supports expressions syntax, so that the following value is absolutely correct:

```jsx
<Popup 
  offset='100% - 100%p + 10'
  ...
/>
```

This PR addresses such limitation.


test/specs/components/Popup/Popup-test.tsx
```diff
@@ -56,4 +61,45 @@ describe('Popup', () => {
...
+
+    const expectOffsetTransformResult = (originalOffset, resultOffset) => {
+      expect(applyRtlToOffset(originalOffset, 'above')).toBe(resultOffset)
+    }
+
+    it('flips sign of simple expressions', () => {
+      expectOffsetTransformResult('100%', '-100%')
+      expectOffsetTransformResult('  2000%p ', '-2000%p')
+      expectOffsetTransformResult('100  ', '-100')
+      expectOffsetTransformResult(' - 200vh', '200vh')
```
layershifter, line 84: You can use `jest.each` there: https://jestjs.io/docs/en/api#testeachtable-name-fn-timeout

src/components/Popup/positioningHelper.ts
```diff
@@ -74,3 +74,30 @@ export default ({
...
+const flipPlusMinusSigns = (offset: string): string => {
+  return offset
+    .replace(/\-/g, '<plus>')
+    .replace(/^(\s*)(?=\d)/, '<minus>')
+    .replace(/\+/g, '<minus>')
+    .replace(/<plus>/g, '+')
+    .replace(/<minus>/g, '-')
+    .trimLeft()
+    .replace(/^\+/, '')
+}
```
layershifter, line 91: This looks a complicated for me, but we have tests. Possibly, we can to add memoization there.

docs/src/examples/components/Popup/Variations/PopupExampleOffset.shorthand.tsx
```diff
@@ -11,33 +11,24 @@ const renderButton = rotateArrowUp => (
...
-              The popup is rendered at {position}-{align}
-              <br />
-              corner of the trigger.
-            </p>
-          ),
-        }}
-        key={`${position}-${align}`}
-      />
-    ))}
+  <Grid columns="repeat(1, 80px)" variables={{ padding: '30px', gridGap: '30px' }}>
```
layershifter: It seems that we don't `repeat` there more.

layershifter (approved):
I played with it locally and manually checked RTL offsets. Works fine 👍

docs/src/examples/components/Popup/Variations/PopupExampleOffset.shorthand.tsx
kuzhelov: absolutely, need to fix

src/components/Popup/positioningHelper.ts
kuzhelov: yes, the tests were introduced exactly for this sake :) not sure about memoization, as all expressions used are of linear complexity - there shouldn't be any significant gain from it.

test/specs/components/Popup/Popup-test.tsx
kuzhelov: have intentionally selected this approach for the following reasons:
- the `jest.each()` is less expressive than explicit name of the check
- we are not interested in `name` arg that `jest.each()` might provide for each case under check
- even given that this is more verbose, might argue that we will read these lines far more often than write them, so I'd better prefer expressiveness here
- this approach is consistent with the one that is already taken for the rest of the tests in this file (which is, also, provide more readable tests code)

PR #556, opened 2018-12-04
feat(chat): chat item gutter and refactoring

bmdalex:
# feat(chat): chat item gutter and refactoring

## Description

This PR implements **solution 1** from #537 and closes #537 

We need to refactor our `Chat` subcomponents, `Chat.Item` and `Chat.Message` for the following reasons:
- need better structuring and more flexibility;
- need to solve specific clients needs when it comes to layout of components and providing very specific content to specific areas;
- `Chat.Message` is too specific (the `avatar` should be more top level)
- `Chat.Item` is not specific enough (needs `gutter`)

## Solution (implemented in this PR)

This PR implements a mix of **solution 1**  from #537 and other discussion I had with the team; here are the main steps:
- removing `avatar` slot from `Chat.Message`
- adding `gutter` slot to `Chat.Item`;
- adding `gutterPosition='start' | 'end'` prop to `Chat.Item`;
- renaming `content` slot to `message` for `Chat.Item`;
- deprecating Children API for `Chat` components;

## API:
```jsx
<Chat items={[
  {
    message: <AnyComponent .../>,
  },
  {
    mine: true,
    gutterPosition: 'end',
    gutter: { content: <Avatar image="" status="" /> },
    message: { content:
      <Chat.Message
        content="Let's have a call"
        author="John Doe"
        timestamp="Today, 11:15 PM"
        mine
      />,
    },
  },
]} />
```

## Known issues:

- need to add `mine` prop to `Chat.Item` as well for styling purposes and deciding on `gutter` positioning; other suggestion was: `gutterPosition: 'before' | 'after'` (***LE**: decided `gutterPosition: 'start' | 'end'`)
- in order to leverage correct styles for `gutter` and `content` slots we will need to specify the `content` and `gutter` props as objects with `content` props instead of just elements after merging #496 (***LE**: we will use verbose `content` props for now and look into #512 to simplify that)
- `chatItemStyles` are too specific and are causing other components to be rendered incorrectly as content (Divider for instance, see below); we might have to go with **solution 2** from #537 and create subcomponents. (***LE**: fixed by using new styles that do not rely on flexbox, by using css children selectors for styling)
screenshot 2018-12-05 at 12 17 27
- `maxWidth: 80%` for `gutter` + `content` requires complex styling in order to be achieved (***LE**: fixed by using new styles that do not rely on flexbox, by using css children selectors for styling)

***LE**: ALL ISSUES ARE HANDLED, GOOD FOR REVIEW

## Legend:

***LE**: later edit

PR #614, opened 2018-12-14
fix(FocusZone): fix last breaking changes and make improvements when using Chat

sophieH29:
Fixes/improvements addressed to FocusZone:
- fix after breaking changes introduced in https://github.com/stardust-ui/react/pull/556 - changed selector in ```chatBehavior```
- revert https://github.com/stardust-ui/react/pull/563 - found a corner case when it can break functionality
- introduced ```shouldResetActiveElementWhenTabFromZone``` prop - when TAB from ```FocusZone```, give possibility to reset ```activeElement```. So when TAB back to ```FocusZone```, ```activeElement``` will be set to initial value, not where we left before. This behavior important for particular cases, for e.g., when using ```VirtualList``` we should always TAB back to last visible message
- moved setting ```activeElement``` from ```defaultTabbableElement``` prop to ```updateIndexes``` method. Before it was in ```componentDidMount```, which would not always work in an exact time when we expect (when using ```VirtualList``` items are rendered with delay after FocusZone's ```componentDidMount``` was called). Using it ```updateIndexes``` method will assure that activeElement was initially set, so we can update tab indexes accordingly to it.

TODO:
- [x] UPDATE CHANGELOG


src/lib/accessibility/Behaviors/Chat/chatBehavior.ts
```diff
@@ -36,8 +36,11 @@ const ChatBehavior: Accessibility = (props: any) => ({
 })
 
 const getLastTabbableElement = (root: HTMLElement): HTMLElement => {
+  const lastVisibleMessage = root.querySelector('[data-last-visible]') as HTMLElement
```
sophieH29: note: this ```chatBehavior``` would be moved from this repo in the future

jurokapsiar (approved):

src/lib/accessibility/Behaviors/Chat/chatBehavior.ts
kohlikohl: Could this be changed to
```suggestion
  const lastVisibleMessage = root.querySelector('[data-last-visible="true"]') as HTMLElement
```
When selecting for `[data-last-visible]` it would select elements that have both `data-last-visible="true"` and `data-last-visible="false"` set.

It is also hard to just render the attribute without values in react as I would need to create a new fork for the whole component. e.g.

```
if(lastVisible) {
    return <Chat data-last-visible />
} else {
    return <Chat />
}
```

src/lib/accessibility/Behaviors/Chat/chatBehavior.ts
```diff
@@ -21,7 +21,7 @@ const ChatBehavior: Accessibility = (props: any) => ({
     props: {
       shouldEnterInnerZone: event => keyboardKey.getCode(event) === keyboardKey.Enter,
       direction: FocusZoneDirection.vertical,
-      shouldHandleKeyDownCapture: false,
+      shouldResetActiveElementWhenTabFromZone: true,
```
kohlikohl, line 24: Could it be that this prop is being passed down to the DOM somehow. I am getting following error:
>warning.js?d89a:33 Warning: React does not recognize the `shouldResetActiveElementWhenTabFromZone` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `shouldresetactiveelementwhentabfromzone` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

kohlikohl (dismissed):

src/lib/accessibility/Behaviors/Chat/chatBehavior.ts
sophieH29: hmm, have you replicated all changes regarding FocusZone, together with types file? If the prop you're using is not on a list of handled props of a component, it might raise such error

sophieH29: oh, you're right, missed to set it in PropTypes. Fixed, thank you!

src/lib/accessibility/Behaviors/Chat/chatBehavior.ts
```diff
@@ -36,8 +36,11 @@ const ChatBehavior: Accessibility = (props: any) => ({
 })
 
 const getLastTabbableElement = (root: HTMLElement): HTMLElement => {
+  const lastVisibleMessage = root.querySelector('[data-last-visible="true"]') as HTMLElement
```
sophieH29, line 39: This behavior will be moved in future

src/lib/accessibility/FocusZone/FocusZone.tsx
```diff
@@ -48,7 +48,7 @@ export class FocusZone extends React.Component<FocusZoneProps> implements IFocus
     defaultTabbableElement: PropTypes.func,
     shouldFocusOnMount: PropTypes.bool,
     shouldFocusFirstElementWhenReceivedFocus: PropTypes.bool,
-    shouldHandleKeyDownCapture: PropTypes.bool,
```
mnajdova, line 51: This prop was not used anywhere? Do we have some tests that can confirm that everything is working as expected. A am scared when we make changes to some core component :)

sophieH29: yes, we do have tests :) this prop was added by me previous week, and I just reverted it as we don't need it anymore

mnajdova: Got it, thanks for the explanation!

mnajdova (approved):
Update the changelog and you are good to merge!

Issue #495, opened 2018-11-19
Docs: use TypeScript and plain JS in examples


layershifter:
### Input

- we want to write examples in TypeScript to get working all TS features and checks
- we want to show examples to user in JS and TS

### Solution

Semantic UI React has [a plugin](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/gulp/plugins/gulp-example-source.js) that simply writes sources to a JSON file.

I have an idea about similar solution, but more advanced. We can use `babel` and `babel-preset-typescript` to generate following structure:

```ts
type ComponentExampleSource = {
  js: {
    children: string,
    shorthand: string,
  },
  ts: {
    children: string,
    shorthand: string,
  },
}

type ComponentExampleSources = {
  [k: string]: ComponentExampleSource
}
```

This will give us ability to get sources of example very easily and create a toggle with JS and TS in docs. BTW, this will allow us to drop `raw-loader` 😄 

We will need to write a small plugin that will transform imports:
```js
import * as _ from 'lodash'
import * as React from 'react'
// =>
import _ from 'lodash'
import React from 'react'
```

Quick POC: https://codesandbox.io/s/yv285k54lx

hughreeling:
Or at least add Typescript specific gotya and examples.  For example:

const centered = {
  textAlign: 'center',
}

doesn't work
Should be

const centered: ICSSInJSStyle = {
  textAlign: 'center',
}

Issue #554, opened 2018-12-04
Docs example can be broken and merge into master without any check


mnajdova:
### Steps
With the latest release and the changes introduced for the Editor, we introduced problems with one of the examples for the Popup which was using react-dom. Currently this is how the example looks.
![image](https://user-images.githubusercontent.com/4512430/49442366-967c0f00-f7c9-11e8-9680-d6f28b2e20f9.png)

### Expected Result
The examples on the docs should always work. We must provide some way of testing this, not just the screens of the examples, as those are tested in an isolated (popout) mode.

### Actual Result
We may introduce problems in the docs example, and those are not caught.

### Version
0.13.1

layershifter:
We can try to render them with `SourceRender` and throw error if it happened.

levithomason:
We had a test in Semantic UI React that simply renders all examples and ensures there were no errors.  We should do the same.

Just note that the example file needs to be rendered _through_ the `ComponentExample` opposed to render the example file source directly.  This way, we are sure the example works in the editor and not just that the example itself works.

PR #617, opened 2018-12-14
docs(Examples): allow to use TS in examples

layershifter:
Refs #495. This PR is a first part.

----

This PR:
- adds ability use TypeScript in our examples via:
  - Babel plugin that transforms imports to JS-like in our examples
  - Gulp plugin that transforms our TS code to plain JS
- drops `raw-loader`

#### 1. JS examples lose some formatting.

Babel is the transpiler so some formatting things are lost.

### Input: TS code
```ts
  state = { isMounted: false }

  createdRef = React.createRef<HTMLButtonElement>()
  functionalRef = null

  handleRef = (node: HTMLButtonElement) => (this.functionalRef = node)

  componentDidMount() {
    this.setState({ isMounted: true })
  }
```

### Babel: JS output
```js
  state = {
    isMounted: false,
  }
  createdRef = React.createRef()
  functionalRef = null
  handleRef = node => (this.functionalRef = node)

  componentDidMount() {
    this.setState({
      isMounted: true,
    })
  }
```

I don't have a better idea how we can it be saved. I tried to use TS compiler there, but it gives more confusing result:

```ts
    constructor() {
        super(...arguments);
        this.state = { isMounted: false };
        this.createdRef = React.createRef();
        this.functionalRef = null;
        this.handleRef = (node) => (this.functionalRef = node);
    }
    componentDidMount() {
        this.setState({ isMounted: true });
    }
```

#### 2. There are no toggle for Typescript

It will be done in a separate PR with a new `SourceManager`.


build/gulp/plugins/gulp-example-source.ts
```diff
@@ -0,0 +1,61 @@
+import * as Babel from '@babel/core'
+import * as gutil from 'gulp-util'
+import * as path from 'path'
+import * as prettier from 'prettier'
+import * as through from 'through2'
+import * as Vinyl from 'vinyl'
+
+import * as prettierConfig from '../../../.prettierrc.json'
+import { ExampleSource } from '../../../docs/src/types'
+import transformStartImport from '../../babel/transform-star-import'
```
mnajdova: transformStartImport -> transformStarImport

build/babel/transform-star-import.ts
```diff
@@ -0,0 +1,36 @@
+import * as T from '@babel/types'
```
kuzhelov: would expect `plugin` word somewhere in the filename - this will make it easier to find them afterwards.

build/gulp/plugins/gulp-example-source.ts
layershifter: Fixed 👍

build/babel/types.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+type BabelPluginVisitorFunction<T> = (path: NodePath<T>) => void
+type BabelPluginVisitor<T> =
+  | BabelPluginVisitorFunction<T>
+  | {
+      exit: BabelPluginVisitorFunction<T>
+    }
+
+export type BabelPlugin = (
```
kuzhelov, line 15: declared name of this type doesn't correspond to its definition: while name is quite generic, the definition is very specific (as it targets import declarations only). Would suggest to make it more generic, as there are all the necessary ingredients for that

layershifter:
After #600 will be merged we can also apply ESLint's fixes and get more accurate output.

build/gulp/plugins/gulp-example-source.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    if (file.isStream()) {
+      cb(new gutil.PluginError(pluginName, 'Streaming is not supported'))
+      return
+    }
+
+    // create a base name: accordion.types.accordion.example.shorthand
+    const basename = file.path
+      .split(path.sep)
+      .slice(-3)
```
kuzhelov: why 3? What this magic number stands for? Also, the comment with example is misleading, as there will be clearly two tokens (splitted by dot) in the `basename` as a result

build/gulp/plugins/gulp-example-source.ts
```diff
@@ -0,0 +1,61 @@
...
+import * as through from 'through2'
+import * as Vinyl from 'vinyl'
+
+import * as prettierConfig from '../../../.prettierrc.json'
+import { ExampleSource } from '../../../docs/src/types'
+import transformStarImport from '../../babel/transform-star-import'
+
+const pluginName = 'gulp-example-source'
+
+const createExample = (file: Vinyl): ExampleSource => {
```
kuzhelov: `createExampleSourceCode`?

build/gulp/plugins/gulp-example-source.ts
```diff
@@ -0,0 +1,61 @@
...
+      .slice(-3)
+      .join('.')
+      .replace(/\.tsx$/, '')
+      .toLowerCase()
+    const source = createExample(file)
+
+    cb(
+      null,
+      new Vinyl({
+        path: `./${basename}.source.json`,
```
kuzhelov: could you suggest which folder this path will be relative to - and why (what defines this)? :)

kuzhelov: if this is seen to be always used as part of the `dest()` gulp's function (and, as a consequence, this function is seen to define the destination dir), then we might omit relative prefix from name

build/gulp/plugins/gulp-example-source.ts
```diff
@@ -0,0 +1,61 @@
+import * as Babel from '@babel/core'
+import * as gutil from 'gulp-util'
+import * as path from 'path'
+import * as prettier from 'prettier'
+import * as through from 'through2'
+import * as Vinyl from 'vinyl'
+
+import * as prettierConfig from '../../../.prettierrc.json'
+import { ExampleSource } from '../../../docs/src/types'
```
kuzhelov, line 8: suspicious thing is that this type is not used anywhere in docs. Is it a work left for the future (i.e. consume transform result)?

build/gulp/tasks/docs.ts
```diff
@@ -218,10 +232,12 @@ task('watch:docs', cb => {
   watch(componentsSrc, series('build:docs:docgen')).on('change', handleWatchChange)
 
   // rebuild example menus
-  watch(examplesSrc, series('build:docs:example-menu'))
+  watch(examplesIndexSrc, series('build:docs:example-menu'))
     .on('change', handleWatchChange)
     .on('unlink', path => handleWatchUnlink('example-menu', path))
 
+  watch(examplesSrc, series('build:docs:example-sources')).on('change', handleWatchChange)
```
kuzhelov: would expect to eliminate removed files from the set of those that will trigger changes as well

layershifter: Absolutely, on the first run I tried to use the model with JSON file per a component, but it broke HMR for examples.

layershifter:
### TODO for next PR

1. Remove examples from `exampleContext`, we should require only `index.tsx` files
2. Create a new context for `*.source.json` files and use it in `ExternalExampleLayout` with `SourceRender` (#554) and in `ContributionPrompt`.

build/gulp/plugins/gulp-example-source.ts
layershifter: Yes, we will use it the future.

build/babel/transform-star-import.ts
layershifter: Agree, `transform-star-import.ts` => `transform-star-import-plugin.ts`

build/babel/types.ts
layershifter: This type should be extended manually, I added a comment as we decided in offline discussion.

build/gulp/plugins/gulp-example-source.ts
layershifter: 👍

build/gulp/plugins/gulp-example-source.ts
layershifter: Sure, removed 👍

build/gulp/plugins/gulp-example-source.ts
layershifter: Reworked this part with `path.relative()`.

build/gulp/tasks/docs.ts
layershifter: Done 👍

mnajdova (approved):

build/gulp/tasks/docs.ts
```diff
@@ -62,7 +69,8 @@ task(
 
 const componentsSrc = [`${paths.posix.src()}/components/*/[A-Z]*.tsx`, '!**/Slot.tsx']
 const behaviorSrc = [`${paths.posix.src()}/lib/accessibility/Behaviors/*/[a-z]*.ts`]
-const examplesSrc = `${paths.posix.docsSrc()}/examples/*/*/*/index.tsx`
+const examplesIndexSrc = `${paths.posix.docsSrc()}/examples/*/*/*/index.tsx`
+const examplesSrc = `${paths.posix.docsSrc()}/examples/*/*/*/!(*index|.knobs).tsx`
```
kuzhelov, line 74: 👍 for filtering out `index` files

docs/src/components/ComponentDoc/ComponentExample/SourceCodeManager.ts
```diff
@@ -84,7 +84,9 @@ class SourceCodeManager {
 
   private safeRequire = (path: string): string | undefined => {
     try {
-      return require(`!raw-loader!../../../examples/${path}`)
+      const filename = `${path.replace(/^components\//, '')}.source.json`
```
kuzhelov: what this is doing?..

build/gulp/plugins/gulp-example-source.ts
```diff
@@ -0,0 +1,59 @@
...
+      cb(null, file)
+      return
+    }
+
+    if (file.isStream()) {
+      cb(new gutil.PluginError(pluginName, 'Streaming is not supported'))
+      return
+    }
+
+    const relativePath = path.relative(examplesPath, file.path)
```
kuzhelov: this logic is used couple of times (also in file unlinking handler) - maybe we could encapsulate it?

kuzhelov (approved):

build/gulp/plugins/gulp-example-source.ts
layershifter: Was moved to a separate function.

PR #618, opened 2018-12-14
feat(theme): create a base theme with Text component example

alinais:
As we will want to create new themes and most of the themes will have commonalities, there is a need to create a base theme that will keep the variables and styles interfaces/ base implementation.

Starting with Text component - the Teams theme is nicely following the base implementation while there are no visible changes in the other existing themes.

The generic base variables for text are:
```
- colors: ColorValues<string>
- disabledColor: string
- errorColor: string
- successColor: string
- five foint sizes
- five font line heights
- five font weights
```
The styles are closely following the variables and props.

Base theme text examples:
image

TODO:
- [ ] update Changelog
~- [ ] improve doc site to show only the needed Variants and examples for specific themes (no atMention, timestamp, important variants for base theme)~


src/themes/base/colors.ts
```diff
@@ -0,0 +1,128 @@
+import { ColorPalette, ContextualColors, NaturalColors } from '../types'
+
```
alinais: Not sure whether we want some base colors to be available at the level of base theme or not. Let's find the right solution.

src/themes/base/components/Text/textStyles.ts
```diff
@@ -0,0 +1,68 @@
...
+      ...(atMention === true && {
+        color: v.atMentionOtherColor,
+      }),
+      ...(atMention === 'me' && {
+        color: v.atMentionMeColor,
+        fontWeight: v.atMentionMeFontWeight,
+      }),
+      ...(disabled && { color: v.disabledColor }),
+      ...(error && { color: v.errorColor }),
+      ...(success && { color: v.successColor }),
```
alinais: It seems to me only natural that the success color should be green and I believe it would be a good idea to have it as base. Thoughts?

src/themes/base/components/Text/textStyles.ts
```diff
@@ -0,0 +1,68 @@
...
+      ...(temporary && { fontStyle: 'italic' }),
+      ...(important && {
+        fontWeight: v.importantWeight,
+        color: v.importantColor,
+      }),
+
+      ...(weight === 'light' && {
```
alinais: What would be the chance that the weight would differ from one theme to another?

src/themes/base/siteVariables.ts
```diff
@@ -0,0 +1,43 @@
...
+export const yellow = colors.yellow[500]
+export const green = colors.green[500]
+export const grey = colors.grey[500]
+export const blue = colors.blue[500]
+
+//
+// FONT SIZES
+//
+export const fontSizes = {
+  smaller: pxToRem(10),
```
alinais: Should these be specific for each theme?

src/themes/teams/index.ts
```diff
@@ -34,11 +37,11 @@ Object.keys(fontIcons).forEach(iconName => {
   icons[iconName] = declareFontBased(fontIcons[iconName])
 })
 
-export default {
+export default mergeThemes(base, {
```
mnajdova, line 40: Shouldn't we merge here the default theme?

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -663,6 +667,11 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
         <Divider section horizontal />
       </Visibility>
     )
+
+    if (excludedThemes && excludedThemes.indexOf(themeName) > -1) {
```
miroslavstastny: In general, this check should be done at the beginning of the function - all the information are available there, there is no need to compute the whole result and then return ''.

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -549,7 +552,7 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
   }
 
   render() {
-    const { children, description, title } = this.props
+    const { children, description, title, excludedThemes } = this.props
```
miroslavstastny: Docsite should document API - for that reason I would not hide examples based on theme capabilities.

In color palette PR, we had a long discussion about themes opting out/in for some library features and decided not to do that now as it can be tricky.

src/themes/teams/components/Text/textVariables.ts
```diff
@@ -1,38 +1,17 @@
-import { ColorValues } from '../../../types'
 import { mapColorsToScheme } from '../../../../lib'
+import { TextVariables } from '../../../base/components/Text/textVariables'
 
-export interface TextVariables {
-  colors: ColorValues<string>
+export interface TeamsTextVariables extends TextVariables {
```
miroslavstastny, line 4: Why are not the remaining variables (`atMentions*`, `important*`, `timestamp*`) part of the base theme?
I think that what is in the component API (these are) it should be in base theme as well.

src/themes/teams/components/Text/textStyles.ts
```diff
@@ -21,7 +21,7 @@ export default {
       temporary,
     },
     variables: v,
-  }: ComponentStyleFunctionParam<TextProps, TextVariables>): ICSSInJSStyle => {
+  }: ComponentStyleFunctionParam<TextProps, TeamsTextVariables>): ICSSInJSStyle => {
```
miroslavstastny, line 9: Remove parts which are already handled by base theme. Teams theme is merged on top of base theme, it should only (re)define variables and overwrite styles where necessary.

miroslavstastny (commented):
Thinking about this from consumers point of view - what is a theme?
1. Is there one fixed component API which I use and then I can freely change themes to change the visual appearance?
2. Or can the theme "extend" component API which I can then use?

The first option it the one we are currently implementing. Take `atMention` as an example:
1. component API implements `atMention` prop,
2. base theme does not handle that prop,
3. teams theme handles the prop.

I think that whatever is in component's API should be handled by the base theme.
Is `atMention` too specific not to be part of the base theme? So why is that in the component's API?

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
kuzhelov: agree to omit this `exclusion` functionality - in contrast, we should rather think about 'extending' approach, where there will be some base theme, and each new one will be able to include its own examples (like `inherit color` which we have for Teams theme). But, once again, it should be an approach that is driven by `extensibility`, not `exclusion` - thus I agree with suggestion to remove these changes

src/themes/teams/components/Text/textStyles.ts
alinais: I removed the duplicated code from styles side. As for variables, the color palettes and the font sizes are different, so keeping the changes in the teams theme as well.

alinais:
We got to the conclusion that as long as the API is exposing some props, the theme should have some minimal styling for it. For the future we should think about creating a simpler API for the base theme that other themes can extend and add its own props.

src/themes/teams/components/Text/textStyles.ts
```diff
@@ -35,7 +20,6 @@ export default {
       ...(disabled && { color: v.disabledColor }),
```
mnajdova: What is the reason that we still have these styles here (for the disabled, error, success etc) that we already have in the base team. Wouldn't it be enough to just change the variables values in teams' theme if they are different?

src/themes/teams/components/Text/textStyles.ts
```diff
@@ -1,30 +1,15 @@
...
-      important,
-      success,
-      timestamp,
-      truncated,
-      temporary,
-    },
+    props: { atMention, color, disabled, error, important, success, timestamp },
     variables: v,
-  }: ComponentStyleFunctionParam<TextProps, TextVariables>): ICSSInJSStyle => {
+  }: ComponentStyleFunctionParam<TextProps, TeamsTextVariables>): ICSSInJSStyle => {
```
mnajdova, line 9: One question, just so that I understand better. If this theme is designed so that it would be merged with the base team, shouldn't the variables here be TextVariables? Or we are designing this as it could also be a standalone theme? (One scenario would be, let's say, I want to use some additional styling using some variable from the base theme, because I know that this theme will always be merged on top of that one..) Not sure if this is valid, but I just want to understand how we are going to use them, because if we depend on that and do not merge it with base, then we will have some failings..

src/themes/teams/components/Text/textStyles.ts
alinais: Missed this one, fixing it.

src/themes/teams/components/Text/textStyles.ts
mnajdova: Disregard my comment, I didn't see that `TeamsTextVariables` is extending the `TeamsVariables`.

mnajdova:
Totally agree with @miroslavstastny points. If something is in the API then it **should be handled in the base theme** (by providing variables for it, and basic styles). Other themes can **extend or change** the variables describing some prop and can add additional styling for them. If we have the important property for example, the base theme can for example make it just bold, and the teams theme can add additional variable for the color etc. The Text is pretty simple component, so not sure how many variables will be added in the teams theme, but if we take a look into the Chat for example, we will have base theme that can create some black-gray-white chat, and teams theme can change the variables to use the specific colors and maybe add additional ones for customizing the styles further.

Agreed that the atMention and timestamp props maybe should be removed from the API, but that is not the scope of this PR for sure...

src/themes/teams/components/Text/textStyles.ts
```diff
@@ -47,43 +25,6 @@ export default {
         color: v.importantColor,
       }),
       ...(color && { color: _.get(v.colors, color) }),
```
miroslavstastny: remove this line, it is already in base theme

src/themes/teams/components/Text/textStyles.ts
```diff
@@ -47,43 +25,6 @@ export default {
         color: v.importantColor,
```
miroslavstastny: remove this line, it is already in base theme

src/themes/teams/components/Text/textStyles.ts
```diff
@@ -1,41 +1,19 @@
...
       ...(atMention === 'me' && {
         color: v.atMentionMeColor,
         fontWeight: v.atMentionMeFontWeight,
       }),
-      ...(disabled && { color: v.disabledColor }),
-      ...(error && { color: v.errorColor }),
-      ...(success && { color: v.successColor }),
-      ...(temporary && { fontStyle: 'italic' }),
       ...(timestamp && {
         color: v.timestampColor,
```
miroslavstastny: remove this line, it is already in base theme

src/themes/teams/components/Text/textStyles.ts
```diff
@@ -1,41 +1,19 @@
...
-  }: ComponentStyleFunctionParam<TextProps, TextVariables>): ICSSInJSStyle => {
+  }: ComponentStyleFunctionParam<TextProps, TeamsTextVariables>): ICSSInJSStyle => {
     return {
-      display: 'inline-block',
-      ...(truncated && truncateStyle),
-      ...(atMention === true && {
-        color: v.atMentionOtherColor,
-      }),
       ...(atMention === 'me' && {
         color: v.atMentionMeColor,
```
miroslavstastny: remove this line, it is already in base theme

mnajdova (approved):
🎁 for 🎄

PR #621, opened 2018-12-17
chore: cache results of vulnerability scans

kuzhelov:
Provided changes introduce caching strategy for vulnerability tests - this caching will allow us to stay within monthly quota limit (200 runs per month, while expectation for introduced strategy is 30-60).

### Caching Strategy
The strategy is the following:
- for each yarn.lock
  - run vulnerability scan if there were no any made today, cache result
  - otherwise (if today's cached result present) just skip vulnerability tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,12 +51,12 @@ jobs:
           command: bash <(curl -s https://codecov.io/bash)
 
       - restore_cache:
-          key: v1-vuln-scans
+          key: v1-vuln-scans-{{ checksum "yarn.lock" }}
       - run:
           name: Vulnerability Tests
           command: yarn test:vulns
       - save_cache:
-          key: v1-vuln-scans
+          key: v1-vuln-scans-{{ checksum "yarn.lock" }}
           paths:
             - .vuln-scans
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,22 +21,6 @@ jobs:
             echo 'export PATH="${PATH}:${HOME}/.yarn/bin"' >> $BASH_ENV
       - checkout
       # because we don't invoke npm (we use yarn) we need to add npm bin to PATH manually
-      - run:
-          name: Add npm bin to PATH
-          command: echo 'export PATH="${PATH}:$(npm bin)"' >> $BASH_ENV
-      - restore_cache:
-          keys:
-            - v1-dependencies-{{ checksum "yarn.lock" }}
-            - v1-dependencies
-      - run:
-          name: Install Dependencies
-          command: yarn
-      - save_cache:
-          key: v1-dependencies-{{ checksum "yarn.lock" }}
-          paths:
-            - ~/.cache/yarn
-            - .yarn-cache
-            - node_modules
 
       - restore_cache:
           key: v1-vuln-scans-{{ checksum "yarn.lock" }}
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -49,9 +49,17 @@ jobs:
...
           name: Report coverage
           command: bash <(curl -s https://codecov.io/bash)
+
+      - restore_cache:
+          key: v1-vuln-scans-{{ checksum "yarn.lock" }}
       - run:
           name: Vulnerability Tests
           command: yarn test:vulns
+      - save_cache:
+          key: v1-vuln-scans-{{ checksum "yarn.lock" }}-{{ epoch }}
```
layershifter, line 60: Is it okay that we restoring a cache with a key without the `epoch` variable?

build/gulp/tasks/test-vulns.ts
```diff
@@ -0,0 +1,65 @@
...
+const { paths } = config
+
+const SCAN_RESULTS_DIR_NAME = '.vuln-scans'
+
+const log = message => debug.log(message)
+log.success = message => debug.log(`✔ ${message}`)
+
+const ensureDirExists = path => {
+  if (!fs.existsSync(path)) {
+    sh(`mkdir -p ${path}`)
```
layershifter, line 19: `fs.mkdirSync(path, { recursive: true })`?

build/gulp/tasks/test-vulns.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+}
+
+const getScanResultsDirPath = () => {
+  return paths.base(SCAN_RESULTS_DIR_NAME)
+}
```
layershifter: Do we actually need it? As I see it can be a simple variable

build/gulp/tasks/test-vulns.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const getTodayScanFilePath = () => {
+  const now = new Date()
+
+  const fileName = `snyk-scanned-${now.getUTCFullYear()}-${now.getUTCMonth() +
+    1}-${now.getUTCDate()}`
```
layershifter: Can we split this to multiple variables? To avoid multiline there

layershifter (approved):
I left small questions there. I prefer to use pure bash commands for such things, but it will be hard to debug them for others.

Let's ship it.

.circleci/config.yml
kuzhelov: this is necessary because otherwise Circle CI won't update the cache entry if it has existed before (this is its feature). To avoid this 'rewrite ban' the following strategy was suggested: https://discuss.circleci.com/t/add-mechanism-to-update-existing-cache-key/9014/12

layershifter: As we discussed, please add a small comment with this link before the `key` line 👍

build/gulp/tasks/test-vulns.ts
kuzhelov: for that we need Node LTS v10 (and this is, actually, the step we **should** make). Agreed to defer it to the follow-up PR, to be absolutely sure that all the necessary accompanying adjustments to the code will be made

.circleci/config.yml
kuzhelov: agreed to introduce comment for that

Issue #620, opened 2018-12-17
Show popup on trigger hover


alinais:
### Problem description
There are cases when you hover on the trigger of the Popup and you'd like to see it. This item is supposed to implement this work.

### MVP
[Fork, update, and replace this pen if you can show the proposed solution]:
https://codesandbox.io/s/2l3n74j9y

PR #622, opened 2018-12-17
feat(Popup): added `on` prop for triggering the popup on click, hover or focus

mnajdova:
This PR aims to solve https://github.com/stardust-ui/react/issues/620 by adding `on` prop on the Popup, that can receive one of the values `click`, `hover`, `focus`, or array of these (except the `click` and `hover` cannot be defined at the same time), thus allowing the user to specify the event that should trigger the popup. In addition to this, there is new property added `mouseLeaveDelay` that is representing the delay in ms for the mouse leave event, before the popup will be closed (There must be some timeout before the mouse is moved form the trigger to the content of the popup, and the popup should not be closed in that period. This property is defining how long that timeout should be - the default is 500ms).

TODO:
- [x]  escape should close the Popup when the on property's click is specified 
- [x]  escape should close the Popup when the on property's focus is specified 
- [x]  when the `hover` value is specified, because of accessibility consistency, we have to open the popup on click (enter/space)
- [ ]  escape should closed the Popup when  the on property's the hover is specified (we should subscribe to the EventStack keyDown for this). This will be addressed in a separate PR, alongside with the other cases of handling the escape pressed key.
- [ ]  focus trap zone with focusable element inside the content's content doesn't work with focus, as focusing back the element is opening the Popup again. Adressed in a follow up PR with together with the escape key handling.


src/components/Popup/Popup.tsx
```diff
@@ -224,10 +256,54 @@ export default class Popup extends AutoControlledComponent<Extendable<PopupProps
...
+
+    const { on } = this.props
+
+    if (on === 'click') {
+      triggerProps.onClick = (e, ...rest) => {
+        this.trySetOpen(!this.state.open, e)
+        _.invoke(triggerElement, 'props.onClick', e, ...rest)
+      }
+    }
+    if (on === 'hover') {
```
jurokapsiar: this might be for a separate PR, but if popup appears on hover, it also needs to appear on click for the case when user uses keyboard or virtual cursor to access the popup

miroslavstastny: On `click` or `focus`? And how can I close that popup?
FWIW, Semantic displays popup on space or Enter with `on='hover'`, it is possible to close it with Esc only (no Enter/space).
With `on='click'` Enter/space can both open and close the popup.
It also supports `on='focus'` (https://react.semantic-ui.com/modules/popup/#usage-focus) and array for `on` prop.

src/components/Popup/Popup.tsx
```diff
@@ -54,6 +54,9 @@ export interface PopupProps
   /** Initial value for 'open'. */
   defaultOpen?: boolean
 
+  /** Delay in ms for the mouse leave event, before the popup will be closed. */
+  mouseLeaveDelay?: number
```
miroslavstastny, line 64: 👍

src/components/Popup/Popup.tsx
mnajdova: Couple of thoughts from my side:
I started with the ideas from Semantic UI, but there are problems when for some popup we have the `hover` and `click` **at the same time** (awkward things are happening because we have both click and hover opening/closing the popup because, so if you go on the popup to click it in order for it to be opened, the hovered is already opening it etc - please try this on SUIR)
I cannot imagine popup that would need to implement the click and hover at the same time (maybe I am wrong with this assumption)
The focus on the other hand, should follow the (click or hover triggering) - I would image that popup with hover would also be opened if we focus it and popup with click event would be opened if we focus on it and press enter/space. 
@miroslavstastny what do you think about this points? @jurokapsiar can you reference some best practices about how should this work?

jurokapsiar: unfortunately, the design pattern (https://www.w3.org/TR/wai-aria-practices-1.1/#tooltip) is not finalized yet, in fact it does not include sufficient amount of information.

Both options that @miroslavstastny mentioned are valid for a popup that appears on hover:
1. usually for popups that are interactive (include focusable elements): when the user **clicks** on the trigger and the popup is not yet opened it should open (mainly for the virtual cursor which does not have hover state and keyboard use case). Such popups usually also grab and trap focus.
2. usually for popups that are not interactive (tooltip): when the user **focuses** the trigger, popup should open. This does not work correctly with virtual cursor since in some combinations, the focus event is not sent, so the information from the tooltip is often included in the aria-label

Esc should always close the popup, which I think we do not have implemented yet.
Enter and Space should perform the same action as click on the focused element - if the trigger is focused, it should open/close the popup (but that is outside of the scope of this PR)

Also, I would consider the tooltip use case as out of scope and would recommend to concentrate on option 1 first.

mnajdova: So, in conclusion are we okay with following the SUIR practices? (Still not sure tough for the hover + click at the same time)

mnajdova: @layershifter is ok with excluding the combination of having click and hover at the same time. Will proceed with this, and the rest will follow the implementation in suir

src/components/Popup/Popup.tsx
```diff
@@ -130,12 +142,36 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
...
+
+  schedulePopupClose = e => {
+    const { mouseLeaveDelay } = this.props
+    this.isPopupClosing = true
+    setTimeout(() => {
+      if (this.isPopupClosing) {
+        this.trySetOpen(false, e)
+      }
+
+      this.isPopupClosing = false
```
sophieH29: nit: to move this assignment under `if` condition. 

```js
 if (this.isPopupClosing) {
         this.trySetOpen(false, e)
        this.isPopupClosing = false
   } 
```
So if `this.isPopupClosing` === `false`, you don't need to re-assign it again to `false`

src/components/Popup/Popup.tsx
```diff
@@ -130,12 +142,36 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
...
+    } else {
+      this.isPopupClosing = false
+      this.trySetOpen(true, e)
+    }
+  }
+
+  schedulePopupClose = e => {
+    const { mouseLeaveDelay } = this.props
+    this.isPopupClosing = true
+    setTimeout(() => {
```
sophieH29: if `mouseLeaveDelay` === `0` maybe we don't need to use `setTimeout` at all? As it still causes some delay.

src/components/Popup/Popup.tsx
```diff
@@ -146,15 +182,19 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
...
       e.stopPropagation()
     },
   }
 
-  private closeAndFocusTrigger = e => {
+  private closeAndFocusTrigger = (e, focusTrigger) => {
     if (this.state.open) {
       this.trySetOpen(false, e, true)
-      _.invoke(this.triggerDomElement, 'focus')
+      focusTrigger && _.invoke(this.triggerDomElement, 'focus')
```
sophieH29: seems to be a bit confusing having name of function `closeAndFocusTrigger`, but it does what it says only in one case. What do you if to change it similar to `setState` structure with callback?

```js
...
...
    closeAndFocusTrigger: e => {
      this.close(e, () => _.invoke(this.triggerDomElement, 'focus'))
      e.stopPropagation()
    },
    close: e => {
      this.close(e)
      e.stopPropagation()
    },
  }
...

private close = (e, onClose?: () => void) => {
    if (this.state.open) {
      this.trySetOpen(false, e, true)
      onClose && onClose()
    }
  }
```

src/components/Popup/Popup.tsx
```diff
@@ -224,10 +262,94 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
...
+        this.setPopupOpen(true, e)
+        _.invoke(triggerElement, 'props.onMouseEnter', e, ...rest)
+      }
+      triggerProps.onMouseLeave = (e, ...rest) => {
+        this.setPopupOpen(false, e)
+        _.invoke(triggerElement, 'props.onMouseLeave', e, ...rest)
+      }
+      triggerProps.onClick = (e, ...rest) => {
+        this.setPopupOpen(true, e)
+        _.invoke(triggerElement, 'props.onClick', e, ...rest)
```
sophieH29, line 293: From an accessibility perspective, shouldn't we open Popup which opens on 'hover', also on 'focus', not on 'click'?

I am thinking about a case when we show a tooltip with additional information on hover. How a keyboard user will guess that he needs to click to see that info? I would expect to show it on focus

mnajdova: No, after talking with @jurokapsiar the decision was, the focus should not be default for Popups that have on hover (it can be optionally added with using `on={['focus', 'hover']}`). The hover Popups, should always be opened on click (this is not the toggling - just simple open on click) - so that the virtual cursors can first focus and then open the popup.

src/components/Popup/Popup.tsx
mnajdova: Will refactor it a bit.

src/components/Popup/Popup.tsx
```diff
@@ -119,6 +125,12 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
     align: PropTypes.oneOf(ALIGNMENTS),
     defaultOpen: PropTypes.bool,
     defaultTarget: PropTypes.any,
+    mouseLeaveDelay: PropTypes.number,
+    on: PropTypes.oneOfType([
+      PropTypes.oneOf(['hover', 'click', 'focus']),
+      PropTypes.arrayOf(PropTypes.oneOf(['click', 'focus'])),
+      PropTypes.arrayOf(PropTypes.oneOf(['hover', 'focus'])),
+    ]),
```
sophieH29, line 139: it might worth to add an example on using a combination of these values.

docs/src/examples/components/Popup/Usage/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+      title="Click"
+      description="A popup can be triggered on click."
+      examplePath="components/Popup/Usage/PopupExampleClick"
+    />
+    <ComponentExample
+      title="Focus"
+      description="A popup can be triggered on focus."
+      examplePath="components/Popup/Usage/PopupExampleFocus"
+    />
+  </ExampleSection>
```
layershifter, line 23: Do we really need a three separated examples for this? My vote:
- add an example that contains three variants of simple popups
- add an example with multiple `on`

src/components/Popup/Popup.tsx
mnajdova: Addressed, thanks!

docs/src/examples/components/Popup/Usage/index.tsx
mnajdova: Addressed!

docs/src/examples/components/Popup/Usage/PopupExampleOnWithFocusTrap.shorthand.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <Button primary>Second</Button>
+    </>
+  ),
+}
+
+const PopupExampleOnWithFocusTrap = () => (
+  <Grid columns={3} variables={{ gridGap: '10px' }}>
+    <Text content={'Click'} weight="bold" />
+    <Text content={'Hover'} weight="bold" />
+    <Text content={'Focus'} weight="bold" />
```
layershifter: Oops, useless `{}` in `content`

mnajdova: Addressed, thanks!

src/components/Popup/Popup.tsx
miroslavstastny: I would disagree with this. Code should be either always sync or always async. Making it conditionally async can be surprising and difficult to debug.

mnajdova: Changed back the initial logic for always setting up a timeout.

docs/src/examples/components/Popup/Usage/PopupExampleOnWithFocusTrap.shorthand.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react'
+import { Button, Popup, Grid, Text, popupFocusTrapBehavior } from '@stardust-ui/react'
+
+const contentWithButtons = {
+  content: (
+    <>
+      <Button onClick={() => alert('1')}>First</Button>
```
layershifter: Looks like a forgotten debug statement

src/components/Popup/Popup.tsx
```diff
@@ -63,6 +71,9 @@ export interface PopupProps
    */
   offset?: string
 
+  /** Events triggering the popup. */
+  on?: PopupEvents | PopupEventsArray
```
layershifter, line 76: This type allows this:

```ts
on={['click', 'click']}
```

[TS playground](http://www.typescriptlang.org/play/#src=export%20type%20PopupEvents%20%3D%20'click'%20%7C%20'hover'%20%7C%20'focus'%0D%0Aexport%20type%20RestrictedClickEvents%20%3D%20'click'%20%7C%20'focus'%0D%0Aexport%20type%20RestrictedHoverEvents%20%3D%20'hover'%20%7C%20'focus'%0D%0Aexport%20type%20PopupEventsArray%20%3D%20RestrictedClickEvents%5B%5D%20%7C%20RestrictedHoverEvents%5B%5D%0D%0A%0D%0Atype%20OnProp%20%3D%20%20PopupEvents%20%7C%20PopupEventsArray%0D%0A%0D%0Aconst%20a%3A%20OnProp%20%3D%20%5B'click'%2C%20'click'%5D%0D%0A)

src/components/Popup/Popup.tsx
layershifter: We should clear this timeout in `componentWillUnmount()`, otherwise we will receive a warning:
```
Warning: setState(…): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component.
```

src/components/Popup/Popup.tsx
mnajdova: I want to restrict using 'click' together with 'hover'. Adding the array ['click', 'click'] is not a bug, so I don't know it we want to add some other restrictions for this.

docs/src/examples/components/Popup/Usage/PopupExampleOnWithFocusTrap.shorthand.tsx
mnajdova: We want to show here that opening alert will not close the popup, that's why we have this, but I am okay with removing it, or changing the content of it. What do you think?

layershifter: My opinion: we should avoid the usage of `alert()` in docs.

mnajdova: Removing it, it didn't bring much value anyways :)

src/components/Popup/Popup.tsx
mnajdova: Added logic for clearing the timeout, thanks!

docs/src/examples/components/Popup/Usage/PopupExampleOnMultiple.shorthand.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react'
+import { Button, Popup } from '@stardust-ui/react'
+
+const PopupExampleOnMultiple = () => (
+  <div>
+    <Popup
+      trigger={<Button icon="expand" content="Click + Focus" aria-label="Click or focus button" />}
+      content="Hello from popup on click!"
+      on={['click', 'focus']}
```
levithomason, line 9: This doesn't work as expected.  Clicking the button makes the popup appear for a short interval and close.  This is because the click invokes an open then the element receives focus which immediately invokes a close.

levithomason (dismissed):

docs/src/examples/components/Popup/Usage/PopupExampleOnMultiple.shorthand.tsx
mnajdova: Thanks for the catch. The problem was that the focus was opening the popup always, but should be only if it comes from the keyboard. It is fixed now.
popup-fix

miroslavstastny (approved):

PR #624, opened 2018-12-17
Chat header prototype

kolaps33:
I am trying to make a different variants of chat header V2.

I am hitting following issues which I need to help with:

resolved by hack:
- using the list as horizontal list, approached by using  float: 'left' on list item
- for accessibility we want to have list of buttons, hard coding in ListItem.tsx  into the 'header' attribute <div> with role button 

what I could not resolved:
- general approach how to handle CSS styles for components
- menu which visually fit for specification is underlined, how to remove underline?
- when menu is open then longer user name is not wrapped correctly... 

  
chat-header-variants


src/lib/accessibility/Behaviors/List/basicListBehavior.ts
```diff
@@ -14,6 +14,14 @@ const basicListBehavior: Accessibility = (props: any) => ({
       role: 'list',
     },
   },
+  focusZone: {
```
jurokapsiar: create a new NavigableList behavior isntead of changing the basicListBehavior

miroslavstastny:
List of identified issues/limitations in Stardust:

- [ ] horizontal list
- [ ] list item (ItemLayout) does not accept shorthand
- [ ] ListItem should have a wrapper (proposal PR: #688)
- [ ] circular button with icon and text
- [ ] underlined menu - why is there the second underline (bellow the whole menu)?
- [ ] underlined menu - space between active text and underline - bigger in v2
- [ ] button with bold text

PR #627, opened 2018-12-18
fix(FocusTrapZone): Do not focus trigger on outside click

sophieH29:
This PR gives possibility for end user to define if the trigger element should be focused on outside click, by using prop ```focusTriggerOnOutsideClick```. The value is ```false``` by default, as for most cases ```Popup``` will use ```FocusTrapZone```, and on outside click, we close Popup and do not want the trigger to be focused, only if Popup was closed using keyboard keys.

TODO: 
- [x] update FTZ changelog


src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -45,6 +45,7 @@ export class FocusTrapZone extends React.Component<FocusTrapZoneProps, {}> {
     firstFocusableSelector: PropTypes.string,
     disableFirstFocus: PropTypes.bool,
     focusPreviouslyFocusedInnerElement: PropTypes.bool,
+    focusTriggerOnOutsideClick: PropTypes.bool,
```
jurokapsiar, line 50: please add an entry to the FocusZone CHANGELOG

jurokapsiar (approved):

docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.shorthand.tsx
```diff
@@ -24,6 +24,10 @@ class PortalExampleFocusTrapped extends React.Component {
             // When 'false', all clicks outside the Portal will be caught and not handled.
             // 'true' by default.
             isClickableOutsideFocusTrap: false,
+            // Indicates whether to focus element which triggered FTZ on outside click
+            // 'false' by default
+            // note: if isClickableOutsideFocusTrap === 'false', focusTriggerOnOutsideClick will not be taken into account.
```
miroslavstastny, line 29: the example is strange as it does exactly this (`isClickableOutsideFocusTrap: false`)

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -233,7 +252,7 @@ export class FocusTrapZone extends React.Component<FocusTrapZoneProps, {}> {
       })
     }
 
-    if (!isClickableOutsideFocusTrap) {
+    if (!isClickableOutsideFocusTrap || !focusTriggerOnOutsideClick) {
```
kuzhelov: it is better to encapsulate this condition under single expression that will be used in both places, like

```js
private shouldHandleOutsideClick = () => 
  !this.props.isClickableOutsideFocusTrap || !this.props.focusTriggerOnOutsideClick

...

if (this.shouldHandleOutsideClick()) {
  ...
}
```

This move will ensure consistency of this policy applied at several places.

docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.tsx
```diff
@@ -25,6 +25,10 @@ class PortalExamplePortal extends React.Component {
             // When 'false', all clicks outside the Portal will be caught and not handled.
             // 'true' by default.
             isClickableOutsideFocusTrap: false,
+            // Indicates whether to focus element which triggered FTZ on outside click
+            // 'false' by default
+            // note: if isClickableOutsideFocusTrap === 'false', focusTriggerOnOutsideClick will not be taken into account.
+            focusTriggerOnOutsideClick: false,
```
kuzhelov, line 31: should be removed, as it doesn't take any effect for this example

docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.shorthand.tsx
sophieH29: this example just shows possible API that can be used for FocusTrapZone as there is no separate component as ```FocusTrapZone``` that clients can lookup for props, unless they'll check code directly.
And it also gives a possibility to play with it, by setting different values for props.
Let me know what you think @miroslavstastny @kuzhelov

docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.tsx
sophieH29: this example just shows possible API that can be used for FocusTrapZone as there is no separate component as ```FocusTrapZone``` that clients can lookup for props, unless they'll check code directly.
And it also gives a possibility to play with it, by setting different values for props.
Let me know what you think @miroslavstastny @kuzhelov

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: make sense, will change

docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.tsx
sophieH29: I'll create an issue for creating proper documentation with examples for FoucsTrapZone and FocusZone. But would leave this change as it is.

hughreeling (approved):

silviuaavram (approved):

alinais (approved):

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -225,20 +228,33 @@ export class FocusTrapZone extends React.Component<FocusTrapZoneProps, {}> {
...
+      clickedElement && this._forceFocusInTrap(ev, clickedElement)
+    } else if (!focusTriggerOnOutsideClick) {
+      const isOutsideFocusTrapZone =
+        this._root.current && !this._root.current.contains(clickedElement)
+      const isOutsideTriggerElement =
+        this._previouslyFocusedElementOutsideTrapZone &&
+        !this._previouslyFocusedElementOutsideTrapZone.contains(clickedElement)
+      if (isOutsideFocusTrapZone && isOutsideTriggerElement) {
+        // set it to NULL, so the trigger will not be focused on componentWillUnmount
+        this._previouslyFocusedElementOutsideTrapZone = null
```
kuzhelov, line 243: here we are implicitly assume that the element that was previously focused is the `trigger` element - while, in general case, it might be not the case. 

The problem I see is that this `focusTriggerOnOutsideClick ` prop targets specific set up and, in general case, might not work as expected by the client. For example, suppose that element other than `_previouslyFocusedElementOutsideTrapZone` has triggered focus zone - in that case `focusTriggerOnOutsideClick` will not work as expected, as `_previouslyFocusedElementOutsideTrapZone` !== `triggerElement`.

Would suggest to rename this prop in a way that it won't use `trigger` word - please, let me know if I am missing something (is it guaranteed that `trigger` is previously focused element in all cases?), but at the moment am pretty sure that `trigger` is a concept that `FocusTrapZone` cannot control anyhow.

kuzhelov (approved):
please, consider to review this comment before merging: https://github.com/stardust-ui/react/pull/627/files#r246936199

src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: you're right, it can be not always an element which triggered focus zone, if a user sets the prop `elementToFocusOnDismiss`, it will replace the trigger element. That seems the only case.
To be honest I have no idea how to name it differently
Something like `focusElementOutsideFTZOnOutsideClick` or `focusPrevioslyFocusedElementOnOutsideClick`? What seems to me even harder to understand :)

I will merge the PR now. If we come up with better idea on naming, it's quick to create a small PR on renaming it.

PR #628, opened 2018-12-18
feat(header): header and header description color prop

bmdalex:
### Description

This PR:
- adds `color` prop to `Header` and `HeaderDescription` components
- creates `Header` color examples

### API

```jsx
<Header color={COLOR} description={{ color: COLOR }}/>
// or
<Header color={COLOR} variables={{ descriptionColor: COLOR }}/>
```

screenshot 2018-12-18 at 15 13 32

### DIFF
--- a/src/themes/teams/components/Header/headerDescriptionStyles.ts
+++ b/src/themes/teams/components/Header/headerDescriptionStyles.ts
@@ -2,10 +2,10 @@ import * as _ from 'lodash'
 
 import { pxToRem } from '../../utils'
 import { ICSSInJSStyle, ComponentSlotStylesInput } from '../../../types'
-import { HeaderProps } from '../../../../components/Header/Header'
-import { HeaderVariables } from './headerVariables'
+import { HeaderDescriptionProps } from '../../../../components/Header/HeaderDescription'
+import { HeaderDescriptionVariables } from './headerDescriptionVariables'
 
-const headerStyles: ComponentSlotStylesInput<HeaderProps, HeaderVariables> = {
+const headerStyles: ComponentSlotStylesInput<HeaderDescriptionProps, HeaderDescriptionVariables> = {
   root: ({ props: p, variables: v }): ICSSInJSStyle => ({
     display: 'block',
     color: _.get(v.colors, p.color, v.color),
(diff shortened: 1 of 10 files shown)


src/themes/teams/components/Header/headerDescriptionStyles.ts
```diff
@@ -1,11 +1,17 @@
...
+import { ICSSInJSStyle, ComponentSlotStylesInput } from '../../../types'
+import { HeaderProps } from '../../../../components/Header/Header'
+import { HeaderVariables } from './headerVariables'
 
-export default {
-  root: ({ variables: v }): ICSSInJSStyle => ({
+const headerStyles: ComponentSlotStylesInput<HeaderProps, HeaderVariables> = {
+  root: ({ props: p, variables: v }): ICSSInJSStyle => ({
     display: 'block',
+    color: _.get(v.colors, p.color, v.color),
```
layershifter, line 11: I don't know what we can do there, but this line looks very confusing 🤔

layershifter (approved):
This continues the work from #597. We will idea with a color scheme to a separate PR.

src/themes/teams/components/Header/headerDescriptionStyles.ts
bmdalex: agreed, but not sure what we can do either.. 😢

src/themes/teams/components/Header/headerDescriptionStyles.ts
```diff
@@ -1,11 +1,17 @@
...
+
 import { pxToRem } from '../../utils'
-import { ICSSInJSStyle } from '../../../types'
+import { ICSSInJSStyle, ComponentSlotStylesInput } from '../../../types'
+import { HeaderProps } from '../../../../components/Header/Header'
+import { HeaderVariables } from './headerVariables'
 
-export default {
-  root: ({ variables: v }): ICSSInJSStyle => ({
+const headerStyles: ComponentSlotStylesInput<HeaderProps, HeaderVariables> = {
```
kuzhelov: `HeaderDescriptionVariables` should be used as a type here

kuzhelov: same applies to prop types, actually

kuzhelov (changes requested):
lets address typings issue before merging

src/themes/teams/components/Header/headerDescriptionStyles.ts
bmdalex: nice catch, fixed

kuzhelov (approved):

Issue #629, opened 2018-12-18
[RFC]: Colors - choosing text color for box components (where color props sets background color)


bmdalex:
### Problem description
We need to automatically change text color for box components (components where `color` prop goes to background color)

What tool to use for deciding text color for different background shades?

### Proposed solution
Use `color` package and `Color(bgColor).isDark` call to determine if the text is light or dark, but it looks like this is not a preferred package to import? see:
https://github.com/stardust-ui/react/pull/597#discussion_r241344252

bmdalex:
**Q: How do we choose the shades of light and dark text colors?**

A: We need to:
- wait for designers to give us the right redlines for what text colors to use;
- we need to create more proposals and iterate on them, @levithomason would be great if you could summarize your thoughts on these proposals

layershifter:
> We need to automatically change text color for box components

I think that we don't need to do something automatically, we should define mapping explicitly. As I see, described issue should be covered by #636.

bmdalex:
closing as this is covered by #636 now

PR #632, opened 2018-12-18
feat(segment): color prop

bmdalex:
### Description

This PR:
- adds `color` prop to `Segment` component
- creates `Segment` color examples

### API

```jsx
<Segment color={COLOR} .../>
```

screenshot 2018-12-18 at 16 03 20

### Known issues:

Segment text color should be dark shade of gray on segments with light background color, tracked by #629

### DIFF
--- a/docs/src/examples/components/Segment/Variations/SegmentExampleColor.shorthand.tsx
+++ b/docs/src/examples/components/Segment/Variations/SegmentExampleColor.shorthand.tsx
@@ -1,0 +1,18 @@
+import React from 'react'
+import _ from 'lodash'
+import { Segment, ProviderConsumer } from '@stardust-ui/react'
+
+const SegmentExampleColor = () => (
+  <ProviderConsumer
+    render={({ siteVariables: { emphasisColors, naturalColors } }) =>
+      _.keys({ ...emphasisColors, ...naturalColors }).map(color => (
+        <>
+          <Segment key={color} color={color} content={_.startCase(color)} inverted />
+          <br />
+        </>
+      ))
+    }
+  />
+)
+
+export default SegmentExampleColor
--- a/docs/src/examples/components/Segment/Variations/SegmentExampleColor.shorthand.tsx
+++ b/docs/src/examples/components/Segment/Variations/SegmentExampleColor.shorthand.tsx
@@ -6,10 +6,7 @@ const SegmentExampleColor = () => (
   <ProviderConsumer
     render={({ siteVariables: { emphasisColors, naturalColors } }) =>
       _.keys({ ...emphasisColors, ...naturalColors }).map(color => (
-        <>
-          <Segment key={color} color={color} content={_.startCase(color)} inverted />
-          <br />
-        </>
+        <Segment key={color} color={color} content={_.startCase(color)} inverted />
       ))
     }
   />
--- a/docs/src/examples/components/Segment/Variations/SegmentExampleColor.shorthand.tsx
+++ b/docs/src/examples/components/Segment/Variations/SegmentExampleColor.shorthand.tsx
@@ -1,5 +1,5 @@
-import React from 'react'
-import _ from 'lodash'
+import * as React from 'react'
+import * as _ from 'lodash'
 import { Segment, ProviderConsumer } from '@stardust-ui/react'
 
 const SegmentExampleColor = () => (
--- a/docs/src/examples/components/Segment/Variations/SegmentExampleColor.tsx
+++ b/docs/src/examples/components/Segment/Variations/SegmentExampleColor.tsx
@@ -1,0 +1,20 @@
+import React from 'react'
+import _ from 'lodash'
+import { Segment, ProviderConsumer } from '@stardust-ui/react'
+
+const SegmentExampleColor = () => (
+  <ProviderConsumer
+    render={({ siteVariables: { emphasisColors, naturalColors } }) =>
+      _.keys({ ...emphasisColors, ...naturalColors }).map(color => (
+        <>
+          <Segment key={color} color={color} inverted>
+            {_.startCase(color)}
+          </Segment>
+          <br />
+        </>
+      ))
+    }
+  />
+)
+
+export default SegmentExampleColor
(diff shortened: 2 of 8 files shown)


docs/src/examples/components/Segment/Variations/SegmentExampleColor.shorthand.tsx
```diff
@@ -0,0 +1,18 @@
...
+import { Segment, ProviderConsumer } from '@stardust-ui/react'
+
+const SegmentExampleColor = () => (
+  <ProviderConsumer
+    render={({ siteVariables: { emphasisColors, naturalColors } }) =>
+      _.keys({ ...emphasisColors, ...naturalColors }).map(color => (
+        <>
+          <Segment key={color} color={color} content={_.startCase(color)} inverted />
+          <br />
+        </>
```
levithomason: For simplicity suggest turning:
```jsx
         <>
           <Segment key={color} color={color} content={_.startCase(color)} inverted />
           <br />
         </>
```
into:
```jsx
<Segment key={color} color={color} content={color} inverted />
```

bmdalex: simplified, thanks

layershifter (approved):
LGTM

src/themes/teams/components/Segment/segmentStyles.ts
```diff
@@ -1,10 +1,13 @@
+import * as _ from 'lodash'
+
 import { SegmentProps } from '../../../../components/Segment/Segment'
 import { ICSSInJSStyle, ComponentSlotStylesInput } from '../../../types'
 import { SegmentVariables } from './segmentVariables'
 
 const segmentStyles: ComponentSlotStylesInput<SegmentProps, SegmentVariables> = {
   root: ({ props: p, variables: v }): ICSSInJSStyle => {
-    const color = p.color || v.color
+    const color = _.get(v.colors, p.color, v.color)
```
kuzhelov: this logic appears in lots of places already - we should generalize it and consume from the single place

docs/src/examples/components/Segment/Variations/SegmentExampleColor.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+import { Segment, ProviderConsumer } from '@stardust-ui/react'
+
+const SegmentExampleColor = () => (
+  <ProviderConsumer
+    render={({ siteVariables: { emphasisColors, naturalColors } }) =>
```
kuzhelov, line 7: are `contextualColors` supported as well as a value for `color` prop? (not saying that those should be included in the example, but would like to clarify that)

kuzhelov: apparently, no - only emphasis and natural ones, according to the way `mapColorsToScheme` is implemented. Could you, please, suggest what is the reason behind that? As a client, I would expect that the following should be possible as well:

```jsx
<Segment color="danger">Do you really want to remove the repo?</Segment>
```

but this won't work.

src/themes/teams/components/Segment/segmentStyles.ts
```diff
@@ -14,7 +17,7 @@ const segmentStyles: ComponentSlotStylesInput<SegmentProps, SegmentVariables> =
         (p.inverted
```
miroslavstastny: `inverted` should work even without specifying color.

src/themes/teams/components/Segment/segmentVariables.ts
```diff
@@ -1,17 +1,22 @@
-import { ComponentVariablesPrepared } from '@stardust-ui/react'
+import { ColorValues } from '../../../types'
+import { mapColorsToScheme } from '../../../../lib'
 
 export interface SegmentVariables {
-  padding: string
-  background: string
-  borderRadius: string
+  colors: ColorValues<string>
```
miroslavstastny, line 5: can we come up with more descriptive name?

docs/src/examples/components/Segment/Variations/SegmentExampleColor.tsx
bmdalex: @kuzhelov https://github.com/stardust-ui/react/pull/597#discussion_r241343539 should clarify that partially, we still do not support `danger` prop; can we clarify this? @layershifter @levithomason

src/themes/teams/components/Segment/segmentStyles.ts
bmdalex: @miroslavstastny this is about the computed color (based on `color` prop, `colors` variable and `color` variable) not just the `color` prop; there's nothing to invert if there's no computed color for the segment

src/themes/teams/components/Segment/segmentVariables.ts
bmdalex: 👍 we had this discussion in other PRs but didn't find one, any suggestions?

src/themes/teams/components/Segment/segmentStyles.ts
bmdalex: how exactly would we do that? one liner function just for this? something like:
#### src/lib/colorUtils.ts:
```typescript
function getColmputedColor<T>(colors: ColorValues<T>, colorProp: string, defaults: T) {
  return _.get(colors, colorProp, defaults)
}
```

would this work @kuzhelov ?

kuzhelov: `getColmputedColor` -> `getComputedColor`?

yes, there is no much gain here, agree - lets leave as it is and see how it could be better generalized (if it should be). Thank you!

docs/src/examples/components/Segment/Variations/SegmentExampleColor.tsx
kuzhelov: lets leave it for a discussion, will make a note about that

kuzhelov (approved):

src/themes/teams/components/Segment/segmentStyles.ts
miroslavstastny: I understand, I was just pointing out the difference between Stardust and SUIR.
In SUIR `<Segment inverted>` is light on black. In Stardust `inverted` is ignored.

miroslavstastny (approved):

src/themes/teams/components/Segment/segmentStyles.ts
bmdalex: oh got it, good point, thanks @miroslavstastny , introduced a fix for that, should work now

PR #637, opened 2018-12-18
fix(FocusZone): Respect defaultTabbable element when FocusZone container gets focus

sophieH29:
```defaultTabbableElement``` should be taken into account, when ```shouldFocusInnerElementWhenReceivedFocus``` is ```true``` and focus goes to FocusZone container.

Here is a decision tree:
- Focus goes to FocusZone container (usually when it's an inner zone of other FocusZone) and ```shouldFocusInnerElementWhenReceivedFocus``` === "true"
- If ```defaultTabbableElement``` is provided by client, try to fetch element and focus it
- If no - focus first focusable element

### CODE REVIEW

**Review by jurokapsiar — APPROVED** [review]

**Review by miroslavstastny — APPROVED** [review]
- rename this test in `FocusZone-test.tsx`:
```js
it('should force focus to first focusable element when FocusZone container receives focus and shouldFocusFirstElementWhenReceivedFocus is set to "true"', () => {
```
- update CHANGELOG

PR #642, opened 2018-12-19
refactor(Label): cleanup docs and simplify render

dvdzkwsk:
This PR cleans up some inconsistencies in the doc strings, and simplifies the render method for readability. There are no functionality changes; this is simply a reduction of source code.


src/components/Label/Label.tsx
```diff
@@ -82,69 +82,45 @@ class Label extends UIComponent<Extendable<LabelProps>, any> {
...
-      }
+    if (childrenExist(children)) {
+      return (
+        <ElementType {...rest} className={classes.root}>
+          {children}
+        </ElementType>
+      )
     }
 
+    const imageElement = Image.create(image, {
```
dvdzkwsk, line 93: Factories already handle falsy values in the first position, so we don't need to handle that ourselves.

src/components/Label/Label.tsx
```diff
@@ -82,69 +82,45 @@ class Label extends UIComponent<Extendable<LabelProps>, any> {
...
       <ElementType {...rest} className={classes.root}>
-        {childrenExist(children) ? (
-          children
-        ) : (
-          <Layout main={content} start={start} end={end} gap={pxToRem(3)} />
-        )}
+        <Layout
+          start={
+            <>
+              {imagePosition === 'start' && imageElement}
```
dvdzkwsk: Remove reassignment of local variables. We can simply render elements inline in their target position.

src/components/Label/Label.tsx
```diff
@@ -25,27 +25,27 @@ export interface LabelProps
...
-  /** A Label can take the width of its container. */
+  /** A Label can take up the width of its container. */
   fluid?: boolean
 
-  /** Label can have an icon. */
+  /** A Label can have an icon. */
   icon?: ShorthandValue
 
-  /** An icon label can format an Icon to appear before or after the text in the label */
+  /** A Label can position its Icon at the start or end of the layout. */
```
dvdzkwsk, line 37: Make terminology consistent. Allowed prop values are "start" and "end", so we should use the same wording in the description.

kuzhelov (approved):
👍 👍

miroslavstastny (approved):

alinais (approved):

kuzhelov:
UTs need to pass before merging

kuzhelov:
UTs are fixed now - there was a problem with filtering logic implemented as part of shorthand tests

dvdzkwsk:
Thanks for the UT fix, @kuzhelov.

layershifter (changes requested):
I really like changes in this PR, superior work 👍 

----

But, there is an issue that causes visual regressions. `Layout` will add a gap if `start`/`end` are present, after these changes they will be always present and it will produce invalid gaps 😭 

We have already discussed this issue with @kuzhelov, we can make following changes:

```tsx
const startImage = imagePosition === 'start' && imageElement
const starIcon = iconPosition === 'start' && iconElement

const hasStart = startImage || startIcon

<Layout start={hasStart && (
  <>
    {startImage}
    {startIcon}
  </>
)}
```

It can be a solution, but I'm open for better ideas.

layershifter (approved):

Issue #95, opened 2018-08-15
Red file names in VS Code? 


bcalvery:
Whenever I open a file in VS Code I'm getting warnings/red text, which is really difficult to ignore. It seems to be caused by the first line `import React from 'react'`, which is showing the error `[ts] Module '"d:/Dev/stardust-react/node_modules/@types/react/index"' has no default export.`.
![image](https://user-images.githubusercontent.com/15043460/44165844-aa66fe00-a07e-11e8-92df-076385e824f4.png)
Changing the first line of the file to `import * as React from 'react'` does fix it. Is this an acceptable change to make in the files I'm working in?

davezuko:
Should be fixed by https://github.com/stardust-ui/react/pull/96.

levithomason:
Indeed.  The longer term fix here is to either (in order of preference):

1. Write examples in fully typed TypeScript but strip the type syntax in public facing examples.
1. Decide to show examples in TypeScript, which I don't think is good for the larger community, yet.
2. Write examples in JS, which is not great since we lose all type safety there.

levithomason:
Closing as #96 takes care of the original post.

PR #644, opened 2018-12-19
docs(Examples): use sources to render dedicated examples, separate contexts

### LINKED ISSUE #554: Docs example can be broken and merge into master without any check

layershifter:
Refs #95, fixes #554.

This PR:
- removes examples from `exampleContext` and will require only `index.tsx` files (`Types/index.tsx`, `Variations/index.tsx`)
- creates a new context for `*.source.json` files and uses it in `ExternalExampleLayout` with `SourceRender`


docs/src/components/ExternalExampleLayout.tsx
```diff
@@ -17,11 +20,26 @@ const ExternalExampleLayout: any = props => {
...
+      source={exampleSource.js}
+      renderHtml={false}
+      resolver={importResolver}
+    >
+      <SourceRender.Consumer>
+        {({ element, error }) => (
+          <>
+            {element}
+            {/* This block allows to see issues with examples as visual regressions. */}
+            {error && <div style={{ fontSize: '5rem', color: 'red' }}>{error.toString()}</div>}
```
layershifter, line 41: It's a little clumsy, but it will show us broken examples as visual regressions.

docs/src/utils/exampleContexts.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Get the Webpack Context for doc site example groups.
```
kuzhelov: nit: suggest to omit 'Get' verb as these exports just provide corresponding context modules, while 'get' action happens on the client side. 

Also, while 'groups' is also correct, would suggest to rename this context to `exampleIndexContext` - it will be easier for the reader to immediately infer the files implied by this context, without any additional necessity to remind terminology that is used by docs.

docs/src/utils/exampleKebabNameToSourceFilename.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Converts kebab-cased-example-name back into the original filename.
+ * @param {string} exampleKebabName
+ */
+const exampleKebabNameToSourceFilename = exampleKebabName => {
+  // button-example           => ButtonExample.source.json
+  // button-example-shorthand => ButtonExample.shorthand.source.json
+  return `${_.startCase(exampleKebabName)
+    .replace(/ /g, '')
```
kuzhelov, line 11: we can be a bit more expressive here. The only source of the whitespace could be either leading or trailing ones of the initial kebab-cased name. So, we might eliminate them before transform will happen, like that:

```js
const trimmedExampleKebabName = exampleKebabName.trim()

return `${_.startCase(trimmedExampleKebabName) ... }...`
```

While this makes code more efficient, I would put efficiency as the least worthy benefit in this case - the primary goal of this move is rather to improve explicitness of where the spaces (that regex aims to eliminate) might happen from.

docs/src/utils/index.tsx
```diff
@@ -1,7 +1,7 @@
 export { default as componentInfoContext } from './componentInfoContext'
 export { default as componentInfoShape } from './componentInfoShape'
-export { default as exampleContext } from './exampleContext'
-export { default as exampleKebabNameToFilename } from './exampleKebabNameToFilename'
+export { exampleGroupsContext, exampleSourcesContext } from './exampleContexts'
+export { default as exampleKebabNameToFilename } from './exampleKebabNameToSourceFilename'
```
kuzhelov: if this is not hard to do (which it seems not), I would suggest to rename the reexported entry so it will match the name of the file it is imported from

docs/src/components/ExternalExampleLayout.tsx
```diff
@@ -17,11 +20,26 @@ const ExternalExampleLayout: any = props => {
...
   if (!examplePath) return <PageNotFound />
-
-  const ExampleComponent = exampleContext(examplePath).default
-  if (!ExampleComponent) return <PageNotFound />
-
-  return <ExampleComponent />
+  const exampleSource: ExampleSource = exampleSourcesContext(examplePath)
+
+  return (
+    <SourceRender
```
kuzhelov, line 30: still not sure why we are relying on the React Context functionality to render source code - but this is definitely not the point of this PR, we could address this moment later (if necessary)

docs/src/components/ComponentDoc/ComponentExamples.tsx
```diff
@@ -30,23 +30,22 @@ export default class ComponentExamples extends React.Component<ComponentExamples
...
-    const indexPath = _.find(allPaths, path =>
+    const indexPath = _.find(exampleGroupsContext.keys(), path =>
       new RegExp(`\/${displayName}\/index\.tsx$`).test(path),
     )
     if (!indexPath) {
       return null
     }
 
-    const ExamplesElement = React.createElement(exampleContext(indexPath).default) as any
+    const ExamplesElement = React.createElement(exampleGroupsContext(indexPath).default) as any
```
kuzhelov: code which also suggests that `exampleIndexContext` would aid readability here. With current name used it was necessary for me to verify that all the index files are bundled by `exampleGroupsContext`

docs/src/components/ComponentDoc/ComponentExamples.tsx
```diff
@@ -82,23 +81,24 @@ export default class ComponentExamples extends React.Component<ComponentExamples
   )
 
   private testExamplesStructure(displayName: string, allPaths: string[]): string[] {
```
kuzhelov: given the return value of this method, it seems to be a good idea to rename it to something that is more descriptive, like `getMissingExamplePaths` - as otherwise it is completely unclear what this method is supposed to return

docs/src/components/ComponentDoc/ComponentExamples.tsx
```diff
@@ -82,23 +81,24 @@ export default class ComponentExamples extends React.Component<ComponentExamples
...
-    const examplesPattern = `\.\/\\w*\/${displayName}[\\w\/]*\/\\w+Example`
-    const allExamplesRegExp = new RegExp(`${examplesPattern}[\\w\.]*\.tsx$`)
+    const examplesPattern = `\.\/${displayName}[\\w\/]*\/\\w+Example`
```
kuzhelov: it seems that there is a problem with the regular expression used here - specifically, with the following part of it:

```
[\w\/]
```

Here it seems that the intent is to match either word character or slash, but the problem is that idle slash at start makes this regex to match `w` character literally.

This one will work correctly:

```
[\w\\]
```

The same problem apparently happens to the rest of the expression, where double slash is used here: `\\w+Example`. Am I missing something?

kuzhelov (commented):
have left several minor comments, but there is one important one that I would like to clarify: https://github.com/stardust-ui/react/pull/644/files#r244656294 . Other than that see these changes being good to merge 👍

docs/src/utils/exampleContexts.ts
layershifter: Agree, renamed to `exampleIndexContext` and removed the `get` verb 👍

docs/src/utils/exampleKebabNameToSourceFilename.ts
layershifter: Proposed solution is not correct because `_.startCase()` will create spaces:
```js
_.startCase('foo-bar')
// => 'Foo Bar'
```

docs/src/components/ComponentDoc/ComponentExamples.tsx
layershifter: I simplified it to:
```js
const examplesPattern = `\./${displayName}/[\\w/]+Example`
```

`examplesPattern` is not regex expression, it's a regular string. I'm not sure that we need there regular expressions at all.

docs/src/components/ExternalExampleLayout.tsx
layershifter: I'm looking for a better solution there.

kuzhelov (approved):

docs/src/components/ComponentDoc/ComponentExamples.tsx
kuzhelov: This is much simpler to analyze, thanks!

still there is a problem, unfortunately - and this comes from the fact that slashes are not escaped in the string expression. This is what string expression will be evaluated to: `./displayName/[w/]+Example` (with `displayName`='displayName')

And this expression creates several problems: 
- each of the path delimiter  slashes should be escaped
- dot at the start will match any character (is it our intent, or we would like to match dot literally? could we narrow down the set of characters to match, if so?)

I would see the following string should be used instead:

```js
// here I suppose that intent is to match dot character literally, 
// please, make any necessary corrections if it is not
const examplesPattern = `[.][/]${displayName}[/][\w/]+Example`
```

Please, just verify once again this regular expression before merging these changes. Thank you!

layershifter: Sorry, merged before this comment 😿 

> escaped in the string expression

They will be escaped in `new Regexp()`:
```js
const displayName = 'Label'
const r = new RegExp(`\./${displayName}/[\\w/]+Example`)

r.toString() // "/.\/Label\/[\w\/]+Example/"
r.test('./Label/Types/LabelExample') // true
```

PR #646, opened 2018-12-19
feat(FocusZone): Enable RTL

sophieH29:
This PR aims to enable RTL mode for FocusZone.
Also contains fixes for bidirectional mode, replicated from fabric-ui


src/lib/accessibility/FocusZone/FocusZone.tsx
```diff
@@ -775,9 +786,20 @@ export class FocusZone extends React.Component<FocusZoneProps> implements IFocus
...
+          let topBottomComparison
+
+          if (this.props.isRtl) {
+            // When in RTL, this comparison should be the same as the one in _moveFocusLeft for LTR.
+            // Going right at a rightmost rectangle will go up a line instead of down a line like in LTR.
+            // This is important, because we want to be comparing the bottom of the target rect
+            // with the top of the active rect.
+            topBottomComparison = targetRect.top.toFixed(3) > activeRect.bottom.toFixed(3)
+          } else {
+            topBottomComparison = targetRect.bottom.toFixed(3) < activeRect.top.toFixed(3)
```
miroslavstastny: All for `.top.`/`.bottom.` seem to be swapped.
For the comment to be true (moveRight in LTR === moveLeft in RTL) and to keep previous behavior for LTR, it should go as follows:
```jsx
if (this.props.isRtl) {
  topBottomComparison = targetRect.bottom.toFixed(3) > activeRect.top.toFixed(3)
} else {
  topBottomComparison = targetRect.top.toFixed(3) < activeRect.bottom.toFixed(3)
}
```

src/lib/accessibility/FocusZone/FocusZone.tsx
```diff
@@ -747,9 +747,20 @@ export class FocusZone extends React.Component<FocusZoneProps> implements IFocus
         this.props.isRtl,
         (activeRect: ClientRect, targetRect: ClientRect) => {
           let distance = -1
+          let topBottomComparison
```
miroslavstastny, line 750: This is not quite a descriptive name.
(don't ask me for proposal for better one)

src/lib/accessibility/FocusZone/FocusZone.tsx
sophieH29: yes, you're right. Thank you very much for that catching it! Changed it

miroslavstastny (approved):
CHANGELOG & go!

PR #647, opened 2018-12-19
feat(Label): add `color` prop

### LINKED ISSUE #636: [RFC] Color prop for components where it can target multiple areas (content, background)
# [RFC] Color prop for components where it can target multiple areas (content, background)

## Problem description
Let's take the example of `Label` component. The `color` prop for a label can have at least 2 targets:
- the color of the text
- the color of the label's background

How should we create props for this type of components?

## Solution 1
Always pick an area that color goes to and change other colors using variables.

**e.g.:**
for `Label` component, the `color` prop changes the background color, and we're using `color` variable to change the text color:

### API:
```jsx
<Label color='red' variables={{ color: '#fff' }} />
```
will create a label with white text on red background.

**Note**: We should rename `color` variable to `textColor` to avoid confusion.

### +:
- simple solution

### -:
- `color` variable receives a CSS color and does not go through our color palette mechanism (we'd need to implement that)
- confusing that we use a prop and a variable to set colors for areas of the same component

## Solution 2
`color` prop receives an object as value in order to identify the area of the component it sets color for:

**e.g.:**
for `Label` component, the `color` prop changes the text and background colors, so we'd have something like this for the `color` prop:

```typescript
type LabelColorProp = {
  text: string
  background: string
} | string
```

### API:
```jsx
<Label color={{ text: 'red', background: 'green' }} />
```
will create a label with red text on green background.

```jsx
<Label color='red' />
```
will create a label with red text on background with default color.

### +:
- more intuitive than Solution 1
- leverages the color palette mechanism to get correct hex codes for colors

### -:
- complex API
- complex solution

## Solution 3
- always pick an area that color goes to
- change other area colors automatically or using a color scheme: https://github.com/stardust-ui/react/pull/558#discussion_r239333426
- accept overrides using variables

**e.g.:**
for `Label` component:
- the `color` prop changes the background color and determines what text color complements the text color **(same as [SUIR](https://react.semantic-ui.com/elements/label/#variations-colored))** based on a tool / predefined color scheme (see: https://github.com/stardust-ui/react/pull/558#discussion_r239333426)
- we're using `color` variable to override the text color set by `color` prop automatically when setting background color

### API
```jsx
<Label color='red' />
```
will create a label with light colored text that fits best on a red background color.
```jsx
<Label color='white' />
```
will create a label with dark colored text that fits best on a white background.

```jsx
<Label color='white' variables={{ color: 'yellow' }} />
```
will create a label with yellow text on white background.

### +:
- leverages the color palette mechanism to get correct hex codes for colors
- simple API

### -:
- complex solution
- error prone (wrong colors in scheme) and hard to maintain
- can get confusing since variables would override props, we can consider other override mechanisms

bmdalex:
### Description

This PR:
- adds `color` prop to `Label` component
- implements basic color scheme for `Label` colors
- creates `Label` color examples
- closes #636 

### Latest updates on the API (@mnajdova)
Final implementation details. The color prop in the label can receive an object containing all parts of the color scheme: foreground, background, border, hover, shadow (we may want to trim this down to different objects in each component, or if we have component that can have multiple colors inside, we can always have the scheme). 
## How it works
1. The user specifies one color and all colors are calculated from the scheme
```
<Label color='red' content='Red' />
```
image

The color can be one of the following: `'primary' | 'secondary' | 'blue' | 'green' | 'grey'

2. The user can specify object as the value for the color, and those colors depending on the part where they are defined (foreground, background) are picked up from the color scheme
```
 <Label color={{foreground: 'red', background: 'green'}} content='Red on orange' />
```
image

3. The user can specify again object as color, and any of the color can be a custom color, or if is not found in the color scheme, will be returned as the actual HTML color.
```
<Label color={{foreground: 'red', background: 'blue'}} content='Red on blue' />
```
As the blue does not exists on the color scheme, the HTML blue color will be returned. The same will be applied if the color provided is in form of a hex value.
image

## How is the color scheme in the variables defined
The color scheme has defaultValues for all parts (foreground, background border etc), but the component defining can override this (for example the Label can define some shades of grey as defaults for the foreground and background). In order for this to work, there is an util method added called `getColorSchemeWithCustomDefaults`. Here is how the color scheme is defined inside the labelVariables:
```
 colorScheme: getColorSchemeWithCustomDefaults(
      {
        foreground: 'rgba(0, 0, 0, 0.6)',  // default label foreground
        background: 'rgb(232, 232, 232)', // default label background
      },
      siteVars.colorScheme,
    ),
```

Here is how all colors available in teams theme looks like:
image


src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -1,20 +1,38 @@
...
+  iconColor: string
+}
+
+export default (siteVars: any): LabelVariables => {
   const color = 'rgba(0, 0, 0, 0.6)'
 
   return {
-    circularRadius: pxToRem(9999),
-    padding: `0 ${pxToRem(4)} 0 ${pxToRem(4)}`,
+    colors: mapColorsToScheme(siteVars, colorVariants => ({
```
bmdalex: @layershifter fyi

kuzhelov:
speaking on the ask for inversion mechanism - we might start with a simple concept of `on` colors where 'on' color (for text/icons/anything that could be placed on top of background) will be provided in two variants: for light background (then `black` color may be used, as an example) and dark ones (for instance, then `white` color may be used for text). This will require the following declaration for `on` color from the client

```js
onColor = {
  light: #333, // will be used on dark background
  dark: #333 // will be used on light background
}
```

Then we might apply logic that will precompute appropriate 'on' color (dark or light one) for each color used in the user's theme. 

Note that if client will need to customize the 'on' colors generated, it might still be possible:

```js
// pseudo-implementation, just to provide an idea
onColor: {
    
  light: #333, // will be used on dark background by default
  dark: #333 // will be used on light background by default
  
  // OVERRIDES
  backgrounds: {
     primary:  <some custom color>, // will be applied for all 'primary' tints (100 - 900)

     // with these variants being used where appropriate for different tints of secondary
     secondary: {
         light: ... 
         dark: ...
     },
     
     // absolute fine-tuning
     pink: {
         ....
         '600': <color for exact tint of pink>
     }  
  }
}
```

bmdalex:
@kuzhelov can you please add your proposal to #636 ?
there's a big discussion on this happening over there

docs/src/examples/components/Label/Variations/LabelExampleColor.shorthand.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { Label, ProviderConsumer } from '@stardust-ui/react'
+
+const LabelExampleColor = () => (
+  <ProviderConsumer
+    render={({ siteVariables: { colorScheme } }) =>
+      _.keys(colorScheme).map(color => (
+        <>
+          <Label key={color} color={color} content={_.startCase(color)} />
+          <br />
+          <br />
```
layershifter: Is there any way to avoid `br`s? May be we can use `Grid` there?

bmdalex: introduced Grid, thx

mnajdova:
Reverting the changes introduced for the Menu, as it is much more complex, having all different variants. Will introduce different PR for the Menu component after merging this one.

mnajdova:
Hey guys, I need some input for overriding colors via variables. For the `Button` component for example, when we say primary in teams theme, it is generating for us the purple `Button` by default. Later, if we want to change the background, we can override the variable `primaryBackgroundColor` and change it's background. For the colors implemented currently for the Label however this does not work. For example if I specify the following:
```
<Label color='red' content='Label' variables={{ color: 'yellow', backgroundColor: 'red' }} />
```
I personally would expect to see label with red background and yellow color, but once the prop color is specified, the background and foreground colored generated from that property cannot be overridden. However, as the `color` prop is kind a different then the `primary` prop of the `Button`, I need your opinion whether this is the expected behavior, or we should remove the prop color in order to see the effect I am expecting... @layershifter @kuzhelov @Bugaa92 would appreciate your opinion on this, before I make any changes.

layershifter:
In current implementation I still see the issue via pollution of `LabelVariables`, we have `colors` with mapping there and we also have `color` and `borderColor` 🤔 

#### Proposal 1: move `color` and `borderColor` to mapping

Our color mapping will contain the `default` (`base`) color, ugly preudocode:

```diff
-export const colorScheme: ColorSchemeMapping = _.mapValues(
+export const colorScheme: ColorSchemeMapping = {..._.mapValues(
  emphasisAndNaturalColors,
  (colorVariants, colorName) => {
  // ...
  }),
+ default: {
+   foreground: 'magenta', // we will use something from color palette
+   border: 'blue',
    //  ...
+ }
,
```

It also means, that we use will `color` with default value:

```diff
-const colors = _.get(v.colorScheme, p.color, {
+const colors = _.get(v.colorScheme, p.color, v.colorScheme.default))
- foreground: v.color,
- background: v.backgroundColor,
-})
```

This change will break the following API and will point to **Proposal 2**:
```jsx
<Label content='Label' variables={{ color: 'yellow', backgroundColor: 'red' }} />
```

#### Proposal 2: allow to pass a scheme as value for `color` prop

If you we to pass a custom color for a Label's border just do it:

```jsx
<Label
  color={{ border: 'red', background: 'black' }}
  content='Label'
/>
```

All other variables will be picked from `default` color:

```js
const colors = _.isObject(p.color)
  ? { ...v.colorScheme.default, ...p.color }
  : _.get(v.colorScheme, p.color, v.colorScheme.default))
```

src/themes/teams/components/Label/labelStyles.ts
```diff
@@ -1,42 +1,61 @@
...
-  image: ({ variables }): ICSSInJSStyle => ({
-    height: variables.height,
-    width: variables.height,
-  }),
-  icon: ({ props }): ICSSInJSStyle => ({
-    ...(props.icon &&
-      typeof props.icon === 'object' &&
-      (props.icon as any).onClick && {
-        cursor: 'pointer',
+const getColorSchemeFromObject = (colorScheme, colors) => {
```
bmdalex: this will probably be extracted in `src/lib/colorUtils.ts`, need it for other components like button and menu

src/themes/teams/components/Label/labelStyles.ts
```diff
@@ -1,42 +1,61 @@
...
+    // if the color scheme contains the color, then get the value from it, otherwise return the color provided
+    const colorSchemeValue = _.get(colorScheme, colors[key], colorScheme.default[colors[key]])
+    result[key] = colorSchemeValue ? colorSchemeValue[key] : colors[key]
+  })
+  return result
+}
+
+const labelStyles: ComponentSlotStylesInput<LabelProps, LabelVariables> = {
+  root: ({ props: p, variables: v }): ICSSInJSStyle => {
+    const colors = _.isObject(p.color)
```
bmdalex: should this also be extracted in some color util (`src/lib/colorUtils.ts`)?

src/lib/commonPropTypes.ts
```diff
@@ -6,11 +6,39 @@ export interface CreateCommonConfig {
   children?: boolean | 'node' | 'element'
   as?: boolean
   className?: boolean
-  color?: boolean
+  color?: boolean | 'single' | 'complex'
```
bmdalex: ~~`single`~~ -> `simple` might be better in conjunction with `complex` what do you think?

mnajdova: Agreed, it should either be `single | multiple` or `simple | complex` (I like more the second combination)

src/themes/base/colors.ts
```diff
@@ -135,11 +143,39 @@ export const contextualColors: ContextualColors = {
...
+
+export const colors: ColorPalette = {
+  ...emphasisAndNaturalColors,
+  ...contextualColors,
 
   black: '#000',
   white: '#fff',
 }
+
+export const colorScheme: ColorSchemeMapping = {
```
bmdalex: what about:

```typescript
export const colorScheme: ColorSchemeMapping = _.mapValues(emphasisAndNaturalColors, (colorVariants, colorName) => {
  const foreground = isLightBackground(colorName) ? colors.black : colorVariants[50]

  return {
    foreground,
    border: foreground,
    shadow: foreground,
    hover: colorVariants[300],
    default: {
      foreground: colors.grey[600],
      border: colors.grey[600],
      shadow: colors.grey[600],
      background: colors.grey[100],
      hover: colors.grey[300],
    },
  }
})
```

src/themes/teams/components/Label/labelStyles.ts
bmdalex: also please add argument types and return type :)

src/themes/teams/utils/index.ts
```diff
@@ -1,7 +1,22 @@
...
 
 export const getSideArrow = (theme: ThemeInput) => {
   const { rtl, siteVariables } = theme
   const { arrowLeft, arrowRight } = siteVariables
   return rtl ? arrowLeft : arrowRight
 }
+
+export const getColorSchemeWithCustomDefaults = (
+  customDefaultValues: Partial<ColorScheme>,
+  colorScheme: ColorSchemeMapping,
```
bmdalex: nit: I'd make `colorScheme` the first argument

mnajdova: Will address this, and also move this to colorUtils

src/themes/teams/components/Label/labelStyles.ts
bmdalex: nit, you can use `_.mapValues` here:

```typescript
const getColorSchemeFromObject = (colorScheme, colors) => _.mapValues(colors, (color, colorName) => {
  // if the color scheme contains the color, then get the value from it, otherwise return the color provided
  const colorSchemeValue = _.get(colorScheme, color, colorScheme.default[color])
  return colorSchemeValue ? colorSchemeValue[colorName] : colors[colorName]
}
```

bmdalex (commented):
great work Marija 👍 please take a look at my comments before we merge

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `on` and `mouseLeaveDelay` props to `Popup` component @mnajdova ([#622](https://github.com/stardust-ui/react/pull/622))
 - Add Dropdown Single Selection variant @silviuavram ([#584](https://github.com/stardust-ui/react/pull/584))
 - Add `MenuDivider` component and `kind` prop to the `items` inside of the `Menu` for creating different components @mnajdova ([#682](https://github.com/stardust-ui/react/pull/682))
+- Add `color` prop to `Label` component @Bugaa92 ([#647](https://github.com/stardust-ui/react/pull/647))
```
layershifter: Should point to `Unreleased`

src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -1,20 +1,33 @@
...
+  endPaddingRight: string
+  height: string
+  iconColor: string
+}
+
+export default (siteVars: SiteVariablesPrepared): LabelVariables => {
   const color = 'rgba(0, 0, 0, 0.6)'
 
   return {
+    colorScheme: getColorSchemeWithCustomDefaults(siteVars.colorScheme, {
```
layershifter, line 20: I am not sure about this idea, can we omit it for initial implementation?

```js
colorScheme: {
  ...siteVars.colorScheme,
  default: {
    ...siteVars.colorScheme,
    foreground: color,
    background: 'rgb(232, 232, 232)',
  }
}
```

src/lib/colorUtils.ts
```diff
@@ -9,3 +17,43 @@ export const mapColorsToScheme = <T>(
...
+
+export const generateColorScheme = (
+  colorProp: ComplexColorPropType,
+  colorScheme: ColorValues<Partial<ColorScheme>>,
+): Partial<ColorScheme> => {
+  return _.isObject(colorProp)
+    ? { ...colorScheme.default, ...getColorSchemeFromObject(colorScheme, colorProp) }
+    : _.get(colorScheme, colorProp as string, {
+        ...colorScheme.default,
+      })
```
layershifter: 1. We don't need destruction there
```js
_.get(colorScheme, colorProp as string, colorScheme.default)
```
2. If we will revert the condition, we will be able to omit `as string`
```
type of colorProp === 'string'
    ? _.get(colorScheme, colorProp, colorScheme.default)
    : { ...colorScheme.default, ...getColorSchemeFromObject(colorScheme, colorProp) }
```

src/lib/colorUtils.ts
```diff
@@ -9,3 +17,43 @@ export const mapColorsToScheme = <T>(
...
+  const colors = _.get(colorScheme, colorProp)
+  return (area: keyof T, defaultColor: string) => (colors ? colors[area] : defaultColor)
+}
+
+export const getColorSchemeFromObject = (
+  colorScheme: ColorValues<Partial<ColorScheme>>,
+  colors: ComplexColorPropType,
+): Partial<ColorScheme> =>
+  _.mapValues(colors, (color, colorName) => {
+    // if the color scheme contains the color, then get the value from it, otherwise return the color provided
```
layershifter, line 31: This idea looks quite confusing to me :(

I have a definition for `blue` in the color scheme, but `blue` is a valid CSS color, https://www.quackit.com/css/css_color_codes.cfm

docs/src/examples/components/Label/Variations/LabelExampleColor.shorthand.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+import { Label, ProviderConsumer, Grid } from '@stardust-ui/react'
+
+const LabelExampleColor = () => (
+  <Grid
```
kuzhelov, line 6: just thoughts aloud: would rather prefer them being aligned horizontally, as it is commonly done by other libraries for tag components - but yes, there is no good layout component for that yet.

image

src/themes/teams/components/Label/labelVariables.ts
mnajdova: Not sure then how will the components define their default scheme, as in the label we have grey foreground and background, but for the button is white background and grey foreground...

src/lib/colorUtils.ts
mnajdova: Totally understand your point, but I don't have better proposal at this moment. We have to allow the user somehow to define the colors by name (the ones we have in the color palette, but still if we don't have them, we would end up with the CSS color.. If they define it as hex value then it is clear, but in the other case, is kind a confusing. Still, we have to use color names for the color prop, so I don't have any other proposal at this moment)

src/lib/commonPropInterfaces.ts
```diff
@@ -25,21 +25,33 @@ export interface UIComponentProps<P = any, V = any>
...
+  | 'yellow'
+  | string
+
+export type ComplexColorPropType =
+  | {
+      foreground?: ColorValue
+      background?: ColorValue
+      border?: ColorValue
+      shadow?: ColorValue
+      hover?: ColorValue
```
kuzhelov: as we've discussed, `hover` is a questionable member here, as it rather represents a state and not a color per se. At the same time, there is no need to introduce this complication for now - as there are several options of how this issue could be addressed

docs/src/examples/components/Label/Variations/LabelExampleColor.shorthand.tsx
mnajdova: Waiting for the Flex component then :)

src/lib/renderComponent.tsx
```diff
@@ -159,6 +159,7 @@ const renderComponent = <P extends {}>(config: RenderConfig<P>): React.ReactElem
 
         const {
           siteVariables = {
+            colorScheme: {},
```
kuzhelov, line 163: it would be great if we will be able to reduce `color` members `siteVariables` _directly_ expose - currently almost each new color-related concept results in the additional prop being added, but what is more important - those props are all cohesively connected with each other, so we might want eventually to introduce a dedicated object for them

CHANGELOG.md
mnajdova: Thanks, moved

src/themes/base/colors.ts
```diff
@@ -135,11 +143,40 @@ export const contextualColors: ContextualColors = {
...
 }
 
-export const colors: ColorPalette = {
-  ...contextualColors,
+const emphasisAndNaturalColors: EmphasisColors & NaturalColors = {
   ...emphasisColors,
   ...naturalColors,
+}
+
+const lightBackgroundColors = ['teal', 'yellow']
```
kuzhelov, line 151: see this being a very brittle thing that is tricky to maintain. I would be great to inline this `isLight` flag value to the color variant itself, so that in case if collection of colors will change (or even if color name will change), we won't bother about
- adding new color to the list / remove color from the list
- change the name of the color accordingly

We definitely should think on some mechanism that will introduce this ability to select foreground color depending on the background.

src/themes/teams/colors.ts
```diff
@@ -142,12 +150,41 @@ export const contextualColors: ContextualColors = {
...
 }
 
-export const colors: ColorPalette = {
-  ...contextualColors,
+const emphasisAndNaturalColors: EmphasisColors & NaturalColors = {
   ...emphasisColors,
   ...naturalColors,
+}
+
+const lightBackgroundColors = ['orange', 'yellow', 'lightGreen', 'postOrange']
```
kuzhelov, line 158: same story here - if we will be able to agree on some way to define the fact that specific color is a 'light' one, we would solve a maintainability issue and won't need to duplicate common related functionality (like `isLightBackground`)

src/lib/colorUtils.ts
layershifter:
```jsx
// uses `red` from schema
<Label color='red' />

// uses CSS `red`, i.e. #FF0000
<Label color={{ background: 'red' }} />

// still uses CSS colors, but you can get them from schema
<Label color={(colorSchema) => ({ background: colorSchema.red.background })} />
```

What about `color` as function?

kuzhelov (approved):

src/lib/colorUtils.ts
mnajdova: Couple of thoughts on this: 
- not sure what would be the use cases where the user will want to use the CSS colors, when they have defined schema for which colors will use
- more-over why should the prop in our API as **default** returns just some CSS color - as a user would make me wonder why we have special API for that

On the other hand, I understand the confusion, but would rather not over-complicate the API now, and wait for complains if there will be any in the future.

src/lib/commonPropInterfaces.ts
mnajdova: Agreed, will remove it for now.

src/lib/colorUtils.ts
```diff
@@ -9,3 +17,41 @@ export const mapColorsToScheme = <T>(
...
+    const colorSchemeValue = _.get(colorScheme, color, colorScheme.default[color])
+    return colorSchemeValue ? colorSchemeValue[colorName] : colors[colorName]
+  })
+
+export const getColorSchemeWithCustomDefaults = (
+  colorScheme: ColorSchemeMapping,
+  customDefaultValues: Partial<ColorScheme>,
+) => {
+  const mergedDefaultValues = {
```
kuzhelov, line 40: in future see this being reduced to a single concept of 'merging color schemes' - this should simplify the scheme object itself, make merging logic simpler and remove `default` prop

src/themes/teams/components/Label/labelStyles.ts
```diff
@@ -1,42 +1,46 @@
...
-    width: variables.height,
-  }),
-  icon: ({ props }): ICSSInJSStyle => ({
-    ...(props.icon &&
-      typeof props.icon === 'object' &&
-      (props.icon as any).onClick && {
-        cursor: 'pointer',
+const labelStyles: ComponentSlotStylesInput<LabelProps, LabelVariables> = {
+  root: ({ props: p, variables: v }): ICSSInJSStyle => {
+    const colors = generateColorScheme(p.color, v.colorScheme)
```
kuzhelov: is it, essentially, `mergeColorSchemes`?

src/lib/colorUtils.ts
bmdalex: totally support @mnajdova 's points

if the color scheme name collides with the CSS color name it's by design and it's expected since the developer should stick to a theme color as much as possible

if the developer reaaaally wants the actual CSS red (which is not recommended since they're supposed to work within the theme constraints), they can still use the hexcode (#FF0000)

src/themes/teams/components/Label/labelStyles.ts
kuzhelov: also, this definitely shouldn't be a user responsibility to bother about this stuff. I would expect that `colors` object will be just directly consumed from the style function, like that (one of the options):

```jsx
root: ({ props, variables, colors }) => ({
   foreground: colors.foreground,
   ...
})
```

src/lib/renderComponent.tsx
bmdalex: not sure I follow your point @kuzhelov can you please expand or offer an example?

src/themes/base/colors.ts
bmdalex: @kuzhelov agreed, what is your proposal?

add `background` as a key to the `ColorVariants` type?

src/themes/teams/colors.ts
bmdalex: see my question above:
https://github.com/stardust-ui/react/pull/647/files#r247172459

src/themes/teams/components/Label/labelStyles.ts
bmdalex: @kuzhelov 
don't forget about the `p.color` parameter coming from the user; `generateColorScheme` takes that argument and returns the `ColorScheme` object for `p.color`; maybe the name is confusing and should be:

`generateColorSchemeForColorProp` or something more concise

mnajdova: I've added this with the latest commit (those are merged in renderComponent). Please take a look.

src/themes/base/colors.ts
kuzhelov: yes, essentially I would propose to reconsider the set of props of the `ColorVariants` type - and just add the only necessary to inline this information currently kept by `lightBackgroundColors` array. It may be a `background` prop, or maybe other set of props (like `light` and `dark`, and this will allow us to properly apply colors on top of each other). 

There are several corner cases that we won't be able to cover with the naive approach, thus I would rather like to open a separate discussion around exact implementation - but generally, yes, would rather see this information as being part of the `ColorVariants` type.

src/lib/renderComponent.tsx
mnajdova: @kuzhelov are you suggesting having maybe one `colors` object that will contain the scheme, colors  and everything related to the coloring mechanism? I would really avoid that refactoring as part of this part, but I can create an issue and open PR after this one gets in.

kuzhelov: @Bugaa92, sure - currently we have the following set of props being exposed by `siteVariables`: `colors`, `contextualColors`, `emphasisColors`, `colorScheme`, `naturalColors`, etc - there are just plenty of them, and all are aimed to describe `color` concept. Here I am proposing to introduce single `color` prop that will be an object with all these being assembled as props:

```js
siteVariables = {
   color: { 
      scheme, 
      contextual,
      natural,
       ...
   }
   ...
}
```

This will aid maintainability (as all closely related aspects will be semantically grouped), as well as will not pollute the `siteVariables` object itself. More than that, on the client perspective it will be much cleaner to see the entire set of objects from which colors could be consumed from, as everything related to `color` will be assembled under single object.

bmdalex: @kuzhelov I like it, but would agree with @mnajdova to postpone this for a follow-up PR and create an issue for it

src/themes/base/colors.ts
bmdalex: @kuzhelov got it, agreed it'd be a better way to do things and was thinking of something similar myself
@layershifter can you weigh in as well?

mnajdova: Will keep a track of this in an issue together will the other proposals for improvements regarding the color mechanisms.

docs/src/examples/components/Label/Variations/LabelExampleColor.shorthand.tsx
levithomason: Always delete as much code as possible:
1. No Grid is necessary, especially since it is a very complex solution to a very simple problem (add a space).
1. Lodash is not necessary to Start Case the color names, just show the prop values (more helpful as well).
1. Only use top level parent components, not the directly exported children (Provider.Consumer vs ProviderConsumer)

image

```diff
import * as React from 'react'
-import * as _ from 'lodash'
-import { Label, ProviderConsumer, Grid } from '@stardust-ui/react'
+import { Label, Provider } from '@stardust-ui/react'

const LabelExampleColor = () => (
-  <Grid
-    columns="auto"
-    styles={{ justifyContent: 'left', justifyItems: 'left' }}
-    variables={{ gridGap: '10px' }}
-  >
    <ProviderConsumer
      render={({ siteVariables: { colorScheme } }) =>
-        _.keys(colorScheme).map(color => (
-          <Label key={color} color={color} content={_.startCase(color)} />
+       Object.keys(colorScheme).map(color => (
+          <span key={color}>
+            <Label color={color} content={color} />{' '}
+          </span>
        ))
      }
    />
-  </Grid>
)
```

src/components/Label/Label.tsx
```diff
@@ -55,7 +58,7 @@ class Label extends UIComponent<ReactProps<LabelProps>, any> {
   static className = 'ui-label'
 
   static propTypes = {
-    ...commonPropTypes.createCommon(),
+    ...commonPropTypes.createCommon({ color: 'complex' }),
```
levithomason, line 61: The `commonPropTypes` abstraction is growing out of hand.  We already have `customPropTypes`.  Also, the function calls are unnecessary.

Always try to delete as much code and work as possible.  Consider this line:

```jsx
...commonPropTypes.createCommon({ color: 'complex' }),
```

1. `commonPropTypes` duplicates the job of `customPropTypes`
1. Requires a function call
1. Requires constructing an options object (of the exact shape that the returned object will contain)
1. Causes indirection, obscures implementation, and requires spreading

We can achieve this much more simply like this:
```jsx
color: customPropTypes.complexColor
```

1. No need to create or import `commonPropTypes` vs `customPropTypes`
1. No need to call a function (`createCommon()`)
1. No need to create an options object
1. Is obvious as the entire implementation is explicit and simple to read

src/lib/colorUtils.ts
levithomason: Clarification, the color prop exists specifically to restrict the developer from deviating from the theme color.  The color prop is by design restrictive and imprecise.  It should not allow defining where or how the color is applied nor what specific color it is, just the name.

Where and how the color is applied is the job of the designer through the theme.  All props should be high-level, natural language, imprecise, and void of implementation.

Think of the `color` prop as used in this statement "Click on the red label."  Now, imagine you ask 10 designers to draw the label.  There are many ways to do this, which is exactly what we want.  We want flexibility through imprecise natural language.

PR #650, opened 2018-12-19
docs(Examples): refactor SourceManager, add toggle for languages

### LINKED ISSUE #495: Docs: use TypeScript and plain JS in examples

### LINKED ISSUE #615: Docs: crash on click on "Shorthand API"
### Steps

1. Open https://stardust-ui.github.io/react/components/ref
2. Click "Try it" on the first example
3. Then click on "Shorthand API"

image

### Expected Result
"Shorthand API" is disabled

### Actual Result
Crash 💣

layershifter:
Fixes #495.
Fixes #615.

---

This PR:
- refactors `SourceManager` to a React component to provide declarative controls
- adds toggle for TypeScript examples
- fixes issues with Children and Shorthand APIs switch

image


docs/src/components/ComponentDoc/ComponentExamples.tsx
```diff
@@ -30,23 +30,22 @@ export default class ComponentExamples extends React.Component<ComponentExamples
    */
   private renderExamples = (): JSX.Element | null => {
     const { displayName } = this.props
-    const allPaths = exampleContext.keys()
 
     // rule #1
-    const indexPath = _.find(allPaths, path =>
+    const indexPath = _.find(exampleGroupsContext.keys(), path =>
```
kuzhelov: while `Groups` is correct, maybe call it `exampleIndexContext`, it will be easier to infer which files it corresponds to. For example, it has take me to check sources of this `exampleGroupsContext` to ensure that `Groups` implies index files only (the thing that is expected by `indexPath ` variable)

layershifter: Was done in #644.

docs/src/components/ComponentDoc/ComponentSourceManager/ComponentSourceManager.ts
```diff
@@ -0,0 +1,138 @@
+import * as _ from 'lodash'
+import * as React from 'react'
+
+import { ExampleSource } from 'docs/src/types'
+import formatCode from 'docs/src/utils/formatCode'
+import { componentAPIs as APIdefinitions, ComponentAPIs } from './componentAPIs'
+import getExampleSource from './getExampeSource'
+
+export type ComponentSourceManagerRenderProps = ComponentSourceManagerState & {
+  handleCodeAPIChange: (newApi: keyof ComponentAPIs) => void
```
mnajdova, line 10: The Shorthand API should be the first selected option. This is the one we want to advertise. With the latest changes the Children API is first.

mnajdova: Also, would recommend the Shorthand API to be the first option, as it was before.

layershifter: Agree, done with the latest commit :)

docs/src/components/ComponentDoc/ComponentSourceManager/ComponentSourceManager.ts
```diff
@@ -0,0 +1,138 @@
...
+    const { examplePath } = props
+    const { componentAPIs, currentCodeAPI, currentCodeLanguage, currentCode: storedCode } = state
+
+    const sourceCodes = componentAPIs[currentCodeAPI].sourceCode
+    const originalCode = sourceCodes[currentCodeLanguage]
+
+    const currentCode = storedCode || originalCode
+    const currentCodePath = examplePath + componentAPIs[currentCodeAPI].fileSuffix
+
+    const codeParser = currentCodeLanguage === 'ts' ? 'typescript' : 'babylon'
```
mnajdova: It seems that this doesn't raise any error in the Typescript mode:
```
const notString: string = 3
```

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -313,121 +285,100 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
...
-    this.setState({ sourceCode: this.sourceCodeMgr.currentCode })
-  }
+  handleCodeLanguageChange = language => () => {
+    const { handleCodeLanguageChange, wasCodeChanged } = this.props
 
-  setApiCodeType = (codeType: SourceCodeType) => {
-    this.sourceCodeMgr.codeType = codeType
-    this.updateAndRenderSourceCode()
+    if (wasCodeChanged) {
+      if (confirm('Lose your changes?')) {
```
mnajdova, line 296: It was awkward to me to lose the changes between switching to Javascript and Typescript. Can't we store both source codes in state? Not sure if this will really be used by the users, but just wondering...

docs/src/components/ComponentDoc/ComponentSourceManager/ComponentSourceManager.ts
layershifter: Yep, we don't have TS checks there. And I'm not sure that we actually need them for now.

I renamed `codeParser` to `prettierParser` to avoid confusion

mnajdova:
One more thought. Now that we are autogenerating the js code, maybe we should disable the `edit on github` option for those examples? It seems kind a strange that edit is navigating you to different file..
image

docs/src/components/ComponentDoc/ComponentSourceManager/ComponentSourceManager.ts
mnajdova: Got it now, I thought it was actually a code parser. I know that the Monaco editor was configurable to parse and show ts errors, so I thought there is something similar here.

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
layershifter: I want to avoid implementation complications there.

If we will get claims about it, we can refactor `ComponentSourceManager` somehow to store both source codes.

mnajdova: Valid point..

layershifter:
> One more thought. Now that we are autogenerating the js code, maybe we should disable the `edit on github` option for those examples? It seems kind a strange that edit is navigating you to different file..
> image

Agree, I pushed changes with this 👍

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -313,148 +285,148 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
...
+            content="Edit"
+            href={ghEditHref}
+            target="_blank"
+          />
+        )}
       </Menu>
     )
   }
 
   renderJSX = () => {
```
mnajdova: For consistency, let's rename this one too to `renderSourceCode`

mnajdova (approved):

PR #651, opened 2018-12-19
feat(icon): color prop

bmdalex:
### Description

This PR:
- adds `color` prop to `Icon` component
- creates `Icon` color examples

### API

```jsx
<Icon color={COLOR} .../>
```

screenshot 2018-12-19 at 19 38 04

### DIFF
--- a/docs/src/examples/components/Icon/Variations/IconExampleColor.shorthand.tsx
+++ b/docs/src/examples/components/Icon/Variations/IconExampleColor.shorthand.tsx
@@ -16,27 +16,17 @@ const IconExampleColor = () => (
       <Icon name="call" bordered variables={{ outline: true }} />
       <Icon name="call-video" bordered variables={{ outline: true }} />
     </div>
-    <Text
-      content={
-        <span>
-          USING THE <code>color</code> VARIABLE:
-        </span>
-      }
-      weight="bold"
-    />
+    <Text weight="bold">
+      USING THE <code>color</code> VARIABLE:
+    </Text>
     <div>
       <Icon name="calendar" bordered variables={{ color: 'violet' }} />
       <Icon name="call" bordered variables={{ color: 'yellowgreen' }} />
       <Icon name="call-video" bordered variables={{ color: 'cornflowerblue' }} />
     </div>
-    <Text
-      content={
-        <span>
-          USING THE <code>borderColor</code> VARIABLE:
-        </span>
-      }
-      weight="bold"
-    />
+    <Text weight="bold">
+      USING THE <code>borderColor</code> VARIABLE:
+    </Text>
     <div>
       <Icon
         name="calendar"
@@ -54,14 +44,9 @@ const IconExampleColor = () => (
         variables={{ color: 'cornflowerblue', borderColor: 'orangered' }}
       />
     </div>
-    <Text
-      content={
-        <span>
-          USING THE <code>color</code> PROP:
-        </span>
-      }
-      weight="bold"
-    />
+    <Text weight="bold">
+      USING THE <code>color</code> PROP:
+    </Text>
     <div>
       <ProviderConsumer
         render={({ siteVariables: { emphasisColors, naturalColors } }) =>
--- a/docs/src/examples/components/Icon/Variations/IconExampleColor.shorthand.tsx
+++ b/docs/src/examples/components/Icon/Variations/IconExampleColor.shorthand.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react'
-import { Icon, Grid, Text } from '@stardust-ui/react'
+import * as _ from 'lodash'
+import { Icon, Grid, Text, ProviderConsumer } from '@stardust-ui/react'
 
 const IconExampleColor = () => (
   <Grid columns="repeat(4, auto)" styles={{ alignItems: 'center' }} variables={{ gridGap: '10px' }}>
@@ -51,6 +52,23 @@ const IconExampleColor = () => (
         name="call-video"
         bordered
         variables={{ color: 'cornflowerblue', borderColor: 'orangered' }}
+      />
+    </div>
+    <Text
+      content={
+        <span>
+          USING THE <code>color</code> PROP:
+        </span>
+      }
+      weight="bold"
+    />
+    <div>
+      <ProviderConsumer
+        render={({ siteVariables: { emphasisColors, naturalColors } }) =>
+          _.take(_.keys({ ...emphasisColors, ...naturalColors }), 3).map(color => (
+            <Icon key={color} name="call" bordered color={color} />
+          ))
+        }
       />
     </div>
   </Grid>
--- a/src/components/Icon/Icon.tsx
+++ b/src/components/Icon/Icon.tsx
@@ -6,6 +6,7 @@ import {
   createShorthandFactory,
   UIComponentProps,
   commonPropTypes,
+  ColorComponentProps,
 } from '../../lib'
 import { iconBehavior } from '../../lib/accessibility/'
 import { Accessibility } from '../../lib/accessibility/types'
@@ -17,7 +18,7 @@ export type IconXSpacing = 'none' | 'before' | 'after' | 'both'
 
 export type IconSize = 'smallest' | 'smaller' | 'small' | 'medium' | 'large' | 'larger' | 'largest'
 
-export interface IconProps extends UIComponentProps {
+export interface IconProps extends UIComponentProps, ColorComponentProps {
   /**
    * Accessibility behavior if overriden by the user.
    * @default iconBehavior
@@ -57,6 +58,7 @@ class Icon extends UIComponent<ReactProps<IconProps>, any> {
     ...commonPropTypes.createCommon({
       children: false,
       content: false,
+      color: true,
     }),
     accessibility: PropTypes.func,
     bordered: PropTypes.bool,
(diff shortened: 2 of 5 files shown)


docs/src/examples/components/Icon/Variations/IconExampleColor.shorthand.tsx
```diff
@@ -53,6 +54,23 @@ const IconExampleColor = () => (
...
       />
     </div>
+    <Text
+      content={
+        <span>
+          USING THE <code>color</code> PROP:
+        </span>
+      }
+      weight="bold"
+    />
```
layershifter:
```jsx
<Text weight="bold">
  USING THE <code>color</code> PROP:
</Text>
```

Will be more compact

src/themes/teams/components/Icon/iconVariables.ts
```diff
@@ -1,31 +1,39 @@
...
   secondaryColor: siteVars.white,
+  redColor: siteVars.red,
   disabledColor: siteVars.gray06,
 
-  redColor: siteVars.red,
-  brandColor: siteVars.brandColor,
+  outline: undefined,
+  horizontalSpace: pxToRem(10),
+  margin: `0 ${pxToRem(8)} 0 0`,
+  sizeModifier: undefined,
```
layershifter: Is there are real need to define `undefined` explicitly?

src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -1,13 +1,16 @@
...
 import { ComponentSlotStylesInput, ICSSInJSStyle, FontIconSpec } from '../../../types'
 import { ResultOf } from '../../../../../types/utils'
-import { IconXSpacing, IconProps } from '../../../../components/Icon/Icon'
+import { IconXSpacing, IconProps, IconSize } from '../../../../components/Icon/Icon'
 import { pxToRem } from '../../utils'
 import { getStyle as getSvgStyle } from './svg'
+import { IconVariables, IconSizeModifier } from './iconVariables'
 
-const sizes = new Map([
+const sizes = new Map<IconSize, number>([
```
layershifter: Can we remove this `Map` at all and use a typed object there?

layershifter (approved):
LGTM, left some comments 👍

src/themes/teams/components/Icon/iconVariables.ts
bmdalex: not really, made it optional and removed from values

src/themes/teams/components/Icon/iconStyles.ts
bmdalex: done, thx

docs/src/examples/components/Icon/Variations/IconExampleColor.shorthand.tsx
bmdalex: agreed, changed in the whole example, thx

docs/src/examples/components/Icon/Variations/IconExampleColor.shorthand.tsx
```diff
@@ -53,6 +44,18 @@ const IconExampleColor = () => (
         variables={{ color: 'cornflowerblue', borderColor: 'orangered' }}
       />
     </div>
+    <Text weight="bold">
+      USING THE <code>color</code> PROP:
+    </Text>
+    <div>
+      <ProviderConsumer
+        render={({ siteVariables: { emphasisColors, naturalColors } }) =>
+          _.take(_.keys({ ...emphasisColors, ...naturalColors }), 3).map(color => (
```
kuzhelov, line 53: this expression is very confusing - there should be a way to simplify this example. Actually, why slicing is necessary here?

src/components/Icon/Icon.tsx
```diff
@@ -25,7 +26,7 @@ export type IconSize =
   | 'huge'
   | 'massive'
 
-export interface IconProps extends UIComponentProps {
+export interface IconProps extends UIComponentProps, ColorComponentProps {
```
kuzhelov, line 21: why do we need to introduce this `ColorComponentProps ` type here? It should be enough just to provide `color: true` to `...commonPropTypes.createCommon({`

src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -1,23 +1,26 @@
...
-  ['small', 14],
-  ['normal', 16],
-  ['large', 20],
-  ['big', 32],
-  ['huge', 40],
-  ['massive', 64],
-])
+import { IconVariables, IconSizeModifier } from './iconVariables'
+
+const sizes: { [key in IconSize]: number } = {
```
kuzhelov, line 13: 👍

src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -86,11 +90,12 @@ const getIconSize = (size, sizeModifier): number => {
   return modifiedSizes[size] && modifiedSizes[size][sizeModifier]
 }
 
-const getIconColor = color => color || 'currentColor'
+const getIconColor = (colorProp: string, variables: IconVariables) =>
+  _.get(variables.colors, colorProp, variables.color || 'currentColor')
```
kuzhelov, line 92: note, this changes semantic of the method. Previously it was a feature that icon will take a color of its surrounding - with `color` variable being completely ignored. Please, ensure the following things before merging:
- corresponding changelog entry is provided in the BREAKING CHANGES section
- there are no visual regressions related to that

src/themes/teams/components/Icon/iconVariables.ts
```diff
@@ -1,31 +1,38 @@
+import { ColorValues } from '../../../types'
+import { mapColorsToScheme } from '../../../../lib'
 import { pxToRem } from '../../utils'
 
-export interface IconVariables {
-  [key: string]: string | number | boolean | undefined
```
kuzhelov, line 4: why this extensibility point was removed? Note that this extensibility is necessary for some scenarios, like, for example, to provide some category to the component that will define its styles:

```jsx
<Icon variables={ chatIcon: true } ... />
```

With these changes types will be broken.

kuzhelov (changes requested):
lets please take another look on removed extensibility point for IconVariables interface

docs/src/examples/components/Icon/Variations/IconExampleColor.shorthand.tsx
bmdalex: screenshot 2018-12-19 at 19 38 04
just to make the example fit nicely, any suggestions for improving, @kuzhelov ?

src/components/Icon/Icon.tsx
bmdalex: @kuzhelov one is the interface, the other one is the propTypes, makes sense?

src/themes/teams/components/Icon/iconStyles.ts
bmdalex: > Previously it was a feature that icon will take a color of its surrounding - with color variable being completely ignored

@kuzhelov I am aware, but are you sure? from what I can see from previous implementation, I am not changing the logic:
```typescript
const getIconColor = color => color || 'currentColor'
```

where `color` param is `variables.color`.

It looks like priority was:
1. `variables.color`
2. `currentColor`

and after my change the priority is:
1. `color` prop
2. `variables.color`
3. `currentColor`

src/themes/teams/components/Icon/iconVariables.ts
bmdalex: the reasons were:
### 1. it doesn't bring any value
Your example would work without any type errors because the type for `variables` prop is not generic enough anyway to take into account `IconVariables` interface.

`IconVariables` is only relevant for 
src/themes/teams/components/Icon/iconVariables.ts

### 2. It was causing the following type error:
screenshot 2018-12-20 at 16 58 32

If you feel strongly that we should add the extensibility back (which I doubt), the error can be fixed by adding `object` to the values type:
screenshot 2018-12-20 at 16 59 28

@kuzhelov what do you think?

src/themes/teams/components/Icon/iconStyles.ts
bmdalex: this is the only change in screener, so no visual regressions:
screenshot 2018-12-20 at 17 08 10

Bugaa92:
@kuzhelov can you please add your proposal to #636 ?
there's a big discussion on this happening over there

src/components/Icon/Icon.tsx
kuzhelov: sorry - yes, absolutely, thanks!

src/themes/teams/components/Icon/iconVariables.ts
kuzhelov: my thoughts have the following motivation
> in case if we will remove extensibility point from variables type, we will break variables' semantics by this move, as variables interface won't be extendable anymore 

If we would like to make this move (note, this is not related to this PR), we should make it for all variables' types altogether, to enforce consistency in our code base. However, given our current vision about the fact that variables are seen to be extendable ones, we shouldn't introduce any changes at least for now

bmdalex: that's fine with me if it moves the PR closer to a merge, reverted the change as advised

docs/src/examples/components/Icon/Variations/IconExampleColor.shorthand.tsx
bmdalex: @kuzhelov can I resolve this?

kuzhelov (approved):

PR #656, opened 2018-12-20
feat(RTL): Enable RTL for keyboard handlers

sophieH29:
This PR aims to enable RTL for key handlers, so then when users define `keyHandlers` in accessibility behaviors, in RTL mode, the keys should be swapped:
`Left` -> `Right`
`Right` -> `Left`

Good example to check how it works is a Menu with submenus.

### DIFF
--- a/src/lib/getKeyDownHandlers.ts
+++ b/src/lib/getKeyDownHandlers.ts
@@ -8,19 +8,19 @@ import {
 } from 'src/lib/accessibility/types'
 import { State, PropsWithVarsAndStyles } from '../themes/types'
 
+const rtlKeyMap = {
+  [keyboardKey.ArrowRight]: keyboardKey.ArrowLeft,
+  [keyboardKey.ArrowLeft]: keyboardKey.ArrowRight,
+}
+
 /**
  * Assigns onKeyDown handler to the Component's part element, based on Component's actions
  * and keys mappings defined in Accessibility behavior
  * @param {AccessibilityActionHandlers} componentActionHandlers Actions handlers defined in a component.
  * @param {KeyActions} behaviorKeyActions Mappings of actions and keys defined in Accessibility behavior.
  * @param {State & PropsWithVarsAndStyles} props The props which are used to invoke onKeyDown handler passed from top.
+ * @param {boolean} isRtl Indicates if Left and Right arrow keys should be swapped in RTL mode.
  */
-
-const rtlKeyMap = {
-  [keyboardKey.ArrowRight]: keyboardKey.ArrowLeft,
-  [keyboardKey.ArrowLeft]: keyboardKey.ArrowRight,
-}
-
 const getKeyDownHandlers = (
   componentActionHandlers: AccessibilityActionHandlers,
   behaviorKeyActions: KeyActions,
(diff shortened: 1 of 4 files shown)


src/lib/getKeyDownHandlers.ts
```diff
@@ -7,17 +8,24 @@ import {
...
+  [keyboardKey.ArrowLeft]: keyboardKey.ArrowRight,
+}
+
 /**
  * Assigns onKeyDown handler to the Component's part element, based on Component's actions
  * and keys mappings defined in Accessibility behavior
  * @param {AccessibilityActionHandlers} componentActionHandlers Actions handlers defined in a component.
  * @param {KeyActions} behaviorKeyActions Mappings of actions and keys defined in Accessibility behavior.
  * @param {State & PropsWithVarsAndStyles} props The props which are used to invoke onKeyDown handler passed from top.
+ * @param {boolean} isRtl Indicates if Left and Right arrow keys should be swapped in RTL mode.
```
mnajdova: Would recommend using the `rtl` name, the same name that is used in the Theme, or `isRtlEnabled` if you want to use the verb :)

src/lib/getKeyDownHandlers.ts
```diff
@@ -34,9 +42,20 @@ const getKeyDownHandlers = (
...
       onKeyDown: (event: KeyboardEvent) => {
         handledActions.forEach(actionName => {
+          let keyCombinations = componentPartKeyAction[actionName].keyCombinations
+
+          if (isRtl) {
+            keyCombinations = keyCombinations.map(keyCombination => {
+              if (rtlKeyMap[keyCombination.keyCode]) {
+                keyCombination.keyCode = rtlKeyMap[keyCombination.keyCode]
+              }
+              return keyCombination
```
mnajdova, line 53: create variable for `rtlKeyMap[keyCombination.keyCode]` and reuse it when assigning it to the keyCode

src/lib/getKeyDownHandlers.ts
```diff
@@ -34,9 +42,20 @@ const getKeyDownHandlers = (
...
+            })
+          }
+
```
layershifter, line 56: We can move this code to a separate function and test it separately:
```js
const keyCombinations = normalizeKeyCombinations(componentPartKeyAction[actionName], isRtl)
```

sophieH29: We can :) But I think this part of the code is very trivial to move it to separate function and test it because it's mainly key mapping.
I tested that correct action handlers are invoked based on key pressed, and this seems to be an important part.
Also, creating one more function leads to having one more `rtl` prop dependency, hence `getKeyDownHandlers` will do nothing about this param but just passing it further.

src/lib/getKeyDownHandlers.ts
sophieH29: would stick for `isRtlEnabled`, thx

miroslavstastny (commented):
Everywhere in Stardust (except styles), we use **start** and **end** to be direction-agnostic, but here we keep using **left** and **right** and remap it for RTL.
Have you thought about using meta key names (**ArrowStart**, **ArrowEnd**) and remap them for both directions accordingly?
Would you consider that more clean or more confusing? @jurokapsiar 
Other than that the PR looks good to me.

sophieH29:
For me it feels more confusing, I would say `start` & `end` is more about a position, not direction. For this would better make sense to have `next` and `previous`. But I don't see a reason to change it now, as it is more an internal logic which is not shown to the clients, all they do is just binding a key to action.

jurokapsiar:
Yes, we considered this. I agree with @sophieH29. Changing the name for ArrowLeft to ArrowPrev is technically more correct, but very unintuitive especially for consumers that are not aware of RTL. In the case of direction in the components, where you control the interface, I agree that we should have start/end/next/prev. On the other hand, people expect to be able to set keyboard handlers for ArrowLeft or ArrowRight.

miroslavstastny (approved):

Issue #610, opened 2018-12-13
AutoControlled component does not handle NOT autocontrolled state


sophieH29:
### Steps
Let's ```componentA``` is derived from ```AutoControlled``` component

```componentA``` has state object
```js
state = {
    someState: true
}
```

And has autocontrolled state and props

```js

props = {
...
autoControlledState: bool
defaultAutoControlledState: bool
...
}
...
static autoControlledProps = ['autoControlledState']
...
```

Steps:
1. Use component 
```js
<componentA defaultAutoControlledState={true} />
```
2. Debug component
3. Check applied state to component

### Expected Result
Expected state is 
```js
state = {
    someState: true,
    autoControlledState: true
}
```

### Actual Result
```js
state = {
    someState: true,
}
```

As a result, ```state``` of the component is not taken into account in ```AutoControlled``` component. Then constructor of ```componentA``` overrides the state value with what was set in ```AutoControlled``` component during initialization.

layershifter:
```js
class SomeComponent extends AutoControlledComponent {
  state = { someState: true }
}
// equals to:
class SomeComponent extends AutoControlledComponent {
  constructor(...args) {
    super(...args)
    console.log(this.state) // { autoControlledState: true }
    this.state = { someState: true }
  }
}

```

The `autoControlledState` variable is assigned inside `constructor()` of `AutoControlledComponent`, [AutoControlledComponent.tsx#L149](https://github.com/stardust-ui/react/blob/v0.15.0/src/lib/AutoControlledComponent.tsx#L149).

The `SomeComponent`'s constructor will be called after it and this means that it will simply reassign the `state` property 👎  More details on [StackOverflow](https://stackoverflow.com/a/43595944/6488546).

More historic details are there https://github.com/Semantic-Org/Semantic-UI-React/pull/1799. The solution is to disallow to use `state` property in components that extends `AutoControlledComponent`, but I don't see any transparent way to do this.

PR #659, opened 2018-12-20
fix(Popup): defaultOpen prop

### LINKED ISSUE #657: Popup defaultOpen is not working
### Steps
Setting defaultOpen on the Popup is not working

### Expected Result
If defaultOpen is set to true on the Popup it should be opened.

### Actual Result
The Popup is not opened.

### Version
0.15.0

### Testcase

mnajdova:
This PR fixes https://github.com/stardust-ui/react/issues/657 - `defaultOpen` prop on the `Popup` is not working.

### DIFF
--- a/src/components/Popup/Popup.tsx
+++ b/src/components/Popup/Popup.tsx
@@ -183,8 +183,6 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
     }
   }
 
-  public state = { target: undefined, open: false }
-
   public componentDidMount() {
     this.updateOutsideClickSubscription()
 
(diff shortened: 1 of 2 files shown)


src/components/Popup/Popup.tsx
```diff
@@ -183,8 +183,6 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
     }
   }
 
-  public state = { target: undefined, open: false }
-
```
sophieH29, line 187: Just to raise attention, there is an issue currently in Autocontrolled component. It doesn't allow to have both autocontrolled and non-controlled state.
Created this 1 week ago - https://github.com/stardust-ui/react/issues/610

Because it might be needed to have an initial state  explicitly set for `target`

kuzhelov (approved):
👍

src/components/Popup/Popup.tsx
mnajdova: the target is also an auto-controlled prop in this case, so it is safe to be removed.

PR #672, opened 2019-01-02
chore(Text): remove unnecessary variable overrides in Teams theme

miroslavstastny:
Teams theme should override only the variables which differ/are added compared to the base theme.
This PR removes redundant variable overrides for `Text` component as these were setting the same value base theme had already set.

No consumer facing change, CHANGELOG entry intentionally not added.

### DIFF
--- a/src/themes/teams/components/Text/textVariables.ts
+++ b/src/themes/teams/components/Text/textVariables.ts
@@ -1,4 +1,3 @@
-import { mapColorsToScheme } from '../../../../lib'
 import { TextVariables } from '../../../base/components/Text/textVariables'
 
 export interface TeamsTextVariables extends TextVariables {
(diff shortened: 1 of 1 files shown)


src/themes/teams/components/Text/textVariables.ts
```diff
@@ -7,41 +6,15 @@ export interface TeamsTextVariables extends TextVariables {
...
+  atMentionOtherColor: siteVariables.brand06,
+  atMentionMeColor: siteVariables.orange04,
+  atMentionMeFontWeight: siteVariables.fontWeightBold,
+  disabledColor: siteVariables.gray06,
+  errorColor: siteVariables.red,
+  importantWeight: siteVariables.fontWeightBold,
+  importantColor: siteVariables.red,
+  successColor: siteVariables.green04,
+  timestampColor: siteVariables.gray04,
+  timestampHoverColor: siteVariables.gray02,
```
mnajdova, line 19: Not for this PR, but maybe now that we have the color palette we should redefine the colors we are using with using variants with number, as it is done for the colors in the base theme.

mnajdova (approved):
👍

PR #673, opened 2019-01-02
fix(Accordion, Menu): consilidating unicode arrow characters across the components

mnajdova:
We have inconsistency of using the unicode arrow characters across different components: Accordinon, Menu, Dropdown etc.

## Accordion
image

## Menu
image

## Dropdown
The problem on the Dropdown is slightly different, as we are using icon name, that may not even be defined in some themes. (see comment https://github.com/stardust-ui/react/pull/584/files#r244750605)
image

This PR fixes this inconsistency for the Accordion and Menu components (as there is already started work on the Dropdown for handling the toggle button, I propose that it should be fixed in that PR).

This is the look of the components after the changes:

## Accordion
image

## Menu:
image

### DIFF
--- a/src/components/Accordion/AccordionTitle.tsx
+++ b/src/components/Accordion/AccordionTitle.tsx
@@ -66,7 +66,7 @@ class AccordionTitle extends UIComponent<ReactProps<AccordionTitleProps>, any> {
 
     return (
       <ElementType {...rest} className={classes.root} onClick={this.handleClick}>
-        {active ? <span>&#9660;</span> : <span>&#9654;</span>}
+        {active ? <span>&#x25BE;</span> : <span>&#x25B8;</span>}
         {content}
       </ElementType>
     )
--- a/src/components/Accordion/AccordionTitle.tsx
+++ b/src/components/Accordion/AccordionTitle.tsx
@@ -66,7 +66,6 @@ class AccordionTitle extends UIComponent<ReactProps<AccordionTitleProps>, any> {
 
     return (
       <ElementType {...rest} className={classes.root} onClick={this.handleClick}>
-        <span className={classes.arrow} />
         {content}
       </ElementType>
     )
--- a/src/components/Accordion/AccordionTitle.tsx
+++ b/src/components/Accordion/AccordionTitle.tsx
@@ -54,7 +54,7 @@ class AccordionTitle extends UIComponent<ReactProps<AccordionTitleProps>, any> {
   }
 
   renderComponent({ ElementType, classes, rest }) {
-    const { active, children, content } = this.props
+    const { children, content } = this.props
 
     if (childrenExist(children)) {
       return (
@@ -66,7 +66,7 @@ class AccordionTitle extends UIComponent<ReactProps<AccordionTitleProps>, any> {
 
     return (
       <ElementType {...rest} className={classes.root} onClick={this.handleClick}>
-        {active ? <span>&#x25BE;</span> : <span>&#x25B8;</span>}
+        <span className={classes.arrow} />
         {content}
       </ElementType>
     )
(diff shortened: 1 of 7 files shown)


notandrew:
I think we should be using the icons instead of unicode -- do we have a plan to move to using the icons?

image

mnajdova:
We will create shorthands for the icon for these, but as the icons are theme dependent, unicode characters  should be the default value. The 'chevron' icon might not exists in some themes. If we add some basic icons in the base theme, which is now under development, then we can safely replace the unicode characters with some icon, but still for the themes that will be implemented from scratch this could still be a problem. Until we decide on this points, adding the same unicode characters and icon shorthands should be enough for covering all points.

src/components/Accordion/AccordionTitle.tsx
```diff
@@ -66,7 +66,7 @@ class AccordionTitle extends UIComponent<ReactProps<AccordionTitleProps>, any> {
 
     return (
       <ElementType {...rest} className={classes.root} onClick={this.handleClick}>
-        {active ? <span>&#9660;</span> : <span>&#9654;</span>}
+        {active ? <span>&#x25BE;</span> : <span>&#x25B8;</span>}
```
sophieH29: Can we move the Unicode characters under some variables with descriptive naming?
for e.g. `const arrowDown = '&#x25BE;'`
If firstly look at code without running, can be a "guess" game to know what does it represent :)

silviuaavram:
Oleksandr had a nice idea about making those arrows non-selectable in my Dropdown PR. Maybe it will be useful to the arrows in this PR as well.

https://developer.mozilla.org/en-US/docs/Web/CSS/user-select

silviuaavram (approved):
Look at the user-select link from Shift.

src/components/Accordion/AccordionTitle.tsx
mnajdova: I moved all usages of the unicode characters in the styles and added constants in the teams theme.

mnajdova:
All comments resolved. I moved all unicode characters usages in the teams theme, and they can be safely move later in the base theme. In addition to this, we can add icon shorthands on all components. Let me know if you have other suggestions.

src/themes/teams/constants.ts
```diff
@@ -0,0 +1,8 @@
+const constants = {
```
alinais: I see these constants as part of the base theme. What do you think? Should we add it there now or later on, when individual components are going to have their base theme implementations?

mnajdova: I added initially the constants file as I was using this in the components' files. Now that we are using this here, they might become part of the siteVariables, I don't think there will be need for additional constants... For sure, the styles responsible for the arrows will go in the base theme.

mnajdova: As for the question of this specific unicode character, my opinion is that we should add them in the base theme, once we move one of these components' styles there. This arrow should create the very basic look of the arrows (not using any specific icons) which is exactly what our base theme is. Later we will add the icon slots, and each theme can further customize this.

alinais (approved):

src/themes/teams/siteVariables.ts
```diff
@@ -98,3 +98,9 @@ export const bodyFontSize = '1.4rem'
 export const bodyBackground = white
 export const bodyColor = black
 export const bodyLineHeight = lineHeightBase
+
+//
+// UNICODE CHARACTERS
+//
+export const arrowRight = '\u25B8'
```
miroslavstastny, line 105: This is not RTL-aware. Please create a tracking issue before merging this PR.

miroslavstastny (approved):

src/themes/teams/siteVariables.ts
mnajdova: Will do, thanks for noticing.

mnajdova: Here it is for refenrence: https://github.com/stardust-ui/react/issues/684

Issue #664, opened 2018-12-21
Expose `styles` in createComponent public API


miroslavstastny:
### Problem description
To compose `createComponent` with standard Stardust components (render Stardust component inside `createComponent`) it is not allowed to pass css-in-js classes to the Stardust component, styles must be used instead (#467).

### Proposed solution
Add `styles` to `createComponent` so that it is possible to pass styles to Stardust component

layershifter:
Fixed in #714.

PR #674, opened 2019-01-02
fix(Prototype):  Chat pane prototype accessibility improvements

kolaps33:


docs/src/prototypes/chatPane/chatPaneContent.tsx
```diff
@@ -26,12 +26,20 @@ class ChatPaneContainer extends React.PureComponent<ChatPaneContainerProps> {
...
-          />
+            aria-label="Message list. In forms mode: press Enter to explore message content, then use Escape to shift focus back to the message"
+          >
+            <div
+              id="chat-pane-reader-text"
+              style={screenReaderMessageContainerStyles}
+              role="heading"
+              aria-level={2}
+            >
+              Message list.
```
sophieH29, line 29: Is it enough descriptive to have just "Message list."? Previously, we used chat's title in naming, like `${chat.title} chat content.`

docs/src/prototypes/chatPane/chatPaneHeader.tsx
```diff
@@ -7,6 +7,17 @@ export interface ChatPaneHeaderProps {
   chat?: ChatData
 }
 
+const screenReaderMessageContainerStyles: React.CSSProperties = {
```
sophieH29: you're defining and using identical `screenReaderMessageContainerStyles` in 3 different files.
Would be better to have it in one place and export there (for e.g. create `styles.ts` and put there), then import these styles where you need.

And maybe to rename it `screenReaderMessageContainerStyles` -> `screenReaderContainerStyles`, as it is used not only for messages.

docs/src/prototypes/chatPane/chatPaneContent.tsx
kolaps33: I was following documentation with naming "Message list". 
The instruction are currently on the "main" element, as we didn't figure out anything better. In old code we set instruction on every single message, what was annoying for users. 
But there will be follow up discussion about instruction.

docs/src/prototypes/chatPane/composeMessage.tsx
```diff
@@ -1,18 +1,35 @@
...
+          role="region"
+          aria-labelledby="chat-compose-reader-text"
+          vertical
+          start={
+            <div>
+              <div
+                role="heading"
+                aria-level={2}
+                id="chat-compose-reader-text"
+                style={style.screenReaderContainerStyles}
```
layershifter: Instead of using styles I will propose to use our `createComponent()` API

docs/src/prototypes/chatPane/composeMessage.tsx
```diff
@@ -1,18 +1,35 @@
...
 class ComposeMessage extends React.Component {
   public render() {
     return (
-      <Layout
-        vertical
-        start={this.renderInput()}
-        main={this.renderToolbar()}
-        styles={{ padding: '16px 32px' }}
-      />
+      <div>
```
layershifter: Do you we need this `div`?

layershifter (changes requested):
LGTM, but I will propose to use `createComponent()` instead of `screenReaderMessageContainerStyles`

layershifter (approved):
Unblocked, #664 wasn't closed yet

PR #675, opened 2019-01-02
 fix(Accordion|Dropdown): fix default props in auto controlled mode

### LINKED ISSUE #610: AutoControlled component does not handle NOT autocontrolled state

layershifter:
Fixes #610.


src/components/Accordion/Accordion.tsx
```diff
@@ -118,8 +118,6 @@ class Accordion extends AutoControlledComponent<ReactProps<AccordionProps>, any>
   static Title = AccordionTitle
   static Content = AccordionContent
 
-  state: any = { activeIndex: [0] }
-
   getInitialAutoControlledState({ exclusive }) {
     return { activeIndex: exclusive ? -1 : [-1] }
```
layershifter, line 122: This line didn't work previously, that's why we have visual regressions there.

mnajdova: To me it seems, according to the changes in screener that that line was indeed working:
image
 as the previous accepted screen has the first element inside active. Not quite sure how did you conclude that it was not working before.. Maybe I am missing something?

mnajdova: I understand that this is part of the autocontroll props and should not be initialized, but can't understand why do you say it was not working previously

mnajdova (approved):

Issue #645, opened 2018-12-19
MenuItem's "onFocus" is not handled as prop


sophieH29:
Currently, ```MenuItem``` handles  onFocus event with this code:

```js
private handleFocus = (e: React.SyntheticEvent) => {
    this.setState(IsFromKeyboard.state())

    _.invoke(this.props, 'onFocus', e, this.props)
  }
```

BUT, ```onFocus``` is not added as handled props to ```propTypes``` and ```MenuItemProps```.
So, when passing ```onFocus``` from the client side, inner handling wouldn't work.

PR #677, opened 2019-01-03
fix(MenuItem): "onFocus" is not handled as prop

sophieH29:
Fix for issue https://github.com/stardust-ui/react/issues/645

### DIFF
--- a/src/components/Menu/MenuItem.tsx
+++ b/src/components/Menu/MenuItem.tsx
@@ -61,12 +61,11 @@ export interface MenuItemProps
   onClick?: ComponentEventHandler<MenuItemProps>
 
   /**
-   * Called on key down pressed.
-   *
+   * Called after user's focus.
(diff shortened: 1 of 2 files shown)


src/components/Menu/MenuItem.tsx
```diff
@@ -61,12 +61,11 @@ export interface MenuItemProps
   onClick?: ComponentEventHandler<MenuItemProps>
 
   /**
-   * Called on key down pressed.
-   *
+   * Called after user's focus.
    * @param {SyntheticEvent} event - React's original SyntheticEvent.
    * @param {object} data - All props.
    */
-  onKeyDown?: ComponentEventHandler<MenuItemProps>
```
sophieH29, line 69: Is not used

layershifter (approved):

layershifter:
The matching issue will not be closed automatically, the correct comment should be `Fixes #645`, https://help.github.com/articles/closing-issues-using-keywords/

PR #678, opened 2019-01-03
 fix(Popup): update positions after each render

### LINKED ISSUE #641: Async Popup doesn't update position and alignment
## Bug Report
When Popup content is loaded asynchronously and it needs to expand, Popup grows in a direction that no longer honors position and alignment

asyncpopup

### Steps
```jsx
import React from 'react'
import { Button, Popup, Segment } from '@stardust-ui/react'

const Component = () => <Segment styles={{minHeight: "300px"}}>Hello World</Segment>

const PopupExample = () => <Popup position="above" align="top" trigger={<Button icon="expand" />} content={{content: <AsyncData data={<Component />} />}} />

const render = data => {
  return !data ? "...loading" : data;
}

class AsyncData extends React.Component<{ render: Function; data?: any }> {
  state = { loading: true }

  componentDidMount() {
    setTimeout(() => { 
      console.log("TimedOut")
      this.setState({loading: false})
      
    }, 1000 )
  }

  public render() {
    const { loading } = this.state
    const { data } = this.props
    return loading ? "loading" : data
  }
} 

export default PopupExample
```

### Expected Result
After content is loaded and Popup resized, it should maintain its position and alignment

### Actual Result
Doesn't maintain position and alignment

### Version
0.15.0

layershifter:
Fixes #641.

----

normal
overflow


package.json
```diff
@@ -169,8 +169,5 @@
   "peerDependencies": {
     "react": "^16.3.0",
     "react-dom": "^16.3.0"
-  },
-  "resolutions": {
-    "create-react-context": "0.2.2"
```
layershifter, line 176: `react-popper` now defines max version of `create-react-context`

https://github.com/FezVrasta/react-popper/releases/tag/v1.3.1

mnajdova (approved):
👍 works like charm :)

miroslavstastny (approved):

kuzhelov:
unfortunately, there is another problem introduced by this fix - now an infinite render loop starts on Popup open.

In fact, we just need to force position update on async operation finish. I would suggest (as a very base step) to stick to the same strategy used by Popper - i.e. let the client explicitly decide when this update should be made.

popup-problem

PR #682, opened 2019-01-04
feat(Menu): adding kind prop for the items and MenuDivider component

### LINKED ISSUE #667: [RFC] Menu separator (vertical and horizontal)
## Feature Request
Implement Vertical Divider
### Problem description
Stardust already provides a [horizontal divider](https://stardust-ui.github.io/react/components/divider), however, it lacks a vertical divider. Supporting a "vertical" prop would come in handy. For instance to separate menu items in toolbars

mnajdova:
This PR adds the kind prop in the items prop in the Menu, for creating different components inside the menu (MenuItem, MenuDivider- component added as part of this PR). The PR is introduced in order to fix https://github.com/stardust-ui/react/issues/667 

For future work, we may leverage this kind property for extracting the MenuItem with submenus as different component, a MenuHeader component etc.

Example:
Vertical pointing menu:
```
import React from 'react'
import { Menu } from '@stardust-ui/react'
const items = [
  {
    key: 'editorials',
    content: 'Editorials',
  },
  {
    key: 'review',
    content: 'Reviews',
  },
  {
    key: 'events',
    content: 'Upcoming Events',
  },
]

const MenuExampleVerticalPointing = () => (
  <Menu defaultActiveIndex={0} items={items} vertical pointing />
)

export default MenuExampleVerticalPointing
```

image

With separator between the last two elements:
```
import React from 'react'
import { Menu } from '@stardust-ui/react'
const items = [
  {
    key: 'editorials',
    content: 'Editorials',
  },
  {
    key: 'review',
    content: 'Reviews',
  },
  {
    key: 'separator',
    kind: 'MenuSeparator',
  },
  {
    key: 'events',
    content: 'Upcoming Events',
  },
]

const MenuExampleKind = () => (
  <Menu defaultActiveIndex={0} items={items} vertical pointing="start" />
)

export default MenuExampleKind
```

image


src/themes/teams/components/Menu/menuItemStyles.ts
```diff
@@ -197,8 +174,6 @@ const menuItemStyles: ComponentSlotStylesInput<MenuItemPropsAndState, MenuVariab
         display: 'flex',
       }),
 
-      ...itemSeparator({ props, variables: v, theme }),
```
miroslavstastny: I am not sure that we want to remove separators in Teams theme menu

mnajdova: I had the same doubt, but wasn't sure how can I show the separator. Will add example with some other menu type that does not have separator by default then... Thanks for the suggestion.

mnajdova: Addressed the comment, added example for some other variants, take a look in the description of the PR.

docs/src/examples/components/Menu/Usage/MenuExampleKind.shorthand.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+import { Menu } from '@stardust-ui/react'
+
+const items = [
+  { key: 'editorials', content: 'Editorials' },
+  { key: 'review', content: 'Reviews' },
+  { key: 'separator', kind: 'MenuSeparator' },
```
layershifter: As I remember we decided to use lowercase without prefixes in the `kind` prop, i.e. `separator`.
And technically, it's a `divider`

src/components/Menu/MenuItem.tsx
```diff
@@ -144,6 +146,7 @@ class MenuItem extends AutoControlledComponent<ReactProps<MenuItemProps>, MenuIt
     defaultMenuOpen: PropTypes.bool,
     onActiveChanged: PropTypes.func,
     inSubmenu: PropTypes.bool,
+    kind: PropTypes.string,
```
layershifter:
```suggestion
    kind: PropTypes.oneOf(['divider']),
```

src/components/Menu/MenuItem.tsx
```diff
@@ -105,6 +105,8 @@ export interface MenuItemProps
 
   /** Indicates whether the menu item is part of submenu. */
   inSubmenu?: boolean
+
+  kind?: string
```
layershifter:
```suggestion
  kind?: 'divider'
```

src/themes/teams/components/Menu/menuSeparatorStyles.ts
```diff
@@ -0,0 +1,19 @@
...
+  root: ({ props: p, variables: v }): ICSSInJSStyle => {
+    return {
+      ...(p.vertical
+        ? {
+            borderTop: `1px solid ${p.primary ? v.primaryBorderColor : v.borderColor}`,
+          }
+        : {
+            borderLeft: `1px solid ${p.primary ? v.primaryBorderColor : v.borderColor}`,
+          }),
+    }
```
layershifter:
```js
const borderColor = p.primary ? v.primaryBorderColor : v.borderColor
const borderType = p.vertical ? 'borderTop' : 'borderLeft'

return {
  [borderType]: `1px solid ${borderColor}`
}
```

We can change naming, but this looks to me more readable and compact :)

src/components/Menu/MenuItem.tsx
mnajdova: Why should the type here be `oneOf` containing divider? MenuItem will actually never be created if the kind prop is divider. For the menu item it can be 'item' according to the convention, or nothing as this is the default component. Am I wrong assuming this?

layershifter: So, `kind: PropTypes.oneOf(['divider', 'item']),`? We should disallow to pass invalid values to `kind`

mnajdova: My point was that in the MenuItem, it should be `oneOf(['item'])`, and in the divider should be `oneOf(['divider'])` as the MenuItem will never be created if the kind prop is something other then item (or null, but the prop is anyways optional - anything else provided should throw an error), and the MenuDivider will never be created if the kind prop is something other then the 'divider'. Would this be sufficient?

mnajdova:
I've added some typings for the kind prop, some custom propTypes, and in the factories the kind prop is removed from the props when creating the component (thanks @layershifter for the suggestions). Please take a look again!

src/components/Menu/MenuItem.tsx
layershifter: Resolved in personal conversation :)

src/components/Menu/Menu.tsx
```diff
@@ -10,13 +10,17 @@ import {
...
 import MenuItem from './MenuItem'
 import { menuBehavior } from '../../lib/accessibility'
 import { Accessibility } from '../../lib/accessibility/types'
 
 import { ComponentVariablesObject } from '../../themes/types'
-import { ReactProps, ShorthandValue } from '../../../types/utils'
+import { ReactProps, ShorthandCollection } from '../../../types/utils'
+import MenuDivider from './MenuDivider'
+
+export type MenuItemKindOptions = 'divider' | 'item'
```
layershifter:
```suggestion
export type MenuItemKinds = 'divider' | 'item'
```

`MenuItemKinds` may be? Or even `MenuShorthandKinds` because they are related to shorthand items, not to `MenuItem`

docs/src/examples/components/Menu/Usage/MenuExampleKind.shorthand.tsx
```diff
@@ -0,0 +1,15 @@
...
+  { key: 'review', content: 'Reviews' },
+  { key: 'divider', kind: 'divider' as MenuItemKindOptions },
+  { key: 'events', content: 'Upcoming Events' },
+]
+
+const MenuExampleKind = () => (
+  <Menu defaultActiveIndex={0} items={items} vertical pointing="start" />
+)
+
+export default MenuExampleKind
```
layershifter:
```suggestion
export default MenuExampleDivider
```

As I understand this example should show us the usage of `MenuDivider`, SUI uses `Content` section of docs to show elements usage: https://semantic-ui.com/collections/menu.html#header

src/components/Menu/Menu.tsx
mnajdova: `MenuShorthandKinds` it is :)

layershifter (approved):

PR #685, opened 2019-01-04
feat(Loader): add component

### LINKED ISSUE #666: [RFC] Loader (Spinner)
## Feature Request
Startdust needs a progress indicator control (e.g. spinner and/or progress bar) that can be used to indicate to users that the app or some subset of its UI is loading.

### Problem description
I'm trying to migrate my app over from the Office UI Fabric to Stardust and have so far been unable to find a replacement for the Spinner control:
https://developer.microsoft.com/en-us/fabric#/components/spinner

I considered animating the refresh icon to create something that looks like a spinner but that doesn't seem to be a good long term solution.

### Proposed solution
Provide a basic Spinner control as part of the Stardust component library.

layershifter:
Fixes #666.

Initial specification: https://github.com/stardust-ui/specifications/pull/5

----

### `Loader` vs `Spinner`, `Indicator`...
See spec.

### Missing `color` prop
Will be added after #647.

### `label` vs `content`/`children`
SUIR calls this text, Fabric calls this label. Also, this part is a slot, so it cannot be directly children.
Opened there for feedback.


layershifter:
@jurokapsiar Juraj, please take a look on this PR. It seems that we need to add there some accessibility attributes.

layershifter:
- `labelPosition` was added to spec
- `indicator` slot is now configurable via prop

Thanks to @mnajdova for suggestions.

levithomason (approved):
👍

PR #688, opened 2019-01-07
Allow targeting of ListItem content

jurokapsiar:
Currently ListItem renders a `<li>` and group of `div`s in it. For accessibility purposes, it needs to behave similar to MenuItem and have a wrapper and item slot.

### DIFF
--- a/src/components/List/List.tsx
+++ b/src/components/List/List.tsx
@@ -128,7 +128,10 @@ class List extends AutoControlledComponent<ReactProps<ListProps>, ListState> {
           const targetComponent = this.itemRefs[index] && this.itemRefs[index].current
           const targetDomNode = ReactDOM.findDOMNode(targetComponent) as any
 
-          targetDomNode && targetDomNode.focus()
+          if (targetDomNode) {
+            const itemElements = targetDomNode.getElementsByClassName('ui-list__item')
+            if (itemElements.length > 0) itemElements.item(0).focus()
+          }
         })
       },
     )
--- a/src/components/List/ListItem.tsx
+++ b/src/components/List/ListItem.tsx
@@ -12,6 +12,8 @@ import ItemLayout from '../ItemLayout/ItemLayout'
 import { listItemBehavior } from '../../lib/accessibility'
 import { Accessibility, AccessibilityActionHandlers } from '../../lib/accessibility/types'
 import { ReactProps, ComponentEventHandler } from '../../../types/utils'
+import Slot from '../Slot/Slot'
+import cx from 'classnames'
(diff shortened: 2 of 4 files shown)


src/components/List/List.tsx
```diff
@@ -128,7 +128,10 @@ class List extends AutoControlledComponent<ReactProps<ListProps>, ListState> {
           const targetComponent = this.itemRefs[index] && this.itemRefs[index].current
           const targetDomNode = ReactDOM.findDOMNode(targetComponent) as any
 
-          targetDomNode && targetDomNode.focus()
+          if (targetDomNode) {
```
levithomason, line 131: Let's add a comment as to _why_ this special behavior is needed.  Accessibility requires arrow keys to focus the inner div of the LI, not the LI itself.  Correct?

src/components/List/List.tsx
```diff
@@ -128,7 +128,10 @@ class List extends AutoControlledComponent<ReactProps<ListProps>, ListState> {
...
+            const itemElements = targetDomNode.getElementsByClassName('ui-list__item')
```
levithomason, line 132: 'ui-list__item' => ListItem.className?

src/components/List/ListItem.tsx
```diff
@@ -132,11 +138,28 @@ class ListItem extends UIComponent<ReactProps<ListItemProps>> {
...
+        {...accessibility.keyHandlers.item}
         {...rest}
+        {...!wrapper && { onClick: this.handleClick }}
       />
     )
+
+    if (wrapper) {
+      return Slot.create(wrapper, {
+        defaultProps: {
+          className: cx('ui-list__item__wrapper', classes.wrapper),
```
levithomason, line 151: Consider reference to ListItem.className here as well opposed to hard coded string.

src/components/List/ListItem.tsx
```diff
@@ -132,11 +138,28 @@ class ListItem extends UIComponent<ReactProps<ListItemProps>> {
         headerMediaCSS={styles.headerMedia}
         contentCSS={styles.content}
         onClick={this.handleClick}
-        {...accessibility.attributes.root}
-        {...accessibility.keyHandlers.root}
+        {...accessibility.attributes.item}
+        {...accessibility.keyHandlers.item}
         {...rest}
+        {...!wrapper && { onClick: this.handleClick }}
```
sophieH29, line 144: you might need to remove `onClick` already added line 140

jurokapsiar:
this is now outdated as we will use Menu variant for the required purpose

PR #689, opened 2019-01-07
fix(ChatMessage): focus outline visible only during keyboard navigation 

kolaps33:
Fix that focus outline is visible only while user is in keyboard navigation.


jurokapsiar (approved):

sophieH29 (approved):

src/components/Chat/ChatMessage.tsx
```diff
@@ -38,12 +40,18 @@ export interface ChatMessageProps
 
   /** Timestamp of the message. */
   timestamp?: ShorthandValue
+
+  onFocus?: ComponentEventHandler<ChatMessageProps>
```
layershifter, line 49: Missing description of the prop.

src/components/Chat/ChatMessage.tsx
```diff
@@ -38,12 +40,18 @@ export interface ChatMessageProps
...
+}
+
+export interface MessageState {
```
layershifter: Should be `ChatMessageState`

layershifter (commented):
Added small comments, please also add a record to `CHANGELOG.md`

src/components/Chat/ChatMessage.tsx
kolaps33: added :)

layershifter (approved):

PR #694, opened 2019-01-08
 fix(Menu): correctly handle disabled item

### LINKED ISSUE #186: Add MenuItem disabled prop
## Feature Request
As a result of https://github.com/stardust-ui/react/pull/83 , it is required for the MenuItem Component to handle the `disabled` prop and act accordingly.
### Problem description
The `disabled` prop has been added to MenuItem, in order to be used to add the `aria-disabled` attribute on the element. However, adding this prop should trigger the actual disabled behaviour for the menu item, similar to how Button behaves, for instance.
### Proposed solution
Provide functionality to the MenuItem component in order to make it act similarly to Button when it receives the `disabled=true`

miroslavstastny:
Adds correct styles and fixes behaviors for disabled menu item. Fixes #186.

- [x] changelog
- [x] discuss behavior tests in `MenuItem-test.tsx` with @kolaps33 / @jurokapsiar  - why is behavior-test not enough?
  - behavior-test is enough, accessibility tests removed from `MenuItem-test.tsx`, accessibility integration smoke test added there.
- [x] improve tests for `IS_FOCUSABLE_ATTRIBUTE` (check it is set only if not `disabled`)
  - also changed how `aria-disabled` is handled - instead of adding `aria-disabled=false` the attribute is omitted.


kolaps33 (approved):

bmdalex (approved):

test/specs/components/Menu/MenuItem-test.tsx
```diff
@@ -46,203 +47,36 @@ describe('MenuItem', () => {
...
-        )
-      })
-    })
-
-    describe('as a ToolbarButton', () => {
-      test('root role should be presentation', () => {
-        const menuItemComponent = mountWithProviderAndGetComponent(
-          MenuItem,
-          <MenuItem accessibility={toolbarButtonBehavior} />,
+          <MenuItem disabled aria-label={ariaLabel} accessibility={accessibility.behavior} />,
```
sophieH29, line 68: It might also worth to check that onClick handler wouldn't be called when `disabled` prop is passed

sophieH29 (approved):

CHANGELOG.md
```diff
@@ -35,7 +36,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `MenuDivider` component and `kind` prop to the `items` inside of the `Menu` for creating different components @mnajdova ([#682](https://github.com/stardust-ui/react/pull/682))
 
 ### Fixes
-- Fix unicode arrow characters to be RTL aware ([#690](https://github.com/stardust-ui/react/pull/690))
+- Fix unicode arrow characters to be RTL aware @mnajdova ([#690](https://github.com/stardust-ui/react/pull/690))
```
layershifter, line 57: 👍

docs/src/examples/components/Menu/States/MenuExampleDisabled.shorthand.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+const items = [
+  { key: 'editorials', content: 'Editorials' },
+  { key: 'review', content: 'Disabled Item', disabled: true },
+  { key: 'events', content: 'Upcoming Events' },
+]
+
+const MenuExampleDisabled = () => <Menu defaultActiveIndex={0} items={items} />
+
+export default MenuExampleDisabled
```
layershifter:
```suggestion
export default MenuExampleItemDisabled
```

May be this name will be more obvious?

layershifter (approved):

test/specs/behaviors/testDefinitions.ts
```diff
@@ -124,20 +124,68 @@ definitions.push({
   },
 })
 
-// Example: Adds attribute 'aria-disabled=true' to 'trigger' component's part based on the property 'disabled'.
+function testMethodConditionallyAddAttribute(
+  parameters,
+  attributeToBeAdded,
```
kuzhelov: would suggest to rearrange these method's args so its intent will be easier to read. The following option might work better:

```js
function testMethodConditionallyAddAttribute(
  parameters,
  component,
  propertyDependsOn,
  attributeToBeAdded,
  valueOfAttributeToBeAddedIfTrue,
  valueOfAttributeToBeAddedOtherwise
)
```

test/specs/behaviors/testDefinitions.ts
```diff
@@ -124,20 +124,68 @@ definitions.push({
...
+  ][attributeToBeAdded]
+  expect(testHelper.convertToBooleanIfApplicable(expectedResultAttributeDefined)).toEqual(
+    testHelper.convertToBooleanIfApplicable(valueOfAttributeToBeAddedIfTrue),
+  )
+}
+
+// Example: Adds attribute 'aria-disabled=true' to 'trigger' component's part if 'disabled' property is true. Does not set the attribute otherwise.
 definitions.push({
-  regexp: /Adds attribute '([a-z A-Z -]+)=([a-z 0-9]+)' to '([a-z -]+)' component's part based on the property '([a-z -]+)'\./g,
+  regexp: /Adds attribute '([a-zA-Z-]+)=([a-z0-9]+)' to '([a-z-]+)' component's part if '([a-z -]+)' property is true\. Does not set the attribute otherwise\./g,
```
kuzhelov, line 191: 👍 for regex fixes

test/specs/behaviors/testDefinitions.ts
```diff
@@ -124,20 +124,68 @@ definitions.push({
...
+    )
+  },
+})
+
+// Example: Adds attribute 'aria-disabled=true' to 'trigger' component's part if 'disabled' property is true. Sets the attribute to 'false' otherwise.
+definitions.push({
+  regexp: /Adds attribute '([a-zA-Z-]+)=([a-z0-9]+)' to '([a-z-]+)' component's part if '([a-z -]+)' property is true\. Sets the attribute to '([a-z0-9]+)' otherwise\./g,
+  testMethod: (parameters: TestMethod) => {
+    const [
+      attributeToBeAdded,
```
kuzhelov, line 214: totally understand that this was introduced before, but in future would suggest to refactor `parameters.props` in a way that it would return an object with named props - otherwise it is very hard to remember (and maintain) an order of more than two variables that are result of array destructuring

kuzhelov (approved):

docs/src/examples/components/Menu/States/MenuExampleDisabled.shorthand.tsx
miroslavstastny: Removing this example, it has been already added by @codepretty in another PR.

test/specs/behaviors/testDefinitions.ts
miroslavstastny: Let's discuss this separately

test/specs/components/Menu/MenuItem-test.tsx
miroslavstastny: We have a test for `onClick` in `Menu-test.tsx`, I will add a test for this there.

test/specs/behaviors/testDefinitions.ts
miroslavstastny: there are more to fix, will do that in a separate PR.

test/specs/behaviors/testDefinitions.ts
miroslavstastny: done, bonus parameter added

PR #698, opened 2019-01-10
fix(utils): export ShorthandRenderer

miroslavstastny:
Export `ShorthandRenderer` type.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -163,3 +163,4 @@ export {
   ChildrenComponentProps,
   ContentComponentProps,
 } from './lib'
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -163,3 +163,4 @@ export {
   ChildrenComponentProps,
   ContentComponentProps,
 } from './lib'
+export { ShorthandRenderer } from '../types/utils'
```
kuzhelov, line 166: yes, returning to your comment about how types are arranged on files - we should definitely consider some grooming there

kuzhelov (approved):

PR #5, opened 2018-07-24
feat(Text): introduce dir attribute to support RTL

kuzhelov:
This proposal is aimed on introducing proper text rendering in RTL mode. Its idea consists of two points:
- introduce `dir="auto"` attribute value to the text rendered by `Text` component
- from now on ensure that all components that are about rendering text information as part of their presentation should use `Text` component to wrap this textual part

PR #704, opened 2019-01-11
feat(rtl): add span element with dir: 'auto' for the strings used in the Stardust components

mnajdova:
This PR aims to fix the issues raised in the following https://github.com/stardust-ui/react/pull/5#issuecomment-439881600 

> In general, we need to get to a state where the text nodes are inside of elements that have dir='auto' attribute. This allows the browser to detect the text direction.
> 
> If the text is composed of multiple parts (typically a translation with parameters), each of the parts needs to be separated in <span dir='auto'> element. This is however out of scope for Stardust and should be handled in the i18n layer of the application.
> 
> Requirements:
>
> - [x] use Text component to render any text inside of standard Stardust components - **NOTE**: this is reworked in a different way, by providing `rtlProps` in the renderComponent function which currently contains only `{dir: 'auto'}` if the children or content in the component are plain string. In the future we may reworked this in a similar way the accessibility are defined, by adding slots and applying these props in different parts of the component
> - [x] Text component renders <span dir='auto>{content}</span> if the type of content is a string
if user provides HTML/React nodes as an input for the component content, the user is responsible for adding dir='auto' by himself. This needs to be documented
> - [ ] revisit applying of truncateStyle in buttonStyles.ts - it needs to be applied even if the content contains <span> or other elements - **NOTE**: This was broken by one of the latest change sin the factories - if there is children inside the factory it is returning just the element provided there.


src/lib/generateContent.tsx
```diff
@@ -0,0 +1,22 @@
...
+import Text from '../components/Text/Text'
+
+export const generateContentSlot = (
+  content: ShorthandValue | ShorthandRenderCallback,
+  options: CreateShorthandOptions,
+) => {
+  if (typeof content === 'string') {
+    return Text.create(content, options)
+  }
+  return Slot.create(content, options)
```
jurokapsiar: maybe we can just return `content` in this case, let's discuss pros and cons.

bmdalex (approved):

src/lib/renderComponent.tsx
```diff
@@ -221,6 +224,10 @@ const renderComponent = <P extends {}>(config: RenderConfig<P>): React.ReactElem
         const classes: ComponentSlotClasses = getClasses(renderer, mergedStyles, styleParam)
         classes.root = cx(className, classes.root, props.className)
 
+        const { children, content } = props
+        const evaluatedChildren = getRtlAwareElement(children)
```
kuzhelov: may suggest to rename these two to something that is less abstract in terms of providing their intent (as `evaluatedChildren` is quite general). Think that something with `rtl` in its name would work better

mnajdova: How about `rtlTransformedChildren` & `rtlTransformedContent`?

src/components/Header/HeaderDescription.tsx
```diff
@@ -36,10 +36,10 @@ class HeaderDescription extends UIComponent<ReactProps<HeaderDescriptionProps>,
     as: 'p',
   }
 
-  renderComponent({ ElementType, classes, unhandledProps }) {
+  renderComponent({ ElementType, classes, unhandledProps, accessibility }) {
     const { children, content } = this.props
     return (
-      <ElementType {...unhandledProps} className={classes.root}>
+      <ElementType {...unhandledProps} {...accessibility.attributes.root} className={classes.root}>
```
kuzhelov: should we put `unhandledProps` as the last ones there - as otherwise there won't be any possibility for overrides opened to the client?

src/components/Button/Button.tsx
```diff
@@ -139,10 +140,10 @@ class Button extends UIComponent<ReactProps<ButtonProps>, ButtonState> {
         {...accessibility.attributes.root}
         {...unhandledProps}
       >
-        {hasChildren && children}
+        {hasChildren && rtlTransformedChildren}
         {!hasChildren && iconPosition !== 'after' && this.renderIcon(variables, styles)}
         {Slot.create(!hasChildren && content, {
-          defaultProps: { as: 'span', className: classes.content },
+          defaultProps: { as: 'span', styles: styles.content },
```
kuzhelov, line 147: 👍

src/components/Accordion/AccordionTitle.tsx
```diff
@@ -53,20 +53,18 @@ class AccordionTitle extends UIComponent<ReactProps<AccordionTitleProps>, any> {
...
+    unhandledProps,
+    rtlTransformedChildren,
+    rtlTransformedContent,
+  }) {
+    const { children } = this.props
 
     return (
       <ElementType {...unhandledProps} className={classes.root} onClick={this.handleClick}>
-        {content}
+        {childrenExist(children) ? rtlTransformedChildren : rtlTransformedContent}
```
kuzhelov: 👍

src/lib/renderComponent.tsx
```diff
@@ -221,6 +224,10 @@ const renderComponent = <P extends {}>(config: RenderConfig<P>): React.ReactElem
         const classes: ComponentSlotClasses = getClasses(renderer, mergedStyles, styleParam)
         classes.root = cx(className, classes.root, props.className)
 
+        const { children, content } = props
+        const rtlTransformedChildren = getRtlTransformedElement(children)
```
kuzhelov: it seems that this step is not necessary in quite majority of rendering cycles. What if we will just provide a function to the client component instead, so that RTL transform would be approached the following way:

```jsx
class SomeComponent extends UIComponent {
   renderComponent({ ..., addRtlSupport }) {   // just one option for naming, also we might think about consuming this function directly from import
      const { children, content } = props
   
       <ElementType>
          {addRtlSupport(childrenExist(children) ? children : content)}
      </ElementType>
   }
}
```

Additional benefit is that we will eliminate ambiguity in choosing where `children` or `content` should be consumed from - as with current approach they should be consumed from props if not transformed, and from render args otherwise.

mnajdova: Will add the `addRtlSupport` as part of the lib, as adding it just to the renderComponent will require from use to always pass it to other factories methods that may need it... Then we can safely just use it as a dependency.

jurokapsiar:
This looks correct to me, however if you look at our first Button example and switch to RTL, the truncation is not correct:

image

It should behave as https://jsfiddle.net/j3n6bezx/ (thanks @sophieH29 for the original fiddle)

The button example works correctly, when `dir='auto'` is applied on the parent box span instead of a separate span. (as the box also has the overflow styles)

src/components/Header/Header.tsx
```diff
@@ -54,26 +55,21 @@ class Header extends UIComponent<ReactProps<HeaderProps>, any> {
...
+    const hasChildren = childrenExist(children)
 
     return (
       <ElementType {...unhandledProps} className={classes.root}>
-        {content}
-        {HeaderDescription.create(description, {
-          defaultProps: {
-            variables: {
-              ...(v.descriptionColor && { color: v.descriptionColor }),
+        {hasChildren ? addRtlSupport(children) : addRtlSupport(content)}
```
kuzhelov: can be simplified to `addRtlSupport(hasChildren ? children : content)`

kuzhelov (approved):

mnajdova:
@jurokapsiar thanks for the explanation, it is fixed now. Please take a look.

src/components/Accordion/Accordion.tsx
```diff
@@ -21,6 +21,8 @@ import {
   ShorthandValue,
   ShorthandRenderFunction,
 } from '../../../types/utils'
+import { childrenDependentRtlAttributes } from '../../lib/rtl'
+import { RtlFunc } from '../../lib/rtl/types'
```
kuzhelov: would expect to consume both type and entity from single module  - could you, please, suggest why this separation is done?

src/components/Form/Form.tsx
```diff
@@ -26,6 +28,12 @@ export interface FormProps extends UIComponentProps, ChildrenComponentProps {
    * @param {object} data - All props.
    */
   onSubmit?: ComponentEventHandler<FormProps>
+
+  /**
+   * Rtl attributes function if overridden by the user.
+   * @default formRtlAttributes
+   */
+  rtlAttributes?: RtlFunc
```
kuzhelov: it seems like it would be more expressive if we'll rename type to `RtlAttributesProvider`, what do you think?

src/components/Header/Header.tsx
```diff
@@ -24,6 +27,12 @@ export interface HeaderProps
 
   /** Align header content. */
   textAlign?: 'left' | 'center' | 'right' | 'justified'
+
+  /**
+   * Rtl attributes function if overridden by the user.
+   * @default radioGroupRtlAttributes
```
kuzhelov: this one lies about default prop

src/components/Header/Header.tsx
```diff
@@ -45,35 +54,33 @@ class Header extends UIComponent<ReactProps<HeaderProps>, any> {
...
 
     return (
-      <ElementType {...unhandledProps} className={classes.root}>
-        {content}
-        {HeaderDescription.create(description, {
-          defaultProps: {
-            variables: {
-              ...(v.descriptionColor && { color: v.descriptionColor }),
+      <ElementType {...rtlAttributes.root} {...unhandledProps} className={classes.root}>
+        {description ? contentElement : addRtlSupport(contentElement)}
```
kuzhelov: don't like the fact that this general approach like 'calculate attributes during rendering and apply them to <ElementType />' breaks here: the `addRtlSupport` logic will unconditionally apply the same 'rtl behavior' (children-dependent logic) to `contentElement`, whatever this is. 

Could we generalize our approach to handle this case as well? Like, for example, introduce `description` prop for RTL attributes provider? Otherwise, I am not sure how this approach might scale to handle similar cases

src/components/Label/Label.tsx
```diff
@@ -65,12 +73,14 @@ class Label extends UIComponent<ReactProps<LabelProps>, any> {
...
     imagePosition: PropTypes.oneOf(['start', 'end']),
     fluid: PropTypes.bool,
+    rtlAttributes: PropTypes.func,
   }
 
   static defaultProps = {
     as: 'span',
     imagePosition: 'start',
     iconPosition: 'end',
+    rtlAttributes: childrenDependentRtlAttributes,
```
kuzhelov: just to clarify for myslef - could you, please, suggest why it cannot be used as a default RTL attributes provider?

kuzhelov: or, I see, nvm

kuzhelov (commented):
have left couple comments - could we discuss them before merge?

src/components/Header/Header.tsx
mnajdova: The general problem is that, in cases when the calculation for the rtl props are failing (we are returning empty object) we want to add `<span dir='auto'>` around the element. I am open to other ideas..

src/components/Header/Header.tsx
mnajdova: Thanks for the catch, will change it.

src/components/Accordion/Accordion.tsx
mnajdova: Will export the types from the index file as well.

src/components/Form/Form.tsx
mnajdova: Agreed, let me rename it.

src/lib/rtl/Layout/layoutRtlAttributes.ts
```diff
@@ -0,0 +1,16 @@
+import { RtlAttributesProvider } from '../types'
```
kuzhelov: cannot find place where this one is used

src/lib/rtlTextContainer.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const rtlTextContainer = {
+  getAttributes: ({
+    condition = true,
+    forElements = [],
+  }: {
+    condition?: boolean
+    forElements: any[]
+  }) => {
+    if (condition) {
```
kuzhelov:
```js
return condition && forElements.some(child => typeof child === 'string')
  ? { dir: 'auto' }
  : {}
```

Yes, it is not that efficient (although still have the same asymptotic complexity;), but it will be more declarative, and given that there are no more than two items expected to be in this array, this seems to be a reasonable price

src/components/Header/Header.tsx
```diff
@@ -54,26 +56,29 @@ class Header extends UIComponent<ReactProps<HeaderProps>, any> {
...
-        {HeaderDescription.create(description, {
-          defaultProps: {
-            variables: {
-              ...(v.descriptionColor && { color: v.descriptionColor }),
-            },
-          },
+      <ElementType
+        {...rtlTextContainer.getAttributes({
+          forElements: [children, content],
+          condition: !description,
```
kuzhelov, line 68: maybe we could use `if` instead of `condition`, as it would be a bit easier to read. However, please, feel free to decide by yourself on that :)

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add and export 'missed call' icon in Teams theme @codepretty ([#748](https://github.com/stardust-ui/react/pull/748))
 - Add `Indicator` component and used it in `MenuItem` and `AccordionTitle` @mnajdova ([#721](https://github.com/stardust-ui/react/pull/721))
 - Expose `renderItem` and `renderSelectedItem` callbacks API for `Dropdown` @layershifter ([#746](https://github.com/stardust-ui/react/pull/746))
+- Add rtl support for the strings used inside the components @mnajdova ([#704](https://github.com/stardust-ui/react/pull/704))
```
kuzhelov: lets write RTL in uppercase here

kuzhelov (approved):

PR #705, opened 2019-01-11
fix(Popup): opened popup causes infinite rendering loop

### LINKED ISSUE #706: Opened Popup causes infinite render loop
### Steps
popup-problem

### Expected Result
There is at least finite amount of rerendering cycles involved in rendering opened popup.

### Actual Result
Infinite React rerendering loop is triggered. 💣

### Version
0.16.1

kuzhelov:
Fixes #706.

## TODO
- [x] provide an example for introduced API
- [x] update CHANGELOG

-----------------

The problem was introduced by #678  - essentially, the root cause is the side effect of `scheduleUpdate()` function that causes React to rerender components tree - and given that each rendering ends up in another render being immediately scheduled, the whole story ends up in an infinite loop. infinite loop.

Provided fix introduces ability to the client to explicitly tell when this popup position update should happen. Please, note that this scenario is fairly rare, so that normally `content` prop of the `Popup` will be used, as it was before. However, in cases where fine control is necessary over the moment of popup position update, the introduced approach now can be used:

```jsx
<Popup 
  trigger={..} 
  renderContent={updatePosition => <AsyncData done={updatePosition} ... />} />
```

This will prevent us from implicit scheduling of render updates and, thus, remove probability of falling into infinite render loops altogether. At the same time, all the user scenarios mentioned in the following issue (#641) will be covered.


src/components/Popup/Popup.tsx
```diff
@@ -411,9 +418,11 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
     popupPositionClasses: string,
     rtl: boolean,
     accessibility: AccessibilityBehavior,
+    // https://popper.js.org/popper-documentation.html#Popper.scheduleUpdate
     { ref, scheduleUpdate, style: popupPlacementStyles }: PopperChildrenProps,
   ) => {
-    const { content } = this.props
+    const { content: propsContent, renderContent } = this.props
+    const content = renderContent ? renderContent(scheduleUpdate) : propsContent
```
mnajdova, line 425: Not sure but, maybe it will be nice if we can have an example with this, especially that we are adding some new prop for the API. Other then that looks good!

mnajdova (approved):

src/components/Popup/Popup.tsx
kuzhelov: totally agree, will work on it!

bmdalex (approved):

miroslavstastny (approved):

PR #711, opened 2019-01-13
feat(chore): run visual tests in parallel

kuzhelov:
Introduces changes that run build and visual tests in parallel.
### Average CI build time change 
| was | now | diff (avg)
|-----|------|-----|
| 12-14 mins | 6-8 mins | 45% |


.circleci/config.yml
```diff
@@ -89,3 +94,41 @@ jobs:
...
               npm publish --tag current
             fi
+
+  visual:
+    docker:
+      - image: circleci/node:8
+    environment:
+      TZ: "/usr/share/zoneinfo/America/Los_Angeles"
+    steps:
+      - run:
```
miroslavstastny, line 104: Instead of copy pasting these steps, I would prefer to extract them to commands:
https://circleci.com/docs/2.0/reusing-config/#authoring-reusable-commands

miroslavstastny (approved):

.circleci/config.yml
kuzhelov: have initially tried this approach, for exactly the same reasons you've mentioned - have repeated again to remind the reasons I have abandoned it. The problem is that `commands` section is a feature of v2.1 of Circle CI config, and the key problem is that to support this version the following setting should be enabled for the project: 

image

Given that this setting is in Preview, as well as the caveat that 'once moved to v2.1 ensure that .config won't be reverted back to v2.0' from here https://github.com/CircleCI-Public/config-preview-sdk/blob/master/README.md#important-21-configuration-caveats (and, of course, those things that work differently between these versions), I would rather wait the moment when these features will pass the 'Preview' stage (due to potential syntax changes). 

Don't get me wrong - while I totally acknowledge the problem of duplication, provided costs seem to outweighs the necessity of dedup the code that is rarely changed - am fully opened to discuss this in person, but for now it seems that better strategy is ti wait for this v2.1 'commands' feature.

PR #429, opened 2018-11-05
breaking(Menu): replace `type` with `primary`/`secondary`

layershifter:
# BREAKING CHANGES

Refs #421.

#### Before

```jsx
<Menu type='primary' />
```

#### After

```jsx
<Menu primary />
```

#### Before

```ts
  typePrimaryActiveColor: string
  typePrimaryActiveBackgroundColor: string
  typePrimaryActiveBorderColor: string

  typePrimaryBorderColor: string
  typePrimaryHoverBorderColor: string
  typePrimaryUnderlinedBorderColor: string
```

### After

```ts
  primaryActiveColor: string
  primaryActiveBackgroundColor: string
  primaryActiveBorderColor: string

  primaryBorderColor: string
  primaryHoverBorderColor: string
  primaryUnderlinedBorderColor: string
```

PR #712, opened 2019-01-14
feat(menu): color prop

bmdalex:
### Description

This PR:
- adds `color` prop to `Menu` component
- creates `Menu` color examples

### API

```jsx
<Menu color={COLOR} .../>
```
 where `COLOR` can be:
- a string: `'primary' | 'secondary' | 'blue' | 'green' | 'grey'
- an object with:
  - key: one of `'foreground' | 'background' | 'border'` to identify the target area of the component
  - value: one of the strings above to describe  the color of the area
e.g.:
```jsx
<Menu color={ foreground: 'pink', background: 'yellow', border: 'green' }} .../>
```
renders:
screenshot 2019-01-14 at 02 47 47

### List of colored menu examples:
screenshot 2019-01-14 at 02 52 10

### DIFF
--- a/docs/src/examples/components/Menu/Variations/MenuExampleColor.shorthand.tsx
+++ b/docs/src/examples/components/Menu/Variations/MenuExampleColor.shorthand.tsx
@@ -1,0 +1,61 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+import { Menu, ProviderConsumer, Grid, Text } from '@stardust-ui/react'
+
+const items = [
+  { key: 'editorials', content: 'Editorials' },
+  { key: 'review', content: 'Reviews' },
+  { key: 'events', content: 'Upcoming Events' },
+]
+
+const iconItems = [
+  { key: 'home', content: 'Home', icon: 'home' },
+  { key: 'users', content: 'Users', icon: 'users' },
+  { key: 'search', icon: 'search' },
+]
+
+const MenuExampleColor = () => (
+  <ProviderConsumer
+    render={({ siteVariables: { colorScheme } }) => {
+      const colorsArr = _.keys(colorScheme)
+      const colors = _.times(7, num => colorsArr[num % colorsArr.length])
+
+      return (
+        <Grid
+          columns="repeat(2, auto)"
+          styles={{ justifyContent: 'left', justifyItems: 'left', alignItems: 'center' }}
+          variables={{ gridGap: '10px' }}
+        >
+          <Text content={`${_.upperCase(colors[0])} DEFAULT MENU:`} weight="bold" />
+          <Menu defaultActiveIndex={0} color={colors[0]} items={items} />
+          <Text content={`${_.upperCase(colors[1])} PILLS MENU:`} weight="bold" />
+          <Menu defaultActiveIndex={0} color={colors[1]} pills items={items} />
+          <Text content={`${_.upperCase(colors[2])} POINTING MENU:`} weight="bold" />
+          <Menu defaultActiveIndex={0} color={colors[2]} pointing items={items} />
+          <Text content={`${_.upperCase(colors[3])} VERTICAL POINTING MENU:`} weight="bold" />
+          <Menu defaultActiveIndex={0} color={colors[3]} vertical pointing items={items} />
+          <Text content={`${_.upperCase(colors[4])} UNDERLINED MENU:`} weight="bold" />
+          <Menu defaultActiveIndex={0} color={colors[4]} underlined items={items} />
+          <Text content={`${_.upperCase(colors[5])} ICON MENU:`} weight="bold" />
+          <Menu defaultActiveIndex={0} color={colors[5]} items={iconItems} />
+          <Text content={`${_.upperCase(colors[6])} ICON ONLY MENU:`} weight="bold" />
+          <Menu
+            defaultActiveIndex={0}
+            color={colors[6]}
+            iconOnly
+            items={iconItems.map(item => _.pick(item, ['key', 'icon']))}
+          />
+          <Text content={`UNDERLINED MENU (mutiple colors):`} weight="bold" />
+          <Menu
+            defaultActiveIndex={0}
+            underlined
+            color={{ foreground: colors[4], background: colors[6], border: colors[1] }}
+            items={items}
+          />
+        </Grid>
+      )
+    }}
+  />
+)
+
+export default MenuExampleColor
--- a/docs/src/examples/components/Menu/Variations/MenuExampleVerticalPointingEnd.shorthand.tsx
+++ b/docs/src/examples/components/Menu/Variations/MenuExampleVerticalPointingEnd.shorthand.tsx
@@ -8,7 +8,7 @@ const items = [
 ]
 
 const MenuExampleVerticalPointingEnd = () => (
-  <Menu defaultActiveIndex={0} items={items} vertical pointing="end" />
+  <Menu defaultActiveIndex={0} items={items} vertical pointing="end" primary />
 )
 
 export default MenuExampleVerticalPointingEnd
--- a/src/lib/colorUtils.ts
+++ b/src/lib/colorUtils.ts
@@ -18,10 +18,11 @@ export const mapColorsToScheme = <T>(
     typeof mapper === 'number' ? String(mapper) : (mapper as any),
   ) as ColorValues<T>
 
-export const getColorSchemeFn = <T>(colorProp: string, colorScheme: ColorValues<T>) => {
-  const colors = _.get(colorScheme, colorProp)
-  return (area: keyof T, defaultColor: string) => (colors ? colors[area] : defaultColor)
-}
+export const getColorFromScheme = <T extends {}>(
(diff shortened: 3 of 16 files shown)


src/themes/types.ts
```diff
@@ -99,12 +99,10 @@ export type ColorPalette = ExtendablePalette<
...
  * A type for the generic color scheme of a component based on CSS property names
  */
-export type ColorScheme = {
-  foreground: string
-  background: string
-  border: string
-  shadow: string
-}
+export type ColorScheme = Record<
+  'foreground' | 'background' | 'border' | 'shadow' | 'lighterBackground',
```
mnajdova, line 103: What is `lighterBackground` representing here? Maybe if we need different background it is result of some state inside the component. Should we consider adding object containing colors for the states: hovered, focus etc?

src/themes/teams/components/Menu/menuItemStyles.ts
```diff
@@ -1,6 +1,13 @@
+import * as _ from 'lodash'
+
 import { getSideArrow } from '../../utils'
-import { pxToRem } from '../../../../lib'
-import { ComponentSlotStyleFunction, ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
+import { pxToRem, getColorFromScheme } from '../../../../lib'
```
mnajdova, line 4: I remember I used `getColorFromScheme` in the renderComponent, so that the user will receive colors object in the styles containing the merged colors. Please take a look on that. If that is not sufficient, we may rework the mechanism there and try to avoid this method in the styles.

src/themes/types.ts
mnajdova: Or maybe it is a color of other type of the Menus?

src/lib/colorUtils.ts
```diff
@@ -61,7 +62,7 @@ export const generateColorScheme = (
   // if the color prop is not defined, but the the color scheme is defined, then we are returning
   // the defaults from the color scheme if they exists
   if (colorScheme) {
-    return colorScheme && colorScheme.default ? colorScheme.default : {}
+    return colorScheme.default || {}
```
mnajdova, line 65: 👍

docs/src/examples/components/Menu/Variations/MenuExampleColor.shorthand.tsx
```diff
@@ -0,0 +1,61 @@
...
+            color={{ foreground: colors[4], background: colors[6], border: colors[1] }}
+            items={items}
+          />
+        </Grid>
+      )
+    }}
+  />
+)
+
+export default MenuExampleColor
```
layershifter, line 61: This example looks overcomplicated to me, may be we can:
- introduce a new `Select` knob
- use it to get select an active color

docs/src/examples/components/Menu/Variations/MenuExampleVerticalPointingEnd.shorthand.tsx
```diff
@@ -8,7 +8,7 @@ const items = [
 ]
 
 const MenuExampleVerticalPointingEnd = () => (
-  <Menu defaultActiveIndex={0} items={items} vertical pointing="end" />
+  <Menu defaultActiveIndex={0} items={items} vertical pointing="end" primary />
```
layershifter, line 11: This looks like a breaking change, is `primary` required now to get the same styling?

codepretty: Why not get rid of 'primary' altogether and just use the new color prop?

src/themes/teams/components/Menu/menuItemStyles.ts
```diff
@@ -328,10 +368,13 @@ const menuItemStyles: ComponentSlotStylesInput<MenuItemPropsAndState, MenuVariab
 
         ...(primary
```
codepretty, line 369: I think this would be greatly simplified (easier to redline & maintain) if we can remove `primary` prop and use just the `color` prop, since in Teams world `primary` is synonymous with just using brand color anyway. I think having `primary` prop makes sense on button and maybe some other components, but not for menu.

src/themes/base/colors.ts
```diff
@@ -169,11 +169,13 @@ export const colorScheme: ColorSchemeMapping = _.mapValues(
       border: foreground,
       shadow: foreground,
       background: colorVariants[500],
+      lighterBackground: colorVariants[300],
```
codepretty, line 172: What is the purpose of lighterBackground? The other keys make sense to me. Should lighterBackground really just be a different value of background?

docs/src/examples/components/Menu/Variations/MenuExampleVerticalPointingEnd.shorthand.tsx
layershifter: Previously, we discussed this as I remember:
- replace `type` with `primary` and `secondary` props, #429
- remove `primary` and `secondary` from `Divider` as it doesn't applies any styling, #558
- on components like `Menu`, `primary` and `secondary` can affect styling, too: https://react.semantic-ui.com/collections/menu/#types-secondary

May be, in this case we should not expose `primary` and `secondary` as colors to avoid confusions.

src/themes/base/colors.ts
bmdalex: `lighterBackground` was just something temporary I came up before we came up with a more robust color palette that would allow different shades of background|foreground|border|etc color for different UI states (active|hovered|focused|etc).

**tl;dr**
will remove it soon and add something like:

```typescript
background: {
  initial: colorVariants[500],
  hover: colorVariants[300],
},
```

docs/src/examples/components/Menu/Variations/MenuExampleVerticalPointingEnd.shorthand.tsx
bmdalex: @layershifter 
@codepretty 

We can agree there's a confusion from having `primary` and `color='primary'` props.
Let me just summarize our options:

### 1. remove `primary` color from `Menu` color palette and use `primary` prop

@layershifter I know we just discussed this, but after trying to code this up I realized how complicated it becomes because of the conditional logic for styles as @codepretty points out

We always need to check if the user set a color and if not default to either a primary variable color or a regular variable color (depending on `primary` prop value).

This cannot be extracted in a function because of all the possible shades of primary and regular colors.


### 2. change the logic for `primary` prop to be reflected by `color='primary'` instead (as @codepretty suggests); `primary` boolean prop does nothing for Teams theme

If we use `color='primary' as @codepretty suggests, the code will indeed be simplified tremendously as we do not have to handle the primary cases anymore.

Also, it looks like the colors are the only different styles for `<Menu primary .../>` so changing it to use `<Menu color='primary' .../> to achieve the purple menus does not cause other regressions!

To avoid confusion we should think about removing `primary` boolean prop or rename either the boolean prop or the name of the color?

What do others think?
@mnajdova @kuzhelov @miroslavstastny

src/themes/teams/components/Menu/menuItemStyles.ts
bmdalex: I agree, pls take a look at https://github.com/stardust-ui/react/pull/712#discussion_r247989391

docs/src/examples/components/Menu/Variations/MenuExampleColor.shorthand.tsx
bmdalex: @layershifter can you show me an example of that? I would give it a shot

docs/src/examples/components/Menu/Variations/MenuExampleVerticalPointingEnd.shorthand.tsx
codepretty: I talked with @levithomason yesterday about this and he also thinks the `primary`, `secondary` prop for Menu should be removed. I like the solution of just using `color` prop for all color style updates.

src/themes/base/colors.ts
codepretty: ah, ok, makes sense!

bmdalex:
closing in favor of #756

src/themes/types.ts
bmdalex: not relevant anymore, see #756

src/themes/teams/components/Menu/menuItemStyles.ts
bmdalex: not relevant anymore, see #756

docs/src/examples/components/Menu/Variations/MenuExampleVerticalPointingEnd.shorthand.tsx
bmdalex: not relevant anymore, see #756
`primary` is not needed as a boolean prop any longer

src/themes/base/colors.ts
bmdalex: not relevant anymore, see #756

Issue #502, opened 2018-11-20
Supporting generating info files for components using the renderComponent function


mnajdova:
## Feature Request
Supporting doc generation for the components created with the createComponent function.

### Problem description
For now, react-docgen-typescript works only with class components. We need to support generating docs for the components created using the renderComponent function.

### Proposed solution
Take a look into the react-docgen-typescript library for adding a custom resolver for the props interfaces.

levithomason:
@kuzhelov Here's the issue that corresponds to the docgen work you're doing.

PR #713, opened 2019-01-14
feat(slot): rename Slot to Box and export

bmdalex:
### Description

This PR:
- renames internal `Slot` component to `Box`
- export `Box` (former `Slot`)


src/components/Segment/Segment.tsx
```diff
@@ -43,7 +43,7 @@ class Segment extends UIComponent<ReactProps<SegmentProps>, any> {
 
     return (
       <ElementType {...unhandledProps} className={classes.root}>
-        {childrenExist(children) ? children : Slot.create(content)}
+        {childrenExist(children) ? children : Component.create(content)}
```
kuzhelov: another misleading point, as result of `Component.create()` is actually an element, not a component

src/components/Component/Component.tsx
```diff
@@ -10,20 +10,20 @@ import {
 import createComponent, { CreateComponentReturnType } from '../../lib/createComponent'
 import { ReactProps } from '../../../types/utils'
 
-export interface SlotProps
-  extends UIComponentProps<SlotProps>,
+export interface ComponentProps
+  extends UIComponentProps<ComponentProps>,
```
kuzhelov: unfortunately, with the changes introduced this type becomes to be very hard to read

kuzhelov (approved):
approving the PR as this rename has been agreed upon - but my personal feeling is that this move introduces even more misleading moments (have pointed a few in the PR) than we have before.

Also, this generic `Component` name doesn't provide any context on its semantics and where it should be used, in contrast to the former `Slot` name.

It would be great if we would wait for others' opinions before merging.

src/components/Chat/ChatItem.tsx
```diff
@@ -63,12 +63,13 @@ class ChatItem extends UIComponent<ReactProps<ChatItemProps>, any> {
 
   private renderChatItem(styles: ComponentSlotStylesPrepared) {
     const { message, gutter, gutterPosition } = this.props
-    const gutterElement = gutter && Slot.create(gutter, { defaultProps: { styles: styles.gutter } })
+    const gutterElement =
```
kuzhelov: this is misleading - `Component.create` reads like a factory method for some `React.Component`, but the element is produced in fact.

src/components/Chat/ChatItem.tsx
```diff
@@ -63,12 +63,13 @@ class ChatItem extends UIComponent<ReactProps<ChatItemProps>, any> {
...
+      gutter && Component.create(gutter, { defaultProps: { styles: styles.gutter } })
```
levithomason: No need to short circuit shorthand factories, they will gracefully handle null/undefined shorthand values.

```suggestion
Component.create(gutter, { defaultProps: { styles: styles.gutter } })
```

src/components/Chat/ChatItem.tsx
levithomason: Although this produces an element, we wanted to avoid `*.createElement()` since it is quite different from the `React.createElement()`.  For all other components, I think the reading makes more sense:

```jsx
Button.create()
Image.create()
...
```

It seems the `Component.create()` is a little more ambiguous since `Component` is so abstract.  It sort of reads like `createComponent()`.

Let's leave it as is for now and see if we get more feedback.

src/components/Component/Component.tsx
levithomason: Yes, Component vs UIComponent is not so clear.  How about if UIComponent becomes BaseComponent?

Another option would be to call `Component` `Box` instead... ?

src/components/Segment/Segment.tsx
levithomason: Per above comments, what if this were `Box.create()`?

test/specs/components/Component/Component-test.ts
```diff
@@ -0,0 +1,33 @@
+import { mountWithProvider as mount } from 'test/utils'
+
+import Component from 'src/components/Component/Component'
+import { isConformant } from 'test/specs/commonTests'
+
+describe('Component', () => {
+  const createComponentComp = (factoryFn: Function, val, options?) =>
+    mount(factoryFn(val, options)).find(Component)
+
+  xdescribe('is conformant', () => {
```
levithomason: Should this be `xdescribe`?

test/specs/components/Component/Component-test.ts
```diff
@@ -0,0 +1,33 @@
...
+    isConformant(Component)
+  })
+
+  describe(`create`, () => {
+    it(`renders a ${Component.defaultProps.as} element with children prop`, () => {
```
levithomason: There is a common test called `rendersChildren()` that I think we can use here.

test/specs/components/Component/Component-test.ts
```diff
@@ -0,0 +1,33 @@
...
+      const testChildren = 'test children'
+      const component = createComponentComp(Component.create, testChildren)
+
+      expect(component.prop('as')).toEqual(Component.defaultProps.as)
+      expect(component.prop('children')).toEqual(testChildren)
+    })
+
+    it(`renders an HTML element with correct prop`, () => {
```
levithomason: I believe the `as` prop is covered in the conformance test.

src/components/Chat/ChatItem.tsx
bmdalex: unfortunately, there was a bug there that did not short circuit the creation of elements; I must have forgot to create an issue for it; I'll check again

test/specs/components/Component/Component-test.ts
bmdalex: yes, we still cannot reenable the tests due to #502 as info files are not created for Slot so conformance tests fail

test/specs/components/Component/Component-test.ts
bmdalex: not in the scope of this PR as it's just renaming files; I'll try to port the `rendersChildren` tests from Semantic UI, though

src/components/Chat/ChatItem.tsx
bmdalex: what about using `Box` as suggested previously by @mnajdova and @layershifter 
@layershifter mentioned `Box` is used successfully in other UI libraries as well:

#### Component:
https://ui.reach.tech/component-component

#### Box:
https://material-ui.com/utils/box/
https://reakit.io/components/box
https://github.com/segmentio/ui-box

I think  renaming `Slot` to `Box` (instead of `Component`) this would remove confusion. As we remember, the semantical problem `Slot` has is that it's describing an entity that is part of something, but we're using `Slot` for wrapping components as well (opposite meaning almost).

What do you think @kuzhelov ?

test/specs/components/Component/Component-test.ts
bmdalex: will remove, thanks

src/components/Component/Component.tsx
bmdalex: agreed, see https://github.com/stardust-ui/react/pull/713/#discussion_r247552617

src/components/Segment/Segment.tsx
bmdalex: agreed, see https://github.com/stardust-ui/react/pull/713/#discussion_r247552617

src/components/Chat/ChatItem.tsx
kuzhelov: @Bugaa92, `Box` is a much better alternative, would totally support this move

src/components/Component/Component.tsx
kuzhelov: @levithomason, would rather choose second option (`Component` -> `Box`), as this provide more specific semantics to this component (instead of `Component` name that, essentially, doesn't  provide any). Thanks!

src/components/Chat/ChatItem.tsx
bmdalex: renamed to `Box`

src/components/Segment/Segment.tsx
bmdalex: renamed to `Box`

bmdalex:
Postponing until solving a dependency

PR #718, opened 2019-01-14
fix(core): replace lodash/fp with lodash

### LINKED ISSUE #703: jest --detectLeaks reports memory leaks
### Steps
```
yarn jest test --detectLeaks
```

### Actual Result
Memory leak reported 💣
image

### Version
0.16.1

miroslavstastny:
This PR is a workaround which fixes #703.
Instead of using `lodash/fp`, stardust now uses `lodash`.
This PR also runs `jest` tests with `--detectLeaks` as part of CI.


src/lib/lodashFp.ts
```diff
@@ -0,0 +1,28 @@
...
+  max: require('lodash/fp/max'),
+  memoize: require('lodash/fp/memoize'),
+  min: require('lodash/fp/min'),
+  pick: require('lodash/fp/pick'),
+  sortBy: require('lodash/fp/sortBy'),
+  startCase: require('lodash/fp/startCase'),
+  sum: require('lodash/fp/sum'),
+  take: require('lodash/fp/take'),
+  trim: require('lodash/fp/trim'),
+}
```
layershifter: We should use only ES imports as we have ES modules build

package.json
```diff
@@ -164,6 +164,7 @@
     "tslint": "^5.11.0",
     "tslint-config-airbnb": "^5.11.1",
     "typescript": "~3.2.2",
+    "weak": "^1.0.1",
```
miroslavstastny: required by `jest --detectLeaks`

src/lib/lodashFp.ts
miroslavstastny: removed

layershifter (approved):

src/lib/customPropTypes.tsx
```diff
@@ -39,27 +37,27 @@ export const suggest = (suggestions: string[]) => {
   const findBestSuggestions = _.memoize((str: string) => {
     const propValueWords = str.split(' ')
 
-    return _.flow(
-      _.map((suggestion: string) => {
+    return _.chain(suggestions)
```
levithomason, line 40: Heads up that `_.chain` causes _all_  (or almost all) of lodash to be imported because all the chainable functions must be available on the returned object, and lodash can't statically detect which will be used (i.e. tree shaking will not work). Recommend refactoring to reduce the weight of our library.

https://medium.com/making-internets/why-using-chain-is-a-mistake-9bc1f80d51ba

PR #721, opened 2019-01-14
feat(MenuItem|AccordionTitle): add ability for removing and customizing the (active|submenu) indicator

### LINKED ISSUE #687: Ability to remove arrow from submenu
### Problem description
When I use menu with 'toolbarButtonBehavior'
and the button has submenu, 
then 'arrow key' next to button is displayed. 

I would like to remove arrow key. 

image

### MVP

mnajdova:
This PR aims to fix https://github.com/stardust-ui/react/issues/687 (allowing the user to remove the submenu indicator), as well is is adding props for customizing the indicator, by providing an icon shorthand. I have few proposal for the API, and would like to hear your opinion on them.

1. Add explicit props for all things
- hideSubmenuIndicator: boolean - indicartes whether the submenu indicators should be shown
- submenuIndicatorVertical: ShorthandValue - shorthand for icon for the submenu indicators in the vertical menus
- submenuIndicatorHorizontal: ShorthandValue - shorthand for icon for the submenu indicators in the horizontal menus

Pros: this API provides you to define all indicators in the root menu, and all these will be propagated to all submenus (vertical and horizontal)
Cons: too many properties added for just one slot

2. Add single property
- submenuIndicator: boolean | ShorthandValue - this is a single prop that defined whether the submenu indicator shuold be shown (true or false), but also allows the user to specify shorthand for the icon in it. 

Props: we are adding just one prop for the submenu indicator, so the user doesn't have to think which one he should apply.
Cons: with this approach, we cannot specify on the root menu what should the submenu indicators in all menus be, as the vertical and horizontal menus are most likely to be different icons. This means, every menu shuold define the submenu indicators in it's first children MenuItems.

**3. Add single property and transform the icon** - this is currently implemented
- submenuIndicator: boolean | ShorthandValue - this is a single prop that defined whether the submenu indicator shuold be shown (true or false), but also allows the user to specify shorthand for the icon in it. For the vertical menus, we can transform this icon 90 degrees.

Props: we are adding just one prop for the submenu indicator, so the user doesn't have to think which one he should apply. Allows the user to specify the submenu indicator on the root menu
Cons: not sure whether this would be expected behavior for the client.

TODO:
- [x] update changelog
- [x] add Indicator examples


src/components/Menu/MenuItem.tsx
```diff
@@ -193,6 +220,10 @@ class MenuItem extends AutoControlledComponent<ReactProps<MenuItemProps>, MenuIt
               defaultProps: { xSpacing: !!content ? 'after' : 'none' },
             })}
           {content}
+          {showSubmenuIndicatorIcon &&
```
layershifter: Do we need `hideSubmenuIndicator` at all?

```jsx
<Menu submenuIndicatorVertical={null} submenuIndicatorVertical={null} />
```

sophieH29:
Thought that came into my mind, users might want to have different icons for open/close state. Do we consider to support it too?

image


Anyway, I would vote for option 2

src/components/UnicodeCharacter/UnicodeCharacter.tsx
```diff
@@ -0,0 +1,63 @@
...
+    hex: PropTypes.string,
+  }
+
+  static defaultProps = {
+    as: 'span',
+  }
+
+  renderComponent({ ElementType, classes, unhandledProps, rtl }) {
+    const { hex } = this.props
+    const hexUnicode = rtl ? rtlMapUnicodeCharacters[hex] || hex : hex
```
mnajdova: We may want to add here disableRtl prop as well...

src/components/Menu/MenuItem.tsx
mnajdova: Nope, removed.

mnajdova:
@sophieH29 thanks for your thoughts. This is currently possible, by adding styles for the MenuItem, not sure if we want to blow up the API, by adding additional things at this moment. If this comes as a requirement in the future, sure we can extend the API. However at this moment, I would restrain from adding more props. Do you have strong opinion that we want to support this at this moment?

src/components/Menu/MenuItem.tsx
```diff
@@ -172,7 +178,21 @@ class MenuItem extends AutoControlledComponent<ReactProps<MenuItemProps>, MenuIt
...
+      icon,
+      wrapper,
+      menu,
+      primary,
+      secondary,
+      active,
+      vertical,
+      submenuIndicator,
+    } = this.props
+    const submenuIndicatorWithDefault = submenuIndicator === undefined ? true : submenuIndicator
```
mnajdova: We cannot specify the default value of the submenuIndicator in the defaultProp, as it is itemShorthand and conflicts with the children. If anybody has other idea, please let me know.

src/components/Accordion/AccordionTitle.tsx
```diff
@@ -47,26 +54,44 @@ class AccordionTitle extends UIComponent<ReactProps<AccordionTitleProps>, any> {
     active: PropTypes.bool,
     index: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
     onClick: PropTypes.func,
+    activeIndicator: customPropTypes.itemShorthand,
+  }
+
+  static defaultProps = {
+    activeIndicator: true,
```
layershifter: My vote is for `indicator`, I'm not sure that we really need there a prefix

sophieH29:
> @sophieH29 thanks for your thoughts. This is currently possible, by adding styles for the MenuItem, not sure if we want to blow up the API, by adding additional things at this moment. If this comes as a requirement in the future, sure we can extend the API. However at this moment, I would restrain from adding more props. Do you have strong opinion that we want to support this at this moment?

Don't have a strong opinion that we need to support it right now, rather out of curiosity and something that we can keep in mind.

src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -120,6 +121,8 @@ const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
 
       ...((bordered || v.borderColor || circular) &&
         getBorderedStyles(circular, v.borderColor || getIconColor(color, v))),
+
+      transform: `${rtl ? 'scaleX(-1)' : ''} rotate(${rtl ? -1 * rotate : rotate}deg)`,
```
codepretty: I love this this addition to Icon!! This is a personal preference, but I find it much easier to quickly understand what the transform is doing here if the code was written in multiple lines. eg)

```
...(rtl && {
    transform: `'scaleX(-1)' rotate(-1 * ${rotate}deg)`,
}),

...(!rtl && {
}),
```

codepretty:
My vote is for the 3rd option, it seems the most minimal to get the job done.

src/themes/base/components/Indicator/indicatorStyles.ts
```diff
@@ -0,0 +1,15 @@
+import * as _ from 'lodash'
+import { ComponentStyleFunctionParam, ICSSInJSStyle } from '../../../types'
+import { IndicatorVariables } from './indicatorVariables'
+import { IndicatorProps } from '@stardust-ui/react'
+
+export default {
+  root: ({
+    props: { color },
```
codepretty: Do we want an indicator to be able to specify it's own color? Since the indicator isn't a standalone component, it makes sense that it would just inherit its color from the current color. Where the current color would depend on the color set by the containing components (ie Menu, Accordion in these cases)

src/components/Indicator/Indicator.tsx
```diff
@@ -0,0 +1,87 @@
...
+      direction && Indicator.directionMap[this.getDirectionBasedOnRtl(rtl, direction)].unicode
+    const contentProps = !icon
+      ? {
+          dangerouslySetInnerHTML: {
+            __html: hexUnicode && this.isHex(hexUnicode) ? `&#x${hexUnicode};` : '',
+          },
+        }
+      : {
+          children: Icon.create(icon, {
+            defaultProps: { color, rotate: Indicator.directionMap[direction].rotation },
```
codepretty: Seems like this is making the assumption that all icons will always be rotated in the same way

mnajdova: I myself had doubts on this, but I don't see any other way how we can easily achieve this (from the other components it's not good practice to change some props for the icon as it is shorthand for another component - Indicator in this case). So we can either provide prop in the Indicator for rotating (similar that we have in the Icon), or we can have assumption that the indicator icon provided will always be the bottom one (and depending on the direction we can rotate it accordingly). I am open to other proposals as well.

src/themes/base/components/Indicator/indicatorStyles.ts
mnajdova: Agree, will remove the color prop from this one. Thanks!

src/themes/teams/components/Icon/iconStyles.ts
mnajdova: Agreed, will refactor it.

src/components/Indicator/Indicator.tsx
```diff
@@ -0,0 +1,87 @@
...
+  commonPropTypes,
+  ColorComponentProps,
+  customPropTypes,
+} from '../../lib'
+import { ReactProps, ShorthandValue } from '../../../types/utils'
+import Icon from '../Icon/Icon'
+
+export interface IndicatorProps extends UIComponentProps, ColorComponentProps {
+  /** The indicator can be direction in different directions. */
+  direction?: 'forward' | 'back' | 'top' | 'bottom'
```
sophieH29: For me, it seems a bit confusable, what is 'bottom' and what is 'back'. I guess it's left but seems not very intuitive.
Maybe just 'right', 'left', 'top', 'bottom' - simple and straightforward

src/components/Indicator/Indicator.tsx
sophieH29: Don't you need to get `direction` here based on rtl, same way as you did above `this.getDirectionBasedOnRtl(rtl, direction)`?

src/components/Accordion/AccordionTitle.tsx
mnajdova: I've added the prefix, so that people will know that the icon they are defining is regarding the active state, but yeah, we can make the name simpler.

src/components/Indicator/Indicator.tsx
```diff
@@ -0,0 +1,87 @@
...
+  UIComponent,
+  UIComponentProps,
+  commonPropTypes,
+  ColorComponentProps,
+  customPropTypes,
+} from '../../lib'
+import { ReactProps, ShorthandValue } from '../../../types/utils'
+import Icon from '../Icon/Icon'
+
+export interface IndicatorProps extends UIComponentProps, ColorComponentProps {
```
sophieH29: Having concerns regarding the name `Indicator` as it can be anything, not specific to mark the element that can expand. Honestly, hard to find any other name, but rather opening a discussion for others ideas.
The only things that came for now `ExpandIndicator`, `HierarchyIndicator`

src/components/Indicator/Indicator.tsx
```diff
@@ -0,0 +1,87 @@
...
+  direction?: 'forward' | 'back' | 'top' | 'bottom'
+
+  /** The indicator can show specific icon if provided. */
+  icon?: ShorthandValue
+}
+
+/**
+ * An indicator is suggest additional content next to the element it is inside.
+ */
+class Indicator extends UIComponent<ReactProps<IndicatorProps>, any> {
```
sophieH29: Having concerns regarding the name `Indicator` as it can be anything, not specific to mark the element that can expand. Honestly, hard to find any other name, but rather opening a discussion for others ideas.
The only things that came for now `ExpandIndicator`, `HierarchyIndicator`

src/components/Indicator/Indicator.tsx
layershifter: `right` and `left` are not RTL safe.

src/components/Indicator/Indicator.tsx
mnajdova: Nope, as the Icon itself is handling rtl, if the theme wants to support that (this changes were added as part of this PR). Let me know if you think otherwise..

src/components/Indicator/Indicator.tsx
mnajdova: Would like to avoid the right and left, as in rtl it won't make sense to use them.. I am open to other suggestion. This one were gathered from the browser arrows for navigation (back and forward). Maybe start and end? What do you think?

src/components/Accordion/AccordionTitle.tsx
mnajdova: After some thoughts, agreed on the name indicator. We are nowhere using prefixes for the shorthands. Thanks for the suggestion!

src/components/Indicator/Indicator.tsx
sophieH29: If Icon is handling it then it's totally fine

src/components/Indicator/Indicator.tsx
sophieH29: Well, we use 'right' and 'left' for the Popup too.
I don't think `start`  `end` is better as it more describes a position in regards to something. If no other ideas then let's leave 'back' and 'forward'. If treat it as video player buttons it makes more sense 😄

alinais:
@sophieH29, @mnajdova  I don't think it is in the scope of this fix to allow other icon variants for the submenu. Let's create a separate feature and address it separately.

src/components/Indicator/Indicator.tsx
miroslavstastny: I vote for `start` and `end` as these are the direction names we use.

src/components/Indicator/Indicator.tsx
miroslavstastny: I would prefer something like this:
```jsx
overrideProps: ({rotate}) => ({rotate: (Indicator.directionMap[direction].rotation || 0) + (rotate || 0)})
```

Then I can use `send` icon as an indicator:
```jsx
<Indicator icon={{name: "send", rotate: 90}} direction="forward" />
```

mnajdova:
> @sophieH29, @mnajdova I don't think it is in the scope of this fix to allow other icon variants for the submenu. Let's create a separate feature and address it separately.

@alinais I agree is not the scope of this fix, but I would rather create another issue and address it as part of this PR, as otherwise we will add just one prop for removing the indicator and then we will introduce breaking changes for refactoring it to this implementation..

src/components/Indicator/Indicator.tsx
mnajdova: @miroslavstastny agree, let's strive for consistency, although it seems really strange for me to have the '**<**' indicator a start, and '**>**' as end.. We can change it later anyway...

docs/src/examples/components/Indicator/Types/IndicatorExampleDirection.shorthand.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react'
+import { Indicator } from '@stardust-ui/react'
+
+const IndicatorExampleDirection = () => (
+  <div>
```
layershifter: May be `Fragment` API?

```suggestion
  <>
```

docs/src/examples/components/Indicator/index.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
+import Types from './Types'
+
+const IndicatorExamples = () => (
+  <div>
```
layershifter:
```suggestion
  <>
```
We can use there Fragment API, `div` is a legacy from SUIR

src/components/Icon/Icon.tsx
```diff
@@ -37,6 +37,9 @@ export interface IconProps extends UIComponentProps, ColorComponentProps {
   /** Name of the icon. */
   name?: string
 
+  /** An icon can be rotated by the degree specified as number. */
+  rotate?: number
```
layershifter, line 41: String looks like a valid value, too:

```jsx
<Indicator rotate='90' />
```

src/components/Indicator/Indicator.tsx
```diff
@@ -0,0 +1,89 @@
...
+  renderComponent({ ElementType, classes, unhandledProps, rtl }) {
+    const { direction, icon, color } = this.props
+    const hexUnicode =
+      direction && Indicator.directionMap[this.getDirectionBasedOnRtl(rtl, direction)].unicode
+    const contentProps = !icon
+      ? {
+          dangerouslySetInnerHTML: {
+            __html: hexUnicode && this.isHex(hexUnicode) ? `&#x${hexUnicode};` : '',
+          },
+        }
```
layershifter: This looks very strange for me, I played with this an it seems that we don't need this at all: https://codesandbox.io/s/1zxj6x8ok3

I've pinned `react@16.3` and `react-dom@16.3` to ensure that it's supported and even tested in IE 😼 

---

image

IE compatible build: https://create-react-app-i4qwctbvw.now.sh/

src/components/Icon/Icon.tsx
mnajdova: The rotate property of the icon can be used for calculation as well, take a look in the Indicator component (the overrideProps function for the Icon). In that case we will need to cast this to number if it was provided as string. Do we really want this?

src/components/Indicator/Indicator.tsx
mnajdova: Yep, refactored. At the start I had here string, that's why I had to do this magic, totally valid and more clear code. Thanks!

layershifter (approved):

PR #726, opened 2019-01-15
fix(Dropdown): Highlights selected option in single selection

silviuaavram:
Highlights selected option when opening dropdown.

Based on https://github.com/paypal/downshift/pull/652 and since now it's released, this PR will be able to perform the necessary fix.

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -523,14 +523,6 @@ export default class Dropdown extends AutoControlledComponent<
         if (state.isOpen && document.activeElement === this.listRef.current) {
           return {} // won't change state in this case.
         }
-      case Downshift.stateChangeTypes.clickItem:
-      case Downshift.stateChangeTypes.keyDownEnter:
-        if (!this.props.search && !this.props.multiple) {
-          this.setState({
-            defaultHighlightedIndex: state.highlightedIndex,
-          })
-        }
-        return changes
       default:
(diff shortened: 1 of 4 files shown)


layershifter (commented):
Downshift has [highlightedIndex](https://github.com/paypal/downshift#highlightedindex), too. Can we implement it as autocontrolled prop?

silviuaavram:
@layershifter let's create a new issue with making those props autocontrolled. It will be a bit more complicated, as our `default` is actually `initial` in Downshift, while `default` there is the value that prop will be used at every change (selection, close/open).

src/components/Dropdown/Dropdown.tsx
```diff
@@ -228,6 +229,7 @@ export default class Dropdown extends AutoControlledComponent<
       focused: false,
       searchQuery: search ? '' : undefined,
       value: multiple ? [] : null,
+      defaultHighlightedIndex: !this.props.search && !this.props.multiple ? null : undefined,
```
mnajdova, line 233: Can we document with some comment why we use null and why undefined?

bmdalex (approved):

jurokapsiar (approved):

src/components/Dropdown/Dropdown.tsx
silviuaavram: added one. is it ok?

mnajdova (approved):

PR #729, opened 2019-01-15
feat(Dropdown): add `loading` prop

layershifter:
# BREAKING CHANGES

### 1. Rename `toggleButton` to `toggleIndicator`

Reason: this slot is not a button, it cannot be focused or be selected in native `select`, too.

#### Before
```jsx
<Dropdown toggleButton />
```

### After
```jsx
<Dropdown toggleIndicator />
```

### 2. `toggleIndicator` is visible by default

Reason: it's a common UI behavior to show toggles by default.

#### Before
```jsx
<Dropdown />
<Dropdown toggleButton />
```

### After
```jsx
<Dropdown />
<Dropdown toggleIndicator={false} />
```

---

### Customizing `loadingMessage`

image

```jsx
<Dropdown loadingMessage={{ content: <Loader size="large" /> }} />
```

image

```jsx
<Dropdown />
```


src/themes/teams/components/Dropdown/dropdownMessageNoResultsStyles.ts
```diff
@@ -0,0 +1,13 @@
...
+import { DropdownMessageNoResultsProps } from '../../../../components/Dropdown/DropdownMessageNoResults'
+import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
+
+const dropdownSelectedItemStyles: ComponentSlotStylesInput<DropdownMessageNoResultsProps, any> = {
+  root: ({ variables: v }): ICSSInJSStyle => ({
+    [`&.${ListItem.className}`]: { backgroundColor: v.listItemBackgroundColor },
+
+    fontWeight: 'bold',
+  }),
+}
```
layershifter: We have a task for cleaning up styles in `Dropdown`, we should separate variables between components and them meaningful.

CHANGELOG.md
```diff
@@ -19,6 +19,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### BREAKING
 - Rename `DropdownLabel` to `DropdownSelectedItem` and extract styles @layershifter ([#725](https://github.com/stardust-ui/react/pull/725))
+- Changes in `Dropdown` @layershifter ([#729](https://github.com/stardust-ui/react/pull/729))
+  - rename `noResultsMessage` prop to `messageNoResults`
+  - `messageNoResults` has default value by default 
+  - rename `toggleButton` prop to `indicator`
```
kuzhelov: could you, please, mention the libraries you've compared against (other than SUIR) - those that server as a reason for this change?

docs/src/examples/components/Dropdown/Variations/DropdownExampleMultipleSearchImageAndContent.shorthand.tsx
```diff
@@ -55,7 +55,6 @@ const DropdownExample = () => (
     getA11yStatusMessage={getA11yStatusMessage}
     search
     getA11ySelectionMessage={getA11ySelectionMessage}
-    noResultsMessage="We couldn't find any matches."
```
kuzhelov: not completely sure that there should be any default value, given that this value needs to be properly localized. The reasoning behind this thought is the following: suppose that I, as a client, am using this component. I am not expecting any English text to show up in my French UI. 

Then if I wouldn't carefully study the API and spot all the places where default values may be provided, I will be trapped by this "alien" locale chiming into my UI. In that case I would rather prefer explicitness in a sense that there won't be any string that I am unaware of being shown to the client.

CHANGELOG.md
layershifter: SUIR calls this prop `icon`, but in the current implementation it's an unicode character.

---

https://react-select.com/props
And we have now the `Indicator` component.

From other side it's that `toggleButton` can behave like a loader.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -155,17 +159,19 @@ export default class Dropdown extends AutoControlledComponent<
     fluid: PropTypes.bool,
     getA11ySelectionMessage: PropTypes.object,
     getA11yStatusMessage: PropTypes.func,
+    indicator: PropTypes.bool,
```
kuzhelov: maybe personal thing, but for me this name makes it being much less intuitive what it is responsible for

docs/src/prototypes/AsyncDropdownSearch/AsyncDropdownSearch.tsx
```diff
@@ -0,0 +1,98 @@
...
+      <div style={{ margin: 20 }}>
+        <Segment>
+          <Header content="Async Dropdown Search" />
+          <p>Use the field to perform a simulated search.</p>
+        </Segment>
+
+        <Segment>
+          <Dropdown
+            fluid
+            indicator={false}
```
kuzhelov: maybe personal thing, but for me this name makes it being much less intuitive what it is responsible for, especially in this context

src/components/Dropdown/Dropdown.tsx
```diff
@@ -370,7 +383,10 @@ export default class Dropdown extends AutoControlledComponent<
     getItemProps: (options: GetItemPropsOptions<ShorthandValue>) => any,
     getInputProps: (options?: GetInputPropsOptions) => any,
   ) {
-    const accessibilityMenuProps = getMenuProps({ refKey: 'innerRef' }, { suppressRefError: true })
+    const { innerRef, ...accessibilityMenuProps } = getMenuProps(
```
kuzhelov, line 398: 👍

src/components/Dropdown/Dropdown.tsx
```diff
@@ -396,49 +412,42 @@ export default class Dropdown extends AutoControlledComponent<
...
+          ...(typeof item === 'object' &&
+            !item.hasOwnProperty('key') && {
+              key: (item as any).header,
+            }),
+        },
+        overrideProps: () => this.handleItemOverrides(item, index, getItemProps),
+      }),
+    )
 
-    if (filteredItems.length > 0) {
```
kuzhelov, line 444: 👍

src/components/Dropdown/DropdownIndicator.tsx
```diff
@@ -0,0 +1,61 @@
...
+    open: PropTypes.bool,
+  }
+
+  handleClick = (e: React.SyntheticEvent) => {
+    _.invoke(this.props, 'onClick', e, this.props)
+  }
+
+  renderComponent({ classes, ElementType, unhandledProps }) {
+    const { loading, open } = this.props
+    const content = open ? String.fromCharCode(9650) : String.fromCharCode(9660)
```
kuzhelov: nit: comments might help quicker understand what these values stand for (it would be a valid response that, most probably, there is a browser engine somewhere close that is ready to provide this answer to reviewer, but still:)

docs/src/prototypes/AsyncDropdownSearch/AsyncDropdownSearch.tsx
kuzhelov: actually, it still seems to me that we might want to split `toggleButton` and `loadingIndicator` on two dedicated slots. In that case it will be possible to the client to define each one's tree (by providing shorthand values to corresponding slots), and, at the same time, it will be possible to achieve the following configs:

- customized 'delayed' loading indicator 
- customized toggle button
- disable each separately from the other

kuzhelov (commented):
overall I support the vector of changes, but would like to discuss couple of moments (where had left comments) before merge

src/components/Dropdown/DropdownIndicator.tsx
layershifter: It was copied from existing code, I will reuse there the `Indicator` component as we have merged it 🦅

src/components/Dropdown/Dropdown.tsx
```diff
@@ -689,7 +696,12 @@ export default class Dropdown extends AutoControlledComponent<
     }
 
     // we don't have event for it, but want to keep the event handling interface, event is empty.
-    _.invoke(this.props, 'onSelectedChange', {}, { ...this.props, value: newValue })
+    _.invoke(
+      this.props,
+      'onSelectedChange',
+      {},
+      { ...this.props, searchQuery: '', value: newValue },
```
layershifter, line 723: Previously we made a reset in `setState()` of `searchQuery`, but this change was not passed in `onSelectedChange`

src/components/Dropdown/Dropdown.tsx
```diff
@@ -287,7 +302,16 @@ export default class Dropdown extends AutoControlledComponent<
                         variables,
                       )
                     : this.renderTriggerButton(styles, getToggleButtonProps)}
-                  {toggleButton && this.renderToggleButton(getToggleButtonProps, classes, isOpen)}
+                  {/* TODO: Make `indicator` a fully working shorthand. */}
+                  {DropdownIndicator.create(indicator && '', {
```
mnajdova: I don't understand what is the connection between the loading and indicator aspects of the Dropdown... We can replace the toggleButton with indicator (totally support this idea, as it would be consistent with the other components), and the indicator definitely can be shorthand for the Indicator component. The loading is completely separated concept that again can be implemented with a shorthand for the Loading component, and the Dropdown can handle when it should show it..

src/components/Dropdown/DropdownIndicator.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  renderComponent({ classes, ElementType, unhandledProps }) {
+    const { loading, open } = this.props
+
+    return (
+      <ElementType className={classes.root} onClick={this.handleClick}>
+        {loading ? (
+          <Loader size="small" />
+        ) : (
+          Indicator.create('', {
```
mnajdova: See my prev comment, let's separate this two aspects...

src/components/Dropdown/Dropdown.tsx
layershifter: Resolved with latest changes 👍

docs/src/prototypes/AsyncDropdownSearch/AsyncDropdownSearch.tsx
layershifter: Resolved with latest changes, `loadingIndicator` will be a separate concept

docs/src/examples/components/Dropdown/Variations/DropdownExampleMultipleSearchImageAndContent.shorthand.tsx
layershifter: Reverted this change, we will not provide a default message

CHANGELOG.md
layershifter: We decided to go with `toggleIndicator` 🚢

src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -73,22 +73,18 @@ const dropdownStyles: ComponentSlotStylesInput<DropdownProps, DropdownVariables>
...
+    width: v.toggleButtonSize,
     cursor: 'pointer',
     backgroundColor: 'transparent',
     margin: 0,
     display: 'flex',
     justifyContent: 'center',
     alignItems: 'center',
     userSelect: 'none',
-    ...(fluid ? { right: 0 } : { left: `calc(${width} - ${toggleButtonSize})` }),
+    ...(p.fluid ? { right: 0 } : { left: `calc(${v.width} - ${v.toggleButtonSize})` }),
```
mnajdova: Let's rename the variable toggleButtonSize -> toggleIndicatorSize (should we consider whether this will be icon and change the value?!?!)

src/components/Dropdown/DropdownMessageNoResults.tsx
```diff
@@ -0,0 +1,36 @@
...
+class DropdownMessageNoResults extends UIComponent<ReactProps<DropdownMessageNoResultsProps>> {
+  static className = 'ui-dropdown__message-no-results'
+  static create: Function
+  static displayName = 'DropdownMessageNoResults'
+  static propTypes = commonPropTypes.createCommon({ children: false })
+
+  renderComponent({ classes, ElementType, unhandledProps }) {
+    const { content } = this.props
+
+    return <ListItem className={classes.root} content={content} {...unhandledProps} />
```
mnajdova: Not sure about this new components. Can't we use the Box component with `as={ListItem}`? We can still create slots in the styles for them... Thoughts?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -287,7 +294,13 @@ export default class Dropdown extends AutoControlledComponent<
                         variables,
                       )
                     : this.renderTriggerButton(styles, getToggleButtonProps)}
-                  {toggleButton && this.renderToggleButton(getToggleButtonProps, classes, isOpen)}
+                  {Indicator.create(toggleIndicator, {
+                    defaultProps: {
+                      direction: isOpen ? 'top' : 'bottom',
+                      onClick: getToggleButtonProps().onClick,
```
mnajdova, line 300: In the Indicator we are not handling the `onClick` prop, so we won't have the user props as a second arg. Will that be a problem in this case?

layershifter: Nope, it comes from Downshift

src/themes/teams/components/Dropdown/dropdownStyles.ts
layershifter: Renamed to `toggleIndicatorSize`

docs/src/components/CodeSnippet.tsx
```diff
@@ -6,24 +6,24 @@ import Editor, { EDITOR_BACKGROUND_COLOR } from './Editor'
...
+  mode?: 'json' | 'jsx' | 'html' | 'sh'
+  value: string | string[] | Object
   style?: React.CSSProperties
 }
 
-const joinToString = (stringOrArray: string | string[]) => {
-  return typeof stringOrArray === 'string' ? stringOrArray : stringOrArray.join('\n')
-}
+const joinToString = (stringOrArray: string | string[]) =>
+  Array.isArray(stringOrArray) ? stringOrArray.join('\n') : stringOrArray
```
kuzhelov: could you, please, suggest why this is better? if we are allowing types different from `string` and `string[]`, lets express this express this explicitly as the arg's type

src/components/Dropdown/Dropdown.tsx
```diff
@@ -133,8 +137,8 @@ export interface DropdownProps extends UIComponentProps<DropdownProps, DropdownS
   /** Sets search query value (controlled mode). */
   searchQuery?: string
 
-  /** Whether toggle button (that shows/hides items list) should be rendered. */
-  toggleButton?: boolean
+  /** Whether a toggle indicator (that shows/hides items list) should be rendered. */
```
kuzhelov: seems that we should introduce some corrections to description (as it is not `boolean` anymore). Something like the following, if we would like to stay consistent with the rest of the code:

```
/* Toggle indicator that shows/hides items list. */
```
or more precise:

```
/* Controls appearance of toggle indicator that shows/hides items list. */
```

src/components/Dropdown/Dropdown.tsx
kuzhelov, line 300: for the sake of consistency, lets rename it to `getToggleIndicatorProps `

src/components/Dropdown/Dropdown.tsx
```diff
@@ -421,50 +424,43 @@ export default class Dropdown extends AutoControlledComponent<
...
     return [
-      noResultsMessage
-        ? {
-            key: 'dropdown-no-results',
-            content: <Text weight="bold" content={noResultsMessage} />,
-            styles: styles.emptyListItem,
-          }
-        : null,
+      ...items,
+      loading && DropdownMessageLoading.create(loadingMessage),
```
kuzhelov: this `DropdownMessageLoading ` component seems to be extremely simple in terms of its rendering logic - couldn't we reuse `Box` component here instead of creating new one?

src/themes/teams/components/Dropdown/dropdownMessageLoadingStyles.ts
```diff
@@ -0,0 +1,11 @@
+import ListItem from '../../../../components/List/ListItem'
+import { DropdownMessageLoadingProps } from '../../../../components/Dropdown/DropdownMessageLoading'
+import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
+
+const dropdownMessageLoadingStyles: ComponentSlotStylesInput<DropdownMessageLoadingProps, any> = {
+  root: ({ variables: v }): ICSSInJSStyle => ({
+    [`&.${ListItem.className}`]: { backgroundColor: v.listItemBackgroundColor },
```
kuzhelov: do we want to consider `ListItem` as a child element only? It seems that this styling is generic enough, so that client would like to see it being applied to either custom `<li>` or even `<div>` element passed as a slot value. Should we omit this selector altogether?

docs/src/components/CodeSnippet.tsx
layershifter: I rewrote this place, we accept a plain object there for JSON. Nice catch 👍

src/components/Dropdown/Dropdown.tsx
layershifter: It's not possible, it's a render prop that comes from Downshift 🐰

src/components/Dropdown/DropdownMessageNoResults.tsx
layershifter: I decided to reuse just `ListItem` there 👍

src/components/Dropdown/Dropdown.tsx
```diff
@@ -437,34 +439,39 @@ export default class Dropdown extends AutoControlledComponent<
...
-        ? {
-            key: 'dropdown-no-results',
-            content: <Text weight="bold" content={noResultsMessage} />,
-            styles: styles.emptyListItem,
-          }
-        : null,
+        }),
+      !loading &&
+        items.length === 0 &&
+        ListItem.create(noResultsMessage, {
```
mnajdova, line 472: 👍

src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -73,22 +74,28 @@ const dropdownStyles: ComponentSlotStylesInput<DropdownProps, DropdownVariables>
     background: listBackgroundColor,
   }),
 
-  emptyListItem: ({ variables: { listItemBackgroundColor } }) => ({
-    backgroundColor: listItemBackgroundColor,
+  loadingMessage: ({ variables: v }): ICSSInJSStyle => ({
+    [`&.${ListItem.className}`]: { backgroundColor: v.listItemBackgroundColor },
```
mnajdova: Why do we need this selector here? As far as I can see we are applying the styles on the ListItem component.

src/components/Dropdown/Dropdown.tsx
layershifter: I reused `ListItem` there

src/themes/teams/components/Dropdown/dropdownMessageLoadingStyles.ts
layershifter: Removed this selector at all

src/themes/teams/components/Dropdown/dropdownStyles.ts
layershifter: Removed it, nice catch 👍

mnajdova (approved):

silviuaavram (approved):

PR #731, opened 2019-01-15
docs(Examples): add export to Codesandbox

layershifter:
docs-sandbox

This PR enables export of examples to Codesandbox ⭐️

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
+++ b/docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
@@ -86,7 +86,7 @@ class ComponentControlsShowCode extends React.Component<
         example={exampleCode}
         dependencies={dependencies}
         /* Magic trick to reload sources on passed code update */
-        key={btoa(exampleCode)}
+        key={exampleCode}
         name={exampleName}
         providedFiles={{
           'index.js': { content: appTemplateJs },
--- a/docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
+++ b/docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
@@ -1,19 +1,12 @@
+import * as _ from 'lodash'
 import * as React from 'react'
 import CodeSandboxer from 'react-codesandboxer'
-import { Menu } from 'semantic-ui-react'
 
-import pkg from 'package.json'
+import { Menu } from 'semantic-ui-react'
 import { ComponentSourceManagerLanguage } from 'docs/src/components/ComponentDoc/ComponentSourceManager'
+import { dependencies } from 'docs/src/components/Playground/renderConfig'
 import { updateForKeys } from 'docs/src/hoc'
 import { appTemplateJs } from './indexTemplates'
-
-const dependencies = {
-  '@stardust-ui/react': pkg.version,
-  lodash: pkg.dependencies.lodash,
-  react: pkg.devDependencies.react,
-  'react-dom': pkg.devDependencies['react-dom'],
-  'react-scripts': 'latest',
-}
 
 type ComponentControlsCodeSandboxProps = {
   exampleCode: string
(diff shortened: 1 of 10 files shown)


docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
```diff
@@ -0,0 +1,117 @@
...
+import CodeSandboxer from 'react-codesandboxer'
+import { Menu } from 'semantic-ui-react'
+
+import pkg from 'package.json'
+import { ComponentSourceManagerLanguage } from 'docs/src/components/ComponentDoc/ComponentSourceManager'
+import { updateForKeys } from 'docs/src/hoc'
+import { appTemplateJs } from './indexTemplates'
+
+const dependencies = {
+  '@stardust-ui/react': pkg.version,
```
kuzhelov: unfortunately, it won't work for localhost :(

kuzhelov: would suggest to consume dependency names from the same source that is used by `importsResolver` of `SourceRender` (with necessary additions) - to ensure that they always will be in sync 👍

brianespinosa:
This is @#*&$^ amazing. A million times easier to help answer questions for people by just spinning up a quick Codesandbox from an example. Awesome stuff @layershifter

docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/indexTemplates.ts
```diff
@@ -0,0 +1,15 @@
...
+import Example from "./example";
+
+const element = document.getElementById("root");
+ReactDOM.render(
+  <Provider theme={themes.teams}>
+    <Example />
+  </Provider>,
+  element
+);
+`.slice(1)
```
kuzhelov: may suggest to introduce comment for the sake of intent being immediately readable

docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
layershifter: Yep, this is unfixable because CodeSandbox installs dependencies from NPM 🐤

layershifter: It's impossible because these deps will fetched from NPM on CodeSandbox side 😭

docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/indexTemplates.ts
layershifter: removed this `slice()` at all to avoid confusions

docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    return (
+      <CodeSandboxer
+        afterDeploy={this.handleDeploy}
+        examplePath="/"
+        example={exampleCode}
+        dependencies={dependencies}
+        /* Magic trick to reload sources on passed code update */
+        key={btoa(exampleCode)}
```
kuzhelov: pass `exampleCode` directly?

kuzhelov (approved):

docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
kuzhelov: yes, but speaking of the keys - those could be kept in sync, right? (like `react`, `lodash`, etc)

layershifter: Oh, I understand your point, will do

kuzhelov:
awesome one - amongst other ones really appreciate the introduced benefit of being able to run Stardust components in 'clean' environment (no third-party static CSS, etc) 👍

PR #734, opened 2019-01-16
chore(tests): fix behavior test regex

miroslavstastny:
Fix regexes uses in behavior test definitions.
Changelog intentionally left blank.

### DIFF
--- a/test/specs/behaviors/testDefinitions.ts
+++ b/test/specs/behaviors/testDefinitions.ts
@@ -44,7 +44,7 @@ definitions.push({
 
 // Example:  Adds role 'menuitem' to 'anchor' component's part
 definitions.push({
-  regexp: /Adds role '(\w+)' to '(\w+)' component's part/g,
+  regexp: /Adds role '(\w+)' to '([\w-]+)' component's part/g,
   testMethod: (parameters: TestMethod) => {
     const [roleToBeAdded, elementWhereToBeAdded] = [...parameters.props]
     const property = {}
@@ -56,7 +56,7 @@ definitions.push({
 // Example: Adds attribute 'tabIndex=0' to 'anchor' component's part.
 //          Adds attribute 'data-is-focusable=true' to 'anchor' component's part.
 definitions.push({
-  regexp: /Adds attribute '([\w-]+)=([\w\d]+)' to '([\w\d]+)' component's part\./g,
+  regexp: /Adds attribute '([\w-]+)=([\w\d]+)' to '([\w-]+)' component's part\./g,
   testMethod: (parameters: TestMethod) => {
     const [attributeToBeAdded, attributeExpectedValue, elementWhereToBeAdded] = [
       ...parameters.props,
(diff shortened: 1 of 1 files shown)


sophieH29 (approved):

test/specs/behaviors/testDefinitions.ts
```diff
@@ -56,7 +56,7 @@ definitions.push({
 // Example: Adds attribute 'tabIndex=0' to 'anchor' component's part.
 //          Adds attribute 'data-is-focusable=true' to 'anchor' component's part.
 definitions.push({
-  regexp: /Adds attribute '([a-z A-Z -]+)=([a-z 0-9]+)' to '([a-z]+)' component's part\./g,
+  regexp: /Adds attribute '([\w-]+)=([\w\d]+)' to '([\w\d]+)' component's part\./g,
```
kuzhelov: is it intentional inclusion of digit in the last regex group? Could you, please, suggest which part? In that case we should make it consistent for other similar places as well (however, not sure that we should include digits here, even given it is technically correct)

kuzhelov: also, please, note that dash ('-') is not included to this component part group's regex - seems that we should use `([\w-]+)` for it

test/specs/behaviors/testDefinitions.ts
```diff
@@ -44,7 +44,7 @@ definitions.push({
 
 // Example:  Adds role 'menuitem' to 'anchor' component's part
 definitions.push({
-  regexp: /Adds role '([a-z]+)' to '([a-z]+)' component's part/g,
+  regexp: /Adds role '(\w+)' to '(\w+)' component's part/g,
```
kuzhelov: this `component's part`'s group regex turns out to be inconsistent across different places where used - it seems that `([\w-]+)` is the expression we should use for it everywhere

test/specs/behaviors/testDefinitions.ts
```diff
@@ -256,7 +256,7 @@ definitions.push({
 
 // Example: Adds attribute 'aria-hidden=true', if there is no 'alt' property provided.
 definitions.push({
-  regexp: /Adds attribute '([\w\-\w \s*]+)=([a-z]+)', if there is no 'alt' property provided\.+/g,
```
kuzhelov, line 259: 👍

test/specs/behaviors/testDefinitions.ts
```diff
@@ -277,7 +277,7 @@ definitions.push({
 
 // Example:  Adds attribute 'aria-expanded=true' based on the property 'open' if the component has 'hasSubtree' property.
 definitions.push({
-  regexp: /Adds attribute '([\w\-\w \s*]+)=([a-z]+)' based on the property '([a-z]+)' if the component has '([a-zA-Z]+)' property./g,
+  regexp: /Adds attribute '([\w-]+)=(\w+)' based on the property '(\w+)' if the component has '(\w+)' property\./g,
```
kuzhelov, line 280: 👍

kuzhelov (approved):
just couple comments

test/specs/behaviors/testDefinitions.ts
miroslavstastny: component part changed to `([\w-]+)` everywhere

PR #735, opened 2019-01-16
fix(factories): remove `render` from default options

layershifter:
This PR fixes issue with always present `render()` function, this breaks passing `isNoop` check in `factories.tsx` and disallows to hide/skip element renders.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### BREAKING
 - Rename `DropdownLabel` to `DropdownSelectedItem` and extract styles @layershifter ([#725](https://github.com/stardust-ui/react/pull/725))
 
+### Fixes
+- Fix remove `render` from default factories options @layershifter ([#735](https://github.com/stardust-ui/react/pull/735))
+
 ### Documentation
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### BREAKING
 - Rename `DropdownLabel` to `DropdownSelectedItem` and extract styles @layershifter ([#725](https://github.com/stardust-ui/react/pull/725))
 
+### Fixes
+- Fix remove `render` from default factories options @layershifter ([#735](https://github.com/stardust-ui/react/pull/735))
```
kuzhelov: nit: we could simplify a bit: `Remove 'render' from ... `

src/lib/factories.ts
```diff
@@ -49,7 +48,7 @@ const mappedProps: { [key in HTMLTag]: ShorthandProp } = {
 /** A more robust React.createElement. It can create elements from primitive values. */
 export function createShorthand(
   Component: React.ReactType,
-  mappedProp: string,
+  mappedProp?: string,
```
kuzhelov, line 51: 👍

kuzhelov (approved):

PR #737, opened 2019-01-16
docs(Prototype): Replace Popup with Submenu in Popover prototype

sophieH29:
After Submenu PR was merged, it makes sense to use it instead of Popup in Popover prototype. What means less code and less complexity


docs/src/examples/components/Menu/Usages/MenuExampleToolbar.shorthand.tsx
```diff
@@ -74,6 +74,7 @@ const items = [
     },
     accessibility: toolbarButtonBehavior,
     'aria-label': 'More options',
+    indicator: false,
```
sophieH29, line 77: removed also indicator in toolbar example, as it's not needed there, based on https://github.com/stardust-ui/react/issues/687

layershifter (approved):
Please add changelog entry before merge 😺

sophieH29:
> Please add changelog entry before merge 😺

There is no need for changelog as it's not a releasable unit, just a prototype :)

PR #740, opened 2019-01-17
fix(typings): synthetic event

kuzhelov:
This PR explicitly specifies generic param (as `HTMLElement`) of type that is used for synthetic event arg in component's event handlers. 

This is necessary for client libraries to avoid explicit cast from `Element` (currently provided by Stardust) to `HTMLElement` which is expected by client's code.

## TODO
- [x] add changelog entry

### DIFF
--- a/types/utils.d.ts
+++ b/types/utils.d.ts
@@ -31,7 +31,10 @@ export type ReactChildren = React.ReactNodeArray | React.ReactNode
 export type ReactPropsStrict<T> = { [K in keyof T]: NullableIfUndefined<T[K]> }
 export type ReactProps<T> = Extendable<ReactPropsStrict<T>>
 
-export type ComponentEventHandler<TProps> = (event: React.SyntheticEvent, data: TProps) => void
+export type ComponentEventHandler<TProps> = (
+  event: React.SyntheticEvent<HTMLElement>,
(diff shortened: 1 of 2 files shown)


miroslavstastny (approved):

types/utils.d.ts
```diff
@@ -31,7 +31,10 @@ export type ReactChildren = React.ReactNodeArray | React.ReactNode
 export type ReactPropsStrict<T> = { [K in keyof T]: NullableIfUndefined<T[K]> }
 export type ReactProps<T> = Extendable<ReactPropsStrict<T>>
 
-export type ComponentEventHandler<TProps> = (event: React.SyntheticEvent, data: TProps) => void
+export type ComponentEventHandler<TProps> = (
```
layershifter, line 34:
```suggestion
export type ComponentEventHandler<TProps, TElem = HTMLElement> = (
```

Can we make this a generic?

kuzhelov: yes, although not sure whether it is necessary to introduce this variability now, as it is pretty much clear that it will always be about HTMLElement for this handler's type - thus, not sure if we should make this type to be customizable (as it will, probably, will be an indication of some improper assumptions)

PR #748, opened 2019-01-19
feat(Icon): Adding the missed call icon

codepretty:
Adding the icon to indicate a missed call in control messages.

image


kuzhelov (approved):
please, just add changelog entry before merge, something like "Add and export 'missed call' icon in Teams theme"

alinais (approved):

PR #751, opened 2019-01-21
docs: improve docs for shorthand props

kuzhelov:
### TODO
- [x] update changelog


CHANGELOG.md
```diff
@@ -29,6 +26,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Features
 - Rename `Slot` component to `Box` and export it @Bugaa92 ([#713](https://github.com/stardust-ui/react/pull/713))
 - Add `Indicator` component and used it in `MenuItem` and `AccordionTitle` @mnajdova ([#721](https://github.com/stardust-ui/react/pull/721))
+- Expose `renderItem` and `renderSelectedItem` callbacks API for `Dropdown` @layershifter ([#746](https://github.com/stardust-ui/react/pull/746))
```
layershifter, line 32: 👍

CHANGELOG.md
```diff
@@ -29,6 +26,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
+- Refine Shorthand docs page content @kuzhelov ([#751](https://github.com/stardust-ui/react/pull/751))
```
layershifter, line 35: May be it should be placed in `Documentation` section?

docs/src/views/ShorthandProps.tsx
```diff
@@ -1,37 +1,263 @@
...
+          There is even shorter way to define default element's props - by using a primitive value.
+          In that case provided shorthand value will be taken as a value for 'default prop' of this
+          default element.
+        </p>
+
+        <p>
+          This could be much easier seen with an example. Here, again, we have a {code('Button')}{' '}
+          element with its icon being defined with shorthand - where provided {code('string')} value
+          will be used as icon's {code('name')}:
+        </p>
```
layershifter: Can we add a section about boolean values and `null`? That this allows to remove a slot completely

docs/src/views/ShorthandProps.tsx
```diff
@@ -1,37 +1,263 @@
...
+          It is quite common for Stardust component to have "slots" which accept shorthand values.
+          For example, {code('Button')} component has an {code('icon')} slot which value defines the
+          icon that will be rendered.
+        </p>
+
+        {codeExample([`<Button icon='chess rook' />`])}
+
+        <p>
+          So, what this string value (i.e. {code("'chess rook'")}) means for {code('Button')}{' '}
+          component and, essentially, how it transforms into rendered {code('<Icon />')} element?
```
mnajdova: In my opinion, this question doesn't bring any value to the user. Let's try to remove it and maybe provide bullet points for explaining this points below.

docs/src/views/ShorthandProps.tsx
```diff
@@ -1,37 +1,263 @@
...
+          `// 'name' and 'size' will be used as <Icon /> element's props`,
+          `<Button icon={{ name: 'chess rook', size: 'small' }} />`,
+        ])}
+
+        {/* SHORTHAND AS A PRIMITIVE VALUE */}
+        {header('Shorthand value as Primitive')}
+        <p>
+          There is even shorter way to define default element's props - by using a primitive value.
+          In that case provided shorthand value will be taken as a value for 'default prop' of this
+          default element.
```
mnajdova: 'default prop' of this default element => 'default prop' of this element. Would avoid the default for the element as well...

docs/src/views/ShorthandProps.tsx
```diff
@@ -1,37 +1,263 @@
...
+          `<>`,
+          `  <Button icon='chess rook' />`,
+          ``,
+          `  {/* has identical effect to the previous one */}`,
+          `  <Button icon={{ name: 'chess rook' }} />`,
+          `</>`,
+        ])}
+        <p>
+          This works because {code('name')} is the default prop of slot's {code('<Icon />')}{' '}
+          element.
```
mnajdova: This works because {code('name')} is the default prop of the {code('<Icon />')} component.

docs/src/views/ShorthandProps.tsx
```diff
@@ -1,37 +1,263 @@
...
+
+        <blockquote>
+          <strong>
+            There is a very important caveat here, though: whenever React Element is directly used
+            as a shorthand value, it becomes client's responsibility to handle some aspects that
+            Stardust was originally responsible for (such as {code('styles')} or{' '}
+            {code('accessibility')}).
+          </strong>{' '}
+          This is because, in contrast to other forms of shorthand values, Stardust-evaluated props
+          cannot be unsafely passed to the element which type is, generally, doesn't allow Stardust
```
mnajdova: cannot be safely?

docs/src/views/ShorthandProps.tsx
```diff
@@ -1,37 +1,263 @@
...
+        </p>
+
+        <ul>
+          <li>it should finish syncronously</li>
+          <li>it should return React Element as a result</li>
+        </ul>
+
+        <p>Thus, in its simplest form, it could be used the following way:</p>
+
+        {codeExample([`<Button icon={() => <Icon name='chess rook' />} />`])}
```
mnajdova: Not sure that this example brings any value... Let's try not to add examples which should never be used.

docs/src/views/ShorthandProps.tsx
```diff
@@ -1,37 +1,263 @@
...
+
+        {codeExample([
+          `<Button icon={render => render(`,
+          `  /* what to render */`,
+          `  { name: 'chess rook' },`,
+          ``,
+          `  /* how to render */`,
+          `  (ComponentType, props) => (`,
+          `    <div class='my-icon-container'>`,
+          `      <i {...props} />`,
```
mnajdova: Maybe we can showcase here the usage of the styles and accessibility, like spreading the accessbility.root elements on the `i` element or using the classes generated?

docs/src/views/ShorthandProps.tsx
kuzhelov: this one provides a visibility of what does it mean to satisfy criterias of this function. If we will start right with the render callback stuff, it won't be absolutely clear what is going on - there might be a false assumption from the client that this 'render' function is, actually, triggers rendering somewhere, while, in fact, it just returns an element.

I would really like to leave this simple explanation of what does it mean to pass function as a shorthand value, and what Stardust expects from this function.

What do you think?

docs/src/views/ShorthandProps.tsx
kuzhelov: a bit reluctant to use this terminology, as `default props` for Component may lead to false associations. Essentially, default element's prop is even more correct, given how this default prop's value is handled

docs/src/views/ShorthandProps.tsx
```diff
@@ -1,37 +1,264 @@
...
 import { NavLink } from 'react-router-dom'
+import { Header } from 'semantic-ui-react'
 import DocPage from '../components/DocPage/DocPage'
 import CodeSnippet from '../components/CodeSnippet'
 import Button from '../../../src/components/Button/Button'
 
+const joinToString = stringOrArray => {
+  if (typeof stringOrArray === 'string') {
+    return stringOrArray
+  }
```
layershifter: `CodeSnippet` may be we should move this directly to the component?

docs/src/views/ShorthandProps.tsx
```diff
@@ -1,37 +1,264 @@
...
+
+  return stringOrArray.join('\n')
+}
+
+const header = content => <Header as="h2">{content}</Header>
+const subheader = content => <Header as="h3">{content}</Header>
+
+const code = content => <code>{content}</code>
+const codeExample = snippet => <CodeSnippet value={joinToString(snippet)} />
```
layershifter: I have a strange feeling about there helpers I understand why there are there, but I also should mention that they are not looking friendly for newcomers 🤔

docs/src/views/ShorthandProps.tsx
```diff
@@ -1,37 +1,264 @@
...
+
 class ShorthandProps extends React.Component {
```
layershifter: We can use there functional component, this will reduce nesting level (less spaces and more compact markup)⭐️

docs/src/views/ShorthandProps.tsx
kuzhelov: yes, thought about the same thing as well..

layershifter (approved):

PR #756, opened 2019-01-22
feat(menu): color prop and complex scheme

bmdalex:
### Description

This PR:
- continues the work started in #712
- adds `color` prop to `Menu` component
- implements complex color scheme
- improves color palette
- creates `Menu` color examples
- fixes menu bugs
	- missing background color
	- wrong color for different states

### API

```jsx
<Menu color={COLOR} .../>
```
 where `COLOR` can be:
- a string: `'primary' | 'secondary' | 'blue' | 'green' | 'grey'
- an object with:
  - key: one of `'foreground' | 'background' | 'border'` to identify the target area of the component
  - value: one of the strings above to describe  the color of the area
e.g.:
```jsx
<Menu color={ foreground: 'pink', background: 'yellow', border: 'green' }} .../>
```
renders:
screenshot 2019-01-14 at 02 47 47

#### 1. Regular theme
screenshot 2019-01-22 at 00 36 21

#### 2. Dark theme
screenshot 2019-01-22 at 00 36 34


docs/src/examples/components/Menu/Variations/MenuExampleColor.shorthand.tsx
```diff
@@ -0,0 +1,61 @@
...
+            color={{ foreground: colors[4], background: colors[5], border: colors[1] }}
+            items={items}
+          />
+        </Grid>
+      )
+    }}
+  />
+)
+
+export default MenuExampleColor
```
bmdalex, line 51: for tracking @layershifter 's comment in: https://github.com/stardust-ui/react/pull/712#discussion_r247543678

> This example looks overcomplicated to me, may be we can:
> 
> * introduce a new `Select` knob
> * use it to get select an active color

will address

src/components/Menu/MenuItem.tsx
```diff
@@ -229,6 +233,7 @@ class MenuItem extends AutoControlledComponent<ReactProps<MenuItemProps>, MenuIt
           {Menu.create(menu, {
             defaultProps: {
               accessibility: submenuBehavior,
+              color,
               vertical: true,
               primary,
```
codepretty, line 239: Should we remove `primary` and `secondary` props in MenuItem and Menu.tsx? And references to `secondary` in menuItemStyles as part of this change?

src/themes/teams/components/Menu/menuVariables.ts
```diff
@@ -1,52 +1,78 @@
...
 }
 
-export default (siteVars: any): MenuVariables => {
-  return {
+export default (siteVars: SiteVariablesPrepared): MenuVariables => {
+  const v: MenuVariables = {
     color: siteVars.gray02,
+    focusedColor: siteVars.black,
+    disabledColor: siteVars.gray06,
+    primaryActiveColor: siteVars.white,
```
codepretty, line 33: Would it make more sense to get rid of specific variables like this `primaryActiveColor` and just set `primary.foreground.active` to `siteVars.white` directly?

src/themes/teams/components/Menu/menuVariables.ts
```diff
@@ -1,52 +1,78 @@
...
-    primaryFocusedColor: siteVars.white,
     primaryFocusedBackgroundColor: siteVars.brand12,
 
+    borderColor: siteVars.gray08,
     primaryBorderColor: siteVars.brand08,
-    primaryHoverBorderColor: siteVars.gray08,
-    primaryUnderlinedBorderColor: siteVars.gray08,
-
-    disabledColor: siteVars.gray06,
+    primaryActiveBorderColor: siteVars.brand,
```
codepretty, line 44: You're using this for iconOnly colors instead of `iconOnlyActiveColor`, so this value should be what that was before (`brand06`) otherwise the style is incorrect in dark theme

image

src/themes/teams-high-contrast/components/Menu/menuVariables.ts
```diff
@@ -1,13 +1,80 @@
...
+
+    backgroundColor: siteVars.black,
     activeBackgroundColor: siteVars.accessibleYellow,
+    focusedBackgroundColor: siteVars.accessibleYellow,
+    primaryActiveBackgroundColor: siteVars.accessibleYellow,
+    primaryFocusedBackgroundColor: siteVars.accessibleYellow,
+
+    borderColor: siteVars.white,
+    primaryBorderColor: siteVars.white,
+    primaryActiveBorderColor: siteVars.white,
```
codepretty, line 46: You're using this color for the icons color when being active, so it should match what `activeColor` is. Otherwise, the active state of the icon menu in high contrast is inaccessible
image

src/themes/teams/components/Menu/menuItemStyles.ts
```diff
@@ -12,61 +19,65 @@ const underlinedItem = (color: string): ICSSInJSStyle => ({
...
+  const { iconOnly, underlined, isFromKeyboard, active, pointing, vertical } = props
+
+  if (iconOnly) {
+    return { color: colors.border.focused }
+  }
+
   if (active && !underlined) return {}
+
+  if (underlined && !isFromKeyboard) {
+    return { color: v.focusedColor }
```
codepretty, line 62: It's a little confusing because these variables are mixed with color scheme variables throughout the file. What is the right paradigm to follow? Do we use color scheme for everything, except when we need a specific exceptions or "override"? It looks like this might be the only time this variable is used. If so and if we're only using non-color scheme variables for specific instances, can we rename this variable to something specific like `v.underlineInitialFocusedColor` or something?

src/themes/teams-dark/components/Menu/menuVariables.ts
```diff
@@ -1,9 +1,19 @@
...
 export default (siteVars: any): MenuVariables => {
   return {
     color: siteVars.white,
+
+    primaryActiveColor: siteVars.black,
+    primaryFocusedColor: siteVars.black,
+
+    backgroundColor: siteVars.black,
   }
 }
```
codepretty, line 19: Why do we not need color scheme in this file, but we do for the contrast variables?

levithomason (changes requested):
This complex color prop design is not inline with the current vision or principles.  See https://github.com/stardust-ui/react/pull/647#discussion_r247304463.

Let's have a sync before merging any more of these changes.

src/themes/teams-dark/components/Menu/menuVariables.ts
levithomason: Before adding any more variables, let's nail down exactly why we need certain variables and make them consistent across components.   To do this, we need user stories and requirements to fulfill.

We should probably have a focused effort on reviewing components and the usefulness of certain variables across many components.  Once we have requirements and have spent some time designing a variable scheme to fill the requirements, we can then start adding more variables.

src/themes/teams/components/Menu/menuVariables.ts
levithomason: Flat variables are preferred for now for consistency, ease of merging, and ease of discovery.

Once we formalize the structure of our design system, we might have nested values like this. For now, let's keep consistent and flat.

docs/src/examples/components/Menu/Variations/MenuExampleColor.shorthand.tsx
levithomason: ### Bare minimum examples

Reference SUIR for examples of clean Menu color examples.  Let's not put so much information in a single example.

https://react.semantic-ui.com/collections/menu/#variations-colored

### Definition only, avoid large permutations

The doc page should only be a "definition" of the component right now. That is a single example for each thing the menu can do, usually one example per prop.

At this point, let's not add examples for all possible permutations of what the menu can do.

src/themes/teams/components/Menu/menuVariables.ts
codepretty: Okay, does that mean we should have the flat variable inherit from the nested color scheme values? Or get rid of the color scheme nested values altogether?

levithomason: We're going to have a meeting on color in the morning.  This needs more discussion that is suitable in this PR's comments.  It will likely be early, would you like to join?

codepretty: Add me to the invite. I would like to join if I can :)

src/components/Menu/MenuItem.tsx
bmdalex: we can do that, but I wanted to have a separate PR for it; it also needs to be discussed again; I'll update the comment after we sync

src/themes/teams/components/Menu/menuVariables.ts
bmdalex: to be updated after we discuss

src/themes/teams/components/Menu/menuVariables.ts
bmdalex: will address after our sync

src/themes/teams-high-contrast/components/Menu/menuVariables.ts
bmdalex: will address after our sync

src/themes/teams/components/Menu/menuItemStyles.ts
bmdalex: we need this to be a variable because this specific color should not change with the `color` prop (it's the same even if color is `red` or `green`)

regarding the name, it's debatable if we want to include something so specific... we need to come up with a naming convention for variables; if we take into account @codepretty 's proposal (`underlineInitialFocusedColor`) we can find the pattern to be:

`${propNames}${state}${area}Color`

where:
- `propNames`: is a concatenation of props that are set to true for the branch the color is used;
- `state`: is the state of the component (initial, focused, hovered, active, disabled, etc)
- `area`: is the area of the component that gets the color (border, background, foreground, etc)

docs/src/examples/components/Menu/Variations/MenuExampleColor.shorthand.tsx
bmdalex: @levithomason 
in SUIR there are not that many versions of Menus; I just introduced a Select Knob at @layershifter 's proposal and now we have examples that look like this:

TODO - introduce screenshot

We can separate these into different example files where we reuse the knob or we can add a knob on every example for the color, but because of the way knobs are implemented we need to have a `.knobs.tsx` file for every example, which is a lot of duplication

what do you think?

bmdalex:
### Closing for now.

## Reasons:

- this work not yet aligned with current priorities
- at the moment, the library is not ready to accept the complex changes this PR introduces

## Findings and TODOs:

### 1. We need to drop complex color prop design and go back to the simpler API:
```jsx
<Menu color='string' ... />
```

#### TODO:
- remove complex color prop
- stick to color prop with string values

### 2. Flat variable names vs nested color scheme (comment: https://github.com/stardust-ui/react/pull/756#discussion_r249959701) :

- the idea of the color scheme itself is good and should be used in our approach after we simplify it 
- not all components have the same color scheme so we should start from defining color schemes for individual components and then try find patterns and extract these into more complex logic
- there's too much abstraction in `renderComponent`, see `generateColorScheme`

#### TODO:
- for now we'll remove the color scheme from variables, because we're not ready for the concept of nested variables (needs more research and experimentation); we don't want to allow users to customize something like this yet
- color scheme will be moved to the styles file of each component
	 
### 3. How do we handle colors values that depend on prop values?

Right now, the color scheme structure doesn't support that; it identifies only colors, areas (background, border, etc) and states (initial, hovered, active, etc)

#### TODO:
- rework the PR to make it the color scheme specific for each component
- we will use a nested structure on the lines of: `colors[colorName][area][state][prop]`, where:
  - `colors` is the variable for colors;
  -` colorName` is one of the names of colors from palette ('primary' | 'secondary' | 'blue' | 'green' | 'grey' | 'orange' | 'pink' | 'purple' | 'teal' | 'red' | 'yellow' | string)
  - `area` is the component area the color belongs to ('foreground' | 'background' | 'border', etc)
  - `state` is the UI state of the component (initial, focused, hovered, active, disabled, etc)
  - `prop` is an identifier for menu prop combinations that are true and need different variants (active background primary color is different than underlined 

### 4. `primary` as name for prop and color is confusing

#### TODO:
- rename `primary` color to `brand`
		
## Color Scheme Proposal:

```typescript
const colorScheme = {
  primary: {
    foreground: {
      initial: colorVariants[XXX],
      active: colorVariants[XXX],,
      disabled: colorVariants[XXX],,
      focused: colorVariants[XXX],,
      hovered: colorVariants[XXX],,
    },
    background: {
      // same props as above
    },
    border: {
      // same props as above
    },
    shadow: {
      // same props as above
    },
  },

  red: {
    // same props as above
  },

   // rest of colors
}
```

PR #758, opened 2019-01-22
fix(Popup): Adding toggle functionality if trigger element is not button

kolaps33:
Currently if the trigger element for the popup is not button, then popup is not open by Enter/Space key.

Specification by @jurokapsiar regarding Mouse vs Keyboard action. 
Mouse hover > then with keyboard "Enter" should open popup and "ESC" close
Focus > then with keyboard "Focus" should open popup and "ESC" close
Click > 
- then with keyboard "Enter" should open popup and "ESC" close
- then with keyboard "Enter" should open popup and "Enter" close (toggle)


levithomason (dismissed):
Let's get a test added that shows this works.  Ensure to make assertions only on what the end user has access to.  That means avoid making assertions that the state changed, and prefer things like checking if the popup is actually open on the page.

sophieH29 (approved):

src/components/Popup/Popup.tsx
```diff
@@ -176,6 +176,10 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
       e.stopPropagation()
     },
     close: e => this.close(e),
+    toggle: e => {
+      e.preventDefault()
```
silviuaavram, line 180: what is this added for?

test/specs/components/Popup/Popup-test.tsx
```diff
@@ -102,4 +111,28 @@ describe('Popup', () => {
       expect(yOffsetTransformed.trim()).toBe(yOffset)
     })
   })
+
+  describe('toggle popup', () => {
+    test('with enter/space key', () => {
```
silviuaavram: I think this test is limited to testing that Enter on trigger opens the popup and Space on trigger closes the popup.

Maybe make 2 separate tests. One for Enter one for Space. Same workflow. Test that key opens the popup and then the same key closes it.

You can even use the same function and create 2 tests with it, just change the key, pass it as parameter.

silviuaavram (changes requested):
Split the test into 2 distinct ones. But the thing you actually test is ok.

src/components/Popup/Popup.tsx
kolaps33: this is when space key trigger open popup...

test/specs/components/Popup/Popup-test.tsx
kolaps33: good point :) implemented two tests one for Enter and second for Space

silviuaavram (approved):

jurokapsiar (approved):

mnajdova:
This has regressions on the popups that have `on="hover"` or `on="focus"`. In this cases, the enter/space should always open the popup not open and close it. We should add the toggling behavior only if we have `click` as a combination in the value of the `on` prop.

jurokapsiar:
>  We should add the toggling behavior only if we have `click` as a combination in the value of the `on` prop.

agree for click. for other combinations we should do the same what click event does - open the popup if not already open.

test/specs/components/Popup/Popup-test.tsx
```diff
@@ -36,6 +38,57 @@ describe('Popup', () => {
   }: PositionTestInput & { rtl?: never }) =>
     testPopupPosition({ align, position, expectedPlacement, rtl: true })
 
+  const getPopupContent = (popup: ReactWrapper) => {
+    return popup.find('.ui-popup__content')
```
kuzhelov: lets use interpolated class names instead of hardcoded strings: 
```js
return popup.find(`${Popup.Content.className}`)
```

test/specs/components/Popup/Popup-test.tsx
```diff
@@ -131,4 +184,11 @@ describe('Popup', () => {
       expect(spy.mock.calls[0][1]).toMatchObject({ open: true })
     })
   })
+
+  describe('open/close popup by keyboard', () => {
+    togglePopupTest('enter', keyboardKey.Enter, 'click')
```
kuzhelov: nit: lets rename it in a way `togglePopupTest ` -> `testPopupToggleOn`, this will make test's code more readable

src/lib/accessibility/Behaviors/Popup/popupBehavior.ts
```diff
@@ -11,6 +11,7 @@ import * as _ from 'lodash'
  * Adds attribute 'aria-disabled=true' to 'trigger' component's part if 'disabled' property is true. Does not set the attribute otherwise.
  */
 const popupBehavior: Accessibility = (props: any) => {
+  const normalizedOn = _.isArray(props.on) ? props.on : [props.on]
```
kuzhelov: would suggest name like `onAsArray` - that would be much more expressive than general 'normalized'

src/lib/accessibility/Behaviors/Popup/popupBehavior.ts
```diff
@@ -33,6 +34,18 @@ const popupBehavior: Accessibility = (props: any) => {
...
           keyCombinations: [{ keyCode: keyboardKey.Escape }],
         },
+        toggle: {
+          keyCombinations: _.includes(normalizedOn, 'click') && [
+            { keyCode: keyboardKey.Enter },
+            { keyCode: keyboardKey.Spacebar },
+          ],
+        },
+        open: {
+          keyCombinations: _.includes(normalizedOn, 'hover') && [
```
kuzhelov: this is misleading - the prop is about taking key combinations, but currently we are starting to handle pseudo states under the same key-combinations-devoted prop

src/lib/accessibility/Behaviors/Popup/popupBehavior.ts
```diff
@@ -11,6 +11,7 @@ import * as _ from 'lodash'
  * Adds attribute 'aria-disabled=true' to 'trigger' component's part if 'disabled' property is true. Does not set the attribute otherwise.
  */
 const popupBehavior: Accessibility = (props: any) => {
```
kuzhelov, line 12: not related to this PR in particular, but could you, please, suggest if there is any reason for not introducing types for props in accessibility behaviors, i.e.:

```jsx
const popupBehavior: Accessibility = (props: PopupProps & PopupState) => {
```

This will alleviate prop rename refactoring changes, as well as clearly suggest what this input `props` arg should consist of.

test/specs/components/Popup/Popup-test.tsx
```diff
@@ -36,6 +38,57 @@ describe('Popup', () => {
...
+        <Popup
+          trigger={<span id={triggerId}> text to trigger popup </span>}
+          content={popupContent}
+          on={onProp}
+        />,
+      )
+      const popupTriggerElement = popup.find(`#${triggerId}`)
+      popupTriggerElement.simulate('keydown', { keyCode: keyboardKeyToOpen })
+      expect(getPopupContent(popup).length).toBe(1)
+      expect(getPopupContent(popup).getElement().props.children).toMatch(popupContent)
```
kuzhelov: this is brittle test - it will break once inner structure of `PopupContent` will change for some reason. We can achieve the same more robustly:

```jsx
const contentId = 'fancy-content-id'

// ....

 const popup = mountWithProvider(
        <Popup
          trigger={<span id={triggerId}> text to trigger popup </span>}
          content={<span id={contentId}/>}
          on={onProp}
        />,
  )

// ...
expect(getPopupContent(popup).find(`#${contentId}`)).length).toBe(1)

```

test/specs/components/Popup/Popup-test.tsx
```diff
@@ -36,6 +38,57 @@ describe('Popup', () => {
...
+    keyboardKeyToOpen: keyboardKey,
+    keyboardKeyToClose: keyboardKey,
+    onProp: PopupEvents,
+  ) => {
+    test(`open popup with ${keyNameToOpen} key and close it with escape key`, () => {
+      const triggerId = 'triggerElement'
+      const popupContent = 'any test content'
+      const popup = mountWithProvider(
+        <Popup
+          trigger={<span id={triggerId}> text to trigger popup </span>}
```
kuzhelov: `text to trigger popup` can be safely omitted - it will be better to do that so the test's arrangement will be more concise

test/specs/components/Popup/Popup-test.tsx
```diff
@@ -36,6 +38,57 @@ describe('Popup', () => {
...
+          content={popupContent}
+          on={onProp}
+        />,
+      )
+      const popupTriggerElement = popup.find(`#${triggerId}`)
+      popupTriggerElement.simulate('keydown', { keyCode: keyboardKey })
```
mnajdova: I would say that, if the span is not focusable we will not be able to repro this in a real usage. Unfortunately, no better idea at this moment..

src/lib/accessibility/Behaviors/Popup/popupBehavior.ts
mnajdova: Why are we using this with 'hover' and not with 'focus' as well. Currently, if the focus is put on a Button, it opens on enter, but if it is a span with tabindex='1' it is not opening on enter... If we use on='focus' and we open the popup and close it with esc, we have no other way to open it afterwards, expect focusing on other element and then back to it... @jurokapsiar is it the desired behavior?

jurokapsiar: I think this is fine. Think of on='focus' as a case for tooltip - you show the tooltip on focus, but on click/enter/space there typically is a different action being executed.

kolaps33: issue reported:
https://github.com/stardust-ui/react/issues/798

src/lib/accessibility/Behaviors/Popup/popupBehavior.ts
jurokapsiar: #804

yarn.lock
```diff
@@ -2969,10 +2969,10 @@ dot-prop@^4.1.0:
   dependencies:
     is-obj "^1.0.0"
 
-downshift@^3.2.0:
-  version "3.2.0"
-  resolved "https://registry.yarnpkg.com/downshift/-/downshift-3.2.0.tgz#820f40575683d3fc521a9ffd409ef4a69cef531d"
-  integrity sha512-IXZ3IB7xU51R04olRGSp8GBmJhPlCL8tW5FjZrJh8e4kB4BGUWsQHZ9tNsO+qNLABieR5DIgYBZ9MncPlwZSDg==
+downshift@3.2.1:
```
kuzhelov: why these changes are necessary?

test/specs/components/Popup/Popup-test.tsx
```diff
@@ -131,4 +159,19 @@ describe('Popup', () => {
       expect(spy.mock.calls[0][1]).toMatchObject({ open: true })
     })
   })
+
+  describe('open/close popup by keyboard', () => {
+    test(`toggle popup with Enter key`, () => {
+      openAndClosePopup(keyboardKey.Enter, keyboardKey.Enter, 'click')
```
kuzhelov: nit: this name might work better: `expectPopupToOpenAndClose`

mnajdova (approved):

kuzhelov (approved):

PR #759, opened 2019-01-22
fix(factories): reverted changed for not passing props to elements in the factories

### LINKED ISSUE #671: Creating a component slot with .create function will not apply styles
## Bug Report
Creating a component slot with .create function will not apply styles from the styles file to that slot.
### Steps
The code from DropdownLabel.tsx:
```jsx
      <Label
        styles={styles.root}
        role="presentation"
        circular
        onClick={this.handleClick}
        content={header}
        icon={Icon.create(icon, {
          defaultProps: {
            'aria-label': 'Remove ${header} from selection.',
            'aria-hidden': false,
            role: 'button',
          },
          overrideProps: this.handleIconOverrides,
        })}
        image={Image.create(image, {
          defaultProps: {
            avatar: true,
          },
        })}
        {...rest}
      />
```

Both Icon and Image will not get styles applied to them from labelStyles.ts.
Both icon and image are shorthands, passed into the Dropdown component. To see the actual bug, go to the Dropdown component in the docs, example Multiple Search with Image and Content.

screenshot 2019-01-02 at 13 07 17

The line 132 in factories.ts might be the cause of this issue:
`
  if (valIsReactElement) {
    return value as React.ReactElement<Props>
  }
`

### Expected Result
Styles should be applied in all cases except when we pass an actual <Image> or <Icon> component to the Dropdown props icon and image.

### Actual Result
Styles are not applied.

### Version
0.15

mnajdova:
This PR is reverting some of the changes introduced in https://github.com/stardust-ui/react/pull/496 regarding not passing props in the factories if the element is a React component, in order to fix https://github.com/stardust-ui/react/issues/671


test/specs/lib/factories-test.tsx
```diff
@@ -524,21 +534,18 @@ describe('factories', () => {
...
-          { commonProp: 'originalElement', originalElementProp: true },
-        )
+      itOverridesDefaultProps(
+        'element',
+        { some: 'defaults', overridden: false },
+        { some: 'defaults', overridden: true },
+        { value: <div {...{ overridden: true } as any} /> },
+      )
+      itOverridesDefaultPropsWithFalseyProps('element', {
+        value: <div {...{ undef: undefined, nil: null, zero: 0, empty: '' } as any} />,
```
levithomason, line 548: Do we need to create these superfluous objects and spread them like this?  It would be much simpler to just pass the props, if possible:

```suggestion
        value: <div undef={undefined} nil={null} zero={0} empty='' />,
```

levithomason (approved):
See one nit comment.

test/specs/lib/factories-test.tsx
mnajdova: Typescript is complaining because these props are not valid for div element, that's why we have this object spread. I can create another object and spread it, but it is much easier indeed to read it if it is inline on the element.

docs/src/views/ShorthandProps.tsx
```diff
@@ -86,14 +86,12 @@ const ShorthandProps = props => (
     <blockquote>
       <strong>
         There is a very important caveat here, though: whenever React Element is directly used as a
-        shorthand value, it becomes client's responsibility to handle some aspects that Stardust was
-        originally responsible for (such as {code('styles')} or {code('accessibility')}).
+        shorthand value, all props that Stardust has created for the slot's Component will be spread
+        on the passed element. This means, you may end up with invalid props applied on HTML
```
kuzhelov: `This means that provided element should be able to handle Stardust props - while this requirement is satisfied for all Stardust components, you should be aware of that when either HTML or any third-party elements are provided.`

The reason I would suggest to rephrase original one is elimination of `invalid props` - in fact, there is nothing invalid in props provided, it is just about element may be not aware of necessity to handle them.

mnajdova: Speaking of this, maybe using the `<i />` is not the best example? Should we consider using another Stardust component, like <Image /> for example?

kuzhelov (approved):

PR #761, opened 2019-01-23
fix(ChatMessage): Author element should always be rendered

sophieH29:
Author element should always be rendered, but hidden in styles, so the screen reader can narrate it.


jurokapsiar (approved):

kolaps33:
tested with JAWS/NVDA looks good :)

miroslavstastny (approved):

PR #762, opened 2019-01-23
fix(Text): refactor variables to match `size` prop

### LINKED ISSUE #454: Text component size-variable match
In version 0.11.0 the Text component supports:
- 7 sizes from "smaller" to "largest".
- 5 variables for overriding text sizes from "fontSizeExtraSmall" to "fontSizeExtraLarge".
- 5 variables for overriding line height from "fontLineHeightExtraSmall" to "fontLineHeightExtraLarge".

The problems with this:
- there are missing variables for some sizes.
- naming doesn't match - should "fontSizeExtraLarge" change "larger" or "largest" size?

layershifter:
Fixes #454.

### TODO

- [x] add CHANGELOG entry


docs/src/examples/components/Text/Variations/TextExampleColor.tsx
```diff
@@ -6,12 +6,10 @@ const TextExampleColor = () => (
   <ProviderConsumer
     render={({ siteVariables: { emphasisColors, naturalColors } }) =>
       _.keys({ ...emphasisColors, ...naturalColors }).map(color => (
-        <>
-          <Text key={color} color={color}>
-            {_.startCase(color)}
-          </Text>
+        <React.Fragment key={color}>
```
layershifter, line 9: Fixes `key` warning 👍

src/themes/teams/siteVariables.ts
```diff
@@ -50,11 +50,13 @@ export const shadowLevel1 = '0 .2rem .4rem -.075rem rgba(0, 0, 0, 0.1)'
 // FONT SIZES
 //
 export const fontSizes = {
+  smallest: pxToRem(8),
   smaller: pxToRem(10),
   small: pxToRem(12),
   medium: pxToRem(14),
   large: pxToRem(18),
   larger: pxToRem(24),
+  largest: pxToRem(28),
```
layershifter: I'm not sure that it's the best idea, should I revert this change?

src/themes/base/siteVariables.ts
```diff
@@ -8,11 +8,13 @@ export { colors, contextualColors, emphasisColors, naturalColors, colorScheme }
 // FONT SIZES
 //
 export const fontSizes = {
+  smallest: '0.8rem',
```
kuzhelov, line 11: while it seems to be logical, the counterpoint is that we, generally, will 'close' the interval of possible values in terms of extensibility - especially when we are talking about `base` theme. 

The problem is that, with this being, essentially, a finite set, we cannot guarantee that client's scenario won't require additional sizes added to the set - and, in that case, we cannot ensure that client will be able to extend `base` theme and introduce, say, 10 (and not 7, as prescribed) font sizes _with meaningful names_ . 

So, because of that I am a bit reluctant to introduce this change now.

src/themes/base/components/Text/textVariables.ts
```diff
@@ -47,8 +51,11 @@ export default (siteVariables): TextVariables => {
     errorColor: siteVariables.colors.red[500],
     successColor: siteVariables.colors.green[500],
 
-    fontSizeExtraSmall: siteVariables.fontSizes.smaller,
-    fontLineHeightExtraSmall: siteVariables.lineHeightExtraSmall,
+    fontSizeSmallest: siteVariables.fontSizes.smallest,
+    fontLineHeightSmallest: siteVariables.lineHeightExtraSmall,
```
kuzhelov: seems that this one wasn't replaced

src/themes/base/siteVariables.ts
```diff
@@ -8,11 +8,13 @@ export { colors, contextualColors, emphasisColors, naturalColors, colorScheme }
...
   smaller: '1rem',
   small: '1.2rem',
   medium: '1.4rem',
   large: '1.8rem',
```
kuzhelov, line 15: frankly, to me it is not absolutely intuitive whether `large` or `larger` is bigger while just having a markup:

```jsx
<Button size='larger' />
```

### Further thoughts

It would be much better to have some range where each value's name would immediately deliver insights on what is the name of next larger item and next smaller one, like, say, in the following case:

```js
xx-small    // or x2-small, to reduce amount of x's
x-small
medium
large
x-large
xx-large
```

Amongst other things it would be great if this range would be extendable, so that client's theme may easily provide additional sizes to it while maintaining the overall order of values.

With this order being established, we might be able to decently handle scenarios when one theme, with wider range of sizes, is switched on another one where this range is smaller:

```jsx
// original theme
const fontSizes = {
  xx-small
  x-small
  medium
  large
  x-large
  xx-large
}

// theme to switch on - note, no values for xx-large/small
const fontSizes = {
  x-small
  medium
  large
  x-large
}
```

The idea here is just to allow next theme to choose nearest size to the one that is requested (so, for `xx-large` new theme will use its `x-large`) - and this is pretty much the same what is happening with fonts in Web (both speaking of sizes and weights).

### To conclude

This was just one of the approaches we might think about - definitely, there may be lots of other ones. At the same time, if we would like to provide client with an intuitive range of values, as well as to support theme switching between two that have different set of sizes defined, we would need to satisfy the following conditions:
- names of values suggest their order
- it is possible to infer next closest value from the name of each one
- there is an extendable range of values

src/themes/teams/siteVariables.ts
layershifter: Reverted: https://github.com/stardust-ui/react/issues/136#issuecomment-447152727

src/themes/base/siteVariables.ts
layershifter: > The sizing scheme we decided to go with for now is: smallest, smaller, small, medium, large, larger, largest.

#640

> For themes that may want 5 sizes, just have the largest/smallest sizes use the same values as larger/smaller to limit them.

https://github.com/stardust-ui/react/issues/136#issuecomment-447152727

src/themes/base/components/Text/textVariables.ts
```diff
@@ -47,8 +51,11 @@ export default (siteVariables): TextVariables => {
...
+
+    fontSizeSmaller: siteVariables.fontSizes.smaller,
+    fontLineHeightSmaller: siteVariables.lineHeightSmaller,
```
kuzhelov, line 58: just to be sure - thus, speaking of the `lineHeight` range, we will have values from `smaller` to `small`, right?

kuzhelov (approved):

src/themes/base/siteVariables.ts
kuzhelov: thanks for these references and clarifications!

PR #765, opened 2019-01-24
docs(Prototype): Popover emojis should be visible to Virtual Cursor

sophieH29:


docs/src/prototypes/chatMessageWithPopover/Popover.tsx
```diff
@@ -36,19 +36,40 @@ class Popover extends React.Component<PopoverProps, PopoverState> {
...
-      display: 'flex',
+      position: 'initial',
+      zIndex: 'initial',
+      opacity: 1,
     },
   })
 
   render() {
+    const overridenToolbarBehavior: Accessibility = (props: any) => {
+      const toolbarFocusZone = toolbarBehavior(props).focusZone
```
jurokapsiar: This can be simplified to:
```
const behavior = toolbarBehavior(props)
behavior.focusZone.props.defaultTabbableElement = {...}
return behavior
```

sophieH29: looks better, thx

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
```diff
@@ -36,19 +36,33 @@ class Popover extends React.Component<PopoverProps, PopoverState> {
...
     '&:hover .smile-emoji': {
-      display: 'flex',
+      position: 'initial',
+      zIndex: 'initial',
+      opacity: 1,
     },
   })
 
   render() {
+    const overridenToolbarBehavior: Accessibility = (props: any) => {
```
kuzhelov: there is no need to create this function on each `render` - this may be put out of the `render` scope

jurokapsiar (approved):

docs/src/prototypes/chatMessageWithPopover/Popover.tsx
```diff
@@ -28,6 +28,14 @@ class Popover extends React.Component<PopoverProps, PopoverState> {
     this.changeFocusState(shouldPreserveFocusState)
   }
 
+  getToolbarBehavior: Accessibility = (props: any) => {
+    const behavior = toolbarBehavior(props)
+    behavior.focusZone.props.defaultTabbableElement = (root: HTMLElement): HTMLElement => {
+      return root.querySelector('[aria-label="thumbs up"]')
+    }
+    return behavior
```
kuzhelov, line 36: isn't it possible to create this object just once and not during each rendering phase?

sophieH29: @kuzhelov this behavior's wrapper works the same way as we'd use behavior imported from the library, mainly - passing a callback function to component, which is called in `renderComponent.tsx` and passing props into it.

sophieH29:
closing in regards of https://github.com/stardust-ui/react/pull/811

PR #767, opened 2019-01-24
feat(ChatItem): add attached prop, renamed gutterPosition to contentPosition

### LINKED ISSUE #588: [RFC] `Chat.Item` `grouped` prop for describing groups of chat items
#### Legend:
- **batch**: group of consecutive messages from the same participant

#### What?

We need a way to describe either:
- batches of chat messages;
- or chat messages that are not the first in a batch.

#### Why?

There are scenarios where there needs to be a clear visual indication that differentiates messages within a batch:
- redundant subcomponents;
- different styling (fitted consecutive messages).

#### Current:
screenshot 2018-12-10 at 17 12 00

#### Expected:
screenshot 2018-12-10 at 17 14 37

##### API:
```jsx
<Chat>
  <Chat.Item
    grouped
    gutter={...}
    content={(
      <Chat.Message
        content="Hi"
        author="Jane Doe"
      />
    )}
  />
  <Chat.Item
    gutter={...}
    content={<Chat.Message content="How are you todau?" author="Jane Doe" timestamp="Yesterday, 10:15 PM" />}
  />
  <Chat.Item
    gutter={...}
    content={<Chat.Message content="*today" author="Jane Doe" timestamp="Yesterday, 10:15 PM" />}
  />
</Chat>
```

##### API:
```jsx
<Chat>
  <Chat.Item
    gutter={...}
    content={(
      // Their group
      <Chat.Message.Group>
        <Chat.Message content="Hi" author="Jane Doe" timestamp="Yesterday, 10:15 PM" />
        <Chat.Message content="How are you today?" author="Jane Doe" timestamp="Yesterday, 10:15 PM" />
        <Chat.Message content="Good, and you?" mine author="Jane Doe" timestamp="Yesterday, 10:15 PM" />
      </Chat.Message.Group>
    )}
  />
  <Chat.Item
    gutter={...}
    content={(
      // My group
      <Chat.Message.Group>
        <Chat.Message content="Good, and you?" mine author="Jane Doe" timestamp="Yesterday, 10:15 PM" />
      </Chat.Message.Group>
    )}
  />
</Chat>
```

#### 3. Try to achieve the visual differences with current capabilities (might not cover all scenarios):

- we will just not add `gutter`, `author` and `timestamp` subcomponents
- we will remove extra margins between messages and replace them with invisible `Divider` components that we'd add manually between batches

mnajdova:
This PR adds the attached prop for the ChatItem, in order for the theme to be able to style differently the messages that are part of a batch. The value of the attached prop can be one of: b,oolean, 'top' or 'bottom'. The PR fixes https://github.com/stardust-ui/react/issues/588 

This is an example of this prop used with the teams theme:

image

image

```
import React from 'react'
import { Avatar, Chat, Divider } from '@stardust-ui/react'
const janeAvatar = {
  image: 'public/images/avatar/small/ade.jpg',
  status: {
    color: 'green',
    icon: 'check',
  },
}
const items = [
  {
    message: {
      content: (
      ),
    },
    contentPosition: 'end',
    attached: 'top',
    key: 'message-id-1',
  },
  {
    message: {
      content: (
        <Chat.Message content="I'm back!" author="John Doe" timestamp="Yesterday, 10:15 PM" mine />
      ),
    },
    contentPosition: 'end',
    attached: true,
    key: 'message-id-2',
  },
  {
    message: {
      content: (
        <Chat.Message
          content="Thanks for waiting!"
          author="John Doe"
          mine
        />
      ),
    },
    contentPosition: 'end',
    attached: 'bottom',
    key: 'message-id-3',
  },
  {
    gutter: {
      content: <Avatar {...janeAvatar} />,
    },
    message: {
      content: <Chat.Message content="Hi" author="Jane Doe" timestamp="Yesterday, 10:15 PM" />,
    },
    attached: 'top',
    key: 'message-id-4',
  },
  {
    gutter: {
      content: <Avatar {...janeAvatar} />,
    },
    message: {
      content: (
        <Chat.Message content="No problem!" author="Jane Doe" timestamp="Yesterday, 10:15 PM" />
      ),
    },
    attached: true,
    key: 'message-id-5',
  },
  {
    gutter: {
      content: <Avatar {...janeAvatar} />,
    },
    message: {
      content: (
        <Chat.Message content="What's up?" author="Jane Doe" timestamp="Yesterday, 10:15 PM" />
      ),
    },
    attached: 'bottom',
    key: 'message-id-6',
  },
  {
    message: {
      content: (
        <Chat.Message
          content="Would you like to grab a lunch?"
          author="John Doe"
          timestamp="Yesterday, 10:16 PM"
          mine
        />
      ),
    },
    contentPosition: 'end',
    key: 'message-id-7',
  },
  {
    gutter: {
      content: <Avatar {...janeAvatar} />,
    },
    message: {
      content: (
        <Chat.Message
          content="Sure! Let's try the new place downtown."
          author="Jane Doe"
        />
      ),
    },
    key: 'message-id-8',
  },
  {
    children: <Divider content="Today" color="primary" important />,
    key: 'message-id-9',
  },
  {
    message: {
      content: (
        <Chat.Message content="Ok, let's go." author="John Doe" timestamp="Today, 11:15 PM" mine />
      ),
    },
    contentPosition: 'end',
    key: 'message-id-10',
  },
]

const ChatExample = () => <Chat items={items} />

export default ChatExample

```


src/themes/teams/components/Chat/chatItemStyles.ts
```diff
@@ -1,24 +1,54 @@
...
+        display: 'none',
+      },
+      [chatMessageTimestampClassname]: {
+        display: 'none',
+      },
+    }),
+    marginBottom: 0,
   }),
 
   gutter: ({ props: p, variables: v }): ICSSInJSStyle => ({
```
layershifter, line 55: Can we follow `p.` pattern, it's more clear that it comes from props

src/components/Chat/ChatMessage.tsx
```diff
@@ -120,9 +120,20 @@ class ChatMessage extends UIComponent<ReactProps<ChatMessageProps>, ChatMessageS
         ) : (
           <>
             {!mine &&
-              Text.create(author, { defaultProps: { size: 'small', styles: styles.author } })}
+              Text.create(author, {
+                defaultProps: {
+                  size: 'small',
+                  styles: styles.author,
+                  className: `${ChatMessage.className}__author`,
```
mnajdova: This was needed for targeting this className in the chat item for displaying or not the author of the message. If we add this prop on the ChatMessage as well, we won't need this.

src/themes/teams/components/Chat/chatItemStyles.ts
```diff
@@ -1,25 +1,76 @@
 import { ICSSInJSStyle, ComponentSlotStylesInput } from '../../../types'
 import { ChatItemVariables } from './chatItemVariables'
 import { ChatItemProps } from '../../../../components/Chat/ChatItem'
+import { pxToRem } from '../../../../lib'
+import ChatMessage from '../../../../components/Chat/ChatMessage'
+
+const chatMessageClassname = `& .${ChatMessage.className}`
```
bmdalex: can we add these values to a constants file somewhere or export them from `ChatMessage.tsx` so there's only one place we change them?

src/themes/teams/components/Chat/chatItemStyles.ts
```diff
@@ -1,25 +1,76 @@
...
-  root: ({ props: p, variables: v }): ICSSInJSStyle => ({
-    position: 'relative',
-    marginTop: v.margin,
-    marginBottom: v.margin,
-  }),
+  root: ({ props: p, variables: v }): ICSSInJSStyle => {
+    const { grouped } = p
+    return {
+      position: 'relative',
+      ...((!grouped || grouped === 'start') && { marginTop: pxToRem(16) }),
```
bmdalex: nit: I like the `p.group` and `v.margin` pseudo-conventions because they give you an idea on where the variable comes from

src/themes/teams/components/Chat/chatItemStyles.ts
```diff
@@ -1,25 +1,76 @@
...
-  }),
+  gutter: ({ props: p, variables: v }): ICSSInJSStyle => {
+    const { grouped } = p
+    return {
+      position: 'absolute',
+      marginTop: v.gutterMargin,
+      [p.gutterPosition === 'end' ? 'right' : 'left']: 0,
+      ...(grouped &&
+        grouped !== 'start' && {
+          visibility: 'hidden',
```
bmdalex: we have `display: 'none'` on the author and timestamp and `visibility: 'hidden'` here, any reason we cannot keep consistency? I'd expect `display: 'none'` to work here as well

src/themes/teams/components/Chat/chatItemStyles.ts
```diff
@@ -1,25 +1,76 @@
...
+  },
 
-  message: ({ props: p, variables: v }): ICSSInJSStyle => ({
-    position: 'relative',
-    marginLeft: v.messageMargin,
-    marginRight: v.messageMargin,
-  }),
+  message: ({ props: p, variables: v }): ICSSInJSStyle => {
+    const { grouped } = p
+    return {
```
bmdalex: very nice 👍

bmdalex (commented):
really nice job, just a few comments to be addressed

src/themes/teams/components/Chat/chatItemStyles.ts
```diff
@@ -1,25 +1,76 @@
 import { ICSSInJSStyle, ComponentSlotStylesInput } from '../../../types'
 import { ChatItemVariables } from './chatItemVariables'
 import { ChatItemProps } from '../../../../components/Chat/ChatItem'
+import { pxToRem } from '../../../../lib'
+import ChatMessage from '../../../../components/Chat/ChatMessage'
+
+const chatMessageClassname = `& .${ChatMessage.className}`
+const chatMessageAuthorClassname = `& .${ChatMessage.className}__author`
+const chatMessageTimestampClassname = `& .${ChatMessage.className}__timestamp`
```
layershifter: nit: it's a selector, not class name 🐤

src/components/Chat/ChatItem.tsx
```diff
@@ -17,6 +17,9 @@ import Box from '../Box/Box'
 import { ComponentSlotStylesPrepared } from '../../themes/types'
 
 export interface ChatItemProps extends UIComponentProps, ChildrenComponentProps {
+  /** Indicates whether the ChatItem is the part of a batch. */
+  grouped?: 'start' | 'middle' | 'end'
```
layershifter: I remember that we discussed this before, but can you please point me to results?
Why we decided to use `grouped` instead of SUI's `attached={true|'top'|'bottom'}`?

https://react.semantic-ui.com/elements/segment/#variations-attached

mnajdova: My bad for not linking the issue in the PR.. Here is the issue with the discussion around this: https://github.com/stardust-ui/react/issues/588 We were considering using grouped, consecutive prop or additional component for representing the groups of messages. We decided there to use the grouped prop (which in the issue is only as an boolean, but it turned out that we must have some different words for describing the first, the last message and the middle messages in a batch. The options: start, end and middle were chosen because the start and end are the options we are using across all other components. Please share some additional proposals if you have any :)

src/themes/teams/components/Chat/chatItemStyles.ts
mnajdova: Yep, will rename those, although still not sure how to export this slot's classnames from the components... Should we add authorClassName, timestampClassName as static fields in the ChatMesage?! Maybe just a constants exported from the ChatMessage (although it will be not consistent as other class names defined).

mnajdova:
Guys I have one problem with adjusting the styles for the grouped messages when we are using the mine prop. Basically, if the ChatMessages have the mine prop, currently we are floating the message to the right and we are changing the color. The problem is, when the ChatItem is grouped and the ChatMessage is mine we should change the border radius of different corners. Looking into the API, maybe the `gutterPosition` prop can be renamed to something like `contentPosition`, or `pushContent` to 'start' | 'end' which will mean that both the gutter and the messages will be pushed to one side. Then we can know on ChatItem level which border corner radius should be altered. What do you think about this? It will be a breaking change, but I am not sure I have any other proposal at this moment. Open for other opinions. (The contentPosition was added, if we decide for different approach will refactor).

Last question, how do we want to export the classNames for the slots? See the author and timestamp classNames in the ChatMessage component.
 @Bugaa92 @kuzhelov @miroslavstastny @layershifter

src/themes/teams/components/Chat/chatItemStyles.ts
bmdalex: > Last question, how do we want to export the classNames for the slots? See the author and timestamp classNames in the ChatMessage component.

what about exporting a constants object from `ChatMessage.tsx`?

```typescript
export const chatMessageMetadata = {
  authorClassName: `${ChatMessage.className}__author`,
  timestampClassname: `${ChatMessage.className}__timestamp`,
}
```

mnajdova: I like this idea of adding metadata objects. Not sure whether we can add this to the UIComponent and use static field for it, as probably every components will have different structure for it, but exporting a constant for now may be enough...

src/themes/teams/components/Chat/chatItemStyles.ts
miroslavstastny: My first guess would be a static field, something like `ChatMessage.selectors.author`.

src/components/Chat/ChatItem.tsx
miroslavstastny: - It's true that we use `start` and `end` but that applies for horizontal direction.
- `true` seems better that `middle` (especially if you consider the default value to be `false`)
- In #588, we have not discussed `attached`. As this term already exists in SUI and seems to be valid among several components, what's the reason for not using that?

src/themes/teams/components/Chat/chatItemStyles.ts
mnajdova: Would like to avoid the name selectors as it indicates that we will have the class selector `.` at the end of the string. We may add something like ChatMessage.classes.author, but if we decide in future to have some different selectors (other then the classes, then maybe your suggestion would make more sense for the user... Just like to mention that if we decide with the selectors, would like to specify the authors and timestamp there as `.ui-chat__item__author` and `.ui-chat__item__timestamp` respectively. @kuzhelov @Bugaa92 @layershifter  what is your opinion on this?

mnajdova: Will add a static field called cssSelectors, and will specify the selectors for the author and timestamp there.

src/components/Chat/ChatItem.tsx
mnajdova: Will add the attached prop

src/themes/teams/components/Chat/chatItemStyles.ts
layershifter: My opinion:
We call `timestamp` and `author` as slots. So may be it can be `slotClassNames` field be defined as a static field on the matching component

layershifter: Later we can introduce a test that asserts that all keys that are presented in styles have matching `slotClassName`

mnajdova: Maybe `slotCssSelectors` will be better name, as in future we may have more complex selectors that are not necessarily classNames. And defining the selector will allow the user to always use them consistently without adding the necessary selector for className for example.. What do you think?

layershifter: 1. As user I want to have CSS class names because I should be able to build custom selectors:
```js
`.${ChatItem.slotCssSelectors.author} + .${Another.slotCssSelectors.slot}`
```

2. We have `className` property on each component, it's strange to have `selectors`

src/components/Chat/ChatItem.tsx
kuzhelov: speaking of the design language choices, there are apps that are using `grouped` term to express this concept (Telegram as an example).

Thing that I don't like with attached is the following combination of prop name and `top` value:

```jsx
<ChatItem attached='start' />
```
This is really hard to reason about what does this prop mean for the element, especially when it will occur in the following context:

```jsx
...
<ChatItem attached />
<ChatItem attached='end' />
<ChatItem attached='start' /> // ???
<ChatItem />
...
```

kuzhelov: to me it seems that original version with `grouped` is something that reads better:

```jsx
...
<ChatItem grouped='start' />
<ChatItem grouped />
<ChatItem grouped />
<ChatItem grouped='end' />
<ChatItem />
...
```

mnajdova: Let's go with `attached: boolean | 'top' | 'bottom'` for now, it satisfies the need for three different attached ChatItems which is sufficient for now, and in the future once we will add this prop in other component, it will be much easier for the users to learn to use it... We may consider changing this in the future if we receive complains for it. Can we agree with this?

kuzhelov: yes, sure

miroslavstastny (approved):

bmdalex (approved):

src/components/Chat/ChatItem.tsx
```diff
@@ -17,11 +17,14 @@ import Box from '../Box/Box'
 import { ComponentSlotStylesPrepared } from '../../themes/types'
 
 export interface ChatItemProps extends UIComponentProps, ChildrenComponentProps {
+  /** Attach ChatItem to other content. */
```
kuzhelov, line 20: would suggest to omit component class names in the description - as it should just convey general purpose of the prop. Something like: `Controls item's relation to other chat items`

src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -2,11 +2,15 @@ import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
 import { ChatMessageProps } from '../../../../components/Chat/ChatMessage'
 import { ChatMessageVariables } from './chatMessageVariables'
 import { screenReaderContainerStyles } from '../../../../lib/accessibility/Styles/accessibilityStyles'
+import { pxToRem } from '../../../../lib'
 
 const chatMessageStyles: ComponentSlotStylesInput<ChatMessageProps, ChatMessageVariables> = {
   root: ({ props: p, variables: v }): ICSSInJSStyle => ({
     display: 'inline-block',
-    padding: v.padding,
+    paddingLeft: v.padding,
```
bmdalex, line 10: @mnajdova this is now causing a problem with Control Messages; the padding values are different there so they need to be updated via the `padding` prop; the problem with this change is that `padding` variable only changes the left and right paddings as we're now hardcoding top padding to 8 and bottom padding to 10

fyi @jurokapsiar - Marija, Juraj has more input here and can show you the issue on Embed side

mnajdova: This was change because the paddingsTop and Bottom are different when the messages are grouped, while the left and right padding are always the same. I am not sure why embed are using the ChatMessage component for the Control messages, but sure let's take a look.

PR #768, opened 2019-01-24
chore(package): update Fela to latest

layershifter:
Measures captured with `yarn cross-env PERF=true gulp perf --times=200` 💻 

| Test name | Measure | Before | After | Diff |
| --------- | ------- | ------ | ----- | ---- |
| Attachment.perf.tsx | avg | 7.43 ms | 5.38 ms | -32% |
| Button.perf.tsx | avg | 6.81 ms | 4.32 ms | -44.74% |
| Divider.perf.tsx | avg | 2.85 ms | 2.12 ms | -29.37% |
| Header.perf.tsx | avg | 1.86 ms | 1.83 ms | -1.62% |
| HeaderDescription.perf.tsx | avg | 3.18 ms | 2.68 ms | -17.06% |
| Icon.perf.tsx | avg | 4.06 ms | 3.18 ms | -24.3% |
| Loader.perf.tsx | avg | 4.62 ms | 3.57 ms | -25.64% |

| Test name | Measure | Before | After | Diff |
| --------- | ------- | ------ | ----- | ---- |
| Attachment.perf.tsx | median | 6.29 ms | 3.97 ms | -45.22% |
| Button.perf.tsx | median | 5.61 ms | 3.25 ms | -53.27% |
| Divider.perf.tsx | median | 2.08 ms | 1.6 ms | -26.08% |
| Header.perf.tsx | median | 0.96 ms | 0.9 ms | -6.45% |
| HeaderDescription.perf.tsx | median | 2.05 ms | 1.6 ms | -24.65% |
| Icon.perf.tsx | median | 3.28 ms | 2.43 ms | -29.77% |
| Loader.perf.tsx | median | 3.58 ms | 2.29 ms | -43.95% |

<details>
  <summary>comp.js</summary>

```js
const _ = require('lodash')

_.mixin({
  sortKeysBy: function(obj, comparator) {
    var keys = _.sortBy(_.keys(obj), function(key) {
      return comparator ? comparator(obj[key], key) : key
    })

    return _.zipObject(
      keys,
      _.map(keys, function(key) {
        return obj[key]
      }),
    )
  },
})

const before = _.sortKeysBy(require('./before.json'))
const after = _.sortKeysBy(require('./after.json'))

console.log('| Test name | Measure | Before | After | Diff |')
console.log('| --------- | ------- | ------ | ----- | ---- | ')

const compDiff = (a, b) => {
  const sign = a > b ? '-' : '+'
  const result = 100 * Math.abs((a - b) / ((a + b) / 2))

  return `${sign}${_.floor(result, 2)}%`
}

const logLine = (b, a, testName, measure) => {
  const before = b[measure]
  const after = a[measure]

  console.log(
    `| ${testName} | ${measure} | ${before} ms | ${after} ms | ${compDiff(before, after)} |`,
  )
}

_.forEach(before, ({ actualTime: bResult }, testName) => {
  const aResult = after[testName].actualTime

  logLine(bResult, aResult, testName, 'median')
})
```
</details>


build/webpack.config.perf.ts
```diff
@@ -37,9 +37,7 @@ const webpackConfig: any = {
         exclude: /node_modules/,
         options: {
           useCache: true,
-          configFileName: __PERF__
-            ? paths.base('build/tsconfig.perf.json')
-            : paths.base('build/tsconfig.docs.json'),
+          configFileName: paths.base('build/tsconfig.perf.json'),
```
layershifter, line 40: This config is used only for perf stuff, cruft

build/webpack.config.perf.ts
```diff
@@ -37,9 +37,7 @@ const webpackConfig: any = {
         exclude: /node_modules/,
         options: {
           useCache: true,
-          configFileName: __PERF__
```
kuzhelov, line 40: 👍

kuzhelov (approved):
lets introduce breaking change entry for that in the CHANGELOG, because, in fact, this change implies potential breaking changes in rendering behavior (as Fela's major version is updated)

alinais (approved):

src/components/Provider/ProviderConsumer.tsx
```diff
@@ -16,7 +16,9 @@ export interface ProviderConsumerProps {
 /**
  * The Provider's Consumer is for accessing theme.
  */
-const ProviderConsumer: React.SFC<ProviderConsumerProps> = props => <FelaTheme {...props} />
+const ProviderConsumer: React.SFC<ProviderConsumerProps> = ({ render }) => (
```
mnajdova, line 19: We had prev here spreading all props, not just the render, not sure if there were other props here, but it may be worth taking a look...

PR #770, opened 2019-01-24
feat: remove props from variables function

kuzhelov:
### TODO
- [x] introduce BREAKING CHANGE entry to changelog

-----

We've agreed to move from providing `props` as an argument to variables function - leaving `siteVariables` being the only one provided. This move will make order of styles resolution (as well as list of dependencies used on each stage) much cleaner and reasonable:
```
siteVariables -> variables(siteVariables) -> styles(props, variables, [siteVariables])    
```

### DIFF
--- a/src/themes/teams/components/RadioGroup/radioGroupItemStyles.ts
+++ b/src/themes/teams/components/RadioGroup/radioGroupItemStyles.ts
@@ -29,23 +29,18 @@ const radioStyles: ComponentSlotStylesInput<
     }),
   }),
 
-  icon: ({ props: p, variables: v, theme }): ICSSInJSStyle => {
-    const { siteVariables: siteVars } = theme
+  icon: ({ props: p, variables: v }): ICSSInJSStyle => ({
+    margin: `0 ${pxToRem(10)} 0 0`,
+    color: p.checked ? v.colorChecked : v.color,
+    backgroundColor: p.checked ? v.colorBackgroundChecked : v.colorBackground,
+    borderColor: p.checked ? v.colorBorderChecked : v.colorBorder,
+    outlineColor: v.color,
 
-    return {
-      margin: `0 ${pxToRem(10)} 0 0`,
-      color: p.checked ? v.colorChecked : v.color,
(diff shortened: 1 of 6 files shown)


src/themes/teams/components/RadioGroup/radioGroupItemStyles.ts
```diff
@@ -3,36 +3,49 @@ import {
...
-  icon: ({ props, variables }): ICSSInJSStyle => ({
-    ...(props.isFromKeyboard && {
-      // this creates both inset and outset box shadow that some readers (NVDA) show when radio is not checked but it is focused
-      boxShadow:
-        `0 0 0 ${pxToRem(1)} ${variables.icon.outlineColor},` +
-        `0 0 0 ${pxToRem(2)} ${variables.icon.outlineColor} inset`,
-    }),
-  }),
+  icon: ({ props: p, variables: v, theme }): ICSSInJSStyle => {
+    const { siteVariables: siteVars } = theme
```
kuzhelov: will simplify this expression

src/themes/teams/components/RadioGroup/radioGroupItemStyles.ts
```diff
@@ -3,36 +3,49 @@ import {
...
+      margin: `0 ${pxToRem(10)} 0 0`,
+      color: p.checked ? v.colorChecked : v.color,
+      backgroundColor: p.checked ? v.colorBackgroundChecked : v.colorBackground,
+      borderColor: p.checked ? v.colorBorderChecked : v.colorBorder,
+      outlineColor: siteVars.brand,
+
+      ...(p.isFromKeyboard && {
+        // this creates both inset and outset box shadow that some readers (NVDA) show when radio is not checked but it is focused
+        boxShadow:
+          `0 0 0 ${pxToRem(1)} ${siteVars.brand},` + `0 0 0 ${pxToRem(2)} ${siteVars.brand} inset`,
```
layershifter: Can we move this to variables to omit `siteVars` usage?

layershifter (approved):
LGTM, left one comment

alinais (approved):

PR #771, opened 2019-01-25
fix(a11y): removed redundant a11y code from Dropdown

silviuaavram:
Cleanup getA11yStatusMessage and A11yStatusMessageOptions once updated versions get released in Downshift.

Now these messages are generated by default as a result of https://github.com/paypal/downshift/pull/629


src/components/Dropdown/Dropdown.tsx
```diff
@@ -71,7 +66,7 @@ export interface DropdownProps extends UIComponentProps<DropdownProps, DropdownS
    * Callback that creates custom accessability message for dropdown status change. Involves changes in highlighted item in the list, selection, toggle status.
    * @param {A11yStatusMessageOptions<ShorthandValue>} messageGenerationProps - Object with properties to generate message from. See getA11yStatusMessage from Downshift repo.
```
layershifter: Oops, `@param` should be fixed, too :)

silviuaavram: fixed, thanks!

layershifter (approved):

mnajdova (approved):
Please update the changelog before merging.

jurokapsiar:
nice cleanup. please make sure to have at least one example that shows how the messages can be provided by the consumer for translation purposes.

silviuaavram:
@jurokapsiar I've added an example in French. I wanted to be clear why those methods are needed for (internationalisation).

jurokapsiar:
Cette liste déroulante est fabuleuse!

PR #772, opened 2019-01-25
 fix(List): make `headerMedia` visible for screen readers

### LINKED ISSUE #463: [Selection list] Time is hidden, when item is hovered or focused
##[Selection list] Time is hidden, when item is hovered or focused 

### Steps
1. Hover the list item 
2. Navigate to the list item by keyboard

### Expected Result
Item is highlighted and text is visible.

### Actual Result
1. Item is highlighted but time text get hidden.
It looks like it was introduce by:
https://github.com/stardust-ui/react/pull/278
I think it would be proper behavior if endMedia exists. But if there is no endMedia I cannot find any reason why text get hidden.

2. Another issue is that hidden time is not narrated by reader. But probably first issue need to be address first and then the second one.

### Version
latest master, before version is v0.11.0

### Testcase

layershifter:
Fixes #463.


kolaps33:
Looks good :) 
Tested with Chrome/JAWS, Firefox/NVDA. Time is now narrated :)

kolaps33 (approved):

sophieH29 (approved):

PR #778, opened 2019-01-28
fix(createRenderer): set `devMode` if ENV is not production

layershifter:
Fixes a small regression after #768 with uneditable styles.

---

### Before

image

### After 

image


src/lib/felaRenderer.tsx
```diff
@@ -8,6 +8,7 @@ import rtl from 'fela-plugin-rtl'
 import { Renderer } from '../themes/types'
 
 const createRendererConfig = (options: any = {}) => ({
+  devMode: process.env.NODE_ENV !== 'production',
```
layershifter, line 11: http://fela.js.org/docs/advanced/RendererConfiguration.html

kuzhelov (approved):

PR #780, opened 2019-01-28
fix(Accordion): panel title layout

kuzhelov:
### TODO
- [x] add docs example
- [x] add changelog entry

------

This PR removes empty line rendered for Accordion panel's title whenever custom block element was introduced:

```jsx
const panels = [
  {
    title: { content: <div>What is a point?</div> },
    content: ...
  },
  ...
]

export default () => <Accordion panels={panels} />
```

### Was

image

### Now

image


src/components/Accordion/AccordionTitle.tsx
```diff
@@ -65,15 +66,15 @@ class AccordionTitle extends UIComponent<ReactProps<AccordionTitleProps>, any> {
     const indicatorWithDefaults = indicator === undefined ? {} : indicator
 
     const contentElement = (
-      <>
-        {Indicator.create(indicatorWithDefaults, {
+      <Layout
+        start={Indicator.create(indicatorWithDefaults, {
```
mnajdova, line 70: Maybe is worth adding example so the screens will catch this?

mnajdova (approved):

src/components/Accordion/AccordionTitle.tsx
kuzhelov: yes, was thinking about the same thing, was waiting for the same opinions on that :) will add, thank you!

PR #786, opened 2019-01-29
fix(dropdown): style improvements

bmdalex:
## fix(dropdown): style improvements

This PR covers the following improvements on Dropdown component styles:
- add scroll in the selected items box (see GIF below)
- corner rounding on listbox
- remove weird rounding on the border bottom of the compose box input
- add shadow around listbox
- set and expose `maxHeight` variables for selected items box and listbox

### Scrolling behavior:
screen recording 2019-01-30 at 14 49 03

### Before:
screenshot 2019-01-29 at 15 15 37

### After:
screenshot 2019-01-29 at 15 15 50

## Known issues:
Technical limitation for setting `maxHeight` on the listbox (list of results), see https://github.com/stardust-ui/react/issues/560#issuecomment-458541526

## Later edit:
1. Fixed `maxHeight` issue - accessibility needs to be tested
2. Accessibility tested and looks fine

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -162,6 +162,7 @@ export default class Dropdown extends AutoControlledComponent<
   private buttonRef = React.createRef<HTMLElement>()
   private inputRef = React.createRef<HTMLElement>()
   private listRef = React.createRef<HTMLElement>()
+  private selectedItemsRef = React.createRef<HTMLDivElement>()
 
   static displayName = 'Dropdown'
 
@@ -281,16 +282,21 @@ export default class Dropdown extends AutoControlledComponent<
                   className={cx(`${Dropdown.className}__container`, classes.container)}
                   onClick={multiple ? this.handleContainerClick.bind(this, isOpen) : undefined}
                 >
-                  {multiple && this.renderSelectedItems()}
-                  {search
-                    ? this.renderSearchInput(
-                        accessibilityRootPropsRest,
-                        getInputProps,
-                        highlightedIndex,
-                        selectItemAtIndex,
-                        variables,
-                      )
-                    : this.renderTriggerButton(styles, getToggleButtonProps)}
+                  <div
+                    ref={this.selectedItemsRef}
+                    className={cx(`${Dropdown.className}__selected-items`, classes.selectedItems)}
+                  >
+                    {multiple && this.renderSelectedItems()}
+                    {search
+                      ? this.renderSearchInput(
+                          accessibilityRootPropsRest,
+                          getInputProps,
+                          highlightedIndex,
+                          selectItemAtIndex,
+                          variables,
+                        )
+                      : this.renderTriggerButton(styles, getToggleButtonProps)}
+                  </div>
                   {Indicator.create(toggleIndicator, {
                     defaultProps: {
                       direction: isOpen ? 'top' : 'bottom',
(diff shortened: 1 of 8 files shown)


src/components/Dropdown/Dropdown.tsx
```diff
@@ -39,6 +40,9 @@ import { screenReaderContainerStyles } from '../../lib/accessibility/Styles/acce
 import ListItem from '../List/ListItem'
 
 export interface DropdownProps extends UIComponentProps<DropdownProps, DropdownState> {
+  /** The initial value for the open state. */
+  defaultIsOpen?: boolean
```
layershifter: We just just `open`, without `is` prefix 💭

bmdalex: @layershifter 
will do, but it's gonna be a bit confusing in the code because we're mapping it to `isOpen` prop from `Downshift`, just a heads up

bmdalex: done

src/components/Dropdown/DropdownSearchInput.tsx
```diff
@@ -8,9 +8,6 @@ import { UIComponentProps } from '../../lib/commonPropInterfaces'
 import Input from '../Input/Input'
 
 export interface DropdownSearchInputProps extends UIComponentProps<DropdownSearchInputProps> {
-  /** Informs the search input about an existing toggle button. */
-  hasToggleButton?: boolean
-
```
layershifter, line 19: ❤️

src/components/Dropdown/Dropdown.tsx
```diff
@@ -253,6 +265,10 @@ export default class Dropdown extends AutoControlledComponent<
           selectedItem={search && !multiple ? undefined : null}
           getA11yStatusMessage={getA11yStatusMessage}
           onStateChange={changes => {
+            if (changes.isOpen !== undefined) {
```
silviuaavram: let's just make a new function here, this one got bigger.

silviuaavram (approved):
just one small comment

src/components/Dropdown/DropdownSearchInput.tsx
```diff
@@ -101,6 +101,7 @@ class DropdownSearchInput extends UIComponent<ReactProps<DropdownSearchInputProp
     } = this.props
     return (
       <Input
+        className={classes.root}
```
layershifter: Should we use `styles` there?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -253,6 +265,10 @@ export default class Dropdown extends AutoControlledComponent<
...
+              this.trySetState({ open: changes.isOpen })
```
miroslavstastny: When adding support for controlled mode, shouldn't we also add `onOpenChange` callback similar to what we have in popup?

src/components/Dropdown/Dropdown.tsx
bmdalex: done

src/components/Dropdown/Dropdown.tsx
bmdalex: I am not adding controlled mode anymore because it's not as straightforward as expected, it was causing a regression and the issue might be on Downshift side.

Anyway, it was not needed anyway for this task, I just needed `isOpen` in styles.

Added this as a task to #791 
see: https://github.com/stardust-ui/react/issues/791#issuecomment-459023880

src/components/Dropdown/DropdownSearchInput.tsx
bmdalex: removed altogether

layershifter (approved):

PR #787, opened 2019-01-29
refactor(Portal): allow to use createRef() API in triggerRef

layershifter:
The main goal of this PR allow to use `createRef` API in `triggerRef`.

### DIFF
--- a/src/components/Portal/Portal.tsx
+++ b/src/components/Portal/Portal.tsx
@@ -56,7 +56,7 @@ export interface PortalProps extends ChildrenComponentProps, ContentComponentPro
   /** Accessibility behavior object to apply on trigger node. */
   triggerAccessibility?: TriggerAccessibility
 
-  /** Assigns a passed ref or called with a ref to the trigger node. */
+  /** Sets trigger node to passed ref. */
   triggerRef?: React.Ref<any>
 
   /**
--- a/src/components/Portal/Portal.tsx
+++ b/src/components/Portal/Portal.tsx
@@ -12,6 +12,7 @@ import {
   ContentComponentProps,
   handleRef,
   rtlTextContainer,
+  customPropTypes,
 } from '../../lib'
 import Ref from '../Ref/Ref'
 import PortalInner from './PortalInner'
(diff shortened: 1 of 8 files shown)


miroslavstastny (approved):

src/components/Portal/Portal.tsx
```diff
@@ -194,9 +194,9 @@ class Portal extends AutoControlledComponent<ReactPropsStrict<PortalProps>, Port
 
   private handleDocumentClick = (e: ReactMouseEvent) => {
     if (
-      !this.portalNode || // no portal
-      doesNodeContainClick(this.triggerNode, e) || // event happened in trigger (delegate to trigger handlers)
-      doesNodeContainClick(this.portalNode, e) // event happened in the portal
+      !this.portalRef.current || // no portal
+      doesNodeContainClick(this.triggerRef.current, e) || // event happened in trigger (delegate to trigger handlers)
```
kuzhelov: this method `doesNodeContainClick` is working with DOM modes only. We should preserve original code of this method and just pass DOM nodes there explicitly - this will make contract of this method to be explicit

src/components/Portal/Portal.tsx
```diff
@@ -175,13 +176,12 @@ class Portal extends AutoControlledComponent<ReactPropsStrict<PortalProps>, Port
...
   private handlePortalRef = (portalNode: HTMLElement) => {
-    this.portalNode = portalNode
+    handleRef(this.portalRef, portalNode)
   }
 
   private handleTriggerRef = (triggerNode: HTMLElement) => {
-    this.triggerNode = triggerNode
-
-    _.invoke(this.props, 'triggerRef', triggerNode)
+    handleRef(this.triggerRef, triggerNode)
```
kuzhelov: once again this strikes my cognitive process :) I really think that `setRef` will be much more intuitive name for this method as, essentially, this is what it does

src/components/Portal/Portal.tsx
```diff
@@ -57,9 +58,9 @@ export interface PortalProps extends ChildrenComponentProps, ContentComponentPro
   /**
    * Called with a ref to the trigger node.
    *
-   * @param {JSX.Element} node - Referred node.
+   * @param {HTMLElement} node - Referred node.
```
kuzhelov: this could be removed now

src/components/Portal/Portal.tsx
```diff
@@ -103,7 +104,7 @@ class Portal extends AutoControlledComponent<ReactPropsStrict<PortalProps>, Port
     onUnmount: PropTypes.func,
     open: PropTypes.bool,
     trigger: PropTypes.node,
-    triggerRef: PropTypes.func,
+    triggerRef: PropTypes.oneOfType([PropTypes.func, PropTypes.object]),
```
kuzhelov: probably, this is a good candidate for custom prop types - would expect this to appear in multiple places

kuzhelov (approved):

PR #788, opened 2019-01-29
feat(Icon): mark icons that should be rotated in rtl for teams theme

mnajdova:
This PR adds mechanism for marking icon definitions that should rotate in rtl. Also the following icons are already marked and will be rotated in rtl:
- bullets
- indent
- leave
- outdent
- redo
- send
- undo

**LTR:** 
image

**RTL:**
image

## Question:
How should we handle icons that have text or numbers? For example the numbers of the number-list icon in rtl are mirrored, which is not something we want... Any ideas will be appreciated. - Will be resolved in another PR.


docs/src/examples/components/Icon/Variations/IconExampleSpace.shorthand.tsx
```diff
@@ -4,29 +4,29 @@ import { Icon, Label } from '@stardust-ui/react'
 const IconExampleSpace = () => (
   <div>
     <p>
-      <Label content="Default" />
+      <Label>Default</Label>
```
mnajdova, line 7: This was change because of console errors for invalid dom structure (div inside p element)

miroslavstastny:
> How should we handle icons that have text or numbers? For example the numbers of the number-list icon in rtl are mirrored, which is not something we want... Any ideas will be appreciated.

I think that for these special cases, the Icon component must render different SVG paths in RTL.

mnajdova:
> How should we handle icons that have text or numbers? For example the numbers of the number-list icon in rtl are mirrored, which is not something we want... Any ideas will be appreciated.
> 
> I think that for these special cases, the Icon component must render different SVG paths in RTL.

@miroslavstastny  thanks for your feedback. Do we want to support this somehow, or should this be user's responsibility then? We may add mapping icon name for RLT for this special icons...

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Fixes
 - Fix layout of `Accordion` panel's title @kuzhelov ([#780](https://github.com/stardust-ui/react/pull/780))
 
+### Features
+- Add mechanism for marking icons that should rotate in rtl in teams theme; marked icons: `send`, `bullets`, `leave`, `number-list`, `outdent`, `redo`, `undo`, `send` @mnajdova ([#788](https://github.com/stardust-ui/react/pull/788))
```
kuzhelov: nit: lets use capitalized 'Teams' instead of 'teams'

src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -122,9 +122,11 @@ const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
       ...((bordered || v.borderColor || circular) &&
         getBorderedStyles(circular, v.borderColor || getIconColor(color, v))),
 
-      ...(rtl && {
-        transform: `scaleX(-1) rotate(${-1 * rotate}deg)`,
-      }),
+      ...(rtl &&
+        !isFontBased &&
```
kuzhelov: we may safely omit this `isFontBased` check, as there is an explicit flag required for those icons that need to be flipped

src/themes/teams/components/Icon/svg/types.d.ts
```diff
@@ -2,15 +2,16 @@ import { SvgIconSpec, ComponentSlotStyleFunction } from '../../../../types'
 import { ObjectOf } from '../../../../../../types/utils'
 import { IconProps } from '../../../../../components/Icon/Icon'
 
-type SvgIconSpecWithStyles = {
+type SvgIconSpecWithMetaData = {
```
kuzhelov: what if we will end up with these names:
- `SvgIconSpec` -> `SvgIconPathSpec`
- `SvgIconSpecWithStyles` -> `SvgIconSpec`

It seems that those will be more descriptive and, thus, easier to maintain.

src/themes/teams/withProcessedIcons.ts
```diff
@@ -1,29 +1,38 @@
...
 
-  const icon: SvgIconSpec = (iconAndMaybeStyles as any).styles
-    ? (iconAndMaybeStyles as SvgIconSpecWithStyles).icon
-    : (iconAndMaybeStyles as SvgIconSpec)
+  const icon: SvgIconSpec = (iconAndMaybeMetaData as any).styles
+    ? (iconAndMaybeMetaData as SvgIconSpecWithMetaData).icon
+    : (iconAndMaybeMetaData as SvgIconSpec)
+
+  const rotateInRtl = (iconAndMaybeMetaData as any).rotateInRtl
+    ? (iconAndMaybeMetaData as any).rotateInRtl
```
kuzhelov: do we need this ternary operator? I would expect it being safe to just directly provide the value of `(iconAndMaybeMetaData as any).rotateInRtl` as `rotateInRtl`

kuzhelov:
> @miroslavstastny thanks for your feedback. Do we want to support this somehow, or should this be user's responsibility then? We may add mapping icon name for RLT for this special icons...

would suggest just to pass `rtl` value to icon's path rendering logic (currently we are just passing `classes` there) - and then it will be up to the client's logic to decide how this case should be processed

CHANGELOG.md
mnajdova: Fixed

src/themes/teams/components/Icon/svg/types.d.ts
mnajdova: I reverted the changes in the public API as you suggested and I am adding classes for these icons

src/themes/teams/withProcessedIcons.ts
mnajdova: Have reverted these changes.

mnajdova:
> would suggest just to pass rtl value to icon's path rendering logic (currently we are just passing classes there) - and then it will be up to the client's logic to decide how this case should be processed

@kuzhelov will created this as other PR once we are certain what we want to do in such cases.

docs/src/examples/components/Icon/Variations/IconExampleSpace.shorthand.tsx
notandrew: Is &lt;Label&gt; referring to an HTML element or a Stardust element?

mnajdova: It is Stardust component: https://stardust-ui.github.io/react/components/label

src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -122,16 +122,22 @@ const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
...
-        transform: `scaleX(-1) rotate(${-1 * rotate}deg)`,
-      }),
-
       ...(!rtl && {
         transform: `rotate(${rotate}deg)`,
       }),
     }
   },
 
+  rotateInRtl: ({ props: p, theme }) => {
```
kuzhelov: nit: this may be simplified a bit (given that there is only one prop we would like to consume from theme):
```js
 rotateInRtl: ({ props: p, theme: { rtl } }) => {
```

src/themes/teams/components/Icon/svg/icons/bullets.tsx
```diff
@@ -1,9 +1,15 @@
...
+import cx from 'classnames'
 
 export default {
   icon: ({ classes }) => (
-    <svg role="presentation" focusable="false" viewBox="8 8 16 16" className={classes.svg}>
+    <svg
+      role="presentation"
+      focusable="false"
+      viewBox="8 8 16 16"
+      className={cx(classes.svg, classes.rotateInRtl)}
```
kuzhelov: 👍

src/themes/teams/components/Icon/iconStyles.ts
layershifter:
```suggestion
  rotateInRtl: ({ props: p, theme: { rtl } }) => ({
    ...(rtl && {
      transform: `scaleX(-1) rotate(${-1 * p.rotate}deg)`,
    }),    
  })
```

We can make this more compact 🤔

kuzhelov (approved):

layershifter (commented):
LGTM, only one thing disturbs me, do we have a better name than `rotateInRtl`? 💭

layershifter (approved):

codepretty:
Is there a case where we have one icon that is flipped in some cases, but not others and it just depends on context? I was reading through [Material Design's bidirectionality docs](https://material.io/design/usability/bidirectionality.html#mirroring-elements) and it appears like it will be based on the icon...?

codepretty:
> > How should we handle icons that have text or numbers? For example the numbers of the number-list icon in rtl are mirrored, which is not something we want... Any ideas will be appreciated.
> 
> I think that for these special cases, the Icon component must render different SVG paths in RTL.

I agree. We have icons that will use different paths for different sizes for example. It would be great to think about both cases when you implement a solution for showing different paths for rtl.

mnajdova:
@codepretty currently for the icons we are providing just the classes, but if we realize that we need more context there (props, theme etc), we can add these things. Let's just go step by step and add them only when necessary.

PR #790, opened 2019-01-29
feat(Dialog): add component

layershifter:
This PR implements basic Dialog component.


src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,253 @@
...
+
+    const dialogContent = (
+      <Ref innerRef={this.contentRef}>
+        <ElementType
+          className={classes.root}
+          {...accessibility.attributes.popup}
+          {...accessibility.keyHandlers.popup}
+          {...unhandledProps}
+        >
+          {Box.create(header, {
```
jurokapsiar: Can we use the `Header` component here? By default we should render H2: https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

Also, would it be possible to allow adding aria-labelledby (which needs an ID reference to the header)?
The logic should be:
- by default, generate aria-labelledby
- if aria-labelledby provided by the user, use that instead
- if aria-label provided, do not generate aria-labelledby

For aria-describedby I would leave it fully on the consumer to provide correct referecne as not all dialogs have clear description.

layershifter: Replaced with `Header`, https://github.com/stardust-ui/react/pull/790/commits/fda25b76837c3d35bc3c45b2c15d626f9d95b8ee

miroslavstastny (approved):

src/components/Dialog/Dialog.tsx
layershifter: Created an issue for `aria`s, #812

PR #792, opened 2019-01-29
chore: add RTL examples for visual testing

mnajdova:
This PR adds mechanism for adding rtl examples for visual test. The added test/examples should end with the extension `.rtl.tsx`, and this is all that is necessary from developer's side (See the example for the `DividerExample.rtl.tsx`). Screener then will maximize this example with rtl enabled and will take screenshot of that.


docs/src/components/ExternalExampleLayout.tsx
```diff
@@ -11,38 +11,41 @@ import {
...
-        {({ element, error }) => (
-          <>
-            {element}
-            {/* This block allows to see issues with examples as visual regressions. */}
-            {error && <div style={{ fontSize: '5rem', color: 'red' }}>{error.toString()}</div>}
-          </>
-        )}
-      </SourceRender.Consumer>
-    </SourceRender>
+    <Provider theme={{ rtl: rtl === 'true' ? true : false }}>
```
mnajdova: Not sure whether we should wrap this with div that will have dir='rtl' if rtl is enabled. We have that logic in the ComponentExample component.

docs/src/components/ExternalExampleLayout.tsx
```diff
@@ -26,23 +27,29 @@ const ExternalExampleLayout: any = props => {
...
-          <>
-            {element}
-            {/* This block allows to see issues with examples as visual regressions. */}
-            {error && <div style={{ fontSize: '5rem', color: 'red' }}>{error.toString()}</div>}
-          </>
-        )}
-      </SourceRender.Consumer>
-    </SourceRender>
+    <Provider theme={{ rtl: isRtlEnabled }}>
+      {/*<div dir={isRtlEnabled ? 'rtl' : undefined}>*/}
```
mnajdova: Not sure whether we should add this wrapper div that will have dir='rtl' if rtl is enabled. We have that logic in the ComponentExample component.

alinais (approved):

docs/src/examples/components/Divider/Rtl/DividerExample.rtl.tsx
```diff
@@ -0,0 +1,6 @@
+import * as React from 'react'
+import { Divider } from '@stardust-ui/react'
+
+const DividerExampleShorthand = () => <Divider content="مثال النص" />
+
+export default DividerExampleShorthand
```
layershifter:
```suggestion
export default DividerExampleRtl
```

May be name should be changed?

docs/src/components/ExternalExampleLayout.tsx
layershifter: We definitely need it.

### Without

image


### With

image

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Fixes
 - Fix layout of `Accordion` panel's title @kuzhelov ([#780](https://github.com/stardust-ui/react/pull/780))
 
+### Features
+- Add mechanism for adding rtl visual tests @mnajdova ([#792](https://github.com/stardust-ui/react/pull/792))
```
layershifter: I'm not sure that we should have an entry in `CHANGELOG` about it, there are no any benefits for user from this change

layershifter (approved):

miroslavstastny (commented):
I vote for adding RTL examples to docsite as well (and switch them to RTL by default)

mnajdova:
@miroslavstastny thanks for the suggestion, I've added RTL examples (switch to RTL by default), and now when you maximize an example, if you were seeing the example in RTL the maximize example will also show it in rtl... Please take a look.

docs/src/examples/components/Button/index.tsx
```diff
@@ -13,6 +14,7 @@ const ButtonExamples = () => (
     <Variations />
     <Groups />
     <Usage />
+    <Rtl />
```
miroslavstastny, line 17: The same change is missing in `Divider/index.tsx`

docs/src/routes.tsx
```diff
@@ -18,7 +18,7 @@ import IntegrateCustomComponents from './views/IntegrateCustomComponents'
 const Router = () => (
   <BrowserRouter basename={__BASENAME__}>
     <Switch>
-      <Route exact path="/maximize/:exampleName" component={ExternalExampleLayout} />
+      <Route exact path="/maximize/:exampleName/:rtl?" component={ExternalExampleLayout} />
```
miroslavstastny, line 21: I don't like `/true` and `/false` in the URL, but let's leave it as is for now and iterate on that later

miroslavstastny (commented):
Please also edit `SECTION_ORDER` in `gulp-example-menu.ts` to show RTL below Usage section just above Performance

docs/src/examples/components/Button/index.tsx
mnajdova: Added, thanks!

PR #794, opened 2019-01-29
feat(Animation): support keyframes as functions and add keyframeParams prop

### LINKED ISSUE #760: Keyframes - Animation override for transform values
## Feature Request
I would like to be able to declare an animation and override the duration, delay, and keyframes when needed. This is particularly valuable in the case of transform animations. I would like to create animation templates for basic slide animations (up, down, left & right).  Since we have elements with different heights and widths that slide in and out of view it would be nice to be able to use one slide animation template but override the transform value as needed so we can account for different sizes of elements.

### Problem description
Imagine we have a sidebar with a width of 300px and another sidebar with a width 400px and we want to animate then in and out of view.  Instead of creating different animations for each use case we could use one Slide_In_Left animation and override one of the transform keyframes to account for the difference in element sizes. See example below:

Sidebar 1 animation="Slide_In_Left" keyframe1="-300px" keyframe2="0px"
Sidebar 2 animation="Slide_In_Left" keyframe1="-400px" keyframe2="0px"

### Proposed solution
With this solution we could have a few animation templates that uses our Teams branded animation curves but are also flexible enough to adapt to any need.

mnajdova:
This PR adds keyframeParams in the definition of the animations. Also, keyframes now can be defined as functions, that will receive the keyframeParams defined in the animation itself or as a prop from the user. There is an example of the usage added as part of the PR. This fixes https://github.com/stardust-ui/react/issues/760


docs/src/examples/components/Animation/Types/AnimationExampleKeyframeParams.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from 'react'
+import { Animation, Icon, Provider } from '@stardust-ui/react'
+
+const colorChanger = {
```
kuzhelov: it will be great to ensure types of the `keyframeParams` and input arg of the `keyframe` match each other (within the same animation object)

src/lib/createAnimationStyles.tsx
```diff
@@ -17,12 +17,18 @@ const createAnimationStyles = (animation: AnimationProp, theme: ThemePrepared) =
         iterationCount,
         playState,
         timingFunction,
+        keyframeParams,
       } = animations[animationName]
 
+      const mergedKeyframeParams =
```
kuzhelov, line 26: sorry, not sure that I completely understand that - shouldn't we actually merge the values that were provided by client with the ones that are defined as default ones for the animation object? I might expect this being useful when, for example, there are multiple params that define start and end position of animated element, and client might want to override just one of them (say, just `startPosition`)

miroslavstastny (commented):
Technically it is obvious, but shouldn't we document anywhere that each unique `keyframeParams` will render as a new `@keyframes`?

src/lib/createAnimationStyles.tsx
mnajdova: Valid point, will add this..

mnajdova:
@miroslavstastny that's a great point. Will add info in the examples of the keyframeParams that new keyframe will be added if the params are changed..
Update I added orange Segment stating this before the examples:
image

miroslavstastny (approved):

PR #795, opened 2019-01-29
fix(Avatar): add white border around avatar in contrast theme

bcalvery:
This change adds properties for setting the border for the avatar, which is used in the contrast theme to draw a white border around the image.
Before:
image
After:
image

Known issues:
The example in the docs with a custom image value (using Icon instead of Image) isn't rendering correctly. It seems like the example is perhaps broken, because it doesn't respect avatar size property.

Default size:
image
Both set to 28:
image


alinais:
The issue is that, for the second example we don't update the padding on the icon according to the size of the avatar. It seems to be an issue on our side.

alinais (approved):

docs/src/examples/components/Avatar/Variations/AvatarExampleImageCustomization.shorthand.tsx
```diff
@@ -10,6 +10,8 @@ const AvatarExampleImageCustomizationShorthand = () => (
     &emsp;
     <Avatar
       image={
+        // This example does not react to the avatar size variable
```
kuzhelov, line 13: 👍 thanks for spotting this!

src/themes/teams-high-contrast/components/Avatar/avatarVariables.ts
```diff
@@ -2,5 +2,7 @@ import { AvatarVariables } from '../../../teams/components/Avatar/avatarVariable
 import { Partial } from 'types/utils'
 
 export default (siteVariables: any): Partial<AvatarVariables> => ({
+  avatarBorderColor: 'white',
```
kuzhelov: would suggest to use color from `siteVariables` here (i.e. `siteVariables.white`), for the sake of consistency - or let me know if there any reason to not use it

kuzhelov (approved):
please, just take a look on this comment before merge: https://github.com/stardust-ui/react/pull/795/files#r252747086 . Thank you!

src/themes/teams-high-contrast/components/Avatar/avatarVariables.ts
bcalvery: Good catch. Thanks!

PR #797, opened 2019-01-30
feat(screener): add screener tests with steps

silviuaavram:
Edited `screener.config.js` to check for test files that have tests with steps. Also edited the `states` property gets created.

Added shorthand for `triggerButton` in `Dropdown` so I can add a test id attribute to it, for better and safer selection in screener (and future e2e) tests. Edit: if we decide to use className, will leave the shorthand for now, it is still useful.


test/screener/tests/dropdown.js
```diff
@@ -0,0 +1,14 @@
+const Steps = require('screener-runner/src/steps')
+const selector = require('../utils').selector
+
+module.exports = {
+  'DropdownExampleSingleSelection.shorthand': [
+    {
+      name: 'Shows List',
+      steps: new Steps()
+        .click(selector('trigger-button'))
```
layershifter: Let's use `className` from them matching slot (i.e. `ui-dropdown__trigger-button`), this will allow to avoid changes in examples

layershifter (commented):
As we already have [`glob`](https://www.npmjs.com/package/glob) in dependencies I have a proposal to place steps files near with examples, i.e:
```
DropdownExampleSingleSelection.shorthand.tsx
DropdownExampleSingleSelection.steps.js
```

Then resolve these files via `glob` and use them, this will also allow to simplify structure of file:
```js
module.exports = (Steps) => ([
  {
     name: 'Shows List',
     steps: new Steps()
       .click(selector('CLASSNAME'))
       .snapshot('Snapshot')
       .end(),
  },
])
```

docs/src/examples/components/Dropdown/Types/DropdownExampleSingleSelection.shorthand.tsx
```diff
@@ -20,6 +20,9 @@ const DropdownExample = () => (
     }}
     placeholder="Select your hero"
     items={inputItems}
+    triggerButton={{
+      'data-testid': 'trigger-button',
+    }}
```
layershifter: Cruft?

screener.config.js
```diff
@@ -28,15 +30,28 @@ const screenerConfig = {
   // screenshot every example in maximized mode
   states: glob
     .sync('docs/src/examples/**/*.tsx', { ignore: ['**/index.tsx', '**/*.knobs.tsx'] })
-    .map(examplePath => {
-      const { name: nameWithoutExtension, base: nameWithExtension } = path.parse(examplePath)
+    .reduce((states, examplePath) => {
+      const { name: nameWithoutExtension, base: nameWithExtension, dir } = path.parse(examplePath)
       const rtl = nameWithExtension.endsWith('.rtl.tsx')
+      const url = `http://localhost:8080/maximize/${_.kebabCase(nameWithoutExtension)}/${rtl}`
```
layershifter, line 36:
```suggestion
      const url = `http://${config.server_host}:${config.server-host}/maximize/${_.kebabCase(nameWithoutExtension)}/${rtl}`
```

We have this vars in config, let's use them

layershifter (approved):
LGTM except two comments, fix them please before merge 🐱 

Also it will be nice to hear feedback from @kuzhelov before merge

docs/src/examples/components/Dropdown/Types/DropdownExampleSingleSelection.shorthand.steps.js
```diff
@@ -0,0 +1,7 @@
+module.exports = [
+  steps => steps.click('.ui-dropdown .ui-button').snapshot('Shows list'),
```
bmdalex, line 2: should we use `Dropdown.className` and `Button.className` to avoid regressions when if these values change.
Same for other occurrences

bmdalex (approved):
looks good, one small comment

mnajdova (approved):
🥇 Let's just create issue for the problems with using JS files here, as we are aware at this moment that we have problems when we want to import something from TS files.

docs/src/examples/components/Dropdown/Types/DropdownExampleSingleSelection.shorthand.steps.js
kuzhelov, line 2: please, do not use hardcoded class names - use string interpolation and `className` prop of corresponding component instead. This will ensure tests' robustness

screener.config.js
```diff
@@ -28,15 +30,28 @@ const screenerConfig = {
   // screenshot every example in maximized mode
   states: glob
```
kuzhelov: please, introduce reference to the corresponding Screener docs file in the comment

screener.config.js
```diff
@@ -28,15 +30,28 @@ const screenerConfig = {
...
+      const url = `http://localhost:8080/maximize/${_.kebabCase(nameWithoutExtension)}/${rtl}`
 
-      return {
-        url: `http://localhost:8080/maximize/${_.kebabCase(nameWithoutExtension)}/${rtl}`,
+      states.push({
+        url,
         name: nameWithExtension,
-      }
-    }),
+        steps: fs.existsSync(`${dir}/${nameWithoutExtension}.steps.js`)
```
kuzhelov, line 42: please, introduce reference to the corresponding Screener docs file in the comment

screener.config.js
```diff
@@ -28,15 +30,28 @@ const screenerConfig = {
...
+      })
+
+      return states
+    }, []),
+}
+
+function getSteps(dir, nameWithoutExtension) {
+  const stepTests = require(`./${dir}/${nameWithoutExtension}.steps`)
+  return stepTests.reduce((stepsAcc, steps) => {
+    return steps(stepsAcc)
```
kuzhelov: nit: this one can be simplified

```js
return stepTests.reduce(
  (stepsAcc, steps) => steps(stepsAcc), 
  new Steps()
)
```

kuzhelov (approved):

docs/src/examples/components/Dropdown/Types/DropdownExampleSingleSelection.shorthand.steps.js
silviuaavram: will take care of this when we refactor for TS

PR #802, opened 2019-01-30
feat: Flex component

kuzhelov:
# Principles
- minimal markup necessary
- readable markup (visual effect should be inferable by reading the markup)
- minimal set of props in vocabulary, with large coverage of use cases
- no props with conflicting effects
- no leaky abstractions
- restricted set of size values (for `gap`, `size`, `padding` props) that should follow foreseen approach to design tokens' names

# Layout example sources used
- typical use cases: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Typical_Use_Cases_of_Flexbox
- http://styleguides.io/

# Important cases to cover
- control item size ratio along main axis: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Controlling_Ratios_of_Flex_Items_Along_the_Main_Ax
- address margins problem on wrap: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Mastering_Wrapping_of_Flex_Items

# API and Vocabulary

### `Flex`
Container element.
- direction
  - `column: boolean`
- align items
  - `hAlign: 'start' | 'center' | 'end' | 'stretch'`,
  - `vAlign: ...`
  - no boolean flag props with conflicting effects (`center`, `top`, ...) - as those will make it impossible to override related effects: `<Flex {...} center /> - there is no guarantee that center alignment will be applied`
- arrange space between items
  - `space: 'between' | 'around' | 'evenly' | ... `
- gap between items
  - `gap: 'gap.small' | 'gap.medium' | 'gap.large'`
  -  the set of values is intentionally limited at the moment, values should be introduced as long as they will be needed
- `inline: boolean`
- `wrap: boolean` - allows items to wrap on the next line
- *`padding: 'padding.medium'`
  - the set of values is intentionally limited at the moment, values should be introduced as long as they will be needed

### `Flex.Item` 
Should be sparingly used, only in cases where styles for individual item should be introduced.

- item size (along flow direction)
  - `size:  'size.small' | 'size.medium' | 'size.half' | 'size.quarter'`
  - `grow: boolean | ValueOf<FlexGrow>`
  - `shrink: boolean | ValueOf<FlexShrink>`
- item alignment (CROSS axis)
  - `align: 'start | center | end | ... '`
- push
  - push item from the rest of content (along MAIN axis)
  - `push: boolean`

# Q&A 
- `Flex` or `Row` (or other name)?
  - at the moment we've agreed to stay with `Flex` and get on par with CSS Flex capabilities (while intentionally limiting some features, such as reordering of child DOM elements)
- prop names
  - provided prop names were aimed to largely mimic the ones provided by Flex CSS spec. At times it might seem that there are clear better naming alternatives (that would make markup more readable), however, we have deferred this renaming till the point when component itself won't suggest (by its name) its close correspondence to Flex.
- semantic names (that aim to match design spec) for some prop values
  - `gap` and `padding` of `Flex`
  - `size` of `Flex.Item`
  - the set of values introduced for each of them now is intentionally limited - it will be updated and, ultimately, rebuilt into common design language spec
- names for semantic prop values ('gap.medium', 'gap.small' - as string keys)
  - it was a conscious decision to provide explicit string keys as prop values - this move, although not being an ideal and most succinct choice for now, will highly alleviate further upcoming variables refactoring, when all the variables will be arranged under common design language schema


docs/src/examples/components/Flex/Types/index.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react'
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
+
+const Types = () => (
+  <ExampleSection title="Types">
+    <ComponentExample
```
kuzhelov, line 7: the set of examples will be rearranged and refined

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,121 @@
...
+  fill?: boolean
+}
+
+/**
+ * Arrange group of items aligned towards common direction.
+ */
+class Flex extends UIComponent<ReactProps<FlexProps>, any> {
+  static Item: typeof FlexItem
+
+  static Gap: any
```
kuzhelov: will take care of type for this guy

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,121 @@
...
+  debug?: boolean
+
+  /** Orders container to fill all parent's space available. */
+  fill?: boolean
+}
+
+/**
+ * Arrange group of items aligned towards common direction.
+ */
+class Flex extends UIComponent<ReactProps<FlexProps>, any> {
```
layershifter:
```suggestion
class Flex extends UIComponent<ReactProps<FlexProps>> {
```

docs/src/examples/components/Flex/index.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
+import Types from './Types'
+
+const FlexExamples = () => (
+  <div>
```
layershifter:
```suggestion
  <>
```

docs/src/examples/components/Flex/index.tsx
```diff
@@ -0,0 +1,10 @@
...
+    <Types />
+  </div>
```
layershifter:
```suggestion
  </>
```

packages/react/src/components/Flex/FlexItem.tsx
```diff
@@ -0,0 +1,105 @@
...
+      })
+    }
+
+    return withStyles(React.Children.only(children), styles, classes)
+  }
+}
+
+export default FlexItem
+
+const withStyles = (element: React.ReactElement<any>, styles, classes): any => {
```
layershifter: Can we choose another name for `withStyles`? The `with` prefix has the strong association with HOCs.

```suggestion
const applyStyles = (element: React.ReactElement<any>, styles, classes): any => {
```

docs/src/examples/components/Flex/Types/FlexColumnsExample.shorthand.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <Flex.Item size="size.half">
+        <Segment content="1/2" />
+      </Flex.Item>
+
+      <Flex.Item size="size.quarter">
+        <Segment content="1/4" />
+      </Flex.Item>
+    </Flex>
+
+    <Flex gap="gap.small" padding="padding.medium" style={{ minHeight: 200 }}>
```
layershifter: We should not confuse users with the `style` prop on Stardust elements, however I don't have a proposal there

docs/src/examples/components/Flex/Types/FlexColumnsExample.shorthand.tsx
```diff
@@ -0,0 +1,42 @@
...
+      </Flex.Item>
+
+      <Flex.Item size="size.half">
+        <Segment content="Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum mollis velit non gravida venenatis. Praesent consequat lectus purus, ut scelerisque velit condimentum eu. Maecenas sagittis ante ut turpis varius interdum. Quisque tellus ipsum, eleifend non ipsum id, suscipit ultricies neque." />
+      </Flex.Item>
+    </Flex>
+  </>
+)
+
+export default FlexExampleShorthand
```
layershifter:
```suggestion
export default FlexExampleColumns
```

docs/src/examples/components/Flex/Types/FlexExampleInput.shorthand.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react'
+import { Flex, Input, Button, Label } from '@stardust-ui/react'
+
+const FlexExampleShorthand = () => (
+  <>
```
layershifter: `Fragment` is redundant there:
```suggestion
```

docs/src/examples/components/Flex/Types/FlexExampleInput.shorthand.tsx
```diff
@@ -0,0 +1,25 @@
...
+          </Flex.Item>
+        </Flex>
+      </Flex.Item>
+
+      <Button content="Load" />
+    </Flex>
+  </>
+)
+
+export default FlexExampleShorthand
```
layershifter:
```suggestion
export default FlexExampleInput
```

docs/src/examples/components/Flex/Types/FlexColumnsExample.shorthand.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react'
```
layershifter: Please rename this file to `FlexExampleColumns.shorthand.tsx`

docs/src/examples/components/Flex/Types/FlexExampleNavMenu.shorthand.tsx
```diff
@@ -0,0 +1,19 @@
...
+        <Button content="Page 1" />
+      </Flex.Item>
+
+      <Button content="Page 2" />
+      <Button content="Page 3" />
+    </Flex>
+  </>
+)
+
+export default FlexExampleShorthand
```
layershifter:
```suggestion
export default FlexExampleNavMenu
```

docs/src/examples/components/Flex/Types/FlexExampleNavMenu.shorthand.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react'
+import { Flex, Button } from '@stardust-ui/react'
+
+const FlexExampleShorthand = () => (
+  <>
```
layershifter:
```suggestion
```

docs/src/examples/components/Flex/Types/FlexItemsAlignmentOptionsExample.shorthand.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react'
+import { Flex, Segment } from '@stardust-ui/react'
+
+const FlexExampleShorthand = () => (
+  <>
```
layershifter:
```suggestion
```

docs/src/examples/components/Flex/Types/FlexItemsAlignmentOptionsExample.shorthand.tsx
```diff
@@ -0,0 +1,40 @@
...
+              <Segment styles={{ width: '30px', height: '30px' }} />
+            </Flex>
+          ))}
+        </Flex>
+      ))}
+    </Flex>
+  </>
+)
+
+export default FlexExampleShorthand
```
layershifter:
```suggestion
export default FlexExampleItemsAlignmentOptions
```

docs/src/examples/components/Flex/Types/FlexItemsAlignmentOptionsExample.shorthand.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react'
```
layershifter: Please rename this file to match: `FlexExample*`

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,119 @@
...
+    debug: PropTypes.bool,
+  }
+
+  renderChildren = gapClasses => {
+    const { column, gap, children } = this.props
+
+    let isFirst = true
+    return React.Children.map(children, (child: React.ReactElement<any>) => {
+      const childElement =
+        child.type && ((child.type as any) as typeof FlexItem).__isFlexItem
```
layershifter, line 95: This can be broken with `react-hot-loader`:
https://github.com/gaearon/react-hot-loader/tree/v4.6.5#checking-element-types
https://github.com/gaearon/react-hot-loader/blob/v4.6.5/src/utils.dev.js#L14

`__isFlexItem` is a static property, so it can be passed correctly, but please check it

---

We can also try to use `react-is`:
```suggestion
        ReactIs.isElement(child) && ((child.type as any) as typeof FlexItem).__isFlexItem
```

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,119 @@
...
+    fill: PropTypes.bool,
+
+    debug: PropTypes.bool,
+  }
+
+  renderChildren = gapClasses => {
+    const { column, gap, children } = this.props
+
+    let isFirst = true
+    return React.Children.map(children, (child: React.ReactElement<any>) => {
```
layershifter:
```suggestion
    return React.Children.map(children, (child: React.ReactElement<any>, index) => {
```

We can use `index === 0` instead of `isFirst`

packages/react/src/themes/teams/components/Flex/utils.ts
```diff
@@ -0,0 +1,15 @@
+export const toFlexAlignment = propValue => {
```
layershifter:
```suggestion
export const toFlexAlignment = (propValue: string) => { // ???
```

Can we define type here explicitly?

packages/react/src/themes/teams/components/Flex/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+  }
+
+  return trimmedValue
+}
+
+export const toFlexItemSizeValues = sizeValue => {
+  return {
+    flexBasis: sizeValue,
+  }
+}
```
layershifter:
```ts
export const toFlexItemSizeValues = (sizeValue: TYPE) => ({
  flexBasis: sizeValue,
})
```

Can be a little simplified

docs/src/examples/components/Flex/Types/FlexColumnsExample.shorthand.tsx
kuzhelov: agree - let me try to address it

packages/react/src/components/Flex/Flex.tsx
kuzhelov: yes - and this is exactly why the flag-based approach is used, instead of just trying to compare types directly: as whoever will be the intermediate that would introduce proxied types (like `react-hot-loader`) - irregardless of this fact it is safe to assume that all the type's props will return the same values as original type (`react-hot-loader`, surely, adheres to that). 

Thus, in this aspect it should be pretty robust approach

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,116 @@
...
+  inline?: boolean
+
+  /** Sets vertical flow direction. */
+  column?: boolean
+
+  /** Allows overflow items to wrap on the next container's line. */
+  wrap?: boolean
+
+  /** Controls items alignment in horizontal direction. */
+  hAlign?: 'start' | 'center' | 'end' | 'stretch'
```
bmdalex, line 22: can we extract these in types as well as we did for `FlexGap` and `FlexPadding`?

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  /** Controls items alignment in vertical direction. */
+  vAlign?: 'start' | 'center' | 'end' | 'stretch'
```
bmdalex, line 25: same as above

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+  /** Defines strategy for distributing remaining space between items. */
+  space?: 'around' | 'between' | 'evenly'
```
bmdalex, line 28: same as above

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,116 @@
...
+    wrap: PropTypes.bool,
+
+    hAlign: PropTypes.oneOf(['start', 'center', 'end', 'stretch']),
+    vAlign: PropTypes.oneOf(['start', 'center', 'end', 'stretch']),
+
+    space: PropTypes.oneOf(['around', 'between', 'evenly']),
+
+    gap: PropTypes.string,
+
+    padding: PropTypes.string,
```
bmdalex: `gap` and `padding` are strings here but we enforce a different type in the interface, aka `FlexGap` and `FlexPadding`; why is that?

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,116 @@
...
+    fill: PropTypes.bool,
+
+    debug: PropTypes.bool,
+  }
+
+  renderChildren = gapClasses => {
```
bmdalex: private method, also should come after public ones, also type for `gapClasses`?

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,116 @@
...
+        child.type && ((child.type as any) as typeof FlexItem).__isFlexItem
+          ? React.cloneElement(child, {
+              flexDirection: column ? 'column' : 'row',
+            })
+          : child
+
+      const renderGap = index !== 0
+      return (
+        <>
+          {renderGap && gap && <Flex.Gap className={gapClasses} />}
```
bmdalex: the `gap` logic can be done only once before the loop;

```typescript
private renderChild: (
  childElement: React.ReactChild,
  renderGap: boolean,
  gapClasses: any // type for classes
) => React.ReactChild = this.props.gap
  ? (childElement, renderGap, gapClasses) => (
      <>
        {renderGap && <Flex.Gap className={gapClasses} />}
        {childElement}
      </>
    )
  : childElement => childElement
```

and use it as:
```typescript
return this.renderChild(childElement, index !== 0, gapClasses)
```
in your React.Children.map function?

Looking at it again it reduces readability a bit so feel free to disregard..

packages/react/src/themes/teams/components/Flex/flexItemVariables.ts
```diff
@@ -0,0 +1,15 @@
...
+
+type SizeValues = { [Key in FlexItemSize]: string }
+
+export type FlexItemVariables = SizeValues
+
+export default (): FlexItemVariables => ({
+  'size.half': '50%',
+  'size.quarter': '25%',
+
+  'size.small': pxToRem(150),
```
bmdalex, line 13: how did we come up with these values?
can we make these part of theme somehow or is that plan for future?

packages/react/src/themes/teams/components/Flex/flexItemStyles.ts
```diff
@@ -0,0 +1,26 @@
...
+import { FlexItemProps } from '../../../../components/Flex/FlexItem'
+
+import { toFlexAlignment, toFlexItemSizeValues } from './utils'
+
+const flexItemStyles: ComponentSlotStylesInput<FlexItemProps, any> = {
+  root: ({ props: p, variables: v }) => {
+    return {
+      ...(p.align && { alignSelf: toFlexAlignment(p.align) }),
+
+      ...(p.size && toFlexItemSizeValues(v[p.size])),
```
bmdalex, line 12: very interesting pattern, prop value as variable key 👍

docs/src/examples/components/Flex/Types/FlexExampleColumns.shorthand.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <Flex.Item size="size.half">
+        <Segment content="1/2" />
+      </Flex.Item>
+
+      <Flex.Item size="size.quarter">
+        <Segment content="1/4" />
+      </Flex.Item>
+    </Flex>
+
+    <Flex gap="gap.small" padding="padding.medium" style={{ minHeight: 200 }}>
```
bmdalex, line 30: what about having `height` as prop? I imagine it was considered already? Disregard in that case

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,116 @@
...
+    return (
+      <ElementType className={classes.root} {...unhandledProps}>
+        {this.renderChildren(classes.gap)}
+      </ElementType>
+    )
+  }
+}
+
+Flex.Item = FlexItem
+Flex.Gap = ({ className }) => <div className={className} />
```
layershifter: Please consider to move it to a separate component and add a `displayName`:

image

packages/react/src/themes/teams/components/Flex/flexItemStyles.ts
```diff
@@ -0,0 +1,26 @@
+import { ComponentSlotStylesInput } from '../../../types'
+import { FlexItemProps } from '../../../../components/Flex/FlexItem'
+
+import { toFlexAlignment, toFlexItemSizeValues } from './utils'
+
+const flexItemStyles: ComponentSlotStylesInput<FlexItemProps, any> = {
```
layershifter:
```suggestion
const flexItemStyles: ComponentSlotStylesInput<FlexItemProps, FlexItemVariables> = {
```

packages/react/src/themes/teams/components/Flex/flexItemVariables.ts
```diff
@@ -0,0 +1,15 @@
+import { pxToRem } from '../../../../lib'
+
+import { FlexItemSize } from '../../../../components/Flex/FlexItem'
+
+type SizeValues = { [Key in FlexItemSize]: string }
```
layershifter:
```suggestion
type SizeValues = Record<FlexItemSize, string>
```

packages/react/src/themes/teams/components/Flex/flexStyles.ts
```diff
@@ -0,0 +1,46 @@
+import { ComponentSlotStylesInput } from '../../../types'
+import { FlexProps } from '../../../../components/Flex/Flex'
+
+import { toFlexAlignment } from './utils'
+
+const flexStyles: ComponentSlotStylesInput<FlexProps, any> = {
```
layershifter:
```suggestion
const flexStyles: ComponentSlotStylesInput<FlexProps, FlexVariables> = {
```

packages/react/src/themes/teams/components/Flex/flexStyles.ts
```diff
@@ -0,0 +1,46 @@
...
+  root: ({ props: p, variables: v }) => {
+    return {
```
layershifter: Can be simplied:

```diff
-root: ({ props: p, variables: v }) => {
-  {
+root: ({ props: p, variables: v }) => ({

File: packages/react/src/themes/teams/components/Flex/flexVariables.ts [hunk 35]
```diff
@@ -0,0 +1,17 @@
+import { pxToRem } from '../../../../lib'
+import { FlexGap, FlexPadding } from '../../../../components/Flex/Flex'
+
+type GapValues = { [Key in FlexGap]: string }
+type PaddingValues = { [Key in FlexPadding]: string }
```
layershifter: We can use `Record` there, too 😺

layershifter (approved):
Left small comments, overall - ❤️ 

---

Don't forget about a changelog entry 👍

bmdalex (commented):
very nice job, pls take a look at my comments

docs/src/examples/components/Flex/Types/FlexExampleColumns.shorthand.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react'
+import { Flex, Segment } from '@stardust-ui/react'
+
+const FlexExampleColumns = () => (
+  <>
+    <Flex gap="gap.small" padding="padding.medium">
```
mnajdova, line 6: What is the reason for defining string options containing dot? At first I thought that `gap`, `padding` and `size` are some variables.

docs/src/examples/components/Flex/Types/FlexExampleItemsAlignment.shorthand.tsx
```diff
@@ -0,0 +1,38 @@
...
+        { hAlign: 'end', vAlign: 'center' },
+        { hAlign: 'end', vAlign: 'end' },
+      ],
+    ].map(rowOfAlignmentProps => (
+      <Flex gap="gap.large">
+        {rowOfAlignmentProps.map((alignmentProps: any) => (
+          <Flex
+            inline
+            {...alignmentProps}
+            style={{ width: '100px', height: '100px', border: '1px dashed grey' }}
```
mnajdova: Not sure but, can we get rid of the border if we are using the debug prop?

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,116 @@
+import * as PropTypes from 'prop-types'
+import * as React from 'react'
+import * as _ from 'lodash'
+
+import { UIComponent, commonPropTypes } from '../../lib'
+import { ReactProps } from '../../types'
+import FlexItem from './FlexItem'
+
+export type FlexGap = 'gap.small' | 'gap.medium' | 'gap.large'
+export type FlexPadding = 'padding.medium'
```
mnajdova: Why we have just one option here. Moreover why is the value medium? It doesn't make sense without any other values in my opinion.

packages/react/src/components/Flex/FlexItem.tsx
```diff
@@ -0,0 +1,109 @@
+import * as React from 'react'
+import * as PropTypes from 'prop-types'
+import cx from 'classnames'
+import { UIComponent, commonPropTypes } from '../../lib'
+import { mergeStyles } from '../../lib/mergeThemes'
+import { Extendable } from '../../types'
+
+export type FlexItemSize = 'size.half' | 'size.quarter' | 'size.small' | 'size.medium'
```
mnajdova: How did we ended up with this values for size? Again confused, why we have values like small and medium, but not big? What these values mean?

packages/react/src/components/Flex/FlexItem.tsx
```diff
@@ -0,0 +1,109 @@
...
+  static className = 'ui-flex__item'
+
+  static displayName = 'FlexItem'
+
+  static propTypes = {
+    ...commonPropTypes.createCommon({
+      content: false,
+    }),
+    align: PropTypes.oneOf(['auto', 'start', 'end', 'center', 'baseline', 'stretch']),
+    size: PropTypes.string,
```
mnajdova: Why not more strict prop type here?

docs/src/examples/components/Flex/Types/FlexExampleColumns.shorthand.tsx
bmdalex: @mnajdova because the prop values need to be unique through the `Flex` component; @kuzhelov is matching their values with variables keys to achieve styling, see https://github.com/stardust-ui/react/pull/802#discussion_r254673039

kuzhelov: acknowledge the issue - yes, it was done exactly for the reasons @Bugaa92 has mentioned. However, indeed, we might do better here - let me address this moment. Thank you!

docs/src/examples/components/Flex/Types/FlexExampleColumns.shorthand.tsx
kuzhelov: yes, actually, it was - and the main reason for not doing it (for now, at least) is that `height` wasn't seen as a close attribute of the layout aspects - in other case, this prop is not specific to layout component (the same way then we might argue about adding `height` to API of Segment). 

Also, note that if Flex needs to be provided with some height while being part of some other component, this will be done by styles, not by the props API:

```jsx
renderComponent = ({ ..., ElementType, styles }) => {
   <ElementType>
       <Flex styles={styles.flex} >
        ....
      </Flex>
  </ElementType>
}

/// myComponentStyles.js
{
  root: ...
  flex: {
    height: '200px'
  }
}
```

packages/react/src/components/Flex/Flex.tsx
kuzhelov: agree  - but lets refine the set of `padding` values further down the road. What I am not sure at the moment is whether `padding` should be specified this way, as it is a 2D thing - what if, for instance, we would need to make vertical padding lesser than horizontal one, as it is usually the case for button? `medium` name was selected just to be consistent with the dictionary we've used so far for size values.

With that being said, I acknowledge the issue, but it was intentionally left visible, as the approach to organise these values (by putting them under general design schema) needs to be introduced as a follow-up work.

packages/react/src/components/Flex/Flex.tsx
kuzhelov: actually, I would rather do other way round - the reason is that introducing types for `hAlign` is seems to be a redundant thing. There is a way to eliminate types of `FlexGap` and `FlexPadding` as well, however - with that we will completely remove redundancy in the types declared

packages/react/src/components/Flex/FlexItem.tsx
kuzhelov: for now just those that were used on the examples were introduced. There is no spec on the values now, it was rather about approach to values introduced for now. However, just to maintain consistency with the rest of the API, lets introduce the same range of size values (`smallest` -> `largest`)

packages/react/src/themes/teams/components/Flex/flexItemVariables.ts
kuzhelov: it is definitely a plan for the future - there are no clear specs on that yet, provided numbers are free to be changed

packages/react/src/themes/teams/components/Flex/flexVariables.ts
kuzhelov: will change but, frankly, this type declaration seems to be more expressive to me - each time I see
```ts
Record<FlexGap, string >
```
it takes me some though effort to rearrange in my brain how resulting type will be composed.

packages/react/src/components/Flex/FlexItem.tsx
kuzhelov: dev reasons - would like to omit duplication before final type was settled, Thanks!

packages/react/src/components/Flex/Flex.tsx
kuzhelov: yes, will be addressed 👍

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -0,0 +1,112 @@
...
+        child.type && ((child.type as any) as typeof FlexItem).__isFlexItem
+          ? React.cloneElement(child, {
+              flexDirection: column ? 'column' : 'row',
+            })
+          : child
+
+      const renderGap = index !== 0
+      return (
+        <>
+          {renderGap && gap && <Flex.Gap className={cx(`${Flex.className}__gap`, gapClasses)} />}
```
mnajdova, line 104: Please add slotClassNames static field in the Flex component and export this className.

packages/react/src/components/Flex/FlexGap.tsx
```diff
@@ -0,0 +1,6 @@
+import * as React from 'react'
+
+const FlexGap: React.FC<any> = ({ className }) => <div className={className} />
+FlexGap.displayName = 'FlexGap'
```
mnajdova, line 4: Or maybe we can define the className for the gap here?

packages/react/src/components/Flex/FlexItem.tsx
```diff
@@ -0,0 +1,107 @@
...
+  // if element is DOM element
+  if (typeof element.type === 'string') {
+    return React.cloneElement(element, {
+      className: cx(element.props.className, classes.root),
+    })
+  }
+
+  // assuming element is Stardust element
+  return React.cloneElement(element, {
+    styles: mergeStyles(element.props.styles || {}, styles.root),
```
mnajdova: Is the `element.props.styles` suppose to be last, so that they can win over the `styles.root`?

packages/react/src/index.ts
```diff
@@ -58,6 +58,9 @@ export {
   DropdownSearchInputProps,
 } from './components/Dropdown/DropdownSearchInput'
 
+export { default as Flex } from './components/Flex/Flex'
```
mnajdova: Oops, missing Prop interfaces, possible other types defined in the files.

packages/react/src/components/Flex/Flex.tsx
kuzhelov: actually, haven't done it for reason - we won't like this thing to be selectable on the client side (and, thus, introduce any kind of selectors for it). More than that, this component will be removed in the future, when negative-margins-based approach will be introduced to handle gaps (instead of populating gap elements to the DOM tree)

packages/react/src/components/Flex/FlexGap.tsx
kuzhelov: unfortunately, this would require to make it `UIComponent` (as it contains logic to handle class names), which it shouldn't be

mnajdova (approved):

PR #803, opened 2019-01-30
fix: preserve outside click subscription for JS frame that processes component update

### LINKED ISSUE #649: The subscription for EventStack should be refactored in the Popup and MenuItem components
## Bug Report

@kuzhelov: The implementation used in the Popup and the MenuItem for subscribing to EventStack has one subtle problem: although it solves the issue of item being opened and closed within the same frame, it also opens possibility for the following behavior:

MenuItem has opened submenu, this opened prop value doesn't change anyhow for the next JS frame.
///------ JS FRAME --------////

render method was called (but submenuOpen wasn't changed changed anyhow, it is still true)

updateOutsideClickSubscription

this.outsideClickSubscription.unsubscribe(), no subscription on outside click (!!)
setTimeout(() => subscription will be made on the next frame)
... some other logic ..

note, there is no click subscription at this moment, but it should be (!!)
////---------------------------////

Would suggest to leave a fix for a separate PR, for both components.

Reference: https://github.com/stardust-ui/react/pull/539#discussion_r242148547

kuzhelov:
### TODO
- [x] update CHANGELOG
------

Fixes #649 .

This PR ensures that `outsideClick` subscription stays enabled for JS frame that processed component updates. Fix is introduced for `MenuItem` and `Popup` components - essentially, for those where `setInterval` function has to be used to simulate browser's behavior of `addEventListener`.


mnajdova (approved):
LGTM

PR #806, opened 2019-01-30
fix(acc): Remove popup trigger button role

jurokapsiar:
Popup trigger can be a button, menu item or other components. It should be the consumers responsibility to provide appropriate role.

This fix is based on requirements for the chat header prototype.


layershifter (approved):
Had similar issue when tried to use `Label` as trigger 👍

PR #808, opened 2019-01-30
feat(menu): Padding variables for horizontal menu

jurokapsiar:
Consumer needs to be able to customize padding in the menu. This PR adds variables that allow to do this for the vertical variant.

I'm not sure if this is the right API as the variable names are bound to the variant (horizontal), so any feedback would be appreciated

### DIFF
--- a/src/themes/teams/components/Menu/menuVariables.ts
+++ b/src/themes/teams/components/Menu/menuVariables.ts
@@ -24,10 +24,7 @@ export interface MenuVariables {
 
   lineHeightBase: string
 
-  horizontalPaddingLeft: string
-  horizontalPaddingRight: string
-  horizontalPaddingTop: string
-  horizontalPaddingBottom: string
+  horizontalPadding: string
 }
 
 export default (siteVars: any): MenuVariables => {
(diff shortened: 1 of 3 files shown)


alinais (approved):

src/themes/teams/components/Menu/menuVariables.ts
```diff
@@ -48,5 +55,10 @@ export default (siteVars: any): MenuVariables => {
     disabledColor: siteVars.gray06,
 
     lineHeightBase: siteVars.lineHeightMedium,
+
+    horizontalPaddingLeft: pxToRem(18),
+    horizontalPaddingRight: pxToRem(18),
+    horizontalPaddingTop: pxToRem(14),
+    horizontalPaddingBottom: pxToRem(14),
```
layershifter: > When four values are specified, the paddings apply to the top, right, bottom, and left in that order (clockwise).
> https://developer.mozilla.org/en-US/docs/Web/CSS/padding#Values

My vote is to keep a single value:

```suggestion
horizontalPadding: `${pxToRem(14)} ${pxToRem(18)} ${pxToRem(14)} ${pxToRem(18)}`
```

layershifter (approved):

PR #811, opened 2019-01-31
feat(Chat): add `actionMenu` prop for `ChatMessage`

layershifter:
### `actions` vs `actionMenu`

Initially, I was thinking about `actions`, but this shorthand points to `Menu` component and it was really confusing why the code below will be broken:
```jsx
<ChatMessage actions={['one', 'two']} />
```

--- 

Also picks changes from #765.


jurokapsiar (approved):

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Features
 - Accessibility for menu divider @jurokapsiar ([#822](https://github.com/stardust-ui/react/pull/822))
+- Add `actionMenu` prop to `ChatMessage` component @layershifter ([#811](https://github.com/stardust-ui/react/pull/811))
```
kuzhelov, line 32: just putting my vote on that: would rather prefer `actions` than `actionsMenu` for the following reasons
- it will be easier to avoid breaking changes in future - for example, if at some point it will become an `actionsPanel`
- it is consistent with the naming taken for `Attachment` component, where there is an `action` slot

kuzhelov: but given the example from description, agree that it is definitely better to use singular `menu` noun in the name of the prop, to properly reflect underlying semantics

kuzhelov: actually, a bit surprised that we are not handling array as 'primitive' shorthand values - is there any particular reason for that? Might expect that this feature is necessary for all the places where collection-container component stays as a default one for shorthand value

docs/src/prototypes/chatMessageWithPopover/ChatWithPopover.tsx
```diff
@@ -1,6 +1,6 @@
 import { Chat, Provider, Avatar } from '@stardust-ui/react'
```
kuzhelov: agree with simpler name used - lets change directory name as well

packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
```diff
@@ -1,6 +1,9 @@
 import { pxToRem } from '../../../../lib'
 
 export interface ChatMessageVariables {
+  actionsBoxShadow: string
```
kuzhelov: lets agree on consistent naming for the slot variables -like `actionMenuXX`

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -1,10 +1,19 @@
...
+  default as ChatMessage,
+  ChatMessageProps,
+  ChatMessageState,
+} from '../../../../components/Chat/ChatMessage'
 import { ChatMessageVariables } from './chatMessageVariables'
 import { screenReaderContainerStyles } from '../../../../lib/accessibility/Styles/accessibilityStyles'
 import { pxToRem } from '../../../../lib'
 
-const chatMessageStyles: ComponentSlotStylesInput<ChatMessageProps, ChatMessageVariables> = {
+const chatMessageActionsClassNameSelector = `& .${ChatMessage.slotClassNames.actionMenu}`
```
kuzhelov: this seems to be used in the single place - lets just inline this expression there, to aid readability

packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -86,20 +102,35 @@ class ChatMessage extends UIComponent<ReactProps<ChatMessageProps>, ChatMessageS
...
   }
 
+  protected actionHandlers: AccessibilityActionHandlers = {
+    // prevents default FocusZone behavior, e.g., in ChatMessageBehavior, it prevents FocusZone from using arrow keys as navigation (only Tab key should work)
+    preventDefault: event => {
+      event.preventDefault()
+    },
+  }
+
   private handleFocus = (e: React.SyntheticEvent) => {
```
kuzhelov: lets remove `private` modifier from this method as well, to preserve local consistency. But totally agree, we need to have general grooming around this scope modifiers story

packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -86,20 +102,35 @@ class ChatMessage extends UIComponent<ReactProps<ChatMessageProps>, ChatMessageS
...
-    },
+  handleBlur = (e: React.FocusEvent) => {
+    const shouldPreserveFocusState = _.invoke(e, 'currentTarget.contains', e.relatedTarget)
+
+    this.changeFocusState(shouldPreserveFocusState)
+    _.invoke(this.props, 'onBlur', e, this.props)
+  }
+
+  changeFocusState = (isFocused: boolean) => {
+    if (this.state.focused !== isFocused) {
```
kuzhelov: why this check is necessary?

CHANGELOG.md
layershifter: I can be wrong, but this decision came from the consistency point: you can pass only `ShorthandValue` or `ShorthandValue[]`. However, I understand your point, had the same feeling about it.

packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -86,20 +102,35 @@ class ChatMessage extends UIComponent<ReactProps<ChatMessageProps>, ChatMessageS
...
     _.invoke(this.props, 'onFocus', e, this.props)
   }
 
-  protected actionHandlers: AccessibilityActionHandlers = {
-    // prevents default FocusZone behavior, e.g., in ChatMessageBehavior, it prevents FocusZone from using arrow keys as navigation (only Tab key should work)
-    preventDefault: event => {
-      event.preventDefault()
-    },
+  handleBlur = (e: React.FocusEvent) => {
+    const shouldPreserveFocusState = _.invoke(e, 'currentTarget.contains', e.relatedTarget)
```
kuzhelov, line 141: could you, please, suggest the case we would like to cover with that?

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
layershifter: Done 👍

packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
layershifter: Done 👍

packages/react/src/components/Chat/ChatMessage.tsx
layershifter: Removed it 👍

packages/react/src/components/Chat/ChatMessage.tsx
layershifter: It comes from the accessibility point: when we will click `Enter` key, the focus will be moved to another element.

### With

chat-with1

### Without

chat-without

CHANGELOG.md
kuzhelov: got it, thank you @layershifter! actually, this original problem (`actions` and necessity to compromise on `actionsMenu`) is a good example that might serve as an inspiration for us to further refine this moment in future

packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -56,6 +62,13 @@ export interface ChatMessageProps
   /** A message can format the badge to appear at the start or the end of the message. */
   badgePosition?: 'start' | 'end'
 
+  /**
+   * Called after user's focus.
```
kuzhelov: we need to update this one :)

packages/react/src/components/Chat/ChatMessage.tsx
kuzhelov: then, frankly, I am a bit confused with `onBlur` event being raised unconditionally - so, for instance, even if `shouldPreserveFocusState` is `true` (and, as a consequence,  even `this.state.focused` remains to be true) - even in this situation we will trigger `blur` event for `ChatMessage`.

So, couple of things that I would like to solve here:
- ensure that I am not missing anything in this thought 
- and, in case if so, it seems that we need to resolve this emerged problem that is about implicit reliance of events subscriptions - we should strive for component's behavioural intents being explicit

Could you, please, to resolve this issue for now, just provide a comment that will describe the intent of these lines? Thank you!

kuzhelov (approved):

packages/react/src/components/Chat/ChatMessage.tsx
layershifter: `this.state.focused` controls is focused the `ChatMessage` on any of its children, when we're navigating with keyboard focused elements changes and there is no way to use `:focus`:

image

May be proper naming can make it better?

kuzhelov: yes, totally agree about naming - this should help. However, it seems that many things are involved here, so it is, probably, much better to address naming and all the related moments in a dedicated PR - what do you think?

PR #816, opened 2019-01-31
fix(Dropdown): rename `combobox` to `root` in `dropdownSearchInputStyles`

### LINKED ISSUE #753: dropdownSearchInputStyles missing `root` key
### Steps
Call `dropdownSearchInputStyles.root()`

### Expected Result
Styles for the root element should be provided.

### Actual Result
```
TypeError: dropdownSearchInputStyles.root is not a function
```

### Version
0.17.0

### Testcase
See [dropdownSearchInputStyles.ts](https://github.com/stardust-ui/react/blob/master/src/themes/teams/components/Dropdown/dropdownSearchInputStyles.ts), there is no `root` key. 💣

### Fix Considerations

It seems there should be parity between the `DropdownSearchInput` and the `Input` components.  However, the Input component conflates the `wrapper` and the `root` style targets.

It also seems there should be parity between components which have a `wrapper` part.  However, the `DropdownSearchInput`, `Input`, and `MenuItem` are all different.  These are different so that the intuitive usage of their component APIs is most correct, however, this seems to cause issues when trying to programmatically loop over the theme and deal with component styles.

It could be that we need to keep intuitive mappings at the component API level and keep programmatically consistent mappings at the style level.  We need more thinking and resolution on this.

layershifter:
Fixes #753.

# BREAKING CHANGES

This PR will affect you only if you're using custom styling for `DropdownSearchInput` component.

### Before

```jsx
const theme = {
  componentStyles: {
    DropdownSearchInput: {
      combobox: { /* your styles */ },
    },
  },
}
```

### After

```jsx
const theme = {
  componentStyles: {
    DropdownSearchInput: {
      root: { /* your styles */ },
    },
  },
}
```


layershifter:
I tried to enforce `root` presence with typings, but it will break overrides via `Provider` 👎

CHANGELOG.md
```diff
@@ -46,6 +49,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Exports `Menu`'s variables as `MenuItem` and `MenuDivider`'s variables in `Teams` theme @mnajdova ([#814](https://github.com/stardust-ui/react/pull/814))
 - Fix `Avatar` - Add white circle border in contrast theme @bcalvery ([#795](https://github.com/stardust-ui/react/pull/795))
 - Dix `MenuDivider` not shown on horizontal `Menu` @mnajdova ([#813](https://github.com/stardust-ui/react/pull/813))
+- Add `root` to `dropdownSearchInputStyles` @layershifter ([#816](https://github.com/stardust-ui/react/pull/816))
```
bmdalex: already have the entry at line 24 :)

bmdalex (approved):

packages/react/src/themes/teams/components/Dropdown/dropdownSearchInputStyles.ts
```diff
@@ -6,6 +6,8 @@ const dropdownSearchInputStyles: ComponentSlotStylesInput<
   DropdownSearchInputProps,
   DropdownVariables
 > = {
+  root: {},
```
mnajdova: What was the motivation for this? As I see in `DropdownSearchInput.tsx` we are not using styles.root, nor classes.root. I don't understand why we need this?

silviuaavram (approved):

mnajdova (approved):

miroslavstastny (approved):

PR #817, opened 2019-01-31
chore: add TS support and remove hardcoded values from visual tests config

kuzhelov:
This PR introduces ability to consume TS modules in Screener configuration file. 

This move allows to consume configuration parameters from `config` module, as well as to remove hardcoded class selectors from specs of interactive tests.


docs/src/examples/components/Dropdown/Types/DropdownExampleSingleSelection.shorthand.steps.ts
```diff
@@ -0,0 +1,13 @@
+import { Dropdown, Button, List } from '../../../../../../src'
```
layershifter, line 1:
```suggestion
import { Dropdown, Button, List } from '@stardust-ui/react'
```

Can we use just this?

If it requires `tsconfig`, this should work: 
```ts
require("ts-node").register({
  project: "build/tsconfig.common.json",
});
```

screener.config.js
```diff
@@ -31,26 +37,37 @@ const screenerConfig = {
...
+      const {
+        name: exampleNameWithoutExtension,
+        base: exampleNameWithExtension,
+        dir: exampleDir,
+      } = path.parse(examplePath)
+
+      const rtl = exampleNameWithExtension.endsWith('.rtl.tsx')
+      const url = `http://${SCREENER_HOST_URL}/maximize/${_.kebabCase(
+        exampleNameWithoutExtension,
+      )}/${rtl}`
```
layershifter:
```suggestion
const exampleUrl = _.kebabCase(exampleNameWithoutExtension)
const url = `http://${SCREENER_HOST_URL}/maximize/${exampleUrl}/${rtl}`
```

Should be more readable

layershifter (approved):

docs/src/examples/components/Dropdown/Types/DropdownExampleSingleSelection.shorthand.steps.ts
kuzhelov: let me check..

kuzhelov: unfortunately, this doesn't help - but, actually, I would also expect that it should. Will take another look tomorrow, thank you!

kuzhelov: taking this as a general problem (import expressions like this one could be met quite often in the code), will clean up this stuff in the dedicated PR. Currently wasn't able to come up with resolution, but will spend additional time today

PR #820, opened 2019-02-01
chore: update imports

layershifter:
This PR aligns internal imports in multiple places.

### DIFF
--- a/docs/src/components/ExternalExampleLayout.tsx
+++ b/docs/src/components/ExternalExampleLayout.tsx
@@ -1,3 +1,4 @@
+import { Provider } from '@stardust-ui/react'
 import * as _ from 'lodash'
 import * as PropTypes from 'prop-types'
 import * as React from 'react'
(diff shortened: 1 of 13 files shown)


docs/src/components/ExternalExampleLayout.tsx
```diff
@@ -1,3 +1,4 @@
+import { Provider } from '@stardust-ui/react'
```
mnajdova, line 1: Can we move imports from our internal project after the imports form external dependencies. It is much clearer when you look them to figure out what is coming form where. This is my personal opinion, let me know what you think.

mnajdova (approved):

docs/src/components/ExternalExampleLayout.tsx
layershifter: I want to introduce ESLint/TSLint and align all imports in future
I placed  `@stardust-ui/react` at first because namespaced imports have more priority in sort

Issue #773, opened 2019-01-25
Performance test plan


miroslavstastny:
# Perf measuring

This document describes workflow used to measure (and avoid regressions in) performance / bundle size.

## Scenarios

| Scenario | Setup | Question it answers
| :--- | :--- | :--- |
| Minimal           | Only required props        | What is the base overhead cost of a component?
| Slots             | All slot props             | What is the cost of slot prop logic? 
| InlineStyles      | Inline style overrides     | What is the cost of inline style overrides? 
| InlineVariables   | Inline variable overrides  | What is the cost of inline variable overrides?
| NoTheme           | No theming system applied  | What is the cost of the theming system? 

- [x] Add Minimal scenarios #2164 
- [ ] Add Slots scenarios
- [ ] Add InlineStyles scenarios
- [ ] Add InlineVariables scenarios
- [ ] Add NoTheme scenarios

## Workflow overview

1. run perf tests as part of CI for every PR build,
2. store the results to external storage,
3. in every PR, compare current results with results from a fork point,
4. (optional), with every release, get current results from external storage, push them to git,
5. show numbers (with history and charts) in docsite.

### Run perf tests

- [x] `yarn build:stats` to measure bundle size locally
- [x] run bundle size tests in CI
- [x] `yarn perf` to measure performance locally
- [x] run perf tests in CI (will the results be stable enough?)
- [x] add perf examples for all components #2164 
- [ ] add perf examples for compound pages (See [Semantic-UI/React layouts](https://react.semantic-ui.com/layouts))
- [ ] add conformance test for perf tests

### External storage

- [x] spin-up external storage

Let's start with a nosql database, with following keys:
- **commit hash**
    - to map 1:1 from git commit to the result
- **git branch**
    - to be able to display stats for a branch without git
    - may not work for branches other than master as branch names can be reused
    - but is it required for a branch?
    - works for master
- **ci build number**
    - to order results on a single branch
    - to distinguish multiple builds on a single commit

### Comparing results in every PR
- [x] save current results to external storage
- [ ] for PR builds, compare current results with fork-point results, add the diff to the PR

As part of each PR build, perf tests are executed, results stored to the external storage and compared with master.

To find correct master commit:
```
git merge-base --fork-point master
```
*Note: This does not work for PR to other branches than master, not tested for PRs from forks!*

We will use DangerJS to compare the results and add them as a comment to the PR.

In the first version it will just log the difference without enforcing any policy.

#### Master builds

[...]

PR #821, opened 2019-02-01
chore: Save package statistics to DB

miroslavstastny:
Part of #773.

Each CI build stores results of `yarn build:stats` to an external DB.
Following data are stored:
- build number,
- GIT commit SHA,
- GIT branch,
- GIT tag (to filter release builds),
- GitHub PR URL,
- build date
- current bundle size statistics.

No magic is currently done on DB side, data is stored as-is. Based on the required queries, we can restructure the data and migrate all existing ones.

No user-facing change, changelog left intentionally blank.


miroslavstastny:
This is repo for the DB API endpoint: https://github.com/miroslavstastny/stardust-stats-azure (to be moved to stardust repo later)

package.json
```diff
@@ -34,6 +34,7 @@
     "release:patch": "yarn prerelease && ta-script npm/release patch && yarn postrelease",
     "prestart": "yarn satisfied --fix yarn",
     "start": "gulp --series dll docs",
+    "stats:save": "gulp stats:save",
```
layershifter, line 37: Can we rename `build:stats` to `stats:build`? (line 5)

```diff
-"build:stats": "gulp stats",
+"stats:build": "gulp stats:build",
```

.circleci/config.yml
```diff
@@ -89,7 +97,7 @@ jobs:
       - run:
           name: Publish npm package
           command: |
-            if [ -n "${npm_TOKEN+x}" ] && [ "${CIRCLE_BRANCH}" == "master" ]; then
+            if [ -n "${npm_TOKEN}" ] && [ "${CIRCLE_BRANCH}" == "master" ]; then
```
layershifter, line 95: This change looks strange, is it correct?

build/gulp/tasks/stats.ts
```diff
@@ -121,3 +123,27 @@ task(
...
+task('stats:save', async () => {
+  const commandLineArgs = _.pick(argv, ['sha', 'branch', 'tag', 'pr', 'build'])
+  const bundleStats = require(paths.docsSrc('bundleStats.json'))[UNRELEASED_VERSION_STRING]
+
+  const options = {
+    method: 'POST',
+    uri: process.env.STATS_URI,
+    body: {
+      sha: process.env.CIRCLE_SHA1,
+      branch: process.env.CIRCLE_BRANCH,
```
layershifter: There values are required as I understand, can we add an assert that they're defined?

.circleci/config.yml
kuzhelov: yes, as `-n` `test` function already correctly handles the values `+x` filter has targeted on

build/gulp/tasks/stats.ts
```diff
@@ -121,3 +123,27 @@ task(
...
+      pr: process.env.CIRCLE_PULL_REQUEST, // optional
+      build: process.env.CIRCLE_BUILD_NUM,
+      ...commandLineArgs, // allow command line overwrites
+      bundleSize: bundleStats,
+      ts: new Date(),
+    },
+    json: true,
+  }
+
+  const ret = await rp(options)
```
kuzhelov: lets omit shortcuts here, as it makes the code a bit unreadable. We could end up with something like (with import aliasing and variable being renamed)

```js
const response = await requestHttp(options)
```

layershifter (approved):
👍

kuzhelov (approved):

.circleci/config.yml
```diff
@@ -74,7 +74,15 @@ jobs:
...
           name: Bundle Statistics
-          command: yarn build:stats
+          command: yarn stats:build
+      - run:
+          name: Save Statistics to DB
+          command: |
+            if [ -n "${STATS_URI}" ]; then
+              yarn stats:save
+            else
+              echo "STATS_URI not set, skipping"
```
levithomason, line 84: Seems this should throw, no?  Do we want silent passes when stats aren't being saved for some reason?

PR #822, opened 2019-02-01
feat(Accessibility): Accessibility for menu divider

jurokapsiar:
This PR adds accessibility behavior to the menu divider


mnajdova (approved):
LGTM, just update the changelog.

src/components/Menu/MenuDivider.tsx
```diff
@@ -28,17 +36,25 @@ class MenuDivider extends UIComponent<ReactProps<MenuDividerProps>, any> {
 
   static defaultProps = {
     as: 'li',
+    accessibility: menuDividerBehavior as Accessibility,
   }
 
   static propTypes = {
     ...commonPropTypes.createCommon({ content: false, children: false, color: true }),
+    accessibility: PropTypes.func,
```
layershifter, line 44: I've seen another definition:
```
```

Can we create a customPropType for this in separate PR? May be create a tracking issue?

src/lib/accessibility/Behaviors/Menu/menuDividerBehavior.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Adds role 'presentation' to 'root' component's part.
+ */
+
+const menuDividerBehavior: Accessibility = (props: any) => ({
+  attributes: {
+    root: {
+      role: 'presentation',
+    },
+  },
+})
```
layershifter, line 17: A little confused with this, we don't have a behavior for `Divider`. How are they differenced?

src/components/Menu/MenuDivider.tsx
jurokapsiar: yes, we used to support object and function, now it is just a function, probably some cleanup is needed. I will create an issue.

src/lib/accessibility/Behaviors/Menu/menuDividerBehavior.ts
jurokapsiar: it's just that we did not get to divider yet. it will be done soon. If it turns out to be the same, we will consolidate.

mnajdova: The `MenuDivider` is part of the list (it is `li`) element and I guess as the other elements in the menu has this role.

PR #823, opened 2019-02-01
feat(ChatMessage): add badge and badgePosition props

mnajdova:
This PR adds the badge and badgePosition ('start' | 'end') props to the ChatMessage. 
TODO
- [x] update changelog
- [x] provide examples for the badge

Here are examples of this:

- Mine message with badge at the end:

```jsx
 <Chat.Message
          content="Hello"
          author="John Doe"
          mine
          badge={{
            icon: 'redbang',
            color: 'red',
          }}
/>
```
image

- Mine message with badge at the start
```jsx
 <Chat.Message
          content="Hello"
          author="John Doe"
          mine
          badge={{
            icon: 'redbang',
            color: 'red',
          }}
         badgePosition="start"
/>
```

image

- Other message with badge at the end

```jsx
       <Chat.Message
          content="Hi"
          author="Jane Doe"
          badge={{
            icon: 'mention',
            color: 'red',
          }}
        />
```

image

- Other message with badge at the start
```jsx
        <Chat.Message
          content="Hi"
          author="Jane Doe"
          badge={{
            icon: 'mention',
            color: 'red',
          }}
          badgePosition="start"
        />
```
image

Example in chat:
image


layershifter (commented):
I am not sure about `badge`, can we go with `label` for now?
And when we will have decision about naming just rename it everywhere?

src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -44,6 +45,17 @@ const chatMessageStyles: ComponentSlotStylesInput<ChatMessageProps, ChatMessageV
       textDecoration: 'underline',
     },
   }),
+  badge: ({ props: { badgePosition } }) => {
+    const sidePosition = badgePosition === 'start' ? 'left' : 'right'
```
layershifter:
```suggestion
    const sidePosition = p.badgePosition === 'start' ? 'left' : 'right'
```

Can we use `p.`? 😸

mnajdova:
@layershifter in my opinion label as a word has nothing semantically with the ChatMessage, although it is shorthand for label (yep, the name had already produced us enough headache). Even if this was shorthand for the Icon component, I would still go with badge, as it is much more clear what it represents...

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -44,6 +45,19 @@ const chatMessageStyles: ComponentSlotStylesInput<ChatMessageProps, ChatMessageV
       textDecoration: 'underline',
     },
   }),
+  badge: ({ props: p, variables: v }) => {
+    const sidePosition = p.badgePosition === 'start' ? 'left' : 'right'
+    return {
+      boxShadow: v.badgeShadow,
+      position: 'absolute',
+      height: pxToRem(24),
```
miroslavstastny: This breaks when you change icon size. Any chance this could work without absolute values?

Something like:
```jsx
      boxShadow: v.badgeShadow,
      position: 'absolute',
      padding: pxToRem(4),
      height: 'auto',
      width: 'auto',
      borderRadius: '50%',
      top: pxToRem(4),
      zIndex: '1',
      [sidePosition]: 0,
      transform: p.badgePosition === 'start' ? 'translateX(-50%)' : 'translateX(50%)',
```

miroslavstastny (commented):
What about Styled Chat Item example?

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
bmdalex: +1 for using transform

bmdalex (approved):
looks good, I vote for having `badge` for name for now
https://ant.design/components/badge/

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
mnajdova: Thanks Miro, I've changed the styles, work with every size of icon, and the default is according to the specs. I've added badge in the styles ChatItem example as well. Let me know if there is anything else.

miroslavstastny (approved):

packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -83,6 +93,7 @@ class ChatMessage extends UIComponent<ReactProps<ChatMessageProps>, ChatMessageS
   static defaultProps = {
     accessibility: chatMessageBehavior,
     as: 'div',
+    badgePosition: 'end',
```
kuzhelov, line 97: not sure about this default - lot of examples where badge appears at start

kuzhelov (approved):
not sure about default badge's position - I would rather suggest to make it `start` one.

PR #827, opened 2019-02-01
feat(theming): exporting slot's classNames form components

### LINKED ISSUE #784: Export slot's classNames as static fields from the Component
### Problem description
For some of the components' slots that does not have dedicated components, we are defining classNames in order for defining css selectors for them. As an example, in the `MenuItem` component, we have custom class for the wrapper element: `ui-menu__item__wrapper` which is not exported anywhere from stardust. In addition to this, we should make sure that all slot's inside the components can be safely selected for applying styles.

### Proposed solution
Let's add the static slotClassNames object to the components that are defining some custom classNames for the slots (See the example in the ChatMessage component as a guidance).

### MVP

mnajdova:
This PR fixes https://github.com/stardust-ui/react/issues/784 by adding slotClassNames static fields int he component which have already specified some custom classNames for the slots.

# BREAKING CHANGES
The classNames for the slots in the ItemLayout component have been renamed. Please replace the following selectors accordingly:
```
import { ItemLayout } from '@stardust-ui/react'

ui-item-layout__header <=> ItemLayout.slotClassNames.header
ui-item-layout__headerMedia <=> ItemLayout.slotClassNames.headerMedia
ui-item-layout__main <=> ItemLayout.slotClassNames.main
ui-item-layout__content <=> ItemLayout.slotClassNames.content
ui-item-layout__contentMedia <=> ItemLayout.slotClassNames.contentMedia
ui-item-layout__media <=> ItemLayout.slotClassNames.media
ui-item-layout__endMedia <=> ItemLayout.slotClassNames.endMedia
```

# Question
Do we want to introduce slot's classes for the slots inside the component which are created with the Box or Text component? It seems reasonable to me, as these components can be heavily used in the content of the components, so selecting the slot's containers might be tricky... On the other hand, they can be styled via the slot's inside the component styles, but still creating selectors for them if necessary can be tricky.. Let me know what you think! 
# Decision
We will add them if there is requirement from the users.


src/components/Layout/Layout.tsx
```diff
@@ -173,9 +185,9 @@ class Layout extends UIComponent<ReactProps<LayoutProps>, any> {
     if (reducing && isSingleArea) {
       const composedClasses = cx(
         classes.root,
-        startArea && 'ui-layout--reduced__start',
-        mainArea && 'ui-layout--reduced__main',
-        endArea && 'ui-layout--reduced__end',
+        startArea && Layout.slotClassNames.reducedStart,
```
mnajdova: Not sure if this is the right convention, as the className here is changed because of a prop...

src/components/ItemLayout/ItemLayout.tsx
```diff
@@ -175,8 +187,8 @@ class ItemLayout extends UIComponent<ReactProps<ItemLayoutProps>, any> {
     const mainArea = renderMainArea(this.props, this.state, classes)
     const endArea = endMedia
 
-    const mergedMediaClasses = cx('ui-item-layout__media', classes.media)
```
mnajdova: Ups, breaking changes introduced. It seems that the classNames of this slots were not correct...

silviuaavram (approved):

CHANGELOG.md
```diff
@@ -17,8 +17,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### BREAKING CHANGES
+- Renamed classNames for the slots inside the `ItemLayout` component @mnajdova ([#827](https://github.com/stardust-ui/react/pull/827))
```
kuzhelov: lets do either 'class names' or '`className`s'. I would suggest the first option, as it is more correct and implementation agnostic :)

CHANGELOG.md
```diff
@@ -17,8 +17,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### BREAKING CHANGES
+- Renamed classNames for the slots inside the `ItemLayout` component @mnajdova ([#827](https://github.com/stardust-ui/react/pull/827)) 
+
 ### Features
 - Accessibility for menu divider @jurokapsiar ([#822](https://github.com/stardust-ui/react/pull/822))
+- Added slotClassNames in `ChatMessage`, `ChatItem`, `Dropdown`, `ItemLayout`, `Layout`, `MenuItem` @mnajdova ([#827](https://github.com/stardust-ui/react/pull/827))
```
kuzhelov: would suggest just 'slot class names'

packages/react/src/components/ItemLayout/ItemLayout.tsx
```diff
@@ -117,8 +129,8 @@ class ItemLayout extends UIComponent<ReactProps<ItemLayoutProps>, any> {
     renderHeaderArea: (props, state, classes) => {
       const { debug, header, headerMedia, truncateHeader, headerCSS, headerMediaCSS } = props
 
-      const mergedClasses = cx('ui-item-layout__header', classes.header)
-      const mediaClasses = cx('ui-item-layout__headerMedia', classes.headerMedia)
+      const mergedClasses = cx(ItemLayout.slotClassNames.header, classes.header)
```
kuzhelov: lets rename these variables to `headerClasses` and `headerMediaClasses` - although, surely, this is a problem that has existed long before this PR

packages/react/src/components/Layout/Layout.tsx
```diff
@@ -173,9 +185,9 @@ class Layout extends UIComponent<ReactProps<LayoutProps>, any> {
     if (reducing && isSingleArea) {
       const composedClasses = cx(
         classes.root,
-        startArea && 'ui-layout--reduced__start',
-        mainArea && 'ui-layout--reduced__main',
-        endArea && 'ui-layout--reduced__end',
+        startArea && Layout.slotClassNames.reducedStart,
```
kuzhelov: to me it seems that there was a problem already there - the class we need to assign for reduced area is rather `'ui-layout__start--reduced`, not the `'ui-layout--reduced__start` that we've had. The reason is that, essentially, it is about marking the same `start` area, but using `reduced` state modifier for it

kuzhelov (approved):

PR #828, opened 2019-02-01
fix(Avatar): alignment of status icon on size changes

kuzhelov:
# TODO
- [x] introduce breaking change entry to CHANGELOG

# Breaking Change

This PR unifies the set of values that could be provided to `size` prop of `Avatar` and `Status` components: instead of `number`s that were originally used, now there is a finite set of string values that could be provided as `size`: 
```
type SizeValue = 'smallest' | 'smaller' | 'small' | 'medium' | 'large' | 'larger' | 'largest'
```

## Necessary changes on the client side
This PR requires the following changes to be made on the client side at places where `size` values were provided to either `Avatar` component, or to the `size` prop of its `status` slot. No changes should be introduced in case if no any value was provided to `size` prop.

### `sizes` of Avatar and `status` slot originally were not in sync
Size of status icon was always remained the same and haven't responded to `Avatar`'s size changes - now size of `status` slot changes alongside with changes to `Avatar`'s size:

```jsx
<Avatar size='larger' status={{ icon: 'check', color: 'green' }} image={...} />
```

#### Was

image

#### Now

image

In order to preserve original behavior size of `status` slot must be set explicitly:

```jsx
<Avatar size='larger' status={{ size='medium', ... }} image={...} />
```

### `Avatar`, value was provided to `size` prop
```jsx
// Before
<Avatar size={32} />

// After
<Avatar size='medium' />
```

### `Avatar`, `size` value is provided to `status` slot
```jsx
// Before
<Avatar status={{ size: 10, ... }}/>

// After
<Avatar status={{ size: 'medium', ... }}/>
```
## Removed types
`AvatarPropsWithDefaults` and `StatusPropsWithDefaults` types were removed from public API. Client should replace them with `AvatarProps` and `StatusProps` at places of usage, accordingly.

---------

# Description

This PR unifies size values for Avatar component - now those values are taken from common Stardust dictionary (`smallest` -> ... -> `largest`). 

Related styles of Teams theme were adjusted accordingly.

#### Was

image

#### Now

image


packages/react/src/themes/teams/components/Status/statusStyles.ts
```diff
@@ -35,16 +35,33 @@ const getTextColor = (state: string, variables: StatusVariables) => {
...
+  smaller: 8,
+  small: 8,
+  medium: 10,
+  large: 12,
+  larger: 14,
+  largest: 16,
+}
+
+export const getSizeStyles = (sizeInPx: number, variables: StatusVariables) => {
+  const sizeInRem = pxToRem(sizeInPx + 2 * ((variables.borderColor && variables.borderWidth) || 0))
```
layershifter:
```suggestion
  const borderWidth = (variables.borderColor && variables.borderWidth) || 0
  const sizeInRem = pxToRem(sizeInPx + 2 * borderWidth)
```

One line will not save us 😸 Can we make it more readable?

packages/react/src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -1,14 +1,14 @@
...
+import { callable, pxToRem, SizeValue } from '../../../../lib'
 import { ComponentSlotStylesInput, ICSSInJSStyle, FontIconSpec } from '../../../types'
 import { ResultOf } from '../../../../types'
-import { IconXSpacing, IconProps, IconSize } from '../../../../components/Icon/Icon'
+import { IconXSpacing, IconProps } from '../../../../components/Icon/Icon'
 import { getStyle as getSvgStyle } from './svg'
 import { IconVariables, IconSizeModifier } from './iconVariables'
 
-const sizes: { [key in IconSize]: number } = {
+const sizes: { [key in SizeValue]: number } = {
```
layershifter:
```suggestion
const sizes: Record<SizeValue, number> = {
```

As I remember this should work

packages/react/src/themes/teams/components/Status/statusStyles.ts
```diff
@@ -35,16 +35,33 @@ const getTextColor = (state: string, variables: StatusVariables) => {
   }
 }
 
-const statusStyles: ComponentSlotStylesInput<StatusPropsWithDefaults, StatusVariables> = {
+export const sizeToPxValue = {
```
layershifter: Can we introduce a type there, i.e.:

```suggestion
export const sizeToPxValue: Record<SizeValue, number> = {
```

packages/react/src/index.ts
```diff
@@ -120,7 +120,7 @@ export {
 export { default as Ref, RefProps } from './components/Ref/Ref'
 export { default as Segment, SegmentProps } from './components/Segment/Segment'
 
-export { default as Status, StatusPropsWithDefaults, StatusProps } from './components/Status/Status'
+export { default as Status, StatusProps } from './components/Status/Status'
```
layershifter, line 142: `StatusPropsWithDefaults` was part of public API, so this is a breaking change, too

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -23,10 +24,10 @@ export interface AvatarProps extends UIComponentProps {
   name?: string
 
   /** Size multiplier. */
-  size?: number
+  size?: SizeValue
```
mnajdova, line 27: Can you check if there are some usages in the prototypes (I am pretty sure there is employee card prototype that has bigger Avatar)

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -23,10 +24,10 @@ export interface AvatarProps extends UIComponentProps {
...
 
   /** Shorthand for the status of the user. */
-  status?: ShorthandValue
+  status?: ShorthandValue<StatusProps>
```
mnajdova, line 30: 👍

mnajdova (approved):
LGTM please just check the prototypes if there are some miss-usages of the size props after these changes.

packages/react/src/components/Avatar/Avatar.tsx
kuzhelov: sure, thanks for heads up!

PR #829, opened 2019-02-01
fix(button): Update button to use semibold weight

notandrew:
Update button text to be semibold instead of regular font weight. Fix https://github.com/stardust-ui/react/issues/862

**Before**
image

**After**
image


src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -279,10 +279,12 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps & ButtonState, any> = {
     }
   },
 
-  content: () => ({
+  // modifies the text of the button
+  content: ({ variables }): ICSSInJSStyle => ({
     overflow: 'hidden',
     textOverflow: 'ellipsis',
     whiteSpace: 'nowrap',
+    fontWeight: variables.fontWeight,
```
codepretty: i think you had mentioned this before, but suggest renaming `variables.fontWeight` to something more specific

kuzhelov: `buttonFontWeight` for now - would it be better?

src/themes/teams/components/Button/buttonVariables.ts
```diff
@@ -9,6 +9,7 @@ export interface ButtonVariables {
   borderRadius: string
   circularRadius: string
   paddingLeftRightValue: number
+  fontWeight: string
```
kuzhelov: shouldn't the value be provided for this variable (or am I missing something?)

codepretty: The value was already provided - https://github.com/stardust-ui/react/blob/master/packages/react/src/themes/teams/components/Button/buttonVariables.ts

packages/react/src/themes/teams/components/Button/buttonVariables.ts
```diff
@@ -74,7 +75,7 @@ export default (siteVars: any): ButtonVariables => {
     maxWidth: pxToRem(280),
     borderRadius: pxToRem(2),
     circularRadius: pxToRem(999),
-    fontWeight: siteVars.fontWeightSemibold,
+    buttonFontWeight: siteVars.fontWeightSemibold,
```
mnajdova: I know there was already discussion about the name, but maybe `contentFontWeight` will be better name, as it is not applied to the icon part. In my opinion` buttonFontWeight` is not any different then `fontWeight`, as it is used inside the button, but doesn't explain that it is applied only in the content.

mnajdova (approved):

PR #832, opened 2019-02-04
chore(tsconfig): use correct paths

layershifter:
Refs #825.

Corrects paths as @kuzhelov mentioned in https://github.com/stardust-ui/react/commit/da0fc96425cc65c9c62960a5dcf24f0008809169#r32149482.

### DIFF
--- a/build/tsconfig.common.json
+++ b/build/tsconfig.common.json
@@ -5,8 +5,9 @@
     "lib": ["es2015", "dom"],
     "baseUrl": "../",
     "paths": {
-      "@stardust-ui/react": ["packages/react/src"],
-      "*": ["packages/react/*", "*", "types/*"]
+      "@stardust-ui/*": ["packages/*/src"],
+      "docs/*": ["docs/*"],
+      "src/*": ["packages/react/src/*"]
     },
     "types": ["node", "jest"],
     "typeRoots": ["../types", "../node_modules/@types"],
--- a/build/tsconfig.common.json
+++ b/build/tsconfig.common.json
@@ -6,7 +6,7 @@
     "baseUrl": "../",
     "paths": {
       "@stardust-ui/react": ["packages/react/src"],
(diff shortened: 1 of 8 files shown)


build/tsconfig.commonjs.json
```diff
@@ -3,5 +3,5 @@
   "compilerOptions": {
     "module": "commonjs"
   },
-  "include": ["../packages/src", "../packages/types", "../types"]
```
layershifter, line 6: Path `packages/types` is not existing, `packages/src` should be `packages/react/src`

build/tsconfig.common.json
```diff
@@ -6,7 +6,7 @@
     "baseUrl": "../",
     "paths": {
       "@stardust-ui/react": ["packages/react/src"],
-      "*": ["*", "types/*"]
+      "*": ["packages/react/*", "*", "types/*"]
```
kuzhelov: lets prioritise 'one-to-one' path mapping option (`*`) as the first thing to try - this will ensure correctness in cases of names collision

kuzhelov (approved):

build/tsconfig.common.json
layershifter: `*` is the second because at first we should look in `packages/react`, the most imports comes from there. There is a performance benefit around 5-10 seconds with this order.

layershifter: Want to mention: I'm happy that we have there a star at all.

kuzhelov: then may we just remove it? 

what I am not happy with is that if some module would suddenly have a name match in `packages/react/*`, then this one will be imported. In that case it will be extremely hard to reason about module's location in the following (pseudo) case:

```js
// these statements might import modules from two different dirs,
// even if both `./one` and `./another` dir would physically exist
import * from './one'
import * from './another'
```

Can we remove star resolution?

PR #833, opened 2019-02-04
chore(package): remove "leven" from dependencies

layershifter:
This PR:
- removes `leven()` export from lib, it's not exported as public API and used only in `customPropTypes`
- removes `leven` package as dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,7 +129,6 @@
     "jest": "^23.6.0",
     "jest-axe": "^3.1.0",
     "json-stable-stringify-without-jsonify": "^1.0.1",
-    "leven": "^2.1.0",
     "lint-staged": "^7.0.2",
     "merge2": "^1.2.2",
     "normalize.css": "^8.0.0",
--- a/packages/react/src/components/Tree/Tree.tsx
+++ b/packages/react/src/components/Tree/Tree.tsx
@@ -7,6 +7,7 @@ import {
   UIComponent,
   childrenExist,
   commonPropTypes,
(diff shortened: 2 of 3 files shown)


packages/react/src/components/Tree/Tree.tsx
```diff
@@ -7,14 +7,14 @@ import {
...
   commonPropTypes,
+  customPropTypes,
   UIComponentProps,
   ChildrenComponentProps,
   rtlTextContainer,
 } from '../../lib'
 import { ShorthandValue, ShorthandRenderFunction, ReactProps } from '../../types'
 import { Accessibility } from '../../lib/accessibility/types'
 import { defaultBehavior } from '../../lib/accessibility'
-import * as customPropTypes from '../../lib/customPropTypes'
```
layershifter, line 17: Oops, wrong import

package.json
```diff
@@ -129,7 +129,6 @@
     "jest": "^23.6.0",
     "jest-axe": "^3.1.0",
     "json-stable-stringify-without-jsonify": "^1.0.1",
-    "leven": "^2.1.0",
```
layershifter, line 132: `leven` package was used in SUIR for [IconSearch](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/src/components/IconSearch/IconSearch.js) component, we don't have such.

_Why? In production mode, i.e. hosted docs `leven()` function from `lib` is nooped_

miroslavstastny (approved):

PR #834, opened 2019-02-04
feat(Icon): Teams theme add static outline and filled classes to SVG icons

### LINKED ISSUE #716: Teams theme: Add outline and filled classes to icons
Add `outline` and filled` classes to icons so that they can be addressed by parent elements (used to fill icon on parent hover).
POC: #364.

Required for features where a component with an icon needs to show that icon's filled state when hovered. For example in #708.

miroslavstastny:
Closes #716.

- [x] Add `teamsIconSlotClassNames.outline` and `filled` to all SVG icons in Teams theme which have outline and/or filled paths,
- [x] do the same for all ProcessedIcons in Teams theme,
- [x] add UT to verify these classes on Teams icons,
- [x] Use `teamsIconSlotClassNames` instead of hardcoded class names in theme styles,
- [x] update CHANGELOG.


packages/react/test/specs/themes/teams/components/Icon/Icon-test.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import icons, {
+  teamsIconSlotClassNames,
+} from '../../../../../../src/themes/teams/components/Icon/svg'
+import processedIcons from '../../../../../../src/themes/teams/components/Icon/svg/processedIndex'
+import { SvgIconSpecWithStyles } from '../../../../../../src/themes/teams/components/Icon/svg/types'
+
+describe('Teams Theme Icon', () => {
+  function testIcon(icon) {
+    const iconFunc: any = (icon as SvgIconSpecWithStyles).icon
```
miroslavstastny: This does not cover all possible Icon flavors, but works for all Teams theme icons.
Instead of implementing universal test now, let it fail and iterate.

packages/react/src/themes/teams/components/Icon/svg/ProcessedIcons/icons-accept.tsx
```diff
@@ -1,16 +1,18 @@
 import * as React from 'react'
+import cx from 'classnames'
 import { TeamsProcessedSvgIconSpec } from '../types'
+import { teamsIconSlotClassNames } from '../index'
```
layershifter: We should move `teamsIconSlotClassNames` to a separate file as it produces circular imports.

See `Project Tests` in CI:
```
(!) Circular dependency:
node_modules/@stardust-ui/react/dist/es/themes/teams/components/Icon/svg/index.js -> 
node_modules/@stardust-ui/react/dist/es/themes/teams/components/Icon/svg/icons/index.js -> 
node_modules/@stardust-ui/react/dist/es/themes/teams/components/Icon/svg/icons/undo.js -> 
node_modules/@stardust-ui/react/dist/es/themes/teams/components/Icon/svg/index.js
```

packages/react/test/specs/themes/teams/components/Icon/Icon-test.ts
```diff
@@ -0,0 +1,36 @@
...
+    const component = shallow(
+      iconFunc({ classes: { outlinePart: 'TEST-OUTLINE', filledPart: 'TEST-FILLED' } }),
```
layershifter: Actually this a React component, correct?

```suggestion
      <SvgIcon classes={{ outlinePart: 'TEST-OUTLINE', filledPart: 'TEST-FILLED' }} />
```

Can we use JSX there?

layershifter (approved):

packages/react/src/themes/teams/components/Icon/svg/ProcessedIcons/icons-number-list.tsx
```diff
@@ -1,16 +1,18 @@
...
 import { TeamsProcessedSvgIconSpec } from '../types'
+import { teamsIconSlotClassNames } from '../teamsIconSlotClassNames'
 
 export default {
   icon: ({ classes }) => (
     <svg role="presentation" focusable="false" viewBox="8 8 16 16" className={classes.svg}>
       <g>
         <path
-          className={classes.outlinePart}
+          className={cx(teamsIconSlotClassNames.outline, classes.outlinePart)}
```
kuzhelov: don't like the fact that this class name data is repeated for every icon while, in fact, it provides the same information as `classes.outlinePart`. We might think about some way to provide `classes` via styles, via dedicated prop - so that this repetition will be avoided. 

```js
const iconStyles = ({ ... }) => ({
  root: ....,
  outlinePart: {
     ...
      classes: ['outline']
  }
})
```

Note that this will be quite safe move (as it is quite easy to filter out this prop before Fela transformations), as well as we will still preserve full control of this aspect being provided to the client's hands (so it will be up to client to decide on class names policy, extend this set, etc)

packages/react/src/themes/teams/components/Icon/svg/ProcessedIcons/icons-accept.tsx
```diff
@@ -1,16 +1,18 @@
...
 import { TeamsProcessedSvgIconSpec } from '../types'
+import { teamsIconSlotClassNames } from '../teamsIconSlotClassNames'
 
 export default {
   icon: ({ classes }) => (
     <svg role="presentation" focusable="false" viewBox="8 8 16 16" className={classes.svg}>
       <g>
         <path
-          className={classes.outlinePart}
+          className={cx(teamsIconSlotClassNames.outline, classes.outlinePart)}
```
kuzhelov: formally these are not _slot_ class names - maybe we could rename them to just `teamsIconClassNames`? This will be pimplier, won't rely in Stardust's approach to naming that might eventually change and easier to consume on the client side - what do you think?

kuzhelov (approved):

PR #837, opened 2019-02-04
chore(package): get working `paths` option for ts-node

layershifter:
A small cut from my ongoing Lerna's PR, fixes the import issue from https://github.com/stardust-ui/react/pull/817#discussion_r252831302.

Adds `tsconfig-paths` as `ts-node` doesn't handle `compilerOptions.paths`:
- https://github.com/TypeStrong/ts-node/issues/138
- https://github.com/TypeStrong/ts-node/pull/254


docs/src/examples/components/Dropdown/Types/DropdownExampleSingleSelection.shorthand.steps.ts
```diff
@@ -1,4 +1,4 @@
-import { Dropdown, Button, List } from '../../../../../../packages/react/src'
+import { Dropdown, Button, List } from '@stardust-ui/react'
```
kuzhelov, line 1: 👍

gulpfile.ts
```diff
@@ -1,10 +1,18 @@
...
+import * as tsPaths from 'tsconfig-paths'
+
+const { compilerOptions } = require('./build/tsconfig.common.json')
 
 // add node_modules/.bin to the path so we can invoke .bin CLIs in tasks
 process.env.PATH =
   process.env.PATH + path.delimiter + path.resolve(__dirname, 'node_modules', '.bin')
 
+tsPaths.register({
+  baseUrl: '.',
```
kuzhelov: shouldn't we use repo's path from config constant here as well? This will ensure consistency in the approach taken (there will be always single source of truth maintained)

kuzhelov (commented):
do like the changes, but would like to get better story on why this functionality wasn't implemented as part of `ts-node` itself (https://github.com/TypeStrong/ts-node/pull/254#issuecomment-269404787) - just to properly assess our risks here (if there are any). Will return back shortly

gulpfile.ts
layershifter: Did it 👍

kuzhelov (approved):
yep, even given that import statements transpilation strategy of this plugin is absolutely different from the strategy of TS compiler,  it does exactly the thing we would like to have in these scenarios

Issue #433, opened 2018-11-05
RFC: how should we structure our monorepo and packages


levithomason:
We will move to a monorepo and publish packages for state, styling, and accessibility at the least.  We also have utilities such as our factories, which may be part of a "core" package.  What packages do we foresee and what modules should be included in them?

***

This list is updated per the discussion below:

- [ ] **`@stardust-ui/prop-types`**
    - use: `/lib/customPropTypes.tsx`
    - [x] Move to `/packages`
    - [ ] Reconcile Stardust changes with SUIR
    - [ ] Write README.md
    - [x] Publish

- [ ] **`@stardust-ui/auto-controlled-component`**
    - use: `/lib/AutoControlledComponent.tsx`
    - [ ] Move to `/packages`
    - [ ] Reconcile Stardust changes with SUIR
    - [ ] Write README.md
    - [ ] Publish

layershifter:
#### `custom-proptypes`

I think that we should move them to a separate package, will allow us to drop this package for production builds at all.

#### `auto-controlled-component` 

`AutoControlledComponent` is also used in SUI, this will allow us to start code-sharing. It also self-sufficient component.

#### `event-stack`

I have already rewritten `EventStack` to TS and it's a separate package, it also has a nice React component that makes life easier and less complex code (see changes in https://github.com/Semantic-Org/Semantic-UI-React/pull/2990). We can simply adopt it there, however I should note that it has own build and test system (browser based testing with Mocha in IE 11, Chrome & Firefox). I'm not that it fits monorepo ideas. May be, we can move it just to the org.

https://github.com/layershifter/event-stack

#### `props-utils`

Not sure about this, but... `getUnhandledProps()`, `partitionHTMLProps()` and `getElementType()` can be used and are used outside SUIR and Stardust.

bmdalex:
#### `Slot` and `createSlotFactory`

We are using these to create component slots (`content`, `icon`, `label` are just a few examples) and we might want to export them from the monorepo since they have the behavior of a factory of (sub)components;

we might want to move these to `src/lib/` side of the object (currently the are in `src/components/Slot.tsx`)

levithomason:
It seems that we might want to separate the React packages from the vanilla/core packages as well.

Slot / createSlotFactory, agree that we should publish something like this.  I think we need to iterate that again before doing so though.  Example issue, we call it "slot" but then we use it as the root `wrapper` element in the `Input`.  Really, this is just a "generic stardust element factory".

[...]

PR #838, opened 2019-02-04
chore: Setup Lerna

layershifter:
Refs #433.

This PR:
- configures Lerna and uses it for releases

---

- [x] `yarn build`
- [x] `yarn perf`
- [x] `yarn start`
- [x] `yarn stats`
- [x] `yarn test`
- [x] test release process
  - https://www.npmjs.com/package/stardust-ui-test-react
  - https://www.npmjs.com/package/stardust-ui-test-react-proptypes
  - [x] `patch` / `major`


.circleci/config.yml
```diff
@@ -91,8 +91,7 @@ jobs:
           command: |
             if [ -n "${npm_TOKEN+x}" ] && [ "${CIRCLE_BRANCH}" == "master" ]; then
               echo "//registry.npmjs.org/:_authToken=${npm_TOKEN}" > ~/project/.npmrc
-              npm version prerelease --preid="${CIRCLE_BUILD_NUM}.${CIRCLE_BRANCH}.`git rev-parse --short HEAD`" --no-git-tag-version
-              npm publish --tag current
+              yarn lerna publish --canary --yes
```
layershifter: If current is `0.19.2`, it will release for example `0.19.3-alpha.5+18e079d3`

build/gulp/tasks/bundle.ts
```diff
@@ -2,70 +2,79 @@ import { task, series, parallel, src, dest } from 'gulp'
...
 })
 
-task('build:dist:es', () => {
+task('bundle:build:es', () => {
   const tsConfig = paths.base('build/tsconfig.es.json')
   const settings = { declaration: true }
-  const typescript = g.typescript.createProject(tsConfig, settings)
 
-  const { dts, js } = src(componentsSrc).pipe(typescript())
-  const types = src(paths.base('types/**'))
```
layershifter, line 56: We don't need to bundle `types` folder now as `types.ts` is placed in `packages/react/src` directory.

build/gulp/tasks/docs.ts
```diff
@@ -245,7 +245,7 @@ task('watch:docs', cb => {
     .on('unlink', path => handleWatchUnlink('component-menu-behaviors', path))
 
   // rebuild images
-  watch(`${config.paths.src()}/**/*.{png,jpg,gif}`, series('build:docs:images')).on(
+  watch(`${config.paths.docsSrc()}/**/*.{png,jpg,gif}`, series('build:docs:images')).on(
```
layershifter, line 245: We don't have images in `src` folder 👀

build/gulp/tasks/bundle.ts
```diff
@@ -2,70 +2,79 @@ import { task, series, parallel, src, dest } from 'gulp'
...
 })
 
-task('clean:dist', parallel('clean:dist:es', 'clean:dist:commonjs', 'clean:dist:umd'))
+task('bundle:clean', parallel('bundle:clean:es', 'bundle:clean:commonjs', 'bundle:clean:umd'))
 
 // ----------------------------------------
 // Build
 // ----------------------------------------
-const componentsSrc = [paths.src('**/*.{ts,tsx}'), `!${paths.src('**/umd.ts')}`]
+const createComponentSrc = () => [
```
kuzhelov: why delayed execution (func) necessary to get this array?

layershifter: If we will execute just `yarn gulp`, the `packageName` variable will undefined and we will get an error here.

---

We can revert this and modify code:
```diff
-const packageName: string = argv.package
+const packageName: string = argv.package || ''
```

build/gulp/tasks/bundle.ts
```diff
@@ -2,70 +2,79 @@ import { task, series, parallel, src, dest } from 'gulp'
...
 
 import config from '../../../config'
+import sh from '../sh'
 
 const { paths } = config
 const g = require('gulp-load-plugins')()
 const { log, PluginError } = g.util
 
+const buildUMD = !!argv.umd
+const packageName: string = argv.package
```
kuzhelov: lets introduce necessary assertions and provide meaningful error description to the client when this value is not provided (otherwise it might be hard after a while to trace the exact problem why `gulp bundle:build:commonjs` has failed). As an alternative we might introduce default value here (as stardust-react is, essentially, the one that is implied quite often)

config.ts
```diff
@@ -34,11 +33,14 @@ const base = (...args) => path.resolve(...[envConfig.path_base, ...args])
 
 const paths = {
   base,
-  src: base.bind(null, envConfig.dir_src),
-  dist: base.bind(null, envConfig.dir_dist),
   dll: base.bind(null, envConfig.dir_dll),
   docsDist: base.bind(null, envConfig.dir_docs_dist),
   docsSrc: base.bind(null, envConfig.dir_docs_src),
+  packageDist: (packageName: string, ...paths: string[]) =>
```
kuzhelov, line 39: would expect `dist` artefacts being placed somewhere not in the `packages` dir - essentially, it might be the same hierarchy of directories (by project name) provided for the top-level project `dist` (like `dist/react`, ...)

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -15,7 +15,7 @@ import { constants } from 'src/lib'
 
 type ComponentMenuItem = { displayName: string; type: string }
 
-const pkg = require('../../../../package.json')
+const pkg = require('../../../../packages/react/package.json')
```
kuzhelov, line 18: 👍

config.ts
layershifter: Actually, I'm not sure that it's possible at all. Lerna will call NPM release process that requires to place files in the package folder. Did I miss something?

kuzhelov: no, you are not missing anything - while it is possible, it would require us to hoist paths specified in `package.json` (`files: '../dist/react'`), and then we have to select between the following:
- current approach, where all the packages are isolated from each other (although, yes, may be populated with dist artefacts)
- alternative one, where the isolation will be, essentially, not that strict, as hoisted (`../dist/react/`) file paths will be provided in the `package.json`

Thus, yes, would totally support taken approach - and if there will be true concerns around `dist` location, we might rethink. Thank you!

.circleci/config.yml
```diff
@@ -86,8 +86,7 @@ jobs:
...
```
kuzhelov: could we preserve `preid` value used before, like
```
lerna publish --canary --preid "${CIRCLE_BUILD_NUM}.${CIRCLE_BRANCH}"

or at least

lerna publish --canary --preid "${CIRCLE_BRANCH}"
```

layershifter: Sure, done 👍 

Will give us: `0.19.3-%CIRCLE_BUILD_NUM%.%CIRCLE_BRANCH%.23+002c6557b`

build/gulp/tasks/bundle.ts
layershifter: Let's go with default value 👍

build/gulp/tasks/bundle.ts
```diff
@@ -2,70 +2,79 @@ import { task, series, parallel, src, dest } from 'gulp'
...
 
 import config from '../../../config'
+import sh from '../sh'
 
 const { paths } = config
 const g = require('gulp-load-plugins')()
 const { log, PluginError } = g.util
 
+const buildUMD = !!argv.umd
+const packageName: string = argv.package || 'react'
```
kuzhelov, line 14: 👍

.circleci/config.yml
```diff
@@ -94,8 +94,7 @@ jobs:
           command: |
             if [ -n "${npm_TOKEN}" ] && [ "${CIRCLE_BRANCH}" == "master" ]; then
               echo "//registry.npmjs.org/:_authToken=${npm_TOKEN}" > ~/project/.npmrc
-              npm version prerelease --preid="${CIRCLE_BUILD_NUM}.${CIRCLE_BRANCH}.`git rev-parse --short HEAD`" --no-git-tag-version
-              npm publish --tag current
+              yarn lerna publish --canary --preid "${CIRCLE_BUILD_NUM}.${CIRCLE_BRANCH}" --yes
```
kuzhelov, line 100: just something to think about in the future - not absolutely sure that we should publish _all_ the packages after each build - actually, even now it creates lots of versions (for `react` package) that we are not consuming. I would see that there will be some trigger logic there (that might rely in the commit message) that would be required to express _explicit_ intent to publish package - and, otherwise, it shouldn't be published at all.

build/gulp/tasks/bundle.ts
```diff
@@ -2,70 +2,79 @@ import { task, series, parallel, src, dest } from 'gulp'
...
-    dts.pipe(dest(paths.dist('es'))),
-    js.pipe(dest(paths.dist('es'))),
-    types.pipe(dest(paths.dist('types'))),
+    dts.pipe(dest(paths.packageDist(packageName, 'es'))),
+    js.pipe(dest(paths.packageDist(packageName, 'es'))),
   ])
 })
 
-task('build:dist:umd', cb => {
+task('bundle:build:umd', cb => {
```
kuzhelov: actually, this might be very confusing if one would try to run

```
gulp bundle:build:umd
```

It seems odd that, for the sake of artefacts being generated, this command should be invoked with the following flag:

```
gulp bundle:build:umd --umd
```

build/gulp/tasks/bundle.ts
```diff
@@ -88,10 +97,11 @@ task('build:dist:umd', cb => {
   })
 })
 
-task('build:dist', parallel('build:dist:commonjs', 'build:dist:es', 'build:dist:umd'))
+task('build:build', parallel('bundle:build:commonjs', 'bundle:build:es', 'bundle:build:umd'))
```
kuzhelov: probably, `bundle:build` - or this name was chosen intentionally (sorry if I've missed the reason - could you, please, clarify?)

layershifter: Typo, will fix

build/gulp/tasks/bundle.ts
layershifter: Will rework it to separate tasks:
- `gulp bundle --package react`
- `gulp bundle:with-umd --package react`

packages/react/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "types": "dist/es/index.d.ts",
+  "sideEffects": false,
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "gulp bundle --package react --umd"
+  },
+  "dependencies": {
+    "@stardust-ui/react-proptypes": "^0.20.0",
```
kuzhelov: just to be sure (although, pretty much confident you've testes `workspaces` feature) - is it the case that both `yarn install` and `lerna bootstrap` properly link inner dependencies?

packages/react/src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -1,3 +1,4 @@
+import * as customPropTypes from '@stardust-ui/react-proptypes'
```
kuzhelov: 👍

tsconfig.json
```diff
@@ -1,12 +1,4 @@
...
-  "include": [
-    "docs/src",
-    "build",
-    "gulp",
-    "packages/react/src",
-    "packages/react/test",
-    "perf/src",
-    "types"
-  ]
+  "include": ["docs/src", "build", "gulp", "packages/*/src", "packages/*/test", "perf/src", "types"]
```
kuzhelov: here is actually a point of my concerns: currently what do we have is that each repo will have the same `tsconfig` - while, at the same time, it would be more beneficial to split these configs for different packages (and, probably, leave the absolutely bare necessary `tsconfig.common` at the root level):

### now
```
tsconfig.common.json
tsconfig.json
  packages
     react
     themes
     ...
```

### better?

```
tsconfig.common.json
  packages
     react
        tsconfig.json
     themes
        tsconfig.json
     ...
```

Here are the reasons why it would be beneficial:
- each project will have an autonomy on introducing specific configuration aspects that are project-specific (like 'include' prop)
- it will make packages independent of each other in the aspect of TS transform. Just, as a matter of contrived example, but still - what if there will be a package that is not written in TS? In that case we won't like it even being considered by TS compiler
- we will improve perf of individual builds - as only the paths that are relevant to the project will be included

Please, let me know what you think

layershifter: Yep, I fully support you're there. I'm going to do this in next PRs

packages/react/package.json
layershifter: Yes, `yarn install` correctly links all dependencies. You can check it because `@stardust-ui/react-proptypes` is not published yet :)

layershifter: When I tried to change the package version (non-local):

image

packages/react-proptypes/test/customPropTypes-test.ts
```diff
@@ -1,4 +1,4 @@
-import { customPropTypes } from 'src/lib'
+import * as customPropTypes from '@stardust-ui/react-proptypes'
```
kuzhelov: stardustPropTypes?

packages/react-proptypes/src/index.ts
```diff
@@ -1,8 +1,9 @@
 import * as _ from 'lodash'
 import * as PropTypes from 'prop-types'
+
 import leven from './leven'
-import { ObjectOf } from '../types'
 
+type ObjectOf<T> = { [key: string]: T }
```
kuzhelov: nit: `Record<string, T>`? :) but please, feel free to reason by yourself if that should be changed

packages/react-proptypes/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@stardust-ui/react-proptypes",
```
kuzhelov: actually, not absolutely convinced that we need to make this split now - could you, please, suggest what are the reasons to introduce this package now? Maybe we could wait for a while with this move - or there is already some client project exists that needs to consume them?

lerna.json
```diff
@@ -0,0 +1,6 @@
+{
+  "npmClient": "yarn",
+  "packages": ["packages/*"],
+  "version": "0.20.0",
+  "useWorkspaces": true
```
kuzhelov, line 5: nit-nit: may suggest to put this option near to `npmClient`, as those are closely related

build/webpack.config.umd.ts
```diff
@@ -1,15 +1,15 @@
 import config from '../config'
 import webpackConfig from '../webpack.config'
 
-const pkg = require('../package.json')
+const pkg = require('../packages/react/package.json')
```
kuzhelov: seems that intent was to require `pkg` by `packageName` as well (instead of `react` being hardcoded)

build/tsconfig.commonjs.json
```diff
@@ -3,5 +3,5 @@
   "compilerOptions": {
     "module": "commonjs"
   },
-  "include": ["../packages/react/src", "../types"]
+  "include": ["../packages/*/src", "../types"]
```
kuzhelov: ideally it would be great to avoid these expensive globs

build/jest/jest.config.common.js
```diff
@@ -8,9 +11,12 @@ module.exports = {
     '^.+\\.(ts|tsx)$': 'ts-jest',
   },
   moduleNameMapper: {
-    'docs/(.*)$': `${__dirname}/../../../docs/$1`,
-    'src/(.*)$': `${__dirname}/../src/$1`,
-    'test/(.*)$': `${__dirname}/../test/$1`,
+    ...lernaAliases(),
+
+    // Legacy aliases, they should not be used in new tests
```
kuzhelov, line 17: just to be clear - it means that we've left it as a follow-up work (I mean, tests grooming), right? In that case, could you, please, create a note in the PR description (or do whatever is necessary), so we won't forget it

build/gulp/tasks/test-projects.tsx
```diff
@@ -16,11 +17,33 @@ const log = (context: string) => (message: string) => {
...
 
-export const publishPackage = async () => {
-  const filename = tmp.tmpNameSync({ prefix: 'stardust-', postfix: '.tgz' })
-  await sh(`yarn pack --filename ${filename}`)
+const installStardustPackages = async (tmpDirectory: string) => {
+  const stardustPackages = lernaAliases({ sourceDirectory: false })
+  const packedPackages = await Promise.all(
+    Object.keys(stardustPackages).map(
+      async (packageName: string): Promise<{ packageName: string; filename: string }> => {
+        const filename = tmp.tmpNameSync({ prefix: `stardust-`, postfix: '.tgz' })
```
kuzhelov: lets include `packageName` in `prefix`

build/gulp/tasks/test-projects.tsx
```diff
@@ -16,11 +17,33 @@ const log = (context: string) => (message: string) => {
   console.log('='.repeat(80))
 }
 
-export const publishPackage = async () => {
-  const filename = tmp.tmpNameSync({ prefix: 'stardust-', postfix: '.tgz' })
-  await sh(`yarn pack --filename ${filename}`)
+const installStardustPackages = async (tmpDirectory: string) => {
+  const stardustPackages = lernaAliases({ sourceDirectory: false })
```
kuzhelov: nit: comment with the intent of specifying this option would be helpful, even simple like

```
// packages/react/src -> packages/react, 
// as lernaAliases append 'src'  by default
```

build/gulp/tasks/test-projects.tsx
```diff
@@ -16,11 +17,33 @@ const log = (context: string) => (message: string) => {
...
 
-  return filename
+        await runIn(directory)(`yarn pack --filename ${filename}`)
+
+        return { packageName, filename }
+      },
+    ),
+  )
+
+  const packageJsonName = path.resolve(tmpDirectory, 'package.json')
```
kuzhelov: name of the var is a bit misleading - could we use either `packageJsonFileName` or `packageJsonPath`?

build/gulp/tasks/test-projects.tsx
```diff
@@ -16,11 +17,33 @@ const log = (context: string) => (message: string) => {
...
+  const packageJson = require(packageJsonName)
+
+  packageJson.resolutions = packageJson.resolutions || {}
+  packedPackages.forEach(({ packageName, filename }) => {
+    packageJson.resolutions[`**/${packageName}`] = `file:${filename}`
+  })
+
+  fs.writeFileSync(packageJsonName, JSON.stringify(packageJson, null, 2))
+
+  const filenames = packedPackages.map(({ filename }) => filename).join(' ')
```
kuzhelov: `filenames` -> `packageNames` (or `stardustPackageNames`)?

build/gulp/tasks/test-projects.tsx
```diff
@@ -16,11 +17,33 @@ const log = (context: string) => (message: string) => {
   console.log('='.repeat(80))
 }
 
-export const publishPackage = async () => {
-  const filename = tmp.tmpNameSync({ prefix: 'stardust-', postfix: '.tgz' })
-  await sh(`yarn pack --filename ${filename}`)
+const installStardustPackages = async (tmpDirectory: string) => {
```
kuzhelov: we might name the input parameter as `testProjectDir` - this will be more descriptive and better reflect the expectations of this function. Also, it will make these lines easier to reason about:

```js
 const packageJsonName = path.resolve(tmpDirectory, 'package.json')
 const packageJson = require(packageJsonName)
```

vs 

```js
 const testProjectPackageJson = require(path.resolve(testProjectDir, 'package.json'))
```

kuzhelov: and, frankly, it seems that this function has more responsibilities than it should - we could split it on the following ones, so that it will be easier to read the sequence of actions in the test method itself (client of this function), something like:

`installStardustPackages` -> 
- `packStardustPackages()`
- `addResolutionPathsForStardustPackages()`
  - this is a new step necessary
- `yarn add @stardust-ui/react` 
  - as this is the only thing that we would like to reference as dependency

kuzhelov (commented):
nice job 👍  

there are just couple fo comments I see important to be resolved/discussed (apart of minor ones and nits):

- should we introduce separate package for prop types now?
- tsconfig files - how those should be organised?

Also, it would be great if the overall changes strategy, as well as the reasoning behind selecting the one, would be briefly reflected in the description - just so that it would be easy to see what these changes are about, as well as what is the overall vision in this area. Specifically, something like: "isolate packages, use yarn workspaces for resolution and linking project packages, use lerna for publishing" - with a bit more details introduced for each. 

Thank you!

packages/react-proptypes/src/index.ts
layershifter: 👍

build/gulp/tasks/test-projects.tsx
layershifter: Picked `stardustPackageFilenames` because they are packed packages file names.

build/gulp/tasks/test-projects.tsx
layershifter: Used `testProjectDir` as the param name 👍

build/gulp/tasks/test-projects.tsx
```diff
@@ -77,8 +109,10 @@ task('test:projects:cra-ts', async () => {
   //////// ADD STARDUST AS A DEPENDENCY ///////
   logger('STEP 2. Add Stardust dependency to test project..')
 
-  await runInTestApp(`yarn add ${packageFilename}`)
-  logger("Stardust is successfully added as test project's dependency.")
+  const packedPackages = await packStardustPackages()
+  await addResolutionPathsForStardustPackages(testAppPath(), packedPackages)
+  await installStardustReactPackage(testAppPath(), packedPackages)
```
kuzhelov: maybe we could just inline the command here? 

```
runIn(testProjectDir)(`yarn add ${packedPackages['@stardust-ui/react']}`)
```

or generalize in a way

```
installPackage(testAppPath(), packedPackages['@stardust-ui/react'])
```

To me both these options seem to be more expressive when read - but, please, feel free to decide whether change should be introduced

build/gulp/tasks/test-projects.tsx
```diff
@@ -16,14 +19,46 @@ const log = (context: string) => (message: string) => {
...
+      const filename = tmp.tmpNameSync({ prefix: `stardust-`, postfix: '.tgz' })
+      const directory = stardustPackages[packageName]
+
+      await runIn(directory)(`yarn pack --filename ${filename}`)
+
+      stardustPackages[packageName] = filename
+    }),
+  )
+
+  return stardustPackages
```
kuzhelov, line 56: it was helpful for debugging purposes to have paths (previously only the path for 'react' package) being displayed on the console

build/gulp/tasks/test-projects.tsx
```diff
@@ -61,9 +96,6 @@ task('test:projects:cra-ts', async () => {
   const logger = log('test:projects:cra-ts')
   const scaffoldPath = paths.base.bind(null, 'build/gulp/tasks/test-projects/cra')
 
-  const packageFilename = await publishPackage()
-  logger(`✔️Package was published: ${packageFilename}`)
```
kuzhelov, line 63: it was helpful for debugging purposes to have paths (previously only the path for 'react' package) being displayed on the console - could we preserve this (in the corresponding step)?

build/gulp/tasks/bundle.ts
```diff
@@ -88,10 +91,10 @@ task('build:dist:umd', cb => {
...
-task('build:dist', parallel('build:dist:commonjs', 'build:dist:es', 'build:dist:umd'))
-
 // ----------------------------------------
 // Default
 // ----------------------------------------
 
-task('dist', series('clean:dist', 'build:dist'))
+task('bundle', series('bundle:clean', parallel('bundle:build:commonjs', 'bundle:build:es')))
+task('bundle:with-umd', series('bundle', 'bundle:build:umd'))
+task('bundle:all', () => sh('lerna run build'))
```
kuzhelov: frankly, was a bit lost in naming convention around `bundle` - shouldn't it be `bundle:build:all`? It seems that this will even aid resolution thought on differences between `bundle:with-umd` and `bundle:all`

This might be also appealing that all the places where `bundle:all` is used are, essentially, `build` tasks

build/webpack.config.umd.ts
```diff
@@ -1,36 +1,39 @@
...
+    devtool: false,
+    entry: {
+      [pkg.name]: paths.packageSrc(packageName, 'umd.ts'),
+    },
+    externals: {
+      react: 'React',
+      'react-dom': 'ReactDOM',
+    },
+    output: {
+      filename: 'stardust-ui-react.min.js',
```
kuzhelov: we should reflect package name here as well
```js
`stardust-ui-${packageName}.min.js`
```

kuzhelov (approved):
actually, changes look good to be merged to me - just left couple of additional comments. Also, it would be great to have someone else's review before merge  (@Bugaa92, @mnajdova, @miroslavstastny)

build/webpack.config.umd.ts
layershifter: 👍

build/gulp/tasks/test-projects.tsx
layershifter: Updated 👍

build/gulp/tasks/test-projects.tsx
layershifter: Updated:
```
test:projects:rollup : ✔️Package "@stardust-ui/react" was packed to C:\Users\olfedias\AppData\Local\Temp\stardust-18148
```

build/gulp/tasks/bundle.ts
layershifter: Renamed:
- `bundle` => `bundle:package`
- `bundle:with-umd` => `bundle:package:with-umd`
- `bundle` => `bundle:all-packages`

Should be much clear now 👍

build/gulp/tasks/test-projects.tsx
```diff
@@ -16,14 +19,42 @@ const log = (context: string) => (message: string) => {
...
+  // as lernaAliases append 'src'  by default
+  const stardustPackages = lernaAliases({ sourceDirectory: false })
+
+  await Promise.all(
+    Object.keys(stardustPackages).map(async (packageName: string) => {
+      const filename = tmp.tmpNameSync({ prefix: `stardust-`, postfix: '.tgz' })
+      const directory = stardustPackages[packageName]
+
+      await runIn(directory)(`yarn pack --filename ${filename}`)
+      logger(`✔️Package "${packageName}" was packed to ${filename}`)
```
kuzhelov, line 50: 👍

build/gulp/tasks/bundle.ts
```diff
@@ -2,70 +2,73 @@ import { task, series, parallel, src, dest } from 'gulp'
...
 })
 
-task('clean:dist:umd', cb => {
-  rimraf(`${config.paths.dist()}/umd/*`, cb)
+task('bundle:clean:umd', cb => {
+  rimraf(`${config.paths.packageDist(packageName)}/umd/*`, cb)
 })
 
-task('clean:dist', parallel('clean:dist:es', 'clean:dist:commonjs', 'clean:dist:umd'))
+task('bundle:clean', parallel('bundle:clean:es', 'bundle:clean:commonjs', 'bundle:clean:umd'))
```
kuzhelov: still a bit confused with the naming, sorry :) Here is the reason for that - in general the following structure for task names is suggested:

`bundle:package:...` - package specific tasks

At the same time we have

`bundle:clean` - and this one turns out to be package-specific as well (this moment is quite confusing)

`bundle:all-packages` - actually, we could safely rename it to `bundle:all`

### Proposal
As a proposal I might suggest something like the following (at the same time, please, feel free to introduce another approach - there are only aforementioned points I would like to be addressed):

```
bundle:package - builds all module resolution versions (I would include umd here)
bundle:package:no-umd - this one will be used in the build chain

bundle:package:clean
bundle:package:clean:umd

bundle:build:all (or bundle:build) - the sole reason I won't suggest to use 'bundle:all' is because the actual action taken as part of this task is different, so that 'bundle:all !== bundle:package for each package', according to the task's logic
```

layershifter: Updated all expect `bundle:all-packages`, it's very clear what task actually does

build/gulp/tasks/bundle.ts
```diff
@@ -88,10 +94,14 @@ task('build:dist:umd', cb => {
...
 // ----------------------------------------
 
-task('dist', series('clean:dist', 'build:dist'))
+task(
+  'bundle:package:no-umd',
+  series('bundle:package:clean', parallel('bundle:package:commonjs', 'bundle:package:es')),
+)
+task('bundle:package', series('bundle:package:no-umd', 'bundle:package:umd'))
+
+task('bundle:all-packages', () => sh('lerna run build'))
```
kuzhelov, line 107: ok, lets compromise on that for now :) in either case, this is quite easy thing to change _in case if needed_. Thank you!

.circleci/config.yml
```diff
@@ -97,8 +97,7 @@ jobs:
...
```
miroslavstastny, line 100: Don't you want to add `--skip-git`?

package.json
```diff
@@ -28,9 +18,9 @@
     "postcommit": "git update-index --again",
     "prerelease": "yarn ci && cross-env NODE_ENV=production yarn build",
     "postrelease": "yarn deploy:docs",
-    "release:major": "yarn prerelease && ta-script npm/release major && yarn postrelease",
-    "release:minor": "yarn prerelease && ta-script npm/release minor && yarn postrelease",
-    "release:patch": "yarn prerelease && ta-script npm/release patch && yarn postrelease",
+    "release:major": "yarn prerelease && lerna publish major && yarn postrelease",
```
miroslavstastny, line 21: Remove `ta-scripts` dependency?

miroslavstastny: `ta-script npm/release` does several things - sanity checks, pushes git tag... are you sure `lerna publish` does all these?

miroslavstastny (commented):
Added some comments re things I am not sure about.

You should also update `CONTRIBUTING.md` to describe how to add/remove packages + some basic rules on what should go to main `package.json` vs `react/package.json`

It would be great if @levithomason could briefly check this as well.

package.json
layershifter: > Remove `ta-scripts` dependency?

Yep, nice catch 👍 

> sanity checks, pushes git tag... are you sure `lerna publish` does all these?

An even more, you can't publish without changes: `lerna notice Current HEAD is already released, skipping change detection`
> pushes git tag: `lerna info git Pushing tags...`

https://github.com/layershifter/react/releases
https://github.com/layershifter/react/commit/308f7e75ea32a8ebf952b28eed94f613bc59b355

.circleci/config.yml
layershifter:
```
Successfully published:
 - stardust-ui-test-react@2.0.1-alpha.2+1f3c4b76
```

It will not create git tag or a version commit.

layershifter:
> You should also update `CONTRIBUTING.md` to describe how to add/remove packages + some basic rules on what should go to main `package.json` vs `react/package.json`

Going to make it in a next PR with extracting `customPropTypes`.

miroslavstastny (approved):

PR #842, opened 2019-02-05
feat(Dropdown): arrow navigation between selected items

silviuaavram:
Allows the navigation between Dropdown selected items using the Arrow Left and Arrow Right keys. Dropdown must be a multiple selection.

Removed the `tabIndex=0` from `Label` as it was not needed for click. @mnajdova please confirm. As a result, 'X' icons are no longer in the tab order, they can only be clicked in order to remove the item. Focus is returned to the search input at removal.

Selected items are focusable now, but are not part of the normal tab order. Can be focused by click or by left/right arrow keys.

- Left Arrow from search input should focus last selected item.
- Right Arrow from last selected item should focus input.
- Left/Right Arrows should switch focus between the items, depending on the currently focused one.
- Delete/Backspace while focus is on a selected item should remove that item and put focus on the search input.

Added a style similar to Teams to the focused item (background colour change).

May need to rename `activeIndex` as I used it for consistency with Accordion/Menu but it may be a bit confusing, accepting feedback.

Adding screener test to check highlighted style on both keydown and click.


packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -638,8 +661,35 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
     ) => {
-      if (keyboardKey.getCode(e) === keyboardKey.Tab && !_.isNil(highlightedIndex)) {
-        selectItemAtIndex(highlightedIndex)
+      switch (keyboardKey.getCode(e)) {
+        case keyboardKey.Tab:
+          if (!_.isNil(highlightedIndex)) {
+            selectItemAtIndex(highlightedIndex)
+          }
+          break
+        case keyboardKey.ArrowLeft:
```
jurokapsiar, line 672: I don't see a handler for the right key, is that intentional? Also, in RTL, right and left keys need to be swapped.

jurokapsiar, line 672: In RTL mode, left/right arrow keys need to be swapped.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -759,10 +792,58 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+    predefinedProps: DropdownSelectedItemProps,
+    DropdownSelectedItemProps: DropdownSelectedItemProps,
+  ) {
+    const { activeIndex, value } = this.state as { activeIndex: number; value: ShorthandValue[] }
+    switch (keyboardKey.getCode(e)) {
+      case keyboardKey.Delete:
+      case keyboardKey.Backspace:
+        this.handleSelectedItemRemove(e, item, predefinedProps, DropdownSelectedItemProps)
+        break
+      case keyboardKey.ArrowLeft:
```
jurokapsiar: In RTL mode, left/right arrow keys need to be swapped. (applies to line 814 as well)

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: made changes. at some point we need to create behaviours.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -595,6 +611,10 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     statusDiv.textContent = statusMessage
   }
 
+  isSelectedItemActive = (index: number): boolean => {
```
bmdalex: private method, right?

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -638,8 +667,35 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
     ) => {
-      if (keyboardKey.getCode(e) === keyboardKey.Tab && !_.isNil(highlightedIndex)) {
-        selectItemAtIndex(highlightedIndex)
+      switch (keyboardKey.getCode(e)) {
+        case keyboardKey.Tab:
+          if (!_.isNil(highlightedIndex)) {
+            selectItemAtIndex(highlightedIndex)
+          }
+          break
+        case rtl ? keyboardKey.ArrowRight : keyboardKey.ArrowLeft:
```
bmdalex: For readability, I'd rather leave the switch case as a simple condition and extract the body of a case in a private method:

something like:

```typescript
private setActiveIndexToLastSelectedItem = () => {
  const { value } = this.state as { value: ShorthandValue[] }
  if (this.props.multiple && value.length > 0) {
    this.trySetState({ activeIndex: value.length - 1 })
  }
}

// in the switch/case
switch (keyboardKey.getCode(e)) {
  // ...
  case keyboardKey.ArrowLeft:
    if (!rtl) {
      this.setActiveIndexToLastSelectedItem()
    }
    break
  case keyboardKey.ArrowRight:
    if (rtl) {
      this.setActiveIndexToLastSelectedItem()
    }
    break
  // ...
}
```
what do you think?

packages/react/test/specs/components/Label/Label-test.tsx
```diff
@@ -13,14 +11,4 @@ describe('Label', () => {
   isConformant(Label)
   labelImplementsShorthandProp('icon', Icon, { mapsValueToProp: 'name' })
   labelImplementsShorthandProp('image', Image, { mapsValueToProp: 'src' })
-
-  describe('icon', () => {
```
bmdalex, line 17: can you please provide a reason for removing this test as a comment?

bmdalex (commented):
@silviuavram - generally looks good, one thing: as we discussed, we'd like have the key nav extracted in the behaviors, can you create an issue or if it already exists reference it in this PR? Thanks!

packages/react/test/specs/components/Label/Label-test.tsx
silviuaavram: what do you mean? I changed the Label not to have tabindex 0 anymore. This test was failing as a result. What am I missing>

bmdalex: got it, thx, fyi @mnajdova

mnajdova: In this case, keyboard-only user will still be able to perform the action that the regular user can perform on click? I don't see how we are handling this. It seems that now is user's responsibility to ensure that..

silviuaavram: we need to do it some other way. having tabindex 0 just because the icon has onClick is not really the way to go.
1. we don't need tabindex 0 to have it clickable.
2. yes, the user can add tabindex 0 to the shorthand. 
3. it's not a button by default, so it should not be in the tab order by default.
4. we can think of some other way to do it.

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: we will refactor this in the future, with behaviours, so any improvement made now will only get removed then. right now I think this is good enough.

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: yes, thanks!

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -47,6 +47,12 @@ export interface DropdownSlotClassNames {
 }
 
 export interface DropdownProps extends UIComponentProps<DropdownProps, DropdownState> {
+  /** The index of the currently active selected item, if dropdown has a multiple selection. */
+  activeIndex?: number
```
sophieH29: Agree that this prop name may be confusing as seems to be referencing to indexes in the list. `activeSelectedIndex`?

silviuaavram: it's a good suggestion :) @mnajdova @kuzhelov @Bugaa92 you agree?

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: ok, maybe I've changed my mind.

packages/react/src/components/Dropdown/Dropdown.tsx
mnajdova: I agree, it's not obvious in terms of the Dropdown component, so using more specific name helps 👍

silviuaavram: renamed as Sophie suggested, thanks!

bmdalex (approved):

docs/src/examples/components/Dropdown/Types/index.tsx
```diff
@@ -5,20 +5,25 @@ import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
 const Types = () => (
   <ExampleSection title="Types">
     <ComponentExample
-      title="Default"
+      title="Selection"
```
mnajdova, line 8: Why we rename the Default example? If the default is selection, better leave it as default :)

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -151,12 +157,12 @@ export interface DropdownProps extends UIComponentProps<DropdownProps, DropdownS
 }
 
 export interface DropdownState {
-  isOpen?: boolean
-  value: ShorthandValue | ShorthandValue[]
-  backspaceDelete: boolean
```
mnajdova, line 159: We don't need backspaceDelete anymore? Are we going to still have the functionality to remove selected items on backspace?

docs/src/examples/components/Dropdown/Types/index.tsx
silviuaavram: I added Selection as it's the same in Semantic. Default is a dropdown that just pops a list, and does not memorise selection :)

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: yes :D

silviuaavram: I think it's because I changed the handler from keyUp to keyDown. it worked, so I just removed that code as it was just a hack to begin with.

jurokapsiar (approved):

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -637,8 +663,27 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
       e: React.SyntheticEvent,
       searchInputProps: DropdownSearchInputProps,
     ) => {
-      if (keyboardKey.getCode(e) === keyboardKey.Tab && !_.isNil(highlightedIndex)) {
-        selectItemAtIndex(highlightedIndex)
+      switch (keyboardKey.getCode(e)) {
```
mnajdova, line 666: Please please create an issue for creating dropdownBehavior. It's really hard to keep track of everything that it happening here.

mnajdova (approved):

packages/react/src/components/Dropdown/Dropdown.tsx
sophieH29: @silviuavram  we can pair on creating behavior, to show how the behaviors work and action handlers are assigned.

silviuaavram: https://github.com/stardust-ui/react/issues/875

PR #846, opened 2019-02-05
fix(avatar): font icons sizing issue

kuzhelov:
This PR fixes appearance of `Icon` component when it is provided into `image` slot of `Avatar` component (originally reported here https://github.com/stardust-ui/react/pull/795): 

```jsx
<>
  { /* benchmark */ }
  <Avatar size='largest' ... /> 

  { /* with font-icon-based 'image' slot */ }
  <Avatar
    size='largest'
    image={<Icon size='largest' ... />}
  />
</>
```

## Was
image

## Now
image

image


packages/react/src/lib/callable.ts
```diff
@@ -1,6 +1,10 @@
+import { Func } from '../types'
+
 // https://jsperf.com/startdust-callable
```
kuzhelov, line 3: we may safely remove this reference as 
- it is pointing to stale document
- we have all the necessary tooling now to measure performance on the Stardust's side

packages/react/src/themes/teams/components/Icon/avatarIconStyles.ts
```diff
@@ -0,0 +1,34 @@
+import * as _ from 'lodash'
+
+import { pxToRem, SizeValue } from '../../../../lib'
+import { IconVariables } from './iconVariables'
+
+import { sizeToPxValue as avatarSizeToPxValue } from '../Avatar/avatarStyles'
+
+const sizeToIconPaddingInPx: Record<SizeValue, number> = {
```
kuzhelov, line 8: these values need to be refined

packages/react/src/lib/callable.ts
layershifter: Let's remove it 👍

packages/react/src/themes/teams/components/Avatar/avatarStyles.ts
```diff
@@ -1,8 +1,8 @@
 import { pxToRem } from '../../../../lib'
 import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
-import { AvatarProps } from '../../../../components/Avatar/Avatar'
+import { AvatarProps } from '../../../..'
```
layershifter, line 3: Can we still use the direct import?

kuzhelov: sure, thanks for catching this thing!

packages/react/src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -124,6 +125,10 @@ const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
       ...(!rtl && {
         transform: `rotate(${rotate}deg)`,
       }),
+
+      [`&.${Avatar.slotClassNames.image}`]: {
+        ...(isFontBased && getAvatarFontIconStyles(size, v)),
+      },
```
layershifter, line 131: Actually, this is a single thing that stops me from the approve button.

I understand _why_ `Avatar` can know about `Icon`, because it's a slot of `Avatar`. But, I want to avoid the reserve coupling because `Icon` is a primitive component  🤔

PR #847, opened 2019-02-05
chore: Run perf tests in CI

miroslavstastny:
Run `yarn perf` as part of Ci build, part of #773.
- CircleCi build now uses docker image with browsers,
- results are NOT stored to external DB, to be handled in a separate PR (after #821 is merged).

No changelog entry, no user-facing change.


.circleci/config.yml
```diff
@@ -6,7 +6,7 @@ general:
 jobs:
   build:
     docker:
-      - image: circleci/node:8
+      - image: circleci/node:8-browsers
```
layershifter, line 9: Not sure about this, `puppeteer` package will install an own browser. Do we need to switch images?

layershifter (approved):

.circleci/config.yml
miroslavstastny: Yes.
Puppeteer installs its own Chrome, but it still depends on some shared libraries which are not installed in the original image:
```
[02:10:55] 'perf:run' errored after 174 ms
[02:10:55] Error: Failed to launch chrome!
/home/circleci/project/node_modules/puppeteer/.local-chromium/linux-609904/chrome-linux/chrome: error while loading shared libraries: libXtst.so.6: cannot open shared object file: No such file or directory
```

bmdalex (approved):

kuzhelov (approved):

PR #852, opened 2019-02-06
feat(Provider): Make Provider a UIComponent

### LINKED ISSUE #586: Provider is not setting styles to theme
### Steps
Create react-app and add Stardust

```jsx
 // index.js

import React from 'react';
import ReactDOM from 'react-dom';
import './index.css';
import App from './App';
import \* as serviceWorker from './serviceWorker';
import { Provider, themes } from '@stardust-ui/react'

ReactDOM.render(
   <Provider theme={themes.teams}>
    <App />
  </Provider>,
   document.getElementById('root')
);
```

```jsx
// App.js

import React, { Component } from 'react';
import logo from './logo.svg';
import './App.css';
import { Text, Provider, themes, Button } from '@stardust-ui/react'

class App extends Component {
  render() {
    return (
      <div className="App">
        <header className="App-header">
          <img src={logo} className="App-logo" alt="logo" />
          <p>
            Edit <code>src/App.js</code> and save to reload.
          </p>
           <Text content="Learn React" />
           <Provider theme={ themes.teamsDark }> { /* <-------- !!! */ }
               <Text content="Build your future" />
           </Provider>
        </header>
      </div>
    );
  }
}

export default App;
```
image

### Expected Result
"Learn React" would be black text on white background - like teams theme
"Build your Future" would be white text on black background - as dark theme

### Actual Result

Theme styles are not visually applied at all.  Currently would need to wrap in segment, or apply manually from styling variables.

### Version
0.14.0

### Testcase

### Suggestion
The Provider should generate div and apply the theme related styles.  Theme could select which inherited styles they apply (colours, fonts, etc)

miroslavstastny:
## BREAKING CHANGE
`<Provider>` component now renders a `<div>` element with `dir` attribute and styles setting `background` and `color` CSS attributes.

Fixes #586.

## Change description
`<Provider>` component now creates a `<div>` element.

#### `Provider` can now contain multiple children

**Before**
```jsx
<Provider theme={themes.teamsDark}>
  <> <!-- React.Fragment or other explicit wrapping element required -->
    <p>Text inside p tag inside dark theme</p>
    <Button content="Stardust button inside dark theme"  />
  </>
</Provider>
```

**After**
```jsx
<Provider theme={themes.teamsDark}>
  <p>Text inside p tag inside dark theme</p>
  <Button content="Stardust button inside dark theme"  />
</Provider>
```

#### Theme colors (`bodyBackground` and `bodyColor`) are correctly applied for all children

**Code**
```jsx
<Provider theme={themes.teamsDark}>
  <p>Text inside p tag inside dark theme</p>
  <Button content="Stardust button inside dark theme"  />
</Provider>
```
**Before**
image

**After**
image

#### There is also `dir` attribute correctly set on the `div` so RTL works without any additional changes:

**Before**
```jsx
<div dir="rtl"> <!-- this div must be explicitly added for RTL to work -->
  <Provider theme={{rtl: true}}>
  </Provider>
</div>
```

**After**
```jsx
<Provider theme={{rtl: true}}> <!-- Provider renders div with dir attribute set -->
</Provider>
```

#### If `rtl` is not defined in the nested `Provider`, it is now inherited from the outer provider

```jsx
<Provider theme={{rtl: true}}> <!-- Provider renders div with dir attribute set -->
  <Provider theme={themes.teamsDark}> <!-- Rtl attribute not set, div is rendered without dir attribute, RTL is inherited from parent -->
  </Provider>
</Provider>
```

#### It is possible to override variables and styles for the `ProviderBox`

```jsx
<Provider variables={{background: 'red'}} styles={{borderBottom: '3px solid blue'}} theme={mergeThemes(themes.teamsDark, {
  siteVariables: {
    bodyBackground: 'salmon'
  },
  componentVariables: {
    ProviderBox: {
      color: 'yellow',
    }
  },
  componentStyles: {
    ProviderBox: {
      root: {
        borderTop: '3px solid green'
      }
    }
  }
})}>
  <p>with custom overrides</p>
</Provider>
```

**After**
image

## Known limitations
Docsite displays `div` created by the example-wrapping Provider as part of the Rendered HTML.


layershifter (commented):
### 1. `ProviderBox`

image

```jsx
<Provider as='div' theme={theme}>
```

This the most confusing part to me. Why we need this? I tried to use styles directly on `Provider` and use `as=div` and it works like a charm 🌸

packages/react/src/lib/renderComponent.tsx
```diff
@@ -155,7 +155,10 @@ const renderComponent = <P extends {}>(config: RenderConfig<P>): React.ReactElem
     <FelaTheme>
       {(theme: ThemePrepared) => {
         if (_.isEmpty(theme)) {
-          logProviderMissingWarning()
+          if (className !== 'ui-provider') {
```
```suggestion
          if (_.isEmpty(theme) && className !== 'ui-provider') {
```

packages/react/test/specs/components/Provider/Provider-test.tsx
```diff
@@ -53,4 +60,76 @@ describe('Provider', () => {
...
+
+  describe('RTL', () => {
+    test('Sets dir="rtl" on the div for RTL theme', () => {
+      const component = mount(
+        <Provider id="top-level-provider" theme={{ rtl: true }}>
+          <span />
+        </Provider>,
+      )
+      const providerDiv = component.find('div#top-level-provider')
+      expect(providerDiv.length).toEqual(1)
```
layershifter:
```suggestion
      expect(providerDiv.exists()).toBe(true)
```

layershifter (commented):
### 2. changes in `mergeRTL()`

This is a second point of my confusion.

```jsx
<Provider theme={{}} />
  <Provider theme={{}} />  // rtl: false
// ---
<Provider theme={{}} />
  <Provider theme={{ rtl: true }} /> // rtl: true
// ---
<Provider theme={{ rtl: true }} />
  <Provider theme={{}} /> // rtl: true
```

Why it should be `undefined`? It will add `dir='ltr'` to the root node, but why this is not correct?
Fela uses `direction` prop on `theme`,  https://github.com/rofrischmann/fela/pull/673. I think that we should do the same thing instead of `rtl: boolean`.

miroslavstastny:
> 2. changes in `mergeRTL()`

The original code works for merging themes in `Provider`, that's correct. Where it breaks is this code:
```js
teamsTheme = mergeThemes(base, {/*...*/})
```
RTL is defined neither in base theme nor in teams overrides but with the original code `mergeThemes()` results in `rtl: false`. And when you use this theme in nested provider it force-overwrites rtl to false and that's not what you want.

miroslavstastny:
> 1. `ProviderBox`

The reason to add `ProviderBox` was to make the `div` created by `Provider` themable. With this approach you can:
1. style the `div` by the `theme` passed as a prop to the `Provider`,
2. override the theme variables for this single div (without affecting div styles of any nested Provider) using `<Provider variables={{...}} ...>`,
3. override styles for this single div using `<Provider styles={{...}} ...>`.

The reason why we cannot do that directly in `Provider` is that Fela classes are computed in `renderComponent` function **before** `Provider.renderComponent()` is called but the new theme is merged in `Provider.renderComponent()`.

As discussed offline with @layershifter the main problems of the approach taken are:
- `Provider` should be as lightweight as possible,
- users can (accidentally) misuse `as` to do something like `<Provider as={Menu} ... />`,
- having a `ProviderBox` component together with `Box` component is confusing.

Alternative proposal (@layershifter) is:
- change `Provider` back to be a regular component, not `UIComponent`
  - that removes `as` = no confusion or misuse
- add `nodiv` (working name, not final) prop to `Provider` to add possibility to opt out of creating the div
- do not pass any props not handled by `Provider` to `ProviderBox` - this makes it impossible to override variables and styles for the `ProviderBox` when creating a `Provider` but when that is required, user can use `noDiv` and create her own `div`
- as a next step try to remove `ProviderBox` component and:
  - recompute Fela classes directly in `Provider.renderComponent()` after the themes are merged
  - create `div` directly in `Provider.renderComponent()`. Not `ElementType` as `Provider` is not `UIComponent` but hardcoded `div` (again, if that does not work for you, you can go with `nodiv`and create whatever element inside the Provider manually).

packages/react/src/components/Provider/Provider.tsx
```diff
@@ -152,9 +171,22 @@ class Provider extends React.Component<ProviderProps> {
...
+            (incomingTheme.rtl !== outgoingTheme.rtl && _.isBoolean(outgoingTheme.rtl))
+          ) {
+            rtlProps.dir = outgoingTheme.rtl ? 'rtl' : 'ltr'
+          }
+
           return (
             <RendererProvider renderer={outgoingTheme.renderer} {...{ rehydrate: false }}>
-              <ThemeProvider theme={outgoingTheme}>{children}</ThemeProvider>
+              <ThemeProvider theme={outgoingTheme}>
+                <ElementType {...unhandledProps} {...rtlProps} className={classes.root}>
```
levithomason: Let's make ElementType ProviderBox so that the user can't break the implementation.  Forward the as prop from the Provider to the ProviderBox.

packages/react/src/components/Provider/Provider.tsx
```diff
@@ -31,12 +37,18 @@ export interface ProviderProps {
 /**
  * The Provider passes the CSS in JS renderer and theme to your components.
  */
-class Provider extends React.Component<ProviderProps> {
-  staticStylesRendered: boolean = false
+class Provider extends UIComponent<ReactProps<ProviderProps>> {
```
levithomason: Once ElementType is replaced (see render) this no longer needs to be extending UIComponent.

👍 thanks Shift

packages/react/src/components/Provider/Provider.tsx
```diff
@@ -64,7 +76,14 @@ class Provider extends React.Component<ProviderProps> {
     children: PropTypes.element.isRequired,
   }
 
+  static defaultProps = {
+    as: ProviderBox,
+  }
```
levithomason, line 67: Can remove after undoing UIComponent

levithomason:
First shipment should be as minimal as possible.  We won't advertise opting out of the styled div until we get requests.  Even then, the initial workaround suggestion will be `as={React.Fragment}`.  However, let's remove it from the breaking change notice for now.

docs/src/app.tsx
```diff
@@ -19,6 +19,7 @@ class App extends React.Component<any, ThemeContextData> {
     return (
       <ThemeContext.Provider value={this.state}>
         <Provider
+          as={React.Fragment}
```
miroslavstastny, line 22: This is temporary hotfix for Docsite when switching themes for examples. We should fix theme switching to apply to examples only, not to the whole docsite.

docs/src/components/ExternalExampleLayout.tsx
```diff
@@ -61,32 +61,24 @@ class ExternalExampleLayout extends React.Component<
       <Provider key={this.state.renderId} theme={theme}>
         <Provider.Consumer
           render={({ siteVariables }) => (
-            <div
-              dir={theme.rtl ? 'rtl' : undefined}
-              style={{
-                color: siteVariables.bodyColor,
```
miroslavstastny, line 67: Reverted #1050, removed unnecessary div.

packages/react/src/lib/renderComponent.tsx
miroslavstastny: As `Provider` is no longer a UIComponent I reverted this change.

packages/react/src/components/Provider/ProviderBox.tsx
```diff
@@ -0,0 +1,32 @@
...
+  className: 'ui-provider__box',
+
+  propTypes: {
+    ...commonPropTypes.createCommon(),
+  },
+
+  render(config, props) {
+    const { ElementType, classes, unhandledProps } = config
+    const { children } = props
+    if (isFragment(<ElementType />)) {
```
layershifter:
```suggestion
    if (ElementType === React.Fragment) {
```

Do we really need `react-is` there?

miroslavstastny: Thanks! Definitely!

layershifter (approved):

PR #855, opened 2019-02-06
feat(Label): added black and white colors for the color prop

### LINKED ISSUE #739: Color Scheme for black color
## Feature Request
We have a scenario where we need a Label to have a black background
image

### Problem description
This does not exist in the current colorSchemes
image

### Proposed solution
Can we please add a color scheme for black color so that I can use it with Label component in my use case.

Also, the color black exported from siteVariables is actually grey and not full black.
`black: naturalColors.grey[900],`
Can we please export fullBlack : #000 in siteVariables as well.

### MVP

mnajdova:
This PR adds the black and white color options for the Label, in order to fix https://github.com/stardust-ui/react/issues/739  
Here is how all the variants for color look:
image

As the white and black are primitive colors, the color scheme for these colors is manually defined:
```
const primitiveColorsScheme = {
  black: {
    foreground: colors.white,
    border: colors.white,
    shadow: colors.white,
    background: colors.black,
  },
  white: {
    foreground: colors.black,
    border: colors.black,
    shadow: colors.black,
    background: colors.white,
  },
}
```
If there are any other ideas, please let me know.

# BREAKING CHANGES (Teams theme)
The value for the `siteVariables.colors.black` was changed from the `grey[900](#252424)` variant to `#000`. If there are usages of this variable, replaced them in the following manner:

`siteVariables.colors.black` <=> `siteVariables.colors.grey[900]`

Whenever you want to use the black color `(#000)` use the variable `siteVariables.colors.black`.


CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `actionMenu` prop to `ChatMessage` component @layershifter ([#811](https://github.com/stardust-ui/react/pull/811))
 - Add `rtl` field in the `SvgIconFuncArg`, and used it in Teams theme's number-list icon ([#851](https://github.com/stardust-ui/react/pull/851))
 - Add single search flavor for `Dropdown` component @Bugaa92 ([#839](https://github.com/stardust-ui/react/pull/839))
+- add `black` and `white` options for the `color` prop in the `Label` component @mnajdova ([#855](https://github.com/stardust-ui/react/pull/855))
```
kuzhelov: nit: lets, please, capitalise this sentence, for the sake of consistency. 

nit-nit:
```
prop in the `Label` component -> prop of the `Label` component
```

mnajdova: Done.

codepretty:
Are you going to update the references to the old `#252424` color of black as well? For example in `buttonVariables.ts`.

mnajdova:
@codepretty as far as I see in `buttonVariables.ts` we are using `siteVariables.black`, which still have the same value as before. I intentionally didn't change the value of that one as it is widely used, and I would introduce lots of breaking changes. If `#000` should be used anywhere, we can update the styles of the components one by one. Let me know if I misunderstood your question.

codepretty:
> @codepretty as far as I see in buttonVariables we are using siteVariables.black, which still have the same value as before. I intentionally didn't change the value of that one as it is widely used, and I would introduce lots of breaking changes. If #000 should be used anywhere, we can update the styles of the components one by one. Let me know if I misunderstood your question.

ah, i see. i missed the update to siteVariables you did to reference colors.grey[900]. do you think it will be confusing to devs using this to know that siteVars.black is different from colors.black without digging in?

mnajdova:
Yep, totally agree with the confusion, but will address this in a separate PR because I want this one to be simple. The basic idea I have in mind is that we should remove all colors we have in siteVariables, and use only the ones from the colors object. We will map all current color in the siteVariables with the corresponding matching from the colors object. As you may guess this will be a big change, and we need to keep in mind to create colors and scheme for the other themes as well (dark and HOC), so that theme switching would work as expected.

packages/react/src/themes/teams/colors.ts
```diff
@@ -163,13 +163,28 @@ export const colors: ColorPalette = {
...
   // Primitive colors
-  black: naturalColors.grey[900],
+  black: '#000',
   white: naturalColors.grey[50],
 }
 
-export const colorScheme: ColorSchemeMapping = _.mapValues(
-  emphasisAndNaturalColors,
-  (colorVariants, colorName) => {
+const primitiveColorsScheme = {
```
layershifter:
```suggestion
const primitiveColorsScheme: Record<keyof PrimitiveColors, ColorScheme> = {
```

layershifter (approved):

PR #856, opened 2019-02-06
docs(screener-io): added documentation on how to write tests with steps

silviuaavram:
Added documentation on how to write tests with the Screener Steps.

At the moment I added a draft change, and I think I can have some feedback while I also look to improve it.

### DIFF
--- a/.github/test-a-feature.md
+++ b/.github/test-a-feature.md
@@ -113,14 +113,14 @@ yarn test:watch
 
 ## Screener Tests
 
-For some components, it is necessary to write screenshot tests in order to check they render properly. For each component example added to the docsite a screenshot test is automatically created. This checks if that the component is rendered in a consistent way, as it checks the visual differences between the previous and the current rendering. We use [screener-io](https://screener.io/) to achieve this.
+For some components, it is necessary to write screenshot tests in order to check that they render properly. For each component example added to the docsite a screenshot test is automatically created. This checks if that the component is rendered in a consistent way, as it looks for visual differences between the previous rendering and the current one. We use [screener-io](https://screener.io/) to achieve screenshot testing.
 
 ### Tests with Steps API
 
-This default test only checks the rendering for the component in its initial state. In order to test the rendering of more complex components, such as a `Dropdown`, screener provides an [api](https://www.npmjs.com/package/screener-runner) to execute actions on the DOM, in a way similar to end-to-end tests. These tests are automatically sent to Screener in our test runs, as long as the tests and their files respect the conventions:
-- the test file should be placed at the same location as the tested component in the docs.
-- the test file should be named exactly as the component file. If `DropdownExample.shorthand.tsx` is to be tested, the screener test file should be `DropdownExample.shorthand.steps.ts`.
-- the tests should be written as an array of callbacks that accept a `steps` parameter, as they will be chained in `screener.config` automatically. The `steps` parameter is actually the `Steps` object from screener, instantiated in `screener.config`.
+This default test only checks the rendering for the component in its initial state. In order to test the rendering of more complex components, such as a `Dropdown`, screener provides an [api](https://www.npmjs.com/package/screener-runner) to execute actions on the DOM, in a way similar to end-to-end tests. These tests are executed by Screener as long as both the tests and their files respect the conventions:
+- the test file should be placed at the same location as the component example under test.
+- the test file should be named exactly as the component example file. If `DropdownExample.shorthand.tsx` is to be tested, the screener test file should be named `DropdownExample.shorthand.steps.ts`.
+- the tests should be written as an array of callbacks that accept a `steps` parameter, as all of them will be chained in `screener.config`. The `steps` parameter is actually the `Steps` object from screener, instantiated in `screener.config`.
 
 #### Example for a test file:
 
@@ -129,9 +129,9 @@ import { Dropdown } from '@stardust-ui/react'
 
 const steps = [
   steps => steps.click(`.${Dropdown.slotClassNames.triggerButton}`)
-    .snapshot('Opens with selected item highlighted'),
+    .snapshot('Opens dropdown list'),
   steps => steps.hover(`.${Dropdown.slotClassNames.itemsList} li:nth-child(2)`)
-    .snapshot('Highlights another item'),
+    .snapshot('Highlights an item'),
 ]
 
 export default steps
--- a/.github/test-a-feature.md
+++ b/.github/test-a-feature.md
@@ -104,3 +104,48 @@ yarn test:watch
 ```
 
 [1]: https://github.com/stardust-ui/react/tree/master/test/specs/commonTests
+
+## Screener Tests
+
+For some components, it is necessary to write screenshot tests in order to check they render properly. For each component added to the docsite, a screenshot test is automatically created. This checks if that the component is rendered in a consistent way, as it checks the visual differences between the previous and the current rendering. We use [screener-io](https://screener.io/) to achieve this.
+
+### Tests with Steps API
+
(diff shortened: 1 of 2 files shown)


.github/test-a-feature.md
```diff
@@ -103,4 +103,49 @@ yarn test
...
+[1]: https://github.com/stardust-ui/react/tree/master/test/specs/commonTests
+
+## Screener Tests
+
+For some components, it is necessary to write screenshot tests in order to check they render properly. For each component added to the docsite, a screenshot test is automatically created. This checks if that the component is rendered in a consistent way, as it checks the visual differences between the previous and the current rendering. We use [screener-io](https://screener.io/) to achieve this.
+
+### Tests with Steps API
+
+This default test only checks the rendering for the component in its initial state. In order to test the rendering of more complex components, such as a `Dropdown`, screener provides an [api](https://www.npmjs.com/package/screener-runner) to execute actions on the DOM, in a way similar to end-to-end tests. These tests are automatically sent to Screener in our test runs, as long as the tests and their files respect the conventions:
+- the test file should be placed at the same location as the tested component in the docs.
```
mnajdova: In my opinion it feels kind a awkward that we are adding tests (steps file) in the docs rather then in the test folder, where all other tests resides. I understand that the initial components are there, but in my opinion maybe it will be worth it to just initialize the component in the same place where the test will reside. Any strong reason for not doing this?

mnajdova: Btw, the comment has nothing to do with the docs, just expressing my thoughts on the procedure

layershifter: We are testing existing examples, it's very useful that we have visual tests in the same place 👀

.github/test-a-feature.md
```diff
@@ -103,4 +103,49 @@ yarn test
...
+- a test can have as many snapshots as needed.
+- before the snapshot is taken, steps are added to reach the state of assertion, with methods from `screener-runner` API (`.click(<selector>)`, `.hover(<selector>)`etc.)
+- tests do not perform any cleanup by default. This means a test is dependent on the state of the component in which the previous test leaves it. In the example above, the second test will receive the `Dropdown` component with the list open.
+
+### Run Screener tests
+In order to run the tests locally, make sure to have a Screener API key saved in the environment variables on your machine. For instance, on MacOS/Linux you can use `export SCREENER_API_KEY=<Your screener key here>`.
+
+When ran locally, not all the tests may be needed to run, only the ones added/edited. This can be achieved by changing the regexp in `screener.config.js`, where the `states` property is created. Make sure not to commit that change though! Results can be viewed in the Screener online dashboard, among all the other runs (for master, pull requests, etc.)
```
mnajdova, line 151: For future work, we may allow user to send the regexp when running the command :)

.github/test-a-feature.md
mnajdova: I would not agree with this one, especially that it is in the docs. If we decide that we don't want to show some example, someone can accidentally remove it, not knowing it was used in testing (yeah, we have the same steps.ts file there as an indicator, and we can remove it just from the ExampleSection) but still, coupling the docs site with tests seems like a red light to me 👀 (watching you too) :D

.github/test-a-feature.md
```diff
@@ -103,4 +103,49 @@ yarn test
 yarn test:watch
 ```
 
-[1]: https://github.com/stardust-ui/react/tree/master/test/specs/commonTests
\ No newline at end of file
+[1]: https://github.com/stardust-ui/react/tree/master/test/specs/commonTests
+
+## Screener Tests
```
sophieH29, line 114: might be worth adding a link to Screener tests to the Content section, line 14

silviuaavram: ran `gulp build:docs:toc` and it did it for me :D

.github/test-a-feature.md
silviuaavram: added https://github.com/stardust-ui/react/issues/860

.github/test-a-feature.md
bmdalex: > If we decide that we don't want to show some example, someone can accidentally remove it, not knowing it was used in testing

How is that different from current behavior? We already have visual tests and removing an example causes screener tests to fail. It's been like this from the beginning.

> coupling the docs site with tests seems like a red light to me
It may be but once again, aren't we doing this already? :) It's just that now we're adding extra steps to already existing tests. At least by having the files here the developer will be aware that THERE ARE some visual tests running.

Also, there is not as if there is something that is going to fail, we'll just have screen regression (nothing new)

@mnajdova what is the alternative in the current state of the project?
The long term solution is E2E tests but I don't know where that is on the list in terms of priority

bmdalex (approved):
GG 👍 thanks @silviuavram

.github/test-a-feature.md
mnajdova: Yep, we are doing it already, but that doesn't mean it's the right thing to do. At least these files, are actually used in the docs, as examples for the users. I guess the initial thought was, let's reuse these files, so till some point I would understand that (we may change this in the future). But now, we are actually adding test steps in the same directory (that are not in any ways used in the docs). The alternative is, we can define these files in the tests directory, and reuse the files from the docs. Then if we decide that for the purpose of the tests we need to change the initial screen, we can create a separate file, that will not necessarily need to be part of the docs. Anyways, I started the discussion with asking whether there is strong reason for not splitting the tests in a separate place (and apparently from the comments, there isn't). 
With restructuring the packages (the docs, react implementation, accessibility etc), we will hit this, so may tackle this at that point.
Anyways, we can iterate on this later, and not block the docs which is the main focus for this PR :)

mnajdova:
We have 99 screen regressions?!?! All new screens are empty pages. I guess is another related to your changes in this PR, but not sure at this moment.. Can you re-trigger the build please? @silviuavram

mnajdova (approved):
LGTM 👍

.github/test-a-feature.md
```diff
@@ -103,4 +109,49 @@ yarn test
...
+
+For some components, it is necessary to write screenshot tests in order to check they render properly. For each component added to the docsite, a screenshot test is automatically created. This checks if that the component is rendered in a consistent way, as it checks the visual differences between the previous and the current rendering. We use [screener-io](https://screener.io/) to achieve this.
```
kuzhelov: nit: For each component => For each component _example_, no comma is necessary in this sentence

silviuaavram: thank you!

.github/test-a-feature.md
```diff
@@ -103,4 +109,49 @@ yarn test
...
+#### Important mentions:
+
+- by convention, each test is written as a different callback and added to the steps array.
+- an actual assertion is performed by taking a `.snapshot(<Your test name here>)`, as the assertion is performed by comparing the screenshot with the one taken on the previous run and considered correct.
+- a test can have as many snapshots as needed.
+- before the snapshot is taken, steps are added to reach the state of assertion, with methods from `screener-runner` API (`.click(<selector>)`, `.hover(<selector>)`etc.)
+- tests do not perform any cleanup by default. This means a test is dependent on the state of the component in which the previous test leaves it. In the example above, the second test will receive the `Dropdown` component with the list open.
+
+### Run Screener tests
+In order to run the tests locally, make sure to have a Screener API key saved in the environment variables on your machine. For instance, on MacOS/Linux you can use `export SCREENER_API_KEY=<Your screener key here>`.
```
layershifter, line 149: And what is for Windows? 🔍 

This should work:
```bash
cross-env SCREENER_API_KEY=<Your screener key here> yarn test:visual
```

silviuaavram: oups. will address this in my arrow navigation PR :)

PR #857, opened 2019-02-07
 chore(docs): use minified versions of Babel

layershifter:
Also fixes CDN with broken `babel.js`.

----

Errors work fine even on minified builds:

babel


docs/src/index.ejs
```diff
@@ -49,8 +49,8 @@
   <script src="https://cdnjs.cloudflare.com/ajax/libs/react-dom/<%= htmlWebpackPlugin.options.versions.reactDOM %>/umd/react-dom<%= __PROD__ ? '.production.min' : '.development' %>.js"></script>
   <script src="https://cdnjs.cloudflare.com/ajax/libs/react-dom/<%= htmlWebpackPlugin.options.versions.reactDOM %>/umd/react-dom-server.browser<%= __PROD__ ? '.production.min' : '.development' %>.js"></script>
 
-  <script src="https://cdn.jsdelivr.net/npm/@babel/standalone@7/babel.js"></script>
-  <script src="https://cdn.jsdelivr.net/npm/@babel/preset-env-standalone@7/babel-preset-env.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/@babel/standalone@7.3/babel.min.js"></script>
```
kuzhelov: shouldn't we use the same version restriction as we have in our `devDependencies` - specifically, `^7.1`? Or this one is not working for the env where you are able to repro the problem (asking because for me it has worked, but I wasn't able to repro even the original problem)?

layershifter: The best way is to use `versions` from `package.json`, but it means that will also need to add `@babel/preset-env-standalone` to deps.

---

In fact, we not using `@babel/standalone` anywhere, so may be we can remove it from dependencies at all 🤔  I'm also not that we need `preset-env` there, `preset-typescript` does much more work.

My proposal:
- remove the usage `@babel/preset-env`
- use `@babel/standalone` version from dependencies

kuzhelov: awesome!

kuzhelov (approved):

docs/src/components/Playground/renderConfig.ts
```diff
@@ -4,16 +4,10 @@ import * as Stardust from '@stardust-ui/react'
...
-      {
-        targets: {
-          browsers: ['last 4 versions', 'not dead'],
-        },
-      },
-    ],
+  plugins: [
+    'proposal-class-properties',
+    ['transform-typescript', { isTSX: true }],
+    'transform-classes',
```
layershifter, line 10: If we will have additional requirements we can simply turn on the `stage-*` preset, they are bundled with `babel-standalone`

PR #858, opened 2019-02-07
feat(teams-theme): clearing duplicate colors from siteVariables

mnajdova:
# Description of the problem
In Teams theme we are exporting colors in different ways
- through the `colors` object (containing the natural, emphasis, primitive colors)
- directly through `siteVariables` (we have fields there like orange04, grey02 etc)

This produces big confusion when using the colors. As an example, we have the black color both as `siteVariables.black`, as well as `siteVariables.colors.black`. These values can be the same or different (the user doesn't know right away, until they check the values directly in the theme). Moreover, when we look into it, the black color is actually in exported as the `siteVariables.colors.grey[900]` variant. Why we have black color, if it is the same as some grey variant (**it's not even black**). 

In addition if we look into the values of the colors `siteVariables.brand` and `siteVariables.brand06`, they are referencing the same color - `siteVariables.colors.primary[500]` (why we need two different aliases for the same color?). We also have some colors defined in the `siteVariables` which are not used at all in the styles and variables (not sure whether they are used on client side) 

The aim of this PR is to start reducing the colors defined directly in the `siteVariables`, as long as we have alias color in the `colors` object. From here then, we can have clear picture of which colors defined in the `siteVariables`, do not have mapping in the color palette, and we can iterate on that. Then, we can start creating color palettes for the other themes (dark and high contracts). 

# Goal
We should aim for clearing all colors directly defined in the siteVariables, and referencing them with something from the colors object.

# BREAKING CHANGES (Teams theme)
Some of the colors from `siteVariables` were removed. Find the mappings below:
```
brand16 <=> colors.primary[50]
brand14 <=> colors.primary[100]
brand12 <=> colors.primary[200]
brand <=> colors.primary[500]
brand06 <=> colors.primary[500]
brand02 <=> colors.primary[900]

green <=> naturalColors.lightGreen[900]
green04 <=> colors.green[900]

black <=> colors.grey[900]
white <=> colors.grey[50]

red <=> colors.red[900]

yellow <=> colors.yellow[900]

orange04 <=> naturalColors.darkOrange[400]
```

In addition, the value of the `siteVariable.colors.black` was changed from `siteVariables.colors.grey[900] (#252424)` to `#000`. If the intention of the user was to have the dark grey, then please replace the usage with `siteVariables.colors.grey[900]`.


packages/react/src/themes/teams/colors.ts
```diff
@@ -10,16 +10,16 @@ import {
...
-    50: '#F4F4FC',
-    100: '#E2E2F6',
-    200: '#BDBDE6',
+    50: '#F4F4FC', // siteVariables.brand16
+    100: '#E2E2F6', // siteVariables.brand14
+    200: '#BDBDE6', // siteVariables.brand12
     300: '#8F90C1',
     400: '#6E70AE',
-    500: '#6264A7',
+    500: '#6264A7', // siteVariables.brand, siteVariables.brand06 (same color?)
```
bmdalex, line 20: yes, it's the same color

packages/react/src/themes/teams-dark/siteVariables.ts
```diff
@@ -19,7 +19,7 @@ export const brand02 = '#e2e2f6'
 export const brand04 = '#bdbde6'
 export const brand06 = '#a6a7dc'
 export const brand08 = '#8b8cc7'
-export const brand12 = brand
+export const brand12 = colors.primary[500]
```
bmdalex: is there a plan to remove the `brand*` names from other themes?
that would help A LOT with the `color` prop story since we won't need any code changes in the styles of the components for other themes (you probably remember the `Menu` mayhem)

bmdalex (approved):
YAAASSS!!

packages/react/src/themes/teams/colors.ts
mnajdova: Thanks for pointing out :D

packages/react/src/themes/teams-dark/siteVariables.ts
mnajdova: Yep, I remember the problems we had there. In my opinion, ideally we will have colors.primary[500] as a color used for the same thing in all themes, but the themes would change what primary[500] means in that context (it might be lighter in dark theme for example...) Not certain 100% for this yet, we have to have the whole picture first..

mnajdova: This one is actually great example that, in dark theme, `colors.primary[500]` is used for `brand12`, although in the light theme it is used as `brand`. We should rearrange the colors primary in dark theme to match these values accordingly by specifying the correct hex values (I guess the values will be lighter then they are in the light theme)

packages/react/src/themes/teams-dark/siteVariables.ts
```diff
@@ -1,34 +1,34 @@
...
-export const gray10 = '#323130'
-export const gray14 = '#292828'
+export const gray02 = '#c8c6c4' // light theme gray06
+export const gray03 = '#b3b0ad' // no mapping color
+export const gray04 = '#8a8886' // no mapping color
+export const gray06 = '#605e5c' // light theme gray03
+export const gray08 = '#484644' // light theme gray02
+export const gray09 = '#3b3a39' // no mapping color
+export const gray10 = '#323130' // no mapping color
+export const gray14 = '#292828' // no mapping color
```
layershifter, line 15: It feels like `gray*` and `brand*` there should extracted as part of color pallete

packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
```diff
@@ -26,18 +26,18 @@ export default (siteVars): ChatMessageVariables => ({
...
   headerMarginBottom: pxToRem(2),
-  contentFocusOutlineColor: siteVars.brand,
+  contentFocusOutlineColor: siteVars.colors.primary[500],
   border: 'none',
   badgeShadow: siteVars.shadowLevel1Darker,
   isImportant: false,
   hasMention: false,
-  hasMentionColor: siteVars.orange04,
-  isImportantColor: siteVars.red,
+  hasMentionColor: siteVars.naturalColors.darkOrange[400],
```
layershifter, line 42:
```suggestion
  hasMentionColor: siteVars.colors.darkOrange[400],
```

packages/react/src/themes/teams/components/Text/textVariables.ts
```diff
@@ -7,14 +7,14 @@ export interface TeamsTextVariables extends TextVariables {
 }
 
 export default (siteVariables): Partial<TeamsTextVariables> => ({
-  atMentionOtherColor: siteVariables.brand06,
-  atMentionMeColor: siteVariables.orange04,
+  atMentionOtherColor: siteVariables.colors.primary[500],
+  atMentionMeColor: siteVariables.naturalColors.darkOrange[400],
```
layershifter, line 11:
```suggestion
  atMentionMeColor: siteVariables.colors.darkOrange[400],
```

packages/react/src/themes/teams/components/Attachment/attachmentVariables.ts
```diff
@@ -26,9 +26,9 @@ export default siteVariables => ({
 
   backgroundColor: siteVariables.gray09,
   backgroundColorHover: siteVariables.gray08,
-  textColor: siteVariables.black,
+  textColor: siteVariables.colors.grey[900],
 
-  progressColor: siteVariables.green,
+  progressColor: siteVariables.naturalColors.lightGreen[900],
```
layershifter, line 31:
```suggestion
  progressColor: siteVariables.colors.lightGreen[900],
```

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -86,7 +86,9 @@ const chatMessageStyles: ComponentSlotStylesInput<
     props: p,
     variables: v,
     theme: {
-      siteVariables: { white },
+      siteVariables: {
+        colors: { white },
+      },
```
layershifter: Actually, should be moved to `chatMessageVariables`

packages/react/src/themes/teams/siteVariables.ts
```diff
@@ -11,35 +11,21 @@ export const htmlFontSize = '10px' // what 1rem represents
...
-export const gray10 = '#F3F2F1'
-export const gray14 = '#FAF9F8'
+export const gray02 = '#484644' // no mapping color
+export const gray03 = '#605E5C' // no mapping color
+export const gray04 = '#979593' // no mapping color
+export const gray06 = '#C8C6C4' // no mapping color
+export const gray08 = '#E1DFDD' // no mapping color
+export const gray09 = '#EDEBE9' // no mapping color
+export const gray10 = '#F3F2F1' // no mapping color
+export const gray14 = '#FAF9F8' // no mapping color
```
layershifter, line 21: Have a mixed feeling about it. If we are trying to rid of colors on `siteVariables` directly, should we keep these?
May be add an evil prefix to them? `__unsafe__gray14`? 😈

mnajdova: This step is just a clean up, so that we can have clear picture of what colors are missing in the color palette. Then the design team can safely update the color palette, following the steps introduced in this PR. I would not want to introduce breaking changes now for these colors, as most likely they are going to be removed soon.

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
mnajdova: What do you mean? We have the white in the `colors` and not directly in the `siteVariables`

layershifter: Sorry, updated comment, it's more clear

packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
mnajdova: Actually I think this was producing TS errors, that's why I have the `naturalColors`. Let me check and will confirm

mnajdova: The problem is that the `NaturalColorsStrict` doesn't contain this color.

mnajdova: Hm, it seems that it's working.. Never mind will change it and report back if I have an issue with that.

packages/react/src/themes/teams-dark/siteVariables.ts
mnajdova: We have grey in the color palette, but the color are not matching these one, after the updates of the color palette, we may be able to reduce them. For the brand colors, currently I am mapping them to the primary, but we may change this to brand in the future.

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
mnajdova: Got it, thanks!

layershifter (approved):

packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
mnajdova: Created issue with proposed fixed: https://github.com/stardust-ui/react/issues/898

PR #859, opened 2019-02-07
fix(themes): do not define fontFaces in Teams theme by default

### LINKED ISSUE #692: Remove references to Segoe Font in Stardust
### Problem description
Currently have hard coded references to westeuropean version of segoe fonts from sharepoint servers.

If users want other regional characters, then these aren't available.

### Proposed solution
Base themes, and stardust itself to use standard browser fonts.

### MVP

layershifter:
Fixes #692.

# BREAKING CHANGES

Teams theme will not import Segoe UI font by default now, please import it manually if you need it.

```jsx
import React from 'react'
import ReactDOM from 'react-dom'
import { Provider, themes } from '@stardust-ui/react'

import App from './App'

const themeWithFonts = {
  ...themes.teams,
  // <=== Fonts definition is here
  fontFaces: [
    {
      name: 'Segoe UI',
      paths: [
        'https://static2.sharepointonline.com/files/fabric/assets/fonts/segoeui-westeuropean/segoeui-regular.woff2',
      ],
      style: { fontWeight: 400 },
    },
    {
      name: 'Segoe UI',
      paths: [
        'https://static2.sharepointonline.com/files/fabric/assets/fonts/segoeui-westeuropean/segoeui-semibold.woff2',
      ],
      style: { fontWeight: 600 },
    },
    {
      name: 'Segoe UI',
      paths: [
        'https://static2.sharepointonline.com/files/fabric/assets/fonts/segoeui-westeuropean/segoeui-bold.woff2',
      ],
      style: { fontWeight: 700 },
    },
  ],
}

ReactDOM.render(
  <Provider theme={themeWithFonts}>
    <App />
  </Provider>,
  document.getElementById('root'),
)

```

### DIFF
--- a/docs/src/index.ejs
+++ b/docs/src/index.ejs
@@ -122,6 +122,24 @@
     .hidden {
       display: none;
     }
+
+    /* Teams fonts */
+
+    @font-face {
+      font-family: 'Segoe UI';
+      font-weight: 400;
+      src: url('https://static2.sharepointonline.com/files/fabric/assets/fonts/segoeui-westeuropean/segoeui-regular.woff2') format('woff2')
+    }
+    @font-face {
+      font-family: 'Segoe UI';
+      font-weight: 600;
+      src: url('https://static2.sharepointonline.com/files/fabric/assets/fonts/segoeui-westeuropean/segoeui-semibold.woff2') format('woff2')
+    }
+    @font-face {
+      font-family: 'Segoe UI';
+      font-weight: 700;
+      src: url('https://static2.sharepointonline.com/files/fabric/assets/fonts/segoeui-westeuropean/segoeui-bold.woff2') format('woff2')
+    }
   </style>
 </body>
 </html>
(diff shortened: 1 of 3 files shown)


docs/src/index.ejs
```diff
@@ -121,6 +121,24 @@
     .hidden {
       display: none;
     }
+
+    /* Teams fonts */
+
+    @font-face {
```
kuzhelov, line 127: ideally would like this knowledge reside in Teams theme, so that this will be a matter of importing these fonts here from some theme's module. This will aid general scenarios where we would need to support other themes - as the approach to consume font data will be unified (to some extent, at least). What do you think?

layershifter: My proposal: leave it as is for now and avoid introducing something until we will need it. For now we have actually only one theme and I don't want to introduce something like `css-loader` to load one style.

If we have a ready simple proposal, share it 🐱

kuzhelov (approved):

layershifter:
After the offline discussion we decided to keep at it as is. `siteVariables.fontFaces` can be easily overrided as they are part of the public API:

```js
import { Provider, themes } from '@stardust-ui/react'

const themeWithFonts = {
  ...themes.teams,
  fontFaces: [],
]
```

PR #861, opened 2019-02-07
fix(Popup): Focus the last focused element outside Popup on ESC

### LINKED ISSUE #831: Popup triggered by MenuItem is focusing the wrapper instead of the root when user presses ESC
## Bug Report
Popup has the ability to focus trigger element when it is closed by ESC key.
This does not work properly with MenuItem - in that case, the element to be triggered needs to be the root (a), not the wrapper (li) element.

### Testcase
https://codesandbox.io/s/mql1160ll8?module=/example.js

sophieH29:
This PR fixes the issue when a complex component is used as a trigger for Popup, but actually, the inner element of that component triggers the Popup. So when press ESC on Popup, the focus should go to the actual focusable element which triggered the Popup, not the whole component itself.

In the next example, `MenuItem` is a trigger for the Popup and this component has this structure
```jsx
<li> - wrapper
   <a></a> // focusable element which actually triggers Popup
</li>

```

```jsx
const items = [
  {
    key: 'editorials',
    content: 'Editorials',
  },
  {
    key: 'review',
    content: 'Reviews',
  },
  {
    key: 'events',
    content: 'Upcoming Events',
  },
].map(item => render => render(item, renderIten))

const renderIten = (MenuItem, props) =>
  <Popup
    content={{ content: <Button content="Click me" /> }}
    trigger={<MenuItem {...props}/>}
  />

const MenuExample = () => <Menu defaultActiveIndex={0} items={items} />
```

So when press ESC, `MenuItem`'s wrapper `<li>` element will receive focus, instead of `<a>`, what breaks focus management on the page.

Closes https://github.com/stardust-ui/react/issues/831

### DIFF
--- a/packages/react/src/components/Popup/Popup.tsx
+++ b/packages/react/src/components/Popup/Popup.tsx
@@ -166,13 +166,15 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
   private outsideClickSubscription = EventStack.noSubscription
 
   private triggerDomElement = null
+  // focusable element which has triggered Popup, can be either triggerDomElement or the element inside it
+  private triggerFocusableDomElement = null
   private popupDomElement = null
 
   private closeTimeoutId
 
   protected actionHandlers: AccessibilityActionHandlers = {
     closeAndFocusTrigger: e => {
-      this.close(e, () => _.invoke(this.triggerDomElement, 'focus'))
+      this.close(e, () => _.invoke(this.triggerFocusableDomElement, 'focus'))
       e.stopPropagation()
     },
     close: e => this.close(e),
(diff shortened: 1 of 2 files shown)


jurokapsiar (approved):

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -474,6 +476,8 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
   }
 
   private trySetOpen(newValue: boolean, eventArgs: any) {
+    // when new state 'open' === 'true', save the last focused element
+    newValue && this.updateTriggerFocusableDomElement()
```
```suggestion
    if (newValue) this.updateTriggerFocusableDomElement()
```

```suggestion
    if (newValue) {
      this.updateTriggerFocusableDomElement()
    }
```

Let's make it more readable.

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -497,4 +501,14 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
       onClose && onClose()
     }
   }
+
+  /**
+   * Save DOM element which had focus before Popup opens.
+   * Can be either trigger DOM element itself or the element inside it.
+   */
+  private updateTriggerFocusableDomElement() {
```
layershifter, line 511: Do we need to introduce a new method for a single use?

sophieH29: I think it makes sense, don't want to mess up `tryOpen` method and still remain better reading

layershifter (approved):

PR #863, opened 2019-02-07
feat(dropdown): inline prop

bmdalex:
### Description:

This PR implements the `inline` prop for `Dropdown` component.

**There will be a follow-up PR with a prototype on how to handle the inline Dropdown in the context of a rich text editor** 

### API

```jsx
<Dropdown inline items={inputItems} placeholder="Select your hero" />
```

### Styling
screen recording 2019-02-07 at 18 36 44

### DIFF
--- a/docs/src/examples/components/Dropdown/Types/DropdownExampleInline.shorthand.tsx
+++ b/docs/src/examples/components/Dropdown/Types/DropdownExampleInline.shorthand.tsx
@@ -1,0 +1,38 @@
+import * as React from 'react'
+import { Dropdown, Header } from '@stardust-ui/react'
+
+const inputItems = [
+  'Bruce Wayne',
+  'Natasha Romanoff',
+  'Steven Strange',
+  'Alfred Pennyworth',
+  `Scarlett O'Hara`,
+  'Imperator Furiosa',
+  'Bruce Banner',
+  'Peter Parker',
+  'Selina Kyle',
+]
+
+const DropdownExample = () => (
+  <>
+    <Header as="h3">Inline:</Header>
+    <div>
+      Some text inline with the{' '}
+      <Dropdown inline items={inputItems} placeholder="Select your hero" /> and more text.
+    </div>
+    <Header as="h3">Inline Search:</Header>
+    <span>
+      Some other text inline with the{' '}
+      <Dropdown
+        inline
+        search
+        items={inputItems}
+        noResultsMessage="We couldn't find any matches."
+        placeholder="Start typing a name"
+      />{' '}
+      and more text.
+    </span>
+  </>
+)
+
+export default DropdownExample
--- a/docs/src/examples/components/Dropdown/Types/DropdownExampleInline.shorthand.tsx
+++ b/docs/src/examples/components/Dropdown/Types/DropdownExampleInline.shorthand.tsx
@@ -13,7 +13,7 @@ const inputItems = [
   'Selina Kyle',
 ]
 
-const DropdownExample = () => (
+const DropdownExampleInline = () => (
   <>
     <Header as="h3">Inline:</Header>
     <div>
@@ -35,4 +35,4 @@ const DropdownExample = () => (
   </>
 )
 
-export default DropdownExample
+export default DropdownExampleInline
--- a/packages/react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react/src/components/Dropdown/Dropdown.tsx
@@ -76,6 +76,9 @@ export interface DropdownProps extends UIComponentProps<DropdownProps, DropdownS
    */
   getA11yStatusMessage?: (options: DownshiftA11yStatusMessageOptions<ShorthandValue>) => string
 
+  /** A dropdown can be formatted to appear inline in the content of other components. */
+  inline?: boolean
+
   /** Array of props for generating list options (Dropdown.Item[]) and selected item labels(Dropdown.SelectedItem[]), if it's a multiple selection. */
   items?: ShorthandValue[]
 
@@ -190,6 +193,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     fluid: PropTypes.bool,
     getA11ySelectionMessage: PropTypes.object,
     getA11yStatusMessage: PropTypes.func,
+    inline: PropTypes.bool,
     items: customPropTypes.collectionShorthand,
     itemToString: PropTypes.func,
     loading: PropTypes.bool,
@@ -333,15 +337,16 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     styles: ComponentSlotStylesInput,
     getToggleButtonProps: (options?: GetToggleButtonPropsOptions) => any,
   ): JSX.Element {
+    const { inline, fluid, triggerButton } = this.props
     const content = this.getSelectedItemAsString(this.state.value)
 
     return (
       <Ref innerRef={this.buttonRef}>
-        {Button.create(this.props.triggerButton, {
+        {Button.create(triggerButton, {
           defaultProps: {
             className: Dropdown.slotClassNames.triggerButton,
             content,
-            fluid: true,
+            fluid: fluid || !inline,
             styles: styles.triggerButton,
             ...getToggleButtonProps({
               onFocus: () => {
@@ -369,7 +374,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     ) => void,
     variables,
   ): JSX.Element {
-    const { searchInput, multiple, placeholder } = this.props
+    const { inline, searchInput, multiple, placeholder } = this.props
     const { searchQuery, value } = this.state
 
     const noPlaceholder =
@@ -378,6 +383,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     return DropdownSearchInput.create(searchInput || {}, {
       defaultProps: {
         placeholder: noPlaceholder ? '' : placeholder,
+        inline,
         variables,
         inputRef: this.inputRef,
       },
--- a/packages/react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react/src/components/Dropdown/Dropdown.tsx
@@ -337,7 +337,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     styles: ComponentSlotStylesInput,
     getToggleButtonProps: (options?: GetToggleButtonPropsOptions) => any,
   ): JSX.Element {
-    const { inline, fluid, triggerButton } = this.props
+    const { triggerButton } = this.props
     const content = this.getSelectedItemAsString(this.state.value)
 
     return (
@@ -346,7 +346,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
           defaultProps: {
             className: Dropdown.slotClassNames.triggerButton,
             content,
-            fluid: fluid || !inline,
+            fluid: true,
             styles: styles.triggerButton,
             ...getToggleButtonProps({
               onFocus: () => {
--- a/packages/react/src/components/Dropdown/DropdownSearchInput.tsx
+++ b/packages/react/src/components/Dropdown/DropdownSearchInput.tsx
@@ -14,6 +14,9 @@ import { UIComponentProps } from '../../lib/commonPropInterfaces'
 import Input from '../Input/Input'
 
 export interface DropdownSearchInputProps extends UIComponentProps<DropdownSearchInputProps> {
+  /** A dropdown search input can be formatted to appear inline in the context of a Dropdown. */
+  inline?: boolean
+
   /** Ref for input DOM node. */
   inputRef?: React.Ref<HTMLElement>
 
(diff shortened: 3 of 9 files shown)


packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -333,15 +337,16 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
 
     return (
       <Ref innerRef={this.buttonRef}>
-        {Button.create(this.props.triggerButton, {
+        {Button.create(triggerButton, {
           defaultProps: {
             className: Dropdown.slotClassNames.triggerButton,
             content,
-            fluid: true,
+            fluid: fluid || !inline,
```
bmdalex: I'm not that sure about this condition; @silviuavram this is about the `Button` component we use for selection `Dropdown`, which should be `fluid` in all cases except `inline`.

However `<Dropdown fluid ../>` should override the `inline` condition anyway, hence the `fluid: fluid || !inline` condition.

I'm thinking to leave just `fluid` at this line and move the logic for `!inline` to the styles, what do you think?

docs/src/examples/components/Dropdown/Types/DropdownExampleInline.shorthand.tsx
```diff
@@ -0,0 +1,38 @@
...
+        items={inputItems}
+        noResultsMessage="We couldn't find any matches."
+        placeholder="Start typing a name"
+      />{' '}
+      and more text.
+    </span>
+  </>
+)
+
+export default DropdownExample
```
layershifter:
```suggestion
export default DropdownExampleInline
```

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -65,7 +89,8 @@ const dropdownStyles: ComponentSlotStylesInput<DropdownProps & DropdownState, Dr
     zIndex: 1000,
     maxHeight: v.listMaxHeight,
     overflowY: 'auto',
-    width: p.fluid ? '100%' : v.width,
+    width: getWidth(p, v),
+    minWidth: 'max-content',
```
layershifter: image

https://developer.mozilla.org/en-US/docs/Web/CSS/min-width
This thing is not supported in Edge & IE at all 💣

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -3,23 +3,49 @@ import { DropdownProps, DropdownState } from '../../../../components/Dropdown/Dr
...
+  ':hover': transparentColorStyle,
+  ':active': transparentColorStyle,
+  ':focus': {
+    ...transparentColorStyle,
+    ':active': transparentColorStyle,
+  },
+}
+
+const getWidth = (p: Props, v: DropdownVariables): string =>
+  p.fluid ? '100%' : p.inline ? 'initial' : v.width
```
layershifter: https://eslint.org/docs/rules/no-nested-ternary

packages/react/src/components/Dropdown/Dropdown.tsx
layershifter: I think that we need to extract it to a separate variable `buttonFluid` and leave a comment, we should keep styles simple 🐱

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -3,23 +3,49 @@ import { DropdownProps, DropdownState } from '../../../../components/Dropdown/Dr
 import { DropdownVariables } from './dropdownVariables'
 import { pxToRem } from '../../../../lib'
 
-const dropdownStyles: ComponentSlotStylesInput<DropdownProps & DropdownState, DropdownVariables> = {
-  root: (): ICSSInJSStyle => ({}),
+type Props = DropdownProps & DropdownState
```
layershifter: Not sure that `Props` is the best name there, `DropdownPropsAndState` may be?

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
bmdalex: damn it..**I saw it used in SUIR though for the same thing**; here are the screenshots for my scenario:

### 1. with `minWidth: 'max-content'` -> list items fit on one line:
screenshot 2019-02-08 at 12 43 14

### 2. without `minWidth: 'max-content'` -> list items inherit width from narrow parent so they don't fit on one line:
screenshot 2019-02-08 at 12 43 25

Any idea how to achieve this?
@layershifter

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: it was just one line in styles, I reverted the condition here; don't think we can make it better at this point :(

docs/src/examples/components/Dropdown/Types/DropdownExampleInline.shorthand.tsx
bmdalex: done, thx

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
bmdalex: done, but we should have this as part of the build process or in dedicated `eslint` so that we get a warning while using it
@layershifter

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
bmdalex: done, thx

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
layershifter: Actually, I checked SUI code again, they use `white-space: nowrap;` on `item`.

### Before in Edge

image

### After in Edge

image

---

```js
    [`&.${ListItem.className}`]: {
      backgroundColor: v.listItemBackgroundColor,
      whiteSpace: 'nowrap',
    },
```

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
layershifter: #600, will check it, actually should be unblocked now :)

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
bmdalex: works like a charm, thank you sir 💯

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -83,7 +117,7 @@ const dropdownStyles: ComponentSlotStylesInput<DropdownProps & DropdownState, Dr
     fontWeight: 'bold',
   }),
 
-  toggleIndicator: ({ variables: v }): ICSSInJSStyle => ({
+  toggleIndicator: ({ props: p, variables: v }): ICSSInJSStyle => ({
```
mnajdova: Why we need props here?

packages/react/src/components/Dropdown/DropdownSearchInput.tsx
```diff
@@ -70,7 +73,7 @@ class DropdownSearchInput extends UIComponent<ReactProps<DropdownSearchInputProp
     }),
     accessibilityInputProps: PropTypes.object,
     accessibilityComboboxProps: PropTypes.object,
-    hasToggleButton: PropTypes.bool,
```
mnajdova, line 73: In which cases were we using this before?

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
bmdalex: we don't, I forgot it there while testing something, removed, thanks!

packages/react/src/components/Dropdown/DropdownSearchInput.tsx
bmdalex: we were not, forgot it there when we removed the `hasToggleButton` prop

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: maybe I don't fully understand what is going on here. If the Dropdown is inline, ok, fine, the whole container will be inline. Why does the Button, which is fluid inside that container, should be affected? As long as you handle the whole inline case through the first `<div>` (the container one), all should be exactly as it was before.

please let me know what I am missing :)

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -3,23 +3,58 @@ import { DropdownProps, DropdownState } from '../../../../components/Dropdown/Dr
...
 import { pxToRem } from '../../../../lib'
 
-const dropdownStyles: ComponentSlotStylesInput<DropdownProps & DropdownState, DropdownVariables> = {
-  root: (): ICSSInJSStyle => ({}),
+type DropdownPropsAndState = DropdownProps & DropdownState
+
+const transparentColorStyle: ICSSInJSStyle = {
+  backgroundColor: 'transparent',
+  borderColor: 'transparent',
+  borderBottomColor: 'transparent',
```
mnajdova, line 11: I hate that we have to specify the borderBottomColor as well :\

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -3,23 +3,58 @@ import { DropdownProps, DropdownState } from '../../../../components/Dropdown/Dr
...
+  ...transparentColorStyle,
+  ':hover': transparentColorStyle,
+  ':active': transparentColorStyle,
+  ':focus': {
+    ...transparentColorStyle,
+    ':active': transparentColorStyle,
+  },
+}
+
+const getWidth = (p: DropdownPropsAndState, v: DropdownVariables): string => {
```
mnajdova, line 24: 👍

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -32,19 +67,14 @@ const dropdownStyles: ComponentSlotStylesInput<DropdownProps & DropdownState, Dr
...
       margin: '0',
       justifyContent: 'left',
       padding: v.comboboxPaddingButton,
       height: pxToRem(30),
-      ...transparentColorStyle,
       ...(p.multiple && { minWidth: 0, flex: 1 }),
-      ':hover': transparentColorStyle,
+      ...transparentColorStyleObj,
       ':focus': {
         ...transparentColorStyle,
```
mnajdova, line 79: Do we really need to apply here again the `transparentColorStyle`?

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -54,7 +84,11 @@ const dropdownStyles: ComponentSlotStylesInput<DropdownProps & DropdownState, Dr
         },
         ':active': transparentColorStyle,
```
mnajdova, line 85: Here again?

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
bmdalex: me 2, any ideas?
It's just horrendous how many styles we need to override, this goes to technical debt I'm afraid

mnajdova: I saw @kuzhelov closed the corresponded issue for this: https://github.com/stardust-ui/react/issues/237 Are we missing something Roman?

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
bmdalex: it's either that or writing a deep merge function...
for `:focus` pseudoselector there are even more styles that we need to override;

everything would be easier if we had a basic `Button`. Not sure what else I can do here at the moment

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
bmdalex: same explanation as in https://github.com/stardust-ui/react/pull/863/files/35b41aea2d575722a43ff00db73186bc5d2aa442#r255066767

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
bmdalex: oh really, let me try then

bmdalex: I removed `borderBottomColor: 'transparent'` and I'm afraid it doesn't work... I shouldn't see the bottom border; see GIF below:

screen recording 2019-02-08 at 14 12 11

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
mnajdova: I understand the issue, thanks.

mnajdova (approved):

Issue #613, opened 2018-12-14
Unable to use Stardust in TS project with --isolatedModules


miroslavstastny:
### Steps
Create react app:
```
npx create-react-app stardust-consumer --typescript
```
Add stardust package, `Provider`, `Button`, run the app

### Expected Result
Works

### Actual Result
App fails to compile because of the two following errors 💣
1. FocusZone (under our control):
```
/stardust-consumer/node_modules/@stardust-ui/react/dist/es/lib/accessibility/FocusZone/FocusZone.types.d.ts
Type error: Ambient const enums are not allowed when the '--isolatedModules' flag is provided.  TS1209

    114 |     preventDefaultWhenHandled?: boolean;
    115 | }
  > 116 | export declare const enum FocusZoneTabbableElements {
        |                           ^
    117 |     /** Tabbing is not allowed */
    118 |     none = 0,
    119 |     /** All tabbing action is allowed */
```

2. Fela (out of our control, currently in all Fela version, up to 10.0.1):
```
stardust-consumer/node_modules/@stardust-ui/react/node_modules/fela/index.d.ts
Type error: Ambient const enums are not allowed when the '--isolatedModules' flag is provided.  TS1209

    11 |   type TEnhancer = (renderer: IRenderer) => IRenderer; //http://fela.js.org/docs/advanced/Enhancers.html
    12 | 
  > 13 |   const enum TSubscribeMessageType {
       |              ^
    14 |     rule = 'RULE',
    15 |     staticObject = 'RULE',
    16 |     keyframes = 'KEYFRAME',
```

### Version
0.14.0

### Why this is important
- `create-react-app` is the essential way how to create the project
- it is not possible to disable `--isolatedModules` there

layershifter:
We cannot build project with Babel because it also not working with `const enum`: https://babeljs.io/docs/en/babel-plugin-transform-typescript

PR #867, opened 2019-02-08
fix(FocusZone): get rid of `const enum`

layershifter:
Refs #613.


layershifter:
Should I add an entry to changelog? This type wasn't exported publicly.

sophieH29:
> Should I add an entry to changelog? This type wasn't exported publicly.

There is a `FocusZone`'s changelog file, please fill the change there.

sophieH29 (approved):

miroslavstastny (approved):

PR #868, opened 2019-02-08
chore(package): add polyfills to correctly work in IE11

### LINKED ISSUE #769: MergeIcons throws in IE11
## Bug Report
`mergeIcons` method from `react/src/lib/mergeThemes.ts` throws an exception on Internet Explorer 11:
`Unable to get property 'apply' of undefined or null reference`

The reason for this is that Internet Explorer does not support `Object.assign`.
### Version
0.17

layershifter:
Fixes #769.


docs/src/components/Playground/renderConfig.ts
```diff
@@ -9,6 +9,7 @@ export const babelConfig = {
     ['transform-typescript', { isTSX: true }],
     'transform-classes',
   ],
+  presets: ['es2015'],
```
layershifter, line 12: Need this to run examples in IE11

docs/src/index.ejs
```diff
@@ -26,6 +26,7 @@
   </script>
 </head>
 <body>
+  <script src="https://cdn.jsdelivr.net/npm/@babel/polyfill@7/dist/polyfill.min.js"></script>
```
layershifter, line 29: Only for docs, see https://github.com/Semantic-Org/Semantic-UI-React/issues/3369

package.json
```diff
@@ -145,7 +145,7 @@
     "react-hot-loader": "^4.1.3",
     "react-router": "^4.1.2",
     "react-router-dom": "^4.1.2",
-    "react-source-render": "^2.0.0-beta.4",
+    "react-source-render": "2.0.0-beta.6",
```
layershifter, line 150: IE11 doesn't support the object destruction, see changes in fixtures: https://github.com/layershifter/react-source-render/commit/264e05af7120fe2e488bd1f79bd255ad0dbdc2df

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -581,7 +581,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
       statusDiv.setAttribute('role', 'status')
       statusDiv.setAttribute('aria-live', 'polite')
       statusDiv.setAttribute('aria-relevant', 'additions text')
-      Object.assign(statusDiv.style, screenReaderContainerStyles)
+      _.assign(statusDiv.style, screenReaderContainerStyles)
```
layershifter: https://lodash.com/docs/4.17.11#assign
> Note: This method mutates object and is loosely based on Object.assign.

miroslavstastny (approved):
We should enforce IE11 compatibility by lint rules

packages/react/src/components/Dropdown/Dropdown.tsx
kuzhelov: would we need to ensure that `Object.assign` won't be used anywhere in the codebase? I would rather expect that we will introduce polyfill that will ensure that even if `Object.assign` is used somewhere in the codebase, it will be handled correctly. Currently we are just introducing implicit rule of using `_.assign` vs `Object.assign` that will just serve as short-term patch.

I do understand that we may introduce lint rules, but for this particular case it seems that much easier solution will be just to introduce a polyfill - this will ensure that all the previous code would work without any additional changes necessary

mnajdova: I support too adding polyfill, it will be better for the developers. And yep, we can add additional polyfills if there are other constraints in the future.

kuzhelov (approved):

PR #869, opened 2019-02-08
fix(styles): expand css shorthands in order for them to be applied properly

### LINKED ISSUE #237: CSS shorthand properties do not work well when styles are overwritten

mnajdova:
Picks changes from https://github.com/stardust-ui/react/pull/296 Fixes https://github.com/stardust-ui/react/issues/237 Please see additional conversation on this PR https://github.com/stardust-ui/react/pull/296 

## Fix
This PR aims to fix #237 by expanding the css shorthands of the theme's components styles and the styles prop provided to the component while they are merged.

I added the css-shorthand-expand library and provided service for expanding the styles. There are still some points that should be addressed:

### 1. The library provides extending for the following shorthands:
```
  font
  padding
  margin
  border
  borderWidth
  borderStyle
  borderColor
  borderTop
  borderRight
  borderBottom
  borderLeft
  borderRadius
  background
  outline
```
This means that we still don't have any mechanism for expanding the other shorthands like:

```
  animation
  column-rule
  columns flex
  flex-flow grid
  grid-area
  grid-column
  grid-row
  grid-template
  list-style
  offset
  overflow
  place-content
  place-items
  place-self
  text-decoration
  transition
```
### 2. It would have been nice if we would be able to measure the performance of the components and conclude whether this changes would impact them.
I need input whether this is worth exploring further, or whether maybe we should just try to use the library when defining the styles for the themes or applying the styles prop on the components.

### Examples which are fixed with this
1. Let's say that we have a component where inside the styles prop, we are providing some extended vs shorthand styles
```
<Segment content="First Segment" styles={{marginBottom: '10px', margin: '0px'}} />
```
The example is simple, but the `margin: '0px'` can be provided based on some condition, so in some cases it would make sense.
The expected result is that the `margin: '0px'` will override the marginBottom. This PR insures that.

2. We have components that in their styles have usages of `marginRight`, `marginLeft` etc. If the user specify in the `styles` of this component (or in some nested `Provider`), that they want `margin: 0px` this should override the above specified `marginRight | Left`.
In the next example we have the `Icon` component, that by default has right margin of `8px`. Using this should override that margin.
```
<Icon name="call-video" styles={{margin: '0px'}} />
```
The prove that this didn't work before is shown in the screener changes for the icon used inside the radio item component. Basically the `radioItemGroupStyles` is overriding the right margin of the icon with the styles for the icon's slot in the following manner: `margin: '0 10px 0 0'`, but this style was never applied, and we were not even aware that the right margin was not overriden.


packages/react/src/lib/felaExpandCssShorthandsPlugin.ts
```diff
@@ -0,0 +1,33 @@
+import * as _ from 'lodash'
+import * as _expand from 'css-shorthand-expand'
+
+// `css-shorthand-expand` is a CJS library, there are known issues with them:
+// https://github.com/rollup/rollup/issues/1267#issuecomment-446681320
+const expand = (_expand as any).default || _expand
+
+export default () => {
```
bmdalex, line 33: NIT: what about defining the method outside instead of using anonymous method?

```typescript
const expandCssShorthands = styles => {
  // function body
}

export default expandCssShorthands
```

reasons:
- more readable
- can leverage automatic import by name

packages/react/src/lib/felaExpandCssShorthandsPlugin.ts
```diff
@@ -0,0 +1,33 @@
...
+      }
+
+      return { ...acc, [cssPropertyName]: cssPropertyValue }
+    }, {})
+  }
+
+  return expandCssShorthands
+}
+
+const transformKebabCaseKeysToCamelCase = obj => {
```
bmdalex: NIT: the method does not require keys to be kebab, I'd suggest:

~~`transformKebabCaseKeysToCamelCase`~~ -> `transformKeysToCamelCase` or maybe `convertKeysToCamelCase`

packages/react/test/specs/lib/felaExpandCssShorthandsPlugin-test.ts
```diff
@@ -0,0 +1,67 @@
+import felaExpandCssShorthandsPlugin from 'src/lib/felaExpandCssShorthandsPlugin'
+
+const expandCssShorthands = felaExpandCssShorthandsPlugin()
+
+describe('felaExpandCssShorthandsPlugin', () => {
```
bmdalex, line 5: nice tests 👍 couple of things:

1. I'd use `toMatchObject` instead of `toEqual` to compare objects so that we don't have to rely on key order as that is a detail of the CSS library and might change; we're also not interested in the order

2. I'd also add a test for what happens when we use an expanded prop together with its shorthand; something like:

```typescript
test('should merge expanded prop with its shorthand', () => {
  const style = {
    marginTop: '3px',
    margin: '10px',
    marginRight: '15px',
  }

  expect(expandCssShorthands(style)).toMatchObject({
    marginTop: '10px', // overridden by margin: '10px'
    marginRight: '15px', // overridden by marginRight: '15px'
    marginBottom: '10px',
    marginLeft: '10px',
  })
})
```

packages/react/src/lib/felaExpandCssShorthandsPlugin.ts
```diff
@@ -0,0 +1,33 @@
+import * as _ from 'lodash'
+import * as _expand from 'css-shorthand-expand'
+
+// `css-shorthand-expand` is a CJS library, there are known issues with them:
+// https://github.com/rollup/rollup/issues/1267#issuecomment-446681320
+const expand = (_expand as any).default || _expand
+
+export default () => {
+  const expandCssShorthands = styles => {
```
bmdalex: NIT: type for `styles`

bmdalex (approved):
yuppieee 👍 few comments but looks good

packages/react/src/lib/felaExpandCssShorthandsPlugin.ts
mnajdova: This is the definition that the fela renderer expect from a plugin. Other internal plugins that we have are defined in the same manner. Would rather not change at this moment.

packages/react/src/lib/felaExpandCssShorthandsPlugin.ts
mnajdova: I checked fela's plugins, and they use here Object as a type.. Not best, but will add it..

packages/react/test/specs/lib/felaExpandCssShorthandsPlugin-test.ts
mnajdova: Added, actually this is really good case for corner cases, and testing whether one style object within itself will correctly apply the styles. Thanks for the suggestion on this one @Bugaa92

packages/react/src/lib/felaExpandCssShorthandsPlugin.ts
```diff
@@ -0,0 +1,33 @@
...
+export default () => {
+  const expandCssShorthands = (styles: Object) => {
+    return _.keys(styles).reduce((acc, cssPropertyName) => {
+      const cssPropertyValue = styles[cssPropertyName]
+
+      if (typeof cssPropertyValue === 'object') {
+        return { ...acc, [cssPropertyName]: expandCssShorthands(cssPropertyValue) }
+      }
+
+      const expandedProps = expand(_.kebabCase(cssPropertyName), String(cssPropertyValue))
```
kuzhelov: may suggest to use string interpolation approach instead of `String(cssPropertyValue)`
```js
`${cssPropertyValue}`
```

packages/react/src/lib/felaExpandCssShorthandsPlugin.ts
```diff
@@ -0,0 +1,33 @@
...
+      if (expandedProps) {
+        return { ...acc, ...convertKeysToCamelCase(expandedProps) }
+      }
+
+      return { ...acc, [cssPropertyName]: cssPropertyValue }
```
kuzhelov, line 49: here we may simplify and use ternary expression:

```js
return { 
  ...acc,  
  ...(expandedProps 
    ? convertKeysToCamelCase(expandedProps) 
    : { [cssPropertyName]: cssPropertyValue })
}
```

packages/react/src/lib/felaExpandCssShorthandsPlugin.ts
```diff
@@ -0,0 +1,33 @@
...
+    }, {})
+  }
+
+  return expandCssShorthands
+}
+
+const convertKeysToCamelCase = obj => {
```
kuzhelov: nit: this could be a one-liner

layershifter (commented):
As this PR introduces a new plugin for Fela I suggest to measure performance before merge 📝

packages/react/src/lib/felaExpandCssShorthandsPlugin.ts
mnajdova: There are multiple conditions there now, so I wouldn't complicate the return statements

layershifter (approved):
We spent some time to make it more performant, so now it's quite fast 👍

PR #876, opened 2019-02-11
feat(FocusZone): Export Focus Zone utilities

sophieH29:
`FocusZone` utilities have been proven to be useful for Stardust users. For e.g, finding next focusable elements or previous tabbable element, etc.. So let's export it!


jurokapsiar (approved):

kuzhelov (commented):
could you, please, suggest what was exact reason (use case) for making these changes? Also, I am concerned about _all_ the functionality of this module being implicitly part of public API - do we really need to export everything?

sophieH29:
> could you, please, suggest what was exact reason (use case) for making these changes? Also, I am concerned about _all_ the functionality of this module being implicitly part of public API - do we really need to export everything?

updated, exporting now 2 functions

kuzhelov (approved):

PR #880, opened 2019-02-11
feat(status): Add status behavior

kolaps33:
We recognized if status is used in context then the status itself is not navigable with screen reader in the reader mode (like virtual cursor, scan mode, browse mode). 
Se we need to put status as img with proper alt attribute. Then the reader navigate on the status element in reader mode.


jurokapsiar (approved):

packages/react/src/lib/accessibility/Behaviors/Status/statusBehavior.ts
```diff
@@ -0,0 +1,19 @@
+import { Accessibility } from '../../types'
+
+/**
+ * @description
+ * The 'img' role is used to identify an element as image. Alt attribute have to be provided on status component.
+ *
+ * @specification
+ * Adds role='img'.
+ */
+
```
layershifter: Should be there an empty line? 🤔

layershifter (approved):
Please also add a changelog entry 👍

packages/react/src/components/Status/Status.tsx
```diff
@@ -28,6 +36,8 @@ export interface StatusProps extends UIComponentProps {
 
 /**
  * A status graphically represents someone's or something's state.
+ * @accessibility
+ * Don't forget to put appropriate 'alt' attribute describing status. This is necessary for correct screen reader behavior.
```
layershifter: If `alt` prop is so necessary, should be a component prop?

miroslavstastny (approved):

PR #883, opened 2019-02-11
fix(pxToRem): update rem size on static styles render

kuzhelov:
### TODO
- [x] update changelog

----

This PR introduces logic that is necessary to update base rem value of `pxToRem` utility (that was previously cached forever on page load) once newly loaded theme is applying static styles.

This PR leaves it to be a separate discussion whether static styles should be allowed for Stardust themes - the sole purpose of these changes is to address the problem of page font size updates after theme load that we currently have.


layershifter (approved):
Let's go with this to fix the issue and try to get a better solution.

PR #885, opened 2019-02-12
feat(Dropdown): add `clearable` prop

layershifter:
Refs #560.

----

### Why to separate props?

`clearable` sets the behavior, while `clearIndicator` allows to customize the slot.


packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -40,6 +40,22 @@ const dropdownStyles: ComponentSlotStylesInput<DropdownPropsAndState, DropdownVa
     }),
   }),
 
+  clearIndicator: ({ variables: v }): ICSSInJSStyle => ({
```
layershifter: I have a bad feeling about this, actually it duplicates styles for `toggleIndicator`. In the same time, we should allow to customize `clearIndicator` separately.

Move these styles to a separate function? 🤔

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -944,10 +979,15 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
 
     return itemToString(value)
   }
+
+  private isValueEmpty = (value: ShorthandValue | ShorthandValue[]) => {
```
bmdalex: haaa! we needed the function now :)
does it need to be part of the class though?

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -748,6 +778,12 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     }
   }
 
+  private handleClear = () => {
+    const { multiple } = this.props
+
+    this.setState({ value: multiple ? [] : '' })
```
bmdalex: this won't be enough for `search` dropdowns, you need to reset the `searchQuery` as well
```suggestion
    this.setState({ value: multiple ? [] : '', searchQuery: search ? '' : undefined })
```
I'd suggest also extracting the logic from `getInitialAutoControlledState` into a common function that we'd call in both places

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -315,13 +336,22 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
-                      direction: isOpen ? 'top' : 'bottom',
-                      onClick: getToggleButtonProps().onClick,
-                      styles: styles.toggleIndicator,
-                    },
-                  })}
+                  {showClearIndicator
+                    ? Icon.create(clearIndicator, {
+                        defaultProps: {
+                          className: Dropdown.slotClassNames.clearIndicator,
+                          onClick: this.handleClear,
```
bmdalex: the search input/ trigger button is not focused after clicking on the clear button; see:

screen recording 2019-02-12 at 14 24 11

you need to call

```typescript
this.tryFocusSearchInput()
this.tryFocusTriggerButton()
```

in the click handler or maybe write a new function

```typescript
private tryFocusTrigger = () => {
  if (this.props.search) {
    this.inputRef.current.focus()
  } else {
    this.buttonRef.current.focus()
  }
}
```

to incorporate the logic of what trigger needs to be focused depending on props

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -40,6 +40,22 @@ const dropdownStyles: ComponentSlotStylesInput<DropdownPropsAndState, DropdownVa
...
+    display: 'flex',
+    justifyContent: 'center',
+
+    backgroundColor: 'transparent',
+    cursor: 'pointer',
+    userSelect: 'none',
+
+    margin: 0,
+    position: 'absolute',
+    right: pxToRem(5),
```
bmdalex: can we extract these in common styles for `clearIndicator` and `toggleIndicator`?

docs/src/examples/components/Dropdown/Types/DropdownExampleClearable.shorthand.steps.ts
```diff
@@ -0,0 +1,15 @@
+import { Dropdown } from '@stardust-ui/react'
+
+const selectors = {
+  clearIndicator: `.${Dropdown.slotClassNames.clearIndicator}`,
+  triggerButton: `.${Dropdown.slotClassNames.triggerButton}`,
+  item: (itemIndex: number) => `.${Dropdown.slotClassNames.itemsList} li:nth-child(${itemIndex})`,
+}
+
+const steps = [
```
bmdalex, line 9: 👍

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: `ShorthandValue` should be enough for the type here since it's already covering the array type
screenshot 2019-02-12 at 14 31 33

bmdalex (changes requested):
pls take a look at comments

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
bmdalex: ah...just saw your comment https://github.com/stardust-ui/react/pull/885#discussion_r255912391

I'd say to extract in one function

packages/react/src/components/Dropdown/Dropdown.tsx
layershifter: Nice catch 👍  Reused existing methods

packages/react/src/components/Dropdown/Dropdown.tsx
layershifter: Going to leave it as class method as it is very coupled with the component.

packages/react/src/components/Dropdown/Dropdown.tsx
layershifter: Reused values from ``getInitialAutoControlledState()`, nice suggestion 👍

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
layershifter: Done 👍

packages/react/src/components/Dropdown/Dropdown.tsx
layershifter: Actually we have a dedicated type for this `ShorthandCollection`, replaced `ShorthandValue[]` with it

bmdalex (approved):
thanks for addressing, looks good!

.github/test-a-feature.md
```diff
@@ -17,6 +17,14 @@ Test a feature
     - [Important mentions:](#important-mentions)
   - [Run Screener tests](#run-screener-tests)
     - [Local run command](#local-run-command)
+- [Behavior tets](#behavior-tets)
```
mnajdova, line 20: Why are these changes here? Bad merge?

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -86,7 +95,7 @@ export interface DropdownProps extends UIComponentProps<DropdownProps, DropdownS
   inline?: boolean
 
   /** Array of props for generating list options (Dropdown.Item[]) and selected item labels(Dropdown.SelectedItem[]), if it's a multiple selection. */
-  items?: ShorthandValue[]
+  items?: ShorthandCollection
```
mnajdova, line 98: Why was this changed?

mnajdova: Never mind I was reading the changes mixed. Good change! 👍

.github/test-a-feature.md
layershifter: It wasn't generated before 🤔

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -192,6 +201,8 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
       content: false,
     }),
     activeSelectedIndex: PropTypes.number,
+    clearable: PropTypes.bool,
```
mnajdova, line 204: I don't like that we have two new props in the API, but I don't have any better proposal for now...

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -226,6 +237,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
 
   static defaultProps: DropdownProps = {
     as: 'div',
+    clearIndicator: 'close',
```
mnajdova, line 240: Please, don't assume that some icon exists, as in some theme they may not. That's the reason we introduced the Indicator component, instead of using the chevron icons inside the components. Can we use here some unicode char by default if no icon i provided?

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -315,13 +337,22 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
-                      direction: isOpen ? 'top' : 'bottom',
-                      onClick: getToggleButtonProps().onClick,
-                      styles: styles.toggleIndicator,
-                    },
-                  })}
+                  {showClearIndicator
+                    ? Icon.create(clearIndicator, {
+                        defaultProps: {
+                          className: Dropdown.slotClassNames.clearIndicator,
+                          onClick: this.handleClear,
```
mnajdova: Shouldn't onClick be on the override props? Then you can do all necessary for the Dropdown, and then invoke the user's onClick if provided.

packages/react/src/components/Dropdown/Dropdown.tsx
layershifter: https://github.com/stardust-ui/react/blob/master/packages/react/src/components/Input/Input.tsx#L149

`Input` component does the same thing actually. We can refactor them separately later.
The main issue with unicode chars, that I was not able to find a good symbol. Do you have a proposal about it?

packages/react/src/components/Dropdown/Dropdown.tsx
layershifter: Good catch 👍

packages/react/src/components/Dropdown/Dropdown.tsx
mnajdova: I am aware that the Input does the same thing, agreed to tackle this in separate PR. This brings me back to the fact that maybe we should have some icons in the base theme, at least for the things we need in the components (close, arrows etc..) Let's create separate issue for this.

layershifter: Opened #896.

docs/src/examples/components/Dropdown/Types/DropdownExampleClearable.shorthand.tsx
```diff
@@ -0,0 +1,20 @@
...
+  'Alfred Pennyworth',
+  `Scarlett O'Hara`,
+  'Imperator Furiosa',
+  'Bruce Banner',
+  'Peter Parker',
+  'Selina Kyle',
+]
+
+const DropdownClearableExample = () => (
+  <Dropdown clearable items={inputItems} placeholder="Select your hero" />
```
mnajdova, line 17: Maybe we should add additional example for showing customization of the `clearableIndicator`?

mnajdova (approved):

docs/src/examples/components/Dropdown/Types/DropdownExampleClearable.shorthand.tsx
layershifter: If we will have request for it, we can introduce it. For now, I am not sure that we need to add example for an every slot, it can make docs unusable ⛸

PR #886, opened 2019-02-12
feat(ListItem): refactor it to use the Flex component

mnajdova:
The goal of this PR is to refactor the `ListItem` component to use the `Flex` component, instead of the `ItemLayout` component. No visual diffs should be added by this PR.

### BREAKING CHANGES
Class names used for the slots in the `ListItem` for Teams themes, have changed. Instead of the `ItemLayout`'s class names, new slot class names for the `ListItem` component are added. Please, in the context of `ListItem`'s styles do the following replacements:
```
ItemLayout.slotClassNames.header <=> ListItem.slotClassNames.header
ItemLayout.slotClassNames.headerMedia <=> ListItem.slotClassNames.headerMedia
ItemLayout.slotClassNames.main <=> ListItem.slotClassNames.main
ItemLayout.slotClassNames.content <=> ListItem.slotClassNames.content
ItemLayout.slotClassNames.contentMedia <=> ListItem.slotClassNames.contentMedia
ItemLayout.slotClassNames.media <=> ListItem.slotClassNames.media
ItemLayout.slotClassNames.endMedia <=> ListItem.slotClassNames.endMedia
```


packages/react/src/themes/teams/components/List/listItemVariables.ts
```diff
@@ -1,4 +1,9 @@
+import { pxToRem } from '../../../../lib'
+
 export default siteVariables => ({
+  minHeight: pxToRem(48),
+  paddingLeft: pxToRem(20),
+  paddingRight: pxToRem(18),
```
layershifter: Single `padding`?

```suggestion
  padding: `0 pxToRem(20) 0 pxToRem(18)`
```

```suggestion
  rootPadding: `0 pxToRem(20) 0 pxToRem(18)`
```

layershifter (approved):

CHANGELOG.md
```diff
@@ -17,16 +17,19 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### BREAKING CHANGES
+- Changed types of the slot's inside the `ListItem` component (`media`, `content`, `contentMedia`, `header`, `headerMedia` and `endMedia`) from `any` to `ShorthandValue` @mnajdova ([#886](https://github.com/stardust-ui/react/pull/886))
+
 ### Features
 - Export `arrow-up`,`arrow-down` and `chat` SVG icon @VyshnaviDasari ([#873](https://github.com/stardust-ui/react/pull/873))
 - Export `FocusZone`'s utilities @sophieH29 ([#876](https://github.com/stardust-ui/react/pull/876))
-- Add `clearable` prop for `Dropdown` @layershifter ([#885](https://github.com/stardust-ui/react/pull/885))
```
alinais: is this line removed by mistake?

packages/react/src/themes/teams/components/Flex/flexVariables.ts
```diff
@@ -8,6 +8,7 @@ export type FlexVariables = GapValues & PaddingValues
 
 export default (): FlexVariables => ({
   // GAP VALUES
+  'gap.list.item': pxToRem(8),
```
alinais: Would be better to name it `gap.smaller`? Thinking of further other usages.

levithomason: Yes, `gap.smaller` probably makes more sense for now.

***

More context:

We are likely to end up organizing our design system values (variables, etc) similar to the CTI (category, type, item) pattern, as seen in Style Dictionary [here](https://amzn.github.io/style-dictionary/#/properties).  In that case, we'd this would be something like `gap.listItem.small`.

This pattern is great because qualitative terms like "small" are relative.   A "smaller gap" between icons is not the same value as a "smaller gap" between vertical page sections.

However, design system properties that involve component names (i.e. `listItem`) would be defined somewhere at the top of the framework.  We won't define values for a one component inside of another component, as in, we won't define ListItem design system values in the Flex component.

This will soon be my primary focus, working on our specification, design language, and design system structure.

mnajdova: I was actually talking with @kuzhelov and he told me that we are planning to add context dependent values for the props here, that's why I added this value, but I would agree that at this moment `gap.smaller` would make more sense. We can update it later. Thanks for the suggestions!

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -28,7 +28,7 @@ export interface FlexProps {
   space?: 'around' | 'between' | 'evenly'
 
   /** Defines gap between each two adjacent child items. */
-  gap?: 'gap.small' | 'gap.medium' | 'gap.large'
+  gap?: 'gap.small' | 'gap.medium' | 'gap.large' | 'gap.smaller'
```
bmdalex: NIT: we can order by size:` 'gap.smaller' | 'gap.small' | 'gap.medium' | 'gap.large'`

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -71,7 +71,7 @@ class Flex extends UIComponent<ReactProps<FlexProps>> {
 
     space: PropTypes.oneOf(['around', 'between', 'evenly']),
 
-    gap: PropTypes.oneOf(['gap.small', 'gap.medium', 'gap.large']),
+    gap: PropTypes.oneOf(['gap.small', 'gap.medium', 'gap.large', 'gap.smaller']),
```
bmdalex: NIT: we can order by size:` 'gap.smaller' | 'gap.small' | 'gap.medium' | 'gap.large'`

packages/react/src/components/List/ListItem.tsx
```diff
@@ -9,27 +9,38 @@ import {
   ContentComponentProps,
   isFromKeyboard,
 } from '../../lib'
-import ItemLayout from '../ItemLayout/ItemLayout'
```
bmdalex, line 12: 👍 how difficult would be now to remove `ItemLayout` and `Layout` from our repo?

packages/react/src/themes/teams/components/List/listItemVariables.ts
```diff
@@ -21,6 +23,9 @@ export interface ListItemVariables {
 
 export default (siteVariables: any): ListItemVariables => {
   return {
+    minHeight: pxToRem(48),
+    rootPadding: `0 ${pxToRem(18)} 0 ${pxToRem(20)}`,
```
bmdalex, line 34: can you add the new props `minHeight`, `rootPadding `, `contentMediaFontSize `, `contentMediaLineHeight ` to `ListItemVariables` interface?

bmdalex (approved):
small comments but looks good 👍

packages/react/src/components/List/ListItem.tsx
mnajdova: I think the Label is the only other component that uses the Layout (I have to check again...) So no big blockers after this will get in..

packages/react/src/themes/teams/components/List/listItemVariables.ts
mnajdova: Sure, this was updated in some of the new PRs that got in..

PR #890, opened 2019-02-12
fix(Flex): handle cases of falsy values provided as children

kuzhelov:


packages/react/src/components/Flex/Flex.tsx
```diff
@@ -90,20 +90,27 @@ class Flex extends UIComponent<ReactProps<FlexProps>> {
   renderChildren = (gapClasses: string) => {
     const { column, gap, children } = this.props
 
-    return React.Children.map(children, (child: React.ReactElement<any>, index) => {
-      const childElement =
-        child.type && ((child.type as any) as typeof FlexItem).__isFlexItem
+    let isFirstElement = true
+    return React.Children.map(children, (child: any) => {
+      const maybeChildElement =
+        child && child.type && ((child.type as any) as typeof FlexItem).__isFlexItem
```
kuzhelov: addressed failures when `child` is `null` - happens when primitive values (other than `string`) are passed

mnajdova (approved):

packages/react/src/components/Flex/Flex.tsx
layershifter:
```jsx
const isFlexItem = child && child.type && ((child.type as any) as typeof FlexItem).__isFlexItem
// or 
const isFlexItem: boolean = _.get(child, 'type.__isFlexItem')
```

packages/react/src/components/Flex/Flex.tsx
```diff
@@ -90,20 +90,27 @@ class Flex extends UIComponent<ReactProps<FlexProps>> {
...
+      if (maybeChildElement) {
+        isFirstElement = false
+      }
+
       return (
-        <>
-          {renderGap && gap && <Flex.Gap className={cx(`${Flex.className}__gap`, gapClasses)} />}
-          {childElement}
-        </>
+        maybeChildElement && (
```
layershifter, line 108: Do we need a condition there?
Without it we will get:
```jsx
<>
  {null}
  {null}
</>
```
But it will not affect the markup

layershifter: Nevermind, 

> If children is null or undefined, this method will return null or undefined rather than an array.
https://reactjs.org/docs/react-api.html#reactchildrenmap

packages/react/src/components/Flex/FlexItem.tsx
```diff
@@ -78,7 +78,7 @@ class FlexItem extends UIComponent<Extendable<FlexItemProps>> {
       })
     }
 
-    return applyStyles(React.Children.only(children), styles, classes)
+    return children ? applyStyles(React.Children.only(children), styles, classes) : children
```
layershifter:
```suggestion
    return _.isNil(children) ? null : applyStyles(React.Children.only(children), styles, classes)
```

May be this will be more clear?

layershifter (approved):

packages/react/src/components/Flex/Flex.tsx
kuzhelov: yes - but, essentially, we won't like to draw anything if falsy value is provided - and this is exactly the effect logic provides. Thus I would rather leave it the way it is now - as a benefit this won't introduce any additional React elements to output elements tree

PR #891, opened 2019-02-12
fix(types): shorthand default props

kuzhelov:
This PR introduces fix for broken check of default prop name provided to `createShorthandFactory` function. Also, it addresses problems of `Loader` and `Indicator` components that were sneaked in due to absence of this check.

### Was (no warning on unknown prop)

image

### Now

image


packages/react/src/lib/createComponent.tsx
```diff
@@ -18,7 +18,7 @@ export interface CreateComponentConfig<P> {
   render: (config: RenderResultConfig<P>, props: P) => React.ReactNode
 }
 
-export type CreateComponentReturnType<P> = React.SFC<P> & {
+export type CreateComponentReturnType<P> = React.FunctionComponent<P> & {
```
layershifter, line 21:
```suggestion
export type CreateComponentReturnType<P> = React.FC<P> & {
```
nit: can be just `FC`

kuzhelov: might change for now (to be consistent with the rest of the code), but, frankly, not completely sure that we should use short alias name version for some types (`FC` instead of `FunctionComponent`) and, at the same time, would have to use long names for the others. Would rather expect that we would use explicit `FunctionComponent` everywhere - what do you think?

layershifter: `FC` has one big advantage, it's more shorter :)

```tsx
const LayoutExample: React.FunctionalComponent<LayoutExampleProps> = ({ vertical, horizontal })
// vs
const LayoutExample: React.FC<LayoutExampleProps> = ({ vertical, horizontal })
```

kuzhelov: this has a disadvantage - it is not that explicit and easy to read :) And the key differentiator to me is the fact that for the rest of React types we still would need to use log versions of type names. Thus, I would rather stay with consistency and expressivity

kuzhelov: just as an example to compare

### now
```js
createShorthandFactory<TFunctionComponent extends React.FunctionComponent>(
  Component: TFunctionComponent,
  mappedProp?: keyof PropsOf<TFunctionComponent>,
)
```

### changed
```js
createShorthandFactory<TFC extends React.FC>(
  Component: TFC,
  mappedProp?: keyof PropsOf<TFC>,
)
```

miroslavstastny: I vote for descriptive (longer) name

miroslavstastny (approved):

PR #894, opened 2019-02-13
chore(package): update fela

### LINKED ISSUE #516: Keyframes are not behaving as expect when RTL is switched
## Bug Report
Switching RTL is not working as expected when some keyframes are used as part of an animation.

### Steps
When some keyframes are defined as part of the animation, the RTL should be apply to them accordingly when it is switched. If we switch the RTL the keyframes are switched, but going back to LTR is not changing them back to the initial implementation

### Expected Result
Switching to RTL, back to LTR should alternate the keyframes.

### Actual Result
Only the first switch is working as expected, as this is the first time the keyframes are overridden. The problems lies in wrong documentation in Fela. Namely, when we are using the rtl renderer we are adding the `rtl_` prefix to the styles (according to the documentation on fela, this should be applied to the keyframes as well: http://fela.js.org/docs/advanced/RendererConfiguration.html), but unfortunatelly, it is not, and the result is that the keyframes with the same name are written twice. I have question posted in their [Gitter](https://gitter.im/rofrischmann/fela) will update this issue after I receive the final answer from them. 

### Version
0.12.0

### LINKED ISSUE #613: Unable to use Stardust in TS project with --isolatedModules

### LINKED ISSUE #782: Docs: RTL switch is broken locally
### Steps
The same is as in #582. `devMode` that was enabled in #778 breaks RTL switch locally.

rtl-switch

### Expected Result
Everything works

### Version
0.19.0

layershifter:
### No more `const enum`

Fixes #613, see https://github.com/rofrischmann/fela/pull/671.

### Fixed animations in RTL

Fixes #516.

#### Before

animation-before

#### After 

animation-after

### `devMode` is working correctly

Fixes #782.

devmode


miroslavstastny (approved):

mnajdova (approved):
Good job! Thanks for addressing the issues on fela's side :)

PR #897, opened 2019-02-14
fix(Dropdown): inserting char at a cursor position will the cursor to end

### LINKED ISSUE #865: bug(dropdown): inserting char at a cursor position will always move the cursor to end
## Bug Report
Inserting a character in the edit text at any cursor position will make the cursor jump to the end of the text afterwards.
### Steps
Write `aaa` in a Search Dropdown.
Move cursor to position one and write `b`.

### Expected Result
Text should be `abaa` and cursor should be after `b`.

### Actual Result
Text is `abaa` but cursor jumped after the text.

### Version
0.20.0

### Recording
ezgif com-video-to-gif

layershifter:
Fixes #865.

----

It's a known issue in `Downshift`, see:
- https://github.com/downshift-js/downshift#oninputvaluechange
- https://github.com/downshift-js/downshift/issues/217

### Before

fix-dropdown-before

### After

fix-dropdown


silviuaavram (approved):
like a charm. thanks a bunch!

PR #900, opened 2019-02-14
feat(dropdown): autocontrolled mode for open state

bmdalex:
# feat(dropdown): autocontrolled mode for open state

## Description:
- introduced `open`, `defaultOpen` props
- introduced `onOpenChange` event handler

## API:
```jsx
<Dropdown
  open={open}
  onOpenChange={handleOpenChange}
  items={inputItems}
/>
```

where `open` value is part of the state of HOC and `handleOpenChange` is the entry for internal changes of `open`.

## Screenshot:

screen recording 2019-02-14 at 15 06 47

## [Blockers:](#description-blockers)

Currently blocked by possible `Downshift` issue related to internal state handling when `isOpen` prop is used in controlled mode.

The `Dropdown` still works but there is a regression when opening the list of items: 

### The first/last item is not highlighted by default anymore after introducing `open` prop.

[**Here**](https://screener.io/v2/states/rkGqtccHX.stardust-ui-react/feat%2Fdropdown-open-prop/1024x768/Chrome/dropdownexampleshorthandtsx-opens-with-selected-item-highlighted-0) is the expected  screener regression

@silviuavram has more details

### DIFF
--- a/docs/src/examples/components/Dropdown/Usage/DropdownExampleControlled.shorthand.tsx
+++ b/docs/src/examples/components/Dropdown/Usage/DropdownExampleControlled.shorthand.tsx
@@ -6,6 +6,10 @@ const inputItems = ['Bruce Wayne', 'Natasha Romanoff', 'Steven Strange', 'Alfred
 class DropdownExampleControlled extends React.Component {
   state = { open: false }
 
+  handleOpenChange = (e, { open }) => {
+    this.setState({ open })
+  }
+
   render() {
     const open = this.state.open
     return (
@@ -19,10 +23,6 @@ class DropdownExampleControlled extends React.Component {
         <Text weight="semibold" content={`Dropdown open state is: "${open}"`} />
       </Flex>
     )
-  }
-
-  handleOpenChange = (e, { open }) => {
-    this.setState({ open })
   }
 }
 
--- a/docs/src/examples/components/Dropdown/Usage/DropdownExampleControlled.shorthand.tsx
+++ b/docs/src/examples/components/Dropdown/Usage/DropdownExampleControlled.shorthand.tsx
@@ -1,0 +1,29 @@
+import * as React from 'react'
+import { Dropdown, Flex, Text } from '@stardust-ui/react'
+
+const inputItems = ['Bruce Wayne', 'Natasha Romanoff', 'Steven Strange', 'Alfred Pennyworth']
+
+class DropdownExampleControlled extends React.Component {
+  state = { open: false }
+
+  render() {
+    const open = this.state.open
+    return (
+      <Flex gap="gap.large" vAlign="center">
+        <Dropdown
+          open={open}
+          onOpenChange={this.handleOpenChange}
+          items={inputItems}
+          placeholder="Select your hero"
+        />
+        <Text weight="semibold" content={`Dropdown open state is: "${open}"`} />
+      </Flex>
+    )
+  }
+
+  handleOpenChange = (e, { open }) => {
+    this.setState({ open })
+  }
+}
+
+export default DropdownExampleControlled
--- a/packages/react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react/src/components/Dropdown/Dropdown.tsx
@@ -384,7 +384,6 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
                   {this.renderItemsList(
                     styles,
                     variables,
-                    open,
                     highlightedIndex,
                     toggleMenu,
                     selectItemAtIndex,
@@ -475,7 +474,6 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
   private renderItemsList(
     styles: ComponentSlotStylesInput,
     variables: ComponentVariablesInput,
-    open: boolean,
     highlightedIndex: number,
     toggleMenu: () => void,
     selectItemAtIndex: (index: number) => void,
@@ -484,6 +482,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     getInputProps: (options?: GetInputPropsOptions) => any,
   ) {
     const { search } = this.props
+    const { open } = this.state
     const { innerRef, ...accessibilityMenuProps } = getMenuProps(
       { refKey: 'innerRef' },
       { suppressRefError: true },
--- a/packages/react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react/src/components/Dropdown/Dropdown.tsx
@@ -62,6 +62,9 @@ export interface DropdownProps extends UIComponentProps<DropdownProps, DropdownS
   /** The initial value for the index of the currently active selected item, in a multiple selection. */
   defaultActiveSelectedIndex?: number
 
+  /** Initial value for 'open' in uncontrolled mode */
+  defaultOpen?: boolean
+
   /** The initial value for the search query, if the dropdown is also a search. */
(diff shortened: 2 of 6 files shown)


bmdalex:
Currently blocked by possible `Downshift` issue related to internal state handling when `isOpen` prop is used in controlled mode.

The `Dropdown` still works but there is a regression when opening the list of items: 

### The first/last item is not highlighted by default anymore after introducing `open` prop.

[**Here**](https://screener.io/v2/states/rkGqtccHX.stardust-ui-react/feat%2Fdropdown-open-prop/1024x768/Chrome/dropdownexampleshorthandtsx-opens-with-selected-item-highlighted-0) is the expected  screener regression

@silviuavram has more details

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -953,14 +954,21 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
-    _.invoke(this.props, 'onSelectedChange', {}, { ...this.props, value })
+    this.setStateAndInvokeHandler({ value }, 'onSelectedChange')
+  }
+
+  /**
+   * Calls trySetState (for autoControlledProps) and invokes event handler exposed to user.
+   * We don't have the event object for most events coming from Downshift se we send an empty event
+   * because we want to keep the event handling interface
+   */
+  private setStateAndInvokeHandler = (newState: Partial<DropdownState>, eventName: string) => {
```
layershifter: 1. Actually, this is a `handlerName`.
2. `handlerName` should be defined in props.
3. It's more intuitive to have it on first place `setStateAndInvokeHandler('onOpenChange', { open: true })`.
4. It should be named as `trySetStateAndInvokeHandler()`.
5. Let's allow to pass event:  `setStateAndInvokeHandler('onOpenChange', null, { open: true })`.

```suggestion
  private trySetStateAndInvokeHandler = (handlerName: keyof DropdownProps, event: any, newState: Partial<DropdownState>) => {
```

docs/src/examples/components/Dropdown/Usage/DropdownExampleControlled.shorthand.tsx
```diff
@@ -0,0 +1,29 @@
...
+    const open = this.state.open
+    return (
+      <Flex gap="gap.large" vAlign="center">
+        <Dropdown
+          open={open}
+          onOpenChange={this.handleOpenChange}
+          items={inputItems}
+          placeholder="Select your hero"
+        />
+        <Text weight="semibold" content={`Dropdown open state is: "${open}"`} />
```
layershifter, line 23: We should use knobs  for `open` there actually.

bmdalex: how exactly?

If I do something like this:

```typescript
import { Dropdown } from '@stardust-ui/react'
import { DropdownExampleControlledKnobsProps } from './DropdownExampleControlled.knobs'

const inputItems = ['Bruce Wayne', 'Natasha Romanoff', 'Steven Strange', 'Alfred Pennyworth']

const DropdownExampleControlled: React.FC<{ knobs: DropdownExampleControlledKnobsProps }> = ({ knobs }) => (
  <Dropdown
    open={knobs.open}
    onOpenChange={() => {}} // redundant
    items={inputItems}
  />
)

export default DropdownExampleControlled
```
`onOpenChange` prop becomes redundant because `open` flag is controlled by the knob now. I'd still like to keep the meaning of the example and showcase both `open` and `onOpenChange` and I saw this pattern being used for:

- [Popup in controlled mode](https://stardust-ui.github.io/react/components/popup#types-popup-controlled-example)
- [List in controlled mode](https://stardust-ui.github.io/react/components/list#types-selectable-controlled)

@layershifter what do you think?

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: great suggestions, done, thanks!

docs/src/examples/components/Dropdown/Usage/DropdownExampleControlled.shorthand.tsx
layershifter: Agree, let's leave it as is.

docs/src/examples/components/Dropdown/Usage/DropdownExampleControlled.shorthand.tsx
```diff
@@ -0,0 +1,29 @@
...
+      </Flex>
+    )
+  }
+
+  handleOpenChange = (e, { open }) => {
+    this.setState({ open })
+  }
```
layershifter: Even in React docs event handlers are placed before `render()`, https://reactjs.org/docs/handling-events.html

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -458,7 +475,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
   private renderItemsList(
     styles: ComponentSlotStylesInput,
     variables: ComponentVariablesInput,
-    isOpen: boolean,
+    open: boolean,
```
layershifter: We can remove `open` from method params because we can get it from `this.state` now 🐰

layershifter (approved):
Only one small comment left, LGTM 👍

docs/src/examples/components/Dropdown/Usage/DropdownExampleControlled.shorthand.tsx
bmdalex: cool, resolving

docs/src/examples/components/Dropdown/Usage/DropdownExampleControlled.shorthand.tsx
bmdalex: ### I think there are 2 POVs here:

1. React with JS -> your example
2. TypeScript and OOP -> my approach

It is true that in your [reference](https://reactjs.org/docs/handling-events.html) the event handlers are placed before `render`, but we have to take into account that:
- your reference is JS code with ES6 (where classes DO NOT have access modifiers);
- we're writing TypeScript code (where classes DO have access modifiers which makes it more OOP-like). Here there is **a common practice to place public methods before protected and protected before private**. 

### Also, is that a good way to structure it necessarily (events handlers before `render`)?

One might argue that when you read the code you want to start with the entry points (the public methods, `render` in this case) and continue to scroll down as if it was a story, without too many jumps to the private methods that are called from the public ones (and constructor, that would also be placed at the beginning of the class).

I'm just expressing my thoughts here, **I'd like to add others to pitch in so we can stick with one convention. It's not what convention we choose that's important but having consistency.**

@layershifter 
@mnajdova 
@kuzhelov 
@miroslavstastny 
@alinais  
@levithomason

layershifter: Convention is actually exists, we should just follow it: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md

My vote is to follow patterns that exists in React world to make the Stardust more friendly for community users.

bmdalex: I did the order according to React convention for now, let's resolve for now and stick to what we have, tracking in #936

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: done, good catch

PR #901, opened 2019-02-14
proto(dropdown): @mention scenario

bmdalex:
### Description

This PR attempts to create a prototype to show users how they can mention people using `@` key in a rich text editor.

I am using a `contentEditable` div as rich text input and the `Dropdown` component to leverage the list of users and filtering through them. The rest of the functionality is in the prototype section 

### Screenshot
screen recording 2019-02-14 at 16 51 35

### Issues:
- #900 and issue in description - **blocker**
- use `createPortal` instead of mounting another element https://github.com/stardust-ui/react/pull/901#discussion_r256901115

### DIFF
--- a/docs/src/prototypes/dropdowns/AsyncDropdownSearch.tsx
+++ b/docs/src/prototypes/dropdowns/AsyncDropdownSearch.tsx
@@ -1,4 +1,4 @@
-import { Divider, Dropdown, DropdownProps, Header, Loader, Segment } from '@stardust-ui/react'
+import { Divider, Dropdown, DropdownProps, Loader } from '@stardust-ui/react'
 import * as faker from 'faker'
 import * as _ from 'lodash'
 import * as React from 'react'
@@ -34,6 +34,8 @@ const createEntry = (): Entry => ({
 // Prototype Search Page View
 // ----------------------------------------
 class AsyncDropdownSearch extends React.Component<{}, SearchPageState> {
+  private searchTimer: number
+
   state = {
     loading: false,
     searchQuery: '',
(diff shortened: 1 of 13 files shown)


docs/src/prototypes/dropdowns/AsyncDropdownSearch.tsx
```diff
@@ -34,25 +34,55 @@ const createEntry = (): Entry => ({
...
   }
 
-  handleSearchQueryChange = (e: React.SyntheticEvent, { searchQuery }: DropdownProps) => {
+  private handleSearchQueryChange = (e: React.SyntheticEvent, { searchQuery }: DropdownProps) => {
     this.setState({ searchQuery })
     this.fetchItems()
   }
 
-  fetchItems = () => {
+  private fetchItems = () => {
```
layershifter, line 85: What is purpose of these changes? Do we need them?
We don't have an agreement on visibility modifiers so let's revert these changes.

docs/src/prototypes/dropdowns/inputWithDropdown.tsx
```diff
@@ -0,0 +1,136 @@
...
+      />
+    )
+  }
+
+  private showDropdown = () => {
+    this.dropdownExists = true
+    insertNodeAtCursorPosition({ id: this.mountNodeId })
+
+    const node = this.getMountNode()
+    ReactDOM.render(
```
layershifter, line 49: Can you please why we can't use `createPortal()`/`PortalInner` component there? What is a reason for another React's root?

docs/src/prototypes/dropdowns/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+    return
+  }
+
+  const elem = document.createElement('span')
+  elem.id = id
+  if (text) {
+    elem.innerText = text
+  }
+  range.insertNode(elem)
+}
```
layershifter, line 36: Is there any third party dependency that does this work? That is fully tested and supported?
Why we can't use [`caret-pos`](https://www.npmjs.com/package/caret-pos)?

docs/src/prototypes/dropdowns/AsyncDropdownSearch.tsx
Bugaa92: > We don't have an agreement on visibility modifiers so let's revert these changes.

I don't get this part, it's pretty obvious to me that in Typescript class members that are not meant to be used outside the scope of the class are private.

> What is purpose of these changes? Do we need them?

I will revert the changes however because the prototype section does not have the JS version so you might argue the change does not bring any benefit here.

docs/src/prototypes/dropdowns/inputWithDropdown.tsx
bmdalex: opened this PR agains wrong branch, moved to: https://github.com/stardust-ui/react/pull/931/files#r258166628

docs/src/prototypes/dropdowns/utils.ts
bmdalex: opened this PR agains wrong branch, moved to: https://github.com/stardust-ui/react/pull/931#discussion_r258165673

docs/src/prototypes/dropdowns/AsyncDropdownSearch.tsx
bmdalex: opened this PR agains wrong branch, moved to: https://github.com/stardust-ui/react/pull/931/files#r258164858

bmdalex:
opened this PR agains wrong branch, moved to #931

PR #907, opened 2019-02-15
fix(Icon): bordered and circular props not working as expected

### LINKED ISSUE #635: Icon - bordered and circular props not working as expected
### Steps
```
  <div>
    <Icon name="call-video" bordered />
    <Icon name="call-video" circular  />
    <Icon name="call-video" circular bordered />
    <Icon name="call-video" />
  </div>
```

### Expected Result
The result that you expected.
image

Bordered always have border
Circular should set only shape to circular (not add border)

### Actual Result
image

### Version
0.15.0

### Testcase

mnajdova:
This PR fixes https://github.com/stardust-ui/react/issues/635 The problem was that when the `circular` prop was provided to the Icon comoponent, there were always borders added as well. This logic is eliminated in this PR.

## BREAKING CHANGES
Using only the `circular` prop on the `Icon` will no longer add the `border` by default. If you want to persist the `border`, you should add the `bordered` prop on the component as well.

For fixing regressions on the Icons that contain the `circular` prop, add the `bordered` prop as well. For example:

`<Icon name='umbrella' circular /> <=> <Icon name='umbrella' circular bordered />`


docs/src/examples/components/Icon/Variations/IconExampleCircular.shorthand.tsx
```diff
@@ -3,17 +3,54 @@ import { Icon } from '@stardust-ui/react'
...
-    <Icon name="coffee" circular />
-    <Icon name="compass outline" circular />
-    <Icon name="area chart" circular />
+    <Icon
+      name="chess rook"
+      circular
+      variables={siteVars => ({
+        backgroundColor: siteVars.colors.grey[500],
+        color: siteVars.colors.white,
+      })}
```
layershifter: These changes look very strange to me: without additional styling `circular` is useless now. May be we can use `bordered` there instead of variables:

```
<Icon bordered circular name="chess rook" />
```

Should mention that SUI provides a shadow by default:
https://semantic-ui.com/elements/icon.html#bordered
https://semantic-ui.com/elements/icon.html#circular

mnajdova: For the examples we have the `bordered` example that shows the combination with `circular` as well. I can change the examples somehow (I also don't like the fact that we are using variables, but otherwise we cannot show  what circular stands for). The thing we need to address is that, in Teams theme, circular may need to appear without border (that's my understand from the created issue).

mnajdova: Maybe we should altogether remove the example of the circular and use the border example to show how it can have rectangular and circular borders.

kuzhelov: share the same concerns as the ones that are expressed by @layershifter - and would rather like to leave `circular` example in docs.

The approach with `variables` that is currently taken could be problematic also in the following aspect - as now we haven't generalized variables' schema (in a way that some variables would be _guaranteed_ to introduced by any theme), we cannot be absolutely sure that each theme will have these variables used in the example, as well as that these variables will introduce expected effect. 

Also, would rather stay against introducing `shadow` _by default_ - as this seem to be not a feature that is absolutely common for icons (Teams, Github, FB, etc as counterexamples)

## Alternatives


### Use `style` for shadowing icons
What if, for now (maybe before the API of `Icon` is extended) we would introduce the following markup for this example:

image

```jsx
// to emphasize icon's boundaries 
const withShadow = { boxShadow: '0px 0px 5px grey' }

...
    <Icon
      name="chess rook"
      circular
      style={withShadow}
  />
```

You might argue that we would use inline styles here, but with variables, essentially, we would do the same. The difference this approach introduces is its simplicity, readability, as well as that we will be absolutely sure about produced effect of this style - and won't rely on colors defined by theme.

### Use `style` to make icons non-transparent
Essentially, the approach is the same as the previous one, with final example's look being the only difference:

image

```jsx
// to emphasize icon's boundaries 
const white = { background: 'white' }

const IconExampleCircular = () => (
  <Flex padding='padding.medium' style={{ background: 'grey' }}>
    <Icon
      name="chess rook"
      circular
      style={white}
    />
   ...
```

### Use `bordered` prop
We may use `bordered` prop in these examples, and leave all the examples of `Bordered` section to be square icons (as, essentially, @layershifter has suggested)

In future, however, we might think to extend Icon's API (by introducing ability to make icon to be non-transparent, like `transparent={false}`) and adjust these examples accordingly. 

What do you, folks, think?

mnajdova: I would vote for using the `boredered` prop, and we can update it later. We are saying all the time not use style, so I wouldn't like us to have it in the examples. Will revert the bordered example as it was before.

kuzhelov: great, totally agree - thank you!

kuzhelov (approved):

layershifter (approved):

bmdalex (approved):

PR #910, opened 2019-02-15
perf: Add `List` performance example

miroslavstastny:
Add performance test for `List` component.


docs/src/examples/components/List/Performance/index.tsx
```diff
@@ -0,0 +1,16 @@
...
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import PerformanceSection from 'docs/src/components/ComponentDoc/PerformanceSection'
+
+const Performance = () => (
+  <PerformanceSection title="Performance">
+    <ComponentExample
+      title="Default"
+      description="A default test."
+      examplePath="components/List/Performance/List.perf"
+    />
```
levithomason: Are the perf results isolated to the `List.perf` and not affected by the use of `ComponentExample`?

levithomason, line 12: Are the perf results isolated to the `List.perf` and not affected by the use of `ComponentExample`?

kuzhelov: yes, those are not affected by `ComponentExample` - only "component under test" is measured:
```jsx
  <Profiler > // from React
     <ComponentUnderTest /> // with styles of Teams theme, for now
  ...
```

layershifter (approved):

kuzhelov (approved):

PR #911, opened 2019-02-15
fix(docs): new theme from switcher not applied when the page is changed

mnajdova:
This PR fixed the issue for the new theme chosen from the theme switcher not being applied when the component page is changed.

## Before
theme-switching-broken

## After
theme-switching-fixed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixes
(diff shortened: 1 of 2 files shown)


alinais (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixes
+- Fix theme switcher not applying the new theme when the component's page is changed @mnajdova ([#911](https://github.com/stardust-ui/react/pull/911))
```
levithomason, line 21: Note, the CHANGELOG should only be used for public changes to the `@stardust-ui/react` package.

Changes to the docs should not get CHANGELOG entries.  Someday, we may want to maintain a docs/CHANGELOG.

PR #913, opened 2019-02-15
feat(Popup): Expose Popup's content Ref

sophieH29:
Currently, there is no way for users to get ref to the Popup content. This PR aims to fix it

### DIFF
--- a/packages/react/src/components/Popup/Popup.tsx
+++ b/packages/react/src/components/Popup/Popup.tsx
@@ -112,7 +112,7 @@ export interface PopupProps
   trigger?: JSX.Element
 
   /** Ref for Popup content DOM node. */
-  popupRef?: React.Ref<HTMLElement>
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Popup/Popup.tsx
```diff
@@ -108,6 +110,9 @@ export interface PopupProps
 
   /** Element to be rendered in-place where the popup is defined. */
   trigger?: JSX.Element
+
+  /** Ref for Popup content DOM node. */
+  popupRef?: React.Ref<HTMLElement>
```
layershifter: Actually, this should be `contentRef` because `Popup` has two slots `trigger` and `content`.

sophieH29: yup, you're right

layershifter (approved):

PR #917, opened 2019-02-18
fix(FocusTrapZone): Do not hide aria-live regions from accessibility tree

### LINKED ISSUE #908: Dropdown aria-live does not work when in Popup with a trap zone or in Dialog
## Bug Report

Dropdown aria-live does not work when in Popup with a trap zone or in Dialog

### Steps
Create an example with Dropdown inside of the Popup trap zone.
Observe aria-live.

### Expected Result
Amounts of available results are narrated

### Actual Result
Amounts of available results are not narrated.
Aria live regions are hidden by aria-hidden by the popup trap focus mechanism. These regions should be excluded.

### Version
x.y.z

### Testcase

sophieH29:
closes https://github.com/stardust-ui/react/issues/908


jurokapsiar (approved):

packages/react/src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -296,13 +296,19 @@ export class FocusTrapZone extends React.Component<FocusTrapZoneProps, {}> {
...
     }
 
-    // loop through all body's children, except the last one - which is the popup
-    for (let index = 0; index < bodyChildren.length - 1; index++) {
-      const element = bodyChildren[index]
-
-      if (element.getAttribute('aria-hidden') !== 'true') {
-        element.setAttribute('aria-hidden', 'true')
-        element.setAttribute(HIDDEN_FROM_ACC_TREE, 'true')
+    for (let index = 0; index < bodyChildren.length; index++) {
```
silviuaavram, line 299: does a for ... of loop work? also with type included. something like:

```suggestion
    for (const currentChild of bodyChildren as HTMLElement[]) {
```

packages/react/src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -296,13 +296,19 @@ export class FocusTrapZone extends React.Component<FocusTrapZoneProps, {}> {
...
+      const currentChild = bodyChildren[index] as HTMLElement
+      const isOrHasFocusTrapZone =
+        currentChild === this._root.current || currentChild.contains(this._root.current)
+      const isAriaLiveRegion = currentChild.hasAttribute('aria-live')
```
silviuaavram, line 303: not a big deal, but no need to calculate this if `isOrHasFocusTrapZone` is false. you can just add it in the if isOrHasFocusTrapZone branch, get value through `getAttribute` and if that `!!<that-value>` is not true, do your setters.

silviuaavram (approved):

packages/react/src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: in case of `isOrHasFocusTrapZone` I don't need to read attribute, but yes, I can invert condition there, but then will need to rename it which I think can be less readable. The reason why I actually put all those calculation in variables is to have good readability of what is happening.
But I can do it for `isAriaLiveRegion` and change it to
```js
const isNotAriaLiveRegion = !currentChild.hasAttribute('aria-live')
```

Just to note, `if that !!<that-value>` won't work if someone will set value to attribute others that 'true', 

```js
document.body.setAttribute('aaa', 'aaaa')
!!document.body.getAttribute('aaa') // <--- return true
```
Please, let me know what you think

packages/react/src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: for..loop works and it's the fastest method of looping  (if for e.g. to compare to forEach). But I like your way too, will change it. Thank you!

jurokapsiar: If you do use for/of, please test with IE, see https://stackoverflow.com/questions/22754315/for-loop-for-htmlcollection-elements

kuzhelov (approved):

packages/react/src/lib/accessibility/FocusZone/FocusTrapZone.tsx
kuzhelov: my vote for explicitness the code already has - this won't be an optimisation in any sense if we would skip an access to this attribute. Thus, if there won't be any alternative solution that would be on par in terms of expressiveness, I would prefer in a way it is now

packages/react/src/lib/accessibility/FocusZone/FocusTrapZone.tsx
sophieH29: We use Babel transpiler of ES6, aren't we?

sophieH29: @layershifter is working on that. In meantime, will keep basic loop

PR #918, opened 2019-02-18
chore(dangerjs): add check for a changelog entry

layershifter:
This PR adds a new function to our `dangerfile` that checks that changelog entries are added to the correct section.


dangerfile.ts
```diff
@@ -1,10 +1,40 @@
...
+
+/**
+ * This function asserts that added entries into the changelog file are placed in the right section.
+ */
+const hasAddedLinesAfterVersionInChangelog = async (): Promise<boolean> => {
+  const changelogContent = fs.readFileSync(path.resolve(__dirname, CHANGELOG_FILE)).toString()
+  const versionLineNumber =
+    changelogContent
+      .split('\n')
+      .findIndex(line => line.startsWith('<!--------------------------------[ ')) - 1
```
miroslavstastny: This seems to be too fragile. Adding or removing a single `-` will break if without being ever noticed.
I would rather search for the second `<!--`.

miroslavstastny (approved):

miroslavstastny:
Tested on #858 which has the real changelog issue, successfully caught! 👍

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixes
 - Display correctly images in portrait mode inside `Avatar` @layershifter ([#899](https://github.com/stardust-ui/react/pull/899))
+- Expose `Popup`'s content Ref @sophieH29 ([#913](https://github.com/stardust-ui/react/pull/913))
```
layershifter, line 25: image

_Next time..._ 😸

dangerfile.ts
layershifter: Updated 👍

PR #920, opened 2019-02-18
fix(Button): use variables in the styles Teams theme

mnajdova:
Couple of refactorings in the styles for misusing variables


packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -215,10 +215,10 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps & ButtonState, any> = {
                   right: `-${pxToRem(primaryBorderWidth * 2)}`,
                   bottom: `-${pxToRem(primaryBorderWidth * 2)}`,
                   left: `-${pxToRem(primaryBorderWidth * 2)}`,
-                  borderWidth: `${pxToRem(primaryBorderWidth)}`,
+                  borderWidth: pxToRem(primaryBorderWidth),
```
kuzhelov, line 218: 👍 thank you!

packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -215,10 +215,10 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps & ButtonState, any> = {
...
                   borderStyle: 'solid',
-                  borderColor: `${primaryBorderColorFocusIndicator}`,
-                  borderRadius: '3px',
+                  borderColor: primaryBorderColorFocusIndicator,
```
kuzhelov, line 220: we might think about making names consistent here as well - as it seems that `focusIndicator` largely states for reflecting 'focus' state aspects. Maybe we could rename it to `primaryBorderColorFocused`?

mnajdova: I agree, but should note that it would mean introducing breaking changes, just because of refactoring? Are we ok with this?

packages/react/src/themes/teams/components/Button/buttonVariables.ts
```diff
@@ -132,5 +133,6 @@ export default (siteVars: any): ButtonVariables => {
     textSecondaryColorHover: siteVars.brand04,
 
     boxShadow: siteVars.shadowLevel1,
+    focusedBorderRadius: pxToRem(3),
```
kuzhelov: just to introduce the same order that is implied by classification we are, essentially, converging to, maybe we might want to mention state at the end (like `borderRadiusFocused`)?

image

kuzhelov (approved):

packages/react/src/themes/teams/components/Button/buttonVariables.ts
mnajdova: Agreed, thanks!

packages/react/src/themes/teams/components/Button/buttonStyles.ts
kuzhelov: no, surely - in that case lets better wait with them. Thank you!

PR #921, opened 2019-02-18
chore: move `customProptypes` to a separate package

layershifter:
Addresses #433.

- moves `customPropTypes` to a separate package
- adds docs about managing packages and dependencies


packages/react/package.json
```diff
@@ -1,29 +1,11 @@
...
-  "sideEffects": false,
-  "files": [
-    "dist"
-  ],
-  "scripts": {
-    "build": "gulp bundle:package --package react"
-  },
+  "bugs": "https://github.com/stardust-ui/react/issues",
   "dependencies": {
+    "@stardust-ui/react-proptypes": "^0.21.1",
```
layershifter: There are no real changes except this line

.github/CONTRIBUTING.md
```diff
@@ -98,6 +105,108 @@ Focused component needs to be clearly visible. This is handled in Stardust by fo
...
+### Add a new package
+
+#### Run `lerna create`
+ 
+You should to run `lerna create` command to create a new package
+
+- we are using `@stardust-ui` namespace on NPM to publish our packages
+- the directory name should not contain any namespace prefix and can be prefixed with the library name if the
+  implementation is not framework agnostic
+- please provide a meaningful description to a package in the matched field
```
kuzhelov, line 122: maybe: "in the _corresponding_ field _of package.json_"?

.github/CONTRIBUTING.md
```diff
@@ -98,6 +105,108 @@ Focused component needs to be clearly visible. This is handled in Stardust by fo
...
+- use `https://github.com/stardust-ui/react/tree/master/packages/__DIRECTORY_NAME__` as `homepage`
+- our packages are published with MIT license, please follow it until you will have a specific requirements
```
kuzhelov: sorry, cannot completely understand what this part of the sentence about:

```
please follow it until you will have a specific requirements
```

Is it some general legal form? I mean, could we just mention something like that:

```
our packages are currently published with MIT license
```

.github/CONTRIBUTING.md
```diff
@@ -98,6 +105,108 @@ Focused component needs to be clearly visible. This is handled in Stardust by fo
...
+homepage: https://github.com/stardust-ui/react/tree/master/packages/react-proptypes
+license: (ISC) MIT
+entry point: (lib/react-proptypes.js)
+git repository: (https://github.com/stardust-ui/react.git)
+```
+
+#### Update `package.json`
+
+After a package will be created we need to add necessary changes to a newly created `package.json`.
+These changes are required to get working package publishing and internal tooling.
```
kuzhelov: nit: These changes are required to _setup_ internal tooling and package publishing

.github/CONTRIBUTING.md
```diff
@@ -98,6 +105,108 @@ Focused component needs to be clearly visible. This is handled in Stardust by fo
...
+```diff
+-  "scripts": {
+-    "test": "echo \"Error: run tests from root\" && exit 1"
+-  },
++  "scripts": {
++    "build": "gulp bundle:package:no-umd --package react-proptypes"
++  },
+```
+
+Don't forget to fill a correct directory name, you can also use `gulp bundle:package` to bundle your package with UMD.
```
kuzhelov: nit: Don't forget to _provide_ (or _fill in_) a correct directory name

.github/CONTRIBUTING.md
```diff
@@ -98,6 +105,108 @@ Focused component needs to be clearly visible. This is handled in Stardust by fo
...
+
+#### Create `tsconfig.json`
+
+If your package uses TypeScript, please also create a new `tsconfig.json` and place it in `packages/__DIRECTORY_NAME__`.
+An example config:
+
+```json
+{
+  "extends": "../../build/tsconfig.common",
```
kuzhelov, line 183: lets make it explicit in the text that this common configuration file is meant to be extended

.github/CONTRIBUTING.md
```diff
@@ -98,6 +105,108 @@ Focused component needs to be clearly visible. This is handled in Stardust by fo
...
+```
+yarn syncpack format
+```
+
+### Add a new dependency
+
+Please always use [`lerna add`][16] to manage all dependencies including internal packages.
+
+```yarn
+lerna add @stardust-ui/react-proptypes packages/react
```
kuzhelov: could we, please, introduce comment for this command that would describe _what_ is the dependency and _to which project_ it is introduced - as these bits are not absolutely evident from the command itself (if not to look in the docs, of course:)

build/gulp/plugins/gulp-example-source.ts
```diff
@@ -4,7 +4,7 @@ import * as prettier from 'prettier'
 import * as through from 'through2'
 import * as Vinyl from 'vinyl'
 
-import * as prettierConfig from '../../../.prettierrc.json'
+const prettierConfig = require('../../../.prettierrc.json')
```
kuzhelov: could you, please, suggest why this change was necessary?

package.json
```diff
@@ -121,6 +121,7 @@
     "screener-runner": "^0.10.30",
     "simulant": "^0.2.2",
     "snyk": "^1.110.2",
+    "syncpack": "^4.3.1",
```
kuzhelov: this is an additional dependency introduced - lets provide a note in the dedicated PR section about that (currently there is one, but lets just make it more visible)

package.json
```diff
@@ -26,7 +26,7 @@
     "stats:build": "gulp stats",
     "stats:save": "gulp stats:save",
     "satisfied": "satisfied --skip-invalid",
-    "pretest": "yarn satisfied",
+    "pretest": "yarn satisfied && yarn syncpack list-mismatches",
```
kuzhelov: could you, please, suggest what is the approx time cost of this change? I mean, what is the relative change of time necessary to pretest commit changes?

packages/react-proptypes/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@stardust-ui/react-proptypes",
+  "description": "Set of custom reusable propTypes for React components.",
```
kuzhelov: would suggest to capitalize `propTypes`->`PropTypes`, to make it evident that this is the same term that is used in React docs: https://reactjs.org/docs/typechecking-with-proptypes.html

package.json
layershifter: > * adds [syncpack](https://github.com/JamieMason/syncpack) to align `package.json` files and our dependencies across packages

It was mentioned in the initial comment. Can you please clarify what changes do you expect?

package.json
layershifter: `pretest` will be run in CI, too. My idea is too check dependencies there, example throw will be:
```
$ syncpack list-mismatches
lodash ^4.17.11, ^4.17.10
```

.github/CONTRIBUTING.md
layershifter: Actually, this will happen on `lerna create` before `package.json` will be created

packages/react-proptypes/package.json
kuzhelov: also, we should make it clear in the description that these PropTypes are, in fact, meant to be Stardust specific (not just general-use ones), as currently we have the following provided by the package: `itemShorthandWithKindProp`, `wrapperShorthand`

.github/CONTRIBUTING.md
layershifter: Updated to: `our packages are currently published with MIT license, please follow it until you will have specific legal requirements`

kuzhelov (approved):
again, not sure that we should make this split now, but the implementation of these changes raises no concerns to me

.github/CONTRIBUTING.md
layershifter: Added:
`This config will extend a common TS config that is used in all packages. You can add specific options for the package here.`

build/gulp/plugins/gulp-example-source.ts
layershifter: Now we have a specific `tsconfig.json` for `react-proptypes` that doesn't include `/build` directory, so it will fail with: 
```
../../build/gulp/plugins/gulp-example-source.ts(7,33): error TS2732: Cannot find modu
le '../../../.prettierrc.json'. Consider using '--resolveJsonModule' to import module
 with '.json' extension
```

Did I miss something?

packages/react-proptypes/package.json
layershifter: Changed to: `Set of custom reusable PropTypes for React components, some of them are specific for Stardust UI.`

layershifter:
Marking as blocked as we decided to include these changes to a release after the next.

PR #923, opened 2019-02-19
chore(package): use Babel for Jest

layershifter:
This PR switches Jest to `babel-jest`. Jest and Babel related stuff was moved to `@stardust-ui/internal-tooling` package, it will host our tooling dependencies and configs.

### Removed dependencies

`ts-jest` was replaced by `babel-jest`.

### Moved dependencies

`@babel/core`, `@babel/preset-typescript`, `jest`, `jest-axe` and matching types were moved to a separate package `@stardust-ui/internal-tooling`.

### Added dependencies

`babel-jest`, `@babel/preset-env`, `@babel/preset-react`, `@babel/plugin-proposal-class-properties`, `@babel/plugin-transform-runtime`.

Minimal set of deps to get working our existing tests 🚀 

### `.babelrc`

Contains a minimal config right now as we are running only tests, will be extended in future PRs.

### `@stardust-ui/internal-tooling`

The goal: have a package that can be required by our consumer and allow to him use our build system and configs. Will be extended with `pikapack` in future PRs to give the final result.

Not sure about naming, `build-tooling`, `build-tools`, `package-tools`?


bmdalex (approved):
👍

packages/react/src/lib/mergeThemes.ts
```diff
@@ -223,9 +223,9 @@ const mergeThemes = (...themes: ThemeInput[]): ThemePrepared => {
     // Use the correct renderer for RTL
     acc.renderer = acc.rtl ? felaRtlRenderer : felaRenderer
 
-    acc.fontFaces = mergeFontFaces(...acc.fontFaces, ...next.fontFaces)
+    acc.fontFaces = mergeFontFaces(...(acc.fontFaces || []), ...(next.fontFaces || []))
```
kuzhelov: actually there is no need to provide default values for accumulated theme (`acc`), as the process is guaranteed to start with initialized fields (and there shouldn't be any field reset back to non-initialized value during the process)

docs/src/examples/components/Dialog/Usage/DialogExampleCallbacks.shorthand.tsx
```diff
@@ -23,7 +23,6 @@ export default class DialogExampleCallbacks extends React.Component<
   }
 
   handleClose = (callbackName: string) => () => {
-    console.log(callbackName)
```
kuzhelov, line 26: 👍

jest.config.js
```diff
@@ -1,5 +1,5 @@
+const commonConfig = require('@stardust-ui/internal-tooling/jest')
```
kuzhelov: nit: may be simplified a bit, given that there is only single place where this is used:

```js
const { coverageReporters } = require( ... )
```

packages/internal-tooling/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@stardust-ui/internal-tooling",
+  "version": "0.23.1",
+  "dependencies": {
+    "@babel/core": "^7.3.4",
```
kuzhelov, line 5: could you, please, mention newly added dependencies in the PR's description?

packages/internal-tooling/babel/index.js
```diff
@@ -0,0 +1,25 @@
+module.exports = {
```
kuzhelov, line 1: not sure that we should introduce additional package for the tooling stuff - I would rather expect it to be something like `scripts/config/build` dir of the root (private) package. Could you, please, suggest why it is better to have this as package?

Another thing where I see package atomicity being broken is its `lerna-alisases` - essentially, this makes this package to be dependent on the env setup where it is used. You might argue that this is called `internal-tooling`, which makes a hint that it is not universal applicable - and I agree here, but then would still have a question of why we need this as package and not just as a content of some tooling directory in our project :)

packages/react/src/lib/mergeThemes.ts
layershifter:
```
 FAIL  packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
  ● Test suite failed to run

    TypeError: Invalid attempt to spread non-iterable instance

      227 |     acc.renderer = acc.rtl ? felaRtlRenderer : felaRenderer
      228 | 
    > 229 |     acc.fontFaces = mergeFontFaces(...acc.fontFaces, ...next.fontFaces)
          |                     ^
      230 | 
      231 |     acc.staticStyles = mergeStaticStyles(...acc.staticStyles, ...next.staticStyles)
      232 | 

      at _nonIterableSpread (../../node_modules/@babel/runtime/helpers/nonIterableSpread.js:2:9)
      at _toConsumableArray (../../node_modules/@babel/runtime/helpers/toConsumableArray.js:8:60)
      at src/lib/mergeThemes.ts:229:21
```

packages/internal-tooling/package.json
layershifter: Done 👍

packages/internal-tooling/babel/index.js
layershifter: Agree with your point, `lerna-alias` is a leaky abstraction in this case. It moved mapping to per-package configs, so the main config doesn't  know anything about Lerna now 👍 

_Why we need a separate package?_ I expressed this in the PR description, I want to provide something like Create React App, but for our end-users to allow create and publish their Stardust components.

packages/react/src/lib/mergeThemes.ts
layershifter: It throws on `next.*` actually, updated to:
```
acc.fontFaces = mergeFontFaces(...acc.fontFaces, ...(next.fontFaces || []))
acc.staticStyles = mergeStaticStyles(...acc.staticStyles, ...(next.staticStyles || []))
```

Better ideas? 🤔

kuzhelov: yes - but isn't it that `acc.fontFaces` is always an array, given the value empty theme is initialized with?

```js
const emptyTheme = {
    siteVariables: {},
    componentVariables: {},
    componentStyles: {},
    fontFaces: [],
    staticStyles: [],
    icons: {},
    animations: {},
  } as ThemePrepared
```

Thus, I would expect that fallback value to be necessary for `next` fonts only (as well as static styles)

packages/internal-tooling/babel/index.js
kuzhelov: great update related to Lerna stuff 👍 

One thing about package introduced - could you, please, just walk me through the client scenario that would need that? I mean, currently I see this scenario to be the following: 
- client already have all the necessary scaffolding to build react project (`create-react-app`)
- client introduces Stardust as a library and just consumes `createComponent` in case if Stardust component needs to be created

So, here I don't see where and how the tooling package might be necessary, given that Stardust components creation is not dependent on the client's set of tooling used - could you, please, elaborate more on that (sorry if I am missing something)

packages/react/src/lib/mergeThemes.ts
layershifter: Got it, have fixed already 👍

packages/internal-tooling/babel/index.js
kuzhelov: have discussed with @layershifter - while it was possible to delay this step of package segregation to the future, there is one very appealing benefit of this move (amongst other ones, but this is the most appealing to me):
- it would allow us to separate `package.json` dependencies on those that are really needed by the logic and those that are necessary for the tooling part (now `internal-tooling` package)

kuzhelov (approved):

Issue #494, opened 2018-11-19
Rewrite our components using `createComponent` utility


bmdalex:
### Problem description

We added the `createComponent` factory that is going to be used for:
- wrap 3rd party components in order to achieve Stardust capabilities (especially styling)
- creating custom Stardust components;
- rewriting existing Stardust components.

**The goal is to rewrite at lest 4-5 components using `createComponent` utility before we export it**, as we might need to make changes in the API and we want to minimize these before we expose it to users.

### Proposed solution

Here is a list of the components that can be picked up for rewriting:
- [ ] Accordion
- [ ] Attachment
- [ ] Avatar
- [ ] Button
- [ ] ButtonGroup
- [ ] Chat
- [ ] ChatItem
- [ ] ChatMessage
- [ ] Divider
- [ ] Form
- [ ] FormField
- [ ] Grid
- [ ] Header
- [ ] HeaderDescription
- [ ] Icon
- [ ] Image
- [ ] Input
- [ ] ItemLayout
- [ ] Label
- [ ] Layout
- [ ] List
- [ ] ListItem
- [ ] Menu
- [ ] MenuItem
- [ ] Portal
- [ ] Popup
- [ ] PopupContent
- [ ] RadioGroup
- [ ] RadioGroupItem
- [ ] Segment
- [ ] Status
- [ ] Text

Please update the list with a reference to your username when you pick it up. It'd be best to have one component per PR unless there are a lot of connected changes in `createComponent` API.
**e.g.:**

#### When you pick up the component:
- [ ] Text taken by @John_Doe 

#### When you finish the work related to the component:
- [x] Text fixed by @John_Doe in PR #999

Once we merge 4-5 components we will create a PR for exporting `createComponent`

bmdalex:
fyi
@kuzhelov 
@layershifter  
@levithomason 
@miroslavstastny 
@mnajdova 

 and
@alinais 
@priyankar205 
for assigning priority

levithomason:
See #475

PR #926, opened 2019-02-19
fix(docs): generate docs for components produced by createComponent() func

### LINKED ISSUE #502: Supporting generating info files for components using the renderComponent function

kuzhelov:
This PR fixes #502.

Introduced changes open way to generate docs descriptions for components declared with lightweight `createComponent` function.

`react-typescript-docgen` was removed from project's dependencies, with the only necessary parsing logic being preserved (with basic clean-up, but without in-depth refactoring being applied).

Apart from implementation changes, this PR doesn't introduce anything which effect would be visible to the client.

Unblocks #494.


build/gulp/plugins/util/docgen.ts
```diff
@@ -0,0 +1,780 @@
...
+
+const defaultPropFilter = (prop, component) => {
+  // skip children property in case it has no custom documentation
+  if (prop.name === 'children' && prop.description.length === 0) {
+    return false
+  }
+  return true
+}
+
+const getComponentSymbolOfType = (type: MaybeIntersectType) => {
```
kuzhelov, line 189: this is where original logic of `react-docgen-typescript` was extended

miroslavstastny (approved):
Tested on `Box` component, no issue found.

build/gulp/plugins/util/docgen.ts
```diff
@@ -0,0 +1,780 @@
+import * as fs from 'fs'
+import * as path from 'path'
+import * as ts from 'typescript'
+
+// currentDirectoryName is used to trim parent fileNames
```
mnajdova, line 22: One small suggestion, can we document somewhere, at least with comment that these things are picked up from react-docgen-typescript? And maybe add comments where we added our changes and why they were necessary. It would be helpful, if we need in future to add something else, or fix something. Looks good otherwise, thanks for these updates, will unblock us!

mnajdova (approved):

build/gulp/plugins/util/docgen.ts
```diff
@@ -0,0 +1,797 @@
+import * as fs from 'fs'
+import * as path from 'path'
+import * as ts from 'typescript'
+
+/**
+ * THIS MODULE is largely based on the parser's logic introduced
```
mnajdova: Thanks a lot for adding this 👍 very good explanation!

Issue #875, opened 2019-02-11
feat(Drodpown): create accessiblity behaviour


silviuaavram:
## Feature Request
At the moment, all accessibility is implemented in Dropdown.tsx and its child components. Encapsulate all this into behaviours.

jurokapsiar:
closing for no activity

PR #927, opened 2019-02-19
feat(mixed): add `accessibility` prop to all components

### LINKED ISSUE #633: 'accessibility' attribute is not available on all Stardust components.
As the preferred way to set Aria role is via Accessibility behaviours which can be passed via `accessibility` component attribute, I expect that every Stardust component would allow it. Segment, nor Text allow it for now.

layershifter:
Fixes #633.


packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -14,6 +16,12 @@ import {
 } from '../../lib'
 
 export interface AvatarProps extends UIComponentProps {
+  /**
+   * Accessibility behavior if overridden by the user.
+   * @default defaultBehavior
+   */
+  accessibility?: Accessibility
```
mnajdova, line 23: Can we add this in interface (or if it should be use in all UI components, maybe even move to the UIComponentProps)? At least common interface would allow us to reuse the same description, although not sure how we can add different @defaults there..

mnajdova: Btw will we still need these @default description, after removing `react-docgen-typescript`? @kuzhelov

packages/react/src/components/Chat/ChatItem.tsx
```diff
@@ -46,19 +54,22 @@ class ChatItem extends UIComponent<ReactProps<ChatItemProps>, any> {
 
   static propTypes = {
     ...commonPropTypes.createCommon({ content: false }),
+    accessibility: customPropTypes.accessibility,
```
mnajdova: I would definitely expect this to be in the commonPropTypes.createCommon factory.. We can use {accessibility: false} if we don't want it there for some components

layershifter: I like `createCommon()`, but https://github.com/stardust-ui/react/pull/647#discussion_r247304323

packages/react/src/components/Avatar/Avatar.tsx
layershifter: I'm note sure that the issues with `@default` will be solved right now while some components have different defaults 🐱

packages/react/src/components/Chat/ChatItem.tsx
mnajdova: Ok, if we agree to remove this abstraction, then we should do this, but having now some property being defined directly, and some of them as result of a function is really confusing. Let's create then issue for this, because I missed that comment and was not aware of it by now...

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -104,7 +104,7 @@ class Accordion extends AutoControlledComponent<ReactProps<AccordionProps>, any>
         }),
       ),
     ]),
-    accessibility: PropTypes.func,
+    accessibility: customPropTypes.accessibility,
```
mnajdova: we can remove this from here now

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -197,6 +197,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
 
   static propTypes = {
     ...commonPropTypes.createCommon({
+      accessibility: false,
```
mnajdova, line 213: Are we sure which component should not have the accessibility property?

packages/react/src/components/Accordion/Accordion.tsx
layershifter: Nice catch 👍

packages/react/src/components/Dropdown/Dropdown.tsx
layershifter: Yep, it's not implemented yet #875

mnajdova (approved):

PR #928, opened 2019-02-19
docs(ComponentExample): remove useless components and cleanups

layershifter:
This PR makes some cleanups in components in our docs.


docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -433,19 +435,22 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
...
         content: 'Reset',
+        key: 'reset',
         onClick: this.resetSourceCode,
         disabled: !wasCodeChanged,
       },
       {
         active: copiedCode, // to show the color
         icon: copiedCode ? { color: 'green', name: 'check' } : 'copy',
         content: 'Copy',
+        key: 'copy',
```
layershifter, line 453: Fixed issues with `key` 🚀

package.json
```diff
@@ -78,7 +78,6 @@
     "express": "^4.15.4",
     "extract-comments": "^1.0.0",
     "faker": "^4.1.0",
-    "fbjs": "^0.8.17",
```
layershifter, line 81: We use `fbjs` only in our docs, it's not a production dependency

package.json
```diff
@@ -102,13 +101,13 @@
...
     "puppeteer": "^1.11.0",
-    "react": "^16.0.0",
+    "react": "^16.8.2",
     "react-ace": "^5.1.2",
     "react-codesandboxer": "^3.1.1",
     "react-custom-scrollbars": "^4.2.1",
     "react-docgen-typescript": "^1.12.2",
     "react-document-title": "^2.0.3",
-    "react-dom": "^16.0.0",
+    "react-dom": "^16.8.2",
```
layershifter: We will use this version only in our docs, we don't increase requirements there

bmdalex (approved):

package.json
kuzhelov: actually, would rather like to reserve the same React version being used both for Stardust, as well as in docs. This would ensure that if some consumer uses React version that satisfies Stardust requirements (say, React v16.4), then her environment would be enough to run all the examples and prototypes that are provided in docs - otherwise this is just a matter of time when Hooks will be used in an example's or prototype's code, and those won't be compatible with client's environment, in contrast to Stardust itself

docs/src/components/ComponentDoc/ComponentControls/ComponentControls.tsx
```diff
@@ -106,12 +102,24 @@ const ComponentControls: React.FC<ComponentControlsProps> = props => {
           )}/${showRtl}`,
           target: '_blank',
           rel: 'noopener noreferrer',
-          onClick: onMaximize,
+          onClick: () => {},
```
kuzhelov: may we make this optional?

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -648,7 +653,6 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
                     showRtl={showRtl}
```
kuzhelov, line 653: I do understand that this is not this PR's problem, but the following markup 

```jsx
<Flex .. >
  <Flex.Item>   <--- no styles here!!
</Flex>
```

looks strange to me. The intent here is just to align rendered items in column - this could be achieved much simpler, with no <Flex.Item> or React Fragment whatsoever:

```
<Flex column>
  <Segment />
  ...
</Flex>
```

kuzhelov (approved):

docs/src/components/ComponentDoc/ComponentControls/ComponentControls.tsx
layershifter: Actually, it's cruft, remove it 👍

package.json
```diff
@@ -102,12 +101,12 @@
     "normalize.css": "^8.0.0",
     "prettier": "^1.15.3",
     "puppeteer": "^1.11.0",
-    "react": "^16.0.0",
+    "react": "^16.3.0",
```
alinais, line 104: `react 16.3.0` and `react-dom 16.3.0` are already approved

package.json
```diff
@@ -118,6 +117,7 @@
     "rimraf": "^2.6.1",
     "satisfied": "^1.1.1",
     "screener-runner": "^0.10.30",
+    "shallowequal": "^1.1.0",
```
alinais, line 120: Is `shallowequal` used for distribution purpose? Or internal?

layershifter: It's used only in docs, see https://github.com/stardust-ui/react/pull/928#discussion_r258100901

PR #679, opened 2019-01-03
docs: Removing Semantic UI from Stardust- DocSite

hughreeling:
Removes all references to Semantic UI from Stardust.

Updates code to use Stardust controls.

Attempts some basic accessibility fixes.

**Breaking change**
Removes reliance on Semantic UI for Font Awesome icons, instead references them directly.
Code using them would need to include them implicitly.

Updated Dark Theme to include variants for List and Menu

Known issues:

- find a way to isolate the component examples in the docSite to not interfere with the docSite theme (use an iframe for this purpose)

- update documentation on limitation of Provider usage (is there any issue if using more than a certain number of Providers?)

- themes switching dropdown  doesn't show the list items on the first render
image

- reenable searching for components in sidebar
image

- active state of the links in the sidebar menu is missing

- fix the theming switch - the examples headers are changing background but not the text
image

- bring back the right floating menu
image

- address the code optimization suggestions marked as `TODO:` in docs

PR #931, opened 2019-02-19
feat(prototypes): mention scenario with dropdown

bmdalex:
### Description

This PR attempts to create a prototype to show users how they can mention people using `@` key in a rich text editor.

I am using a `contentEditable` div as rich text input and the `Dropdown` component to leverage the list of users and filtering through them. The rest of the functionality is in the prototype section 

### Screenshot
screen recording 2019-02-14 at 16 51 35

### Issues:
- #900 and issue in [description](https://github.com/stardust-ui/react/pull/900#issue-253086393) - **blocker**
- styling regression when using the filtering capability caused by #679 

### Later edit:
all issues and blocking items for this PR are fixed


docs/src/prototypes/dropdowns/AsyncDropdownSearch.tsx
```diff
@@ -34,25 +34,55 @@ const createEntry = (): Entry => ({
...
   }
 
-  handleSearchQueryChange = (e: React.SyntheticEvent, { searchQuery }: DropdownProps) => {
+  private handleSearchQueryChange = (e: React.SyntheticEvent, { searchQuery }: DropdownProps) => {
     this.setState({ searchQuery })
     this.fetchItems()
   }
 
-  fetchItems = () => {
+  private fetchItems = () => {
```
bmdalex: https://github.com/stardust-ui/react/pull/901#discussion_r256900108 from @layershifter :

_What is purpose of these changes? Do we need them?
We don't have an agreement on visibility modifiers so let's revert these changes._

bmdalex: @layershifter 

> We don't have an agreement on visibility modifiers so let's revert these changes.

I don't get this part, it's pretty obvious to me that in Typescript class members that are not meant to be used outside the scope of the class are private.

> What is purpose of these changes? Do we need them?

I will revert the changes however because the prototype section does not have the JS version so you might argue the change does not bring any benefit here.

docs/src/prototypes/dropdowns/utils.ts
```diff
@@ -0,0 +1,36 @@
...
+    return
+  }
+
+  const elem = document.createElement('span')
+  elem.id = id
+  if (text) {
+    elem.innerText = text
+  }
+  range.insertNode(elem)
+}
```
bmdalex, line 51: https://github.com/stardust-ui/react/pull/901#discussion_r256902962 from @layershifter 

Is there any third party dependency that does this work? That is fully tested and supported?
Why we can't use [`caret-pos`](https://www.npmjs.com/package/caret-pos)?

bmdalex: @layershifter Just wanted to keep things minimal and avoid  introducing extra dependencies just for a prototype, clients using this code will end up needing it too.

docs/src/prototypes/dropdowns/inputWithDropdown.tsx
```diff
@@ -0,0 +1,136 @@
...
+      />
+    )
+  }
+
+  private showDropdown = () => {
+    this.dropdownExists = true
+    insertNodeAtCursorPosition({ id: this.mountNodeId })
+
+    const node = this.getMountNode()
+    ReactDOM.render(
```
bmdalex: https://github.com/stardust-ui/react/pull/901#discussion_r256901115 from @layershifter 

Can you please why we can't use `createPortal()`/`PortalInner` component there? What is a reason for another React's root?

bmdalex: no reason, 1st iteration, let me try that, thanks for the suggestion

docs/src/prototypes/dropdowns/AsyncDropdownSearch.tsx
kuzhelov: agree with both sides here. Actually, if we would be absolutely pedantic here and adhere to the flavour of OO paradigm introduced by TS, we indeed should use scope modifiers for components' methods - to showcase which ones meant to be accessible from outside as part of public API.

At the same time, the main objection around having this burden of scope modifiers was reasonable fact that this is pretty rare scenario (that, almost for sure would be invalid in Stardust's case) when we would need to access methods of instantiated React component from outside world. Given this reasonable fact, it was considered to be unnecessary to have this scope modifiers being provided to React components.

That being said, just for the sake of moving forward, lets stick to the thing we've agreed upon (i.e. not introduce scope modifiers for React components) - although, as @Bugaa92 has noticed, this is the prototype's code where this is reasonable approach. And, if there is still strong objection against that (or, for certain, if there is any use case that showcases importance of having these modifiers) - @Bugaa92, please, feel free to schedule some time to discuss that 👍

layershifter: Personally for me: there is no big difference, but I _expect_ to see handlers and helpers method before `render()` methods as it's done in React docs: https://reactjs.org/docs/handling-events.html

docs/src/prototypes/dropdowns/dataMocks.ts
```diff
@@ -0,0 +1,53 @@
+export interface AtMention {
+  header: string
+  image: string
+  content: string
+}
+
+export const atMentionItems: AtMention[] = [
```
layershifter:
```suggestion
export const atMentionItems: AtMentionItem[] = [
```

Type name is not represent the object, `AtMentionItem` should be more clear

docs/src/prototypes/dropdowns/dataMocks.ts
```diff
@@ -0,0 +1,53 @@
...
+  {
+    header: 'Peter Parker',
+    image: 'public/images/avatar/small/daniel.jpg',
+    content: 'Partner Software Engineer',
+  },
+  {
+    header: 'Selina Kyle',
+    image: 'public/images/avatar/small/ade.jpg',
+    content: 'Graphic Designer',
+  },
```
layershifter: Actually, we can use `faker` to create these data. Can we use it?

docs/src/prototypes/dropdowns/index.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+import { PrototypeSection, ComponentPrototype } from '../Protoypes'
```
layershifter:
```suggestion
import { PrototypeSection, ComponentPrototype } from '../Prototypes'
```

Oops, a typo ⌨️

docs/src/prototypes/dropdowns/inputWithDropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+class InputWithDropdownExample extends React.Component<{}, InputWithDropdownState> {
+  private readonly mountNodeId = 'dropdown-mount-node'
+  private readonly dropdownInputSelector = `#${this.mountNodeId} .${Input.slotClassNames.input}`
+  private readonly editorStyle: React.CSSProperties = {
+    backgroundColor: 'lightgrey',
+    padding: '5px',
+    minHeight: '100px',
+    outline: 0,
+  }
```
layershifter: `editorStyle` can be a usual variable

docs/src/prototypes/dropdowns/inputWithDropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+    this.setState({ searchQuery, dropdownValue })
+  }
+
+  private handleSearchQueryChange = (e: React.SyntheticEvent, { searchQuery }: DropdownProps) => {
+    this.setInputElementSize(searchQuery.length)
+    this.setState({ searchQuery })
+  }
+
+  private handleInputKeyDown = (e: React.KeyboardEvent) => {
+    const code = keyboardKey.getCode(e)
```
layershifter:
```suggestion
    const keyCode = keyboardKey.getCode(e)
```

docs/src/prototypes/dropdowns/inputWithDropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+        noResultsMessage="We couldn't find any matches."
+      />,
+      node,
+    )
+  }
+
+  private hideDropdownAndRestoreEditor = () => {
+    removeElement(this.getMountNode())
+
+    this.tryFocusEditor()
```
layershifter:
```ts
_.invoke(this.contendEditableRef.current, 'focus')
```

docs/src/prototypes/dropdowns/inputWithDropdown.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+    return ReactDOM.createPortal(
+      <Dropdown
+        defaultOpen={true}
+        inline
+        search
+        items={atMentionItems}
+        toggleIndicator={null}
+        searchInput={{
+          input: { autoFocus: true },
```
layershifter: Can we avoid `setInputElementSize()` and just pass `size` as prop?

```ts
searchInput={{
  input: {
    autoFocus: true,
    size: dropdownOpen ? 0 : searchQuery.length.
  },
}}

File: docs/src/prototypes/dropdowns/inputWithDropdown.tsx (hunk 3 above)
Comment (bmdalex) [reply docs/src/prototypes/dropdowns/inputWithDropdown.tsx]: implemented using `createPortal`, works as expected, good call

File: docs/src/prototypes/dropdowns/AsyncDropdownSearch.tsx (hunk 1 above)
Comment (bmdalex) [reply docs/src/prototypes/dropdowns/AsyncDropdownSearch.tsx]: let's close for now and stick to what we have, tracking in #933

File: docs/src/prototypes/dropdowns/dataMocks.ts (hunk 4 above)
Comment (bmdalex) [reply docs/src/prototypes/dropdowns/dataMocks.ts]: done, thanks

File: docs/src/prototypes/dropdowns/dataMocks.ts (hunk 5 above)
Comment (bmdalex) [reply docs/src/prototypes/dropdowns/dataMocks.ts]: done, thanks

File: docs/src/prototypes/dropdowns/inputWithDropdown.tsx (hunk 7 above)
Comment (bmdalex) [reply docs/src/prototypes/dropdowns/inputWithDropdown.tsx]: done, thanks

File: docs/src/prototypes/dropdowns/inputWithDropdown.tsx (hunk 8 above)
Comment (bmdalex) [reply docs/src/prototypes/dropdowns/inputWithDropdown.tsx]: done, thanks

File: docs/src/prototypes/dropdowns/inputWithDropdown.tsx (hunk 9 above)
Comment (bmdalex) [reply docs/src/prototypes/dropdowns/inputWithDropdown.tsx]: done, thanks

File: docs/src/prototypes/dropdowns/inputWithDropdown.tsx (hunk 10 above)
Comment (bmdalex) [reply docs/src/prototypes/dropdowns/inputWithDropdown.tsx]: yup, with the portal approach it's possible, good call; when I was rendering another node it didn't work; also, size needs to be min 1 so I did `searchQuery.length + 1`

File: docs/src/prototypes/Prototypes.tsx [hunk 11]
```diff
@@ -0,0 +1,29 @@
...
+
+interface PrototypeSectionProps {
+  title?: string
+}
+
+interface ComponentPrototypeProps extends PrototypeSectionProps {
+  description?: string
+}
+
+export const PrototypeSection: React.SFC<ComponentPrototypeProps> = props => (
```
layershifter:
```suggestion
export const PrototypeSection: React.FC<ComponentPrototypeProps> = props => (
```

SFC was deprecated :)

docs/src/prototypes/Prototypes.tsx
```diff
@@ -0,0 +1,29 @@
...
+  <Box style={{ margin: 20 }}>
+    {props.title && <Header as="h1">{props.title}</Header>}
+    {props.children}
+  </Box>
+)
+
+export const ComponentPrototype: React.SFC<ComponentPrototypeProps> = props => (
```
layershifter:
```suggestion
export const ComponentPrototype: React.FC<ComponentPrototypeProps> = props => (
```

docs/src/prototypes/dropdowns/inputWithDropdown.tsx
```diff
@@ -0,0 +1,109 @@
...
+  padding: '5px',
+  minHeight: '100px',
+  outline: 0,
+}
+
+class InputWithDropdown extends React.Component<{}, InputWithDropdownState> {
+  private readonly initialState: InputWithDropdownState = {
+    dropdownOpen: false,
+    searchQuery: '',
+  }
```
layershifter: NIT: `initialState` can be extracted to a variable as it doesn't have any coupling with the class

docs/src/prototypes/dropdowns/inputWithDropdown.tsx
```diff
@@ -0,0 +1,109 @@
...
+  render() {
+    return (
+      <>
+        <div
+          contentEditable
+          ref={this.contendEditableRef}
+          onKeyUp={this.handleEditorKeyUp}
+          style={editorStyle}
+        />
+        {this.renderDropdown()}
```
layershifter: NIT: Do we really need `renderDropdown()`, it looks quite simple to be inlined

docs/src/prototypes/dropdowns/inputWithDropdown.tsx
```diff
@@ -0,0 +1,109 @@
...
+          onSearchQueryChange={this.handleSearchQueryChange}
+          noResultsMessage="We couldn't find any matches."
+        />
+      </CustomPortal>
+    )
+  }
+
+  private hideDropdownAndRestoreEditor = (cb?: () => void) => {
+    this.setState(this.initialState, () => {
+      this.tryFocusEditor()
```
layershifter:
```suggestion
      _.invoke(this.contendEditableRef.current, 'focus')
```

NIT: `tryFocusEditor()`  used only once, may be we can remove it?

packages/react/src/components/Dropdown/DropdownSearchInput.tsx
```diff
@@ -122,8 +124,8 @@ class DropdownSearchInput extends UIComponent<ReactProps<DropdownSearchInputProp
           onBlur: this.handleInputBlur,
           onKeyDown: this.handleInputKeyDown,
           ...accessibilityInputProps,
+          ...unhandledProps.input,
```
layershifter, line 127: Looks weird, but I understand _why_ it's done.

docs/src/prototypes/dropdowns/inputWithDropdown.tsx
bmdalex: I think this is a matter of personal preference; I prefer keeping it more verbose, it makes it obvious to the reader what the code does without to understand details.

packages/react/src/components/Dropdown/DropdownSearchInput.tsx
bmdalex: yea..no better idea at this point

docs/src/prototypes/dropdowns/inputWithDropdown.tsx
bmdalex: with the latest refactoring..no, we don't need it, inlined it, thanks

docs/src/prototypes/Prototypes.tsx
bmdalex: done, thanks

docs/src/prototypes/Prototypes.tsx
bmdalex: done, thanks

docs/src/prototypes/dropdowns/inputWithDropdown.tsx
bmdalex: agreed on the fact there is no coupling, but the concept of `state` doesn't make sense outside of the component...

with that argument (`it doesn't have any coupling with the class`) most of `readonly` members in TS do not have any coupling with their class (are merely constants) but we add them as members because they make sense just in that class

docs/src/prototypes/dropdowns/AsyncDropdownSearch.tsx
```diff
@@ -68,33 +68,26 @@ class AsyncDropdownSearch extends React.Component<{}, SearchPageState> {
...
-            searchQuery={searchQuery}
-            toggleIndicator={false}
-            value={value}
-          />
-          <Divider />
-          <CodeSnippet mode="json" value={this.state} />
-        </Segment>
-      </div>
+      <>
+        <Dropdown
```
kuzhelov: for this experience might suggest to introduce
- comment to example's code
- maybe better additional `<p>` in between of `Dropdown` and `CodeSnippet`, or add text to `Divider`

that would explain that `CodeSnippet` part rendered is necessary to reflect state of `fetchedData` - otherwise it was a bit confusing to me an, frankly, originally I thought that this is just a docs bug :) In either case, it would be helpful to explicitly say to the client that this is aimed to represent state of _fetched_ data.

image

Another small suggestion might be about reducing amount of data entries fetched - I think that even 5 (with names that would suggest reasonably good distribution) should work fine :) Key benefit is that it won't be necessary for the client to make long scroll to get to the next example (if this data section would be collapsable, it would help as well).

kuzhelov: also, when there is nothing found in the list, there is the following view of Dropdown produced:

image

This representation (with narrow white ribbon below the search field) seems to be a bit confusing - I think that we could easily fix it with `noResultsMessage` prop:

image

docs/src/prototypes/dropdowns/MentionsWithDropdown.tsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+const editorStyle: React.CSSProperties = {
+  backgroundColor: 'lightgrey',
+  padding: '5px',
+  minHeight: '100px',
+  outline: 0,
+}
+
+class MentionsWithDropdown extends React.Component<{}, MentionsWithDropdownState> {
```
kuzhelov, line 24: this is just about visual representation - frankly, it was not absolutely evident to me that this grey box below is introduced as an _editable area_. 

image


I do understand that this was done to keep styles (and amount of code) minimal - and totally support that 💯 But maybe we could improve a bit in this aspect, by doing the following steps:

- change wording of description a bit to explicitly mention that area is editable
- use brighter styles that would differentiate area from regular rectangle

image


Please, feel free to decide on whether changes should be done, and which ones specifically. Let me just, for the sake of reference, share changes that I have tried on my side to achieve this appearance:

```js
const editorStyle: React.CSSProperties = {
  backgroundColor: '#eee',
  padding: '5px',
  minHeight: '100px',
  outline: 0,
  border: '1px dashed grey',
  borderRadius: '5px'
}
```

docs/src/routes.tsx
```diff
@@ -62,9 +62,9 @@ const Router = () => (
           />,
           <DocsLayout
             exact
-            key="/prototype-async-dropdown-search"
-            path="/prototype-async-dropdown-search"
-            component={require('./prototypes/AsyncDropdownSearch/index').default}
+            key="/prototype-dropdowns"
```
kuzhelov, line 65: 👍

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -101,7 +101,10 @@ export interface DropdownProps extends UIComponentProps<DropdownProps, DropdownS
   items?: ShorthandCollection
 
   /**
-   * Function to be passed to create string from selected item, if it's a shorthand object. Used when dropdown also has a search function.
+   * Function that converts an item to string. Used when dropdown has the search boolean prop set to true.
```
kuzhelov, line 104: 👍

docs/src/prototypes/Prototypes.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react'
+import { Box, Header, Segment } from '@stardust-ui/react'
+
+interface PrototypeSectionProps {
+  title?: string
+}
+
+interface ComponentPrototypeProps extends PrototypeSectionProps {
+  description?: string
```
layershifter:
```suggestion
  description?: React.ReactNode
```

docs/src/prototypes/Prototypes.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react'
+import { Box, Header, Segment } from '@stardust-ui/react'
+
+interface PrototypeSectionProps {
+  title?: string
```
layershifter:
```suggestion
  title?: React.ReactNode
```

Please allow to pass elements.

docs/src/prototypes/dropdowns/MentionsWithDropdown.tsx
```diff
@@ -0,0 +1,109 @@
...
+        />
+        <PortalAtCursorPosition open={dropdownOpen}>
+          <Dropdown
+            defaultOpen={true}
+            inline
+            search
+            items={atMentionItems}
+            toggleIndicator={null}
+            searchInput={{
+              input: { autoFocus: true, size: searchQuery.length + 1 },
```
kuzhelov, line 53: seems that `size` prop won't work, as there is no `size` prop declared for `DropdownSearchInput` (or for the `Input`). Am I missing something? I think it would be safe for us just to omit this prop

docs/src/prototypes/dropdowns/MentionsWithDropdown.tsx
```diff
@@ -0,0 +1,109 @@
...
+            onSearchQueryChange={this.handleSearchQueryChange}
+            noResultsMessage="We couldn't find any matches."
+          />
+        </PortalAtCursorPosition>
+      </>
+    )
+  }
+
+  private handleEditorKeyUp = (e: React.KeyboardEvent) => {
+    if (!this.state.dropdownOpen && keyboardKey.getCode(e) === keyboardKey.AtSign) {
```
kuzhelov: this condition will open dropdown even if `2` is typed - I would rather suggest to introduce simplest possible logic of checking for `@` symbol explicitly

docs/src/prototypes/dropdowns/MentionsWithDropdown.tsx
```diff
@@ -0,0 +1,109 @@
...
+    dropdownOpen: false,
+    searchQuery: '',
+  }
+
+  private contendEditableRef = React.createRef<HTMLDivElement>()
+
+  state = this.initialState
+
+  render() {
+    const { dropdownOpen, searchQuery } = this.state
```
kuzhelov: cannot spot where `searchQuery` is used

packages/react/src/components/Dropdown/DropdownSearchInput.tsx
kuzhelov: share the same feeling as @layershifter expressed 😕

kuzhelov (commented):
nice work 👍 just couple of things that I would like to be addressed before merge (apart from minor comments:
- sizing for `input` slot of dropdown input (could we remove this `size` altogether?): https://github.com/stardust-ui/react/pull/931/files#r259296667
- at-mentions dropdown is opened on `2` character typed: https://github.com/stardust-ui/react/pull/931/files#r259297611

docs/src/prototypes/dropdowns/AsyncDropdownSearch.tsx
bmdalex: @kuzhelov `AsyncDropdownSearch.tsx` is not part of this PR;
 this file was just moved so that all dropdown related prototypes are in one place

docs/src/prototypes/Prototypes.tsx
bmdalex: done, thx

docs/src/prototypes/Prototypes.tsx
bmdalex: done, thx

docs/src/prototypes/dropdowns/MentionsWithDropdown.tsx
bmdalex: `size` is an HTML attribute:

_unsigned long: Returns / Sets the element's size attribute, containing visual size of the control. This value is in pixels unless the value of type is text or password, in which case, it is an integer number of characters. Applies only when type is set to text, search, tel, url, email, or password; otherwise it is ignored.
(see: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement)_

We need it to resize the input while typing; useful for mentioning people in the middle of text:
screen recording 2019-02-22 at 14 48 06

docs/src/prototypes/dropdowns/MentionsWithDropdown.tsx
bmdalex: done

docs/src/prototypes/dropdowns/MentionsWithDropdown.tsx
bmdalex: I like it, applied the text change and style on content editable div, thanks!

bmdalex:
thanks @kuzhelov , here are my responses:

> * sizing for `input` slot of dropdown input (could we remove this `size` altogether?): https://github.com/stardust-ui/react/pull/931/files#r259296667

see: https://github.com/stardust-ui/react/pull/931#discussion_r259348495

> * at-mentions dropdown is opened on `2` character typed: https://github.com/stardust-ui/react/pull/931/files#r259297611

fixed

docs/src/prototypes/dropdowns/MentionsWithDropdown.tsx
kuzhelov: got it, thanks for explanation - actually, the fact that we are overlapping with DOM property's name seems a bit disturbing to me. But this is, definitely, a discussion aside from this PR.

kuzhelov (approved):
please, just introduce changelog entry before merging 👍

PR #932, opened 2019-02-19
chore(dangerjs): check dependency changes

miroslavstastny:
As part of CI, package dependencies are verified - if relevant change is detected (new `package.json` added, new dependency added, version of existing dependency changed in any `package.json`) warning is added to the PR.

### Example

image

### DIFF
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -3,29 +3,87 @@ import * as fs from 'fs'
 import * as path from 'path'
 import * as _ from 'lodash'
 
+const CHANGELOG_FILE = 'CHANGELOG.md'
+
+/**
+ * This function asserts that added entries into the changelog file are placed in the right section.
+ */
+const hasAddedLinesAfterVersionInChangelog = async (): Promise<boolean> => {
+  const changelogContent = fs.readFileSync(path.resolve(__dirname, CHANGELOG_FILE)).toString()
+  const versionLineNumber = changelogContent
+    .split('\n')
+    .findIndex(line => line.startsWith('<!--') && line.endsWith('-->'))
+
+  const changelogDiff = await danger.git.structuredDiffForFile(CHANGELOG_FILE)
+  const addedLines = changelogDiff.chunks.reduce((acc, chunk) => {
+    const filteredLines = chunk.changes.filter(change => change.type === 'add')
+
+    return acc.concat(filteredLines)
+  }, [])
+
+  return addedLines.some(line => line.ln >= versionLineNumber)
+}
+
+async function getChangedDependencies(filepath, dependenciesKey = 'dependencies') {
+  const diff = await danger.git.JSONDiffForFile(filepath)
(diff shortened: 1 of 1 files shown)


dangerfile.ts
```diff
@@ -37,3 +40,53 @@ if (!hasChangelog) {
...
+    warn(`New package added: ${filepath}. Make sure you have OSS approval before merging!`)
+  })
+
+danger.git.modified_files
+  .filter(filepath => filepath.match(/\bpackage\.json$/))
+  .forEach(async filepath => {
+    const diff = await danger.git.JSONDiffForFile(filepath)
+    if (diff.dependencies) {
+      const before = { ...diff.dependencies.before, ..._.zipObject(diff.dependencies.added) }
+      const after = diff.dependencies.after || {}
```
layershifter: Actually, we should also check `peerDependencies`, too

dangerfile.ts
```diff
@@ -37,3 +40,53 @@ if (!hasChangelog) {
...
+      const changedDependencies = _.reduce(
+        before,
+        (result, value, key) => {
+          return value === after[key] || !after[key]
+            ? result
+            : { ...result, [key]: { before: value, after: after[key] } }
+        },
+        {},
+      )
```
layershifter: I think that we can extract this to a separate function, i.e. `getChangedDependencies(filepath)`

dangerfile.ts
```diff
@@ -37,3 +40,53 @@ if (!hasChangelog) {
...
+
+danger.git.created_files
+  .filter(filepath => filepath.match(/\bpackage\.json$/))
+  .forEach(filepath => {
+    warn(`New package added: ${filepath}. Make sure you have OSS approval before merging!`)
+  })
+
+danger.git.modified_files
+  .filter(filepath => filepath.match(/\bpackage\.json$/))
+  .forEach(async filepath => {
```
layershifter:
```js
danger.git.modified_files
  .filter(filepath => filepath.match(/\bpackage\.json$/))
  .reduce(async (acc, filepath) => {
    const changedDependencies = await getChangedDependencies(filepath)

    if (Object.keys(changedDependencies).length > 0) {
      acc[filename] = changedDependencies
    }

    return acc
  }, {})
  .reduce((hasWarning, changedDependencies, filename) => {
    // create markdown
    markdown(changesTabled)

    if (hasWarning) return true
    warn('Package dependencies changed. Make sure you have approval before merging!')
  }, false)
```

I expect something like this, less functions and more clear steps
/cc @kuzhelov

alinais (approved):

layershifter (approved):
Only one comment:
`hasAddedLinesAfterVersionInChangelog`, `markdownChangedDependencies`, `checkDependencyChanges` as these functions don't need any stuff from context, can we move them outside the `export default`? So we will have only the run flow in `export default`.

PR #933, opened 2019-02-19
docs(menu): show disabled state for menu items in toolbar/submenu

codepretty:
There were no examples of a disabled menu for iconOnly and icon+text menu. I had requests from design to update the disabled color for icons, but the styling is correct. Adding as an example to the docs for a more complete example.

**Updated menu as a toolbar example**
image


packages/react/src/themes/teams/components/Icon/svg/icons/index.ts
```diff
@@ -118,6 +119,7 @@ export default {
   'participant-remove': participantRemove,
   'phone-clock': phoneClock,
   quote,
+  'read-aloud': readAloud,
```
bmdalex: is `read-aloud` icon exported because it's needed or just for this example?

bmdalex (approved):

docs/src/examples/components/Menu/Usages/MenuExampleToolbar.shorthand.tsx
```diff
@@ -76,14 +77,30 @@ const items = [
...
       items: [
-        { key: '1', content: 'item1' },
+        {
+          key: '1',
+          content: 'item1',
+          icon: {
+            name: 'read-aloud',
+            variables: { outline: true },
+          },
+        },
```
layershifter: Can be just `outline: true` (#1002):
```suggestion
            outline: true,
```

docs/src/examples/components/Menu/Usages/MenuExampleToolbar.shorthand.tsx
```diff
@@ -76,14 +77,30 @@ const items = [
     indicator: false,
     menu: {
       items: [
-        { key: '1', content: 'item1' },
+        {
+          key: '1',
+          content: 'item1',
+          icon: {
+            name: 'read-aloud',
+            variables: { outline: true },
```
layershifter: Can be just `outline: true` (#1002):
```suggestion
            outline: true,
```

layershifter (approved):
If `Icon` is exported, we should have a separate changelog entry

packages/react/src/themes/teams/components/Icon/svg/icons/index.ts
codepretty: ah, true. i'm not sure this feature is supported yet.

docs/src/examples/components/Menu/Variations/MenuExampleVertical.shorthand.tsx
```diff
@@ -7,6 +7,6 @@ const items = [
   { key: 'events', content: 'Upcoming Events' },
 ]
 
-const MenuExampleVertical = () => <Menu defaultActiveIndex={0} items={items} vertical primary />
+const MenuExampleVertical = () => <Menu defaultActiveIndex={0} items={items} vertical />
```
bmdalex, line 10: @codepretty 
### tl;dr

My only concern would be that the same `Menu` variation with the `primary` prop **will not be visually tested by screener anymore**. We can proceed if we're ok with this

### More details:
```jsx
​​​​​​​const MenuExampleVertical = () => <Menu defaultActiveIndex={0} items={items} vertical primary />
```

renders:

screenshot 2019-03-08 at 19 21 36

while

```jsx
const MenuExampleVertical = () => <Menu defaultActiveIndex={0} items={items} vertical />
```

renders

screenshot 2019-03-08 at 19 21 49

The 1st screenshot won't be tested anymore so any regressions in the code for styles for `vertical` & `primary` props combo won't be caught.

However, I acknowledge this is a **limitation of our testing infrastructure** that will hopefully be solved once we have a resolution on #1000

bmdalex (approved):
👍

docs/src/examples/components/Menu/Variations/MenuExampleVertical.shorthand.tsx
codepretty: It would be great to have our screener tests be able to run through all the permutations of props/variations we care about - it looks like that is the concern of #1000 . If we're concerned with testing and finding regressions for Teams theme menus though, I think not using the `primary` prop in our testing scenario is fine as the gray menu is what we care about.

bmdalex: > all the permutations of props/variations we care about

A nice thing to do would be to gather all these permutations for all components and place them under each theme. We can have some sort of big test config object with the following structure:

```typescript
type ThemeName = 'base' | 'teams' | 'teamsDark' | 'teamsHighContrast'
type ComponentName = 'List' | 'Menu' | 'Portal' /* ...rest of components */
type ObjectOfAny = Record<string, any>

type ComponentTestObjEntry = string | ObjectOfAny
type ComponentTestObj = { [key in ThemeName]?: ComponentTestObjEntry[] }
type TestObject = { [key in ComponentName]?: ComponentTestObj }

export const testObj: TestObject = {
  List: { /* list props to test */ },
  Menu: {
    teams: [
      'primary',
      'vertical',
      'pointing',
      { pointing: 'start' },
      'primary vertical',
      { pointing: 'start', primary: true },
      { pointing: 'start', primary: true, vertical: true },
      // ...
    ],
    teamsDark: [
      'primary',
      'vertical',
      'pointing',
      // ...
    ],
    teamsHighContrast: [ /* Menu HC props to test */ ]
  },
  Portal: { /* portal props to test */ },
}
```
I'm going to propose this idea in #1000 and then maybe you can help us gather these permutations, @codepretty

PR #937, opened 2019-02-20
fix: temporary revert of unverified dependencies

miroslavstastny:
Reverts latest upgrade of `fela` and removes `mdn-polyfills`.
This libraries will be added again with the new approval process.


docs/src/index.ejs
```diff
@@ -26,7 +26,6 @@
   </script>
 </head>
 <body>
-  <script src="https://cdn.jsdelivr.net/npm/@babel/polyfill@7/dist/polyfill.min.js"></script>
```
layershifter, line 29: Not sure that we should revert this as it affects only our doc site.

layershifter (approved):

docs/src/index.ejs
kuzhelov: yes, agree with the point - but would rather see this being an atomic change: it would be easier to explain this PR as "reverting these two commits", rather than say that "it is partial revert, here are the things that were preserved". Would vote for forfeiting absolute correctness in favour of management simplicity here

kuzhelov (approved):
lets just introduce changelog entry (also, given that Fela upgrade change was introduced as a breaking one, we might want to consider to do the same thing with temp downgrade story)

PR #939, opened 2019-02-20
fix(Accordian Indicator): Updating themeing

bcalvery:
Latest changes:
image

Goal to match Teams control messages:
image

Issues:
Keyboard navigation
Content Title content isn't actionable
Need to support an empty state (and hide Indicator when empty)


packages/react/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -55,24 +59,31 @@ class AccordionTitle extends UIComponent<ReactProps<AccordionTitleProps>, any> {
...
     const contentElement = (
       <Layout
-        start={Indicator.create(indicatorWithDefaults, {
-          defaultProps: {
-            direction: active ? 'bottom' : 'end',
-            styles: styles.indicator,
-          },
-        })}
+        start={
+          <Layout
```
bmdalex, line 77: please do not use `Layout` component as it's going to be deprecated and it renders a lot of extra DOM nodes; take a look at `Flex` component as it might solve your scenario

docs/src/examples/components/Accordion/Usage/AccordionPanelCustomTitleExample.shorthand.tsx
```diff
@@ -19,6 +19,29 @@ class AccordionPanelCustomTitleExample extends React.Component {
...
+              <strong>Someone</strong> has added <strong>Someone Else</strong> to the team
+            </div>
+          ),
+          icon: { name: 'participant-add', variables: { outline: true } },
+        },
+        content: {
+          key: 'peopleadded',
+          content: (
+            <div>
+              <div style={{ padding: '0 0 .5em 0' }}>
```
bmdalex, line 35: please do not change example styles to match specific theme;

I'd suggest reverting these changes in this file and create a prototype (use #931 as a reference on how to create one)

packages/react/src/components/Accordion/AccordionContent.tsx
```diff
@@ -20,6 +20,9 @@ export interface AccordionContentProps
   /** Whether or not the content is visible. */
   active?: boolean
 
+  /** Whether or not content should be indented */
+  indented?: boolean
```
bmdalex, line 24: Can you please:
- explain why this prop is relevant in describing the content of an accordion? 
- describe the scenario where this prop is needed?

We need a strong argument with examples (preferably from the larger web community) that a prop is describing a component before we can accept it as a valid prop

@kuzhelov 
@layershifter 
@miroslavstastny 
@mnajdova 
@alinais

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -157,6 +157,7 @@ class Accordion extends AutoControlledComponent<ReactProps<AccordionProps>, any>
     _.each(panels, (panel, index) => {
       const { content, title } = panel
       const active = this.isIndexActive(index)
+      const indented = title.hasOwnProperty('icon')
```
bmdalex, line 160: this is a very specific detail and does not look like functionality that is generally applicable for any Accordion; can you achieve this on the theme side (`packages/react/src/themes/teams/components/Accordion/accordionStyles.ts`)

packages/react/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -37,6 +38,9 @@ export interface AccordionTitleProps
 
   /** Shorthand for the active indicator. */
   indicator?: ShorthandValue
+
+  /** Shorthand for the icon. */
+  icon?: ShorthandValue
```
bmdalex, line 43: @kuzhelov
@layershifter
@miroslavstastny
@mnajdova
@alinais

do we have any agreement on the `icon` prop for `Accordion`?

bmdalex (changes requested):
please take a look at comments

packages/react/src/components/Accordion/AccordionTitle.tsx
codepretty: What about creating a new example under Usages? If it's an example, then screenshot tests will capture any regressions - I don't think prototypes have that same benefit

docs/src/examples/components/Accordion/Usage/AccordionPanelCustomTitleExample.shorthand.tsx
codepretty: What about creating a new example under Usages? If it's an example, then screenshot tests will capture any regressions - I don't think prototypes have that same benefit

packages/react/src/components/Accordion/AccordionTitle.tsx
layershifter: We should not introduce a new slot there, `Indicator` already covers it:

image
https://codesandbox.io/s/n725p7rz54?module=/example.js

docs/src/examples/components/Accordion/Usage/AccordionPanelCustomTitleExample.shorthand.tsx
bmdalex: > What about creating a new example under Usages?

I also don't think it's a good practice.

The point of the Examples section is to provide **simple and generic code snippets** to show clients how to **use** our components. It is not intended to show them a very particular set of customizations to achieve a very particular flavor for that component.

> If it's an example, then screenshot tests will capture any regressions - I don't think prototypes have that same benefit

They don't but these screenshots tests are not meant to capture regressions for styles inserted in the examples themselves, but for the whole theme (so for styles that are introduced in the top level theme object for the theme that is currently displayed in the docs site). Hope this makes sense.

packages/react/src/components/Accordion/AccordionContent.tsx
bcalvery: So I definitely don't like the feeling of having this prop. It is an artifact of the way Accordion renders the Title and Content of each panel separately within the Accordion. The purpose of Indented is to react to the presence of the Icon (that is a question in itself) and indent the accordion content accordingly.
With and without Icon:
image

packages/react/src/components/Accordion/AccordionTitle.tsx
bcalvery: The Icon is intended to be separate form the indicator. But maybe Adding Icon in general is not right. Maybe making Accordion usable for the control messages was the wrong direction?

PR #941, opened 2019-02-21
feat(perf): add 'filter' flag to perf tests

kuzhelov:
This PR introduces filtering support to perf tests - this functionality is useful in scenarios of perf issues investigations.

Introduced `filter` flag filters perf component examples based on name prefix (case-agnostic).

### Usage
Run perf tests for `Header` component only:
```
yarn perf --filter Header
```


perf/types.tsx
```diff
@@ -1,6 +1,6 @@
 declare global {
   interface Window {
-    runMeasures: () => Promise<ProfilerMeasureCycle>
+    runMeasures: (filter?: string) => Promise<ProfilerMeasureCycle>
```
kuzhelov, line 3: later we might think about `config` arg as an object, but for now has selected the simplest approach

layershifter (commented):
Can we use `nanomatch`/`**match` use allow use globs?

```js
import nm from 'nanomatch'

const performanceExampleNames: string[] = _.shuffle(
  nm.match(performanceExamplesContext.keys(), filter)
)

kuzhelov:
@layershifter, actually thought about it - but do we really need globs? I mean, if we would consider those cases we are interested in, it is mostly about filtering on 'by component' basis, or 'by example' basis - and both of these scenarios are achievable with current simplest implementation. I would rather stay with the simplest necessary implementation for now

layershifter (approved):

miroslavstastny (approved):

PR #943, opened 2019-02-21
feat(FocusZone): Export FocusZone types

sophieH29:
As per request, export FocusZone types, so that users can refer to them if they want to override the focus zone in accessibility behaviors.


mnajdova (approved):
Be sure to update the changelog before merging

PR #945, opened 2019-02-22
fix(styles): removed redundant margins from components and examples

bmdalex:
# fix(styles): removed redundant margins from components and examples

## Description

This PR is pretty straightforward. It:
- removes the `marginRight` styles from the following components:
  - Button
  - Icon
  - Dialog (as consequence, as it was using 2 `Button` components next to each other in the implementation)
- fixes all the examples grouping these components by using `Flex` to achieve the necessary spacing previously provided by `marginRight`

## Screenshots: 

These screenshot outline how spacing is achieved in `Button` examples where 2 `Button` components are placed next to each other

### Before (with `marginRight`):
screenshot 2019-02-22 at 10 25 31

### After (with `gap="gap.small"`):
screenshot 2019-02-22 at 10 24 47

### DIFF
--- a/docs/src/examples/components/Animation/Types/AnimationExampleKeyframeParams.tsx
+++ b/docs/src/examples/components/Animation/Types/AnimationExampleKeyframeParams.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react'
-import { Animation, Icon, Provider, ThemeAnimation } from '@stardust-ui/react'
+import { Animation, Flex, Icon, Provider, ThemeAnimation } from '@stardust-ui/react'
 
 const colorChanger: ThemeAnimation<{ fromColor: string; toColor: string }> = {
   keyframe: ({ fromColor, toColor }) => ({
@@ -17,7 +17,7 @@ const colorChanger: ThemeAnimation<{ fromColor: string; toColor: string }> = {
 
 const AnimationExample = () => (
   <Provider theme={{ animations: { colorChanger } }}>
-    <div>
+    <Flex gap="gap.smaller">
       <Animation name="colorChanger">
         <Icon name="umbrella" circular bordered />
       </Animation>
@@ -27,7 +27,7 @@ const AnimationExample = () => (
       <Animation name="colorChanger" keyframeParams={{ toColor: 'black' }}>
         <Icon name="umbrella" circular bordered />
       </Animation>
-    </div>
+    </Flex>
   </Provider>
 )
 
--- a/packages/react/src/components/Button/ButtonGroup.tsx
+++ b/packages/react/src/components/Button/ButtonGroup.tsx
@@ -16,6 +16,7 @@ import {
 import { Accessibility } from '../../lib/accessibility/types'
 import { defaultBehavior } from '../../lib/accessibility'
 import Button from './Button'
+import Flex from '../Flex/Flex'
 
 export interface ButtonGroupProps
   extends UIComponentProps,
@@ -76,14 +77,16 @@ class ButtonGroup extends UIComponent<ReactProps<ButtonGroupProps>, any> {
 
     return (
       <ElementType {...unhandledProps} className={classes.root}>
-        {_.map(buttons, (button, idx) =>
-          Button.create(button, {
-            defaultProps: {
-              circular,
-              styles: this.getStyleForButtonIndex(styles, idx === 0, idx === buttons.length - 1),
-            },
-          }),
-        )}
+        <Flex gap="gap.smaller">
+          {_.map(buttons, (button, idx) =>
+            Button.create(button, {
+              defaultProps: {
+                circular,
+                styles: this.getStyleForButtonIndex(styles, idx === 0, idx === buttons.length - 1),
+              },
+            }),
+          )}
+        </Flex>
       </ElementType>
     )
   }
--- a/packages/react/src/components/Button/ButtonGroup.tsx
+++ b/packages/react/src/components/Button/ButtonGroup.tsx
@@ -76,18 +76,16 @@ class ButtonGroup extends UIComponent<ReactProps<ButtonGroupProps>, any> {
     }
 
     return (
-      <ElementType {...unhandledProps} className={classes.root}>
-        <Flex gap="gap.smaller">
-          {_.map(buttons, (button, idx) =>
-            Button.create(button, {
-              defaultProps: {
-                circular,
-                styles: this.getStyleForButtonIndex(styles, idx === 0, idx === buttons.length - 1),
-              },
-            }),
-          )}
-        </Flex>
-      </ElementType>
+      <Flex as={ElementType} gap="gap.smaller" {...unhandledProps} className={classes.root}>
+        {_.map(buttons, (button, idx) =>
+          Button.create(button, {
+            defaultProps: {
+              circular,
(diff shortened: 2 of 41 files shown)


packages/react/src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -101,7 +101,6 @@ const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
     return {
       backgroundColor: v.backgroundColor,
       display: 'inline-block',
-      marginRight: v.marginRight,
```
bmdalex, line 104: here is where we remove redundant margin for `Button`

packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -81,7 +81,7 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps & ButtonState, any> = {
       alignItems: 'center',
       position: 'relative',
       padding: `0 ${pxToRem(paddingLeftRightValue)}`,
-      margin: `0 ${pxToRem(8)} 0 0`,
+      margin: 0,
```
bmdalex: here is where we remove redundant margin for `Icon`

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -211,15 +212,15 @@ class Dialog extends AutoControlledComponent<ReactProps<DialogProps>, DialogStat
             },
             overrideProps: {
               content: (
-                <>
+                <Flex gap="gap.small" hAlign="end">
```
bmdalex: here we use `Flex` to fix side effect from removing margin from `Button`

packages/react/src/themes/teams/components/Input/inputVariables.ts
```diff
@@ -26,7 +26,7 @@ export default (siteVars): InputVariables => ({
   fontSize: siteVars.fontSizes.medium,
 
   iconPosition: 'absolute',
-  iconRight: pxToRem(2),
+  iconRight: pxToRem(10),
```
bmdalex, line 29: removed the margin of 8px from `Icon` meaning we need to add the difference here

packages/react/src/themes/teams/components/Attachment/attachmentVariables.ts
```diff
@@ -21,8 +23,8 @@ export type AttachmentVariables = {
 }
 
 export default siteVariables => ({
-  padding: 8,
-  iconSpace: 12,
+  padding: pxToRem(8),
+  iconSpace: pxToRem(12),
```
bmdalex, line 27: there we're not working previously 😕

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -211,15 +212,15 @@ class Dialog extends AutoControlledComponent<ReactProps<DialogProps>, DialogStat
             },
             overrideProps: {
               content: (
-                <>
+                <Flex gap="gap.smaller" hAlign="end">
```
bmdalex, line 215: @kuzhelov this means introducing an extra DOM node, any suggestions on how to avoid this?
I can only think about adding a `marginRight` variable and override it for `Button` when needed

also @layershifter as code owner of this component

packages/react/src/components/Button/ButtonGroup.tsx
```diff
@@ -76,14 +77,16 @@ class ButtonGroup extends UIComponent<ReactProps<ButtonGroupProps>, any> {
...
       <ElementType {...unhandledProps} className={classes.root}>
-        {_.map(buttons, (button, idx) =>
-          Button.create(button, {
-            defaultProps: {
-              circular,
-              styles: this.getStyleForButtonIndex(styles, idx === 0, idx === buttons.length - 1),
-            },
-          }),
-        )}
+        <Flex gap="gap.smaller">
```
bmdalex: @kuzhelov this means introducing an extra DOM node, any suggestions on how to avoid this?
I can think about:
- adding a `marginRight` variable and override it for `Button` when needed
- or specifying
```typescript
defaultProps: {
  as: Flex,
}

// ...

renderComponent() {
  // ...
  return <ElementType gap="gap.smaller" {...unhandledProps} className={classes.root} ... />
}
```

kuzhelov: yes, I would expect the second approach being used here as, essentially, this is what we would like to achieve - i.e. to use `Flex` as a container.

However, this will introduce one subtle problem (that, depending on the use case, might become not that subtle to spot:). Suppose that client wants to achieve the following
- use <span> as wrapper (this is the place where example seems to be contrived, but I am pretty sure that there might be more appealing scenario)
- preserve layout aspects

Note that this use case won't be covered even with the second approach suggested - however, we might use its core idea being a bit extended. To cover mentioned use case I would suggest to use the following implementation strategy:

```jsx
renderComponent({ ElementType, classes, unhandledProps, ... }) {
  ...
  return <Flex as={ElementType} gap="gap.smaller" {...unhandledProps} className={classes.root} ... />
}
```

packages/react/src/components/Dialog/Dialog.tsx
kuzhelov: I think that here it is absolutely fine to introduce additional DOM node - as, if you would imagine what one would do to achieve the same effect with plain HTML, it would still be about introducing container DOM node

packages/react/src/themes/teams/components/Button/buttonStyles.ts
kuzhelov: could we omit this line altogether?

kuzhelov (commented):
just couple of small changes I would suggest to introduce
- for ButtonGroup consider to use variant of alternative approach you've suggested: https://github.com/stardust-ui/react/pull/945/files#r259471117
- remove margins from `Button` styles altogether: https://github.com/stardust-ui/react/pull/945/files#r259272014

packages/react/src/components/Button/ButtonGroup.tsx
bmdalex: sounds good, applied `Flex as={ElementType} gap="gap.smaller"`, thanks

packages/react/src/components/Dialog/Dialog.tsx
bmdalex: cool, thanks for looking into it

packages/react/src/themes/teams/components/Button/buttonStyles.ts
bmdalex: done, thanks

docs/src/examples/components/Animation/Types/AnimationExampleKeyframeParams.tsx
```diff
@@ -17,7 +17,7 @@ const colorChanger: ThemeAnimation<{ fromColor: string; toColor: string }> = {
 
 const AnimationExample = () => (
   <Provider theme={{ animations: { colorChanger } }}>
-    <div>
+    <Flex gap="gap.smaller">
```
bmdalex, line 20: fyi @mnajdova

kuzhelov (approved):

PR #947, opened 2019-02-22
feat(Prototypes): add MenuButton

### LINKED ISSUE #877: Allow creating Button with attached Menu
## Feature Request

Allow creating Button with attached Menu. The functionality should follow [ARIA Navigation Menu example](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html) with regards to the rendered DOM structure and keyboard handling. The only difference is that the menu should not be opened on hover, but on click.

### Requirements
 - menu open on enter/space and focus goes to first item in menu
-(Optional) Down Arrow: opens the menu and moves focus to the first menu item.
-(Optional) Up Arrow: opens the menu and moves focus to the last menu item.
- menu is closed on esc key and focus goes back to the button 
- if he press TAB key, focus goes naturally to next focusable element after menu and menu is closed
- if he press shift+tab key focus goes naturally to previous focusable element and menu is closed.
- able to do position of menu the same as for popup

### DOM structure

```
<div class="menu_button">
  <button id="menubutton"
          aria-haspopup="true"
          aria-controls="menu2">
    WAI-ARIA Quick Links
  </button>
   <!-- ul/li/a structure rendered by Menu component -->
  <ul id="menu2"
      role="menu"
      aria-labelledby="menubutton">
    <li role="none">
      <a role="menuitem" href="https://www.w3.org/">
        W3C Home Page
      </a>
    </li>
```

layershifter:
Fixes #877.


kolaps33:
nice work :) 👍 
tested from accessibility point of view, it works as expected... Keyboard behavior as well reader experience.

kolaps33 (approved):

docs/src/prototypes/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,198 @@
...
+    const { accessibility, button, disabled, menu, placement } = this.props
+    const { menuOpen } = this.state
+    const accessibilityBehavior: AccessibilityBehavior = accessibility({
+      ...this.props,
+      ...this.state,
+    })
+
+    return (
+      <MenuButtonWrapper onKeyDown={this.handleKeyDown}>
+        <Manager>
```
kuzhelov: nit: lets import it as `PopperManager`, to make things a bit easier to reason about, same might say about `Reference` -> `PopperReference`

docs/src/prototypes/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,198 @@
...
+          <Reference>
+            {({ ref }) => (
+              <Ref
+                innerRef={(buttonNode: HTMLButtonElement) => {
+                  // @ts-ignore
+                  this.buttonRef.current = buttonNode
```
kuzhelov: actually, quite concerned about us doing that. If some day this will be really implemented as read-only prop (at JS level), we would have real problems with that. I would rather expect us following the interface that is suggested by React types

docs/src/prototypes/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,198 @@
...
+        menuOpen: !prevState.menuOpen,
+      }))
+    },
+  })
+
+  handleDocumentClick = (e: MouseEvent) => {
+    const { menuOpen } = this.state
+    const target = e.target as HTMLElement
+    const isInside =
+      _.invoke(this.buttonRef.current, 'contains', target) &&
```
kuzhelov: this should be `or` condition - actually, now it will never evaluate to `true`

docs/src/prototypes/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,198 @@
...
+  render() {
+    const { accessibility, button, disabled, menu, placement } = this.props
+    const { menuOpen } = this.state
+    const accessibilityBehavior: AccessibilityBehavior = accessibility({
+      ...this.props,
+      ...this.state,
+    })
+
+    return (
+      <MenuButtonWrapper onKeyDown={this.handleKeyDown}>
```
kuzhelov: just as a matter of thought - should we, instead, remove this `MenuButtonWrapper` component and make `MenuButton` to be the one that is created as Stardust component? I mean, I do understand that the fact `state` is used in `MenuButton`, potentially, had forced towards using wrapper as Stardust component, but now 
- we are defining accessibility as a prop of `MenuButton`  component
- and instead of letting Stardust to process the accessibility, we are using this internal logic here

I would expect things top remain encapsulated - so that, in case if there will be any update to the way how `accessibility` is processed, this update will reach `MenuButton` component as well.

What do you think?

docs/src/prototypes/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,198 @@
...
+        menuOpen: false,
+      })
+    }
+
+    if (shouldOpen) {
+      this.setState({ lastKeyCode: keyCode, menuOpen: true })
+    }
+  }
+
+  handleMenuItemOverrides = accessibilityBehavior =>
```
kuzhelov: lets accept `menuItemAccessibilityBehavior` as an argument of this method: this will decouple it from the parent accessibility object of `menuButton`, as well as will make it easier to infer the intent at the place of usage:

```js
Menu.create(menu, {
   defaultProps: {
      ...accessibilityBehavior.attributes.menu,
     overrideProps: {
       items: this.handleMenuItemOverrides(accessibilityBehavior.menuItem),
       ...
```

docs/src/prototypes/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,198 @@
...
+import * as PopperJS from 'popper.js'
+import * as React from 'react'
+import { Manager, Reference, Popper } from 'react-popper'
+
+import { focusButton, focusMenuItem, focusNearest } from './focusUtils'
+import menuButtonBehavior from './menuButtonBehavior'
+import MenuButtonWrapper from './MenuButtonWrapper'
+
+export interface MenuButtonProps {
+  accessibility?: Accessibility
```
kuzhelov: maybe we should omit this prop from `MenuButton`'s API - because, essentially, we don't expect this behavior to be overriden, given that `MenuButton` component (which is not even a Stardust component) is introduced just for the sake of matching the exact behavior described here: https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html .

docs/src/prototypes/MenuButton/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const { placement } = this.state
+
+    return (
+      <PrototypeSection title="MenuButton">
+        <ComponentPrototype
+          title="A sample MenuButton"
+          description={
+            <>
+              A sample MenuButton, follows{' '}
+              <a href="https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html">
```
kuzhelov, line 29: 👍

docs/src/prototypes/MenuButton/MenuButton.tsx
kuzhelov: a bit concerned that, in contrast to any other Stardust component, we are making this one to be not extendable by any user-defined attributes. Was it done intentionally?

docs/src/prototypes/MenuButton/MenuButton.tsx
layershifter: Makes a sense to me, will be more obvious, done 👍

docs/src/prototypes/MenuButton/MenuButton.tsx
layershifter: Replaced `buttonRef` with `buttonNode` and `menuRef` with `menuNode`, agree with you there :)

---

Going to introduce some ref utils to solve these situations:
```jsx
<Ref innerRef={mergeRef(this.buttonRef, ref)}>
//
const buttonRef = toRef(buttonNode)
```

docs/src/prototypes/MenuButton/MenuButton.tsx
layershifter: Great catch 👍

docs/src/prototypes/MenuButton/MenuButton.tsx
layershifter: Reworked this part, now we're passing: `(menuItemAccessibilityAttributes: AccessibilityAttributes) => {}`

docs/src/prototypes/MenuButton/MenuButton.tsx
layershifter: Replaced with:
```jsx
<div style={{ boxSizing: 'border-box',  display: 'inline-block' }} />
```

docs/src/prototypes/MenuButton/MenuButton.tsx
layershifter: Agree, removed 👍

docs/src/prototypes/MenuButton/MenuButton.tsx
layershifter: Yep, I removed `MenuButtonWrapper` as it doesn't make any reasonable value, but I don't see any way to make `MenuButton` as Stardust component except the HOC.

packages/react/src/index.ts
```diff
@@ -201,11 +201,23 @@ export {
...
-import { getFirstTabbable, getLastTabbable } from './lib/accessibility/FocusZone/focusUtilities'
-export const FocusZoneUtilities = { getFirstTabbable, getLastTabbable }
+import {
+  getFirstTabbable,
+  getLastTabbable,
+  getNextElement,
+  getPreviousElement,
+  focusAsync,
+} from './lib/accessibility/FocusZone/focusUtilities'
+export const focusZoneUtilities = {
```
kuzhelov: actually, this is a breaking change :( My full support to introduce it, but either
- by changing changelog entry accordingly
- by means of dedicated PR

docs/src/prototypes/MenuButton/focusUtils.ts
```diff
@@ -0,0 +1,16 @@
+import { focusZoneUtilities, Menu } from '@stardust-ui/react'
+
+export const focusMenuItem = (menuRef: HTMLUListElement, order: 'first' | 'last') => {
+  const selector = `.${Menu.Item.slotClassNames.wrapper}:${order}-child .${Menu.Item.className}`
```
kuzhelov, line 4: yes, this is something that might strike a user, as `.${Menu.Item.className}:${order}-child` is the most intuitive thing to try

kuzhelov (approved):

PR #951, opened 2019-02-22
fix(docs): fix conflict Fela and FontAwesome

### LINKED ISSUE #950: [Docs] font-family gets randomized in docSite examples
### Steps
After the refactoring of the docSite to be based on Stardust UI, some of our examples are randomly showing Font Awesome 5 Free font instead of default Stardust's Segoe. 

### Expected Result
The same font-family is shown in all the examples.

### Actual Result
image
In some examples the FontAwesome 5 Free appears to be set as default.

### Version
master

### Testcase

layershifter:
Fixes #950.

### DIFF
--- a/packages/react/src/lib/felaRenderer.tsx
+++ b/packages/react/src/lib/felaRenderer.tsx
@@ -7,6 +7,13 @@ import rtl from 'fela-plugin-rtl'
 
 import { Renderer } from '../themes/types'
 
+// Blacklist contains a list of classNames that are used by FontAwesome
+// https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use
+const blacklistedClassNames = ['fa', 'fas', 'far', 'fal', 'fab']
+
+const filterClassName = (className: string): boolean =>
+  blacklistedClassNames.indexOf(className) === -1
+
 const createRendererConfig = (options: any = {}) => ({
   devMode: process.env.NODE_ENV !== 'production',
   plugins: [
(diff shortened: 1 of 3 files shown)


packages/react/src/lib/felaRenderer.tsx
```diff
@@ -24,6 +31,7 @@ const createRendererConfig = (options: any = {}) => ({
     felaPluginFallbackValue(),
     ...(options.isRtl ? [rtl()] : []),
   ],
+  filterClassName,
```
layershifter, line 34: https://github.com/rofrischmann/fela/blob/v10.2.1/packages/fela/src/generateClassName.js#L25

miroslavstastny (approved):
Funny :-)

alinais (approved):

packages/react/src/lib/felaRenderer.tsx
```diff
@@ -7,6 +7,13 @@ import rtl from 'fela-plugin-rtl'
 
 import { Renderer } from '../themes/types'
 
+// Blacklist contains a list of classNames that are used by FontAwesome
+// https://fontawesome.com/how-to-use/on-the-web/referencing-icons/basic-use
+const blacklistedClassNames = ['fa', 'fas', 'far', 'fal', 'fab']
```
bmdalex, line 12: a bit concerned about these as they can change at any time in fontawesome and we have no control over that..

packages/react/src/lib/felaRenderer.tsx
robinweser: Just as I see it here: you can now use the rtl plugin with a theme.direction property to trigger rtl transformation (check the plugin docs) that way you dont have to conditionally add the plugin

layershifter: Yep, seen these changes. Going to visit this place in a separate PR 👍

packages/react/src/lib/felaRenderer.tsx
kuzhelov: agree - especially given that each theme might load CSS libraries with different set of class names that should be blacklisted. Would expect that this `blacklisted` information will be defined by theme, not (only) general Stardust logic - and, thus, would be extendable

PR #952, opened 2019-02-22
fix(docs): Fela and fontAwesome classes conflict

### LINKED ISSUE #950: [Docs] font-family gets randomized in docSite examples

bmdalex:
## fix(docs): Fela and fontAwesome classes conflict

Fixes #950 

Fix is proposed by @rofrischmann (creator of Fela) [here](https://github.com/rofrischmann/fela/issues/192#issuecomment-276111308)

### Downside:

The prefix is added to all class names:
screenshot 2019-02-22 at 15 47 28

### EDIT:
Alternative to #951

### DIFF
--- a/packages/react/src/lib/felaRenderer.tsx
+++ b/packages/react/src/lib/felaRenderer.tsx
@@ -9,6 +9,7 @@ import { Renderer } from '../themes/types'
 
 const createRendererConfig = (options: any = {}) => ({
   devMode: process.env.NODE_ENV !== 'production',
+  selectorPrefix: '_', // https://github.com/rofrischmann/fela/issues/192
   plugins: [
     // is necessary to prevent accidental style typos
     // from breaking ALL the styles on the page


miroslavstastny:
What about #951?

bmdalex:
@miroslavstastny haha, let's vote? 😆

packages/react/src/lib/felaRenderer.tsx
```diff
@@ -9,6 +9,7 @@ import { Renderer } from '../themes/types'
 
 const createRendererConfig = (options: any = {}) => ({
   devMode: process.env.NODE_ENV !== 'production',
+  selectorPrefix: '_', // https://github.com/rofrischmann/fela/issues/192
```
miroslavstastny, line 12: Move to the same place where we set it for RTL (line 29)

bmdalex: @miroslavstastny mm.. `selectorPrefix` is not overridden by RTL, or are you suggesting that just so we have it consistent and in one place?

bmdalex:
closing in favor of #951

packages/react/src/lib/felaRenderer.tsx
bmdalex: nvm we're going with #951

PR #953, opened 2019-02-22
fix(ChatMessage): prevent hover over invisible actions area

kuzhelov:
## TODO
- [x] update changelog

---------

This PR fixes undesired effect when actions menu is showed up while user hovering over invisible to her area. This has resulted in weird user experience (consider on-hover appearance of actions menu for the lower chat message):

chat-message-problem


jurokapsiar:
focus and hover styles (when you hover on the right edge of the message) are broken after the change:

image

kuzhelov:
@jurokapsiar, thanks a lot for spotting that - have updated styles to address these issues:

chat-message-problem

## submenu's accessibility

chat-message-problem

jurokapsiar (approved):

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -58,10 +59,14 @@ const chatMessageStyles: ComponentSlotStylesInput<
...
     boxShadow: v.actionMenuBoxShadow,
-    opacity: p.focused ? 1 : 0,
     position: 'absolute',
     right: v.actionMenuPositionRight,
     top: v.actionMenuPositionTop,
+    overflow: p.focused ? 'visible' : 'hidden',
+
+    // squash actions menu to prevent accidental hovers over its invisible area
+    width: p.focused ? 'auto' : 0,
+    opacity: p.focused ? 1 : 0,
```
layershifter, line 76: nit: reorder `opacity` and `width` by alpha order 🐱

layershifter (approved):
Don't forget about a changelog entry 🐱

PR #954, opened 2019-02-22
fix(docs): createComponent typings and documentation

### LINKED ISSUE #946: Wrong documentation for `createComponent`
### Steps
- user wants to create a new component;
- user goes to documentation: https://stardust-ui.github.io/react/integrate-custom-components
- users picks first example:
screenshot 2019-02-22 at 12 26 28
- user pastes example in his TS file

### Expected Result
Example works as expected without errors

### Actual Result
We have compilation error on our own example from documentation 💣
screenshot 2019-02-22 at 12 25 30
screenshot 2019-02-22 at 12 25 44

### Version
v0.21.1

bmdalex:
### fix(docs): createComponent typings and documentation

FIxes #946 by:
- improving typings for `createComponent`
- improving `/integrate-custom-components` section in Documentation

### DIFF
--- a/packages/react/src/lib/createComponent.tsx
+++ b/packages/react/src/lib/createComponent.tsx
@@ -13,7 +13,7 @@ export interface CreateComponentConfig<P> {
   shorthandPropName?: string
   defaultProps?: Partial<P>
   handledProps?: string[]
-  propTypes?: React.WeakValidationMap<P>
+  propTypes?: React.ValidationMap<P>
   actionHandlers?: AccessibilityActionHandlers
   focusZoneRef?: (focusZone: FocusZone) => void
   render: (config: RenderResultConfig<P>, props: P) => React.ReactNode
--- a/packages/react/src/lib/createComponent.tsx
+++ b/packages/react/src/lib/createComponent.tsx
@@ -5,6 +5,7 @@ import renderComponent, { RenderResultConfig } from './renderComponent'
 import { AccessibilityActionHandlers } from './accessibility/types'
 import { FocusZone } from './accessibility/FocusZone'
 import { createShorthandFactory } from './factories'
+import { ObjectOf } from '../types'
 
 export interface CreateComponentConfig<P> {
   displayName: string
@@ -12,7 +13,7 @@ export interface CreateComponentConfig<P> {
   shorthandPropName?: string
   defaultProps?: Partial<P>
   handledProps?: string[]
-  propTypes?: React.ValidationMap<P>
+  propTypes?: React.WeakValidationMap<P>
   actionHandlers?: AccessibilityActionHandlers
   focusZoneRef?: (focusZone: FocusZone) => void
   render: (config: RenderResultConfig<P>, props: P) => React.ReactNode
@@ -22,7 +23,7 @@ export type CreateComponentReturnType<P> = React.FunctionComponent<P> & {
   create: Function
 }
 
-const createComponent = <P extends {} = {}, S extends {} = {}>({
+const createComponent = <P extends {} = ObjectOf<any>>({
   displayName = 'StardustComponent',
   className = 'ui-stardust-component',
   shorthandPropName = 'children',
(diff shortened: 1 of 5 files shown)


packages/react/src/lib/createStardustComponent.tsx
```diff
@@ -18,13 +20,13 @@ export interface CreateStardustComponentConfig<P> {
   actionHandlers?: AccessibilityActionHandlers
 }
 
-const createComponent = <P extends {} = {}, S extends {} = {}>({
+const createComponent = <P extends {} = ObjectOf<any>>({
```
bmdalex: I was also considering the more correct fix:
```suggestion
const createComponent = <P extends {} = React.HTMLAttributes<HTMLElement>>({
```
but for some reason we don't seem to be fans of typing our components as extensions of HTML elements 😕

alinais (approved):

packages/react/src/lib/createComponent.tsx
```diff
@@ -22,7 +23,7 @@ export type CreateComponentReturnType<P> = React.FunctionComponent<P> & {
   create: Function
 }
 
-const createComponent = <P extends {} = {}, S extends {} = {}>({
+const createComponent = <P extends {} = ObjectOf<any>>({
```
kuzhelov: seems that the intent is a bit different here: if we would like to restrict type to objects, I would suggest the following expression for that:
```js
const createComponent = <P extends ObjectOf<any> = {}>({ ...
```

packages/react/src/lib/createStardustComponent.tsx
kuzhelov: yes - lets just start with loose but correct types for now, and steadily restrict them after. The same suggestion here:
```js
const createComponent = <P extends ObjectOf<any> = {}>({ ...
```

packages/react/src/lib/createComponent.tsx
bmdalex: @kuzhelov - your suggestion does not address that

the intent is to have `P` have `ObjectOf<any` by default so that we don't have the errors I mentioned in #946 for the case when we do not supply the props interface to `createComponent` 

From #946 description:

### Actual Result
We have compilation error on our own example from documentation 💣
screenshot 2019-02-22 at 12 25 30
screenshot 2019-02-22 at 12 25 44

packages/react/src/lib/createStardustComponent.tsx
bmdalex: see https://github.com/stardust-ui/react/pull/954#discussion_r260265890

kuzhelov: yes - as a result, have agreed with @Bugaa92 to introduce the following type:

```js
const createComponent = <P extends ObjectOf<any> = {}>({ ..
```
This will prevent the following things:
```js
const Foo = createComponent<number>(...)
const Foo = createComponent<string>(...)
```
and, at the same time, will work properly for the scenarios @Bugaa92 has mentioned 👍

packages/react/src/lib/createComponent.tsx
kuzhelov: yes - as a result, have agreed with @Bugaa92 to introduce the following type:

```js
const createComponent = <P extends ObjectOf<any> = {}>({ ..
```
This will prevent the following things:
```js
const Foo = createComponent<number>(...)
const Foo = createComponent<string>(...)
```
and, at the same time, will work properly for the scenarios @Bugaa92 has mentioned 👍

bmdalex: after discussions with @kuzhelov we agreed to go with `P extends ObjectOf<any> = any`

EDIT: haha one minute faster

packages/react/src/lib/createStardustComponent.tsx
bmdalex: after discussions with @kuzhelov we agreed to go with `P extends ObjectOf<any> = any`

kuzhelov (approved):

Issue #942, opened 2019-02-21
[RFC] ChatMessage - 'reactions' slot


mnajdova:
### Problem description
We got request for adding additional slot to the ChatMessages for the reaction on the message.

### Teams reactions:
![image](https://user-images.githubusercontent.com/4512430/53176470-6be26b00-35ee-11e9-9f46-7503cc8e6b2a.png)

### Skype:
![image](https://user-images.githubusercontent.com/4512430/53176568-a51adb00-35ee-11e9-95c2-33a1d060461c.png)

### FB messanger:
![image](https://user-images.githubusercontent.com/4512430/53176941-68031880-35ef-11e9-8f39-c93f807188f9.png)

### Github post:
![image](https://user-images.githubusercontent.com/4512430/53176834-2a05f480-35ef-11e9-8fdb-dd0e95227dc1.png)

### Proposed solution
Seeing this is a common thing across messaging platforms, I would agree that it can be something added on the ChatMessage. The common thing between the reactions is that they have reaction (Icon maybe?) and number of reactions of that type. We can create ChatMessageReaction encapsulating this, and the ChatMessage may contain shorthand collection for reaction, which can be positioned on the top or bottom of the message.

So it would look something like

```jsx
<Chat.Message 
  author='John Doe' 
  timestamp='02/21' 
  content='Message with reactions' 
  reactions=[{icon: 'like', count: '15'}, {icon: 'angry', count: '1K'}] 
  reactionsPosition='top'
/>
```

I intentionally set the count as a string, because the number if big can be shorten - we may discuss this if it doesn't make sense in general. 

### MVP
[Fork, update, and replace this pen if you can show the proposed solution]:
https://codesandbox.io/s/2l3n74j9y

bmdalex:
Looks good to me @mnajdova 

Few things:

### 1. Is `reactionsPosition` needed?

I am not sure about `reactionsPosition` prop; from your examples it looks like each theme will decide where to position the reactions; also it seems that it'd be pretty difficult to reason about:

- what possible values `reactionsPosition` will have; there can be so many: top, bottom, before, after, combinations of these, etc
- if a reaction has `reactionsPosition='bottom-after' how do we place that relative to the bottom right corner (above it or to the left of it) => **this indicates `reactionsPosition` might be a theme aspect**
- we don't want to enforce all theme to implement different values of `reactionsPosition` if they only need one (vast majority of cases)

### 2. Is `reactionsPosition` enough?

[...]

PR #959, opened 2019-02-25
feat(Reaction): add Reaction component and use it in the ChatMessage

mnajdova:
The aim of this PR is to implement the RFC proposed here: https://github.com/stardust-ui/react/issues/942 
The PR adds the following things:

- `Reaction` component, which contains `icon` and `count` (maybe this should be shorthand of the `Button` component?) 
- `reactions` collection shorthand in the `ChatMessage`. 
- `reactionsPosition` prop in the `ChatMessage` with the values `'start'` and `'end'`, to indicate whether they should appear before or after the content in the generated HTML structure.

Chat message with reactions:
image


packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -65,17 +65,24 @@ const chatMessageStyles: ComponentSlotStylesInput<
   }),
 
   author: ({ props: p, variables: v }): ICSSInJSStyle => ({
+    marginTop: pxToRem(5),
```
mnajdova: I had to add this here in order to resolve some screen regressions, but I am pretty sure this will produce some differences on client's side. Not sure why this happened...

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -74,8 +74,8 @@ const chatMessageStyles: ComponentSlotStylesInput<
     marginBottom: v.headerMarginBottom,
   }),
 
-  content: ({ variables: v }): ICSSInJSStyle => ({
-    display: 'block',
+  content: ({ props: p, variables: v }): ICSSInJSStyle => ({
+    display: p.reactions && p.reactionsPosition === 'end' ? 'inline-block' : 'block',
```
mnajdova: I had to conditionally change the display prop on the content, because if the reactionsPosition is 'end' in Teams theme, the reactions should appear after the content. No better idea in the moment (tried added float elements, but then the paddings are broken...) I didn't want to create additional element for wrapping this, so this seemed like the only option.

packages/react/src/themes/teams/components/Chat/chatItemStyles.ts
```diff
@@ -22,6 +23,9 @@ const getChatMessageEvaluatedStyles = (p: ChatItemProps) => ({
       paddingTop: pxToRem(5),
       paddingBottom: pxToRem(7),
       ...getPositionStyles(p),
+      [chatMessageContentClassNameSelector]: {
```
mnajdova, line 26: I had to conditionally change the display prop on the content, because if the author and timestamp are not visible in Teams theme, the reactions should appear after the content. No better idea in the moment (tried added float elements, but then the paddings are broken...) I didn't want to create additional element for wrapping this, so this seemed like the only option.

docs/src/examples/components/Reaction/index.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
+import Types from './Types'
+
+const ReactionExamples = () => (
+  <div>
+    <Types />
+  </div>
```
layershifter:
```suggestion
  </>
```

docs/src/examples/components/Reaction/index.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
+import Types from './Types'
+
+const ReactionExamples = () => (
+  <div>
```
layershifter:
```suggestion
  <>
```

docs/src/examples/components/Reaction/Types/index.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
```
layershifter, line 3:
```suggestion

import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
```

nit: add empty line, please 🐱

packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -143,14 +153,40 @@ class ChatMessage extends UIComponent<ReactProps<ChatMessageProps>, ChatMessageS
...
+          return Reaction.create(reaction, {
+            defaultProps: {
+              styles: styles.reaction,
+              as: 'span',
+            },
+          })
+        })}
+      </Box>
+    )
+  }
```
layershifter: Should we introduce a separate component `ReactionGroup`?

```jsx
<ReactionGroup items={[]} />
```

```jsx
ReactionGroup.create(reactions)
```

packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -200,12 +237,17 @@ class ChatMessage extends UIComponent<ReactProps<ChatMessageProps>, ChatMessageS
...
+            {reactionsPosition === 'start' && reactions && this.renderReactions(styles)}
+
             {Box.create(content, {
               defaultProps: {
                 className: ChatMessage.slotClassNames.content,
                 styles: styles.content,
               },
             })}
+
+            {reactionsPosition === 'end' && reactions && this.renderReactions(styles)}
```
layershifter:
```jsx
const reactionsElement = reactions && this.renderReactions(styles)
```

```suggestion
            {reactionsPosition === 'end' && reactionsElement}
```

packages/react/src/components/Reaction/Reaction.tsx
```diff
@@ -0,0 +1,102 @@
...
+  icon?: ShorthandValue
+
+  /** A reaction can have count for the number of reaction. */
+  count?: number | string
+}
+
+/**
+ * A segment is used to create a grouping of related content.
+ */
+class Reaction extends UIComponent<ReactProps<ReactionProps>, any> {
```
layershifter:
```suggestion
class Reaction extends UIComponent<ReactProps<ReactionProps>> {
```

packages/react/src/components/Reaction/Reaction.tsx
```diff
@@ -0,0 +1,102 @@
...
+  static slotClassNames: ReactionSlotClassNames
+
+  static displayName = 'Reaction'
+
+  static propTypes = {
+    ...commonPropTypes.createCommon({
+      content: false,
+    }),
+    icon: customPropTypes.itemShorthand,
+    count: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
layershifter: Should be `count` just a regular `content`? I don't see a big difference here.

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -108,6 +108,16 @@ const chatMessageStyles: ComponentSlotStylesInput<
       transform: p.badgePosition === 'start' ? 'translateX(-50%)' : 'translateX(50%)',
     }
   },
+  reactions: ({ variables: v }) => ({
+    marginLeft: v.reactionsMarginLeft,
+    float: 'right',
+  }),
+  reaction: {
+    cursor: 'pointer',
```
layershifter: I think that `cursor` should be moved to `reactionStyles`

packages/react/src/components/Chat/ChatMessage.tsx
mnajdova: I've added the reactions condition here, because if I don't have it, I will create an empty Box there with some styles... I can move the condition to the function though... Now that I think about it, maybe ReactionGroup would be a good way to go...

packages/react/src/components/Reaction/Reaction.tsx
mnajdova: I thought it would be awkward having number as a type for the content.. In the context of Reaction, count made more sense in my opinion. If you have strong opinion on this, I may change it..

layershifter: I'm not sure that we should be very strict there, another case is the localisation:

```jsx
<Reaction icon="thumbs down" count="1.2K" /> // fine
<Reaction icon="thumbs down" count={<Trans count='1.2' />} />// bomb
<Reaction icon="thumbs down" content={<Trans count='1.2' />} />// works and still consistent
```

docs/src/examples/components/Reaction/Types/ReactionExample.shorthand.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react'
+import { Reaction, Flex } from '@stardust-ui/react'
+
+const ReactionExample = () => (
+  <Flex gap="gap.small">
```
bmdalex: I'm normalizing the spacing between items to 8px using `gap="gap.smaller"`, see: https://github.com/stardust-ui/react/pull/945/files#r260245496

can we do the same here?

bmdalex (approved):
minor comments, everything else looks like @layershifter covered it, nice job 👍

packages/react/src/components/Reaction/Reaction.tsx
mnajdova: Agreed, let's keep consistency here, it can be shorthand as every other slot..

docs/src/examples/components/Reaction/Types/ReactionExample.shorthand.tsx
mnajdova: I added the ReactionGroup component, will get rid of this one.

packages/react/src/components/Chat/ChatMessage.tsx
mnajdova: I've added the `ReactionGroup` component, and this has been resolved with it.

packages/react/src/components/Chat/ChatMessage.tsx
mnajdova: Added.

docs/src/examples/components/Reaction/Types/ReactionGroupExample.shorthand.tsx
```diff
@@ -0,0 +1,19 @@
...
+  <ReactionGroup
+    items={[
+      {
+        icon: 'thumbs up',
+        content: 10,
+      },
+      {
+        icon: 'thumbs down',
+        content: '2K',
+      },
```
layershifter:
```
{ icon: 'thumbs up', content: 10 },
{ icon: 'thumbs down', content: '2K' },
```

nit: I think it can be more compact

packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -75,6 +77,11 @@ export interface ChatMessageProps
    * @param {object} data - All props.
    */
   onFocus?: ComponentEventHandler<ChatMessageProps>
+
+  /** Reaction group applied to the message. */
+  reactionGroup?: ShorthandValue
+
+  reactionGroupPosition?: 'start' | 'end'
```
layershifter, line 85: Missing description?

layershifter (approved):

docs/src/examples/components/Chat/Types/ChatMessageExampleStyled.shorthand.tsx
```diff
@@ -1,6 +1,8 @@
 import * as React from 'react'
 import { Avatar, Chat, Provider } from '@stardust-ui/react'
 
+const reactions = [{ icon: 'thumbs up', count: 5 }, { icon: 'thumbs down', count: '1K' }]
```
kuzhelov: nit-nit: just for the sake of keep optimistic and opened to `Hi` messages, lets rebalance numbers of `thumbs up` and `thumbs down` reactions ;)

packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -104,12 +112,15 @@ class ChatMessage extends UIComponent<ReactProps<ChatMessageProps>, ChatMessageS
     timestamp: customPropTypes.itemShorthand,
     onBlur: PropTypes.func,
     onFocus: PropTypes.func,
+    reactionGroup: customPropTypes.itemShorthand,
```
kuzhelov, line 115: client's voice in me would say that originally taken `reactions` name would work better - although we might want to be absolutely pedantic in terms of matching prop names with component types used as a default value, I would rather expect that these two concepts will be decoupled from each other, in general. These are the main reasons for that:
- `reactions` prop name looks nicer to the client
- in general we might foresee that, _for some reason_, `ReactionsGroup` name that was originally taken for this component will change. Then, if we would still strive for strict correspondence between these two, we would need to change the name of the prop in the API as well - and, actually, it might be not the only place where this change would be necessary

docs/src/examples/components/Chat/Content/ChatExampleReactionGroup.shorthand.tsx
```diff
@@ -0,0 +1,58 @@
...
+    key: 'message-1',
+  },
+  {
+    attached: 'bottom',
+    contentPosition: 'end',
+    key: 'message-2',
+    message: {
+      content: (
+        <Chat.Message
+          reactionGroup={{ items: [{ icon: 'thumbs up', content: '8' }] }}
```
kuzhelov, line 30: just as a matter of thought - here, also if I would be a client, would expect the following syntax being possible as well, without any syntax noise `items` prop is currently introducing:

```jsx
<Chat.Message
  reactions={[ 
     { icon: ...},
     { ... }
  ]}
/>
```

As a proof for this desire we might take an initial state of `Reactions` proposal described in the corresponding RFC: https://github.com/stardust-ui/react/issues/942 .

Should note that this issue applies not just to this case exclusively - thus, we might want to think about a general solution, as this is, essentially, about adjusting the logic of how shorthand values are interpreted

layershifter: Actually, it's the same issue as `actionMenu` 💭

docs/src/examples/components/Chat/Types/ChatMessageExampleStyled.shorthand.tsx
mnajdova: Will do

packages/react/src/components/Chat/ChatMessage.tsx
mnajdova: Yep, the reason I renamed it is correlated with the next comment you've added... 

- I would have `reactions={{items: [....]}}` which looks even more awkward.. With preserving the name `reactionGroup`, I would feel less bad for having items there... 
- At this moment, it seems more consistent with the other `shorthands`. 

These are the only reasions why I have chosen this name. I may change it, but then we will have on the same component the props: `actionMenu` and `reactions` (inconsistency even in the same component's API). Let me know what do you think...

Additional thought, maybe for some of the components that represent group of other components, like the Menu, ButtonGroup, ReactionGroup etc, we may create special shorthand that will be list and will be applied on the collection shorthand that they represent. This would solve lots of these inconsistency and awkward api...

docs/src/examples/components/Chat/Content/ChatExampleReactionGroup.shorthand.tsx
mnajdova: Please see my comment: https://github.com/stardust-ui/react/pull/959#discussion_r260761334 Generally at this moment I don't have better proposal, so would like to preserve consistency, so that when we start fixing this we can fix all occurrences and again will have consistency around these collection components

kuzhelov (approved):

packages/react/src/components/Chat/ChatMessage.tsx
kuzhelov: yes, absolutely agree - lets just keep this in mind and defer to follow-up PR the necessary changes for shorthand to be able to process array values. This would solve this problem, as well as the problems that @layershifter has reminded us about here: https://github.com/stardust-ui/react/pull/959/files#r260743449

mnajdova: Here is the created issue for reference: https://github.com/stardust-ui/react/issues/984

PR #960, opened 2019-02-25
fix(RTL): apply static styles using the top-most Fela renderer

### LINKED ISSUE #948: RTL Provider breaks styles
### Steps
```jsx
const buttonStyles = {
    borderLeftColor: 'red',
    borderLeftWidth: '5px',
}

const outerRtl = true
const innerRtl = true

export default () => (
        <Provider theme={mergeThemes(themes.teams, {rtl:outerRtl})}>
            <div>
                <Button content={`outer rtl=${outerRtl.toString()}`} styles={buttonStyles} primary />
                <Provider theme={mergeThemes(themes.teams, {rtl: innerRtl})}>
                    <Button content={`inner rtl=${innerRtl.toString()}`} styles={buttonStyles} primary />
                </Provider>
            </div>
        </Provider>
)
```

### Expected Result
Correctly works for all combinations of `outerRtl` and `innerRtl`

### Actual Result
Results, differ, also depends on `devMode` in `felaRenderer` 💣

#### `devMode: true`
localhost_3000_
localhost_3000_ 1
localhost_3000_ 2
localhost_3000_ 3

With inner provider completely removed:

localhost_3000_ 10
localhost_3000_ 11

#### `devMode: false`
localhost_3000_ 4
localhost_3000_ 5
localhost_3000_ 6
localhost_3000_ 7

With inner provider completely removed:
localhost_3000_ 8
localhost_3000_ 9

### Version
0.21.1
Chrome 72.0.3626.109 Mac

miroslavstastny:
Static styles were always applied on LTR Fela renderer, fixes #948.
Now the first (top level) `Provider` decides which renderer to use, caches it, and all other Providers reuse that renderer to apply static styles.
- [x] update changelog


bmdalex (approved):

packages/react/src/lib/felaRenderer.tsx
```diff
@@ -37,7 +37,10 @@ const createRendererConfig = (options: any = {}) => ({
   ...(options.rendererId ? { rendererId: options.rendererId } : {}),
 })
 
-export const felaRenderer: Renderer = createRenderer(createRendererConfig())
+// { rendererId: 'default' } is a temporary workaround for https://github.com/stardust-ui/react/issues/948#issuecomment-466404630
+export const felaRenderer: Renderer = createRenderer(
+  createRendererConfig({ rendererId: 'default' }),
+)
```
layershifter, line 43: I am not fan of this solution, but the fix in Fela is not ready. Can we mark this as todo to allow find this item easier in the future?

layershifter (approved):

PR #961, opened 2019-02-25
fix(Status): fix order of applied props

### LINKED ISSUE #958: Status: can't override `role` attribute
## Bug Report

I can't override `role` attribute.

### Steps
```jsx
<Status role='presentation' />
```

### Expected Result
```html
<span role="presentation" class="ui-status co ec ch ci hz ia y ib ic"></span>
```

### Actual Result
```html
<span role="img" class="ui-status co ec ch ci hz ia y ib ic"></span>
```
💣

### Version
master

layershifter:
Fixes #958.

### DIFF
--- a/packages/react/test/specs/commonTests/isConformant.tsx
+++ b/packages/react/test/specs/commonTests/isConformant.tsx
@@ -336,13 +336,9 @@ export default (Component, options: Conformant = {}) => {
       expect(element.prop('role')).toBe(role)
     })
 
-    test('attributes passed by consumer has more weight', () => {
-      const wrapper = mount(
-        <Component
-          {...{ ...requiredProps, [IS_FOCUSABLE_ATTRIBUTE]: false }}
-          accessibility={noopBehavior}
-        />,
-      )
+    test("client's attributes override the ones provided by Stardust", () => {
+      const wrapperProps = { ...requiredProps, [IS_FOCUSABLE_ATTRIBUTE]: false }
+      const wrapper = mount(<Component {...wrapperProps} accessibility={noopBehavior} />)
       const element = getComponent(wrapper)
 
       expect(element.prop(IS_FOCUSABLE_ATTRIBUTE)).toBe(false)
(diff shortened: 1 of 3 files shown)


packages/react/test/specs/commonTests/isConformant.tsx
```diff
@@ -319,18 +329,24 @@ export default (Component, options: Conformant = {}) => {
...
-      })
-
       const wrapper = mount(<Component {...requiredProps} accessibility={noopBehavior} />)
       const element = getComponent(wrapper)
 
+      expect(element.prop(IS_FOCUSABLE_ATTRIBUTE)).toBe(true)
       expect(element.prop('role')).toBe(role)
     })
+
+    test('attributes passed by consumer has more weight', () => {
```
kuzhelov: nit-nit: maybe we could just state supported scenario, like `client's attributes override the ones provided by Stardust`

packages/react/test/specs/commonTests/isConformant.tsx
```diff
@@ -319,18 +329,24 @@ export default (Component, options: Conformant = {}) => {
...
+      const wrapper = mount(
+        <Component
+          {...{ ...requiredProps, [IS_FOCUSABLE_ATTRIBUTE]: false }}
```
kuzhelov: would suggest to make it a bit more readable: 

```jsx
const clientAttributes = {
   [IS_FOCUSABLE_ATTRIBUTE]: false,
   accessibility: noopBehavior
}

const wrapper = mount(<Component {...requiredProps} {...clientAttributes} />
...
```

packages/react/test/specs/commonTests/isConformant.tsx
kuzhelov: 👍 thanks for adding this one!

kuzhelov (approved):
would like @jurokapsiar to take a look on these changes, just to be absolutely sure this strategy won't get into conflict with existing accessibility behaviours

PR #962, opened 2019-02-25
fix(ChatMessage): styles

kuzhelov:


packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -14,19 +14,27 @@ const chatMessageStyles: ComponentSlotStylesInput<
 > = {
   root: ({ props: p, variables: v }): ICSSInJSStyle => ({
     display: 'inline-block',
+    position: 'relative',
+
+    marginLeft: p.mine ? pxToRem(140) : 0,
```
mnajdova: Shouldn't this be `100px`? I think we should calculate with rem being `10px` when converting this values.

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -14,19 +14,27 @@ const chatMessageStyles: ComponentSlotStylesInput<
...
+    marginRight: !p.mine ? pxToRem(140) : 0,
```
mnajdova: `100px`?

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -14,19 +14,27 @@ const chatMessageStyles: ComponentSlotStylesInput<
...
+    maxWidth: `calc(100% - ${pxToRem(140)})`,
```
mnajdova: `calc(100% - ${pxToRem(100)})`

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -14,19 +14,27 @@ const chatMessageStyles: ComponentSlotStylesInput<
...
+    minWidth: pxToRem(140),
```
mnajdova: `pxToRem(100)`

packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
```diff
@@ -4,13 +4,13 @@ export interface ChatMessageVariables {
   actionMenuBoxShadow: string
   actionMenuPositionRight: string
   actionMenuPositionTop: string
-  width: string
```
bmdalex, line 7: @kuzhelov we're removing variables, doesn't this mean it's a breaking change?

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
bmdalex: can we have this magic `140` (or `100`, or even better, the whole margin value, `pxToRem(140`)) value in one place only (instead of 4) to:
- decrease risk of regression if it gets changed
- decrease the number of changes if it becomes a variable

bmdalex (commented):
looks good, couple of comments

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
kuzhelov: yes, lets do that :)

packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
kuzhelov: actually, this is a change introduced to the consumer's code (without any change provided to Stardust's approach to styling) - thus I wouldn't expect increment of Stardust version because of that

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -12,47 +12,59 @@ const chatMessageStyles: ComponentSlotStylesInput<
...
-      },
-    }),
+    return {
+      display: 'inline-block',
+      position: 'relative',
+
+      marginLeft: p.mine ? marginOffset : 0,
+      marginRight: !p.mine ? marginOffset : 0,
+      maxWidth: `calc(100% - ${marginOffset})`,
+      minWidth: marginOffset,
```
mnajdova: Would introduce the marginOffset as a variable, but it is used as minWidth as well, which will mean maybe we need variables... Just thinking out loud :) Maybe it's better if it stays as it is..

mnajdova (approved):

packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
bmdalex: makes sense

bmdalex (approved):
lgtm 👍

PR #966, opened 2019-02-26
feat(dropdown): control highlightedIndex from Dropdown

### LINKED ISSUE #903: feat(Drodpown): add means to highlight first option by default
Provide functionality to make first option to be highlighted when search query changes.

Even when there is another option highlighted, changing the search query should move highlight to first option.

Can be achieved by controlling `highlightedIndex` and using `onSearchQuery` prop, but it involves controlling the prop so we may come up with another way.

silviuaavram:
Means to control `highlightedIndex` prop and fixes https://github.com/stardust-ui/react/issues/903


packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -616,12 +632,42 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+        if (!this.props.multiple && !this.props.search) {
+          if (changes.selectedItem) {
+            newState['highlightedIndex'] = this.props.items.indexOf(changes.selectedItem)
+          } else if (this.state.value) {
+            newState['highlightedIndex'] = this.props.items.indexOf(this.state.value)
+          }
+        } else {
+          newState['highlightedIndex'] = this.props.resetHighlightedIndex || null
+        }
+      }
```
layershifter, line 664: Sorry, but I lost there 😴 If you will try to cover this code with UTs you will not survive 🦇 

1. Can we extract code that calculates only `highlightedIndex` based on changes/props?
2. Can we dispatch DOM effects in `componentDidUpdate()`? I.e. in a single place based on the component's state.
```jsx
componentDidUpdate(prevProps, prevState) {
  if (!prevState.open && this.state.open) {
    if (search) {
      this.listRef.current.focus()
    }
    if (multiple) {
      this.triggerRef.current.focus()
    }
  }
}
```

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -244,6 +254,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     placeholder: PropTypes.string,
     renderItem: PropTypes.func,
     renderSelectedItem: PropTypes.func,
+    resetHighlightedIndex: PropTypes.number,
```
layershifter: I don't understand why we need this prop when we can control the `highlightedIndex` directly, can you please clarify?

silviuaavram: we need it so that users don't have to make a whole different `highlightedIndex` controlling logic when they just want for the Dropdown to always open at a specified index. For example Add People picker in WebClient, that opens always with first item highlighted.

layershifter: For me it looks like a case with controlled usage, I think that we should not introduce new props that do exactly the same thing.

/cc @kuzhelov

silviuaavram: I totally agree. If you look at Downshift's documentation, for controlled props, they have also two helpers, `initial` and `default`. `initial` works similar to our `default` while `default` works similarly to this `reset`. I personally think both of them are useful, along with the actual controlling prop. so if we can converge to a solution, that'd be great.

silviuaavram: followed up with this https://github.com/stardust-ui/react/issues/970 in which we can discuss more on the subject.

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: I extracted the code, you can take a look. as for the DOM effects, it's out of scope, but we can create a separate issue. I also want to move the methods around in the Dropdown, since it's kind of chaos there. should be render, then renderWhatevers, then event handlers, then helpers. Or something like that, but grouped better, since it's a 1100 line file.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1074,6 +1054,42 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+  private getHighlightedIndexOnArrowKeyOpen = (
+    changes: StateChangeOptions<ShorthandValue>,
+  ): number => {
+    if (changes.type === Downshift.stateChangeTypes.keyDownArrowUp) {
+      const itemsLength = this.getItemsFilteredBySearchQuery().length
+      return _.isNumber(this.state.highlightedIndex)
+        ? (() => {
+            const newIndex = this.state.highlightedIndex - 1
+            return newIndex < 0 ? itemsLength - 1 : newIndex
+          })()
```
layershifter: What is a actual reason to introduce IIFE there? I also see that it's the same function as on line 1073 🤔

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1074,6 +1054,42 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+    if (changes.type === Downshift.stateChangeTypes.keyDownArrowDown) {
+      const itemsLength = this.getItemsFilteredBySearchQuery().length
+      return _.isNumber(this.state.highlightedIndex)
+        ? (() => {
+            const newIndex = this.state.highlightedIndex + 1
+            return newIndex >= itemsLength ? 0 : newIndex
+          })()
+        : 0
+    }
+    return undefined
```
layershifter: This method returns `undefined` while `getHighlightedIndexOnClose()` returns `null` 💭

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: will change it, I had it when this was in the onStateChange.

packages/react/src/components/Dropdown/Dropdown.tsx
layershifter: May be in this case we should return the current value?

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -617,11 +634,26 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
 
   private handleStateChange = (changes: StateChangeOptions<ShorthandValue>) => {
     if (changes.isOpen !== undefined && changes.isOpen !== this.state.open) {
-      this.trySetStateAndInvokeHandler('onOpenChange', null, { open: changes.isOpen })
+      const newState: { open: boolean; highlightedIndex?: number } = { open: changes.isOpen }
```
layershifter: Now we have these methods that are much more readable 👍 
Can we simplify more?

```jsx
const highlightedIndex = changes.isOpen ? this.getHighlightedIndexOnArrowKeyOpen(changes) : this.getHighlightedIndexOnClose(changes)
const newState = { open: changes.isOpen, highlightedIndex }

if (changes.isOpen && !this.props.search) {
  this.listRef.current.focus()
}
```

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: it's like this because when the menu opens by arrow up/down, we want to modify the highlighted index already saved in state. otherwise, it needs to be left alone, so it will be opened at whatever it was set in the close function.

when the menu closes, we want the index to be either selectedItem for single selection, 0 in case boolean prop is passed with true, or null if nothing else, so we make sure the dropdown opens with no highlight.

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: the only solution is to have `const newState = { open: changes.isOpen, ...{highlightedIndex} }` to make sure that the highlightedIndex is not passed as `undefined`. otherwise Downshift will complain that we are transforming a controlled prop to uncontrolled.

silviuaavram:
should also aim to fix https://github.com/stardust-ui/react/issues/979

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -617,11 +634,26 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+        }
+        if (!this.props.search) {
+          this.listRef.current.focus()
+        }
+      } else {
+        newState.highlightedIndex = this.getHighlightedIndexOnClose()
+      }
+
+      this.trySetState(newState)
+      _.invoke(this.props, 'onOpenChange', null, { ...this.props, ...newState })
```
bmdalex: logic on lines 651 and 652 was encapsulated in the method that used to be at line 620; can you please use that again?

```typescript
this.trySetStateAndInvokeHandler('onOpenChange', null, newState)
```

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1023,6 +1054,42 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
   private isValueEmpty = (value: ShorthandValue | ShorthandCollection) => {
     return _.isArray(value) ? value.length < 1 : !value
   }
+
+  private getHighlightedIndexOnArrowKeyOpen = (
+    changes: StateChangeOptions<ShorthandValue>,
+  ): number => {
+    // if open by ArrowUp, index should change by -1.
+    if (changes.type === Downshift.stateChangeTypes.keyDownArrowUp) {
```
bmdalex: we were using switch/case for this type of conditions; let's aim for consistency and do the same here

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1023,6 +1054,42 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+      const itemsLength = this.getItemsFilteredBySearchQuery().length
+      if (_.isNumber(this.state.highlightedIndex)) {
+        const newIndex = this.state.highlightedIndex - 1
+        return newIndex < 0 ? itemsLength - 1 : newIndex
+      }
+      return itemsLength - 1
+    }
+    // if open by ArrowDown, index should change by +1.
+    if (changes.type === Downshift.stateChangeTypes.keyDownArrowDown) {
```
bmdalex: this should have been if/else statement anyway as the conditions are independent

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1023,6 +1054,42 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+      const itemsLength = this.getItemsFilteredBySearchQuery().length
+      if (_.isNumber(this.state.highlightedIndex)) {
+        const newIndex = this.state.highlightedIndex + 1
+        return newIndex >= itemsLength ? 0 : newIndex
+      }
+      return 0
```
bmdalex: can we encapsulate logic form lines 1072-1077 (and 1063-1068) in smaller methods? logic is complex and because of branching it becomes hard to read

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1023,6 +1054,42 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+    }
+    // if not opened by Arrow, index should not change at all.
+    return undefined
+  }
+
+  private getHighlightedIndexOnClose = (): number => {
```
bmdalex, line 1079: NIT: for readability:
```typescript
private getHighlightedIndexOnClose = (): number => {
  const { highlightFirstItemOnOpen, items, multiple, search } = this.props
  const { value } = this.state

  if (!multiple && !search && value) {
    // in single selection, if there is a selected item, highlight it.
    return items.indexOf(value)
  }

  if (highlightFirstItemOnOpen) {
    // otherwise, if highlightFirstItemOnOpen prop is provied, highlight first item
    return 0
  }

  // otherwise, highlight no item
  return null
}
```

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1023,6 +1054,42 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+    if (!this.props.multiple && !this.props.search) {
+      // in single selection, if there is a selected item, highlight it.
+      if (this.state.value) {
+        return this.props.items.indexOf(this.state.value)
```
bmdalex: it'd be nice to keep the index of the selected items in the state so that we don't have to perform a search every time; this would become something like:
```typescript
return this.state.selectedItemIndex // or this.state.valueIndex
```

bmdalex (approved):
some comments, good job 👍

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: there are just 4 lines each, and a separate method with some more if-else to change values depending on that `stateChangeTypes` ... I would rather leave it like this.

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: thanks, will do

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: will change.

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: thanks, will use that version

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: I don't know. that thing is used on selection. so once, twice maybe. then on each opening when dropdown is single selection, only when is already selected. once, twice maybe.

should we bother adding more complexity?

bmdalex: then it's fine, no need to optimize, let's keep it simple, thx

layershifter (approved):

PR #969, opened 2019-02-26
feat(Loader): add `delay` prop

### LINKED ISSUE #888: Loader needs a delay property
### Problem description
In some cases, loading happens immediately, in which case, we don't want the loading spinner to be rendered at all.

### Proposed solution
Add a delay property to the Loader component to indicate how long to wait before showing it

layershifter:
Fixes #888. See https://github.com/stardust-ui/react/issues/888#issuecomment-467360455.


packages/react/src/components/Loader/Loader.tsx
```diff
@@ -24,9 +24,15 @@ export interface LoaderProps extends UIComponentProps, ColorComponentProps {
...
   accessibility?: Accessibility
 
+  /** Time in milliseconds after component mount before spinner is visible. */
+  delay?: number
+
   /** A loader can contain an indicator. */
   indicator?: ShorthandValue
 
+  /** Loaders can appear inline with content. */
+  inline?: boolean
```
layershifter, line 34: Added missing prop definition

packages/react/src/components/Loader/Loader.tsx
```diff
@@ -51,27 +61,61 @@ class Loader extends UIComponent<ReactProps<LoaderProps>> {
...
     indicator: customPropTypes.itemShorthand,
+    inline: PropTypes.bool,
     label: customPropTypes.itemShorthand,
     labelPosition: PropTypes.oneOf(['above', 'below', 'start', 'end']),
     size: customPropTypes.size,
   }
 
   static defaultProps = {
     accessibility: loaderBehavior,
+    delay: 0,
```
johannao76, line 74: Should be provide a more reasonable default? Maybe 100ms?

packages/react/src/components/Loader/Loader.tsx
```diff
@@ -51,27 +61,61 @@ class Loader extends UIComponent<ReactProps<LoaderProps>> {
...
+      visible: this.props.delay === 0,
+    }
+  }
+
+  componentDidMount() {
+    const { delay } = this.props
+
+    if (delay > 0) {
+      this.delayTimer = window.setTimeout(() => {
+        this.setState({ visible: true })
```
johannao76, line 95: Seems like we should clear the timeout here too and set this.delayTimer = null

packages/react/src/components/Loader/Loader.tsx
```diff
@@ -51,27 +61,61 @@ class Loader extends UIComponent<ReactProps<LoaderProps>> {
...
+      }, delay)
+    }
+  }
+
+  componentWillUnmount() {
+    clearTimeout(this.delayTimer)
```
johannao76, line 101: Check if this.delayTimer is not null/undefined before calling clearTimeout

johannao76 (approved):
Thanks for adding this!

packages/react/src/components/Loader/Loader.tsx
layershifter: By default it should `0` to keep existing behavior.

packages/react/src/components/Loader/Loader.tsx
layershifter: They will never collide because it's called in `componentDidMount()`, there are no obvious reasons to clear it.

packages/react/src/components/Loader/Loader.tsx
layershifter: Can you please clarify why need it?

> Passing an invalid ID to clearTimeout() silently does nothing; no exception is thrown.

https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/clearTimeout#Notes

### IE 11

image

### Chrome

image

mnajdova:
LGTM, only thing is that maybe we should add test for this behavior?

docs/src/examples/components/Loader/Usage/LoaderExampleDelay.tsx
```diff
@@ -0,0 +1,8 @@
+import { Loader } from '@stardust-ui/react'
+import * as React from 'react'
+
+const LoaderExampleDelay: React.FC<{ knobs: { mounted: boolean } }> = ({ knobs }) => (
+  <div style={{ minHeight: 24 }}>{knobs.mounted && <Loader delay={500} />}</div>
```
mnajdova, line 5: Maybe use here Flex with padding instead of div?

mnajdova (approved):

docs/src/examples/components/Loader/Usage/LoaderExampleDelay.tsx
layershifter: Actually, I don't think that it's a case for `Flex`. We require there `minHeight` because `Loader` will render `null` with `delay`, the same behavior as there: https://evergreen.segment.com/components/spinner/

Did I miss something?

layershifter: I want to avoid this effect:

loader-delay

mnajdova: Aha, got it! Let's leave it as it is then..

PR #972, opened 2019-02-26
fix(perf): get current commit tag in CI

miroslavstastny:
Fix the way how we get a tag of a current commit in CI.
Changelog left blank.


.circleci/config.yml
```diff
@@ -82,7 +82,7 @@ jobs:
           name: Save Statistics to DB
           command: |
             if [ -n "${STATS_URI}" ]; then
-              yarn stats:save --tag=`git tag --points-at`
+              yarn stats:save --tag=`git tag --points-at HEAD`
```
miroslavstastny, line 85: It does not work without the `HEAD` in CI

bmdalex (approved):

build/gulp/tasks/stats.ts
```diff
@@ -148,7 +148,10 @@ function readCurrentBundleStats() {
 }
 
 task('stats:save', async () => {
-  const commandLineArgs = _.pick(argv, ['sha', 'branch', 'tag', 'pr', 'build'])
+  const commandLineArgs = _.pickBy(
+    _.pick(argv, ['sha', 'branch', 'tag', 'pr', 'build']),
+    val => val !== '', // ignore empty strings
```
layershifter, line 153: NIT: 
```suggestion
    _.identity, // ignore empty strings
```

layershifter (approved):

build/gulp/tasks/stats.ts
miroslavstastny: The reason why I haven't used `_.idenitity` is that it removes all falsy values:

```js
_.pickBy({stringVal: '', numberVal: 0, booleanVal: false}, _.identity)
{}

_.pickBy({stringVal: '', numberVal: 0, booleanVal: false}, val => val !== '')
{numberVal: 0, booleanVal: false}
```

So if you want to be able to specify `--build=0` on command line, you cannot use `_.identity`.

layershifter: 👍

PR #976, opened 2019-02-26
fix(avatar): sizing issues

kuzhelov:
## !! Important follow-up

Please, note that `box-sizing` issue is, essentially, the one that we might expect for other components to appear. This becomes to be an issue after static styles of Semantic UI were eliminated - one of the rules they had introduced was `box-sizing` one.

## TODO 
- [x] update changelog

--------

Appearance of Avatar component in Teams High Contrast theme was fixed. Specifically, the following 
aspects were addressed:

## Avatar sizing issue

When theme switch to Teams Hight Contrast (and, apart from other style changes, `border` being introduced to Avatar), Avatar's size was increased by border's thickness value:

### Teams default theme (size is 32px)

image

### Teams High Contrast theme (size has changed to 32px + 2 * 2px)

image

This has happened because of absence of `box-sizing: border-box` CSS rule.

## Font-based `Icon` component 

When is provided into `image` slot of `Avatar` component (originally reported here https://github.com/stardust-ui/react/pull/795): 

### Was
image

### Now
image

image


packages/react/src/themes/teams/components/Avatar/avatarStyles.ts
```diff
@@ -26,6 +27,7 @@ const avatarStyles: ComponentSlotStylesInput<AvatarProps, any> = {
     }
   },
   image: ({ variables: v }): ICSSInJSStyle => ({
+    boxSizing: 'border-box',
```
mnajdova, line 30: I saw that you added the same thing in the image, that would mean that we don't need i there.

packages/react/src/themes/teams/components/RadioGroup/radioGroupItemStyles.ts
```diff
@@ -31,6 +31,7 @@ const radioStyles: ComponentSlotStylesInput<
 
   icon: ({ props: p, variables: v }): ICSSInJSStyle => ({
     margin: `0 ${pxToRem(10)} 0 0`,
+    padding: 0,
```
mnajdova: I am confused with this change... Can you explain what this was added?

packages/react/src/themes/teams/components/Avatar/avatarStyles.ts
kuzhelov: yes, but this is, actually, necessary, as not only the `Image` component could be rendered here. Essentially, we would like to ensure that all sizing props defined for this slot will be applied in the expected way

mnajdova (approved):

packages/react/src/themes/teams/components/RadioGroup/radioGroupItemStyles.ts
kuzhelov: yes, have addressed related visual regression in icon's styles, thank you!

PR #980, opened 2019-02-27
chore(screener): various improvements

layershifter:
### Add `--filter` arg to `yarn test:visual`

image

### Adds typings for steps

This simplify adding steps because you will have autocomplete 😺 

```tsx
const steps: ScreenerSteps = [
  steps =>
    steps
      .click(selectors.triggerButton)
      .click(selectors.item(3))
      .snapshot('Selects an item'),
]
```

### Adds automatic reset

The component will be automatically reset on each step, no confusions with steps more ⭐️

### DIFF
--- a/build/gulp/tasks/screener.ts
+++ b/build/gulp/tasks/screener.ts
@@ -1,4 +1,5 @@
 import { task, series } from 'gulp'
+import { argv } from 'yargs'
 
 import sh from '../sh'
 import config from '../../../config'
@@ -10,6 +11,9 @@ const { paths } = config
 // ----------------------------------------
 
 task('screener:runner', cb => {
+  // screener-runner doesn't allow to pass custom options
+  if (argv.filter) process.env.SCREENER_FILTER = argv.filter
+
   // kill the server when done
   sh(`screener-runner --conf ${paths.base('screener.config.js')}`)
     .then(() => {
--- a/screener.config.js
+++ b/screener.config.js
@@ -1,10 +1,13 @@
 require('ts-node').register()
 
+const chalk = require('chalk')
 const _ = require('lodash')
 const glob = require('glob')
+const minimatch = require('minimatch')
 const path = require('path')
 const fs = require('fs')
 const tsPaths = require('tsconfig-paths')
+const Keys = require('screener-runner/src/keys')
 const Steps = require('screener-runner/src/steps')
 
 const { default: config } = require('./config')
@@ -16,6 +19,17 @@ tsPaths.register({
 })
 
 const SCREENER_HOST_URL = `${config.server_host}:${config.server_port}`
+
+const examplePaths = glob.sync('docs/src/examples/**/*.tsx', {
+  ignore: ['**/index.tsx', '**/*.knobs.tsx'],
+})
+const pathFilter = process.env.SCREENER_FILTER
+const filteredPaths = minimatch.match(examplePaths, pathFilter || '*', { matchBase: true })
+
+if (pathFilter) {
+  console.log(chalk.bgGreen.black(' --filter '), pathFilter)
+  filteredPaths.forEach(filteredPath => console.log(`${_.repeat(' ', 10)} ${filteredPath}`))
+}
 
 // https://github.com/screener-io/screener-runner
 const screenerConfig = {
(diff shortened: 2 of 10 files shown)


screener.config.js
```diff
@@ -41,38 +55,40 @@ const screenerConfig = {
...
-    ? require(`./${dir}/${nameWithoutExtension}.steps`)
-        .default.reduce((stepsAcc, steps) => steps(stepsAcc), new Steps())
-        .end()
-    : undefined
+  if (fs.existsSync(stepsSpecModulePath)) {
+    const stepsDefinition = require(`./${dir}/${nameWithoutExtension}.steps`).default
+
+    return stepsDefinition
+      .reduce((stepsAcc, steps) => {
+        return steps(stepsAcc, Keys).executeScript(`window.resetExternalLayout()`)
```
bmdalex, line 88: how much time does this take; we have to be careful because it's executed for every screenshot

types/screener.d.ts
```diff
@@ -0,0 +1,66 @@
...
+  rightArrow: string
+  downArrow: string
+}
+
+type ScreenerRunnerSteps = {
+  /** This will click on the first element matching the provided css selector. */
+  click(selector: string): ScreenerRunnerSteps
+
+  /** This will capture a visual snapshot. */
+  snapshot(name: string): ScreenerRunnerSteps
```
bmdalex: has an optional parameter as well:
```suggestion
  snapshot(name: string, options?: any): ScreenerRunnerSteps
```

types/screener.d.ts
```diff
@@ -0,0 +1,66 @@
...
+  mouseDown(selector: string): ScreenerRunnerSteps
+
+  /** This will release the mouse button. selector is optional. */
+  mouseUp(selector: string): ScreenerRunnerSteps
+
+  /** This will set cursor focus on the first element matching the provided css selector. */
+  focus(selector: string): ScreenerRunnerSteps
+
+  /** This will set the value of the input field matching the provided css selector. */
+  setValue(selector: string, value: string): ScreenerRunnerSteps
```
bmdalex: has an optional parameter as well:
```suggestion
  setValue(selector: string, value: string, options?: any): ScreenerRunnerSteps
```

bmdalex (approved):
looks good to me, just few comments

screener.config.js
layershifter: It's still runs around 6 minutes, this stuff is actually very light-weight :)

types/screener.d.ts
```diff
@@ -0,0 +1,66 @@
...
+  return: string
+  shift: string
+  tab: string
+  leftArrow: string
+  upArrow: string
+  rightArrow: string
+  downArrow: string
+}
+
+type ScreenerRunnerSteps = {
```
kuzhelov: rename to `ScreenerStepComposer` or `ScreenerStepBuilder`? Otherwise it is very hard to read the last two declarations of this file:
- ScreenerSteps -> many of ScreenerStep
- ScreenerStep -> (steps, keys) // ?

build/gulp/tasks/screener.ts
```diff
@@ -10,6 +11,9 @@ const { paths } = config
 // ----------------------------------------
 
 task('screener:runner', cb => {
+  // screener-runner doesn't allow to pass custom options
+  if (argv.filter) process.env.SCREENER_FILTER = argv.filter
```
kuzhelov, line 15: just to be sure - isn't it possible to pass argument here and capture it in `screener.config` via `argv`?

screener.config.js
kuzhelov, line 88: just to clarify - is it intentional that `executeScript` is not introduced to  `ScreenerRunnerSteps` type?

types/global.d.ts
```diff
@@ -17,4 +17,5 @@ declare module '*.json' {
 
 declare interface Window {
   prettierPlugins: any
+  resetExternalLayout: () => void
```
kuzhelov: shouldn't we declare this method as optional one, given that this type is globally applied to all scenarios and not just screener tests?

kuzhelov (approved):

types/screener.d.ts
layershifter: `ScreenerStepBuilder` looks like the correct one 👍

build/gulp/tasks/screener.ts
layershifter: Exactly 😭 

```
yarn run v1.13.0
$ C:\Users\olfedias\WebstormProjects\react\node_modules\.bin\screener-runner --filter

  error: unknown option `--filter'

info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

screener.config.js
layershifter: Missed it in type definitions, added 👍

types/global.d.ts
layershifter: Good catch 👍

PR #981, opened 2019-02-27
feat(lib): export `getNextElement`, `getPreviousElement` and `focusAsync`

layershifter:
This PR:
- add exports `getNextElement`, `getPreviousElement` and `focusAsync` to `focusZoneUtilities`.
- exports `ShorthandValue`

### DIFF
--- a/packages/react/src/index.ts
+++ b/packages/react/src/index.ts
@@ -213,7 +213,7 @@ import {
   getPreviousElement,
   focusAsync,
 } from './lib/accessibility/FocusZone/focusUtilities'
-export const focusZoneUtilities = {
+export const FocusZoneUtilities = {
   getFirstTabbable,
   getLastTabbable,
   getNextElement,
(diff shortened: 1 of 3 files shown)


packages/react/src/index.ts
```diff
@@ -201,11 +201,23 @@ export {
...
-import { getFirstTabbable, getLastTabbable } from './lib/accessibility/FocusZone/focusUtilities'
-export const FocusZoneUtilities = { getFirstTabbable, getLastTabbable }
+import {
+  getFirstTabbable,
+  getLastTabbable,
+  getNextElement,
+  getPreviousElement,
+  focusAsync,
+} from './lib/accessibility/FocusZone/focusUtilities'
+export const focusZoneUtilities = {
```
kuzhelov: just one thing - shouldn't it be a breaking change, given this code on the client's side:

```jsx
import { FocusZoneUtilities, ... } from '@stardust-ui/react'
```

kuzhelov (approved):

packages/react/src/index.ts
layershifter: Oh, I see, great catch 👍

PR #934, opened 2019-02-20
fix: (menu & menuitem): vertical menu theming updates

jaanus03:
The aim of this this PR is to update vertical menu styles to make it look great again.

**Current vertical menu styles:** 
screenshot 2019-02-20 at 14 30 13
screenshot 2019-02-20 at 14 38 26
screenshot 2019-02-20 at 14 38 42

**New vertical menu styles:**
screenshot 2019-02-20 at 14 43 16
screenshot 2019-02-20 at 14 43 04
screenshot 2019-02-20 at 14 42 46

PR #983, opened 2019-02-27
feat(screenshot tests): menu states and keyboard navigation

bmdalex:
### Description

`Menu` is a complex component when it comes to styling combined with different states.

This PR adds screenshot tests for `Menu` in different UI states in order to reduce risk of regressions, especially with upcoming PRs in that area such as: #934 . It contains test for:
- `Menu` items in `hover` state
screenshot 2019-02-27 at 15 49 03

- `Menu` items in `active` state
screenshot 2019-02-27 at 15 49 12

- `Menu` items in `focus` state
screenshot 2019-02-27 at 15 49 26

- `Menu` with keyboard navigation to outline keyboard focus
screenshot 2019-02-27 at 15 49 49


docs/src/examples/components/Menu/commonSteps.ts
```diff
@@ -0,0 +1,30 @@
...
+    steps
+      .keys(selectors.item(2), vertical ? keys.downArrow : keys.rightArrow)
+      .snapshot('Navigates to next item (focus state styles)'),
+  (steps, keys) =>
+    steps
+      .keys(selectors.item(3), vertical ? keys.upArrow : keys.leftArrow)
+      .snapshot('Navigates to previous item (active and focus state styles)'),
+]
+
+export default getSteps
```
silviuaavram: why didn't the linter complain about the lack of end empty line?

silviuaavram (approved):
cool!

docs/src/examples/components/Menu/commonSteps.ts
bmdalex: github doesn't show it I think, locally it's there..

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `getNextElement`, `getPreviousElement` and `focusAsync` to exported as `FocusZoneUtilities` @layershifter ([#981](https://github.com/stardust-ui/react/pull/981))
 - Add `Reaction` and `ReactionGroup` components @mnajdova ([#959](https://github.com/stardust-ui/react/pull/959))
 - Add `reactionGroup` and `reactionGroupPosition` props to the `ChatMessage` component @mnajdova ([#959](https://github.com/stardust-ui/react/pull/959))
+- Add screener tests with steps for `Menu` @Bugaa92 ([#983](https://github.com/stardust-ui/react/pull/983))
```
layershifter: I don't see any user facing changes there 🤔

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1026,11 +1028,12 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
 }
 
 Dropdown.slotClassNames = {
-  container: `${Dropdown.className}__container`,
   clearIndicator: `${Dropdown.className}__clear-indicator`,
-  triggerButton: `${Dropdown.className}__trigger-button`,
+  container: `${Dropdown.className}__container`,
+  indicator: `${Dropdown.className}__indicator`,
```
layershifter: Let's use `toggleIndicator`

docs/src/examples/components/Dropdown/Types/DropdownExampleSearchMultiple.shorthand.steps.ts
```diff
@@ -1,19 +1,23 @@
-import { Dropdown, List, Indicator, Input, Label } from '../../../../../../packages/react/src'
-import * as Keys from 'screener-runner/src/keys'
+import { Dropdown, DropdownSearchInput } from '@stardust-ui/react'
+
+const selectors = {
+  indicator: `.${Dropdown.slotClassNames.indicator}`,
```
layershifter: https://github.com/stardust-ui/react/pull/983/files#r261273038

docs/src/examples/components/Menu/Slots/MenuExampleIconOnly.shorthand.steps.ts
```diff
@@ -0,0 +1,3 @@
+import getSteps from '../commonSteps'
```
layershifter:
```suggestion
import getSteps from '../commonScreenerSteps'
```

docs/src/examples/components/Menu/Slots/MenuExampleIconOnly.shorthand.steps.ts
```diff
@@ -0,0 +1,3 @@
...
+
+export default getSteps()
```
layershifter:
```suggestion
export default getScreenerSteps()
```

It's clear when they defined, but it's clear when we just export them :)

docs/src/examples/components/Menu/Variations/index.tsx
```diff
@@ -9,6 +9,11 @@ const Variations = () => (
       description="A vertical menu displays elements vertically."
       examplePath="components/Menu/Variations/MenuExampleVertical"
     />
+    <ComponentExample
+      title="Vertical &amp; Pointing"
+      description="A vertical menu displays elements vertically and point to show its relationship to nearby content."
+      examplePath="components/Menu/Variations/MenuExampleVerticalPointing"
```
layershifter, line 14: Please remove `title` and `description` there, it's a simple variation of mixed props.

docs/src/examples/components/Menu/commonSteps.ts
```diff
@@ -0,0 +1,30 @@
...
+  vertical?: boolean
+}
+
+const menuSelector = `.${Menu.className}`
+const selectors = {
+  menu: menuSelector,
+  item: (itemIndex: number) => `${menuSelector} li:nth-child(${itemIndex}) a`,
+}
+
+const commonSteps = [
```
layershifter: It there any reason to keep this as a separate variable?

docs/src/examples/components/Menu/commonSteps.ts
```diff
@@ -0,0 +1,30 @@
+import { Menu } from '@stardust-ui/react'
+
+interface StepsOptions {
+  vertical?: boolean
+}
+
+const menuSelector = `.${Menu.className}`
+const selectors = {
+  menu: menuSelector,
+  item: (itemIndex: number) => `${menuSelector} li:nth-child(${itemIndex}) a`,
```
layershifter:
```suggestion
  menu: `.${Menu.className}`,
  item: (itemIndex: number) => `.${Menu.className} li:nth-child(${itemIndex}) a`,
```

We can omit `menuSelector` as it doesn't make selectors more readable

docs/src/examples/components/Menu/Variations/index.tsx
bmdalex: @layershifter why not keep the title at least? users will get confused

CHANGELOG.md
bmdalex: not really but we've been adding CHANGELOG entries for this type of work before, I'm ok with removing

docs/src/examples/components/Menu/Variations/index.tsx
layershifter: We should show more in less examples, for example:
https://react.semantic-ui.com/collections/menu/#types-vertical

May be we can even combine, `MenuExampleVertical` and `MenuExampleVerticalPointing` to a single example as it's done there:
https://stardust-ui.github.io/react/components/menu#types-pointing

CHANGELOG.md
layershifter: https://github.com/stardust-ui/react/pull/911#discussion_r257390850
It looks like the same situation...

docs/src/examples/components/Menu/commonSteps.ts
bmdalex: to illustrate it's not changing and we don't recreate this part of the array every time we call `getSteps`

docs/src/examples/components/Menu/commonSteps.ts
bmdalex: I'd argue it doesn't make it less readable either, so why not reuse. It's really NIT so I'd rather not change it

docs/src/examples/components/Menu/commonSteps.ts
layershifter: For me: it's not the right place for micro optimizations, but it will be much easier to read all steps in the single place. _NIT_

docs/src/examples/components/Menu/Variations/index.tsx
layershifter: Should mention that in #695 was done the great clean up in the same direction.

bmdalex: > We should show more in less examples, for example:
https://react.semantic-ui.com/collections/menu/#types-vertical

I don't really like that TBH, I have to open the example and looks what props have changed.

> May be we can even combine, `MenuExampleVertical` and `MenuExampleVerticalPointing` to a single example as it's done there:
https://stardust-ui.github.io/react/components/menu#types-pointing

That's a bit better but then I need to write different set of steps in `docs/src/examples/components/Menu/Variations/MenuExampleVerticalPointing.shorthand.steps.ts`

to select the 2nd menu + I know @miroslavstastny is not a fan of having multiple examples in one screenshot:
https://github.com/stardust-ui/react/issues/992#issuecomment-468228767

What about moving the title Vertical & Pointing to description and removing the title then?

CHANGELOG.md
bmdalex: exactly, we need more consistency here, removed

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: done, thx

docs/src/examples/components/Dropdown/Types/DropdownExampleSearchMultiple.shorthand.steps.ts
bmdalex: done, thx

docs/src/examples/components/Menu/Variations/index.tsx
layershifter: I'm not sure that `description` provides something useful in this case.

/cc @mnajdova @kuzhelov

docs/src/examples/components/Menu/Slots/MenuExampleIconOnly.shorthand.steps.ts
bmdalex: done, thanks

docs/src/examples/components/Menu/Slots/MenuExampleIconOnly.shorthand.steps.ts
bmdalex: done, thanks

docs/src/examples/components/Menu/commonSteps.ts
bmdalex: fair enough, let's put them in one place

docs/src/examples/components/Menu/Variations/index.tsx
bmdalex: ### 1. now
screenshot 2019-02-28 at 18 06 11

### 2. @layershifter proposal
screenshot 2019-02-28 at 18 06 40

### 3. compromise:
screenshot 2019-02-28 at 18 07 31

or a short description...

@layershifter 
@mnajdova 
@kuzhelov 
@miroslavstastny 

vote? 😕

mnajdova: I see the point that both of you has made... I would go with only title here, as we previously already describe what vertical and what pointing means. This way, the user will immediately know which props are provided. What do you think?

mnajdova: Also, let's keep consistency with the other examples, I know that the page for the Menu component is pretty long, so try to avoid adding more text if there is already description for something already.

mnajdova: Sorry for leaving so many components, I promise it's the last one :D We don't have to show all examples, if we are adding them just so we can have screen tests...

docs/src/examples/components/Menu/Slots/MenuExample.shorthand.tsx
```diff
@@ -7,6 +7,6 @@ const items = [
   { key: 'events', content: 'Upcoming Events' },
 ]
 
-const MenuExample = () => <Menu defaultActiveIndex={0} items={items} />
+const MenuExample = () => <Menu defaultActiveIndex={0} items={items} primary />
```
mnajdova, line 10: Why are we adding primary on all of the examples? I know that in Teams theme it is more used, but with this example we are laying of what we are showing (we say default Menu, when indeed we are showing the primary variations)

packages/react/src/components/Dropdown/DropdownSearchInput.tsx
```diff
@@ -131,6 +137,11 @@ class DropdownSearchInput extends UIComponent<ReactProps<DropdownSearchInputProp
   }
 }
 
+DropdownSearchInput.slotClassNames = {
```
mnajdova, line 140: 👍

docs/src/examples/components/Menu/commonScreenerSteps.ts
```diff
@@ -0,0 +1,25 @@
...
+interface StepsOptions {
+  vertical?: boolean
+}
+
+const selectors = {
+  menu: `.${Menu.className}`,
+  item: (itemIndex: number) => `.${Menu.className} li:nth-child(${itemIndex}) a`,
+}
+
+const getScreenerSteps = ({ vertical }: StepsOptions = {}) => [
```
mnajdova: I like that this logic is extracted and can be reuse 👍

docs/src/examples/components/Menu/Variations/index.tsx
bmdalex: > We don't have to show all examples, if we are adding them just so we can have screen tests...

@mnajdova good point, how can we do that?

docs/src/examples/components/Menu/Slots/MenuExample.shorthand.tsx
bmdalex: I know, good that you opened this discussion; I want to see screener regression for teams theme otherwise they are not caught at all, what's the alternative here?

docs/src/examples/components/Menu/Variations/index.tsx
layershifter: My vote is for "2".
In the future will be cool to make this area more compact if the example missing a `title` and `description`. Can this satisfy all?

image

mnajdova: @layershifter now that the title is not making anything worse (we still have the same height, I would add it) We may restructure this in future, maybe use some collapsable if we want to see all variations in vertical or something like that...
@Bugaa92 you can still add files in the examples folders, but you don't need to import them in the index.tsx files of the Variations/Types and actually show them on the docs...

docs/src/examples/components/Menu/Slots/MenuExample.shorthand.tsx
mnajdova: See my comment: https://github.com/stardust-ui/react/pull/983#discussion_r261301783

bmdalex: ### So it looks like we came again to the discussion about screener visual tests and using the examples as input:

For this PR I need more things to be tested than what we have examples for; for instance I need to test all menus with `primary` prop being supplied.

However, we might not want to change all examples to have `primary` prop (which is what I did in this PR) just so that screener will create screenshots for the primary menus.

### Ideal solution:

Best thing would be to move all screener related stuff from  `docs/src/examples/` somewhere under `tests/screener/` or `tests/screenshots`.

### Proposal:

For now I'm thinking of following @mnajdova 's suggestions and add example files that I'm not exporting in `index.tsx` so that they are not being shown on the docs site.

I'm also considering marking them with `.test.tsx` extension and write the steps in `test.stetps.ts`

e.g. this file will be reverted and we'll have this structure:
- `docs/src/examples/components/Menu/Slots/MenuExample.shorthand.tsx` -> this file reverted
- `docs/src/examples/components/Menu/Slots/MenuExamplePrimary.test.tsx` -> menu with `primary` prop used for screenshot tests but not exported in `index.tsx`
- `docs/src/examples/components/Menu/Slots/MenuExamplePrimary.steps.ts` -> steps for `MenuExamplePrimary.test.tsx`

It'd be great if we could add these steps directly in `MenuExamplePrimary.test.tsx` so that we don't create more files?

docs/src/examples/components/Menu/Variations/index.tsx
bmdalex: let's see what we decide here first: https://github.com/stardust-ui/react/pull/983#discussion_r261310975

layershifter (approved):

docs/src/examples/components/Dropdown/Types/DropdownExampleSearchMultiple.shorthand.steps.ts
```diff
@@ -1,23 +1,23 @@
...
-      .keys(`.${Input.slotClassNames.input}`, keys.leftArrow)
+      .click(selectors.toggleIndicator)
+      .click(selectors.item(2))
+      .click(selectors.toggleIndicator)
+      .click(selectors.item(2))
+      .keys(selectors.input, keys.leftArrow)
       .snapshot('Selects last selected element'),
-  (steps, keys) =>
-    steps
-      .click(`.${Dropdown.className} .${Indicator.className}`)
```
bmdalex, line 14: pretty sure that was a strange merge previously so I removed it, it was just repeating the steps above

docs/src/examples/components/Menu/Variations/index.tsx
bmdalex: added just title for now, as Marija said, it's not taking extra space and brings no harm

docs/src/examples/components/Menu/Slots/MenuExample.shorthand.tsx
bmdalex: extracted in #1000 and will address in another PR

docs/src/examples/components/Dropdown/Types/DropdownExampleSearchMultiple.shorthand.steps.ts
bmdalex: fixed

PR #949, opened 2019-02-22
fix(Popup): proper handling of clicks in nested Popups

layershifter:
Fixes #1013.

----

### Remove existing `EventStack`

Was completely replaced by `@stardust-ui/react-component-event-listener` & `@stardust-ui/react-component-nesting-registry`.

PR #985, opened 2019-02-27
fix(Popup): close previous popup on enter key

### LINKED ISSUE #866: Popup: previous popup is not closed on `Enter` key
## Bug Report

Previously opened `Popup` is not closed on `Enter` key click.

### Steps
1. Open https://stardust-ui.github.io/react/components/popup#variations-position
2. Click on any button
3. Press `Tab` and then `Enter` keys

### Expected Result
Only one popup is open.

### Actual Result
Two opened popups.

popup-enter

### Version
0.20.0

jongsue:
fixes #866 

I found the line that makes this bug

I tried to make the test case for this, but still failed it
even though it seems working fine
previous-popup-close

Popup's updateOutsideClickSubscription method has setTimeout
the test case runs before setTimeout callback

I hope someone can help me to make the test case pass


layershifter:
Actually, we should use `EventStack` now to handle this, but it's not fully correct. I have a prototype that solves this issue (https://codesandbox.io/s/9yw3vq6774) and will be implemented in #949.

For now going to push changes to get it working.

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -251,9 +254,29 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
...
           },
         )
+        this.outsideKeySubscription = EventStack.subscribe(
+          'keydown',
+          e => {
+            const keyCode = keyboardKey.getCode(e)
+
+            const matchingKey = keyCode === keyboardKey.Enter || keyboardKey.Spacebar
+            const isOutsidePopupElement =
+              this.popupDomElement && !this.popupDomElement.contains(e.target)
```
bmdalex: can you extract this logic from `updateOutsideClickSubscription` so we can reuse here?

bmdalex (approved):
thanks for fixing 👍 one comment
fyi @kuzhelov

jongsue:
I did extra method to reuse it as @Bugaa92 said

Thank you guys for the reviews
I've learned a lot of good points here

kuzhelov:
please, just update `Fixes` section of changelog before merge 👍

Issue #978, opened 2019-02-27
Popup is leaking unnecessary key events


jurokapsiar:
### Steps
Open the testcase below
Click on "Click here"
Click at the end of textarea in Popup
Press shift+home to select the text

### Expected Result
Whole text in the text area gets selected

### Actual Result
Active item in the menu changes to the first item

### Version
0.22.1

### Testcase
https://codesandbox.io/s/moyloyor9y

PR #987, opened 2019-02-27
fix(Popup): Do not propagate keydown events outside Popup's content

sophieH29:
Closes https://github.com/stardust-ui/react/issues/978

No need to propagate keydown events outside Popup, allow only keyboard actions to execute

### DIFF
--- a/packages/react/src/components/Popup/Popup.tsx
+++ b/packages/react/src/components/Popup/Popup.tsx
@@ -180,7 +180,6 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
   protected actionHandlers: AccessibilityActionHandlers = {
     closeAndFocusTrigger: e => {
       this.close(e, () => _.invoke(this.triggerFocusableDomElement, 'focus'))
-      e.stopPropagation()
     },
     close: e => this.close(e),
     toggle: e => {
@@ -439,8 +438,12 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
     const popupWrapperAttributes = {
       ...(rtl && { dir: 'rtl' }),
       ...accessibility.attributes.popup,
-      ...accessibility.keyHandlers.popup,
-
+      onKeyDown: (e: KeyboardEvent) => {
+        // don't need to propagate keydown events outside Popup
+        // allow only keyboard actions to execute
+        accessibility.keyHandlers.popup.onKeyDown(e)
+        e.stopPropagation()
+      },
       className: popupPositionClasses,
       style: popupPlacementStyles,
       ...this.getContentProps(),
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Popup/Popup.tsx
```diff
@@ -439,8 +438,12 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
     const popupWrapperAttributes = {
       ...(rtl && { dir: 'rtl' }),
       ...accessibility.attributes.popup,
-      ...accessibility.keyHandlers.popup,
-
+      onKeyDown: (e: KeyboardEvent) => {
+        // No need to propagate keydown events outside Popup
+        // allow only keyboard actions to execute
+        _.invoke(accessibility.keyHandlers.popup, 'onKeyDown', e)
+        e.stopPropagation()
```
bmdalex, line 445: is this testable? would be great to have a UT since we had problems with Popup

bmdalex (approved):
👍

packages/react/src/components/Popup/Popup.tsx
layershifter: Yep, you can do:
```
const stopPropagation = jest.fn()

wrapper.find('CONTENT').simulate('keyDown', { stopPropagation })
expect(stopPropagation).toHaveBeenCalledTimes(1)
```

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -439,8 +438,12 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
     const popupWrapperAttributes = {
       ...(rtl && { dir: 'rtl' }),
       ...accessibility.attributes.popup,
-      ...accessibility.keyHandlers.popup,
-
+      onKeyDown: (e: KeyboardEvent) => {
```
```suggestion
      onKeyDown: (e: React.KeyboardEvent) => {
```

layershifter (approved):

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -439,8 +438,12 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
     const popupWrapperAttributes = {
       ...(rtl && { dir: 'rtl' }),
       ...accessibility.attributes.popup,
-      ...accessibility.keyHandlers.popup,
-
+      onKeyDown: (e: React.KeyboardEvent) => {
```
mnajdova, line 441: I know it is too late now, but I was just wondering whether our suggestion for the users would have been override the `onKeyDown` for the Popup and stop propagating if you want? This seem too much specific to the client's needs that we had at that moment.. @layershifter @sophieH29

PR #989, opened 2019-02-28
feat(Box): optimize

kuzhelov:
- [x] no test regressions
- [x] no Screener regressions

## Approach
To prevent breaking changes, this PR has preserved original `Box` component's functionality _exported_ from Stardust. 

At the same time, lightweight version of `Box` component was introduced, all the common pieces between these two were reused and placed in the `common.ts` file. While these two components have the same API and, in most cases, are interchangeable, the reason previous version of `Box` was preserved is that lightweight `Box` doesn't support style customizations via theming.

While this `theme-based styles` support for `Box` is not used in Stardust anywhere, however, it might be used by Stardust clients. Eventually I might see it being organised the following way:
- there would be internal utility component in Stardust. It will be used as one of the basis building blocks (the same way `Box` was used), and will have minimal abstraction costs
- component with reacher functionality will be exported to clients - so that all the Stardust features (e.g. theme-based styling) will be supported but, at the same time, it would imply higher perf costs.

With changes made in this PR Stardust consumes light-weight `Box` component internally.

### List with 10 elements, median from 10x50 runs
- with unoptimized `Box` - 67ms
- **with optimized `Box` - 58ms (~13%)**
- *with `ItemLayout` - 57ms (~14%)

### List with 50 elements, median from 10x50runs
- with unoptimized `Box` - 235ms
- **with optimized `Box` - 200ms (~14%)**
- *with `ItemLayout` - 203ms (~13%)

## Opened questions
- primarily, naming
  - `Box` and `BoxHeavy` vs `BoxLight` and `Box`
  - different display names are suggested in the PR (`Box` and `BoxHeavy`), to make it clear from React Debug Tools that unoptimized Box is used - should we use the same instead for both?
  - same class name is suggested for both (to make these equal to static CSS) - should we use different ones?
- additional suggestions to optimize things while preserving correctness 👍 

## Areas for potential improvements
- optimize `Flex` rendering logic
  - optimize styles rendering - most probably we won't need theme mount, as `Flex` styles shouldn't be customized by theme
  - there is no need to calculate other Stardust bits, such as `accessibility`, etc
  - use `margins` instead of `Flex.Gap` element - will allow to gain ~5% of perf improvement
- generally optimize rendering logic of Stardust components 
  - do not calculate Stardust bits that are not necessary for element
  - *do not mount FelaTheme element each time, in some cases `theme` could be passed as prop


packages/react/src/components/Box/Box.tsx
```diff
@@ -1,49 +1,70 @@
...
+  }
+
+  return render({
+    ElementType,
+    props,
+    renderer: theme.renderer,
+  })
+}
+
+const handledProps = Object.keys(boxPropTypes)
```
bmdalex: maybe move to common? it took me a while to understand where it's used

packages/react/src/components/Box/Box.tsx
```diff
@@ -1,49 +1,70 @@
...
-        {...unhandledProps}
-        className={classes.root}
-      >
-        {childrenExist(children) ? children : content}
-      </ElementType>
+const Box: React.FunctionComponent<BoxProps> & { create: Function } = props => {
+  const { as: ElementType, theme } = props
+
+  if (!theme) {
+    return withThemeContext(theme =>
```
bmdalex: NIT: inline `withThemeContext`? it's only used once:

```jsx

if (theme) {
  return render({
    ElementType,
    props,
    renderer: theme.renderer,
  })
}

return (
  <FelaTheme>
    {theme => render({
      ElementType,
      props,
      renderer: theme.renderer,
    })}
  </FelaTheme>
)

File: packages/react/src/components/Box/Box.tsx [hunk 3]
```diff
@@ -1,49 +1,70 @@
...
+  props: BoxProps
+  renderer: FelaRenderer
+}
+
+const render = ({ ElementType, props, renderer }: BoxRenderConfig) => {
+  const { styles, className: predefinedClasses, content, children } = props
+
+  const unhandledProps = getUnhandledProps({ handledProps }, props)
+  const classes = {
+    root: renderer.renderRule(callable(styles), props),
```
bmdalex, line 21: I guess the main optimization is somewhere here, can you explain in a few words how it's achieved?

packages/react/src/components/Box/Box.tsx
```diff
@@ -1,49 +1,70 @@
...
-
-    return (
-      <ElementType
-        {...rtlTextContainer.getAttributes({ forElements: [children, content] })}
-        {...unhandledProps}
-        className={classes.root}
-      >
-        {childrenExist(children) ? children : content}
-      </ElementType>
+const Box: React.FunctionComponent<BoxProps> & { create: Function } = props => {
```
bmdalex, line 37: nit: ~~`React.FunctionComponent`~~ -> `React.FC`

packages/react/src/lib/getDisplayName.ts
```diff
@@ -0,0 +1,9 @@
+const getDisplayName = (Component: React.ReactType) => {
```
bmdalex: where is this method used?

packages/react/src/index.ts
```diff
@@ -16,7 +16,7 @@ export { default as Attachment, AttachmentProps } from './components/Attachment/
 
 export { default as Avatar, AvatarProps } from './components/Avatar/Avatar'
 
-export { default as Box, BoxProps } from './components/Box/Box'
+export { default as Box, BoxProps } from './components/Box/BoxHeavy'
```
bmdalex, line 19: why do we export `BoxHeavy` and what does it bring in addition to the new `Box` that we still need it?

packages/react/src/components/Box/Box.tsx
kuzhelov: this a merge artefact, thank you!

packages/react/src/lib/getDisplayName.ts
kuzhelov: this a merge artefact, thank you!

packages/react/src/index.ts
kuzhelov: the main reason is mentioned in the description:

> To prevent breaking changes, this PR has preserved original `Box` component's functionality _exported_ from Stardust.

Essentially, we would like to prevent Stardust from breaking changes to the clients that might already used theme-based styling approach to `Box`. Yes, while `Box` theme-based styles are not introduced anywhere in Stardust, it might be the case that client had already introduced some style overrides that target `Box` component at places of custom client's usage.

Consider the following example:

```jsx
// this might be an approach taken to circumvent current Stardust problem
// of positioning support
<Box variables={{ absolutePositioned: true }}>
   {someContent}
</Box>
```

with client's styles
```js
export const boxStyles = {
  root: ({ variables }) => ({ 
     variables.absolutePositioned && {
        position: absolute
    }
  })
```

Note that in that case it is always the case that client _won't like to target inner Stardust `Box` elements_ - the intent is to introduce styles to only those boxes that are used on client side. If we would export light version of `Box` that doesn't support theme-based styling, then breaking change will be introduced to the client. However, if we would export the same Box component as before, there won't be any breaking changes happen

packages/react/src/components/Box/Box.tsx
kuzhelov: would rather leave it as it is, due to its readability

packages/react/src/components/Box/Box.tsx
kuzhelov: note that here we imply that styles is an already prepared object, so that we don't need to
- consume theme styles, merge them
- merge theme variables
- resolve styles with merged variables

The thing we are requiring by optimised box is set of already resolved styles - and this is, essentially, the thing that we always have in Stardust code

packages/react/src/components/Box/Box.tsx
levithomason: Please avoid all abbreviations in the repo.  Code is for humans to read.  Let minifiers and uglifiers make it small and unreadable.

levithomason:
I think the solution here is too complex.  We don't want to make optimizations that involve forking or deviating individual components.  We also don't want to introduce any new APIs or ideas or require any refactoring in our optimizations. We should focus on removing complexity and special cases and put a hard stop on adding any special cases, like a light weight Box.

Improvements made should be first be focused on core pieces such as mergeThemes, factories, UIComponent, ProviderConsumer, etc. Areas from which all other abstractions are derived.  If these hot paths are optimized, every component will benefit drastically.

Example, just by `_.memoize()`ing the functions in `mergeThemes()` almost all component render times are reduced by around 75-95%:

https://github.com/stardust-ui/react/pull/1006
```
==============================
Button.perf.tsx
BEFORE  AFTER  SAVED  
42.83   2.89   93%    
43.53   1.57   96%    
10.57   1.11   89%    
86.06   19.54  77%    

==============================
Header.perf.tsx
BEFORE  AFTER  SAVED  
1.13    1.3    -15%   
0.69    0.66   4%     
0.45    0.46   -2%    
4.71    7.63   -62%   

==============================
Icon.perf.tsx
BEFORE  AFTER  SAVED  
8.34    1.33   84%    
8.69    0.56   94%    
3.27    0.38   88%    
16.25   7.99   51%    

==============================
Loader.perf.tsx
BEFORE  AFTER  SAVED  
25.84   1.86   93%    
31.44   1.14   96%    
5.93    0.82   86%    
48.51   11.24  77%    

==============================
HeaderDescription.perf.tsx
BEFORE  AFTER  SAVED  
7.35    2.25   69%    
5.86    1.3    78%    
2.53    0.89   65%    
15.79   16.41  -4%    

==============================
Attachment.perf.tsx
BEFORE  AFTER  SAVED  
35.33   2.47   93%    
26.54   1.44   95%    
7.94    1.04   87%    
68.63   11.54  83%    

==============================
Divider.perf.tsx
BEFORE  AFTER  SAVED  
8.26    1.18   86%    
8.42    0.58   93%    
3.27    0.41   87%    
16.49   11.22  32%    

==============================
List.perf.tsx
BEFORE  AFTER  SAVED  
68.27   68.78  -1%    
67.67   65.39  3%     
54.4    56.3   -3%    
98.21   99.29  -1%    

==============================
Chat.perf.tsx
BEFORE  AFTER  SAVED  
284.64  40.03  86%    
276.29  39.44  86%    
208.04  30.11  86%    
426.62  82.98  81% 
```

kuzhelov:
closing as this strategy but based on static API for React context was implemented for the entire Stardust library: #1163

PR #991, opened 2019-02-28
fix(Icon): Emoji & settings icon update

jaanus03:
Fixed the emoji icon filled version & added settings icon.

Current filled emoji icon:
screenshot 2019-02-28 at 12 01 00

New emoji icon:
screenshot 2019-02-28 at 12 08 50

New settings icon:
screenshot 2019-02-28 at 12 12 45


bmdalex (approved):
looks good, just add `exportedAs: 'settings',` to
`packages/react/src/themes/teams/components/Icon/svg/ProcessedIcons/icons-settings.tsx`

following #585

codepretty (approved):

packages/react/src/themes/teams/components/Icon/svg/ProcessedIcons/icons-settings.tsx
```diff
@@ -18,4 +18,5 @@ export default {
     </svg>
   ),
   styles: {},
+  exportedAs: 'settings',
```
codepretty, line 21: We might want to consider renaming this to 'cog'. We want to start moving to a generic name icons that match what the icon looks like and not use-specific naming that we currently do in Teams.

For example in Font Awesome it is called 'cog' - https://fontawesome.com/icons?d=gallery&q=cog

layershifter (approved):
Please add a changelog entry

PR #993, opened 2019-02-28
fix(Ref): update `node` if it was changed

### LINKED ISSUE #990: Ref: will not pass actual node on component update
## Bug Report

The `Ref` will not update actual node on component update.

### Steps
1. Open https://codesandbox.io/s/2om9k2ny7p
2. Click `Switch element`

### Expected Result
` "nodeName": "H1",`

### Actual Result
 "nodeName": "BUTTON", 💣

### Version
0.22.2

layershifter:
Fixes #990.


packages/react/src/components/Ref/RefFindNode.tsx
```diff
@@ -21,8 +21,21 @@ export default class RefFindNode extends React.Component<RefFindNodeProps> {
...
+    handleRef(this.props.innerRef, this.prevNode)
+  }
+
+  componentDidUpdate() {
+    const currentNode = ReactDOM.findDOMNode(this)
+
+    if (this.prevNode !== currentNode) {
+      this.prevNode = currentNode
+      handleRef(this.props.innerRef, currentNode)
+    }
```
layershifter, line 38: React calls an update only node was changed:
https://codesandbox.io/s/m4rmz4m3x

mnajdova (approved):

johannao76 (commented):
Thanks for fixing this so quickly!

PR #994, opened 2019-02-28
docs(ChatMessage): prototype of message reactions with Popup

mnajdova:
This PR adds prototype of how the reactions can be used with Popup, to display the users who reacted on the message. On hovering of the reaction, users can see the list which appears in Popup.

image

All prototypes related to the chat messages are now moved under one prototype:
image


docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -405,6 +405,13 @@ class Sidebar extends React.Component<any, any> {
         to: '/important-and-mention-messages',
         styles: menuItemStyles,
       },
+      {
+        key: 'message-reactions-with-popup',
```
bmdalex: sidebar becomes pretty full of prototypes, can we try grouping chat related prototypes by context the same we are doing for:

- dropdowns: http://localhost:8080/prototype-dropdowns
- popups: http://localhost:8080/prototype-popups

bmdalex (approved):
code looks good 👍  but I have to say the design is really ugly... 😿 
one comment

docs/src/components/Sidebar/Sidebar.tsx
mnajdova: Will do, for design yep, we decided because of accessibility reasons to replace the List with Menu, so it is ugly at this moment. Will address it, thanks!

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -227,7 +231,7 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
         {this.state.open &&
           Popup.isBrowserContext &&
           popupContent &&
-          ReactDOM.createPortal(popupContent, document.body)}
+          (inline ? popupContent : ReactDOM.createPortal(popupContent, document.body))}
```
layershifter: Can we move this change to a separate PR? And cover it with UT

mnajdova: Sure thing, I've even added new FocusZone component.. All of these component related changes will be moved to separate PRs, once the accessibility team confirms that this work as expected.

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -0,0 +1,245 @@
...
+    this.props.onFocusCapture && this.props.onFocusCapture(ev)
+    if (ev.target !== ev.currentTarget) {
+      // every time focus changes within the trap zone, remember the focused element so that
+      // it can be restored if focus leaves the pane and returns via keystroke (i.e. via a call to this.focus(true))
+      this._previouslyFocusedElementInTrapZone = ev.target as HTMLElement
+    }
+  }
+
+  private _onKeyboardHandler = (ev: React.KeyboardEvent<HTMLDivElement>): void => {
+    this.props.onKeyDown && this.props.onKeyDown(ev)
```
sophieH29: We might not need this,  as `AutoFocusZone` should just focus inner element on mount, it should be still possible to TAB out from it

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -0,0 +1,245 @@
...
+      ev.preventDefault()
+      ev.stopPropagation()
+    } else if (!ev.shiftKey && _lastTabbableChild === ev.target) {
+      focusAsync(_firstTabbableChild)
+      ev.preventDefault()
+      ev.stopPropagation()
+    }
+  }
+
+  private _forceFocusInTrap = (ev: Event, triggeredElement: HTMLElement) => {
```
sophieH29: We might not need this,  as `AutoFocusZone` should just focus inner element on mount, it should be still possible to TAB out from it

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -0,0 +1,245 @@
...
+    }
+  }
+
+  private _handleOutsideFocus = (ev: FocusEvent): void => {
+    const focusedElement = document.activeElement as HTMLElement
+    focusedElement && this._forceFocusInTrap(ev, focusedElement)
+  }
+
+  private _handleOutsideClick = (ev: MouseEvent): void => {
+    const clickedElement = ev.target as HTMLElement
```
sophieH29: We might not need this,  as `AutoFocusZone` should just focus inner element on mount, it should be still possible to TAB out from it

docs/src/prototypes/MessageReactionsWithPopup/ReactionPopup.tsx
```diff
@@ -0,0 +1,77 @@
...
+        }
+        inline
+        content={{
+          content: (
+            <Ref
+              innerRef={(popupNode: HTMLElement) => {
+                this.popup = popupNode
+              }}
+            >
+              <AutoFocusZone firstFocusableSelector={'ui-menu__item:first-child'}>
```
sophieH29: by default it will try to focus first focusable element in zone, so it should work without specifying `firstFocusableSelector` of first menu item

docs/src/prototypes/MessageReactionsWithPopup/ReactionPopup.tsx
```diff
@@ -0,0 +1,77 @@
...
+                <Menu
+                  items={['Jane Doe', 'John Doe']}
+                  vertical
+                  data-is-focusable={true}
+                  variables={{ borderColor: 'transparent' }}
+                  onKeyDown={this.handleKeyDownOnMenu}
+                />
+              </AutoFocusZone>
```
sophieH29: I think better to add `AutoFocusZone` through accessibility behavior (as we done with `FocusTrapZone`). Because if user use `popupFocusTrapBehvaior` and `AutoFocusZone`, there will be 2 zones rendered with overlapping logic. We should limit user's actions to using either `FocusTrapZone` or `AutoFocusZone`

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -0,0 +1,245 @@
...
+  focusAsync,
+} from './focusUtilities'
+
+import { AutoFocusZoneProps } from './AutoFocusZone.types'
+import getUnhandledProps from '../../getUnhandledProps'
+import getElementType from '../../getElementType'
+import * as customPropTypes from '../../customPropTypes'
+
+/** AutoFocusZone is used to focus the first element inside it's children */
+export class AutoFocusZone extends React.Component<AutoFocusZoneProps, {}> {
```
sophieH29: `AutoFocusZone` has common logic with `FocusTrapZone` - focusing inner element on componentDidMount(). `FocusTrapZone` can extend `AutoFocusZone`'s props and reuse that logic. So basically we'll get
```
<FocusTrapZone>
   <AutoFocusZone>
   </AutoFocusZone>
</FocusTrapZone>
```

Note: we need only one `<div>` rendered

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -0,0 +1,113 @@
...
+      // focus on the last item that had focus in the zone before we left the zone
+      focusAsync(this._previouslyFocusedElementInTrapZone)
+      return
+    }
+
+    const focusSelector =
+      firstFocusableSelector &&
+      (typeof firstFocusableSelector === 'string'
+        ? firstFocusableSelector
+        : firstFocusableSelector())
```
layershifter:
```diff
-   const focusSelector =
-      firstFocusableSelector &&
-      (typeof firstFocusableSelector === 'string'
-        ? firstFocusableSelector
-        : firstFocusableSelector())
+ const focusSelector = callable(firstFocusableSelector)
```

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -0,0 +1,113 @@
...
+      <ElementType
+        {...unhandledProps}
+        className={className}
+        ref={this.createRef}
+        aria-labelledby={ariaLabelledBy}
+        onFocusCapture={this._onFocusCapture}
+      >
+        {this.props.children}
+      </ElementType>
+    )
```
layershifter:
```diff
-      <ElementType
-        {...unhandledProps}
-        className={className}
-        ref={this.createRef}
-        aria-labelledby={ariaLabelledBy}
-        onFocusCapture={this._onFocusCapture}
-      >
-        {this.props.children}
-      </ElementType>
-    )
+<Ref innerRef={this.elementRef}>
+      <ElementType
+        {...unhandledProps}
+        className={className}
+        aria-labelledby={ariaLabelledBy}
+        onFocusCapture={this._onFocusCapture}
+      >
+        {this.props.children}
+      </ElementType>
+    )
+</Ref>
```

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+import { getNextElement, focusAsync } from './focusUtilities'
+
+import { AutoFocusZoneProps } from './AutoFocusZone.types'
+import getUnhandledProps from '../../getUnhandledProps'
+import getElementType from '../../getElementType'
+import * as customPropTypes from '../../customPropTypes'
+
+/** AutoFocusZone is used to focus the first element inside it's children */
+export class AutoFocusZone extends React.Component<AutoFocusZoneProps, {}> {
```
layershifter:
```suggestion
export class AutoFocusZone extends React.Component<AutoFocusZoneProps> {
```

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -0,0 +1,113 @@
...
+  private _root: { current: HTMLElement | null } = { current: null }
+  private _previouslyFocusedElementOutsideTrapZone: HTMLElement
+  private _previouslyFocusedElementInTrapZone?: HTMLElement
+
+  private createRef = elem => (this._root.current = ReactDOM.findDOMNode(elem) as HTMLElement)
+
+  static propTypes = {
+    as: customPropTypes.as,
+    className: PropTypes.string,
+    ariaLabelledBy: PropTypes.string,
```
layershifter: `ariaLabelledBy` and `className` should be unhandled props

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -0,0 +1,113 @@
...
+    firstFocusableSelector: PropTypes.string,
+    disableFirstFocus: PropTypes.bool,
+    focusPreviouslyFocusedInnerElement: PropTypes.bool,
+  }
+
+  static defaultProps: AutoFocusZoneProps = {
+    as: 'div',
```
layershifter: Doesn't make any sense, `as` defaults to `div`

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -20,51 +12,34 @@ import * as customPropTypes from '../../customPropTypes'
 
 /** AutoFocusZone is used to focus the first element inside it's children */
 export class AutoFocusZone extends React.Component<AutoFocusZoneProps, {}> {
-  private static _focusStack: AutoFocusZone[] = []
   private _root: { current: HTMLElement | null } = { current: null }
   private _previouslyFocusedElementOutsideTrapZone: HTMLElement
```
sophieH29: `_previouslyFocusedElementOutsideTrapZone ` it is not set anywhere. In FTZ, it is set usually to document.active element, so on `componentWillUnmount` we can focus it, basically does same as in Popup's action - `closeAndFocusTrigger`. I don't think we need it here.

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
mnajdova: Removed from here.

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
mnajdova: Removed from here.

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
mnajdova: Can you check the latest changes? `FocusTrapZone`, now returns in the render method `AutoFocusZone`. Is that what you mean with this comment?

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
mnajdova: This is not `UIComponent`, that's why we have this as part of `defaultProps`.

layershifter: https://github.com/stardust-ui/react/blob/master/packages/react/src/lib/getElementType.tsx#L41
It handled in `getElementType()`

mnajdova: Yep, I forgot we are directly using it. Thanks!

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
mnajdova: I removed it, but I am a bit concern, as this may introduce some regressions. Previously we had the following code:
```
  public componentDidMount(): void {
    const { disableFirstFocus = false } = this.props

    if (
      !this._root.current.contains(this._previouslyFocusedElementOutsideTrapZone) &&
      !disableFirstFocus
    ) {
      this._findElementAndFocusAsync()
    }
  }
```
Now we don't have the fist condition here...

sophieH29: yep, if look in FTZ `_previouslyFocusedElementOutsideTrapZone` is used in many places and is calculated there. Here it doesn't make sense. But we need somehow to respect this condition when `FocusTrapZone` will use `AutoFocusZone`

docs/src/prototypes/MessageReactionsWithPopup/ReactionPopup.tsx
mnajdova: I can see that we have added `focusTrap` as a prop to the behavior, but I cannot see how, depending on this prop we are providing the `FocusTrapZone`. I can see that in `renderComponent` we are handling the `focusZone` prop, but I can't find anything regarding the `focusTrap`.

mnajdova: I added this i the Popup.. I don't like that we are handling it there, but no other idea at the moment..

docs/src/prototypes/MessageReactionsWithPopup/ReactionPopup.tsx
mnajdova: You are right, I removed this selector

docs/src/prototypes/MessageReactionsWithPopup/ReactionPopup.tsx
```diff
@@ -0,0 +1,80 @@
...
+}
+
+class ReactionPopup extends React.Component<any, any> {
+  state = {
+    open: false,
+  }
+
+  reactionNode: HTMLButtonElement
+
+  popup: HTMLElement
```
layershifter: Cruft?

docs/src/prototypes/MessageReactionsWithPopup/ReactionPopup.tsx
```diff
@@ -0,0 +1,80 @@
...
+  }
+  return `${numberOfPersons} people reacted this message with a ${emojiType} emoji. Open menu to see people who reacted.`
+}
+
+class ReactionPopup extends React.Component<any, any> {
+  state = {
+    open: false,
+  }
+
+  reactionNode: HTMLButtonElement
```
layershifter: You can use `createRef()` there

docs/src/prototypes/MessageReactionsWithPopup/ReactionPopup.tsx
```diff
@@ -0,0 +1,80 @@
...
+      // Imported from the MenuButton prototype
+      focusNearest(this.reactionNode, e.shiftKey ? 'previous' : 'next')
+    }
+  }
+
+  handleOpenChange = (e, { open }) => {
+    this.setState({
+      open,
+    })
+  }
```
layershifter: Can be one-liner 😸

docs/src/prototypes/MessageReactionsWithPopup/ReactionPopup.tsx
```diff
@@ -0,0 +1,80 @@
...
+                  vertical
+                  variables={{ borderColor: 'transparent' }}
+                  onKeyDown={this.handleKeyDownOnMenu}
+                />
+              </Ref>
+            ),
+          }}
+          open={this.state.open}
+          onOpenChange={this.handleOpenChange}
+          on={'hover'}
```
layershifter:
```suggestion
          on='hover'
```

docs/src/prototypes/MessageReactionsWithPopup/index.tsx
```diff
@@ -0,0 +1,80 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+import { Avatar, Chat } from '@stardust-ui/react'
+import ReactionPopup from './ReactionPopup'
+
+const reactions = [{ icon: 'thumbs up', content: '1K' }, { icon: 'thumbs down', content: 2 }]
+const reactionsWithPopup = _.map(reactions, reaction => render =>
+  render(reaction, (Component, props) => <ReactionPopup Component={Component} props={props} />),
+)
```
layershifter: Do we need `reactionsWithPopup()`? Can we just use `{ as: ReactionPopup, icon: 'thumbs up', content: '1K' }`?

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -451,11 +455,17 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
       ...popupWrapperAttributes,
     } as FocusTrapZoneProps
 
+    const autoFocusProps = {
+      ...(typeof accessibility.autoFocus === 'boolean' ? {} : accessibility.autoFocus),
+      ...popupWrapperAttributes,
+    } as FocusTrapZoneProps
```
layershifter: Can be this part of behavior? We have there access to props

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { AutoFocusZoneProps } from './AutoFocusZone.types'
+import getUnhandledProps from '../../getUnhandledProps'
+import getElementType from '../../getElementType'
+import * as customPropTypes from '../../customPropTypes'
+import { callable } from '../../index'
+import Ref from '../../../components/Ref/Ref'
+
+/** AutoFocusZone is used to focus the first element inside it's children */
+export class AutoFocusZone extends React.Component<AutoFocusZoneProps, {}> {
+  private _root = React.createRef<HTMLElement>()
```
layershifter:
```suggestion
  private root = React.createRef<HTMLElement>()
```

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -0,0 +1,90 @@
...
+  private _previouslyFocusedElementInTrapZone?: HTMLElement
```
layershifter:
```suggestion
  private previouslyFocusedElementInTrapZone?: HTMLElement
```

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -0,0 +1,90 @@
...
+    firstFocusableSelector: PropTypes.string,
+    disableFirstFocus: PropTypes.bool,
+    focusPreviouslyFocusedInnerElement: PropTypes.bool,
+  }
+
+  public componentDidMount(): void {
+    const { disableFirstFocus = false } = this.props
+
+    if (!disableFirstFocus) {
+      this._findElementAndFocusAsync()
```
layershifter:
```suggestion
      this.findElementAndFocusAsync()
```

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -0,0 +1,90 @@
...
+      { handledProps: [..._.keys(AutoFocusZone.propTypes)] },
+      this.props,
+    )
+    // TODO: Remove `as` there after the issue will be resolved:
+    // https://github.com/Microsoft/TypeScript/issues/28768
+    const ElementType = getElementType({}, this.props) as React.ComponentClass<AutoFocusZoneProps>
+
+    return (
+      <Ref innerRef={this._root}>
+        <ElementType {...unhandledProps} onFocusCapture={this._onFocusCapture}>
```
layershifter:
```suggestion
        <ElementType {...unhandledProps} onFocusCapture={this.onFocusCapture}>
```

packages/react/src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -57,46 +53,38 @@ export class FocusTrapZone extends React.Component<FocusTrapZoneProps, {}> {
...
-    }
-
     this._hideContentFromAccessibilityTree()
     this._subscribeToEvents()
   }
 
   public render(): JSX.Element {
-    const { className, ariaLabelledBy } = this.props
+    const {
+      as,
```
layershifter: It seems that `as` is unhandled prop in this case and can be removed from `propTypes`

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
sophieH29: yes 👍

docs/src/prototypes/MessageReactionsWithPopup/ReactionPopup.tsx
mnajdova: Yep, removed

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -62,6 +62,9 @@ export interface PopupProps
   /** Alignment for the popup. */
   align?: Alignment
 
+  /** If this prop is used, the Popup will propagate the key down event, otherwise it won't. */
+  allowKeyDownPropagation?: boolean
```
mnajdova: See my comment: https://github.com/stardust-ui/react/pull/987/files#r263011991 No better proposal for now..

PR #995, opened 2019-02-28
feat(Popup): Set aria-modal for both Dialog and Popup with focus trap

sophieH29:
Both `Dialog` and `Popup` with focus trap behavior, should have set `aria-modal=true`

### DIFF
--- a/packages/react/src/lib/accessibility/Behaviors/Popup/popupFocusTrapBehavior.ts
+++ b/packages/react/src/lib/accessibility/Behaviors/Popup/popupFocusTrapBehavior.ts
@@ -13,7 +13,8 @@ import popupBehavior from './popupBehavior'
  */
 const popupFocusTrapBehavior: Accessibility = (props: any) => {
   const behaviorData = popupBehavior(props)
-  behaviorData.attributes.popup = {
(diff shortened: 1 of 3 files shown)


bmdalex (approved):
👍

packages/react/src/lib/accessibility/Behaviors/Popup/popupFocusTrapBehavior.ts
```diff
@@ -8,11 +8,17 @@ import popupBehavior from './popupBehavior'
...
+ * Adds attribute 'aria-modal=true' to 'popup' component's part.
  * Traps focus inside component.
  */
-const popupFocusTrapBehavior: Accessibility = (props: any) => ({
-  ...popupBehavior(props),
-  focusTrap: true,
-})
+const popupFocusTrapBehavior: Accessibility = (props: any) => {
+  const behaviorData = popupBehavior(props)
+  behaviorData.attributes.popup = {
```
layershifter, line 16: If we will define `attributes.popup` in `popupBehavior()` we will get a 💣 

```suggestion
  behaviorData.attributes.popup = behaviorData.attributes.popup || {}
  behaviorData.attributes.popup['aria-modal'] = true
```

layershifter (approved):

PR #996, opened 2019-02-28
chore(shorthands): allow arrays as shorthand value

### LINKED ISSUE #984: Improve shorthands for component that contains collection shorthands
### Problem description
When we are using components that contains collection of other components (like `Menu`, or some group components - `ButtonGroup`), we should provide different processing of the shorthand if an array is provided. For example, let's take a look on the `actionMenu` prop in the `ChatMessage`:

```jsx
<ChatMessage 
  author='John Doe' 
  timestamp='27/2' 
  actionMenu={{items: ['like', 'dislike']}} 
  content='Content of the message'
/>
```

This is not really user friendly API, as the `actionMenu` should probably be referred as `actions`. Moreover the user has to create an object with the `items` as a key to finally define the actions they want to have on the `ChatMessage`.

### Proposed solution
Much clear API for achieving the same thing, would be the following:

```jsx
<ChatMessage 
  author='John Doe' 
  timestamp='27/2' 
  actions={['like', 'dislike']}
  content='Content of the message'
/>
```

With this the user should be able to define the same thing as above. Furthermore, if they want to customize the menu, they can still use the same old object for defining the `Menu`:

```jsx
<ChatMessage 
  author='John Doe' 
  timestamp='27/2' 
  actions={{items: ['like', 'dislike'], vertical: true}}
  content='Content of the message'
/>
```

For this to work, we may add additional parameter to the `createShorthandFactory`, that will handle arrays, the same way as we currently handle the string literals. The signature would be something like this:

```jsx
Menu.create = createShorthandFactory(
  Menu, 
  'children', /* what is the default prop when string is passed */
  'items' /* what is the default prop when array is passed */
)
ButtonGroup.create = createShorthandFactory(
  ButtonGroup, 
  'content', /* what is the default prop when string is passed */ 
  'buttons' /* what is the default prop when array is passed */
)
```

mnajdova:
This PR fix https://github.com/stardust-ui/react/issues/984 by allowing arrays to be used together with the shortand props which are representing collection.


docs/src/examples/components/Chat/Content/ChatExampleReactionGroup.shorthand.tsx
```diff
@@ -10,7 +10,7 @@ const items = [
     message: {
       content: (
         <Chat.Message
-          reactionGroup={{ items: reactions }}
+          reactions={reactions}
```
layershifter: Breaking change?

packages/react/src/lib/factories.ts
```diff
@@ -164,6 +178,13 @@ function createShorthandFromValue(
     props[mappedProps[overrideProps.as || defaultProps.as] || mappedProp || 'children'] = value
   }
 
+  // Map prop for array value
+  if (valIsArrayObject) {
+    props[
+      mappedProps[overrideProps.as || defaultProps.as] || mappedPropAsArray || 'children'
```
layershifter: We have partially the same condition on 178 line. Can we extract `mappedProps[overrideProps.as || defaultProps.as]` to something meaningful?

layershifter:
I like the idea with object config, we can apply it to other factories (`createShorthand`, `createShorthandFromValue`), too.

layershifter:
> This PR solves Issue #984

It's not enough to close the matching issue 😸 
https://help.github.com/en/articles/closing-issues-using-keywords

docs/src/examples/components/Chat/Content/ChatExampleReactionGroup.shorthand.tsx
mnajdova: I added this as a showcase... Maybe will create different PR for changing this later...

kuzhelov: yes, lets introduce changes to simplify all the related component APIs as a separate PR - however, probably, it is better to do it sooner than later :)

packages/react/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -90,6 +90,6 @@ class AccordionTitle extends UIComponent<ReactProps<AccordionTitleProps>, any> {
   }
 }
 
-AccordionTitle.create = createShorthandFactory(AccordionTitle, 'content')
+AccordionTitle.create = createShorthandFactory({ Component: AccordionTitle, mappedProp: 'content' })
```
kuzhelov, line 93: 👍

docs/src/examples/components/Chat/Content/ChatExampleReactionGroup.shorthand.tsx
mnajdova: Will revert this changes from this PR, is getting pretty big as it is..

packages/react/src/lib/factories.ts
```diff
@@ -46,70 +46,101 @@ const mappedProps: { [key in HTMLTag]: ShorthandProp } = {
...
+  Component,
+  mappedProp,
+  mappedArrayProp,
+  valueOrRenderCallback,
+  options: optionsPram,
+}: {
+  Component: React.ReactType
+  mappedProp?: string
+  mappedArrayProp?: string
+  valueOrRenderCallback?: ShorthandValue | ShorthandRenderCallback
```
mnajdova, line 59: Not sure about this name here... Would rather have value, as it is much simpler and shorter...

packages/react/src/lib/factories.ts
```diff
@@ -46,70 +46,101 @@ const mappedProps: { [key in HTMLTag]: ShorthandProp } = {
...
+  options?: CreateShorthandOptions
+}): React.ReactElement<Props> | null | undefined {
+  const options = optionsPram || CREATE_SHORTHAND_DEFAULT_OPTIONS
```
layershifter:
```diff
- options: optionsPram,
+ options = CREATE_SHORTHAND_DEFAULT_OPTIONS,
```

```suggestion
```

layershifter (approved):

packages/react/src/lib/factories.ts
```diff
@@ -119,19 +149,21 @@ function createShorthandFromValue(
 
   const valIsPrimitive = typeof value === 'string' || typeof value === 'number'
   const valIsPropsObject = _.isPlainObject(value)
+  const valIsArrayObject = _.isArray(value)
```
miroslavstastny: what is `ArrayObject`?

mnajdova: Will change it to `valueIsArray` :)

PR #997, opened 2019-02-28
fix(test): remove hardcoded magic numbers

kuzhelov:
These changes remove magic numbers that were used in tests to pass over 'service/helper' elements (such as `ThemeProvider`, `FelaTheme`, etc) during child tree traversal. As a result, the logic of conformance tests was made to be more robust and clear. Should note in addition, that corresponding comments over the sections with magic numbers were already obsolete and had not reflected real state of things.

As an additional benefit, these changes would guarantee that all the performance optimizations we might consider in future (such as not rendering `FelaTheme` in case if it is not necessary) won't affect test's logic anyhow, so we wouldn't need to switch back and forth and make yet another adjustments to magic numbers used.


packages/react/src/lib/getDisplayName.ts
```diff
@@ -0,0 +1,9 @@
+const getDisplayName = (Component: React.ReactType) => {
+  return (
+    (Component as React.ComponentType).displayName ||
+    (Component as React.ComponentType).name ||
+    (typeof Component === 'string' && Component.length > 0 ? Component : 'Unknown')
+  )
+}
+
+export default getDisplayName
```
layershifter: I will be there again with this link: https://github.com/gaearon/react-hot-loader#checking-element-types
We can't rely on `displayName`, so my proposal: let's move this util to `tests` directory to avoid accidental usages in component's code.

packages/react/test/specs/commonTests/isConformant.tsx
```diff
@@ -4,6 +4,13 @@ import * as React from 'react'
 import { ReactWrapper } from 'enzyme'
 import * as ReactDOMServer from 'react-dom/server'
 
+import { ThemeProvider, FelaTheme } from 'react-fela'
+
+import Ref from 'src/components/Ref/Ref'
+import RefFindNode from 'src/components/Ref/RefFindNode'
+import Box from 'src/components/Box/Box'
```
layershifter:
```suggestion
import { Box, Ref, RefFindNode } from '@stardust-ui/react'
```

packages/react/test/specs/commonTests/isConformant.tsx
```diff
@@ -53,40 +58,40 @@ export default (Component, options: Conformant = {}) => {
...
+    ...[ThemeProvider, FelaTheme, Ref, RefFindNode],
+    ...(usesWrapperSlot && [Box]),
+  ].map(getDisplayName)
+
+  const toNextNonTrivialChild = (from: ReactWrapper) => {
+    let current = from.childAt(0)
+
+    if (!current) return current
+
+    while (helperComponentNames.indexOf(current.name()) >= 0) {
```
layershifter: nit: may be recursion in this case will be more readable?

packages/react/test/specs/commonTests/isConformant.tsx
```diff
@@ -53,40 +58,40 @@ export default (Component, options: Conformant = {}) => {
 
   const componentType = typeof Component
 
-  // This is added because the component is mounted
+  const helperComponentNames = [
+    ...[ThemeProvider, FelaTheme, Ref, RefFindNode],
+    ...(usesWrapperSlot && [Box]),
```
mnajdova: A bit concern with this... I know currently we use the Box for wrapper, but maybe in the config we can provide which Component is used as Wrapper, having default to be Box.

packages/react/test/specs/commonTests/isConformant.tsx
```diff
@@ -53,40 +58,40 @@ export default (Component, options: Conformant = {}) => {
...
+  ].map(getDisplayName)
+
+  const toNextNonTrivialChild = (from: ReactWrapper) => {
```
layershifter, line 65: It will be great to extract this code outside the `isConformant()` function.

packages/react/test/specs/commonTests/isConformant.tsx
```diff
@@ -53,40 +58,40 @@ export default (Component, options: Conformant = {}) => {
...
+    let topLevelChildElement = toNextNonTrivialChild(componentElement)
 
-    if (component.type() === FocusZone) {
+    // passing through Focus Zone wrappers
+    if (topLevelChildElement.type() === FocusZone) {
       // another HOC component is added: FocuZone
-      component = component.childAt(0) // skip through <FocusZone>
-      if (component.prop(FOCUSZONE_WRAP_ATTRIBUTE)) {
-        component = component.childAt(0) // skip the additional wrap <div> of the FocusZone
+      topLevelChildElement = topLevelChildElement.childAt(0) // skip through <FocusZone>
```
mnajdova, line 82: Is there a reason why FocusZone is not in the `helperComponentsNames`?

kuzhelov: yes - just because the logic itself is a bit different for the focus zone. Wanted to encapsulate the most trivial logic of skipping `service` components - for focus zone we have different behavior, however. And while we might introduce something like `skippedProps` to traversal method's config, it will then obfuscate the component type for which this config is necessary (like, these checks are necessary for FocusZone-related components only - note that currently this 'FocusZone-related' is defined by only _when_ method is called, and not by the type/name of the component)

packages/react/test/specs/commonTests/isConformant.tsx
kuzhelov: lets try that :)

packages/react/test/specs/commonTests/isConformant.tsx
kuzhelov: would rather like to leave this logic here, so that all the bits related to traversal logic would be kept at the same module. Is there any place where we need this logic as well now?

packages/react/src/lib/getDisplayName.ts
kuzhelov: sorry, just not sure why this won't work? yes, while types are not strictly equal (type vs proxied wrapper), there should be still the same set of React-specific static props preserved, such as `displayName`  (otherwise it won't be possible for React Debugger Tools to show display name). Don't get me wrong: I am fine to put this function to tests dir, but just don't see the problems with hot loader (have checked that locally) - am I missing something?

packages/react/test/specs/commonTests/isConformant.tsx
layershifter: Or may be the whole `getComponent()`... To concentrate `isConformant()` on assertions

packages/react/test/specs/commonTests/isConformant.tsx
kuzhelov: have done it intentionally, as some of the components (like `RefFindNode`) might be private ones. Also, this way of importing components is consistent with the imports of used in `*-test` files - and this approach conveys the fact that we might want to use internal Stardust components in tests

packages/react/test/specs/commonTests/isConformant.tsx
kuzhelov: would rather like to wait with segregating this logic, as cannot see any other place where it might be useful now. Yes, agree with the fact that it will make `isConformant` module to be focused on assertions, but, at the same time, it will introduce a module with the functionality that currently suits and specific for the only client (`isConformant`). 

What if we would wait with this split a bit when there will be more appealing reason? Another point is that we are not making things any worse with that, as this logic was always part of conformance tests, and there were no other module so far interested in this functionality

packages/react/src/lib/getDisplayName.ts
layershifter: > here should be still the same set of React-specific static props preserved
Yep, you're correct. 

https://github.com/Semantic-Org/Semantic-UI-React/pull/382#discussion_r75185603

May be this will be more correct case. I want to prevent a case when someone will rely on `
displayName` in production code.

See example: https://codesandbox.io/s/l77xq5y8nq

mnajdova (approved):

packages/react/src/lib/getDisplayName.ts
kuzhelov: lets prevent this - in either case, totally agree that we won't like this function to be used in prod code 👍

layershifter (approved):

PR #1002, opened 2019-03-01
feat(Icon): add `outline` prop

### LINKED ISSUE #793: Icon variable `outline` should be a prop
### Problem description
Setting outline variable on an Icon is common enough that I think it would make more sense and be a much simpler design to have outline be a prop instead of variable.
```
<Icon name="call-video" variables={{ 'outline': true }} />
```

### Proposed solution
A much simpler way to describe an Icon that shows its filled state would be...

```
<Icon name="call-video" outline />
```

layershifter:
Fixes #793.

# BREAKING CHANGES

This PR removes the `iconVariables.outline` variable and introduces the `outline` prop.

### Before 

```jsx
<Icon name="calendar" variables={{ outline: true }} />
```

### After

```jsx
<Icon name="calendar" outline />
```

### DIFF
--- a/docs/src/prototypes/IconViewer/index.tsx
+++ b/docs/src/prototypes/IconViewer/index.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react'
 import { Provider, Grid, Divider, Header, Icon, Menu, Segment } from '@stardust-ui/react'
+import CopyToClipboard from 'docs/src/components/CopyToClipboard'
 import themeWithProcessedIcons from 'src/themes/teams/withProcessedIcons'
 import { TeamsProcessedSvgIconSpec } from 'src/themes/teams/components/Icon/svg/types'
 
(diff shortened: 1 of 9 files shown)


docs/src/prototypes/IconViewer/index.tsx
```diff
@@ -94,7 +94,7 @@ class IconViewerExample extends React.Component<any, {}> {
                         .sort()
                         .map(name => (
                           <div key={`${name}-outline`} style={cellStyles}>
-                            <Icon name={name} variables={{ outline: true }} />
+                            <Icon name={name} outline />
                             <br />
                             <code>{name.replace(processedIconsNamePrefix, '')} outline</code>
```
mnajdova, line 111: I don't  like that here we are kind a suggesting that the outline is part of the name :\ Not related to your changes, but would rather see something like `<Icon name=`${name.replace(processedIconsNamePrefix, '')}` outline />` or some object `{name: `${name.replace(processedIconsNamePrefix, '')}`, outline: true}`, because I was asked multiple times by people about what the name `like outline` for example for the Icon's doesn't work. It is not obvious that the outline is something else...

mnajdova (approved):
LGTM

docs/src/prototypes/IconViewer/index.tsx
layershifter: copy

Added copy 😺

docs/src/prototypes/IconViewer/index.tsx
```diff
@@ -13,9 +14,20 @@ const renderStardustIconName = (icon, isOutline = false) => {
...
+          => {maybeExportedAs} {isOutline && 'outline'}
+        </code>
+        <br />
+        <CopyToClipboard
+          render={(active, onClick) => (
+            <button onClick={onClick} style={{ fontSize: 10 }} title="Copy usage">
+              {active ? '✔' : 'Copy'}
+            </button>
+          )}
+          value={`<Icon name="${maybeExportedAs}" ${isOutline ? 'outline' : ''} />`}
```
kuzhelov, line 28: 👍

kuzhelov (approved):

PR #1004, opened 2019-03-01
fix(dropdown): applied roving tabindex technique to selected dropdown items

silviuaavram:
Without it, when in a Popup and a selected item is focused, shift tab will make focus escape the Popup, as FocusTrapZone will not handle it as first focusable item in the Popup (due to the lack of tabindex="0").

Also it does not hurt to be able to come back to the selected item via tab/shift tab once you tabbed/shift tabbed away.

### DIFF
--- a/packages/react/src/components/Dropdown/DropdownSelectedItem.tsx
+++ b/packages/react/src/components/Dropdown/DropdownSelectedItem.tsx
@@ -68,7 +68,7 @@ class DropdownSelectedItem extends UIComponent<ReactProps<DropdownSelectedItemPr
   static displayName = 'DropdownSelectedItem'
   static create: Function
   static slotClassNames: DropdownSelectedItemSlotClassNames
-  static className = 'ui-dropdown__selected-item'
+  static className = 'ui-dropdown__selecteditem'
 
   static propTypes = {
     ...commonPropTypes.createCommon({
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Dropdown/DropdownSelectedItem.tsx
```diff
@@ -90,7 +90,10 @@ class DropdownSelectedItem extends UIComponent<ReactProps<DropdownSelectedItemPr
 
   componentDidUpdate(prevProps: DropdownSelectedItemProps) {
     if (!prevProps.active && this.props.active) {
+      this.itemRef.current.setAttribute('tabindex', '0')
```
mnajdova: We definitely should start extracting behavior for the Dropdown component, we have too much coupling of accessibility concerns in the component... That way we would test this things separately.

mnajdova (approved):
Approved, just update the changelog, and let's start with extracting these things to behavior

layershifter (changes requested):
What it is the reason to use DOM API?

---

```jsx
<Label tabIndex={active ? 0 : -1} />
```

https://codesandbox.io/s/wnnlqm80w8?module=/example.js

layershifter (approved):

PR #1016, opened 2019-03-05
docs(Editor): improve performance and add TS support to CodeSandbox export

layershifter:
### 1. Editor performance

- I disabled code formatting on each rerender
- used `_.debounce()` directly in the `Editor` component

### 2. CodeSandbox export improvements
- much correct description

image
- TS code is exportable now, too 🎉

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
+++ b/docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
@@ -1,12 +1,11 @@
-import * as _ from 'lodash'
 import * as React from 'react'
 import CodeSandboxer from 'react-codesandboxer'
 
 import { ComponentSourceManagerLanguage } from 'docs/src/components/ComponentDoc/ComponentSourceManager'
-import { imports } from 'docs/src/components/Playground/renderConfig'
 import { updateForKeys } from 'docs/src/hoc'
-import { appTemplateJs } from './indexTemplates'
+import { appTemplateJs, appTemplateTs } from './indexTemplates'
 import LabelledButton from '../ComponentButton'
+import createPackageJson from './createPackageJson'
 
 type ComponentControlsCodeSandboxProps = {
   exampleCode: string
@@ -17,6 +16,7 @@ type ComponentControlsCodeSandboxProps = {
 
 type ComponentControlsCodeSandboxState = {
   exampleCode: string
+  examplePath: string
   sandboxUrl: string
 }
 
@@ -26,6 +26,7 @@ class ComponentControlsShowCode extends React.Component<
 > {
   state = {
     exampleCode: '',
+    examplePath: '',
     sandboxUrl: '',
   }
 
@@ -35,12 +36,14 @@ class ComponentControlsShowCode extends React.Component<
   ): Partial<ComponentControlsCodeSandboxState> {
     return {
       exampleCode: props.exampleCode,
+      examplePath: props.exampleLanguage === 'ts' ? '/example.tsx' : '/example.js',
       sandboxUrl: props.exampleCode === state.exampleCode ? state.sandboxUrl : '',
     }
   }
 
   handleDeploy = (embedUrl: string, sandboxId: string) => {
-    const sandboxUrl = `https://codesandbox.io/s/${sandboxId}?module=/example.js`
+    const { examplePath } = this.state
+    const sandboxUrl = `https://codesandbox.io/s/${sandboxId}?module=${examplePath}`
 
     this.setState({ sandboxUrl })
   }
@@ -53,11 +56,11 @@ class ComponentControlsShowCode extends React.Component<
 
   render() {
     const { active, exampleLanguage, exampleCode, exampleName } = this.props
-    const { sandboxUrl } = this.state
+    const { examplePath, sandboxUrl } = this.state
 
-    if (exampleLanguage === 'ts') {
-      return <LabelledButton label="CodeSandbox" iconName="connectdevelop" active={active} />
-    }
+    const main = exampleLanguage === 'ts' ? 'index.tsx' : 'index.js'
+    const appTemplate = exampleLanguage === 'ts' ? appTemplateTs : appTemplateJs
+    const template = exampleLanguage === 'ts' ? 'create-react-app-typescript' : 'create-react-app'
 
     if (sandboxUrl) {
       return (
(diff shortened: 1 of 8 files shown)


docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -389,20 +380,29 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
...
+          />
+        )),
       {
         disabled: currentCodeLanguage !== 'ts',
         icon: 'github',
         content: 'Edit',
         href: ghEditHref,
         rel: 'noopener noreferrer',
         target: '_blank',
+        title: currentCodeLanguage !== 'ts' ? 'You can edit source only in TypeScript' : undefined,
```
layershifter, line 397: Added friendly description 😺

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -389,20 +380,29 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
...
+              <Component
+                {...props}
+                active={active}
+                icon={active ? 'check' : 'copy'}
+                onClick={onClick}
+              />
+            )}
+            value={currentCode}
+          />
+        )),
```
layershifter, line 389: Reused existing `CopyToClipboard` component

docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
```diff
@@ -73,17 +76,15 @@ class ComponentControlsShowCode extends React.Component<
     return (
       <CodeSandboxer
         afterDeploy={this.handleDeploy}
-        examplePath="/"
+        examplePath={examplePath}
         example={exampleCode}
-        dependencies={_.mapValues(imports, () => 'latest')}
-        /* Magic trick to reload sources on passed code update */
-        key={exampleCode}
```
layershifter, line 80: Was fixed in `react-codesandboxer@3.1.2`

docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/createPackageJson.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import { imports } from 'docs/src/components/Playground/renderConfig'
+import { ComponentSourceManagerLanguage } from 'docs/src/components/ComponentDoc/ComponentSourceManager'
+
+const name = 'stardust-ui-example'
+const description =
+  'An exported example from Stardust UI React, https://stardust-ui.github.io/react/'
+const dependencies = _.mapValues(imports, () => 'latest')
+const devDependencies = {
+  '@types/lodash': 'latest',
```
mnajdova, line 11: why we have here types for `lodash`, but we don't have it inside the dependencies?

docs/src/components/ComponentDoc/ComponentSourceManager/ComponentSourceManager.ts
```diff
@@ -111,7 +99,12 @@ export default class ComponentSourceManager extends React.Component<
   }
 
   handleCodeFormat = (): void => {
-    this.setState(prevState => ({ currentCode: prevState.formattedCode }))
+    const { currentCode, currentCodeLanguage } = this.state
```
mnajdova, line 115: This means that whenever the users click this button, we will prettify it, we just won't show them when it should be prettified right?

docs/src/components/Editor/Editor.tsx
```diff
@@ -102,18 +102,29 @@ class Editor extends React.Component<EditorProps> {
 
     // focus editor when editor it becomes active
     if (active !== previousPros.active && active) {
-      this.editorRef.editor.focus()
+      this.editorRef.current.editor.focus()
     }
   }
 
+  handleChange = _.debounce((value: string, e) => {
```
mnajdova, line 109: ❤️

docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
```diff
@@ -53,11 +56,11 @@ class ComponentControlsShowCode extends React.Component<
...
     const { active, exampleLanguage, exampleCode, exampleName } = this.props
-    const { sandboxUrl } = this.state
+    const { examplePath, sandboxUrl } = this.state
 
-    if (exampleLanguage === 'ts') {
-      return <LabelledButton label="CodeSandbox" iconName="connectdevelop" active={active} />
-    }
+    const main = exampleLanguage === 'ts' ? 'index.tsx' : 'index.js'
+    const appTemplate = exampleLanguage === 'ts' ? appTemplateTs : appTemplateJs
+    const template = exampleLanguage === 'ts' ? 'create-react-app-typescript' : 'create-react-app'
```
mnajdova: Now I understand how this works :D

docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
```diff
@@ -53,11 +56,11 @@ class ComponentControlsShowCode extends React.Component<
...
   render() {
     const { active, exampleLanguage, exampleCode, exampleName } = this.props
-    const { sandboxUrl } = this.state
+    const { examplePath, sandboxUrl } = this.state
 
-    if (exampleLanguage === 'ts') {
-      return <LabelledButton label="CodeSandbox" iconName="connectdevelop" active={active} />
-    }
+    const main = exampleLanguage === 'ts' ? 'index.tsx' : 'index.js'
+    const appTemplate = exampleLanguage === 'ts' ? appTemplateTs : appTemplateJs
```
mnajdova, line 62: Now I understand how this works :D

docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/createPackageJson.ts
layershifter: These `devDependencies` are used for an exported example. I don't see any way to get them automatically like `dependencies`.

`@types/*` should be placed in `devDependencies` as CodeSandbox will check for them there.

docs/src/components/ComponentDoc/ComponentSourceManager/ComponentSourceManager.ts
layershifter: Exactly

layershifter: Not super-friendly, but it's better than prettify on each render to show the status of a single button...

mnajdova: Sure, totally agree. I was just checking if I understand this correctly..

miroslavstastny: Prettify on each render is bad, agree.
But to check if code **changed since it was formatted** would be cheap and way better than `wasCodeChanged` which checks if current code differs from the original code.

layershifter: Reworked this stuff as you proposed 👍

miroslavstastny (approved):

PR #1017, opened 2019-03-05
feat(Popup): add inline prop

mnajdova:
Added inline prop for rendering the content of the Popup, next to the trigger element.


miroslavstastny (approved):

docs/src/examples/components/Popup/Types/PopupExampleInline.shorthand.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+const PopupExample = () => (
+  <Popup
+    trigger={<Button icon="expand" />}
+    content="This popup is rendered next to the trigger."
+    inline
+  />
+)
+
+export default PopupExample
```
layershifter:
```suggestion
export default PopupExampleInline
```

packages/react/test/specs/components/Popup/Popup-test.tsx
```diff
@@ -247,4 +247,23 @@ describe('Popup', () => {
       document.body.removeChild(attachTo)
     })
   })
+
+  describe('inline', () => {
+    test('renders the content in the document body the inline prop is not provided', () => {
+      mountWithProvider(<Popup trigger={<button />} content="Content" open={true} />)
+      expect(document.body.firstElementChild.classList.contains(Popup.Content.className)).toEqual(
+        true,
+      )
```
layershifter:
```jsx
const contentElement = document.body.firstElementChild

expect(contentElement.classList.contains(Popup.Content.className)).toEqual(true)
```

Will be more readable

layershifter (approved):

PR #1018, opened 2019-03-05
feat(Tree): Adding exclusive flag

priyankar205:
Adding the exclusive flag. So that only one subtree can be open at a time


docs/src/examples/components/Tree/Types/TreeExclusiveExample.shorthand.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+const titleRenderer = (Component, { content, open, hasSubtree, ...restProps }) => (
+  <Component open={open} hasSubtree={hasSubtree} {...restProps}>
+    {hasSubtree && <Icon name={open ? 'arrow down' : 'arrow right'} />}
+    <span>{content}</span>
+  </Component>
+)
+
+const TreeExclusiveExample = () => (
+  <Tree items={items} renderItemTitle={titleRenderer} exclusive={true} />
```
mnajdova: You don't need `exclusive={true}`, in the component, just exclusive is enough.
`<Tree items={items} renderItemTitle={titleRenderer} exclusive />`

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -78,20 +90,24 @@ class TreeItem extends AutoControlledComponent<ReactProps<TreeItemProps>, TreeIt
...
+      this.setState({
         open: !this.state.open,
       })
+      _.invoke(this.props, 'onOpenChanged', e, this.props)
       _.invoke(predefinedProps, 'onClick', e, titleProps)
     },
   })
 
   renderContent() {
     const { items, title, renderItemTitle } = this.props
```
mnajdova: `const { items, title, renderItemTitle, exclusive } = this.props`
and then use the `exclusive` it would be much more readable

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -78,20 +90,24 @@ class TreeItem extends AutoControlledComponent<ReactProps<TreeItemProps>, TreeIt
...
-    const { open } = this.state
-
+    const open = this.props.exclusive ? this.props.open : this.state.open
```
mnajdova: It looks like the open prop should be auto controlled prop, then you can make decision of when it should be changed. Even if it is auto controlled, you can control it from the parent component.

packages/react/src/components/Tree/TreeTitle.tsx
```diff
@@ -62,10 +71,10 @@ class TreeTitle extends UIComponent<ReactProps<TreeTitleProps>> {
     return (
       <ElementType
         className={classes.root}
-        onClick={this.handleClick}
         {...accessibility.attributes.root}
         {...rtlTextContainer.getAttributes({ forElements: [children, content] })}
         {...unhandledProps}
+        onClick={this.handleClick}
```
layershifter:
```jsx
className={classes.root}
onClick={this.handleClick}
```

Please move it  and place it after the `className` prop

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -52,21 +63,22 @@ export interface TreeItemState {
   open?: boolean
 }
 
-class TreeItem extends AutoControlledComponent<ReactProps<TreeItemProps>, TreeItemState> {
+class TreeItem extends UIComponent<ReactProps<TreeItemProps>, TreeItemState> {
```
layershifter: Why this change is necessary? This means that controlled mode will not work **(!)**, see example:
https://codesandbox.io/s/845j354jx2?module=/example.js

Please revert it.

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -52,21 +63,22 @@ export interface TreeItemState {
...
-
   static propTypes = {
     ...commonPropTypes.createCommon({
       content: false,
     }),
     defaultOpen: PropTypes.bool,
     items: customPropTypes.collectionShorthand,
+    index: PropTypes.number,
+    exclusive: PropTypes.bool,
+    onOpenChanged: PropTypes.func,
```
```suggestion
    onOpenChange: PropTypes.func,
```

We should go with `onOpenChange` as it was defined in other components.

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -78,20 +90,24 @@ class TreeItem extends AutoControlledComponent<ReactProps<TreeItemProps>, TreeIt
...
+  }
+
   handleTitleOverrides = predefinedProps => ({
     onClick: (e, titleProps) => {
       e.preventDefault()
-      this.trySetState({
+      this.setState({
         open: !this.state.open,
       })
+      _.invoke(this.props, 'onOpenChanged', e, this.props)
```
layershifter: `onOpenChange` should be invoked only when it has sub items:

```jsx
if (hasSubtree) _.invoke(this.props, 'onOpenChanged', e, this.props)
```

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -78,20 +90,24 @@ class TreeItem extends AutoControlledComponent<ReactProps<TreeItemProps>, TreeIt
     accessibility: defaultBehavior,
   }
 
+  public state = {
+    open: false,
+  }
```
layershifter: As this component should be autocontrolled, please use `getInitialAutoControlledState()`

```jsx
  getInitialAutoControlledState({ multiple, search }: DropdownProps): DropdownState {
    return {
      open: false,
    }
  }
```

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -52,21 +63,22 @@ export interface TreeItemState {
...
 
-  static autoControlledProps = ['open']
-
   static propTypes = {
     ...commonPropTypes.createCommon({
       content: false,
     }),
     defaultOpen: PropTypes.bool,
     items: customPropTypes.collectionShorthand,
+    index: PropTypes.number,
```
layershifter, line 80: Please add `index` to `TreeItemProps`

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -50,24 +53,43 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
+  state = {
+    clickedIndex: -1,
+  }
+
   public static defaultProps = {
     as: 'ul',
     accessibility: defaultBehavior,
   }
 
+  handleTreeItemOverrides = predefinedProps => ({
```
```suggestion
  handleTreeItemOverrides = (predefinedProps: TreeItemProps) => ({
```

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -50,24 +53,43 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
+    onOpenChanged: (e, treeItemProps) => {
```
```suggestion
    onOpenChanged: (e: React.SyntheticEvent, treeItemProps: TreeItemProps) => {
```

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -50,24 +53,43 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
       content: false,
     }),
+    exclusive: PropTypes.bool,
     items: customPropTypes.collectionShorthand,
     renderItemTitle: PropTypes.func,
     rtlAttributes: PropTypes.func,
   }
 
+  state = {
+    clickedIndex: -1,
```
layershifter: Please use `activeIndex`

layershifter:
This PR contains the chicken-egg problem as I mentioned originally in #479: we have control of `activeIndex`/`clickedIndex` in `Tree` and `open` in `TreeItem` 🤔 

I expect from API something like this:

```jsx
const items = [
  {
    key: '1',
    activeIndex: 0, // < ---
    title: 'one',
    items: [
      {
        key: '2',
        activeIndex: 0, // < ---
        title: 'one one',
        items: [
          {
            key: '3',
            title: 'one one one',
          },
        ],
      },
    ],
  },
  {
    key: '4',
    title: 'two',
    items: [
      {
        key: '5',
        title: 'two one',
      },
    ],
  },
]

<Tree items={items} activeIndex={1} />
```

I suggest to move all `activeIndex` functionality to `TreeItem` and use factory in `Tree`:

```jsx
render() {
  return TreeItem.create({ as, items })
}
```

@mnajdova what do you think?

packages/react/src/components/Tree/TreeItem.tsx
priyankar205: @layershifter What do you mean by this?

priyankar205: Done

mnajdova:
> This PR contains the chicken-egg problem as I mentioned originally in #479: we have control of activeIndex/clickedIndex in Tree and open in TreeItem 🤔

I agree with your point. I would restrain for suggesting what the changes should be (although I agree with you). The goal of it should be, it should be able to auto control it's state and allow the user to control it. With this changes I don't think we are following this pattern. @priyankar205 please consider the comments from @layershifter - let's make this component follow the stardust principles implemented in all other components.

priyankar205:
@mnajdova @layershifter No other component supports nested subcomponents. I will try to make the suggested changes.

priyankar205:
> This PR contains the chicken-egg problem as I mentioned originally in #479: we have control of `activeIndex`/`clickedIndex` in `Tree` and `open` in `TreeItem` 🤔
> 
> I expect from API something like this:
> 
> ```js-jsx
> const items = [
>   {
>     key: '1',
>     activeIndex: 0, // < ---
>     title: 'one',
>     items: [
>       {
>         key: '2',
>         activeIndex: 0, // < ---
>         title: 'one one',
>         items: [
>           {
>             key: '3',
>             title: 'one one one',
>           },
>         ],
>       },
>     ],
>   },
>   {
>     key: '4',
>     title: 'two',
>     items: [
>       {
>         key: '5',
>         title: 'two one',
>       },
>     ],
>   },
> ]
> 
> <Tree items={items} activeIndex={1} />
> ```
> 
> I suggest to move all `activeIndex` functionality to `TreeItem` and use factory in `Tree`:
> 
> ```js-jsx
> render() {
>   return TreeItem.create({ as, items })
> }
> ```
> 
> @mnajdova what do you think?

If I move activeIndex to the TreeItem then how do I control the sibling TreeItems. How do I close the rest when the exclusive flag is there

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -39,7 +48,7 @@ export interface TreeProps extends UIComponentProps, ChildrenComponentProps {
 /**
  * Allows users to display data organised in tree-hierarchy.
  */
-class Tree extends UIComponent<ReactProps<TreeProps>> {
+class Tree extends AutoControlledComponent<ReactProps<TreeProps>, any> {
```
layershifter: Let's create an interface for the state:

```tsx
export interface TreeState {
  activeIndex: number[] | number
}
```

```suggestion
class Tree extends AutoControlledComponent<ReactProps<TreeProps>, TreeState> {
```

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +78,47 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
     accessibility: defaultBehavior,
   }
 
+  static autoControlledProps = ['activeIndex']
+
+  getInitialAutoControlledState({ defaultActiveIndex, exclusive }) {
```
```suggestion
  getInitialAutoControlledState({ defaultActiveIndex, exclusive }): TreeState {
```

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +78,47 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
+    }
+  }
+
+  computeNewIndex = index => {
+    const { activeIndex } = this.state
+    const { exclusive } = this.props
+
+    if (exclusive) return index === activeIndex ? -1 : index
+    // check to see if index is in array, and remove it, if not then add it
+    return _.includes(activeIndex, index) ? _.without(activeIndex, index) : [...activeIndex, index]
```
```suggestion
    return _.includes(activeIndex as number[], index) ? _.without(activeIndex as number[], index) : [...activeIndex as number[], index]
```

As we apply checks on `exclusive`, TS doesn't know that `activeIndex` is `number`. @mnajdova do you have better ideas? 💭

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +78,47 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
 
+  static autoControlledProps = ['activeIndex']
+
+  getInitialAutoControlledState({ defaultActiveIndex, exclusive }) {
+    return {
+      activeIndex: defaultActiveIndex === undefined ? (exclusive ? -1 : []) : defaultActiveIndex,
+    }
+  }
+
+  computeNewIndex = index => {
```
```suggestion
  computeNewIndex = (index: number) => {
```

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +78,47 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
+      _.invoke(predefinedProps, 'onOpenChange', e, treeItemProps)
+    },
+  })
+
   renderContent() {
-    const { items, renderItemTitle } = this.props
+    const { items, renderItemTitle, exclusive } = this.props
 
-    return _.map(items, item =>
+    return _.map(items, (item, index) =>
```
```suggestion
    return _.map(items, (item: ShorthandValue, index: number) =>
```

To clarify types

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +78,47 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
+      this.trySetState({
+        activeIndex: this.computeNewIndex(index),
+      })
+      _.invoke(predefinedProps, 'onOpenChange', e, treeItemProps)
+    },
+  })
+
   renderContent() {
-    const { items, renderItemTitle } = this.props
+    const { items, renderItemTitle, exclusive } = this.props
```
layershifter, line 118:
```suggestion
    const { items, renderItemTitle, exclusive } = this.props
    const { activeIndex } = this.state
```

Let's desctruct `activeIndex `

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +78,47 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
-    return _.map(items, item =>
+    return _.map(items, (item, index) =>
       TreeItem.create(item, {
         defaultProps: {
+          index,
+          exclusive,
           renderItemTitle,
+          open: exclusive
+            ? index === this.state.activeIndex
+            : _.includes(this.state.activeIndex, index),
```
```tsx
open: exclusive ? index === activeIndex : _.includes(activeIndex as number[], index),
```

Now can be one-liner 👍

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +78,47 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
         },
+        overrideProps: this.handleTreeItemOverrides,
       }),
```
layershifter, line 131: Let's also add shorthand factory, we will use it in the `TreeItem`:

```diff
+Tree.create = createShorthandFactory({ Component: Tree, mappedArrayProp: 'items' })
+
export default Tree
```

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -104,7 +113,13 @@ class TreeItem extends AutoControlledComponent<ReactProps<TreeItemProps>, TreeIt
...
           overrideProps: this.handleTitleOverrides,
         })}
-        {hasSubtree && open && <Tree items={items} renderItemTitle={renderItemTitle} />}
+        {hasSubtree && open && (
+          <Tree
+            items={items}
+            renderItemTitle={renderItemTitle}
+            exclusive={!!this.props.exclusive}
+          />
+        )}
```
layershifter: Factory will handle falsy values for us, so we can just do this:
```tsx
{open && Tree.create(items, {
  defaultProps: {
    exclusive,
    renderItemTitle,
  },
})}
```

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -78,20 +91,16 @@ class TreeItem extends AutoControlledComponent<ReactProps<TreeItemProps>, TreeIt
...
+      _.invoke(this.props, 'onOpenChange', e, this.props)
       _.invoke(predefinedProps, 'onClick', e, titleProps)
     },
   })
 
   renderContent() {
-    const { items, title, renderItemTitle } = this.props
-    const { open } = this.state
-
+    const { items, title, renderItemTitle, open } = this.props
```
```suggestion
    const { exclusive, items, title, renderItemTitle, open } = this.props
```

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -78,20 +91,16 @@ class TreeItem extends AutoControlledComponent<ReactProps<TreeItemProps>, TreeIt
...
   }
 
-  handleTitleOverrides = predefinedProps => ({
+  handleTitleOverrides = (predefinedProps: TreeItemProps) => ({
     onClick: (e, titleProps) => {
       e.preventDefault()
-      this.trySetState({
-        open: !this.state.open,
-      })
+      _.invoke(this.props, 'onOpenChange', e, this.props)
```
layershifter: Actually, now this is `onTitleClick`

```suggestion
      _.invoke(this.props, 'onTitleClick', e, this.props)
```

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +78,47 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
+    return _.includes(activeIndex, index) ? _.without(activeIndex, index) : [...activeIndex, index]
+  }
+
+  handleTreeItemOverrides = predefinedProps => ({
+    onOpenChange: (e: React.SyntheticEvent, treeItemProps: TreeItemProps) => {
+      const { index } = treeItemProps
+      this.trySetState({
+        activeIndex: this.computeNewIndex(index),
+      })
+      _.invoke(predefinedProps, 'onOpenChange', e, treeItemProps)
```
```tsx
    onTitleClick: (e: React.SyntheticEvent, treeItemProps: TreeItemProps) => {
      this.trySetState({ activeIndex: this.computeNewIndex(treeItemProps.index) })
      _.invoke(predefinedProps, 'onTitleClick', e, treeItemProps)
    },
```

https://github.com/stardust-ui/react/pull/1018/files#r263292996

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -67,6 +77,9 @@ class TreeItem extends AutoControlledComponent<ReactProps<TreeItemProps>, TreeIt
     }),
     defaultOpen: PropTypes.bool,
     items: customPropTypes.collectionShorthand,
+    index: PropTypes.number,
+    exclusive: PropTypes.bool,
+    onOpenChange: PropTypes.func,
```
```suggestion
    onTitleClick: PropTypes.func,
```

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -25,12 +30,21 @@ export interface TreeItemProps extends UIComponentProps, ChildrenComponentProps
...
   defaultOpen?: boolean
 
+  /** The index of the item among its sibbling */
+  index: number
+
   /** Array of props for sub tree. */
   items?: ShorthandValue[]
 
+  /** Callback for toggling the current tree item as open element in the menu. */
+  onOpenChange?: ComponentEventHandler<TreeItemProps>
```
```suggestion
  onTitleClick?: ComponentEventHandler<TreeItemProps>
```

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -25,12 +30,21 @@ export interface TreeItemProps extends UIComponentProps, ChildrenComponentProps
...
   /** Initial open value. */
   defaultOpen?: boolean
 
+  /** The index of the item among its sibbling */
+  index: number
+
   /** Array of props for sub tree. */
   items?: ShorthandValue[]
 
+  /** Callback for toggling the current tree item as open element in the menu. */
```
```suggestion
  /** Called when a tree title is clicked. */
```

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +78,47 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
+  computeNewIndex = index => {
+    const { activeIndex } = this.state
+    const { exclusive } = this.props
+
+    if (exclusive) return index === activeIndex ? -1 : index
+    // check to see if index is in array, and remove it, if not then add it
+    return _.includes(activeIndex, index) ? _.without(activeIndex, index) : [...activeIndex, index]
+  }
+
+  handleTreeItemOverrides = predefinedProps => ({
```
```suggestion
  handleTreeItemOverrides = (predefinedProps: TreeItemProps) => ({
```

docs/src/examples/components/Tree/Types/index.tsx
```diff
@@ -14,6 +14,11 @@ const Types = () => (
       description="A Tree with customized title rendering."
       examplePath="components/Tree/Types/TreeTitleCustomizationExample"
     />
+    <ComponentExample
+      title="Exclusive Flag"
```
layershifter:
```suggestion
      title="Exclusive"
```

CHANGELOG.md
```diff
@@ -20,6 +20,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Features
 - Add `inline` prop in the `Popup` for rendering the content next to the trigger element @mnajdova ([#1017](https://github.com/stardust-ui/react/pull/1017))
 
+- Add `exclusive` prop in the `Tree` for expanding one tree item at a time
```
layershifter, line 25: Please remove empty line before, 22

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +83,48 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
+      ? _.without(activeIndex as number[], index)
+      : [...(activeIndex as number[]), index]
+  }
+
+  handleTreeItemOverrides = (predefinedProps: TreeItemProps) => ({
+    onTitleClick: (e: React.SyntheticEvent, treeItemProps: TreeItemProps) => {
+      const { index } = treeItemProps
+      this.trySetState({
+        activeIndex: this.computeNewIndex(index),
+      })
```
```diff
-      const { index } = treeItemProps
-      this.trySetState({
-        activeIndex: this.computeNewIndex(index),
-      })
 this.trySetState({ activeIndex: this.computeNewIndex(treeItemProps.index) })
```

nit: 4 lines vs 1

layershifter (approved):

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +83,45 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
     accessibility: defaultBehavior,
   }
 
+  static autoControlledProps = ['activeIndex']
```
mnajdova, line 86: 👍

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +83,45 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
+    const { activeIndex } = this.state
 
-    return _.map(items, item =>
+    return _.map(items, (item: ShorthandValue, index: number) =>
       TreeItem.create(item, {
         defaultProps: {
+          index,
+          exclusive,
           renderItemTitle,
+          open: exclusive ? index === activeIndex : _.includes(activeIndex as number[], index),
```
mnajdova: Normalize the `activeIndex` as an array before doing this checks. If I don't use the exclusive, but use just defaultActiveIndex={3}, this may break. Do something like:
```
const activeIndexAsArray = _.isArray(activeIndex) ? activeIndex : [activeIndex]
```

and then always use the `_.include`

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -78,20 +91,16 @@ class TreeItem extends AutoControlledComponent<ReactProps<TreeItemProps>, TreeIt
     accessibility: defaultBehavior,
   }
 
-  handleTitleOverrides = predefinedProps => ({
+  handleTitleOverrides = (predefinedProps: TreeItemProps) => ({
```
mnajdova: Isn't the `predefinedProps` of type `TreeTitleProps`? @layershifter

layershifter: Yep, my typo, great catch 👍

packages/react/src/components/Tree/Tree.tsx
priyankar205: This is consistent with the accordion.
screen shot 2019-03-07 at 6 09 57 pm

mnajdova: Still, the user may provide the example showed above. If this is how it is implemented the Accordion, it is another story and we may fix it there as well.

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +83,53 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
 
+  static autoControlledProps = ['activeIndex']
+
+  getInitialAutoControlledState({ exclusive }): TreeState {
+    return {
+      activeIndex: exclusive ? -1 : [],
+    }
+  }
+
+  getActiveIndexAsArray() {
```
```suggestion
  getActiveIndexAsArray(): number[] {
```

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +83,53 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
+
+  computeNewIndex = (index: number) => {
+    const { exclusive } = this.props
+
+    if (exclusive) return index
+    const activeIndexAsArray = this.getActiveIndexAsArray()
+    // check to see if index is in array, and remove it, if not then add it
+    return _.includes(activeIndexAsArray as number[], index)
+      ? _.without(activeIndexAsArray as number[], index)
+      : [...(activeIndexAsArray as number[]), index]
```
layershifter: Now we don't need `as number`

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +83,53 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
+      this.trySetState({ activeIndex: this.computeNewIndex(treeItemProps.index) })
+      _.invoke(predefinedProps, 'onTitleClick', e, treeItemProps)
+    },
+  })
+
   renderContent() {
-    const { items, renderItemTitle } = this.props
+    const { items, renderItemTitle, exclusive } = this.props
+    const { activeIndex } = this.state
+    const activeIndexAsArray = this.getActiveIndexAsArray()
```
```diff
-const { activeIndex } = this.state
-const activeIndexAsArray = this.getActiveIndexAsArray()
+const activeIndexes = this.getActiveIndexAsArray()
```

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -60,14 +83,53 @@ class Tree extends UIComponent<ReactProps<TreeProps>> {
...
-    return _.map(items, item =>
+    return _.map(items, (item: ShorthandValue, index: number) =>
       TreeItem.create(item, {
         defaultProps: {
+          index,
+          exclusive,
           renderItemTitle,
+          open: exclusive
+            ? index === activeIndex
+            : _.includes(activeIndexAsArray as number[], index),
```
```diff
-          open: exclusive
-            ? index === activeIndex
-            : _.includes(activeIndexAsArray as number[], index),
+         open  : _.includes(activeIndexes, index),

packages/react/src/components/Tree/Tree.tsx
layershifter: I also suggest to rename to `getActiveIndexes()`

packages/react/src/components/Tree/Tree.tsx
priyankar205: For exclusive activeIndex is not an array. So this condition needs to be there.

layershifter: `this.getActiveIndexes()` will always return array, even with `exclusive=true`

PR #1020, opened 2019-03-06
feat(Label): add label type: badge

jaanus03:
This PR adds an optional type 'badge' to Label component that looks and acts a little different from existing types.

- Has a new optional field 'additionalContent' 
- Looks are redlined for Teams theme

badge-redlines

### DIFF
--- a/packages/react/src/components/Label/Label.tsx
+++ b/packages/react/src/components/Label/Label.tsx
@@ -19,6 +19,7 @@ import Icon from '../Icon/Icon'
 import Image from '../Image/Image'
 import Layout from '../Layout/Layout'
 import Text from '../Text/Text'
+import Box from '../Box/Box'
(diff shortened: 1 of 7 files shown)


packages/react/src/components/Label/Label.tsx
```diff
@@ -120,15 +138,26 @@ class Label extends UIComponent<ReactProps<LabelProps>, any> {
       },
       overrideProps: this.handleIconOverrides,
     })
+    const additionalContentElement = additionalContent
```
mnajdova: The factories are handling the null values, you can just use:
```
const additionalContentElement = Text.create(additionalContent, {
  defaultProps: {
    styles: styles.additionalContent,
    variables: variables.additionalContent,
  },
})
```

PR #1028, opened 2019-03-07
fix(Popup): Do not propagate keyboard event only when focus trap behavior is used

sophieH29:


packages/react/test/specs/components/Popup/Popup-test.tsx
```diff
@@ -266,4 +263,44 @@ describe('Popup', () => {
...
+
+      // when popup open, check that stopPropagation is called when keyboard events are invoked
+      const stopPropagation = jest.fn()
+      const popupContentElement = getPopupContent(popup)
+      popupContentElement.simulate('keyDown', { stopPropagation })
+      if (shouldStopPropagation) {
+        expect(stopPropagation).toHaveBeenCalledTimes(1)
+      } else {
+        expect(stopPropagation).not.toBeCalled()
+      }
```
layershifter:
```
expect(stopPropagation).toHaveBeenCalledTimes(Number(shouldStopPropagation))
// or 
expect(stopPropagation).toHaveBeenCalledTimes(shouldStopPropagation ? 1 : 0)
```

packages/react/test/specs/components/Popup/Popup-test.tsx
```diff
@@ -266,4 +263,44 @@ describe('Popup', () => {
...
+          content={<span id={contentId} />}
+          accessibility={behavior}
+        />,
+      )
+
+      // check popup open on key press
+      const popupTriggerElement = popup.find(`#${triggerId}`)
+      popupTriggerElement.simulate('keydown', { keyCode: keyboardKey.Enter })
+
+      expect(getPopupContent(popup).exists()).toBe(true)
```
layershifter: Not sure that we need this assertion there.

packages/react/test/specs/components/Popup/Popup-test.tsx
```diff
@@ -266,4 +263,44 @@ describe('Popup', () => {
       expect(contentElement.classList.contains(Popup.Content.className)).toEqual(true)
     })
   })
+
+  describe('keyboard event propagation', () => {
+    const expectPopupToStopPropagation = (
```
mnajdova: Would rather to rename it to `expectPopupToHandleStopPropagation` as this tests both cases of when it should stop and not.

mnajdova (approved):
One small comment, other then that ❤️

packages/react/test/specs/components/Popup/Popup-test.tsx
sophieH29: nice 👍

packages/react/test/specs/components/Popup/Popup-test.tsx
sophieH29: you're right, will remove

packages/react/test/specs/components/Popup/Popup-test.tsx
sophieH29: will rename

layershifter (approved):

PR #1031, opened 2019-03-07
fix(dropdown): fix getA11ySelectionMessage usages across examples

bmdalex:
## fix(dropdown): fix getA11ySelectionMessage usages across examples

This PR restores the initial usages of `getA11ySelectionMessage` method. This method should be used only for `Dropdown` components that have the `multiple` prop


bmdalex:
@silviuavram will continue this work

PR #1032, opened 2019-03-07
fix(dropdown): use Portal to create a11y selection message container

silviuaavram:
Because the status div was created only when the first selection message was generated, the first time this happened the reader did not announce it.

Now creating the selection status div at component mount, if the Dropdown contains the `getA11ySelectionMessage` prop.

Using the `Popup` component to create the message container element.

Updating the examples and added back the add/delete messages to the multiple selection.


layershifter (changes requested):
Please refactor this using our `Portal` component or `ReactDOM.createPortal()` API, I don't any reasons to use there DOM API directly.

https://reactjs.org/docs/portals.html

----

```tsx
render() {
  return (
    <>
      {this.renderDropdown()}
      <Portal open={!!this.state.a11ySelectionStatus}>
        <div role='status' aria-live='polite' style={screenReaderContainerStyles}>
          {this.state.a11ySelectionStatus}
        </div>
      </Portal>
    <>
  )
}
```

silviuaavram:
@layershifter last time I wanted to do that it did not work as expected. will give it another go but if it is too complicated I will drop it. the scope of the PR is to fix a bug for embed.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -648,23 +661,6 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     return filteredItems
   }
 
-  private setA11yStatus = (statusMessage: string) => {
```
bmdalex, line 651: 👍

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -34,4 +34,15 @@ describe('Dropdown', () => {
       )
     })
   })
+
+  describe('getA11ySelectionMessage', () => {
+    it('creates message container element', () => {
+      mountWithProvider(<Dropdown options={[]} getA11ySelectionMessage={{}} />)
+      expect(
+        document.querySelector(
+          `[role="status"][aria-live="polite"][aria-relevant="additions text"]`,
```
bmdalex, line 43: 👍

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -402,6 +405,16 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
             )
           }}
         </Downshift>
+        <Portal open={!!this.props.getA11ySelectionMessage}>
```
bmdalex, line 408: nit: `this.props` is destructured at line 311, you can move `getA11ySelectionMessage` there

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -402,6 +405,16 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+          <div
+            role="status"
+            aria-live="polite"
+            aria-relevant="additions text"
+            style={screenReaderContainerStyles}
+          >
+            {this.state.a11ySelectionStatus}
```
bmdalex, line 415: nit: `this.state` is destructured at line 312, you can move `a11ySelectionStatus ` there

bmdalex (approved):
👍 looks good, 2 nit

bmdalex:
@silviuavram @layershifter 
depending on the decision to render the `a11ySelectionStatus` in a `Portal` or not we should standardize (and maybe even add to dev documentation) this knowledge.

I'm sure other components will need to implement something similar or already do?

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: thanks!

layershifter (approved):

layershifter:
May be we need to change PR title/changelog entry? 🤔

PR #1042, opened 2019-03-11
fix(types): make *Props interface extendable

### LINKED ISSUE #1026: *Props interfaces are not extendable
### Steps
1. Import `ListItemProps`
2. Try to add unhandled prop:
```tsx
const item: ListItemProps = {
  key: 'foo',
  role: 'bar',
}
```

### Expected Result
It works.

### Actual Result
```
Error:(70, 3) TS2322: Type '{ key: string; title: string; }' is not assignable to type 'ListItemProps'.
  Object literal may only specify known properties, and 'key' does not exist in type 'ListItemProps'.
```

### Version
0.23.0

### Proposal
Reorganize types like it's done in SUIR:
- strict `ListItemPropsStrict`
- extendable `ListItemProps`

layershifter:
Fixes #1026.


packages/react/src/lib/commonPropInterfaces.ts
```diff
@@ -18,6 +18,8 @@ export interface AnimatedComponentProps {
 export interface UIComponentProps<P = any, V = any>
   extends StyledComponentProps<P, V>,
     AnimatedComponentProps {
+  [key: string]: any
```
layershifter, line 21: All of our components (except, `Flex`) that actually should have extendable props are using this interface

bmdalex (approved):
👍

PR #1043, opened 2019-03-11
feat(screenshot tests): add support for common themes

bmdalex:
# feat(screenshot tests): add support for common themes

## Description

This PR adds support for adding screenshot tests for the most common themes using the screener API

### 1. `themes` prop together with `steps`:
In your `Component.steps.ts` file introduce a config file that can contain the following props:
  - `steps`: the array of callback representing the test steps as before
  - `themes`: an array of strings representing the theme applied to the component when taking the screenshot; by default, all screenshots are taken for `themes` theme.

```jsx
const config: ScreenerTestsConfig = {
  themes: ['teams', 'teamsDark', 'teamsHighContrast'],
  steps: [sb => sb.click('#sel').snapshot('#snap')],
}

export default config
```
### 2. `themes` prop together without `steps`:
The theme switching mechanism should work even without the `steps` prop:

```jsx
const config: ScreenerTestsConfig = {
  themes: ['teams', 'teamsDark', 'teamsHighContrast']
}

export default config
````
## NOTE:

This PR focuses only on adding the functionality of testing components in other themes with minimal examples presented for:
- [`themes` with `steps`](https://github.com/stardust-ui/react/pull/1043/files#r264679205)
- [`themes` WITHOUT `steps`](https://github.com/stardust-ui/react/pull/1043/files#r264679650)

The tests themselves for the other components will be added in subsequent PRs.

### CODE REVIEW

File: docs/src/app.tsx [hunk 1]
```diff
@@ -3,33 +3,15 @@ import * as _ from 'lodash'
...
-    // State also contains the updater function so it will
-    // be passed down into the context provider
-    this.state = {
-      themeName: 'teams',
-      changeTheme: this.changeTheme,
-    }
+class App extends React.Component<any, ThemeContextData> {
+  // State also contains the updater function so it will
+  // be passed down into the context provider
+  state: ThemeContextData = {
```
bmdalex, line 12: can be simplified even more and use just `themeName` as state, but maybe it's easier to read like this

docs/src/examples/components/Dropdown/Types/DropdownExampleSearchMultiple.shorthand.steps.ts
```diff
@@ -8,17 +9,16 @@ const selectors = {
...
-      .click(selectors.toggleIndicator)
-      .click(selectors.item(2))
-      .click(selectors.toggleIndicator)
-      .click(selectors.item(2))
-      .keys(selectors.input, keys.leftArrow)
-      .snapshot('Selects last selected element')
-      .hover(selectors.selectedItem(1))
-      .snapshot('Hovers first selected element'),
-]
+const steps: ScreenerSteps = applyCommonThemesBeforeStep((steps, keys) =>
```
layershifter: I don't like this change 👎 I suggest to add `themes` as handlable export:
```
import { themes } from '@stardust-ui/react'

export const themes = [themes.teamsDark]
// or
export { themes }
```

docs/src/components/ExternalExampleLayout.tsx
```diff
@@ -49,34 +53,53 @@ class ExternalExampleLayout extends React.Component<
...
-                  {element}
-                  {/* This block allows to see issues with examples as visual regressions. */}
-                  {error && (
-                    <div style={{ fontSize: '5rem', color: 'red' }}>{error.toString()}</div>
+      <Provider key={this.state.renderId} theme={theme}>
+        <Provider.Consumer
+          render={({ siteVariables }) => (
+            <div
+              dir={theme.rtl ? 'rtl' : undefined}
+              style={{
```
bmdalex, line 66: Unfortunately this is needed in order to see the actual changes in other themes, and that's because `color` and `backgroundColor` are not set for a lot of components in other themes.

However, this means the styles add changes for all screenshots so we will see changes in all of them...

types/screener.d.ts
```diff
@@ -16,54 +18,59 @@ type ScreenerRunnerKeys = {
...
-  end(): ScreenerStepBuilder
-}
+    /** This will return the steps to be run. */
+    end(): ScreenerStepBuilder
 
-//
-// Typings for `*.steps.tsx`
-//
+    /** This will switch the theme */
+    switchTheme(themeName: ThemeName): ScreenerStepBuilder
```
layershifter: I expect have `theme: ThemeInput` as value:

```js
import { themes } from '@stardust-ui/react'

steps.switchTheme(themes.dark)
```

docs/src/components/ExternalExampleLayout.tsx
```diff
@@ -49,34 +53,53 @@ class ExternalExampleLayout extends React.Component<
...
-          >
-            <SourceRender.Consumer>
-              {({ element, error }) => (
-                <>
-                  {element}
-                  {/* This block allows to see issues with examples as visual regressions. */}
-                  {error && (
-                    <div style={{ fontSize: '5rem', color: 'red' }}>{error.toString()}</div>
+      <Provider key={this.state.renderId} theme={theme}>
+        <Provider.Consumer
```
bmdalex, line 62: this approach as taken from `docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx`, line 588:
```jsx
<Provider.Consumer
  render={({ siteVariables }) => {
    return (
      <Segment
        dir={showRtl ? 'rtl' : undefined}
        className={`rendered-example ${this.getKebabExamplePath()}`}
        styles={{
          padding: '2rem',
          color: siteVariables.bodyColor,
          backgroundColor: siteVariables.bodyBackground,
          ...(showTransparent && {
            backgroundImage:
              'url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs+9AAAAKUlEQVQoU2NkYGAwZkAD////RxdiYBwKCv///4/hGUZGkNNRAeMQUAgAtxof+nLDzyUAAAAASUVORK5CYII=")',
            backgroundRepeat: 'repeat',
          }),
        }}
      >
        <SourceRender.Consumer>{({ element }) => element}</SourceRender.Consumer>
      </Segment>
    )
  }}
/>
```

 I think we should have the same approach in both cases (use segment with `2rem` padding) and extract logic in separate component. I'll probably do that, what do you think?

docs/src/examples/components/Button/Types/ButtonExample.shorthand.steps.ts
```diff
@@ -0,0 +1 @@
+export { commonThemesSteps as default } from '../../../screenerStepsUtils'
```
bmdalex: this is the current approach for testing `ButtonExample.shorthand.tsx` screenshot in all themes, we plan to simplify it, see ISSUES in Description

docs/src/examples/components/Dropdown/Rtl/DropdownExample.rtl.steps.ts
```diff
@@ -1,17 +1,17 @@
...
 
-const steps: ScreenerSteps = [
-  steps =>
-    steps
-      .click(selectors.triggerButton)
-      .snapshot('RTL: Shows list')
-      .click(selectors.item(3))
-      .snapshot('RTL: Selects an item'),
-]
+const steps: ScreenerSteps = applyCommonThemesBeforeStep(steps =>
```
bmdalex: this is the current approach for testing `DropdownExample.shorthand.rtl.tsx` screenshot in all themes, we plan to simplify it, see ISSUES in Description

docs/src/examples/components/Dropdown/Types/DropdownExampleSearchMultiple.shorthand.steps.ts
bmdalex: fixed and applied change as discussed

bmdalex: fixed and applied change as discussed

layershifter (commented):
Can we add new tests in a separate PR (`ButtonExample.shorthand.steps.ts`, ButtonExampleEmphasis.shorthand.steps.ts`)?

docs/src/examples/components/Button/Types/ButtonExample.shorthand.steps.ts
bmdalex: fixed

build/screener/config.js
```diff
@@ -0,0 +1,2 @@
+require('ts-node').register()
+module.exports = require('./screener.config').default
```
layershifter: 1. I propose to move `tsPaths` there, too:

```diff
+tsPaths.register({
+  baseUrl: config.path_base,
+  paths: compilerOptions.paths,
+})
```

2. I propose to rename this back to `screener.config.js` while, `screener.config.ts` will be `screener.states.ts` and will provide only `states` for config section. What do think?

build/screener/steps.ts
```diff
@@ -0,0 +1,34 @@
...
+import { reduceSteps, applyThemesBeforeStep } from './utils'
+
+/** converts the arguments into a script passed to Steps.executeScript method
+ * e.g.: steps.executeOnWindow('switchTheme', 'teams')
+ * will call steps.executeScript('window.switchTheme("teams"))
+ */
+Steps.prototype.executeOnWindow = function (methodName: string, ...args: string[]) {
+  const computerArgs = args.length ? `"${args.join('", "')}"` : ''
+  return this.executeScript(`window.${methodName}(${computerArgs})`)
+}
```
layershifter: This method still looks useless, we need it in two cases and the direct usage is much obvious 🤔 I propose to get rid of it for now.

build/screener/steps.ts
```diff
@@ -0,0 +1,34 @@
...
+    return undefined
+  }
+
+  const { steps, themes } = require(stepsModulePath).default as ScreenerTestsConfig
+  const themedStep = applyThemesBeforeStep(reduceSteps(steps), themes)
+
+  return themedStep(new Steps(), keys).end()
+}
+
+export default getSteps
```
layershifter: Please rename it to `getScreenerSteps()` and file, too.

docs/src/components/ExternalExampleLayout.tsx
layershifter: These changes will be done in #852, as you can see it has the same regressions 💣

build/screener/config.js
bmdalex: my plan was to get rid of the `.js` file altogether but I think screener API does not accept that.. why not have most of the code in Typescript if it's possible? it's a typescript project after all.

Moving all the code to typescript gave me a better understanding of what's happening there as well since I get instant access to the API through correct typings.

What would be the reasons to go back to JS?

docs/src/components/ExternalExampleLayout.tsx
layershifter: I propose, to add them right now, but before this PR. What do you think?

docs/src/components/ExternalExampleLayout.tsx
layershifter: Let's leave as is, because #852 will apply styles and we will not need it at all 🌟

build/screener/steps.ts
bmdalex: I can agree to that. will change

build/screener/steps.ts
bmdalex: will do, we have strange file name conventions... would this be `screener.steps.ts` or `screenerSteps.ts`

docs/src/components/ExternalExampleLayout.tsx
bmdalex: I propose, to add them right now, but before this PR. 

@layershifter not sure I understand, you propose to merge #852 before this or? who is `them` in this context? 😆

docs/src/components/ExternalExampleLayout.tsx
bmdalex: oh, that answers my question above..hmmm the tests will be useless until then, I don't really like that especially that the #852 is not that active anymore

docs/src/components/ExternalExampleLayout.tsx
bmdalex: got it, you mean to remove the changes and wait for #852 wel..the tests will be useless until then, I don't really like that especially that the #852 is not that active anymore

docs/src/examples/components/Dropdown/Rtl/DropdownExample.rtl.steps.ts
bmdalex: fixed, we have better approach now

build/screener/steps.ts
bmdalex: done

build/screener/steps.ts
bmdalex: done

docs/src/components/ExternalExampleLayout.tsx
bmdalex: agreed to split this in 2 PRs to avoid screen regressions here, the new PR will depend on #852

docs/src/examples/components/Menu/Slots/MenuExample.shorthand.steps.ts
```diff
@@ -1,3 +1,8 @@
 import getScreenerSteps from '../commonScreenerSteps'
 
-export default getScreenerSteps()
+const config: ScreenerTestsConfig = {
+  themes: ['teams', 'teamsDark', 'teamsHighContrast'],
```
bmdalex, line 4: Example of adding `themes` prop together with `steps`

docs/src/examples/components/Button/Types/ButtonExample.shorthand.steps.ts
```diff
@@ -0,0 +1,3 @@
+const config: ScreenerTestsConfig = { themes: ['teams', 'teamsDark', 'teamsHighContrast'] }
```
bmdalex, line 1: Example of adding `themes` prop without `steps`; theme switching should still work

build/screener/config.js
bmdalex: resolved per our discussion

types/screener.d.ts
bmdalex: agreed that `themeName` is fine for now

docs/src/components/ExternalExampleLayout.tsx
bmdalex: agreed to split this in 2 PRs to avoid screen regressions here, the new PR will depend on #852

docs/src/components/ExternalExampleLayout.tsx
bmdalex: resolving, tracked in newly opened #1050

docs/src/components/ExternalExampleLayout.tsx
bmdalex: resolving, tracked in newly opened #1050

bmdalex:
## `ts-lint` vs `prettier` issue 😕 

### 1. `ts-lint` complains about missing whitespace before function parens so it adds it with the `git commit` script:

Screenshot 2019-03-12 at 18 12 30

### 2. `prettier` sees that whitespace as an issue so the build fails 😠  
Screenshot 2019-03-12 at 18 21 26

#### `yarn prettier:fix` removes the space and the cycle continues....

Screenshot 2019-03-12 at 18 15 55


How can I fix this? Ideas?
@layershifter @mnajdova @miroslavstastny 

### LATER EDIT:

Added names to functions even though it looks strange.. seems to have fixed it.

layershifter:
> How can I fix this? Ideas?

Switch to ESLint 😸 Actually, your fix is fine

UPD: Should be fixed with `tslint-config-prettier` from #1052

.github/test-a-feature.md
```diff
@@ -128,24 +128,29 @@ For some components, it is necessary to write screenshot tests in order to check
...
 - the test file should be placed at the same location as the component example under test.
 - the test file should be named exactly as the component example file. If `DropdownExample.shorthand.tsx` is to be tested, the screener test file should be named `DropdownExample.shorthand.steps.ts`.
-- the tests should be written as an array of callbacks that accept a `steps` parameter, as all of them will be chained in `screener.config`. The `steps` parameter is actually the `Steps` object from screener, instantiated in `screener.config`.
+- the tests should be written as a config file that can contain the following props:
+  - `steps`: an array of callbacks that accept a `builder` (step builder) parameter, as all of them will be chained in `screener.config`. The `builder` parameter is actually the `Steps` object from screener, instantiated in `screener.config`.
```
layershifter:
```suggestion
  - `steps`: an array of callbacks that accept a `builder` (step builder) parameter, as all of them will be chained in `screener.config.js`. The `builder` parameter is actually the `Steps` object from screener, instantiated in `screener.config.js`.
```

.github/test-a-feature.md
```diff
@@ -128,24 +128,29 @@ For some components, it is necessary to write screenshot tests in order to check
...
+  - `themes`: an array of strings representing the theme applied to the component when taking the screenshot; by default, all screenshots are taken for `themes` theme.
```
layershifter:
```suggestion
  - `themes`: an array of strings representing the theme applied to the component when taking the screenshot; by default, all screenshots are taken for `Teams` theme.
```

.github/test-a-feature.md
```diff
@@ -154,12 +159,12 @@ export default steps
 - an actual assertion is performed by taking a `.snapshot(<Your test name here>)`, as the assertion is performed by comparing the screenshot with the one taken on the previous run and considered correct.
 - a test can have as many snapshots as needed.
 - before the snapshot is taken, steps are added to reach the state of assertion, with methods from `screener-runner` API (`.click(<selector>)`, `.hover(<selector>)`etc.)
-- tests do not perform any cleanup by default. This means a test is dependent on the state of the component in which the previous test leaves it. In the example above, the second test will receive the `Dropdown` component with the list open.
+- tests perform any cleanup by default. This means each test is independent of the state of the component from previous tests.
```
layershifter:
```suggestion
- tests perform cleanup by default. This means each test is independent of the state of the component from previous tests.
```

.github/test-a-feature.md
```diff
@@ -154,12 +159,12 @@ export default steps
...
 
 ### Run Screener tests
 In order to run the tests locally, make sure to have a Screener API key saved in the environment variables on your machine. For instance, on MacOS/Linux you can use `export SCREENER_API_KEY=<Your screener key here>`.
 
-When ran locally, not all the tests may be needed to run, only the ones added/edited. This can be achieved by changing the regexp in `screener.config.js`, where the `states` property is created. Make sure not to commit that change though! Results can be viewed in the Screener online dashboard, among all the other runs (for master, pull requests, etc.)
+When ran locally, not all the tests may be needed to run, only the ones added/edited. This can be achieved by changing the regexp in `screener.config.ts`, where the `states` property is created. Make sure not to commit that change though! Results can be viewed in the Screener online dashboard, among all the other runs (for master, pull requests, etc.)
```
layershifter:
```suggestion
When ran locally, not all the tests may be needed to run, only the ones added/edited. This can be achieved by changing the regexp in `screener.config.js`, where the `states` property is created. Make sure not to commit that change though! Results can be viewed in the Screener online dashboard, among all the other runs (for master, pull requests, etc.)
```

layershifter (approved):

PR #1046, opened 2019-03-12
chore(package): update @types and TS

layershifter:
Extracted from #1042.


package.json
```diff
@@ -128,7 +128,7 @@
     "tsconfig-paths": "^3.7.0",
     "tslint": "^5.11.0",
     "tslint-config-airbnb": "^5.11.1",
-    "typescript": "~3.2.2",
+    "typescript": "^3.3.3333",
```
bmdalex, line 131: yaaay 👍

packages/react/src/components/Flex/FlexItem.tsx
```diff
@@ -80,9 +80,8 @@ class FlexItem extends UIComponent<Extendable<FlexItemProps>> {
       })
     }
 
-    return _.isNil(children)
-      ? children
-      : applyStyles(React.Children.only(children), styles, classes)
+    if (_.isNil(children)) return children
```
bmdalex, line 83: we cannot use ternary operator anymore or was it your preference here?

bmdalex (approved):
looks good, hope it's safe to make this change at this time though 👍

packages/react/src/components/Flex/FlexItem.tsx
layershifter: It becomes unreadable:

```js
return _.isNil(children)
  ? children
  : applyStyles(
    React.Children.only(children) as React.ReactElement<any>,
    styles, 
    classes,
  )
```

PR #1052, opened 2019-03-12
feat(events): add EventListener and StackableEventListener components

layershifter:
Extracted from #949.

This PR provides the first part of updated event handling.

### `<EventListener />`

Provides declarative abstraction for `.addEventListener()`, it's using the React's lifecycle to bind and unbind at the right time. Similar to [`react-event-listener`](https://github.com/oliviertassinari/react-event-listener).

### `<StackableEventListener />`

The sames as `EventListener`, but will notify only on the last mounted component.

```js
<StackableEventListener />
<StackableEventListener /> // this will be noticed
```

### Example

```jsx
import { documentRef, EventListener } from '@stardust-ui/react-component-event-listener'

// ---

<EventListener listener={(e) => console.log('CLICK', e)} targetRef={documentRef} type="click" />
```

### Hooks story

It's ready to be converted to be Hooks based, but currently we don't want to increase React's version in peer dependencies.

UPD: I restored hooks under `/hooks` directory, they aren't exported. Saved to switch to them shortly.


packages/react-component-event-listener/src/listenerRegistries.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  registries[type].add(listener)
+}
+
+export const remove = (type: EventTypes, listener: EventHandler<EventTypes>) => {
+  if (registries[type]) {
+    registries[type].delete(listener)
+
+    if (registries[type].size === 0) {
+      registries[type] = null
```
miroslavstastny:
```suggestion
      delete registries[type]
```

packages/react-component-event-listener/src/listenerRegistries.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+  }
+}
+
+export const dispatchable = (type: EventTypes, listener: EventHandler<EventTypes>) => {
```
miroslavstastny: `isDispatchable`

miroslavstastny (approved):

PR #1054, opened 2019-03-12
fix(dropdown): change header and content types from string to shorthand

silviuaavram:
Requested by People team.

Also to be in sync with the changes of the prop types in list.

made `content` and `header` props from `DropdownItem` and `header` from `DropdownSelectedItem` to be shorthand.


packages/react/src/components/Dropdown/DropdownSelectedItem.tsx
```diff
@@ -25,7 +25,7 @@ export interface DropdownSelectedItemProps extends UIComponentProps<DropdownSele
   active?: boolean
 
   /** Header of the selected item. */
-  header?: string
+  header?: ShorthandValue
```
mnajdova, line 27: Be careful with this one, as the Label's content is not shorthand value. You may need to create the element here before sending it to the `Label` component.

silviuaavram: thank you, will take a look.I just drafted changes for now, as People Team requested a change here.

packages/react/src/components/Dropdown/DropdownItem.tsx
```diff
@@ -19,10 +19,10 @@ export interface DropdownItemProps extends UIComponentProps<DropdownItemProps> {
   active?: boolean
 
   /** Item's content. */
-  content?: string
+  content?: ShorthandValue
```
bmdalex, line 24: pls also export:
```typescript
export DropdownItem.slotClassNames = {
  content: `${DropdownItem.className}__content`,
  header: `${DropdownItem.className}__header`,
}
```
same as it's done here: https://github.com/stardust-ui/react/blob/7bd74262315cdc03679662be8551b9fa5e3b9eb4/packages/react/src/components/Dropdown/Dropdown.tsx#L1104

There are users that reported that they cannot target the header/content in order to style them so this PR is the best opportunity to add this. Otherwise, pls. let me know once it's fixed and I'll create issue + PR

fyi @mnajdova

packages/react/src/components/Dropdown/DropdownItem.tsx
```diff
@@ -85,6 +93,12 @@ class DropdownItem extends UIComponent<ReactProps<DropdownItemProps>, any> {
   }
 }
 
+DropdownItem.slotClassNames = {
+  content: `${DropdownItem.className}__content`,
```
bmdalex, line 109: you need to pass these as `className` in the `defaultProps` of the slot you're creating:
e.g.:
```typescript
className: Dropdown.slotClassNames.searchInput
```
from https://github.com/stardust-ui/react/blob/b50927bbdf51797ea5596e3d825d3b343db9b85f/packages/react/src/components/Dropdown/Dropdown.tsx#L503

packages/react/src/components/Dropdown/DropdownItem.tsx
```diff
@@ -70,21 +79,36 @@ class DropdownItem extends UIComponent<ReactProps<DropdownItemProps>, any> {
     return (
       <ListItem
         className={classes.root}
-        header={header}
+        header={Box.create(header, {
+          defaultProps: {
+            className: DropdownItem.slotClassNames.header,
```
mnajdova, line 83: Can we apply here also: `styles.header`? That way people will be able to define a slot in the styles for the DropdownItem if they need some overrides. The same applies for the `media` and `content`

packages/react/src/components/Dropdown/DropdownSelectedItem.tsx
```diff
@@ -126,13 +129,20 @@ class DropdownSelectedItem extends UIComponent<ReactProps<DropdownSelectedItemPr
...
     const imageElement = Image.create(image, {
       defaultProps: {
         avatar: true,
+        className: DropdownSelectedItem.slotClassNames.image,
+      },
+    })
+    const contentElement = Box.create(header, {
+      defaultProps: {
+        as: 'span',
+        className: DropdownSelectedItem.slotClassNames.header,
```
mnajdova, line 143: Same here, please add `styles.content`, `styles.image` and `styles.icon` to the appropriate slots. The first way of styling we want to advertise is the slot's based. The `classNames` should be used just as an addition to this, not first way of styling slots.

packages/react/src/components/Dropdown/DropdownItem.tsx
```diff
@@ -55,8 +63,8 @@ class DropdownItem extends UIComponent<ReactProps<DropdownItemProps>, any> {
     }),
     accessibilityItemProps: PropTypes.object,
     active: PropTypes.bool,
-    content: PropTypes.string,
-    header: PropTypes.string,
+    content: customPropTypes.itemShorthand,
```
bmdalex, line 62: `content` and `header` need to become an actual shorthand so you need to use the shorthand API for this when passing them to `ListItem`. See how we do it in `ListItem.tsx` or follow the snippet below:

```jsx
<ListItem
  className={classes.root}
  header={Box.create(header, {
    defaultProps: {
      className: DropdownItem.slotClassNames.header,
      styles: styles.header,
    },
  })}
  onClick={this.handleClick}
  media={Image.create(image, {
    defaultProps: {
      className: DropdownItem.slotClassNames.image,
      avatar: true,
    },
  })}
  content={Box.create(content, {
    defaultProps: {
      className: DropdownItem.slotClassNames.content,
      styles: styles.content,
    },
  })}
  {...accessibilityItemProps}
  {...unhandledProps}
/>
```

In this way we don't need `slotClassNames` to be exported because user can target the slots directly in order to add their custom styles:

```jsx
<Provider
  theme={{
    componentStyles: {
      DropdownItem: {
        root: { /* custom styles */},
        header: { /* custom styles */},
        content: { /* custom styles */},
      },
    },
  }}
>
```

@mnajdova can you confirm?

bmdalex (dismissed):

docs/src/examples/components/Dropdown/Types/DropdownExampleMultiple.shorthand.steps.ts
```diff
@@ -5,7 +5,7 @@ const selectors = {
   item: (itemIndex: number) => `.${Dropdown.slotClassNames.itemsList} li:nth-child(${itemIndex})`,
   removeItemIcon: (itemIndex: number) =>
     `.${Dropdown.slotClassNames.selectedItems} span:nth-child(${itemIndex}) .${
-      DropdownSelectedItem.slotClassNames.removeIcon
+      DropdownSelectedItem.slotClassNames.icon
```
mnajdova, line 8: Please add **BREAKING CHANGE** section in the changelog, as you renamed the slot class name icon and in the PR description add steps of how the breaking changes should be resolved (mainly just replace the usages of  `DropdownSelectedItem.slotClassNames.removeIcon` with `DropdownSelectedItem.slotClassNames.icon`

mnajdova (approved):

PR #1057, opened 2019-03-13
feat(theme): add box-sizing to the global styles (Teams theme)

mnajdova:
This PR adds the `box-sizing: border-box` to all elements, as well before and auto pseudo elements in teams theme, by providing a rule in the global styles. Parts of the styles of the components, had to be updated. We can still see some screener differences, but I see these as fixes, rather then regressions. @codepretty @bcalvery @notandrew let me know what you think.

### BREAKING CHANGES:
The `box-sizing: border-box' prop is added for all elements, so check for regressions to all components that has style overrides affecting the size of the components, or adding paddings and margings of the elements.


packages/react/src/themes/base/components/Loader/loaderStyles.ts
```diff
@@ -52,8 +52,6 @@ export default {
       borderStyle: 'solid',
       borderWidth: v.borderSizes[p.size],
 
-      boxSizing: 'border-box',
-
```
layershifter: Not sure about this, it's defined in Base theme

layershifter (approved):

notandrew (approved):

packages/react/src/themes/base/components/Loader/loaderStyles.ts
mnajdova: My mistake, thought it was in team's theme. Will revert it.

bmdalex:
pls remove `boxSizing` style in `alertStyles.ts` before merging: see https://github.com/stardust-ui/react/pull/1063#discussion_r266396254

packages/react/src/themes/teams/components/Dropdown/dropdownVariables.ts
```diff
@@ -30,12 +30,12 @@ export default (siteVars): DropdownVariables => ({
...
   color: siteVars.bodyColor,
   comboboxPaddingButton: `0 ${_12px_asRem}`,
-  comboboxFlexBasis: '50px',
+  comboboxFlexBasis: pxToRem(50),
   listBackgroundColor: siteVars.colors.white,
   listBorderRadius: _3px_asRem,
   listPadding: `${pxToRem(8)} 0`,
   listBoxShadow: `0 .2rem .6rem 0 ${siteVars.gray06}`,
-  listMaxHeight: '20rem',
+  listMaxHeight: pxToRem(296),
```
miroslavstastny, line 38: just curious how 20rem became 296px?

miroslavstastny (approved):

packages/react/src/themes/teams/components/Dropdown/dropdownVariables.ts
mnajdova: So previously on the Dropdown list, we had padding on the top and 6 elements in the list (one list item has height of 48px) => (6*48px) + 8px = 296px :) We don't want to use rem anyway and the height was a bit different because of the `box-sizing`, so that is why I converted it to px.

codepretty (approved):

packages/react/src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -107,7 +107,7 @@ const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
       speak: 'none',
       verticalAlign: 'middle',
 
-      ...(!isFontBased && { boxSizing: 'border-box' }),
+      ...(isFontBased && { boxSizing: 'content-box' }),
```
kuzhelov: would suggest to make it part of the `getFontStyles` result - by putting it near the following related props:

```js
return {
    ...

    boxSizing: 'content-box',
    width: sizeInRems,
    height: sizeInRems,

   ...
  }
```

kuzhelov (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### BREAKING CHANGES
+- added `box-sizing: border-box` to all elements, as well as before and after pseudo elements in Teams theme @mnajdova ([#1057](https://github.com/stardust-ui/react/pull/1057))
```
kuzhelov: nit-nit: lets capitalize this entry, as well as provide it in the Present tense: `Add box-sizing: border-box to ..` - just to be consistent with CHANGELOG contents

miroslavstastny (approved):

Issue #936, opened 2019-02-20
Access modifiers and order of members in React components


bmdalex:
# Access modifiers and order of members in React components

The goal of this issue is to come up with certain guidelines and best practices for writing React components when it comes to:

### Arguments for:
- we're using TypeScript as main language for this project so we strive to leverage functionality (access modifiers in this case);
- showcase the public API;
- helpful for avoiding dead code (it's easy to forget about an unused method that is `public` by default and it won't be detected by the linter if we don't use the `private` modifier);

### Arguments against:
- closer to ES6 and React standards with ES6
- easier to adopt by JS devs to it includes a larger community
- possibility that with hooks we won't write as many class components

### a. React convention:

- described [here](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md):
    1. static methods and properties
    2. lifecycle methods
    3. custom methods
    4. `render` method

- looks like the generic convention adopted by the React community

### b. TypeScript / OOP convention:

- simplified version (static members come before instance members):
  1. private members
  2. public members
  3. constructor
  4. public methods
  5. private methods

- we're writing TypeScript code so we should take this into account when making a decision;

- is having events handlers before `render` method a good way to structure it necessarily?
One might argue that when you read the code you want to start with the entry points (the public methods, `render` in this case) and continue to scroll down as if it was a story, without too many jumps to the private methods that are called from the public ones (and constructor, that would also be placed at the beginning of the class).
_(full comment [here](https://github.com/stardust-ui/react/pull/900/files/418eb37a22e86f2b1f548d98b4a10761266b154e#r258443297))_

layershifter:
Convention is actually exists, we should just follow it: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md

My vote is to follow patterns that exists in React world to make the Stardust more friendly for community users.

levithomason:
Let's not spend cycles on this one.  Classifying as bikeshedding.  We have many important functional milestones to hit.  Let's get those in order and allow automation / fix to handle formatting things.

- We'll use remove TSLint for ESLint: #600 
- Render method last (community standard)
- Remove access modifiers (don't make sense in React)

Also, reiterating, let's not discuss this one anymore, but let's automate it, onward!  :)

jurokapsiar:
closing as there was no activity

PR #1063, opened 2019-03-15
feat(alert): v1

### LINKED ISSUE #1061: [RFC]: `Alert` | `Banner ` |  `Message` component
## Problem description

We need a component that can offer users a visual representation similar to an alert regarding different states of the application for specific scenarios:
- there's a connectivity issue in the app
- some users are not available to chat
- an item needs extra visual attention because it might be urgent, etc

## Proposed solutions

Other libraries tackle this problem by introducing a dedicated component. You can see some examples in the next section (**Other Libraries Screenshots**, after **Proposed solutions**)

### Name:
1. Alert (used by most librariies) - **front runner** 
2. Message (used by [SUIR](https://react.semantic-ui.com/collections/message/))
3. Banner (used by [Attlassian](https://atlaskit.atlassian.com/packages/core/banner))

### v1:
```typescript
export interface AlertProps {

  /** Button shorthand for the action slot. */
  action?: ShorthandValue<ButtonProps>

  /** Generic animation property for applying different theme animations. */
  animation?: AnimationProp

  /** Controls Alert's relation to neighboring items. */
  attached?: boolean | 'top' | 'bottom'

  /** Content for childrenApi. */
  children?: TChildren

  content?: TContent
 
  /** An alert may be formatted to display a danger message. */
  danger?: boolean

  /** An alert may be formatted to display information. */
  info?: boolean

  /** An alert may be formatted to display a successful message. */
  success?: boolean

  /** An alert may be formatted to display warning message. */
  warning?: boolean
}
```

### v2 (Auto Controlled Component):
```typescript
interface AlertPropsV2 extends AlertProps {
  /** The title of the alert **/
  header?: string

  /** An alert may be formatted to display an error or negative message. */
  error?: boolean
  negative?: boolean // EXPERIMENTAL - not sure we will add both

  /** An alert may be formatted to display a positive message. */
  positive?: boolean // EXPERIMENTAL - synonym for success  not sure we will add it
  
  /** Defines whether Alert is displayed. */
  open?: boolean

  /** Event for request to change 'open' value. */
  onOpenChange?: ComponentEventHandler<PopupProps>

  /** EXPERIMENTAL - some Alerts might need different style for all types/intents */
  appearance?: 'basic' | 'card'
}
```

### [Bootstrap](https://getbootstrap.com/docs/4.2/components/alerts/) - Alert:

- class names for describing the emphasis:`primary`, `secondary`, `success`, `danger`, `warning`, `info`, `light`, `dark` preceded by `alert-` suffix
- can have `header` and `content` (similar to Stardust's [`Header` with description](https://stardust-ui.github.io/react/components/header#variations-description))
- close button

### [Reactstrap](https://reactstrap.github.io/components/alerts/) - Alert:

- color:`primary`, `secondary`, `success`, `danger`, `warning`, `info`, `light`, `dark`
- can have `header` and `content` (similar to Stardust's [`Header` with description](https://stardust-ui.github.io/react/components/header#variations-description))
- `isOpen`: boolean prop for visibility state
- `toggle`: handler for controlling visibility
- `fade`: boolean prop for fade animation
- close button
- visually the same as Bootstrap

### [Ant Design](https://ant.design/components/alert/) - Alert:

Property | Description | Type | Default
-- | -- | -- | --
afterClose | Called when close animation is finished | () => void | -
banner | Whether to show as banner | boolean | false
closable | Whether Alert can be closed | boolean | -
closeText | Close text to show | string\|ReactNode | -
description | Additional content of Alert | string\|ReactNode | -
icon | Custom icon, effective when showIcon is true | ReactNode | -
message | Content of Alert | string\|ReactNode | -
showIcon | Whether to show icon | boolean | false, in bannermode default is true
type | Type of Alert styles, options: success, info, warning, error | string | info, in bannermode default is warning
onClose | Callback when Alert is closed | (e: MouseEvent) => void | -

### [Evergreen](https://evergreen.segment.com/components/alert/) - Alert:

Property | Type | Descriptio
-- | -- | --
child | renunion | The content of the alert. When a string is passed it is wrapped in a `<Text size={400} />` component.
intent | enum = 'none', 'success', 'warning', 'danger' | The intent of the alert.
title | node | The title of the alert.
hasTrim | bool = true | When true, show a border on the left matching the type.
hasIcon | bool = true | When true, show a icon on the left matching the type,
isRemoveable | bool = false | When true, show a remove icon button.
onRemove | func | Function called when the remove button is clicked.
appearance | enum = 'default', 'card' | The appearance of the alert.

### [VuetifyJs](https://vuetifyjs.com/en/components/alerts) - Alert:

Screenshot 2019-03-14 at 20 32 32

### [Semantic-UI React](https://react.semantic-ui.com/collections/message/) - Message:

Name |	Type |	Description
-- | -- | --
as | {custom}	| An element type to render as (string or function).
attached | {bool, enum}	| A message can be formatted to attach itself to other content, Enums: 'bottom', 'top'
children | {node}	| Primary content.
className | {string}	 | Additional classes.
color | {enum}	 | A message can be formatted to be different colors. Enums: 'red', 'orange', 'yellow', 'olive', 'green', 'teal', 'blue', 'violet', 'purple', 'pink', 'brown', 'grey', 'black'
compact | {bool}	| A message can only take up the width of its content.
content | {custom}	 | Shorthand for primary content.
error | {bool}	| A message may be formatted to display a negative message. Same as `negative`.
floating | {bool}	| A message can float above content that it is related to.
header | {custom}	| Shorthand for MessageHeader.
hidden | {bool}	 | A message can be hidden.
icon | {custom, bool}	| A message can contain an icon.
info |{bool}	 | A message may be formatted to display information.
list |{custom}	 |Array shorthand items for the MessageList. Mutually exclusive with children.
negative | {bool}	| A message may be formatted to display a negative message. Same as `error`.
onDismiss | {func}	| A message that the user can choose to hide. Called when the user clicks the "x" icon. This also adds the "x" icon.
positive | {bool}	| A message may be formatted to display a positive message. Same as `success`.
size | {enum}	| A message can have different sizes
success | {bool}	| A message may be formatted to display a positive message. Same as `positive`.
visible | {bool} |	A message can be set to visible to force itself to be shown.
warning |{bool}	|A message may be formatted to display warning messages.

### [Fabric](https://developer.microsoft.com/en-us/fabric#/components/messagebar#Variants) - MessageBar:

Screenshot 2019-03-14 at 20 47 14

### [Material UI](https://material-ui.com/demos/snackbars/) - Snackbar:

Name | Value | Description
-- | -- | --
action | node  | The action to display.
anchorOrigin | enum | The anchor of the Snackbar: { horizontal: enum: 'left' \| 'center' \| 'right', vertical: enum: 'top' \| 'bottom'} 
autoHideDuration | number | The number of milliseconds to wait before automatically calling the onClose function. onClose should then set the state of the open prop to hide the Snackbar. This behavior is disabled by default with the null value.
ClickAwayListenerProps | object |   | Properties applied to the ClickAwayListener element.
ContentProps | object |   | Properties applied to the SnackbarContent element.
message | node |   | The message to display.
onClose | func | Callback fired when the component requests to be closed. Typically onClose is used to set state in the parent component, which is used to control the Snackbar open prop. 
open | bool |   | If true, Snackbar is open.

### [Bootstrap](https://getbootstrap.com/docs/4.2/components/alerts/) and [Reactstrap](https://reactstrap.github.io/components/alerts/) - Alert:

Screenshot 2019-03-14 at 19 58 06

### [Ant Design](https://ant.design/components/alert/) - Alert:

Screenshot 2019-03-14 at 20 18 19

### [Evergreen](https://evergreen.segment.com/components/alert/) - Alert:

Screenshot 2019-03-14 at 20 23 25
Screenshot 2019-03-14 at 20 23 36

### [VuetifyJs](https://vuetifyjs.com/en/components/alerts) - Alert:

Screenshot 2019-03-14 at 20 30 01

### [Semantic-UI React](https://react.semantic-ui.com/collections/message/) - Message:
Screenshot 2019-03-14 at 20 34 09
Screenshot 2019-03-14 at 20 34 16
Screenshot 2019-03-14 at 20 34 24

### [Fabric](https://developer.microsoft.com/en-us/fabric#/components/messagebar#Variants) - MessageBar:

Screenshot 2019-03-14 at 20 45 58

### [Material UI](https://material-ui.com/demos/snackbars/) - Snackbar:
Screenshot 2019-03-15 at 00 57 36

bmdalex:
# Alert component V1

## Description

This PR resolves #1061 by implementing the `Alert` component.

### `info`

An alert may be formatted to display information.

Screenshot 2019-03-15 at 21 25 19

### `warning`

An alert may be formatted to display a warning message.

Screenshot 2019-03-15 at 21 25 30

### `success`

An alert may be formatted to display a successful message.

Screenshot 2019-03-15 at 21 25 38

### `danger`

An alert may be formatted to display a danger message.

Screenshot 2019-03-15 at 21 25 45

### `attached`

An Alert can be can be formatted to attach itself to other content.

Screenshot 2019-03-15 at 21 25 53

### `action`

An Alert can provide the user with an action.

Screenshot 2019-03-15 at 21 26 01

### DIFF
--- a/docs/src/examples/components/Alert/Rtl/index.tsx
+++ b/docs/src/examples/components/Alert/Rtl/index.tsx
@@ -1,0 +1,12 @@
+import * as React from 'react'
+
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import NonPublicSection from 'docs/src/components/ComponentDoc/NonPublicSection'
+
+const Rtl = () => (
+  <NonPublicSection title="Rtl">
+    <ComponentExample examplePath="components/Alert/Rtl/AlertExample.rtl" />
+  </NonPublicSection>
+)
+
+export default Rtl
--- a/docs/src/examples/components/Alert/Types/index.tsx
+++ b/docs/src/examples/components/Alert/Types/index.tsx
@@ -1,0 +1,15 @@
+import * as React from 'react'
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
+
+const Types = () => (
+  <ExampleSection title="Types">
+    <ComponentExample
+      title="Default"
+      description="A default Alert."
+      examplePath="components/Alert/Types/AlertExample"
+    />
+  </ExampleSection>
+)
+
+export default Types
--- a/docs/src/examples/components/Alert/Types/index.tsx
+++ b/docs/src/examples/components/Alert/Types/index.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react'
+
 import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
 import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
 
--- a/docs/src/examples/components/Alert/Variations/AlertExampleAttached.shorthand.tsx
+++ b/docs/src/examples/components/Alert/Variations/AlertExampleAttached.shorthand.tsx
@@ -1,0 +1,14 @@
+import * as React from 'react'
+import { Alert, Input } from '@stardust-ui/react'
+
+const AlertExampleShorthand = () => (
+  <>
+    <Alert attached content="This is a top attached alert" />
+    <Input fluid placeholder="Name..." />
+    <br /> <br />
+    <Input fluid placeholder="Surname..." />
+    <Alert attached="bottom" content="This is a bottom attached alert" />
+  </>
+)
+
+export default AlertExampleShorthand
--- a/packages/react/src/components/Alert/Alert.tsx
+++ b/packages/react/src/components/Alert/Alert.tsx
@@ -14,40 +14,33 @@ import {
 import { RenderResultConfig } from 'src/lib/renderComponent'
 import { defaultBehavior } from '../../lib/accessibility'
 import { Accessibility } from '../../lib/accessibility/types'
-import { ComponentEventHandler, ReactProps, ShorthandValue } from '../../types'
+import { ReactProps, ShorthandValue } from '../../types'
 import Box from '../Box/Box'
 import Button, { ButtonProps } from '../Button/Button'
 
 export interface AlertSlotClassNames {
   content: string
-  closeButton: string
+  action: string
 }
 
-export interface AlertProps extends UIComponentProps, ContentComponentProps {
+export interface AlertProps extends UIComponentProps, ContentComponentProps<ShorthandValue> {
   /**
    * Accessibility behavior if overridden by the user.
    * @default defaultBehavior
    */
   accessibility?: Accessibility
 
+  /** Button shorthand for the action slot. */
+  action?: ShorthandValue<ButtonProps>
+
   /** Controls Alert's relation to neighboring items. */
   attached?: boolean | 'top' | 'bottom'
-
-  /** Button used for dismissing the Alert. */
-  closeButton?: ShorthandValue<ButtonProps>
 
   /** An alert may be formatted to display a danger message. */
   danger?: boolean
 
   /** An alert may be formatted to display information. */
   info?: boolean
-
-  /**
-   * The callback used when the user chooses to hide the Alert.
-   * @param {SyntheticEvent} event - React's original SyntheticEvent.
-   * @param {object} data - All props and proposed value.
-   */
-  onClose?: ComponentEventHandler<AlertProps>
 
   /** An alert may be formatted to display a successful message. */
   success?: boolean
@@ -58,6 +51,9 @@ export interface AlertProps extends UIComponentProps, ContentComponentProps {
 
 /**
  * A Alert displays information that explains nearby content.
+ * @accessibility
+ * Other considerations:
+ *  - if Alert contains action slot, textual representation needs to be provided by using 'title', 'aria-label' or 'aria-labelledby' attributes
  */
 class Alert extends UIComponent<ReactProps<AlertProps>> {
   static displayName = 'Alert'
@@ -65,38 +61,24 @@ class Alert extends UIComponent<ReactProps<AlertProps>> {
 
   static slotClassNames: AlertSlotClassNames = {
     content: `${Alert.className}__content`,
-    closeButton: `${Alert.className}__closeButton`,
+    action: `${Alert.className}__action`,
   }
 
   static propTypes = {
-    ...commonPropTypes.createCommon(),
+    ...commonPropTypes.createCommon({ content: 'shorthand' }),
+    action: customPropTypes.itemShorthand,
     attached: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf(['top', 'bottom'])]),
-    closeButton: customPropTypes.itemShorthand,
     danger: PropTypes.bool,
     info: PropTypes.bool,
-    onClose: PropTypes.func,
     success: PropTypes.bool,
     warning: PropTypes.bool,
   }
 
-  static defaultProps = {
-    accessibility: defaultBehavior,
-    closeButton: { icon: 'close' },
-  }
-
-  handleCloseButtonOverrides = (predefinedProps: ButtonProps) => ({
-    onClick: (e: React.SyntheticEvent, buttonProps: ButtonProps) => {
-      this.handleOnClose(e)
-      _.invoke(predefinedProps, 'onClick', e, buttonProps)
-    },
-  })
-
-  handleOnClose = (e: React.SyntheticEvent) => {
-    _.invoke(this.props, 'onClose', e, this.props)
-  }
+  static defaultProps = { accessibility: defaultBehavior }
 
   renderContent = ({ styles }: RenderResultConfig<AlertProps>) => {
-    const { closeButton, content, onClose } = this.props
+    const { action, content } = this.props
+
     return (
       <>
         {Box.create(content, {
@@ -105,16 +87,14 @@ class Alert extends UIComponent<ReactProps<AlertProps>> {
             styles: styles.content,
           },
         })}
-        {onClose &&
-          Button.create(closeButton, {
-            defaultProps: {
-              iconOnly: true,
-              text: true,
-              className: Alert.slotClassNames.closeButton,
-              styles: styles.closeButton,
-            },
-            overrideProps: this.handleCloseButtonOverrides,
-          })}
+        {Button.create(action, {
+          defaultProps: {
+            iconOnly: true,
+            text: true,
+            className: Alert.slotClassNames.action,
+            styles: styles.action,
+          },
+        })}
       </>
     )
   }
@@ -122,6 +102,7 @@ class Alert extends UIComponent<ReactProps<AlertProps>> {
   renderComponent(config: RenderResultConfig<AlertProps>) {
     const { accessibility, classes, ElementType, unhandledProps } = config
     const { children, content } = this.props
+
     return (
       <ElementType
         className={classes.root}
--- a/packages/react/src/components/Alert/Alert.tsx
+++ b/packages/react/src/components/Alert/Alert.tsx
@@ -15,9 +15,13 @@ import { RenderResultConfig } from 'src/lib/renderComponent'
 import { defaultBehavior } from '../../lib/accessibility'
 import { Accessibility } from '../../lib/accessibility/types'
 import { ComponentEventHandler, ReactProps, ShorthandValue } from '../../types'
-import { Box, Icon } from '../..'
+import Box from '../Box/Box'
+import Button, { ButtonProps } from '../Button/Button'
 
-export type AlertSlotClassNames = Record<'content' | 'closeIcon', string>
+export interface AlertSlotClassNames {
+  content: string
+  closeButton: string
+}
 
 export interface AlertProps extends UIComponentProps, ContentComponentProps {
   /**
@@ -29,8 +33,8 @@ export interface AlertProps extends UIComponentProps, ContentComponentProps {
   /** Controls Alert's relation to neighboring items. */
   attached?: boolean | 'top' | 'bottom'
 
-  /** Icon used for dismissing the Alert, 'close' by default. */
-  closeIcon?: ShorthandValue
+  /** Button used for dismissing the Alert. */
+  closeButton?: ShorthandValue<ButtonProps>
 
   /** An alert may be formatted to display a danger message. */
   danger?: boolean
@@ -55,19 +59,19 @@ export interface AlertProps extends UIComponentProps, ContentComponentProps {
 /**
  * A Alert displays information that explains nearby content.
  */
-class Alert extends UIComponent<ReactProps<AlertProps>, {}> {
+class Alert extends UIComponent<ReactProps<AlertProps>> {
   static displayName = 'Alert'
   static className = 'ui-alert'
 
   static slotClassNames: AlertSlotClassNames = {
     content: `${Alert.className}__content`,
-    closeIcon: `${Alert.className}__closeIcon`,
+    closeButton: `${Alert.className}__closeButton`,
   }
 
   static propTypes = {
     ...commonPropTypes.createCommon(),
     attached: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf(['top', 'bottom'])]),
-    closeIcon: customPropTypes.itemShorthand,
+    closeButton: customPropTypes.itemShorthand,
     danger: PropTypes.bool,
     info: PropTypes.bool,
     onClose: PropTypes.func,
@@ -77,7 +81,7 @@ class Alert extends UIComponent<ReactProps<AlertProps>, {}> {
 
   static defaultProps = {
     accessibility: defaultBehavior,
-    closeIcon: 'close',
+    closeButton: { icon: 'close' },
   }
 
   renderComponent(config: RenderResultConfig<AlertProps>) {
@@ -95,8 +99,8 @@ class Alert extends UIComponent<ReactProps<AlertProps>, {}> {
     )
   }
 
-  private renderContent = ({ styles, variables }: RenderResultConfig<AlertProps>) => {
-    const { closeIcon, content, onClose } = this.props
+  renderContent = ({ styles }: RenderResultConfig<AlertProps>) => {
+    const { closeButton, content, onClose } = this.props
     return (
       <>
         {Box.create(content, {
@@ -106,28 +110,28 @@ class Alert extends UIComponent<ReactProps<AlertProps>, {}> {
           },
         })}
         {onClose &&
-          Icon.create(closeIcon, {
+          Button.create(closeButton, {
             defaultProps: {
-              className: Alert.slotClassNames.closeIcon,
-              styles: styles.icon,
-              variables: variables.icon,
+              iconOnly: true,
+              text: true,
+              className: Alert.slotClassNames.closeButton,
+              styles: styles.closeButton,
             },
-            overrideProps: this.handleCloseIconOverrides,
+            overrideProps: this.handleCloseButtonOverrides,
           })}
       </>
     )
   }
 
-  private handleCloseIconOverrides = (predefinedProps: AlertProps) => ({
-    onClick: (e: React.SyntheticEvent) => {
+  handleCloseButtonOverrides = (predefinedProps: ButtonProps) => ({
+    onClick: (e: React.SyntheticEvent, buttonProps: ButtonProps) => {
       this.handleOnClose(e)
-      _.invoke(predefinedProps, 'onClick', e, this.props)
+      _.invoke(predefinedProps, 'onClick', e, buttonProps)
     },
-    ...(predefinedProps.onClick && { tabIndex: '0' }),
   })
 
-  private handleOnClose = (e: React.SyntheticEvent) => {
-    _.invoke(this.props, 'onClose', e, { ...this.props })
+  handleOnClose = (e: React.SyntheticEvent) => {
+    _.invoke(this.props, 'onClose', e, this.props)
   }
 }
 
--- a/packages/react/src/components/Alert/Alert.tsx
+++ b/packages/react/src/components/Alert/Alert.tsx
@@ -84,19 +84,15 @@ class Alert extends UIComponent<ReactProps<AlertProps>> {
     closeButton: { icon: 'close' },
   }
 
-  renderComponent(config: RenderResultConfig<AlertProps>) {
-    const { accessibility, classes, ElementType, unhandledProps } = config
-    const { children, content } = this.props
-    return (
-      <ElementType
-        className={classes.root}
-        {...accessibility.attributes.root}
-        {...rtlTextContainer.getAttributes({ forElements: [children, content] })}
-        {...unhandledProps}
-      >
-        {childrenExist(children) ? children : this.renderContent(config)}
-      </ElementType>
-    )
+  handleCloseButtonOverrides = (predefinedProps: ButtonProps) => ({
+    onClick: (e: React.SyntheticEvent, buttonProps: ButtonProps) => {
+      this.handleOnClose(e)
+      _.invoke(predefinedProps, 'onClick', e, buttonProps)
+    },
+  })
+
+  handleOnClose = (e: React.SyntheticEvent) => {
+    _.invoke(this.props, 'onClose', e, this.props)
   }
 
   renderContent = ({ styles }: RenderResultConfig<AlertProps>) => {
@@ -123,15 +119,19 @@ class Alert extends UIComponent<ReactProps<AlertProps>> {
     )
   }
 
-  handleCloseButtonOverrides = (predefinedProps: ButtonProps) => ({
-    onClick: (e: React.SyntheticEvent, buttonProps: ButtonProps) => {
-      this.handleOnClose(e)
-      _.invoke(predefinedProps, 'onClick', e, buttonProps)
-    },
-  })
-
-  handleOnClose = (e: React.SyntheticEvent) => {
-    _.invoke(this.props, 'onClose', e, this.props)
+  renderComponent(config: RenderResultConfig<AlertProps>) {
+    const { accessibility, classes, ElementType, unhandledProps } = config
+    const { children, content } = this.props
+    return (
+      <ElementType
+        className={classes.root}
+        {...accessibility.attributes.root}
+        {...rtlTextContainer.getAttributes({ forElements: [children, content] })}
+        {...unhandledProps}
+      >
+        {childrenExist(children) ? children : this.renderContent(config)}
+      </ElementType>
+    )
   }
 }
 
--- a/packages/react/src/components/Alert/Alert.tsx
+++ b/packages/react/src/components/Alert/Alert.tsx
@@ -1,0 +1,134 @@
+import * as _ from 'lodash'
+import * as PropTypes from 'prop-types'
+import * as React from 'react'
+
+import {
+  UIComponent,
+  UIComponentProps,
+  ContentComponentProps,
+  commonPropTypes,
+  customPropTypes,
+  childrenExist,
+  rtlTextContainer,
+} from '../../lib'
+import { RenderResultConfig } from 'src/lib/renderComponent'
+import { defaultBehavior } from '../../lib/accessibility'
+import { Accessibility } from '../../lib/accessibility/types'
+import { ComponentEventHandler, ReactProps, ShorthandValue } from '../../types'
+import { Box, Icon } from '../..'
+
+export type AlertSlotClassNames = Record<'content' | 'closeIcon', string>
+
+export interface AlertProps extends UIComponentProps, ContentComponentProps {
+  /**
+   * Accessibility behavior if overridden by the user.
+   * @default defaultBehavior
+   */
+  accessibility?: Accessibility
+
+  /** Controls Alert's relation to neighboring items. */
+  attached?: boolean | 'top' | 'bottom'
+
+  /** Icon used for dismissing the Alert, 'close' by default. */
+  closeIcon?: ShorthandValue
+
+  /** An alert may be formatted to display a danger message. */
+  danger?: boolean
+
+  /** An alert may be formatted to display information. */
+  info?: boolean
+
+  /**
+   * The callback used when the user chooses to hide the Alert.
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props and proposed value.
+   */
+  onClose?: ComponentEventHandler<AlertProps>
+
+  /** An alert may be formatted to display a successful message. */
+  success?: boolean
+
+  /** An alert may be formatted to display a warning message. */
+  warning?: boolean
+}
+
+/**
+ * A Alert displays information that explains nearby content.
+ */
+class Alert extends UIComponent<ReactProps<AlertProps>, {}> {
+  static displayName = 'Alert'
+  static className = 'ui-alert'
+
+  static slotClassNames: AlertSlotClassNames = {
+    content: `${Alert.className}__content`,
+    closeIcon: `${Alert.className}__closeIcon`,
+  }
+
+  static propTypes = {
+    ...commonPropTypes.createCommon(),
+    attached: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf(['top', 'bottom'])]),
+    closeIcon: customPropTypes.itemShorthand,
+    danger: PropTypes.bool,
+    info: PropTypes.bool,
+    onClose: PropTypes.func,
+    success: PropTypes.bool,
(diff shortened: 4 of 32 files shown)


packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,134 @@
...
+      this.handleOnClose(e)
+      _.invoke(predefinedProps, 'onClick', e, this.props)
+    },
+    ...(predefinedProps.onClick && { tabIndex: '0' }),
+  })
+
+  private handleOnClose = (e: React.SyntheticEvent) => {
+    _.invoke(this.props, 'onClose', e, { ...this.props })
+  }
+}
```
layershifter, line 117: https://github.com/stardust-ui/react/issues/936#issuecomment-470205650

docs/src/examples/components/Alert/Types/index.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
```
layershifter, line 3:
```suggestion

import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
```

_nit_

docs/src/examples/components/Alert/Variations/AlertExampleAttached.shorthand.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { Alert, Input } from '@stardust-ui/react'
+
+const AlertExampleShorthand = () => (
+  <>
+    <Alert attached content="This is a top attached alert" />
+    <Input fluid placeholder="Name..." />
+    <br /> <br />
+    <Input fluid placeholder="Surname..." />
+    <Alert attached="bottom" content="This is a bottom attached alert" />
+  </>
```
layershifter, line 11: Can we use `Flex` with `gap` there?

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,134 @@
...
+  commonPropTypes,
+  customPropTypes,
+  childrenExist,
+  rtlTextContainer,
+} from '../../lib'
+import { RenderResultConfig } from 'src/lib/renderComponent'
+import { defaultBehavior } from '../../lib/accessibility'
+import { Accessibility } from '../../lib/accessibility/types'
+import { ComponentEventHandler, ReactProps, ShorthandValue } from '../../types'
+import { Box, Icon } from '../..'
```
layershifter: Please use direct imports to avoid circular dependencies:

```
import Box from '../Box/Box`
import Icon from '../Icon/Icon`
```

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+export type AlertSlotClassNames = Record<'content' | 'closeIcon', string>
```
layershifter: I like `Record`, but let's be consistent with other components there and use `interface`

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,134 @@
...
+   * Accessibility behavior if overridden by the user.
+   * @default defaultBehavior
+   */
+  accessibility?: Accessibility
+
+  /** Controls Alert's relation to neighboring items. */
+  attached?: boolean | 'top' | 'bottom'
+
+  /** Icon used for dismissing the Alert, 'close' by default. */
+  closeIcon?: ShorthandValue
```
layershifter:
```suggestion
  closeIcon?: ShorthandValue<IconProps>
```

Please :)

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,134 @@
...
+  success?: boolean
+
+  /** An alert may be formatted to display a warning message. */
+  warning?: boolean
+}
+
+/**
+ * A Alert displays information that explains nearby content.
+ */
+class Alert extends UIComponent<ReactProps<AlertProps>, {}> {
```
layershifter:
```suggestion
class Alert extends UIComponent<ReactProps<AlertProps>> {
```

_nit_

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,134 @@
...
+      </>
+    )
+  }
+
+  private handleCloseIconOverrides = (predefinedProps: AlertProps) => ({
+    onClick: (e: React.SyntheticEvent) => {
+      this.handleOnClose(e)
+      _.invoke(predefinedProps, 'onClick', e, this.props)
+    },
+    ...(predefinedProps.onClick && { tabIndex: '0' }),
```
layershifter: Oops, should be this part of accessibility behavior? Can we add todo in this case?

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,134 @@
...
+              variables: variables.icon,
+            },
+            overrideProps: this.handleCloseIconOverrides,
+          })}
+      </>
+    )
+  }
+
+  private handleCloseIconOverrides = (predefinedProps: AlertProps) => ({
+    onClick: (e: React.SyntheticEvent) => {
```
layershifter:
```suggestion
    onClick: (e: React.SyntheticEvent, iconProps: IconProps) => {
```

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,134 @@
...
+      this.handleOnClose(e)
+      _.invoke(predefinedProps, 'onClick', e, this.props)
```
layershifter:
```suggestion
      _.invoke(predefinedProps, 'onClick', e, iconProps)
```

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,134 @@
...
+              styles: styles.icon,
+              variables: variables.icon,
+            },
+            overrideProps: this.handleCloseIconOverrides,
+          })}
+      </>
+    )
+  }
+
+  private handleCloseIconOverrides = (predefinedProps: AlertProps) => ({
```
layershifter:
```suggestion
  private handleCloseIconOverrides = (predefinedProps: IconProps) => ({
```

packages/react/src/themes/teams/components/Alert/alertStyles.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+const alertStyles: ComponentSlotStylesInput<AlertProps, AlertVariables> = {
+  root: ({ props: p, variables: v, theme: { siteVariables } }): ICSSInJSStyle => {
+    return {
+      display: 'flex',
+      position: 'relative',
+      width: '100%',
+      boxSizing: 'border-box',
+      border: '1px solid',
```
layershifter: We need use variable and pxToRem :)

packages/react/src/themes/teams/components/Alert/alertStyles.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+}
+
+const alertStyles: ComponentSlotStylesInput<AlertProps, AlertVariables> = {
+  root: ({ props: p, variables: v, theme: { siteVariables } }): ICSSInJSStyle => {
+    return {
+      display: 'flex',
+      position: 'relative',
+      width: '100%',
+      boxSizing: 'border-box',
```
layershifter: Can we add there a TODO item for https://github.com/stardust-ui/react/pull/1057?

packages/react/src/themes/teams/components/Alert/alertStyles.ts
```diff
@@ -0,0 +1,89 @@
...
+      backgroundColor: colors.grey[50], // $app-white
+      borderColor: naturalColors.lightGreen[900], // $app-green
+    }
+  }
+
+  if (p.warning) {
+    return {
+      color: colors.grey[700],
+      backgroundColor: colors.grey[50], // $app-white
+      borderColor: colors.yellow[900], // $app-yellow
```
layershifter, line 47: I think that this should this part of variables:
```
Record<'info', 'danger', { color: string, backrgroundColor: string }>
```

layershifter:
NIT: I'm not sure that that `info` should be `grey`, I think that it should be `blue` actually... 🤔

packages/react/src/themes/teams/colors.ts
```diff
@@ -39,16 +39,16 @@ export const naturalColors: NaturalColors = {
...
-    100: '#E6E6E6',
-    200: '#CDCCCC',
-    300: '#B8B8B8',
-    400: '#A2A2A2',
-    500: '#8C8C8C',
-    600: '#747373',
-    700: '#5F5E5E',
-    800: '#404040',
-    900: '#252424', // siteVariables.black
+    50: '#FFFFFF', // $app-white
```
mnajdova: I would restrain changing the colors in the palette for now, it should be design responsibility. Were all these changed colors not even used?

mnajdova:
@layershifter that was my first thought as well, but as the changes are implemented in the teams theme, then whatever are the requirements for that theme should be there. If the changes were in base theme, then we could have argue about what the colors should be. I wouldn't change this currently.

docs/src/examples/components/Alert/Rtl/index.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react'
+
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import NonPublicSection from 'docs/src/components/ComponentDoc/NonPublicSection'
+
+const Rtl = () => (
+  <NonPublicSection title="Rtl">
```
mnajdova, line 7: `NonPublicSection` <3 this, I missed when this was added. Good job!

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,134 @@
...
+} from '../../lib'
+import { RenderResultConfig } from 'src/lib/renderComponent'
+import { defaultBehavior } from '../../lib/accessibility'
+import { Accessibility } from '../../lib/accessibility/types'
+import { ComponentEventHandler, ReactProps, ShorthandValue } from '../../types'
+import { Box, Icon } from '../..'
+
+export type AlertSlotClassNames = Record<'content' | 'closeIcon', string>
+
+export interface AlertProps extends UIComponentProps, ContentComponentProps {
```
mnajdova: I know I am jumping a bit late on this, but would expect the Alert component to have title as well.

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,134 @@
...
+  /**
+   * Accessibility behavior if overridden by the user.
+   * @default defaultBehavior
+   */
+  accessibility?: Accessibility
+
+  /** Controls Alert's relation to neighboring items. */
+  attached?: boolean | 'top' | 'bottom'
```
mnajdova, line 37: 👍

docs/src/examples/components/Alert/Variations/AlertExampleAttached.shorthand.tsx
bmdalex: we can but code will double in size for the same effect:
```jsx
const AlertExampleShorthand = () => (
  <Flex gap="gap.smaller" column>
    <Flex.Item>
      <>
        <Alert attached content="This is a top attached alert" />
        <Input fluid placeholder="Name..." />
      </>
    </Flex.Item>
    <Flex.Item>
      <>
        <Input fluid placeholder="Surname..." />
        <Alert attached="bottom" content="This is a bottom attached alert" />
      </>
    </Flex.Item>
  </Flex>
)
```
=> 14 lines vs 7 lines currently:
```jsx
const AlertExampleShorthand = () => (
  <>
    <Alert attached content="This is a top attached alert" />
    <Input fluid placeholder="Name..." />
    <br /> <br />
    <Input fluid placeholder="Surname..." />
    <Alert attached="bottom" content="This is a bottom attached alert" />
  </>
)
```
still wanna do it?

packages/react/src/components/Alert/Alert.tsx
bmdalex: hmm..ok, this is what I get with automatic imports, why is that?

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,134 @@
...
+    danger: PropTypes.bool,
+    info: PropTypes.bool,
+    onClose: PropTypes.func,
+    success: PropTypes.bool,
+    warning: PropTypes.bool,
+  }
+
+  static defaultProps = {
+    accessibility: defaultBehavior,
+    closeIcon: 'close',
```
mnajdova: I would really try to avoid using default values for icon's name as they are part of the theme, but I don't have better proposal at this moment.. We have the issue in other components as well, I am just expressing my thoughts here..

packages/react/src/themes/teams/siteVariables.ts
```diff
@@ -51,17 +54,16 @@ export const fontSizes = {
...
-export const infoStatusTextColor = colors.grey[50]
-export const warningStatusBackgroundColor = colors.yellow[900]
-export const warningStatusTextColor = colors.grey[50]
-export const errorStatusBackgroundColor = colors.red[900]
-export const errorStatusTextColor = colors.grey[50]
-export const unknownStatusBackgroundColor = gray04
-export const unknownStatusTextColor = colors.grey[50]
-
+export const successStatusBackgroundColor = colors.grey[50] // $app-white
+export const successStatusTextColor = colors.green[900] // $app-green-04
```
mnajdova: Thanks for adding comments about the corresponding colors from design!

packages/react/src/components/Alert/Alert.tsx
bmdalex: I will, but in this way how are we going to use new TS features? :)
have a PR for all project dedicated just to changing interfaces that have the same type for value to record?

packages/react/src/components/Alert/Alert.tsx
layershifter: We decided to skip this for first implementation. There is also open discussion for `title` vs `header`...

packages/react/src/components/Alert/Alert.tsx
bmdalex: I removed it as we need to use `<Button iconOnly text icon={closeIcon}>`, but I am not sure now how to handle the `closeIcon` prop; I think it needs to become `closeButton` and use shorthand; what do you think?

packages/react/src/components/Alert/Alert.tsx
layershifter: Personally I think, that yes it will be better to have one PR for it. And introduce specific type, like `SlotClassNames<T>` :)

docs/src/examples/components/Alert/Variations/AlertExampleAttached.shorthand.tsx
layershifter: Valid point, let's keep it as is

packages/react/src/themes/teams/components/Alert/alertStyles.ts
bmdalex: agreed but our convention is to have a flat structure for variables for now; we had this discussion before in #756 (see: https://github.com/stardust-ui/react/pull/756#discussion_r249983095)

@levithomason looks like the need for nested variables is back

packages/react/src/components/Alert/Alert.tsx
layershifter: Automatic imports sometimes are weird, this case I think you got it because `Box` in `src/index.ts` is named import while in `src/Box/Box.ts` it's default.

packages/react/src/components/Alert/Alert.tsx
layershifter: If it represents `Button`, I'm fine to have `closeButton`

packages/react/src/components/Alert/Alert.tsx
bmdalex: we thought `content` is enough for v1, we plan to add it in v2
see #1061 description and

> skip header and statusIcon/icon for initial implementation
from https://github.com/stardust-ui/react/issues/1061#issuecomment-473248610

@mnajdova we can reopen the discussion now if you feel there is a need to support this even in v1; fyi @layershifter

packages/react/src/themes/teams/components/Alert/alertStyles.ts
bmdalex: to be removed in https://github.com/stardust-ui/react/pull/1057#issuecomment-473872372

packages/react/src/themes/teams/components/Alert/alertStyles.ts
layershifter: So, may be we can just use something like `colorBackrgroundRootDanger`, https://amzn.github.io/style-dictionary/#/properties?id=category-type-item

@mnajdova do you have any agreement on variables?

mnajdova: I am starting the work for refactoring the color palette and color scheme. For now I would rather leave it as is, and would revert the changed color palette, so that we will have easier migration from here.

packages/react/src/components/Alert/Alert.tsx
bmdalex: will follow up on it, thanks!

packages/react/src/components/Alert/Alert.tsx
bmdalex: 😿 same thoughts exactly but we need this right now

packages/react/src/components/Alert/Alert.tsx
bmdalex: done..I really hope ESlint will detect unused code

packages/react/src/components/Alert/Alert.tsx
bmdalex: done, thx

packages/react/src/components/Alert/Alert.tsx
bmdalex: done, thx

packages/react/src/components/Alert/Alert.tsx
bmdalex: changed to `closeButton`

packages/react/src/components/Alert/Alert.tsx
bmdalex: done, thx

packages/react/src/components/Alert/Alert.tsx
bmdalex: done, thx

packages/react/src/themes/teams/components/Alert/alertStyles.ts
bmdalex: done, thx

packages/react/src/components/Alert/Alert.tsx
bmdalex: done, thx

packages/react/src/components/Alert/Alert.tsx
bmdalex: done, thx

packages/react/src/components/Alert/Alert.tsx
bmdalex: done, thx

docs/src/examples/components/Alert/Types/index.tsx
bmdalex: thx, done

packages/react/src/components/Alert/Alert.tsx
bmdalex: let's not forget `title` is an HTML attribute

packages/react/src/themes/teams/components/Alert/alertStyles.ts
bmdalex: @layershifter 
I believe for the users it's easier to edit the `backgroundColor` variable
```jsx
<Alert variables={{ backgroundColor: 'red' }} ... />
```

rather than adding the `danger` prop and a strange variable name like `colorBackrgroundRootDanger`:
```jsx
<Alert danger variables={{ colorBackrgroundRootDanger: 'red' }} ... />
```

In this way, we also avoid introducing **15 variables** from **5 variations** (default, info, danger, success, warning) X **3 colors** (text, background, border). Instead we introduce just 3 variables (`color`, `backgroundColor`, `borderColor`),

So I'd also vote to keep it like this.

layershifter: Agree, let's skip it for now.

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  handleCloseButtonOverrides = (predefinedProps: ButtonProps) => ({
+    onClick: (e: React.SyntheticEvent, buttonProps: ButtonProps) => {
+      this.handleOnClose(e)
+      _.invoke(predefinedProps, 'onClick', e, buttonProps)
+    },
+  })
+
+  handleOnClose = (e: React.SyntheticEvent) => {
```
layershifter: Can you please also:
- place `handleCloseButtonOverrides` and `handleOnClose` before `renderComponent`
- place `renderContent` before `renderComponent`

---

- `handleCloseButtonOverrides`
- `handleOnClose`
- `renderContent`
- `renderComponent`

layershifter (approved):
👍

packages/react/src/themes/teams/colors.ts
bmdalex: reverted changes in colors, should be ok now

packages/react/src/components/Alert/Alert.tsx
bmdalex: done

packages/react/src/components/Alert/Alert.tsx
mnajdova: Let's leave it for a next iteration (I agree header is better name)

packages/react/src/themes/teams/siteVariables.ts
```diff
@@ -25,7 +25,10 @@ export const brand08 = '#8B8CC7' // no mapping color
 
 export const magenta = '#B24782' // no mapping color
 export const orchid = '#943670' // no mapping color
-export const red08 = '#F3D6DB' // no mapping color
+
+export const red = colors.red[900] // no mapping color $app-red
+export const red08 = '#FCF4F6' // no mapping color $app-red-08
+export const red10 = '#F3D6DB' // no mapping color $app-red-10
```
mnajdova: Oops the value of `red08` was moved to `red10`

packages/react/src/themes/teams/components/Alert/alertStyles.ts
```diff
@@ -0,0 +1,91 @@
...
+    color: v.color,
+    backgroundColor: v.backgroundColor,
+    borderColor: v.borderColor,
+  }
+}
+
+const alertStyles: ComponentSlotStylesInput<AlertProps, AlertVariables> = {
+  root: ({ props: p, variables: v, theme: { siteVariables } }): ICSSInJSStyle => {
+    return {
+      display: 'flex',
```
mnajdova: Is there a reason why we are not using the Flex component as a root of the Alert? We are adding some styles that are already there.

packages/react/src/themes/teams/components/Alert/alertStyles.ts
```diff
@@ -0,0 +1,91 @@
...
+  siteVars: SiteVariablesPrepared,
+): React.CSSProperties => {
+  const { colors } = siteVars
+  const naturalColors: NaturalColors = siteVars.naturalColors
+
+  if (p.danger) {
+    return {
+      color: siteVars.red,
+      backgroundColor: siteVars.red10,
+      borderColor: siteVars.red08,
```
mnajdova, line 23: Maybe we need to update the values here, as the `red08` was changed and it shouldn't be.

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,138 @@
...
+    const { closeButton, content, onClose } = this.props
+    return (
+      <>
+        {Box.create(content, {
+          defaultProps: {
+            className: Alert.slotClassNames.content,
+            styles: styles.content,
+          },
+        })}
+        {onClose &&
```
mnajdova: I don't like that we are using the `onClose` prop here as a decision of whether we want to render the `closeButton`. The user may provide just the `closeButton`, and add `onClick` event there, but still the button will not be shown. We may revisit this logic in a follow up PR.

mnajdova (approved):
Left few final comments. And please don't forget to update the changelog before merging.

packages/react/src/components/Alert/Alert.tsx
bmdalex: good point... I was thinking about it but wanted to avoid adding one more prop; the alternative would be to add a `closable` prop and do this:
```jsx
<Alert content="Close me!" closable onClose={() => /* code for closing */} />
```

but then we might end up with the following scenario if the user does not provide `onClose` or `onClick` on the `closeButton` slot:

```jsx
<Alert content="Close me!" closable />
```

**User clicks on the icon but it doesn't close because the component does not handle its visibility state**. At least in the current implementation it's a bit more obvious that the user needs write some code in the `onClose` handler for the close to happen.

## Proposed solution for v2:

Probably it'd be better to make this component autocontrolled; we can have there an internal `open` flag exported as an auto-controlled prop for the visibility of the `Alert`. In that case, the scenario above will be solved by doing:

- uncontrolled mode => the `Alert` will close because of internal implementation (similar to click outside in a popup):
```jsx
<Alert content="Close me!" closable />
```

- or controlled mode => the `Alert` will close because of the parent component code:
```jsx
class AlertExampleClosable extends React.Component<{}, { open: boolean }> {
  state = { open: true }

  render() {
    return (
      <Alert
        open={this.state.open}
        content="Close me"
        onClose={() => this.setState({ open: false })}
      />
    )
  }
}
```

@mnajdova @layershifter what do you guys think?

packages/react/src/themes/teams/components/Alert/alertStyles.ts
bmdalex: here it's correct, the problem is in `siteVariables` only, thanks again for catching it.

packages/react/src/themes/teams/siteVariables.ts
bmdalex: great catch, fixed

packages/react/src/components/Alert/Alert.tsx
mnajdova: Makes sense!

layershifter: My vote of ACC 👍

alinais (approved):

packages/react/src/themes/teams/components/Alert/alertStyles.ts
bmdalex: I was considering it and decided to avoid for 2 reasons:

### 1. Markup becomes complex

```typescript
renderComponent(config: RenderResultConfig<AlertProps>) {
  const { accessibility, classes, ElementType, unhandledProps, styles } = config
  const { children, content, onClose, closeButton } = this.props
  const hasChildren = childrenExist(children)

  return (
    <Flex
      as={ElementType}
      gap="gap.smaller"
      vAlign="center"
      className={classes.root}
      {...rtlTextContainer.getAttributes({ forElements: [children, content] })}
      {...unhandledProps}
    >
      {hasChildren && children}
      {!hasChildren && [
        <Flex.Item grow>
          {Box.create(content, {
            defaultProps: {
              key: Alert.slotClassNames.content,
              className: Alert.slotClassNames.content,
              styles: styles.content,
            },
          })}
        </Flex.Item>,
        onClose &&
          Button.create(closeButton, {
            defaultProps: {
              key: Alert.slotClassNames.closeButton,
              iconOnly: true,
              text: true,
              className: Alert.slotClassNames.closeButton,
              styles: styles.closeButton,
            },
            overrideProps: this.handleCloseButtonOverrides,
          }),
      ]}
    </Flex>
  )
}
```

### 2. We're tying the `Alert` implementation with the flex display which is more a theme concern

Other themes might not want to have a flex display. It is true they could override this with styles but it can become messy

What do you think, @mnajdova ?

bmdalex: agreed to not use `Flex`

packages/react/src/components/Alert/Alert.tsx
bmdalex: added to the RFC #1061 - I will create one for follow-up (Alert v2 or something)

packages/react/src/components/Alert/Alert.tsx
bmdalex: removed defaults

PR #1066, opened 2019-03-17
fix(Menu): Add aria posinset and setsize, hide indicator from narration

### LINKED ISSUE #982: Acc - Reader - "VoiceOver keys" - VoiceOver narrate wrongly position in menu
### Steps
Navigate to the menu component:
https://stardust-ui.github.io/react/components/menu
OSX/Chrome/VoiceOver
Navigate through the each menu on the page using VoiceOver keys. (Control+Option+right/left arrow key)

### Expected Result
position in the menu is narrated correctly as  
1 of [total number of menu items]
2 of [total number of menu items]
3  of [total number of menu items]
etc...

### Actual Result
VoiceOver narrate position incorrectly on some menus, like:
2 of 6
4 of 6
6 of 6
even menu has only 3 menu items.

### Version
x.y.z

### Testcase

jurokapsiar:
This fixes menu items narration especially for older versions of Chromium.
- explicitly adds aria-posinset and aria-setsize attributes to the menu item root
- changes default behavior of Indicator component to IconBehavior

Might also fix #982 (but verification is needed)


miroslavstastny (approved):
👍 for code changes
@kolaps33 could you please review/verify accessibility?

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -152,6 +152,13 @@ class Menu extends AutoControlledComponent<ReactProps<MenuProps>, MenuState> {
       indicator,
     } = this.props
     const { activeIndex } = this.state
+    let itemsCount = 0
+    _.forEach(items, item => {
+      if (getKindProp(item, 'item') !== 'divider') {
+        itemsCount++
+      }
+    })
+    let itemPosition = 0
```
layershifter, line 156:
```js
const itemsCount = _.filter(items, item => getKindProp(item, 'item') !== 'divider').length
```

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -57,6 +57,12 @@ export interface MenuItemProps
   /** MenuItem index inside Menu. */
   index?: number
 
+  /** MenuItem position inside Menu (skipping separators). */
+  itemPosition?: number
+
+  /** MenuItem count inside Menu. */
+  itemsCount?: number
```
layershifter, line 64: These props will be added to public API actually, I'm fine with this for now, but I'm not sure that components should be so coupled with accessibility concerns.

kolaps33:
Tested with following:
ElectronFiddle + Jaws a NVDA 
Fx + NVDA 
Ch + NVDA 
Edge + Narrator 

narrating position correctly :)

OSX/Chrome/VoiceOver still exists the issue:
https://github.com/stardust-ui/react/issues/982

kolaps33 (approved):

packages/react/src/components/Menu/MenuItem.tsx
jurokapsiar: yes, I am aware. but as far as I know, we do not have any other option to pass the information from parent menu to the menu item behaviors.

PR #1069, opened 2019-03-18
feat(theme): updating color palette and scheme

### LINKED ISSUE #707: Improving the color mechanism and organization
## Improvement for the color mechanism

This issue will keep track of the possible improvements for the color definitions and the mechanism used for creating the color scheme.
1. The siteVariables are polluted with lots of different concepts related to colors: scheme, contextualColors, emphasisColors etc. We can create one object in the siteVariables, where all these values will be assembled. For more details, please the thread in this https://github.com/stardust-ui/react/pull/647#discussion_r247159414
2.  We need better way for generating the right foreground colors based on the background. Currently we are storing a list of light color and depending on that, the foreground colors are generated. Please see the suggestion on this thread for some proposals on this: https://github.com/stardust-ui/react/pull/647#discussion_r247162443
3. Currently in the colorScheme we have different object default that contains the default colors. This makes the structure of the scheme strange:
```
const colorScheme = {
      foreground: '...',
      border: '..',
      shadow: '..',
      default: {
        foreground: colors.grey[600],
        border: colors.grey[600],
        shadow: colors.grey[600],
        background: colors.grey[100],
      },
    }
```
We should consider other mechanism for creating defaults and merging them with the colors used in the components then this.

### LINKED ISSUE #898: Improve the typings of the colors
### Problem description
Currently the colors object defined in the teams theme has the following type:
```
export type ColorPalette = ExtendablePalette<
  EmphasisColorsStrict & ContextualColorsStrict & NaturalColorsStrict & PrimitiveColors
>
```
where the `NaturalColorsStrict` is defined in the following manner:
```
type NaturalColorsStrict = Partial<{
  blue: ColorVariants
  green: ColorVariants
  grey: ColorVariants
  orange: ColorVariants
  pink: ColorVariants
  purple: ColorVariants
  teal: ColorVariants
  red: ColorVariants
  yellow: ColorVariants
}>
```

This produces problem when the user wants to specify additional colors in the color palette (like there are currently `lightGreen` and `darkOrange` in teams theme). The problem is that, we cannot use these colors from the colors object, because the typings are not containing these colors, so we always have to import them from `naturalColors`.

### Proposed solution
One possible solution that will allow the users to specify custom colors is to add generic for custom color types, so that the color palette can be defined in the following manner:

```
export type ColorPalette<T={}> = ExtendablePalette<
  EmphasisColorsStrict & ContextualColorsStrict & NaturalColorsStrict & PrimitiveColors & T
>

export type CustomThemeColors = Partial<{
  darkOrange: ColorVariants
  lightGreen: ColorVariants
}>
// then in the theme...
const colors: Color<CustomThemeColors> = {...}
```

mnajdova:
**NOTE: these changes are mostly scoped to Teams theme only. The only core Stardust functionality change, is that the `colors` object is not available anymore in the style functions, the variables should be the only place defining values for the styles.**

### Colors updates
First of all, let me start with the motivation for starting this work on the colors. We need to deliver the following functionalities in the definition of the styles:
1.	The colors should be safe for using by developers, in a mean that they should be abstracted away from theme switching. This means that if a developer uses some color in the component’s styles, it should work in all three themes out of the box .
2.	We should have structured way of how the colors available in the app can be accessed. This means, that the color name itself should have some meaning, instead of the values that we currently have (like `gray04`, or `primary08`) which doesn’t mean anything in the mind of the developer. We are trying to move more towards names like _brand foreground color_, or _brand foreground hover_ color, that will express better the intent of its usage.
3.	There should be one source of truth for all colors available in the application, so that designers can pick from a predefined set of colors when creating the design tokens, and colors should not be used if they don’t exist in  this set (this will move us to consistent and predictable designs)

With these in mind, we will introduce two things for the Light, Dark and High Contrast Teams themes. 
- **color palette**, which is the set of all colors available in the application, and will not change in any of the themes
- **color schemes**, which will define the design tokens for colors trough out all different themes, with the following naming structure: color => area => state. For example, the brand color can be expressed in the following format:
```
// the colors object used here is basically the color palette that we mention before
{
  foreground: colors.brand[600],
  background: colors.brand[50],
  border: colors.brand[600],
  shadow: colors.brand[600],

  foregroundHover: colors.white,
  backgroundHover: colors.brand[300],
  borderHover: colors.brand[600],
  shadowHover: colors.brand[600],

  foregroundActive: colors.white,
  backgroundActive: colors.brand[500],
  borderActive: colors.brand[600],
  shadowActive: colors.brand[50],

  foregroundFocus: colors.brand[50],
  backgroundFocus: colors.brand[600],
  borderFocus: colors.brand[600],
  shadowFocus: colors.brand[600],

  foregroundFocusWithin: colors.brand[50],
  backgroundFocusWithin: colors.brand[600],
  borderFocusWithin: colors.brand[600],
  shadowFocusWithin: colors.brand[600],

  foregroundDisabled: colors.brand[50],
  backgroundDisabled: colors.brand[600],
  borderDisabled: colors.brand[600],
  shadowDisabled: colors.brand[600],
}
```

#### The changes applied in the PR
We introduced one global color palette, created by design, that contains all colors that should be used in the Teams application. This will not be changed across different themes, it will always be the same object.

Apart from this, we created three color schemes, one in each theme, which are identical in their structure (so people can pick values from them safely), but the actual values of the picked token will have different values for colors. For example, this is how the brand part of the color scheme, looks in light, hoc and dark theme:
 
image
image

What this means is that, if a developer use in their styles use: `colorScheme.primary.backgroundHover`, it will correctly be mapped to #8DBDE6 in light/dark theme, and to #000 in hoc theme. This is Stardust’s proposal of how the design tokens can be organized, we still have to work with design to conclude on this. As an example of possible change in the future is that, we may introduce inverted schemas for each of the themes.

Fixes https://github.com/stardust-ui/react/issues/898 and fixes https://github.com/stardust-ui/react/issues/707

### Introducing colors in components.
Let's say that you want to introduce Foo component that is able to color itself with theme being switched. How should you start with this:

1. Design should specify in the specs which colors should be used for this component, for example:

```
// variables, default theme
return {
    atMentionMeColor: siteVariables.colorScheme.pink.foreground,
    atMentionOtherColor: siteVariables.colorScheme.default.foreground1,
    importantColor: siteVariables.colorScheme.red.foreground,
   ...
}
```

This colors will be mapped correctly in all different themes, so the user won't have to do the overrides. 

2. If the colors you need don't map in the color scheme, then take a look on the Design Spec and use colors from siteVariables, as a first thing? Like

```
// variables, default theme
return {
    atMentionMeColor: siteVariables.colors.pink[500], // those are palette colors
    atMentionOtherColor: siteVariables.colors.grey[600],
    importantColor: siteVariables.colors.red[500]
   ...
}
```
Then, to override colors for different themes, according to the Design Spec:

```
// variables, Dark Theme
return {
    atMentionMeColor: siteVariables.colors.red[500], // overriding with palette color
   ...
}
```

```
// variables, HC Theme
return {
    atMentionMeColor: siteVariables.accessibleYellow,
   ...
}
```
On the other hand, if you are developing component that should support the color prop, then you may use directly the color scheme, pick the color that is provided by the prop, and use directly the values from the color scheme (again those should be approved by design for the component being developed).

### BREAKING CHANGES:
I will introduce here the mappings for the colors of all the three Teams themes to the corresponding value in the global color palette. 

#### Light theme
```
export const gray02 = colors.grey[500]
export const gray03 = colors.grey[450]
export const gray04 = colors.grey[350]
export const gray06 = colors.grey[250]
export const gray08 = colors.grey[200]
export const gray09 = colors.grey[150]
export const gray10 = colors.grey[100]
export const gray14 = colors.grey[50]

export const brand04 = colors.brand[800]
export const brand08 = colors.brand[500]

export const magenta = '#B24782' // no mapping color (should we update this to pink[500] - '#B4009E')
export const orchid = colors.pink[600]

export const red = colors.red[400]
export const red08 = colors.red[100] // no mapping color $app-red-08 (updated to '#F3D6D8')
export const red10 = colors.red[50]
```
#### Dark theme
```
export const black = colors.grey[800]

export const gray02 = colors.grey[250]
export const gray03 = colors.grey[300]
export const gray04 = colors.grey[400]
export const gray06 = colors.grey[450]
export const gray08 = colors.grey[500]
export const gray09 = colors.grey[550]
export const gray10 = colors.grey[600] // no mapping color (updated to '#323131')
export const gray14 = colors.grey[700]
export const gray850 = colors.grey[850] // not safe for theme consistency (these colors are not mapped in all themes)
export const gray600 = colors.grey[600] // not safe for theme consistency (these colors are not mapped in all themes)

export const brand = colors.brand[600]
export const brand02 = colors.brand[200]
export const brand04 = colors.brand[300]
export const brand06 = colors.brand[400]
export const brand08 = colors.brand[500]
export const brand12 = colors.brand[600]
export const brand14 = colors.brand[800]
export const brand16 = colors.brand[900]

export const orange04 = colors.orange[300]
export const magenta = '#cf6098' // no mapping color (should we update this to pink[300] - '#DE569A')
export const orchid = '#ae3d84' // no mapping color
export const red = '#d74654' // no mapping color (should we update this to red[300] - '#E73550')
export const red08 = '#4f232b' // no mapping color (should we update this to red[900] - '#1E040A')
export const green04 = colors.green[200]
```
### HC theme
```
export const black = colors.black // not safe for theme consistency (we should use here some foreground/background color from the color scheme, and mapped that to black in the HC theme)
export const white = colors.white // not safe for theme consistency (we should use here some foreground/background color from the color scheme, and mapped that to white in the HC theme)
export const green04 = colors.green[200]
```

## Previous color palette mappings

```
primary: { // was renamed to brand
  50: '#F4F4FC', // NO CHANGE
  100: '#E2E2F6', => colors.brand[200]
  200: '#BDBDE6', => colors.brand[300]
  300: '#8F90C1', //  no mapping color - consult design if used
  400: '#6E70AE', //  no mapping color - consult design if used
  500: '#6264A7', => colors.brand[600]
  600: '#55578D', //  no mapping color - consult design if used
  700: '#4A4C78', //  no mapping color - consult design if used
  800: '#414265', //  no mapping color - consult design if used
  900: '#33344A', // NO CHANGE
},
green: {
  50: '#D3E4DB', //  no mapping color - consult design if used
  100: '#C4DCCF', //  no mapping color - consult design if used
  200: '#B3D1C1', //  no mapping color - consult design if used
  300: '#9DC4AF', //  no mapping color - consult design if used
  400: '#8CBAA1', //  no mapping color - consult design if used
  500: '#7BB093', //  no mapping color - consult design if used
  600: '#68A584', //  no mapping color - consult design if used
  700: '#579A75', //  no mapping color - consult design if used
  800: '#458F67', //  no mapping color - consult design if used
  900: '#237B4B', => colors.green[600]
},
grey: {
  50: '#FFFFFF', // NO CHANGE, recommended to use colors.white
  100: '#E6E6E6', //  no mapping color - consult design if used
  200: '#CDCCCC', //  no mapping color - consult design if used
  300: '#B8B8B8', //  no mapping color - consult design if used
  400: '#A2A2A2', //  no mapping color - consult design if used
  500: '#8C8C8C', //  no mapping color - consult design if used
  600: '#747373', //  no mapping color - consult design if used
  700: '#5F5E5E', //  no mapping color - consult design if used
  800: '#404040', //  no mapping color - consult design if used
  900: '#252424', // colors.grey[750] - '#252424' - slightly different color, approved by design
},
orange: {
  50: '#FEF9F7', //  no mapping color - consult design if used
  100: '#FDF0EB', //  no mapping color - consult design if used
  200: '#FAE3DA', //  no mapping color - consult design if used
  300: '#F8D3C5', //  no mapping color - consult design if used
  400: '#F5C6B3', //  no mapping color - consult design if used
  500: '#F4B8A1', //  no mapping color - consult design if used
  600: '#F2AD92', //  no mapping color - consult design if used
  700: '#F0A081', //  no mapping color - consult design if used
  800: '#ED8E6A', //  no mapping color - consult design if used
  900: '#E97548', => colors.orange[300]
},
pink: {
  50: '#E8BDD5', //  no mapping color - consult design if used
  100: '#E1ABC8', //  no mapping color - consult design if used
  200: '#DA9EBF', //  no mapping color - consult design if used
  300: '#D693B8', //  no mapping color - consult design if used
  400: '#D28BB2', //  no mapping color - consult design if used
  500: '#CA7BA6', //  no mapping color - consult design if used
  600: '#C775A3', //  no mapping color - consult design if used
  700: '#C06597', //  no mapping color - consult design if used
  800: '#BA598F', //  no mapping color - consult design if used
  900: '#B34A84', //  no mapping color - consult design if used
},
red: {
  50: '#F2D1D7', //  no mapping color - consult design if used
  100: '#ECBDC5', //  no mapping color - consult design if used
  200: '#E8AFB9', //  no mapping color - consult design if used
  300: '#E39EAA', //  no mapping color - consult design if used
  400: '#DE8D9B', //  no mapping color - consult design if used
  500: '#D97B8C', //  no mapping color - consult design if used
  600: '#D56B7E', //  no mapping color - consult design if used
  700: '#D05B70', //  no mapping color - consult design if used
  800: '#CC4B61', //  no mapping color - consult design if used
  900: '#C4314B', => colors.red[400]
},
yellow: {
  50: '#FEF5D0', //  no mapping color - consult design if used
  100: '#FDF1BE', //  no mapping color - consult design if used
  200: '#FDEEAE', //  no mapping color - consult design if used
  300: '#FBEA9D', //  no mapping color - consult design if used
  400: '#FCE78E', //  no mapping color - consult design if used
  500: '#FAE37C', //  no mapping color - consult design if used
  600: '#FAE06C', //  no mapping color - consult design if used
  700: '#F9DC58', //  no mapping color - consult design if used
  800: '#F9D844', //  no mapping color - consult design if used
  900: '#F8D22A', => colors.yellow[400]
},
darkOrange: {
  50: '#F9ECEA', => colors.orange[50]
  100: '#ECBCB3', //  no mapping color - consult design if used
  200: '#E29C8F', //  no mapping color - consult design if used
  300: '#D97B69', //  no mapping color - consult design if used
  400: '#CC4A31', => colors.orange[400]
  500: '#C5472F', //  no mapping color - consult design if used
  600: '#B7432D', //  no mapping color - consult design if used
  700: '#A73D29', //  no mapping color - consult design if used
  800: '#983927', //  no mapping color - consult design if used
  900: '#833122', => colors.orange[700]
},
lightGreen: {
  50: '#E7F2D9', //  no mapping color - consult design if used
  100: '#DFEECD', //  no mapping color - consult design if used
  200: '#D8EAC1', //  no mapping color - consult design if used
  300: '#CDE3B0', //  no mapping color - consult design if used
  400: '#C6DFA4', //  no mapping color - consult design if used
  500: '#BDDB96', //  no mapping color - consult design if used
  600: '#B4D689', //  no mapping color - consult design if used
  700: '#ACD17B', //  no mapping color - consult design if used
  800: '#A1CC6B', //  no mapping color - consult design if used
  900: '#92C353', => colors.green[200]
},
magenta: {
  50: '#E8D4E0', //  no mapping color - consult design if used
  100: '#DEC0D2', //  no mapping color - consult design if used
  200: '#D4ADC5', //  no mapping color - consult design if used
  300: '#CDA0BC', //  no mapping color - consult design if used
  400: '#C491B0', //  no mapping color - consult design if used
  500: '#BB7FA4', //  no mapping color - consult design if used
  600: '#B26E98', //  no mapping color - consult design if used
  700: '#AA5F8D', //  no mapping color - consult design if used
  800: '#A14F82', //  no mapping color - consult design if used
  900: '#953872', //  no mapping color - consult design if used
},
postOrange: {
  50: '#FDF6F7', //  no mapping color - consult design if used
  100: '#FBF2F3', //  no mapping color - consult design if used
  200: '#FAEEEF', //  no mapping color - consult design if used
  300: '#F9ECED', //  no mapping color - consult design if used
  400: '#F8E9EA', //  no mapping color - consult design if used
  500: '#F7E5E6', //  no mapping color - consult design if used
  600: '#F7E2E4', //  no mapping color - consult design if used
  700: '#F5DEE0', //  no mapping color - consult design if used
  800: '#F5DBDD', //  no mapping color - consult design if used
  900: '#F3D6D8', => colors.red[100]
},
```

#### Renamed variables
- textVariables.ts
```
  colors => colorScheme
```

- dividerVariables.ts
```
  colors => colorScheme
```

- menuVariables.ts
```
  verticalBorderColor: string - REMOVED, was not used anywhere

  focusedBorder => borderColorFocus
  focusedOutline => outlineColorFocus
  focusedBackgroundColor => backgroundColorFocus

  hoverBackgroundColor => backgroundColorHover

  activeColor => colorActive
  activeBackgroundColor => backgroundColorActive
  iconOnlyActiveColor => iconOnlyColorActive

  disabledColor => colorDisabled
  verticalItemBorder => verticalItemBorderWidth + verticalItemBorderColor

  // these primary related colors are deleted and are now picked from the color scheme
  primaryActiveColor: string
  primaryActiveBackgroundColor: string
  primaryActiveBorderColor: string
  primaryFocusedColor: string
  primaryFocusedBackgroundColor: string
  primaryHoverBorderColor: string
  primaryUnderlinedBorderColor: string
```

## Other breaking changes
- Every usage of `color='primary'` should be replace with `color='brand'`
- `colors` object is not available in the style function anymore, the `colorScheme` object should be used instead.


packages/react/src/themes/types.ts
```diff
@@ -17,19 +17,18 @@ import { Extendable, ObjectOf, ObjectOrFunc } from '../types'
...
-  300: string
-  400: string
-  500: string
-  600: string
-  700: string
-  800: string
-  900: string
-}
-
+export type ColorVariants = Extendable<{
```
layershifter: Can be just:
```suggestion
export type ColorVariants = Extendable<Partial<{
```

docs/src/examples/components/Label/Variations/LabelExampleColor.shorthand.tsx
```diff
@@ -1,10 +1,11 @@
 import * as React from 'react'
 import { Label, Provider } from '@stardust-ui/react'
+import * as _ from 'lodash'
 
 const LabelExampleColor = () => (
   <Provider.Consumer
-    render={({ siteVariables: { colorScheme } }) =>
-      Object.keys(colorScheme).map(color => (
+    render={({ siteVariables: { primitiveColors, emphasisColors, naturalColors } }) =>
+      _.keys({ ...primitiveColors, ...emphasisColors, ...naturalColors }).map(color => (
```
mnajdova, line 8: Extracting only the colors we had previously in the example

docs/src/examples/components/Menu/Variations/index.tsx
```diff
@@ -14,9 +14,9 @@ const Variations = () => (
       examplePath="components/Menu/Variations/MenuExampleVerticalPointing"
     />
     <ComponentExample
-      title="Primary"
-      description="A menu can have a primary color."
-      examplePath="components/Menu/Variations/MenuExamplePrimary"
+      title="Color"
```
mnajdova: We can show more colors examples now instead of just primary menu

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -94,6 +94,7 @@ class Dialog extends AutoControlledComponent<ReactProps<DialogProps>, DialogStat
     ...commonPropTypes.createCommon({
       children: false,
       content: true,
+      color: true,
```
mnajdova, line 100: Fixed of the prop types to match the props interface

packages/react/src/components/Label/Label.tsx
```diff
@@ -67,7 +63,7 @@ class Label extends UIComponent<ReactProps<LabelProps>, any> {
   static className = 'ui-label'
 
   static propTypes = {
-    ...commonPropTypes.createCommon({ color: 'complex' }),
+    ...commonPropTypes.createCommon({ color: true }),
```
mnajdova, line 66: The color prop now always has a type of string, so we are removing all other color types.

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -23,7 +24,7 @@ import MenuDivider from './MenuDivider'
 
 export type MenuShorthandKinds = 'divider' | 'item'
 
-export interface MenuProps extends UIComponentProps, ChildrenComponentProps {
+export interface MenuProps extends UIComponentProps, ChildrenComponentProps, ColorComponentProps {
```
mnajdova: Added color prop to all Menu* components, as a showcase of how the color prop should be implemented.

packages/react/src/lib/commonPropInterfaces.ts
```diff
@@ -29,32 +29,7 @@ export interface UIComponentProps<P = any, V = any>
...
-  | {
-      foreground?: TColorValue
-      background?: TColorValue
-      border?: TColorValue
-      shadow?: TColorValue
-    }
-  | TColorValue
-
-export interface ColorComponentProps<TColor = ColorValue> {
+export interface ColorComponentProps<TColor = string> {
```
mnajdova, line 32: The color prop is simple string - it is theme's responsibility of how it will be handle.

packages/react/src/lib/renderComponent.tsx
```diff
@@ -153,11 +152,6 @@ const renderComponent = <P extends {}>(config: RenderConfig<P>): React.ReactElem
 
         const {
           siteVariables = {
```
mnajdova: Removing all color related types from the site variables.

packages/react/src/themes/types.ts
```diff
@@ -128,11 +153,6 @@ export type State = ObjectOf<any>
 // ========================================================
 
 export interface SiteVariablesInput extends ObjectOf<any> {
```
mnajdova: No typings regarding the colors inside the site variables - those should be theme dependent.

packages/react/src/themes/types.ts
```diff
@@ -92,18 +91,44 @@ export type PrimitiveColors = Partial<{
 type ExtendablePalette<T> = T &
   { [K in keyof T]?: K extends keyof PrimitiveColors ? string : ColorVariants }
 
-export type ColorPalette = ExtendablePalette<
-  EmphasisColorsStrict & ContextualColorsStrict & NaturalColorsStrict & PrimitiveColors
+export type ColorPalette<T = {}> = ExtendablePalette<
```
mnajdova, line 98: Making the color palette extended - this typings are used only internally in the theme definition

docs/src/prototypes/alerts/AnimatedBannerAlert.tsx
```diff
@@ -20,17 +20,17 @@ const BannerAlert: React.FunctionComponent<BannerAlertProps> = props => {
 
         if (oof) {
           return {
-            color: siteVars.orchid,
-            backgroundColor: '#F9F5F8',
-            borderColor: '#DCADC7',
+            color: siteVars.colors.pink[600],
+            backgroundColor: '#F9F5F8', // where is this color coming from?
```
bmdalex: @mnajdova these colors are not yet introduced by design team, but they are correct, you can sync with them maybe they have names for them already

packages/react/src/themes/base/components/Text/textVariables.ts
```diff
@@ -1,8 +1,5 @@
-import { ColorValues } from '../../../types'
-import { mapColorsToScheme } from '../../../../lib'
-
 export interface TextVariables {
-  colors: ColorValues<string>
+  colorScheme: any
```
bmdalex: type?

packages/react/src/themes/teams-dark/components/Button/buttonVariables.ts
```diff
@@ -6,20 +6,34 @@ export default (siteVars: any): Partial<ButtonVariables> => {
     colorActive: siteVars.colors.white,
     colorHover: siteVars.colors.white,
     colorFocus: siteVars.colors.white,
+    colorDisabled: siteVars.colors.grey[450],
     backgroundColor: 'transparent',
-    borderColorFocus: siteVars.black,
+    backgroundColorHover: siteVars.colors.grey[700],
```
bmdalex, line 11: These keys seem a bit random to me. Are `450, 550, 850, etc` keys defined for all colors? How does the user know the color is defined or not without just trying and seeing what happens?

packages/react/src/themes/teams/components/Divider/dividerStyles.ts
```diff
@@ -5,22 +5,19 @@ import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
 import { DividerPropsWithDefaults } from '../../../../components/Divider/Divider'
 import { DividerVariables } from './dividerVariables'
 
-const beforeAndAfter = (
-  color: string,
-  size: number,
-  variables: DividerVariables,
-): ICSSInJSStyle => ({
+const beforeAndAfter = (size: number, variables: DividerVariables, colors): ICSSInJSStyle => ({
```
bmdalex: `colors` type?

packages/react/src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -16,11 +14,17 @@ export interface LabelVariables {
 export default (siteVars: SiteVariablesPrepared): LabelVariables => {
   const color = 'rgba(0, 0, 0, 0.6)'
 
+  const colorScheme = {
+    ...siteVars.colorScheme,
+    default: {
+      ...siteVars.colorScheme.default,
```
bmdalex: isn't this a bit dangerous to do explicitly? maybe extract in a `mergeColors` function? (only if you think it's going to be reused/needed again)

packages/react/src/themes/teams/components/Menu/menuItemStyles.ts
```diff
@@ -17,83 +17,89 @@ const underlinedItem = (color: string): ICSSInJSStyle => ({
...
-      ? {
-          color: v.primaryFocusedColor,
-          background: v.primaryFocusedBackgroundColor,
-        }
-      : {
-          color,
-          background: v.hoverBackgroundColor,
-        }),
-
+    color: primary || color ? colorScheme.foregroundFocus : v.colorActive,
```
bmdalex: I see this kind of logic throughout the whole file, but sometimes `color` props as priority, other times a variable does.

How do we decide that?

packages/react/src/themes/teams/components/Menu/menuItemStyles.ts
```diff
@@ -269,8 +280,11 @@ const menuItemStyles: ComponentSlotStylesInput<MenuItemPropsAndState, MenuVariab
       underlined,
       vertical,
       disabled,
+      color,
     } = p
 
+    const colorScheme = v.colorScheme[color ? color : primary ? 'primary' : 'grey']
```
bmdalex: this logic keeps repeating: `color || (primary ? 'primary' : COLOR_NAME)` can we extract in function?

packages/react/src/themes/teams/components/Menu/menuStyles.ts
```diff
@@ -11,19 +11,22 @@ import {
...
       margin: 0,
       padding: 0,
+      color: v.color || colorScheme.backgroundDefault,
+      backgroundColor: v.backgroundColor || 'inherit',
       listStyleType: 'none',
       ...(iconOnly && { alignItems: 'center' }),
       ...(vertical && {
         flexDirection: 'column',
-        backgroundColor: v.verticalBackgroundColor,
+        backgroundColor: v.verticalBackgroundColor || theme.siteVariables.colors.white,
```
bmdalex: why don't we take this color from `colorScheme`?

packages/react/src/themes/teams/components/Menu/menuVariables.ts
```diff
@@ -1,89 +1,89 @@
...
 
-    verticalBackgroundColor: siteVars.colors.white,
+    verticalBackgroundColor: undefined,
     verticalItemPadding: `${pxToRem(9)} ${pxToRem(16)} ${pxToRem(9)} ${pxToRem(16)}`,
     verticalBoxShadow: siteVars.shadowLevel3,
     verticalDividerMargin: `${pxToRem(8)} 0`,
-    verticalItemBorder: `solid ${pxToRem(2)} transparent`,
+    verticalItemBorderWidth: pxToRem(2),
+    verticalItemBorderColor: 'transparent',
+    verticalPointingBorderColor: siteVars.colorScheme.primary.borderActive, // TODO: why is the default having primary color?!
```
bmdalex: > TODO: why is the default having primary color?!

most likely implementation mistake. probably the author didn't really care about default menus as most of them are used with `primary` prop set

packages/react/src/themes/teams/siteVariables.ts
```diff
@@ -1,29 +1,17 @@
...
-export const brand08 = '#8B8CC7' // no mapping color
-
-export const magenta = '#B24782' // no mapping color
-export const orchid = '#943670' // no mapping color
-
-export const red = colors.red[900] // no mapping color $app-red
-export const red08 = '#F3D6DB' // no mapping color $app-red-08
-export const red10 = '#FCF4F6' // no mapping color $app-red-10
+export {
+  colors,
```
bmdalex: do we really need to export all?
How do I know if I need to use `colors` vs `colorScheme` vs other type? As a user I'm confused by having so many choices

packages/react/src/themes/types.ts
```diff
@@ -92,18 +94,44 @@ export type PrimitiveColors = Partial<{
...
+  EmphasisColorsStrict & ContextualColorsStrict & NaturalColorsStrict & PrimitiveColors & T
 >
 
 /**
  * A type for the generic color scheme of a component based on CSS property names
  */
 export type ColorScheme = {
```
bmdalex: nit:

define keys as type and use `Record`
```typescript
type ComponentAreaName = 'foregroundDefault' | 'backgroundDefault' | 'borderDefault' | ...
export type ColorScheme = Record<ComponentAreaName, string>
```

bmdalex (commented):
looks awesome, can't wait to have this merged

docs/src/prototypes/alerts/AnimatedBannerAlert.tsx
mnajdova: I wouldn't add them currently as those are used in the prototypes, but once we have them in the variables, we may ask for updates. @codepretty do you have some input on this one?

packages/react/src/themes/teams-dark/components/Button/buttonVariables.ts
mnajdova: The color palette guide will show the users all colors available per theme. No other idea at this moment (maybe we should have different names?) I am open for suggestions

packages/react/src/themes/base/components/Text/textVariables.ts
mnajdova: Thanks for noticing, will have to take a look on how this will reflect on the base theme

packages/react/src/themes/teams/siteVariables.ts
mnajdova: We will have a guide for the colors usages. Basically, it should be something like this:
1. Whenever you need to use some color for a component, you should introduce variable for it, following the convention [aria][state], and if there is are props involved: [prop1][prop2][aria][state], for example: backgroundActive, or primaryBackgroundActive.
2. If your component supports color prop, then you should use color scheme, as you need to have all these areas + states defined per color
3. If you have a prop that somehow defines a color - then introduce a variable for it if it is not dependent on the color (like verticalBackgroundActive), or if it should be changed by altering colors, you may define two color schemes, default and vertical color scheme for example.
4. Last but not least :), you should have variables so that the user will in the end be able to change any of the areas + state. 

This is ideally how it should work. We cannot change all these in one go, currently we changed the Menu as one of the more complex components, as a showcase that this will work, but we need to take care of the implementation that we had before, and not introduce regressions. This is why currently in some case the color prop wins, in other case variable (this is how it previously behaved). As this evolves we will have one way of defining this. 

For the end user is important that they will have **one source of truth for colors** (one global color palette), and they will **always have to have at least variables for overriding the colors differently in all themes**. No color should be used directly in the styles.

packages/react/src/themes/teams/components/Menu/menuVariables.ts
mnajdova: I would not necessarily say mistake, if in teams theme the vertical pointing menu should always look like this. I added the comment as a red flag, is it may not be the intended behavior. If the combination of these two props always results in the same border color, then this variable is enough, and the variable will always win...

packages/react/src/themes/teams/components/Menu/menuItemStyles.ts
mnajdova: See my comments: https://github.com/stardust-ui/react/pull/1069#discussion_r274023858 and https://github.com/stardust-ui/react/pull/1069#discussion_r274025046

packages/react/src/themes/teams/components/Menu/menuStyles.ts
mnajdova: Actually we should take this color from the variable (my bad for introducing color directly in the styles :\) The vertical menu's should always have solid background, provided by the theme. The color prop is changing the border's, the background of the active menu items etc. Good catch though..

packages/react/src/themes/teams/components/Label/labelVariables.ts
mnajdova: IMHO this is much explicit and clear instead of having a function that will do that, in the end this is just object spread of the default values.. However I see your point, and someone can easily wrongly define the defaults... @miroslavstastny what is your thought on this?

packages/react/src/themes/teams/components/Menu/menuItemStyles.ts
mnajdova: Yep, pretty ugly, because I am using a prop to decide whether I should pick some color (primary)... Will extract this in a separate function.

packages/react/src/themes/teams/components/Label/labelVariables.ts
mnajdova: Added colorUtils in the teams theme for handling this. Not sure if it should be exported somewhere from stardust though..

DESIGN_CONSIDERATIONS.md
```diff
@@ -0,0 +1,77 @@
+## Optimize APIs and patterns for developers over designers
+
+Why?
+- There are thousands of them.  The cost of features and bugs is far greater.
```
kuzhelov: could we, please, be a bit more specific here? What stays behind `them`? the amount of APIs? In that case, could we point to specific examples (briefly, no code is needed) - because, otherwise, it is not absolutely clear in terms of problem's criterias (i.e. 'thousands' - really thousands? what, exactly, makes the situation confusing to the client?)

DESIGN_CONSIDERATIONS.md
```diff
@@ -0,0 +1,77 @@
...
+
+How?
+- Prefer readable and meaningful typed intellisense theme information, opposed to cryptic or unorganized theme information that requires lookups or design assistance to understand.
```
kuzhelov: not sure that we should mention Intellisense here, as we are referring to specific tool with that; `typed` and `discoverable types` should be enough

DESIGN_CONSIDERATIONS.md
mnajdova: This changes were introduced by @levithomason Will revert them from this PR, and we can create another dedicated PR where he can respond to your questions/comments.

docs/src/components/ColorBox.tsx
```diff
@@ -48,7 +48,7 @@ export const colorBoxStyles: ComponentSlotStylesInput<ColorBoxProps, ColorBoxVar
     backgroundColor: p.value,
     border: '1px solid transparent',
     borderRadius: p.rounded && '.25rem',
-    color: Color(p.value).isDark() ? v.colorWhite : v.colorBlack,
+    color: p.value !== undefined && Color(p.value).isDark() ? v.colorWhite : v.colorBlack,
```
layershifter, line 58: image


May it will be better just to skip undefined values?

docs/src/views/ColorPalette.tsx
```diff
@@ -47,78 +96,139 @@ const ColorPalette = () => (
...
 
-          <Header as="h2">Text colors</Header>
+          <Header as="h2">Color scheme</Header>
+          <p>
+            Now that we know how the color palette is defined, let's see how we can use these
+            values, so that they will work correctly when different themes will be applied. As part
+            of each theme, we are defining <b>color scheme</b>, which will define the design tokens
+            usages of the different colors from the palette, that make sense for the developers. For
+            example, let's take a look of one color scheme defined for primary in light theme and
+            high contrast theme.
```
layershifter: I want to mention that Stardust contains Teams stuff, but it's not required:

```suggestion
            Teams light as contrast themes.
```

docs/src/views/ColorPalette.tsx
```diff
@@ -47,78 +96,139 @@ const ColorPalette = () => (
...
+          </p>
+          <Grid columns={3}>
+            <Header as="h3" content="Design token" />
+            <Header as="h3" content="Light theme" />
+            <Header as="h3" content="HC theme" />
+          </Grid>
+          <Grid columns={1} variables={{ gridGap: '2rem' }}>
```
layershifter:
```suggestion
```

docs/src/views/ColorPalette.tsx
```diff
@@ -47,78 +96,139 @@ const ColorPalette = () => (
...
+            <ColorSchemes themes={[themes.teams, themes.teamsHighContrast]} name={'primary'} />
+          </Grid>
```
layershifter:
```suggestion
```

`Grid` is cruft there because `ColorSchemes` has own grid.

docs/src/views/ColorPalette.tsx
```diff
@@ -47,78 +96,139 @@ const ColorPalette = () => (
...
-              ),
+            const theme = {
+              componentStyles: {
+                Box: {
+                  root: ({ theme: { siteVariables } }) => {
+                    const colorScheme = siteVariables.colorScheme
+                    return {
+                      color: colorScheme.primary.foregroundDefault,
+                      backgroundColor: colorScheme.primary.backgroundDefault,
+                      border: '1px solid ' + colorScheme.primary.borderDefault,
```
layershifter:
```suggestion
                      border: `1px solid ${colorScheme.primary.borderDefault}`,
```

nit

docs/src/views/ColorPalette.tsx
```diff
@@ -47,78 +96,139 @@ const ColorPalette = () => (
...
+                        backgroundColor: colorScheme.primary.backgroundHover,
+                        border: '1px solid ' + colorScheme.primary.borderHover,
+                      },
+                    }
+                  },
+                },
+              },
+            }
+
+            const colorSchemeExample = () => (
```
layershifter:
```suggestion
            const ColorSchemeExample = () => (
```

docs/src/views/ColorPalette.tsx
```diff
@@ -47,78 +96,139 @@ const ColorPalette = () => (
...
+                 <Text content={'Box in light theme'} />
+                 <Box content={'LIGHT THEME - HOVER ME'}/>
+                 <Text content={'Box in high contrast theme'} />
+                 <Provider theme={themes.teamsHighContrast}>
+                   <Box content={'HIGH CONTRAST THEME - HOVER ME'}/>
+                 </Provider>
+              </Provider>
+            )
+
+            export default btnExample
```
layershifter:
```suggestion
            export default ColorSchemeExample
```

docs/src/components/ColorSchemes.tsx
```diff
@@ -0,0 +1,47 @@
...
+          elements.push(<ColorBox name={key} key={`${i}schema`} size="small" value={undefined} />)
+          isFirstElement = false
+        }
+        const value = themes[i].siteVariables.colorScheme[name][key]
+        elements.push(<ColorBox key={`${i}${key}`} size="small" value={value} />)
+      }
+    })
+    return (
+      <div className={classes.root}>
+        <Grid columns={themes.length + 1} styles={{ backgroundColor: '#f2f2f2' }}>
```
layershifter: image

I suggest to use more advanced template, it will give more space for tokens:
```
const columns = `auto ${_.times(themes.length, () => '200px').join(' ')}`
```

docs/src/components/ColorSchemes.tsx
```diff
@@ -0,0 +1,47 @@
...
+      let isFirstElement: boolean = true
+      for (let i = 0; i < themes.length; i++) {
+        if (isFirstElement) {
+          elements.push(<ColorBox name={key} key={`${i}schema`} size="small" value={undefined} />)
+          isFirstElement = false
+        }
+        const value = themes[i].siteVariables.colorScheme[name][key]
+        elements.push(<ColorBox key={`${i}${key}`} size="small" value={value} />)
+      }
+    })
```
layershifter: nit: I tried to rewrite it in more functional way:

```jsx
const colorSchemes = _.map(
  themes,
  theme => theme.siteVariables.colorScheme[name]
);
const elements = _.flatMap(_.head(colorSchemes), (_, token) => [
  <ColorBox
    name={token}
    key={`${token}schema`}
    size="small"
    value={undefined}
  />,
  ..._.map(colorSchemes, (colorScheme, i) => (
    <ColorBox key={`${token}${i}`} size="small" value={colorScheme[token]} />
  ))
]);

```

layershifter: Actually, one more note:

image

We provide way to copy color values, but what about tokens?
What about option for copy: `theme.siteVariables.colorScheme[name]`?

docs/src/views/ColorPalette.tsx
```diff
@@ -47,78 +96,139 @@ const ColorPalette = () => (
...
+                  <Box content={'HIGH CONTRAST THEME - HOVER ME'} />
+                </Provider>
+              </Provider>
             )}
-          </Grid>
+          />
+
+          <p>
+            The color scheme is just a recommendation of how the color design tokens can be
+            organized. If neeeded, you can add multiple color schemes per theme, like inverted, or
```
layershifter:
```suggestion
            organized. If needed, you can add multiple color schemes per theme, like inverted, or
```

docs/src/views/ColorPalette.tsx
```diff
@@ -47,78 +96,139 @@ const ColorPalette = () => (
...
+            specific to the parts of the application that looks different. If your design team has
+            provided you with different names for the design tokens, you can use those, by
+            introducing values in the siteVariables, just be sure that they will be mapped to the
```
layershifter:
```suggestion
            introducing values in the <code>siteVariables</code>, just be sure that they will be mapped to the
```
_nit_

packages/react/src/themes/base/components/Text/textStyles.ts
```diff
@@ -22,7 +21,11 @@ export default {
     },
     variables: v,
   }: ComponentStyleFunctionParam<TextProps, TextVariables>): ICSSInJSStyle => {
+    const colors = v.colorScheme[color] || v.colorScheme['default']
```
layershifter:
```suggestion
    const colors = v.colorScheme[getColorSchemeKey(p.color)]
```
?

mnajdova: This is in base theme, the function `getColorSchemeKey` is based on colors in teams theme :\ I may introduce a similar method for the base theme as well.

packages/react/src/themes/teams/components/Header/headerDescriptionStyles.ts
```diff
@@ -6,12 +6,15 @@ import { HeaderDescriptionVariables } from './headerDescriptionVariables'
...
 
 const headerStyles: ComponentSlotStylesInput<HeaderDescriptionProps, HeaderDescriptionVariables> = {
-  root: ({ props: p, variables: v }): ICSSInJSStyle => ({
-    display: 'block',
-    color: _.get(v.colors, p.color, v.color),
-    fontSize: pxToRem(22),
-    fontWeight: 400,
-  }),
+  root: ({ props: p, variables: v }): ICSSInJSStyle => {
+    const colors = v.colorScheme[p.color]
```
layershifter, line 10:
```suggestion
    const colors = v.colorScheme[getColorSchemeKey(p.color)]
```

?

docs/src/components/ColorSchemes.tsx
mnajdova: The tokens are just human readable (the actual token may be something like `colorScheme.primary.borderDefault`, so not sure if this will bring any value to the user...

packages/react/src/themes/teams/components/Label/labelStyles.ts
```diff
@@ -4,17 +4,20 @@ import { pxToRem } from '../../../../lib'
 import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
 import { LabelProps } from '../../../../components/Label/Label'
 import { LabelVariables } from './labelVariables'
+import { getColorSchemeKey } from '../../colors'
 
 const labelStyles: ComponentSlotStylesInput<LabelProps, LabelVariables> = {
-  root: ({ props: p, variables: v, colors }): ICSSInJSStyle => {
+  root: ({ props: p, variables: v }): ICSSInJSStyle => {
+    const colors = v.colorScheme[getColorSchemeKey(p.color)]
```
layershifter:
```suggestion
    const colors = getColorScheme(v.colorScheme, p.color)
```

Just suggestion :) Why we can grab it directly?

packages/react/src/themes/teams/components/Header/headerDescriptionStyles.ts
mnajdova: We don't want that logic here, as this component implements fallback variable. We want if this is actually null to use the variable instead. I am open for other suggestions though...

layershifter: Let's address them in feature and may be we will decide to prefer schemes instead of component's fallback values 🚶

mnajdova: Agreed!

docs/src/components/ColorSchemes.tsx
kuzhelov: understand the reasons of @layershifter proposal - however, would also vote for leaving it as is for now. One of the reasons for that is that I wouldn't want to propose the specific API at this point in the copy-paste functionality, as it may change in future. I think that we are providing enough information to the client for being able to consume the color now, even if not in the most convenient form.

At the same time, agree that this might be improved in future

docs/src/components/ColorBox.tsx
mnajdova: For this I wasn't sure myself. I wanted to emphasize that we are missing some of the colors, but I agree that it is confusing for the community to see these values empty values. @kuzhelov @layershifter what do you think?

docs/src/views/ColorPalette.tsx
mnajdova: `let's take a look of one color scheme defined for primary in light theme and high contrast theme.` 
=>
`let's take a look of one color scheme defined for primary in Teams light theme and high contrast themes.`

Did you mean this?

packages/react/src/themes/teams/components/Label/labelStyles.ts
mnajdova: I've introduced this change, looks much cleaner!

packages/react/src/themes/base/components/Text/textStyles.ts
mnajdova: Will leave this for now as it is, so at least we can spot the difference between base/teams themes

packages/react/src/themes/teams/components/Label/labelStyles.ts
```diff
@@ -4,17 +4,20 @@ import { pxToRem } from '../../../../lib'
...
     return {
       display: 'inline-flex',
       alignItems: 'center',
       overflow: 'hidden',
       height: v.height,
       lineHeight: v.height,
-      color: colors.foreground,
-      backgroundColor: colors.background,
+      color: colors.backgroundDefault,
+      backgroundColor: colors.foregroundDefault,
```
hnager: Previously the color scheme constructed for Labels was conditional based on whether or not the color was considered dark or light. It looks as if that code was removed, so I'm curious to see how the Label examples look with the new color scheme mapping. Do you have a screenshot?

hnager: I see now that Teams grafts some lightColorOverrides for certain natural colors. It still would be good to see what the end result looks like (if you have a screenshot)

mnajdova: Hey, here is a screenshot of the Label's color example: 
image
The important thing is that, for each color we are defining the scheme, by defining the foreground + background default colors. As we are the one defining the scheme, these colors can be properly created, with taking into account the contrast betwone the two. We are still using in some cases the white as a foreground, and in others the black color, but still the scheme is not defined, and currently we aim to have correct values for the things which we are actually using (and no regressions are introduced on those example - default, black and white colors). Btw maybe the orange and green should have white foreground, but at this moment, we don't care much about that so it's open of changes that potentially will be introduced by design

hnager: I wonder if orange and yellow should be considered "light" any more. The 600 variant for each is actually pretty dark. But, like you said, we are in control of the color scheme mapping so we can adjust it however necessary.

packages/react/src/themes/teams/components/Menu/menuStyles.ts
```diff
@@ -38,14 +42,14 @@ export default {
         !iconOnly &&
         !(pointing && vertical) &&
         !underlined && {
-          border: `${v.borderWidth} solid ${v.borderColor}`,
-          ...(primary && {
-            border: `${v.borderWidth} solid ${v.primaryBorderColor}`,
-          }),
+          // primary has hardcoded grey border color
+          border: `${v.borderWidth} solid ${
+            primary ? v.primaryBorderColor : v.borderColor || colorScheme.border
```
miroslavstastny: This is an exercise on operator precedence

packages/react/src/themes/colorUtils.ts
```diff
@@ -0,0 +1,18 @@
+import { ColorSchemeMapping, ColorSchemeMappingOverrides } from './types'
+
+export const extendColorScheme = (
```
miroslavstastny, line 3: No unit test?

packages/react/src/themes/teams/components/Menu/menuStyles.ts
mnajdova: My intention was this: `primary ? v.primaryBorderColor : (v.borderColor || colorScheme.border)`  Is this more obvious? :)

packages/react/src/themes/colorUtils.ts
mnajdova: Will add. Thanks for reminding me :)

packages/react/src/themes/teams/components/Menu/menuVariables.ts
```diff
@@ -1,89 +1,103 @@
...
+        foregroundHover: siteVars.colors.white,
+        backgroundHover: siteVars.colors.brand[300],
+        foregroundActive: siteVars.colors.white,
+        borderActive: siteVars.colors.brand[600],
+        backgroundActive1: siteVars.colors.brand[500], // it's 600 in the color scheme
+        foregroundFocus: siteVars.colors.white,
+        backgroundFocus1: siteVars.colors.brand[300],
+      },
+    }),
+    color: siteVars.colors.grey[500],
```
miroslavstastny, line 64: Is there a plan for future refactoring of the color palette/variables?

I am confused with which variable is used for what and when:
- `colorScheme.foreground` is never used, we use `variables.color`
- for active primary item background goes from color scheme (`colorsScheme.backgroundActive1`) and can be overriden by variable (`variables.backgroundColorActive`), color goes from color scheme (`colorScheme.foregroundActive`) but cannot be overridden.
- `v.backgroundColorActive || colorScheme.backgroundActive1` and the variable defaults to `undefined`, but `v.underlinedBorderColor || colorScheme.backgroundActive1`  and the variable is set?

docs/src/views/ColorPalette.tsx
```diff
@@ -25,20 +59,39 @@ const ColorPalette = () => (
...
           </p>
           <p>
-            Each theme should match our color palette types fully. This will allow you to use our
-            theming features completely and keep your palette structured.
+            There are two things important in order for colors to work transparently when the theme
+            switching is in play:
           </p>
+          <ul>
+            <li>
+              <b>{link('Color scheme', '#color-scheme')}</b> - design tokens for all colors used in
```
miroslavstastny: I would swap **Color scheme** with **Color palette** so that the order corresponds with the explanation below.

docs/src/views/ColorPalette.tsx
```diff
@@ -47,78 +100,154 @@ const ColorPalette = () => (
...
-                <ColorBox name={color} rounded value={colors[color][500]} />
+                <ColorBox name={color} rounded value={colors[color][600]} />
               </div>
             ))}
           </Grid>
 
-          <Header as="h2">Emphasis colors</Header>
-          <p>This part of the palette includes primary and secondary colors.</p>
+          <Header as="h3">Emphasis colors</Header>
+          <p>This part of the palette may includes primary (brand) and secondary colors.</p>
```
miroslavstastny: include

packages/react/src/themes/teams/components/Menu/menuVariables.ts
mnajdova: Yep, there is a plan for that, but if we were about to do that now, it would took a long time. We will go component by component, and we will update the styles to use the colorScheme values. Usually, they will be used in the following way:
```
{
  root: ({variables}) => {
    const colors = getColorScheme(variables.colorScheme, color, primary);
    return {
      ...,
      color: v.color || colors.foregroundColor,
    }
  },
  ...
}
```

The variables would default to undefined, and we need to come up with a way of how to show it in the docs.

For colors that should never change (even if different color or primary prop is provided), we can have directly value in the variable.

packages/react/src/themes/teams/components/Label/labelStyles.ts
```diff
@@ -2,17 +2,20 @@ import { pxToRem } from '../../../../lib'
 import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
 import { LabelProps } from '../../../../components/Label/Label'
 import { LabelVariables } from './labelVariables'
+import { getColorScheme } from '../../colors'
 
 const labelStyles: ComponentSlotStylesInput<LabelProps, LabelVariables> = {
-  root: ({ props: p, variables: v, colors }): ICSSInJSStyle => {
+  root: ({ props: p, variables: v }): ICSSInJSStyle => {
+    const colors = getColorScheme(v.colorScheme, p.color)
```
miroslavstastny, line 9: nit: result of `getColorScheme()` is stored to `colors` variable in 3 files and to `colorScheme` in another 4. Let's choose one name and stick with that.

packages/react/src/themes/colorUtils.ts
mnajdova: Added.

docs/src/examples/components/Label/Variations/LabelExampleColor.shorthand.tsx
miroslavstastny: `DividerExample` loops through emphasis and natural colors only (no primitive) is the difference intentional?

miroslavstastny: `TextExample` as well

packages/react/src/themes/types.ts
```diff
@@ -213,6 +238,7 @@ export interface ComponentSlotStylesPrepared<TProps = {}, TVars = {}>
   extends ObjectOf<ComponentSlotStyleFunction<TProps, TVars>> {}
 
 export interface ComponentSlotClasses extends ObjectOf<string> {}
+
```
miroslavstastny, line 241: I don't understand this change but would definitely remove the export above or below :-)

packages/react/src/themes/teams/components/Label/labelStyles.ts
mnajdova: The `colors` as a name is used everywhere now. It looks more natural to use `colors.foreground`, `colors.backgroundDefault` etc.

packages/react/src/themes/types.ts
mnajdova: I think it was suppose to be just empty row, but I see now that we have the same line twice:
`export interface ComponentSlotClasses extends ObjectOf<string> {}`
Will remove one of them...

mnajdova: @miroslavstastny do you agree?

miroslavstastny: Yup, that's what I've meant.

miroslavstastny (approved):

packages/react/src/themes/base/components/Text/textVariables.ts
```diff
@@ -1,8 +1,7 @@
-import { ColorValues } from '../../../types'
-import { mapColorsToScheme } from '../../../../lib'
+import { ColorSchemeMapping } from '../../../types'
 
 export interface TextVariables {
-  colors: ColorValues<string>
+  colorScheme: ColorSchemeMapping
```
kuzhelov, line 4: this common type implies that supported `ColorNames` (in case if `color` props is supported by component) should be the same for any component of the app. In other words, if there is `pink` in color names, it should be supported by all components in corresponding schemes.

Won't it be a problem if, for instance, we would need to support extended range of colors for some particular component? Or this case is not considered intentional, as design spec rules explicitly won't allow this to happen? What would be our options to support this case, if necessary?

I see `extendColorScheme` method, so I might expect that we will introduce only minimal set of color options for base scheme, and will allow components extend this set of necessary, without affecting others - am I right on that?

kuzhelov: also, related question: currently, for any component that supports `color` prop, it is possible to introduce the following markup :

```jsx
<Text color={any_color_from_color_names} />
<Segment ... > // same thing
```

Wondering of we would like to introduce this much flexibility to the client - as, essentially, only subset of these color variants of `Text` (or `Segment`) components is supposed to be used in the app.

Of course, client might omit defining colors, and then default one will be applied on render - but then markup will be misleading.

Could you, please, suggest our plans to handle this situations?

packages/react/src/themes/base/components/Text/textVariables.ts
```diff
@@ -1,8 +1,7 @@
-import { ColorValues } from '../../../types'
```
kuzhelov, line 1: could you, please, walk me through client's workflow on this. So, I would like to introduce `Foo` component that is able to color itself with theme being switched. What is the source I need to use colors from? 

Am I right that I am supposed to take a look on the Design Spec and use colors from `siteVariables`, as a first thing? Like 

```js
// variables, default theme
return {
    atMentionMeColor: siteVariables.colors.pink[500], // those are palette colors
    atMentionOtherColor: siteVariables.colors.grey[600],
    importantColor: siteVariables.colors.red[500]
   ...
}
```

Then, to override colors for different themes, I need to override corresponding variables, according to the Design Spec:
```js
// variables, Dark Theme
return {
    atMentionMeColor: siteVariables.colors.red[500], // overriding with palette color
   ...
}
```

And then, if I would like to support color variants for my component (i.e. `color` prop), then this is the moment when I need to plug `colorScheme`, right - otherwise I don't need it?

```js
// variables, default theme
return {
    colorScheme: siteVariables.colorScheme,
    atMentionMeColor: siteVariables.colors.pink[500], // those are palette colors
    atMentionOtherColor: siteVariables.colors.grey[600],
    importantColor: siteVariables.colors.red[500]
   ...
}
```

packages/react/src/themes/base/components/Text/textVariables.ts
mnajdova: Very good question. As part of this PR, I just scoped this color value for the prop, to the colors we actually have, and I am defaulting this to the default styles if the color is not in this set. Ideally, we would have color values available defined per component, but this was still not specified by design. After we have these specs, we can update the component styles to handle only the colors they are suppose to. These changes were indeed just to unify the colors, and return default styles if something not expected is provided.

packages/react/src/themes/base/components/Text/textVariables.ts
mnajdova: Even if the component doesn't support color, the color scheme defines the colors that should be used across the app. Design should specify in the specs which colors should be used for this component, for example:

```
// variables, default theme
return {
    atMentionMeColor: siteVariables.colorScheme.pink.foreground, // those are palette colors
    atMentionOtherColor: siteVariables.colorScheme.default.foreground1,
    importantColor: siteVariables.colorScheme.red.foreground,
   ...
}
```
This colors will be mapped correctly in all different themes, so the user won't have to do the overrides. Currently we have only the default and primary color schemes, but design is working on defining the schemes for the other colors as well.

On the other hand, if the component supports the color prop, you are right, they can leverage further the color scheme, by picking the color provided by the color prop.

kuzhelov: got it - so, essentially, the vector of Design Spec rules reduces to the following:

- wherever it is possible, use colors from color sheme
- if rules of color scheme doesn't apply, define color from color palatte

Am I right here?

packages/react/src/themes/base/components/Text/textVariables.ts
kuzhelov: got it, thanks!

packages/react/src/themes/base/components/Text/textVariables.ts
mnajdova: > wherever it is possible, use colors from color sheme

Yep, totally right
> if rules of color scheme doesn't apply, define color from color palatte

Yep, and override them to the correct values in all different themes.

PR #1072, opened 2019-03-19
fix(accessibility): apply accessibility handlers with `applyAccessibilityKeyHandlers()`

### LINKED ISSUE #799: Popup: accessibility behavior silently overrides user's props
## Bug Report

It's not possible to handle events on elements passed to `trigger` prop.
See, https://github.com/stardust-ui/react/blob/v0.19.1/src/components/Popup/Popup.tsx#L377

### Steps
1. Open example, https://codesandbox.io/s/1v66l6m647
2. Focus "Working" button, click "Enter" key, you will get "RAW KEYDOWN"
3. Focus "Working" button, click "Enter" key

### Expected Result
On step 3 you should get "NEVER HAPPEN!"

### Actual Result
`onKeyDown` handler was not called.

### Version
0.19.1

layershifter:
Fixes #799.


layershifter:
## Option 2: `mergeAccessibilityHandlers()` / `applyAccessibilityHandlers()`

Add function that will only merge `keyHanders`.
UPD, we decided:
```js
{...applyAccessibilityKeyHandlers(
  accessibility.keyHanders.root,
  { onKeyDown: this.handleKeyDown, ...unhandedProps },
)}
```

packages/react/src/lib/getKeyDownHandlers.ts
```diff
@@ -57,8 +54,6 @@ const getKeyDownHandlers = (
 
           eventHandler && eventHandler(event)
         })
-
-        _.invoke(props, 'onKeyDown', event, props)
```
layershifter, line 61: This is wrong because it will be applied to each slot, see mini repro:
https://codesandbox.io/s/3kplmnz95q?module=/example.js
_When you will click `Enter` on dots, you will key `onKeyDown` twice_

packages/react/src/lib/applyAccessibilityKeyHandlers.ts
```diff
@@ -0,0 +1,18 @@
+import * as _ from 'lodash'
+
+import { Props } from '../types'
+import { KeyboardHandler, OnKeyDownHandler } from './accessibility/types'
+
+const applyAccessibilityKeyHandlers = (keyHandlers: OnKeyDownHandler, passedProps: Props) =>
```
mnajdova: Is the second argument really always going to be `Props`? If we want to use this functionality with the slots as well, should we receive here `ShorthandValue`? I know that the Popup has special handling of the `triggerProps`, but usually this is not the case...

layershifter: Done, it makes sense because we also deal with slots 👍

packages/react/src/lib/applyAccessibilityKeyHandlers.ts
```diff
@@ -0,0 +1,24 @@
+import * as _ from 'lodash'
+
+import { Props, ShorthandValue } from '../types'
+import { KeyboardHandler, OnKeyDownHandler } from './accessibility/types'
+
+const applyAccessibilityKeyHandlers = (
+  keyHandlers: OnKeyDownHandler,
+  value: Props | ShorthandValue,
+) => {
+  const slotProps: Props = typeof value === 'object' ? value : {}
```
mnajdova: Let's not ignore the other cases of shorthand value. Let's refactor this in the same way it is done in the factories, something like:
```  
  const valIsReactElement = React.isValidElement(value)

  const slotProps =
    (valIsReactElement && (value as React.ReactElement<Props>).props) ||
    (valIsPropsObject && (value as Props)) ||
    {}
```

mnajdova (approved):

PR #1068, opened 2019-03-18
chore(package): update to Jest 24

layershifter:
This PR updates Jest to latest. Picked from #923.
No user facing changes.

PR #1076, opened 2019-03-19
fix(jest): avoid test double run

layershifter:
No user-facing changes, fixes small regression introduced in #1068.

----

### Before

image

### After

image

### DIFF
--- a/build/jest/jest.config.common.js
+++ b/build/jest/jest.config.common.js
@@ -5,7 +5,6 @@ const setupFilesAfterEnv = [
 ]
 
 module.exports = {
-  rootDir: `${__dirname}/../../`,
   coverageDirectory: './coverage/',
   coverageReporters: ['json', 'lcov'],
   testRegex: '/test/.*-test\\.tsx?$',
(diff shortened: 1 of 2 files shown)


packages/react/jest.config.js
```diff
@@ -1,4 +1,14 @@
...
+  ...commonConfig,
   name: 'react',
+  moduleNameMapper: {
+    ...commonConfig.moduleNameMapper,
+    'docs/(.*)$': `<rootDir>/../../docs/$1`,
+
+    // Legacy aliases, they should not be used in new tests
+    'src/(.*)$': `<rootDir>/src/$1`,
+    'test/(.*)$': `<rootDir>/test/$1`,
+  },
```
layershifter, line 13: These mapping is specific for `@stardust-ui/react` and should not be used in new packages.

build/jest/jest.config.common.js
```diff
@@ -5,7 +5,6 @@ const setupFilesAfterEnv = [
 ]
 
 module.exports = {
-  rootDir: `${__dirname}/../../`,
```
layershifter, line 8: Now we have separate `rootDir`s for each package.

miroslavstastny (approved):

PR #1081, opened 2019-03-20
feat(Box): add examples and docs

layershifter:
This PR adds docs for `Box` component.


build/gulp/tasks/docs.ts
```diff
@@ -234,7 +234,9 @@ task('watch:docs', cb => {
       const sourceFilename = getRelativePathToSourceFile(filePath)
       const sourcePath = config.paths.docsSrc('exampleSources', sourceFilename)
 
-      fs.unlinkSync(sourcePath)
+      try {
+        fs.unlinkSync(sourcePath)
+      } catch (e) {}
```
layershifter, line 239: Fixes the crash of `yarn start` when it tries to remove non-existing JSON file.

docs/src/examples/components/Box/Types/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+        For layouts and positioning please prefer <Link to="/components/flex">Flex</Link> and{' '}
+        <Link to="/components/grid">Grid</Link> components, we also have{' '}
+        <Link to="/components/text">Text</Link> component to wrap text. In other cases please try to
+        use existing components and apply styles via theming features. You also can{' '}
+        <Link to="/integrate-custom-components">create your own custom component</Link> for custom
+        behaviors.
+      </p>
+      <p>
+        Remember that styles that applied directly to any component are not friendly for theme
+        switching.
```
mnajdova: Maybe we should add here that we prefer variables, that way those can be handled accordingly in all different themes. Also, not sure about using `styles` everywhere in the docs page.. Would rather use provider and variables, and show the correct way of styling the `Box` component.

layershifter: Updated to:

```
Remember that <code>styles</code> that applied directly to any component are not friendly for theme switching, prefer to use <code>variables</code> instead of <code>styles</code> for overrides.
```

---

Can't agree about `Provider`: for me it looks strange that I need to have `Provider` to apply styles on a single `Box` 🤔

layershifter: @mnajdova proposal:

```jsx
<Provider 
  theme={{
    componentStyles: {
      Box: ({variables: v}) => ({...(v.semiTransparent && {opacity: '0.5'})})
    }
  }}
>
  <Box content='Semi transparent box' variables={{semiTransparent: true}}/>
</Provider>
```

---

@Bugaa92 @miroslavstastny what do you think?

packages/react/src/components/Box/Box.tsx
```diff
@@ -17,7 +17,8 @@ export interface BoxProps
     ChildrenComponentProps {}
 
 /**
- * A Box is a basic component (no default styles)
+ * A Box is an abstract layout component, is frequently used for slots in other Stardust components.
```
bmdalex, line 20: I'd remove this comment completely and try to integrate it in the `Alert` we have in `docs/src/examples/components/Box/Types/index.tsx`

the page looks strange with the 2 descriptions:

image

docs/src/examples/components/Box/Types/index.tsx
Bugaa92: > Can't agree about Provider: for me it looks strange that I need to have Provider to apply styles on a single Box 🤔

I agree with @layershifter on this one. If we check how we use `Box` in our code base it's always something like:

```jsx
/* #1. Box for creating slots: */
{Box.create(SOME_SLOT, {
  defaultProps: {
    styles: SOME_STYLES,
  },
})}

/* #2. Box with `styles` prop: */
<Box styles={SOME_STYLES} />

/* 3. Box with `style` prop: */
<Box styles={SOME_STYLES} />
```

I don't see any instances where we rely on `Provider` for a single case usages of `Box`

> Would rather use provider and variables, and show the correct way of styling the Box component.

regarding @mnajdova 's proposal, I'd say an example with `Provider` will make much more sense if we provide more usages of the customized `Box`. It's also closer to a real scenario and that's what we want to show to our users:

```jsx
<Provider
  theme={{
    componentStyles: {
      Box: ({ variables: v }) => ({
        root: {
          ...(v.dashed && { border: '1px dashed #ccc' }),
          ...(v.semiTransparent && { opacity: '0.5' }),
          ...(v.color && { color: v.color }),
        },
      }),
    },
  }}
>
  <Box variables={{ dashed: true }}> {/* container */}
    {/* semi transparent paragraph */}
    <Box content="Here are your pills: " variables={{ semiTransparent: true }} />
    <Box content="Red pill" variables={{ color: 'red' }} /> {/* red paragraph */}
    <Box content="Blue pill" variables={{ color: 'blue' }} /> {/* blue paragraph */}
  </Box>
</Provider>

Comment (bmdalex) [reply docs/src/examples/components/Box/Types/index.tsx]: I DELETED MY OWN COMMENT BY ACCIDENT;

WILL POST A NEW ONE STAY TUNED

File: packages/react/src/components/Box/Box.tsx (hunk 3 above)
Comment (layershifter) [reply packages/react/src/components/Box/Box.tsx:20]: image

I moved this warning upper. I don't to remove existing description for component because we have an assertion for it in UTs:

image

File: docs/src/examples/components/Box/Types/index.tsx (hunk 2 above)
Comment (bmdalex) [reply docs/src/examples/components/Box/Types/index.tsx]: ## One-off Box with safe `styles`

> for me it looks strange that I need to have Provider to apply styles on a single Box 🤔

agreed with @layershifter on this one. If we take a look at how we use `Box` internally we have:

```jsx
/* #1. Box to create slots: */
Box.create(SOME_SLOT, {
  defaultProps: {
    styles: SOME_STYLES,
  },
})

/* #2. one-off Box with `styles` prop: */
<Box styles={SOME_STYLES} ../>

```

So if we agree these are good patterns we should advertise them in examples with one mention:

#### Theme switching won't work so CSS props changing colors should not be used.

## `Box` with `Provider` for more complex scenarios

> Would rather use provider and variables, and show the correct way of styling the Box component.

I think that is the preferred way to use `Box` for more complex apps where there are multiple scenarios you want to accomplish. Therefore, I think the examples provided by @layershifter 

```jsx
<Provider 
  theme={{
    componentStyles: {
      Box: ({variables: v}) => ({...(v.semiTransparent && {opacity: '0.5'})})
    }
  }}
>
  <Box content='Semi transparent box' variables={{semiTransparent: true}}/>
</Provider>
```

should be more complex and realistic so that the users understand it's the preferred way to use `Box` in apps where it needs to accomplish multiple scenarios; something on the lines of:

```jsx
import { Box, Provider, ComponentStyleFunctionParam } from '@stardust-ui/react'

const BoxShorthandExample = () => (
  <Provider
    theme={{
      componentStyles: {
        Box: {
          root: ({
            variables: v,
            theme: { siteVariables },
          }: ComponentStyleFunctionParam<any, any>) => ({
            ...(v.padded && { padding: '10px' }),
            ...(v.dashed && { border: `1px dashed ${siteVariables.gray02}` }),
            ...(v.semiTransparent && { opacity: '0.5' }),
            ...(v.themed && {
              color: siteVariables.bodyBackground,
              backgroundColor: siteVariables.brand04,
            }),
            ...(v.danger && { backgroundColor: siteVariables.red08 }),
          }),
        },
      },
    }}
  >
    {/* semi transparent Box as container */}
    <Box variables={{ dashed: true, padded: true }}>
      {/* Semi transparent text announcing different boxes */}
      <Box content="Here are your boxes: " variables={{ semiTransparent: true }} />

      {/* Themed box: */}
      <Box content="Themed box" variables={{ themed: true }} />

      {/* Danger box: */}
      <Box content="Danger box" variables={{ danger: true }} />
    </Box>
  </Provider>
)

export default BoxShorthandExample
```

that will also ensure theme switching works and will look like this:

### regular theme

Screenshot 2019-03-21 at 14 16 27

### dark 

Screenshot 2019-03-21 at 14 16 38

## tl;dr

My proposal is to create more examples of `Box` usages based on the user's needs:
### 1. example for one-off thing => use `Box` with `styles` but don't use CSS props that change colors as it will break theme switching
### 2. example more complex app with multiple `Box` scenarios => use `Box` with `Provider` and define `variables`

bmdalex: @layershifter @mnajdova what do you think?

layershifter: > I think that is the preferred way to use Box for more complex apps where there are multiple scenarios you want to accomplish. 

I think that example should contain less lines as possible and be simple. Your case with styling looks more like custom component because it uses design words like `padded` and `danger`, i.e `<Area danger />`

If I will move `theme` outside render we will have only custom theme and variables on `Box` which styles can possibly conflict with existing:

```jsx
<Provider>
  <Box />
</Provider>
// It's fine.
```
```jsx
const = {{ Box: { root: { padding: 50 } }}
//
<Provider>
  <Alert />
  <Box />
</Provider>
// Then I will just add `Alert` and new padding, styles of its slots can be suddenly change
```

packages/react/src/components/Box/Box.tsx
bmdalex: oh, ok, then this is an improvement anyway 👍

docs/src/examples/components/Box/Types/index.tsx
bmdalex: > I think that example should contain less lines as possible and be simple

Maybe my example was not the best but what I was trying to point out is that in this case if we make the example to a minimum we might hurt ourselves and send users the same wrong idea we sent with `Provider` usages before:

### that it's ok to use a `Provider` to customize just one component only

```jsx
<Provider ../>
  <Box ../> {/* just one Box in an isolated Provider */}
</Provider>
```

I'm afraid we might end up seeing a lot of isolated usages like this when the real usage is something like:

```jsx
<Provider ../> {/* top most provider at the root of the app */}

// ...

<Box ../> {/* Box usage somewhere down in the React Tree */}
```

If we reaaaaally want to keep examples to a minimum, let's provide another section where we don't have this constraint (similar to `/integrate-custom-components`) where we go into more details about `Box` usages.

What do you think @mnajdova @layershifter  ?

bmdalex (approved):

docs/src/examples/components/Box/Types/index.tsx
miroslavstastny: What's the main purpose of the `Box` for a consumer? It's to create an element (`div` by default) and be able to style it with styles going through fela. For that, I think it is correct to use `styles` directly on the component.

Then there are more advanced scenarios which include theme switching and for these you want to switch from `styles` to `variables` + `Provider`. But this is rather a general concept, not anything `Box`-specific. So why that should go to `Box` examples?

Moreover as @Bugaa92 wrote - I see examples with `Provider` + single component as quite misleading, because people then do this to modify single instance of a component which is an anti-pattern.

The problem is that developers rarely read documentation, they just copy&paste from examples to production code without realising that these two are different things :-/

PR #1082, opened 2019-03-20
docs(ComponentPlayground): add playground to docs

layershifter:
# Playground

playground-stage2

_Inspired by [BluePrint JS docs](https://blueprintjs.com/docs/#core/components/button)._

## New knobs

There cool knobs are powered by Context and hooks, it means that we will be to replace old knobs with them and get:
- working examples with knobs in full screen mode
- working examples with knobs on CodeSandbox
- will be able to create controlled components with knobs

Heavily inspired by [Retoggle](https://github.com/Raathigesh/retoggle).

They are implemented in `@stardust-ui/docs-components` and will start our work for shared docs experience 🚀 I'm going to move editor and snippet components there, it will allow us to reduce dependencies on root `package.json`, too.

_They use less dependencies as possible, to they are independent from Fela or Stardust._ 

## How to implement new playground?

Actually, each playground is a simple functional component with knobs, so just create a new `Playground.tsx` file in examples directory and it will work like a 💎 

`docs/src/examples/components/Header/Playground.tsx`
```tsx
import { useSelectKnob, useStringKnob } from '@stardust-ui/docs-components'
import { Header } from '@stardust-ui/react'

const HeaderPlayground: React.FunctionComponent = () => {
  const [as] = useSelectKnob({
    name: 'as',
    initialValue: 'h1',
    values: ['h1', 'h2', 'h3', 'h4', 'h5'],
  })
  const [content] = useStringKnob({
    name: 'content',
    initialValue: 'A sample header',
  })

  return <Header as={as} content={content} />
}

export default HeaderPlayground
```

This solution is temporary, in the future we are going to generate playgrounds in automated mode. But, we should start with something 🐎 

## Next steps

- Replace existing knobs with new
- Make `CodeSnippet`s copy-pastable, generated JSX should be also exportable to CodeSandbox


packages/docs-components/src/knobs/defaultComponents.tsx
```diff
@@ -0,0 +1,71 @@
...
+const KnobSelect: React.FunctionComponent<KnobComponentProps> = props => (
+  <select
+    onChange={(e: React.ChangeEvent<HTMLSelectElement>) => {
+      props.setValue(e.target.value)
+    }}
+    value={props.value}
+  >
+    {props.values &&
+      props.values.map(option => (
+        <option key={option} value={option}>
```
jurokapsiar, line 45: Have you considered using `<Dropdown>`?

layershifter: These components are part of `docs-component` package which should be consumable even without Stardust.

I am going to add mapping with our components in a separate PR.

PR #1083, opened 2019-03-20
feat(prototypes): banner alerts with different intents

bmdalex:
### Description:

This PR contains:
- in `docs/src/prototypes/alerts/AnimatedBannerAlert.tsx`:  `BannerAlert`, a customized prototype using `Alert` component that can express different intents through boolean props
- in `docs/src/prototypes/alerts/BannerAlerts.tsx`: a view showing an `AnimatedBannerAlert` custom component on top of a compose box

### Demo:
banner-alerts

### DIFF
--- a/docs/src/prototypes/alerts/AnimatedBannerAlert.tsx
+++ b/docs/src/prototypes/alerts/AnimatedBannerAlert.tsx
@@ -1,0 +1,102 @@
+import * as React from 'react'
+import { Alert, Animation, SiteVariablesInput, AlertProps } from '@stardust-ui/react'
+
+interface BannerAlertProps extends AlertProps {
+  oof?: boolean
+  critical?: boolean
+  urgent?: boolean
+}
+
+/**
+ * Customized Alert that can express different intents through boolean props:
+ * oof, urgent, critical
+ */
+const BannerAlert: React.FunctionComponent<BannerAlertProps> = props => {
+  const { oof, critical, urgent, ...rest } = props
+  return (
+    <Alert
+      variables={(siteVars: SiteVariablesInput) => {
+        const appWhite = siteVars.colors.grey[50]
+
+        if (oof) {
+          return {
+            color: siteVars.orchid,
+            backgroundColor: '#F9F5F8',
+            borderColor: '#DCADC7',
+          }
+        }
+
+        if (urgent) {
+          return {
+            color: appWhite,
+            backgroundColor: siteVars.red,
+            borderColor: siteVars.red,
+          }
+        }
+
+        if (critical) {
+          return {
+            color: appWhite,
+            backgroundColor: siteVars.gray03,
+            borderColor: siteVars.gray02,
+          }
+        }
+
+        return {}
+      }}
+      {...rest}
+    />
+  )
+}
+
+interface AnimatedBannerAlertProps extends BannerAlertProps {
+  closable?: boolean
+}
+
+interface AnimatedBannerAlertState {
+  open: boolean
+}
+
+/**
+ * Needs to have 'slideDown' animation defined in parent Provider
+ */
+class AnimatedBannerAlert extends React.Component<
+  AnimatedBannerAlertProps,
+  AnimatedBannerAlertState
+> {
+  private readonly initialState: AnimatedBannerAlertState = { open: true }
+
+  state = this.initialState
+
+  componentDidUpdate(prevProps: AnimatedBannerAlertProps) {
+    const p = prevProps
+    const { info, danger, oof, critical, urgent } = this.props
+
+    if (
+      p.info !== info ||
+      p.danger !== danger ||
+      p.oof !== oof ||
+      p.critical !== critical ||
+      p.urgent !== urgent
+    ) {
+      this.setState(this.initialState)
+    }
+  }
+
+  handleClick = () => this.setState({ open: false })
+
+  render() {
+    const { open } = this.state
+    const { closable, ...rest } = this.props
+
+    return closable ? (
+      <Animation name={open ? '' : 'slideDown'}>
+        <BannerAlert {...rest} action={{ icon: 'close', onClick: this.handleClick }} />
+      </Animation>
+    ) : (
+      <BannerAlert {...rest} />
+    )
+  }
+}
+
+export default AnimatedBannerAlert
--- a/docs/src/prototypes/alerts/AnimatedBannerAlert.tsx
+++ b/docs/src/prototypes/alerts/AnimatedBannerAlert.tsx
@@ -56,18 +56,16 @@ interface AnimatedBannerAlertProps extends BannerAlertProps {
 /**
  * Needs to have 'slideDown' animation defined in parent Provider
  */
-class AnimatedBannerAlert extends React.Component<AnimatedBannerAlertProps, { open: boolean }> {
-  render() {
-    const { open, ...rest } = this.props
+const AnimatedBannerAlert: React.FunctionComponent<AnimatedBannerAlertProps> = props => {
+  const { open, ...rest } = props
 
-    if (open === undefined) return <BannerAlert {...rest} />
+  if (open === undefined) return <BannerAlert {...rest} />
 
-    return (
-      <Animation name={open ? '' : 'slideDown'}>
-        <BannerAlert {...rest} />
-      </Animation>
-    )
-  }
+  return (
+    <Animation name={open ? '' : 'slideDown'}>
+      <BannerAlert {...rest} />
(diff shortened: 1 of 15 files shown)


layershifter:
There is an autoclose bug:

glitch

**_UPD: nevermind, my branch was outdated..._**

docs/src/prototypes/alerts/AnimatedBannerAlert.tsx
```diff
@@ -0,0 +1,102 @@
...
+            borderColor: siteVars.gray02,
+          }
+        }
+
+        return {}
+      }}
+      {...rest}
+    />
+  )
+}
```
layershifter, line 50: It looks confusing that we have a dedicated component to handle only variables.

My proposal let's replace it with a separate function:
```js
const getBannerAlertVariables = (props: BannerAlertProps) => (siteVars: SiteVariablesInput) => {
  const appWhite = siteVars.colors.grey[50]

  if (props.oof) {}
  // ...

  return {}
}
```

And then we can use it:
```js
const { closable, oof, info, urgent, ...rest } = this.props
const { open } = this.state
const variables = getBannerAlertVariables({ oof, info, urgent })

return closable ? (
  <Animation name={open ? '' : 'slideDown'}>
    <Alert {...rest} action={{ icon: 'close', onClick: this.handleClick }} variables={variables} />
   </Animation>
) : (
  <Alert {...rest} variables={variables} />
)
```

docs/src/prototypes/alerts/AnimatedBannerAlert.tsx
```diff
@@ -0,0 +1,102 @@
...
+    if (
+      p.info !== info ||
+      p.danger !== danger ||
+      p.oof !== oof ||
+      p.critical !== critical ||
+      p.urgent !== urgent
+    ) {
+      this.setState(this.initialState)
+    }
+  }
```
layershifter: Instead of manual reset, we can use `key: bannerName` in `BannerAlerts.tsx`. This will remount component and will reset state, so we will be able to remove `componentDidUpdate()` completely.

docs/src/prototypes/alerts/BannerAlerts.tsx
```diff
@@ -0,0 +1,88 @@
...
+  selectedBanner: RadioGroupItemProps
+}
+
+class BannerAlerts extends React.Component<{}, BannerAlertsState> {
+  private readonly initialState: BannerAlertsState = { selectedBanner: bannerRadioItems[0] }
+
+  state = this.initialState
+
+  render() {
+    const { selectedBanner } = this.state
```
layershifter: Can we store `bannerName` in state instead of props? Currently, it shows a bad practice because the state will contain event handlers and children:
image

docs/src/prototypes/alerts/AnimatedBannerAlert.tsx
bmdalex: > It looks confusing that we have a dedicated component to handle only variables.

The point of that is to show users how they can compose the `Alert` to create a `BannerAlert` that just introduces new intent options (`oof`, `critical`, `urgent`). Decoupling it like that means they can just copy paste it t without needing the `Animation` part.

What do you think?

Any way to make it more obvious that this is my intention?

layershifter: I don't see there a big difference because there are no complicated logic. Opened to other opinions, @mnajdova @kuzhelov

docs/src/prototypes/alerts/AnimatedBannerAlert.tsx
bmdalex: done, great suggestion 💯

docs/src/prototypes/alerts/BannerAlerts.tsx
bmdalex: ooops, fixed, thanks!

docs/src/prototypes/alerts/AnimatedBannerAlert.tsx
```diff
@@ -0,0 +1,73 @@
...
+}
+
+interface AnimatedBannerAlertProps extends BannerAlertProps {
+  open?: boolean
+}
+
+/**
+ * Needs to have 'slideDown' animation defined in parent Provider
+ */
+class AnimatedBannerAlert extends React.Component<AnimatedBannerAlertProps, { open: boolean }> {
```
layershifter: We don't have state more:

```suggestion
class AnimatedBannerAlert extends React.Component<AnimatedBannerAlertProps> {
```

And now it can be `FunctionalComponent`,

docs/src/prototypes/alerts/BannerAlerts.tsx
```diff
@@ -0,0 +1,101 @@
...
+    to: { transform: 'translateY(100%)', display: 'none' },
+  },
+  duration: '.3s',
+  iterationCount: '1',
+  timingFunction: 'linear',
+  fillMode: 'forwards',
+}
+
+interface BannerAlertsState {
+  selectedBannerName: string
```
layershifter:
```suggestion
  selectedBannerName: BannerName 
```

This allow to get rid of `as BannerName` in some places

packages/react/src/themes/teams/components/Alert/alertStyles.ts
```diff
@@ -72,9 +72,13 @@ const alertStyles: ComponentSlotStylesInput<AlertProps, AlertVariables> = {
 
     ...((p.attached === 'top' || p.attached === true) && {
       borderRadius: `${v.borderRadius} ${v.borderRadius} 0 0`,
+      marginBottom: '-1px',
```
mnajdova: Would introduce the `borderWidth` with variable and use here `-${v.borderWidth}`

docs/src/prototypes/alerts/BannerAlerts.tsx
```diff
@@ -0,0 +1,101 @@
...
+              attached: true,
+              [selectedBannerName]: true,
+              content: getBannerContent(selectedBannerName as BannerName),
+              ...(isAlertClosable(selectedBannerName as BannerName) && {
+                open,
+                action: { icon: 'close', onClick: this.closeSelectedBanner },
+              }),
+            }}
+          />
+          <ComposeMessage attached="bottom" />
```
mnajdova: It looks strange that we need to add some additional prop to the component that we want the Alert to be attached to.

docs/src/prototypes/alerts/AnimatedBannerAlert.tsx
mnajdova: Actually no strong opinion on this one, as it's pretty easy to understand what is happening in both scenarios.

packages/react/src/themes/teams/components/Alert/alertStyles.ts
bmdalex: @mnajdova good that you brought this up, that's exactly what I wanted to do from the beginning, but when I set `borderWidth` to `pxToRem(1)` the margin was not applied because of the inexact calculation. I can still do this but the value will be `1px` for `borderWidth` variable.

Are we ok with that?

docs/src/prototypes/alerts/BannerAlerts.tsx
bmdalex: I actually think it makes sense because the `ComposeMessage` has rounded borders at the top by default so we're attaching the components to each other.

It's done the same way in: https://react.semantic-ui.com/collections/message/#variations-attached

image

thoughts?

docs/src/prototypes/alerts/AnimatedBannerAlert.tsx
bmdalex: haha you were faster to review that I was to change the code 😆  done already

docs/src/prototypes/alerts/BannerAlerts.tsx
bmdalex: oops missed this, thanks! applied

packages/react/src/themes/teams/components/Alert/alertStyles.ts
bmdalex: fixed with better solution as discussed with @kuzhelov

layershifter (approved):

docs/src/prototypes/alerts/BannerAlerts.tsx
bmdalex: found solution with @mnajdova and @kuzhelov thanks guys, resolving

PR #1086, opened 2019-03-21
fix(Flex.Item): shrink issue

kuzhelov:
This PR introduces fix for `shrink` prop of `Flex.Item` component. Previously value of `0` was improper interpreted, so that resulting flex item was remaining shrinkable even with this value provided

```jsx
// no effect - still shrinkable :(
<Flex.Item shrink={0}>
  ...
</Flex.Item>
```


mnajdova (approved):
Looks good!

miroslavstastny (approved):

packages/react/src/themes/teams/components/Flex/flexItemStyles.ts
```diff
@@ -11,7 +11,7 @@ const flexItemStyles: ComponentSlotStylesInput<FlexItemProps, FlexItemVariables>
 
       ...(p.size && toFlexItemSizeValues(v[p.size])),
 
-      ...(p.shrink && { flexShrink: p.shrink }),
+      ...(typeof p.shrink === 'number' && { flexShrink: p.shrink }),
```
bmdalex, line 14: I know this is already merged, sorry for not being able to review before that. However, there are a few things I'd like to mention.

### 1. what about the global values as string?
I think we should allow users to pass these too (see: https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink#Syntax)
image

### 2. what about the true boolean value?
From the new code it looks like
```jsx
<Flex.Item shrink>
  {/* some element */}
</Flex.Item />
```
will have no effect; to me that's not expected user behavior since the type of `shrink` prop is still `boolean | number`

kuzhelov: Let me start with the second question first - yes, this is exactly the effect you've described. The logic is correct as`<Flex.Item shrink />` is, essentially, equivalent to <Flex.Item />, given that flex items are shrinkable by default (we've done it to correspond to CSS specs).

Speaking of the first question - we might want to introduce support for mentioned props later, but for now extra prop values from the scope of `Flex` component were omitted in order to achieve the following: 

> address as many scenarios with the least set of props/values necessary

Thus, might suggest to leave it as is for now and return to the question of introducing these prop values (or use another solution approaches) when there will be an appealing use case for that.

PR #1087, opened 2019-03-21
fix(types): update accessibility types to match React's

layershifter:
This PR fixes accessibility types.


packages/react/src/lib/accessibility/types.ts
```diff
@@ -85,52 +85,52 @@ export type AriaRole =
...
+  'aria-posinset'?: number
+  'aria-rowcount'?: number
+  'aria-rowindex'?: number
+  'aria-rowspan'?: number
 }
 
 export interface AccessibilityAttributes extends AriaWidgetAttributes, AriaRelationshipAttributes {
   role?: AriaRole
-  tabIndex?: string
+  tabIndex?: number
```
layershifter, line 133: `tabIndex` is number even in TS definitions: https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L7779

packages/react/src/lib/accessibility/types.ts
```diff
@@ -85,52 +85,52 @@ export type AriaRole =
...
   role?: string
-  'aria-autocomplete'?: string
-  'aria-checked'?: string | boolean
-  'aria-current'?: string
-  'aria-disabled'?: string | boolean
-  'aria-expanded'?: string | boolean
-  'aria-haspopup'?: string
-  'aria-hidden'?: string | boolean
-  'aria-invalid'?: string
+  'aria-autocomplete'?: 'none' | 'inline' | 'list' | 'both'
```
kuzhelov, line 88: could you, please, point to the React type that this prop's type has been taken from? Asking just because it would be much better if this kind of relationship (React prop's type for accessibility's type) would be explicit

layershifter: `HTMLAttributes`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1475
Do you want to leave there a comment with this type?

jurokapsiar (approved):

packages/react/src/lib/accessibility/types.ts
jurokapsiar: @kuzhelov by explicit, do you mean a compile time dependency on React? Please note that behaviors should be plain javascript (or typescript) theoretically usable with other frameworks as well.

kuzhelov: @jurokapsiar, yes, exactly - I would expect that there would be this dependency expressed explicitly. 

>  Please note that behaviors should be plain javascript (or typescript) theoretically usable with other frameworks as well.

While I do understand that we need this to be framework-agnostic, it seems a bit misleading that we are trying to generalize types based on `React` ones. The thing that I am worrying is that now what we are doing, essentially, is introducing this dependency you are talking about, with the only difference that it is _implicit_, compared to the approach I've suggested before. 

Ideally, would expect some sort of common types library (TS `lib.dom.d.ts`?) for HTML attributes that we could rely on, without having React as a dependency

layershifter: > that we are trying to generalize types based on React ones.

But React types are correct in this case...

kuzhelov (approved):

PR #1089, opened 2019-03-21
fix(Flex): FlexProps extends UIComponentProps

### LINKED ISSUE #1078: FlexProps and FlexItemProps do not extend UIComponentProps
## Bug Report

`FlexProps` and `FlexItemProps` does not extend `UIComponentProps` and `ChildrenComponentProps`

### Version
0.23.1

miroslavstastny:
`FlexProps` now extends `UIComponentProps`.
Fixes #1078


packages/react/src/components/Flex/Flex.tsx
```diff
@@ -2,11 +2,11 @@ import * as PropTypes from 'prop-types'
...
 import * as _ from 'lodash'
 
-import { UIComponent, commonPropTypes } from '../../lib'
+import { UIComponent, commonPropTypes, UIComponentProps, ChildrenComponentProps } from '../../lib'
 import { ReactProps } from '../../types'
 import FlexItem from './FlexItem'
 
-export interface FlexProps {
+export interface FlexProps extends UIComponentProps, ChildrenComponentProps {
   [key: string]: any
```
layershifter: This can be removed now 🔧 

```suggestion
```

packages/react/src/components/Flex/FlexItem.tsx
```diff
@@ -2,11 +2,11 @@ import * as React from 'react'
...
 import cx from 'classnames'
 import * as _ from 'lodash'
-import { UIComponent, commonPropTypes } from '../../lib'
+import { UIComponent, commonPropTypes, UIComponentProps, ChildrenComponentProps } from '../../lib'
 import { mergeStyles } from '../../lib/mergeThemes'
-import { Extendable } from '../../types'
+import { ReactProps } from '../../types'
 
-export interface FlexItemProps {
+export interface FlexItemProps extends UIComponentProps, ChildrenComponentProps {
```
layershifter:
```suggestion
export interface FlexItemProps extends UIComponentProps, ChildrenComponentProps<React.ReactElement<any> | ({ styles: TYPE, classes: string }) => React.ReactElement<any>> {
```

layershifter: `children` definition should be corrected as I see

kuzhelov (approved):

miroslavstastny:
This does not fix pointing menu which uses `::after` on selected item.

layershifter (approved):

PR #1090, opened 2019-03-21
fix(renderer): disable `devMode` for Fela by default

### LINKED ISSUE #1088: Disable Fela `devMode` by default
### Problem description
Fela with `devMode=true` is very slow, we should disable this behavior for consumers by default.

layershifter:
Fixes #1088.

---

When Fela's renderer is in `devMode` (default before), the `insertRule()` function uses a de-optimized code path (`insertRuleInDevMode.js`) for writing style updates to the style nodes.  It replaces the text of the style node in the head, opposed to using the browser's optimized `CSSStyleSheet.insertRule()` method.

It's friendly for development tools because you can edit styles, but it provides wrong development experience for consumers because components are much slower.

---

## `devMode=true`

image

## `devMode=false`

image


miroslavstastny (approved):
Not sure I like it, but

kuzhelov (approved):

PR #1091, opened 2019-03-21
docs: layout guide

kuzhelov:
Introduces first version of Stardust Layout guide. 

This version is specifically focused on 
- delivering principles that should guide towards selecting one layout component over another
- providing key "dont's" to prevent improper use of Stardust components

Apart from that, several style issues in Stardust guides were addressed.


docs/src/views/Layout.tsx
```diff
@@ -0,0 +1,212 @@
...
+  grid: link('Grid', '/components/grid'),
+  segment: link('Segment', '/components/segment'),
+}
+
+export default () => (
+  <DocPage title="Layout">
+    <Header as="h2">Overview</Header>
+    <p>The following components are introduced by Stardust to handle layout aspects:</p>
+    <ul>
+      <li>{links.flex} - designed for layout in one dimension, either a row or a column.</li>
```
kuzhelov, line 30: seems that `Box` component, which is currently seen as primitive layout wrapper (0-D layout) should be also added to the scope here

kuzhelov: agreed to not advertise Box component as a way to handle layout aspects

docs/src/components/GuidesNavigationFooter.tsx
```diff
@@ -0,0 +1,45 @@
...
+          content={previous.name}
+          icon="arrow left"
+          iconPosition="before"
+          primary
+          to={previous.url}
+        />
+      )}
+      {next && (
+        <Button
+          as={NavLink}
```
layershifter: Actually, can be just `Link` because `NavLink` offers additional styling that will never work: https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/NavLink.md

docs/src/views/Layout.tsx
```diff
@@ -0,0 +1,212 @@
...
+          </RelativePositioned>
+        )
+      }}
+    />
+    <p>
+      While it might seem that the intent is addressed with the approach taken, however, this is
+      wrong for the following reason:
+    </p>
+    <blockquote>
+      Stardust makes it very important to{' '}
```
layershifter, line 108:
```suggestion
      Stardust UI makes it very important to{' '}
```

_nit_

docs/src/views/Layout.tsx
```diff
@@ -0,0 +1,212 @@
...
+      <li>
+        do I only need to control the layout by row <strong>or</strong> column? – if yes, use{' '}
+        {code('Flex')}.
+      </li>
+      <li>
+        do I need to control the layout by row <strong>and</strong> column? – if yes, use{' '}
+        {code('Grid')}.
+      </li>
+    </ul>
+    <Header as="h3">Note on Segment component</Header>
```
layershifter: Actually, we should have the same note about `Box` component

docs/src/views/Layout.tsx
```diff
@@ -0,0 +1,212 @@
...
+      considered for removal.
+    </blockquote>
+    <ExampleSnippet
+      render={() => (
+        <Flex gap="gap.small">
+          <Button content="Accept" />
+          <Flex.Item>
+            <Button content="Deny" />
+          </Flex.Item>
+        </Flex>
```
layershifter, line 154: I like the way how it's done on `overreacted.io`:
image

---

image

Something like this will be really nice :)

docs/src/views/Layout.tsx
```diff
@@ -0,0 +1,212 @@
...
+      `}
+    />
+    <ExampleSnippet
+      render={() => {
+        const OpaqueBackground = ({ children, color }) => (
+          <Segment styles={{ backgroundColor: color }} content={{ children }} />
+        )
+
+        return (
+          <OpaqueBackground color="red">
```
layershifter: image


nit: Can use we less toxic color? 🌞 It triggers all my focus to it

docs/src/views/Layout.tsx
layershifter: nit: We can move this section after `Flex recipes`

docs/src/views/Layout.tsx
```diff
@@ -0,0 +1,212 @@
...
+
+const links = {
+  flex: link('Flex', '/components/flex'),
+  flexItem: link('Flex.Item', '/components/flex'),
+  grid: link('Grid', '/components/grid'),
+  segment: link('Segment', '/components/segment'),
+}
+
+export default () => (
+  <DocPage title="Layout">
```
layershifter: I think that `Layout components` will be more descriptive

docs/src/views/Layout.tsx
kuzhelov: ### Notes on Box component
speaking of the `Box` component - it seems that the first thing we should do is rather fix its description. I've intentionally consulted about whether the `Box` component is seen now as a layout one, given the recent change of its description: https://github.com/stardust-ui/react/blob/master/packages/react/src/components/Box/Box.tsx#L20

Thus, lets just make things consistent - change the description of `Box` component and put an update to the guides then - otherwise it will just create conflicting thoughts in the client's head.

### Why not put after Flex recipies
This is, actually, done intentionally. Given current usage statistics, the problems that are described in this 'Notes' section is much more serious (especially taking their conequences into account) rather than the ones introduced by overuse of, say, `Flex.Item`. Thus, it is very important to make this note earlier than later. 

Another point relates to where this paragraph semantically attributes to: it definitely should be a part of the section where Stardust layout components are enumerated, with the corresponding use-cases for each - and thus I would rather leave it here and won't introduce a separate section for it.

docs/src/views/Layout.tsx
kuzhelov: it turns out that we are using Stardust/Stardust UI in our docs interchangeably. I would say that this is ok (due to that, when repeated frequently in text, "Stardust UI" is too formal and introduce unnecessary reading 'bumps'). In the context of this sentence "Stardust" refers to rather the set of underlying _principles_ - and, thus, this version (and "Stardust UI" one) is used, much like it is done in the README.

Thus, I would leave it as it is, take existing naming convention and just follow it

docs/src/views/Layout.tsx
kuzhelov: yes, fully agree - I would say, anything that would allow us to have comments preserved would be already a huge win :) for now can do either of two:
- use the approach currently taken
- or write code twice - one for example's code, one for the actual rendering

Given these options, have chosen the first approach because, essentially, it will at least guarantee that these two things (code and rendered output) won't go out of sync

docs/src/views/Layout.tsx
kuzhelov: sure :)

kuzhelov: image

docs/src/views/Layout.tsx
layershifter: We added note about `Segment`, it's cool 👍 

But, I want avoid the following thing: "_Okay, I can't use `Segment`... Oh, we have have `Box`, let's do `<Box styles={{ display: 'flex' }} />`!_" I.e. `Box` can be abused for layouts purposes, too.

docs/src/views/Layout.tsx
layershifter: Let's leave as is for now, duplication is more evil 😈

layershifter (approved):

docs/src/views/Layout.tsx
kuzhelov: agree - then, if we agree on fixing `Box` description (which now states that it is a _layout_ component), then let me do that and introduce a dedicated section to the `Layout` guide

hughreeling:
Could we add linked to the layout guide in the Flex, Segment, Box, Layout components?  (Put in description as link?)

Let's also update description of Box and Segment and explicitly say what they're used for and not used for.

Having this information duplicated isn't a bad thing....

PR #1093, opened 2019-03-22
feat(components): align slotClassNames property for all components

bmdalex:
### Description

This PR:
- introduces ability to select submenus from `MenuItem` components: **`MenuItem.slotClassNames.menu`**
- aligns `slotClassNames` property on composed components in order to be able to target subcomponents directly;
**e.g.:** a `Menu` item can be targeted referencing:
  - it's `className` directly: `MenuItem.className` // even before this PR
  - the item as a slot of the parent component: **`Menu.slotClassNames.item`** // introduced in this PR

### DIFF
--- a/packages/react/src/components/Header/Header.tsx
+++ b/packages/react/src/components/Header/Header.tsx
@@ -95,6 +95,7 @@ class Header extends UIComponent<ReactProps<HeaderProps>, any> {
         {!hasChildren &&
           HeaderDescription.create(description, {
             defaultProps: {
+              className: Header.slotClassNames.description,
               variables: {
                 ...(v.descriptionColor && { color: v.descriptionColor }),
               },
--- a/packages/react/src/components/Header/Header.tsx
+++ b/packages/react/src/components/Header/Header.tsx
@@ -18,6 +18,10 @@ import { Accessibility } from '../../lib/accessibility/types'
 import { defaultBehavior } from '../../lib/accessibility'
 import { ReactProps, ShorthandValue } from '../../types'
 
+export interface HeaderSlotClassNames {
+  description: string
+}
+
 export interface HeaderProps
   extends UIComponentProps,
     ChildrenComponentProps,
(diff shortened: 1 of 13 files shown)


packages/react/src/components/Header/Header.tsx
```diff
@@ -18,6 +18,10 @@ import { Accessibility } from '../../lib/accessibility/types'
 import { defaultBehavior } from '../../lib/accessibility'
 import { ReactProps, ShorthandValue } from '../../types'
 
+export interface HeaderSlotClassNames {
+  description: string
```
kuzhelov, line 22: a very important thing is the contract that stays behind these class names: what we are promising to the client is that in the rendered tree there will be always a `& .${slotClassNames.description}` element rendered in case if `description` is provided.

However, the following example will break these guarantees:
```jsx
<Header description={<Text>Some description</Text>} ... />
```
image


As we might see, there is no `.ui-header__description` element rendered in the tree, while `description` was actually provided. This problem would occur anytime one would use different component type for the slot.

See this to be an important thing to fix before merge.

bmdalex: This is definitely a problem but it's not introduced by this API. We already have this behavior for all the cases where we're using the `slotClassNames` approach. I cannot think of a solution for this rather than telling clients to avoid children API?

@kuzhelov @mnajdova

bmdalex: discussed with @kuzhelov , I will introduce a fix for this

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -65,9 +70,14 @@ export interface TreeItemProps extends UIComponentProps, ChildrenComponentProps
...
   static create: Function
 
+  static displayName = 'TreeItem'
+
   static className = 'ui-tree__item'
 
-  static displayName = 'TreeItem'
+  static slotClassNames: TreeItemSlotClassNames = {
+    title: TreeTitle.className,
+    tree: `${TreeItem.className}__tree`,
```
kuzhelov: seems that this part rather have a `subtree` semantics. The reason I am worrying about that is because client might wrongly think that this selector is for tree itself, not for its subtree part. May we rename it?

bmdalex: I'd say your recommendation makes sense. let's see also what's @mnajdova 's take on this, we were discussing it for `submenu` vs `menu` so we need to align

kuzhelov (approved):

packages/react/src/components/Tree/TreeItem.tsx
bmdalex: discussed with @mnajdova as well and applied @kuzhelov 's suggestion

PR #1096, opened 2019-03-25
feat(screenshot tests): menu, button, icon & divider component theme tests

jaanus03:
This PR adds screener tests for the Menu, Button, Icon & Divider components to cover additional states & themes.


docs/src/examples/components/Button/commonScreenerSteps.ts
```diff
@@ -0,0 +1,11 @@
+import { Button } from '@stardust-ui/react'
+
+const button = `.${Button.className}`
+
+const getScreenerSteps = (): ScreenerSteps => [
+  builder => builder.hover(button).snapshot('Hover state of the first button'),
```
bmdalex: we usually use 3rd person and verbs for the snapshot text; not sure it's the best convention but let's keep consistent for now:
`.snapshot('Hovers the first button')`

same for all the other instances, I won't spam with comments 😄

docs/src/examples/components/Menu/Usages/MenuExampleToolbar.shorthand.steps.ts
```diff
@@ -0,0 +1,40 @@
...
+  vertical?: boolean
+}
+
+const selectors = {
+  menu: `.${Menu.className}`,
+  item: (itemIndex: number) => `.${Menu.className} li:nth-child(${itemIndex}) a`,
+  lastItem: `.${Menu.className} li:last-child a`,
+}
+
+const getScreenerSteps = ({ vertical }: StepsOptions = {}): ScreenerSteps => [
```
bmdalex: first 4 snapshots in this method are the same as in `getScreenerSteps` from  `docs/src/examples/components/Menu/commonScreenerSteps.ts`; let's reuse that and do:

```typescript
import { Menu } from '@stardust-ui/react'

import getScreenerSteps from '../commonScreenerSteps'

const selectors = {
  menu: `.${Menu.className}`,
  item: (itemIndex: number) => `.${Menu.className} li:nth-child(${itemIndex}) a`,
  lastItem: `.${Menu.className} li:last-child a`,
}

const config: ScreenerTestsConfig = {
  steps: [
    (builder, keys) =>
      getScreenerSteps()[0](builder, keys)
        .click(selectors.lastItem)
        .snapshot('Clicks on the last item and opens submenu')

        .keys(selectors.lastItem, keys.downArrow)
        .snapshot('Focuses on the first element in the submenu'),
  ],
}

export default config
```

I will think of some way to make this more readable, but better to reuse what we have

bmdalex (approved):
👍

PR #1097, opened 2019-03-25
feat(Loader): add support for SVG animations, update in Teams theme

### LINKED ISSUE #1038: Loader needs to work with SVG animation
### Steps
The loader is built for a basic loading animation.

### Expected Result
We need to support an SVG and a more complicated animation.

See https://codepen.io/teams-ux-engineers/pen/1c068c6d136debb86a7164054a2c21bf

layershifter:
Fixes #1038.


packages/react/src/themes/teams/components/Loader/loaderVariables.ts
```diff
@@ -0,0 +1,62 @@
...
+    largest: pxToRem(-6912),
+  },
+  svgWidths: {
+    smallest: pxToRem(38.4),
+    smaller: pxToRem(38.4),
+    small: pxToRem(38.4),
+    medium: pxToRem(57.6),
+    large: pxToRem(115.2),
+    larger: pxToRem(115.2),
+    largest: pxToRem(115.2),
```
layershifter: @codepretty can you please provide correct sizes for these elements to make them less magic?

packages/react/src/components/Loader/Loader.tsx
```diff
@@ -105,14 +105,22 @@ class Loader extends UIComponent<ReactProps<LoaderProps>, LoaderState> {
     const { indicator, label } = this.props
     const { visible } = this.state
 
+    const svgElement =
+      styles.svg && Box.create({}, { defaultProps: { className: 'container', styles: styles.svg } })
```
mnajdova: I don't like the fact the we are changing the structure of the element depending of something that exists in the styles. The user should be aware trough the API of the component that they can specify svg element... More over the same can be created trough the indicator slot anyways. Can you suggest why do you see these changes necessary?

packages/react/src/components/Loader/Loader.tsx
```diff
@@ -102,18 +118,30 @@ class Loader extends UIComponent<ReactProps<LoaderProps>, LoaderState> {
   }
 
   renderComponent({ ElementType, classes, accessibility, variables, styles, unhandledProps }) {
-    const { indicator, label } = this.props
+    const { indicator, label, svg } = this.props
     const { visible } = this.state
 
+    const svgElement = Box.create(svg, {
+      defaultProps: { className: Loader.slotClassNames.indicator, styles: styles.svg },
```
mnajdova: Oops, we are sending the wrong `className` here

packages/react/src/components/Loader/Loader.tsx
```diff
@@ -102,18 +118,30 @@ class Loader extends UIComponent<ReactProps<LoaderProps>, LoaderState> {
   }
 
   renderComponent({ ElementType, classes, accessibility, variables, styles, unhandledProps }) {
-    const { indicator, label } = this.props
+    const { indicator, label, svg } = this.props
     const { visible } = this.state
 
+    const svgElement = Box.create(svg, {
```
mnajdova, line 124: I am wondering whether the better shorthand for the svg would be the `Icon` component.. Can you elaborate why we are using the `Box` here?

packages/react/src/themes/teams/components/Loader/loaderStyles.ts
```diff
@@ -0,0 +1,85 @@
...
+import { ComponentStyleFunctionParam, ICSSInJSStyle } from '../../../types'
+import { LoaderVariables } from './loaderVariables'
+
+export default {
+  indicator: ({
+    props: p,
+    variables: v,
+  }: ComponentStyleFunctionParam<LoaderProps, LoaderVariables>): ICSSInJSStyle => ({
+    // Reset existing styles from base theme
+    animationName: 'none',
```
mnajdova, line 11: I wish we didn't have to reset here all existing styles from the base theme, but I don't have any other proposal. The styles for the base theme seems reasonable for it's usage

packages/react/src/themes/teams/components/Loader/loaderStyles.ts
```diff
@@ -0,0 +1,85 @@
...
+      animationFillMode: 'both',
+      animationPlayState: 'running',
+      animationTimingFunction: 'steps(60, end)',
+      animationIterationCount: 'infinite',
+    }
+
+    return {
+      ...outerAnimation,
+
+      ':before': {
```
mnajdova, line 72: How does this works? why we have animations for the `svg` and the `:before`? Can we somehow move this to the `indicator` and the `svg` accordingly?

mnajdova: Is this because of the `svg` that is used?

packages/react/src/themes/teams/components/Loader/loaderStyles.ts
layershifter: Yep, it's absolutely not funny, but I don't have a better idea now...

packages/react/src/themes/teams/components/Loader/loaderStyles.ts
layershifter: Idea with `svg` slot works fine, but [Teams loader](https://codepen.io/teams-ux-engineers/pen/1c068c6d136debb86a7164054a2c21bf) requires specific markup, i.e need one more element. Here is match:
image

I'm opened for better idea...

packages/react/src/components/Loader/Loader.tsx
layershifter: Can you please clarify why it can be `Icon`?  For example, this loader is an extra big SVG:

image

And actually, I don't see any way how spinners below can be used as icons:

image

mnajdova: Make sense, I mention icon only because it is an svg

mnajdova (approved):

Issue #964, opened 2019-02-25
Grid: Keyboard navigation not working properly for grid items when focus set programmatically


johannao76:
### Steps
When setting focus programmatically to an item inside the Grid component, subsequently trying to use keyboard navigation sets focus on the wrong item.

Repro:
- Run the sample provided under the Tescase section of this bug report
- Enter the value 2 in the textbox and hit Enter
- Noticed that item at index #2 is now focused
- Hit the down arrow to navigate to the item below

### Expected Result
- Item below (index: 6) should be focused 

### Actual Result
- Item below is not highlighted. Item at index 4 is selected instead

### Version
0.21.1

### Testcase
https://codesandbox.io/s/9y5qylv8o4

PR #1098, opened 2019-03-25
fix(FocusZone): Update tabindexes and alignment when item is focused programatically

sophieH29:
Fixes https://github.com/stardust-ui/react/issues/964

Changes made:
- _onFocus handler should only reassign active element to a new one element, do not reassign the same
- when an item is focused programmatically, tabindexes should be updated accordingly and focus alignment updated when a direction is bidirectional
- `gridBehavior` should not be circular


jurokapsiar (approved):

mnajdova (approved):
Would recommend adding unit tests for these to avoid regressions in the future.

packages/react/test/specs/lib/FocusZone-test.tsx
```diff
@@ -1286,4 +1286,34 @@ describe('FocusZone', () => {
...
+      </div>,
+    )
+
+    const focusZone = ReactDOM.findDOMNode(component)!.firstChild as Element
+
+    const buttonA = focusZone.querySelector('#a') as HTMLElement
+    const buttonB = focusZone.querySelector('#b') as HTMLElement
+
+    // ButtonA should be focussed.
+    ReactTestUtils.Simulate.focus(buttonA)
```
miroslavstastny, line 1311: nit: I would check `tabIndex` also before the first call to `focus()`

miroslavstastny (approved):

PR #1100, opened 2019-03-25
fix(ChatMessage): show timestamp if reactionGroup is provided

mnajdova:
There is update for the reactions + timestamp in Teams theme. When there are reaction on the message, the timestamp should always be shown.


packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -50,6 +50,9 @@ export interface ChatMessageProps
   /** Menu with actions of the message. */
   actionMenu?: ShorthandValue
 
+  /** Controls messages's relation to other chat messages. It is automatically set from the ChatItem */
```
kuzhelov: nit-nit: lets end description with dot, also, we might consider to use `by` instead of `from`, something like:

```
 Controls messages's relation to other chat messages. Is automatically set by the ChatItem.
```

packages/react/src/components/Chat/ChatItem.tsx
```diff
@@ -89,23 +90,62 @@ class ChatItem extends UIComponent<ReactProps<ChatItemProps>, any> {
...
+  getMessageAugmentedWithChatMessageProps = styles => {
+    const { message, attached } = this.props
+    const messageElement = Box.create(message, {
+      defaultProps: {
+        className: ChatItem.slotClassNames.message,
+        styles: styles.message,
+      },
+    })
+
+    // the box element is ChatMessage
```
kuzhelov: comment may be a bit misleading - what we are testing here is not the `Box` element, but rather if messageElement is ChatMessage. Lets just omit `box` from the comment text

packages/react/src/components/Chat/ChatItem.tsx
```diff
@@ -89,23 +90,62 @@ class ChatItem extends UIComponent<ReactProps<ChatItemProps>, any> {
...
+    if (this.checkIfElementIsChatMessage(messageElement)) {
```
kuzhelov: nit: we may shorten the name of predicate method:

```js
if (this.isChatMessageElement(messageElement)) {
   ..
}
```

packages/react/src/components/Chat/ChatItem.tsx
```diff
@@ -89,23 +90,62 @@ class ChatItem extends UIComponent<ReactProps<ChatItemProps>, any> {
...
+
+    // the content element is ChatMessage
+    if (this.checkIfElementIsChatMessage(messageElement, 'content')) {
+      return this.cloneElementWithCustomProps(messageElement, { attached }, 'content')
+    }
+    return messageElement
+  }
+
+  checkIfElementIsChatMessage = (element, prop?) => {
+    return _.get(element, `${prop ? `props.${prop}.` : ''}type.__isChatMessage`)
```
kuzhelov: I would suggest to make this method's logic as simple as possible - then it would be much easier to reason of what this method is supposed to do at the place of usage:

```jsx
// just single parameter
isChatMessageElement = (element) => ...
```

As a next step, we might notice that this method relies on the aspects that are provided by another type, i.e. `ChatMessage`'s `__isChatMessage` static prop, which name might change. Thus, I would suggest to put this check as part of the `ChatMessage` component's logic:

```jsx
class ChatMessage extends ... {
  static isChatMessageElement(element) => ...
}
```

And, finally - while I do see the benefit of introducing second arg to original method to avoid duplication of prop fetching logic, we might just want to introduce a dedicated method for props fetching for that:

```jsx
getElementProp = (element, propName) => { .. }
```

once again, with preserving the benefit of reducing duplication, we would have yet additional one: this method is quite easy to reason about on the client's side.

As a result, we might end with something like:

```jsx
 if (ChatMessage.isTypeOfElement(messageElement) { .. }
 if (ChatMessage.isTypeOfElement(getElementProp(messageElement, 'content')) { .. }
```

packages/react/src/components/Chat/ChatItem.tsx
```diff
@@ -89,23 +90,62 @@ class ChatItem extends UIComponent<ReactProps<ChatItemProps>, any> {
...
+        styles: styles.message,
+      },
+    })
+
+    // the box element is ChatMessage
+    if (this.checkIfElementIsChatMessage(messageElement)) {
+      return this.cloneElementWithCustomProps(messageElement, { attached })
+    }
+
+    // the children element is ChatMessage
```
kuzhelov: given that `children`, theoretically, may be a function, the following comment would be more correct:

> the children is ChatMessage element

the same applies to the `content`

kuzhelov (approved):

packages/react/src/components/Chat/ChatItem.tsx
```diff
@@ -89,23 +92,58 @@ class ChatItem extends UIComponent<ReactProps<ChatItemProps>, any> {
...
-        {Box.create(message, {
-          defaultProps: { className: ChatItem.slotClassNames.message, styles: styles.message },
-        })}
+        {messageElement}
         {contentPosition === 'end' && gutterElement}
       </>
     )
   }
+
+  getMessageAugmentedWithChatMessageProps = styles => {
```
kuzhelov: just to better deliver the main intent of this method - lets rename if to something like `setAttachedPropValueForChatMessage`

packages/react/src/components/Chat/ChatItem.tsx
```diff
@@ -89,23 +92,58 @@ class ChatItem extends UIComponent<ReactProps<ChatItemProps>, any> {
...
+
+    // the children is ChatMessage
+    if (ChatMessage.isTypeOfElement(getElementProp(messageElement, 'children'))) {
+      return this.cloneElementWithCustomProps(messageElement, { attached }, 'children')
+    }
+
+    // the content is ChatMessage
+    if (ChatMessage.isTypeOfElement(getElementProp(messageElement, 'content'))) {
+      return this.cloneElementWithCustomProps(messageElement, { attached }, 'content')
+    }
```
layershifter, line 135: Can you please clarify this for me? Will we add `attached` prop only is `element`/`children`/`content` are `ChatItem`?
But, any HOC will break this: https://codesandbox.io/s/n4mm1ny55l 💣

mnajdova: Yes, by default we will propagate the attached in these three use cases (which we see as the most common) - the attached prop can always be applied on the ChatMessage if it is wrapped... I am open for refactoring it if we have some better ideas.. (once the template mechanism will be available to the user, we can safely omit these changes)

PR #1103, opened 2019-03-26
fix(build): remove prepush

### LINKED ISSUE #1102: Do not check git tags in prepush
### Steps
```sh
$ gp              
husky > npm run -s prepush (node v8.11.3)

[11:22:15] Requiring external module ts-node/register
[11:22:24] Using gulpfile ~/devel/stardust-ui-react/gulpfile.ts
[11:22:24] Starting 'git:prepush'...
[11:22:24] Finished 'git:prepush' after 19 ms
```

prepush checks tags. The check was necessary to keep integrity of tags after moving from semantic-ui/react. It is hopefully no longer necessary, remove the step to save up to 10s with every push

miroslavstastny:
Remove prepush step as it is no longer needed.
Fixes #1102


mnajdova:
I am still seeing the pre-push failing sometimes, and I have to remove and pull all tags again... Not sure if this is the case with other developers..

kuzhelov:
in that case lets take a look on @mnajdova 's case and leave this functionality till the end of investigation - @miroslavstastny, what do you think?

miroslavstastny:
We do not have an agreement that we can remove this.

miroslavstastny:
We believe we have an agreement now, reopening...

layershifter (approved):

PR #1104, opened 2019-03-26
fix(docs): fix snippets in Layout docs

layershifter:
image

Fixes issue in https://stardust-ui.github.io/react/layout.
The root reason is missing display names in production.


docs/src/components/ExampleSnippet/ExampleSnippet.tsx
```diff
@@ -24,6 +24,13 @@ const ExampleSnippet = ({ render = () => null, value }: ExampleSnippetProps) =>
   const string =
     value ||
     reactElementToJSXString(element, {
+      displayName: (element: React.ReactElement<any>): string =>
```
layershifter, line 27: https://github.com/algolia/react-element-to-jsx-string#reactelementtojsxstringreactelement-options

layershifter: Original function uses `No Display Name`, that breaks `prettier.parser.babylon`

https://github.com/algolia/react-element-to-jsx-string/blob/0b17eaf3decafac13f276e705d4c850418cd4bb7/src/parser/parseReactElement.js#L19

miroslavstastny (approved):
Changelog entry for DOC?

layershifter:
> Changelog entry for DOC?

Not sure, this looks more like a hotfix...

docs/src/views/Layout.tsx
```diff
@@ -76,6 +76,7 @@ export default () => (
           </OpaqueBackground>
         )
       }}
+      value={`<OpaqueBackground color="lightyellow">This is a bad approach to opaque background :(</OpaqueBackground>`}
```
kuzhelov, line 79: for this component we need to introduce display name as well, will address in a separate PR

PR #1108, opened 2019-03-26
feat(Embed|Video): add components

stuartlong:
# API Proposal
The goal of this proposal is to provide components similar to Image that work with video sources. 

1. The Video component is a very basic wrapper around the <video tag>
2. The Embed component is a more complex one that displays either an embedded video (GIF) or an embedded iFrame

## Video Component
Essentially just the normal <video> tag that supports `width` and `height` as variables.

```tsx
 <Video
      src="https://raw.githubusercontent.com/bower-media-samples/big-buck-bunny-480p-30s/master/video.mp4"
      poster="https://raw.githubusercontent.com/bower-media-samples/big-buck-bunny-480p-30s/master/poster.jpg"
      variables={{ width: '600px' }}
  />
```

### Problems

1. Video is played through default HTML5 video controls. These controls are accessible, but only via the tab key. Ideally, a user would tab to the video and then use the arrow keys to navigate through the video controls, but not sure how to get this to work within the stardust framework

2. Muted attribute doesn't work. I can't get the muted attribtue to appear on the outputted html:
    ```tsx
      <Video muted />
    ```
    and
    ```tsx
    <Video muted="true" />
    ```
    both result in <video /> being rendered. I think this is because Stardust is seeing "muted" as a boolean   so it turned `muted="true"` and `muted` into `muted=""` which the browser strips out because it seem to explicitly want `muted` or `muted="true"`.

## Embed Component
This component displays either an Image or Video comppnent based on whether the GIF is currently playing. We do this swap out due to memory issues with the <video> control in Chrome. The component also adds a play/pause icon superimposed over the GIF:
image

If a video isn't provided, an iFrame can be provided to embed instead

The component also adds key handling for space / enter when focused on the gif to play/pause it.

```tsx
 <Emned
      video="https://raw.githubusercontent.com/bower-media-samples/big-buck-bunny-480p-5s/master/video.mp4"
      poster="https://raw.githubusercontent.com/bower-media-samples/big-buck-bunny-480p-5s/master/poster.jpg"
      variables={{ width: '600px' }}
      active={true}
  />
```
### Problems
1. Displaying SVG as ::after content. I display the play/pause button as an `::after` pseudoelement so it doesn't take up any slots or room in the DOM. However, I couldn't find a way to use the stored icons SVG and instead had to hard code the svg within the styles file itself.

## TODO

- [ ] Conformance test
- [ ] Stardust base theme
- [ ] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [ ] Confirm RTL usage
- [ ] W3 accessibility check
- [ ] Stardust accessibility check
- [ ] Update the CHANGELOG.md


layershifter (commented):
@stuartlong great work, however from components perspective I want to add some separation there.

### Video

We can definitely add this component and then iterate on it to apply styling to controls and add more props.

### VideoGif

Instead of it, I propose to add `Embed` component that will allow to handle not only videos, but even iframes. See for reference, https://react.semantic-ui.com/modules/embed/

1. Rename to `VideoGif` to `Embed`
2. Add `video` shorthand slot, add `iframe` shorthand slot and use `Box` component

```
render() {
  const { iframe, video } = this.props

  return (
    <>
      {Video.create(video)}
      {Box.create(iframe, { defaultProps: { as: 'iframe' } })}
     </>
   )
}
```
3. As we are trying to match SUIR API, I suggest to rename `poster` prop to `placeholder` and use shorthand for `Image`:
```
render() {
  const { placeholder } = this.props

  return (
    <>
      {Image.create(placeholder)}
     </>
   )
}
```

docs/src/components/Playground/renderConfig.ts
```diff
@@ -6,6 +6,7 @@ import * as ReactDOM from 'react-dom'
 export const babelConfig = {
   plugins: [
     'proposal-class-properties',
+    'proposal-object-rest-spread',
```
layershifter: Please merge latest `master`, this change is already there 😺

packages/react/src/components/Video/Video.tsx
```diff
@@ -0,0 +1,95 @@
...
+ * An video is a graphicical and audio representation of something.
+ * @accessibility
+ * If Video should be visible to screen readers, textual representation needs to be provided in 'alt' property.
+ *
+ * Other considerations:
+ *  - when alt property is empty, then Narrator in scan mode navigates to image and narrates it as empty paragraph
+ *  - when image has role='presentation' then screen readers navigate to the element in scan/virtual mode. To avoid this, the attribute "aria-hidden='true'" is applied by the default image behavior
+ *  - when alt property is used in combination with aria-label, arialabbeledby or title, additional screen readers verification is needed as each screen reader handles this combination differently.
+ */
+class Video extends UIComponent<ReactProps<VideoProps>, any> {
```
layershifter:
```suggestion
class Video extends UIComponent<ReactProps<VideoProps>> {
```

packages/react/src/components/Video/Video.tsx
```diff
@@ -0,0 +1,95 @@
...
+   * Accessibility behavior if overridden by the user.
+   * @default videoBehavior
+   * */
+  accessibility?: Accessibility
+
+  /** Video source URL. */
+  src?: string
+
+  /** Image source URL for when video isn't playing */
+  poster?: string
```
layershifter: Cruft?

packages/react/src/components/Video/Video.tsx
```diff
@@ -0,0 +1,95 @@
...
+  /** Whether to display the native video controls */
+  controls?: boolean
+
+  /** Whether the video should automatically restart after it ends */
+  loop?: boolean
+}
+
+/**
+ * An video is a graphicical and audio representation of something.
+ * @accessibility
```
layershifter: I suggest to drop accessibility part for initial implementation at all and introduce it later with sync with our accessibility team. What do you think?

packages/react/src/components/Video/VideoGif.tsx
```diff
@@ -0,0 +1,136 @@
...
+  static defaultProps = {
+    as: 'span',
+    accessibility: videoGifBehavior as Accessibility,
+  }
+
+  public state = {
+    isPlaying: this.props.playing,
+  }
+
+  shouldComponentUpdate(nextProps: VideoGifProps, nextState: VideoGifState): boolean {
```
layershifter: image

---

We have `AutoControlledComponent` for this 😺 

1. Remove `shouldComponentUpdate()` at all
2. Use `AutoControlledComponent`:

```diff
-class VideoGif extends UIComponent<ReactProps<VideoGifProps>, VideoGifState> {
+class VideoGif extends AutoControlledComponent<ReactProps<VideoGifProps>, VideoGifState> {
```
3. Add `autoControlledProps`:
```
static autoControlledProps = ['playing']
```
4. Add `defaultPlaying` to `VideoGifProps` and `propTypes`
5. Use `playing` from `this.state` 🚀  and replace `this.setState({ playing: X })` with `this.trySetState({ playing: X })`

packages/react/src/components/Video/VideoGif.tsx
```diff
@@ -0,0 +1,136 @@
...
+  accessibility?: Accessibility
+
+  /** Video source URL. */
+  src?: string
+
+  /** Image source URL for when video isn't playing */
+  poster?: string
+
+  /** whether the gif should be playing */
+  playing?: boolean
```
layershifter: Instead of `playing` I suggest to use `active` prop 👍

packages/react/src/themes/teams/components/Icon/svg/icons/pause.tsx
```diff
@@ -0,0 +1,16 @@
...
+  icon: ({ classes }) => (
+    <svg role="presentation" focusable="false" viewBox="0 0 32 32" className={classes.svg}>
+      <g className={cx(teamsIconClassNames.filled, classes.filledPart)}>
+        <path d="M12.5,22V10h1v12H12.5z" />
+        <path d="M18.5,22V10h1v12H18.5z" />
+      </g>
+    </svg>
+  ),
+  styles: {},
+} as TeamsProcessedSvgIconSpec
```
layershifter: We definitely need these icons, but can we introduce them in a separate PR? Even right now 🙏

packages/react/src/components/Video/VideoGif.tsx
stuartlong: Interesting, I'm only worried that won't work for the case where the video state is playing and the component parent changes the defaultPlaying prop to false. In that case, we'd want to set the playing state to false. That's really the only reason I implemented shouldComponentUpdate

packages/react/src/themes/teams/components/Icon/svg/icons/pause.tsx
stuartlong: Sure, will do

packages/react/src/components/Video/Video.tsx
stuartlong: Sounds good for Video. For Video Gif, I'd like to keep in the key actions so that space/enter on GIFs still play/pause them. But if there's some issue there too we can take that up after the accessibility sync.

packages/react/src/components/Video/Video.tsx
stuartlong: Which part?

docs/src/components/Playground/renderConfig.ts
stuartlong: Mhm, ya not sure how this got here on my fork

packages/react/src/themes/teams/components/Icon/svg/icons/pause.tsx
stuartlong: Actually, it looks like these exist in ProcessedIcons already. What's the difference?

packages/react/src/themes/teams/components/Video/videoGifStyles.ts
```diff
@@ -0,0 +1,57 @@
...
+        background: `0 no-repeat rgba(0,0,0,.25)`,
+        backgroundPositionX: props.isPlaying ? pxToRem(3) : pxToRem(3),
+        borderWidth: pxToRem(2),
+        borderStyle: 'solid',
+        borderColor: 'transparent',
+        fontWeight: 600,
+        fontSize: pxToRem(24),
+        opacity: props.isPlaying ? 0 : 1,
+        transition: 'opacity .22s ease-in-out',
+        content: `url("data:image/svg+xml;charset=utf-8,%3Csvg%20xmlns%3D%22http%3A%2F%2Fwww.w3.org%2F2000%2Fsvg%22%20viewBox%3D%220%200%2032%2032%22%3E%3C${
```
stuartlong: @layershifter is there a way I can use the actual icon files here instead of hard coding it?

packages/react/src/components/Video/VideoGif.tsx
stuartlong: Basically whichever one changed last should be the source of truth

packages/react/src/components/Video/VideoGif.tsx
stuartlong: `active` or `defaultPlaying`, since you suggested that below 😃

docs/src/components/Playground/renderConfig.ts
layershifter: Merged `origin/master` 👍

packages/react/src/components/Video/Video.tsx
layershifter: `poster` prop is not used in this component :)

packages/react/test/specs/components/Embed/Emed-test.tsx
```diff
@@ -0,0 +1,7 @@
+import { isConformant } from 'test/specs/commonTests'
```
layershifter: Typo in filename:
`packages/react/test/specs/components/Embed/Emed-test.tsx` => `packages/react/test/specs/components/Embed/Embed-test.tsx`

packages/react/src/themes/teams/components/embed/embedVariables.ts
```diff
@@ -0,0 +1,5 @@
+export default siteVariables => ({
```
layershifter: Let's introduce a type for this:

```
export interface EmbedVariables {
  width: number
  height: number
  focusBorderColor: string
}
```

packages/react/src/themes/teams/components/embed/embedStyles.ts
```diff
@@ -0,0 +1,57 @@
...
+        margin: `${pxToRem(-30)} 0 0 ${pxToRem(-30)}`,
+        textAlign: 'center',
+        pointerEvents: 'none',
+      },
+      ':hover::after': {
+        opacity: 1,
+      },
+    }
+  },
+} as ComponentSlotStylesInput<EmbedProps & EmbedState, any>
```
layershifter:
```suggestion
} as ComponentSlotStylesInput<EmbedProps & EmbedState, EmbedVariables>
```

packages/react/src/themes/teams/components/embed/embedStyles.ts
```diff
@@ -0,0 +1,57 @@
+import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
+import { pxToRem } from 'src/lib'
+import { EmbedProps, EmbedState } from '../../../../components/Embed/Embed'
+
+export default {
+  root: ({ props, variables }): ICSSInJSStyle => {
```
layershifter:
```suggestion
  root: ({ props: p, variables: v }): ICSSInJSStyle => {
```

We decided to use `p` & `v` as it is more obvious in styles

packages/react/src/themes/teams/components/embed/embedStyles.ts
```diff
@@ -0,0 +1,57 @@
...
+    const isFromKeyboard = { props }
```
layershifter:
```suggestion
```

packages/react/src/themes/teams/components/embed/embedStyles.ts
```diff
@@ -0,0 +1,57 @@
...
+    return {
+      boxSizing: 'border-box',
+      display: 'inline-block',
+      verticalAlign: 'middle',
+      position: 'relative',
+      cursor: 'pointer',
+      width: variables.width,
+      height: variables.height || 'auto',
+      outline: 0,
+      ...(isFromKeyboard && {
```
layershifter:
```suggestion
      ...(p.isFromKeyboard && {
```

packages/react/src/themes/teams/components/embed/embedStyles.ts
```diff
@@ -0,0 +1,57 @@
...
+        ':focus': {
+          '::after': {
+            borderColor: variables.focusBorderColor,
+            opacity: 1,
+          },
+        },
+      }),
+      '::after': {
+        background: `0 no-repeat rgba(0,0,0,.25)`,
+        backgroundPositionX: props.isPlaying ? pxToRem(3) : pxToRem(3),
```
layershifter:
```suggestion
        backgroundPositionX: pxToRem(3),
```

packages/react/src/themes/teams/components/Video/videoVariables.ts
```diff
@@ -0,0 +1,4 @@
+export default () => ({
+  width: undefined,
+  height: undefined,
+})
```
layershifter, line 9: Let's add an interface for these variables

packages/react/src/lib/accessibility/Behaviors/Embed/embedBehavior.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Enter/space keys play and pause the gif respectively
+ *
+ * @specification
+ * Sets 'aria-hidden' unless alt or title attribute provided
+ */
+const embedBehavior: Accessibility = (props: any) => ({
+  attributes: {
+    root: {
+      tabIndex: '0',
+      'aria-hidden': props['alt'] || props['title'] ? undefined : true,
```
layershifter:
```suggestion
      'aria-hidden': props.alt || props.title ? undefined : true,
```

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -0,0 +1,145 @@
...
+        {Box.create(iframe, { defaultProps: { as: 'iframe' } })}
+      </ElementType>
+    )
+  }
+
+  protected actionHandlers: AccessibilityActionHandlers = {
+    performClick: event => this.handleClick(event),
+  }
+
+  private handleClick = e => {
```
layershifter:
```suggestion
  handleClick = e => {
```

We decided drop visibility modifiers in our code and enforce order by plugin, but for please use following order:

```
static ...

actionHandlers
handleClick 

render
```

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -0,0 +1,145 @@
...
+          : Image.create(placeholder, {
+              defaultProps: {
+                styles: styles.image,
+                variables: {
+                  width: variables.width,
+                  height: variables.height,
+                },
+              },
+            })}
+        {Box.create(iframe, { defaultProps: { as: 'iframe' } })}
```
layershifter: I want to correct logic there:
```
{playing ? (
  <>
    {Box.create(iframe, { defaultProps: { as: 'iframe' } })}
    {Video.create(video)}
  </>
) :  Image.create(placeholder)
```

We want display `Video` or `iframe`

packages/react/src/components/Video/Video.tsx
```diff
@@ -0,0 +1,74 @@
...
+        controls={controls}
+        autoPlay={autoPlay}
+        loop={loop}
+        muted={muted}
+      />
+    )
+  }
+}
+
+Video.create = createShorthandFactory({ Component: Video })
```
layershifter:
```suggestion
Video.create = createShorthandFactory({ Component: Video, mappedProp: 'src' })
```

This will allow us to use `<Embed video='http://youtube/1.mpg' />`

docs/src/examples/components/Embed/Types/EmbedExample.shorthand.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react'
+import { Embed, Flex, Text } from '@stardust-ui/react'
+
+const EmbedExample = () => (
+  <Flex column>
+    <Embed
+      placeholder="https://raw.githubusercontent.com/bower-media-samples/big-buck-bunny-480p-5s/master/poster.jpg"
+      src="https://raw.githubusercontent.com/bower-media-samples/big-buck-bunny-1080p-5s/master/video.mp4"
```
layershifter:
```suggestion
      video="https://raw.githubusercontent.com/bower-media-samples/big-buck-bunny-1080p-5s/master/video.mp4"
```

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+  /** Image source URL for when video isn't playing */
+  placeholder?: string
+
+  /** whether the gif should be playing */
+  playing?: boolean
+
+  iframe?: ShorthandValue
+
+  video?: ShorthandValue
```
layershifter:
```suggestion
  /** Shorthand for the video. */
  video?: ShorthandValue
```

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -0,0 +1,145 @@
...
+   * */
+  accessibility?: Accessibility
+
+  /** Video source URL. */
+  src?: string
+
+  /** Image source URL for when video isn't playing */
+  placeholder?: string
+
+  /** whether the gif should be playing */
```
layershifter:
```suggestion
  /** whether the gif should be playing. */
```
_nit_

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -0,0 +1,145 @@
...
+  /**
+   * Accessibility behavior if overridden by the user.
+   * @default embedBehavior
+   * */
+  accessibility?: Accessibility
+
+  /** Video source URL. */
+  src?: string
+
+  /** Image source URL for when video isn't playing */
```
layershifter:
```suggestion
  /** Image source URL for when video isn't playing. */
```
nit

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+export interface EmbedProps extends UIComponentProps {
+  /**
+   * Accessibility behavior if overridden by the user.
+   * @default embedBehavior
+   * */
+  accessibility?: Accessibility
+
+  /** Video source URL. */
+  src?: string
```
layershifter: We can remove `src` as we will have correctly working shorthand, https://github.com/stardust-ui/react/pull/1108/files#r272260321

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -0,0 +1,145 @@
...
+        {...unhandledProps}
+        className={classes.root}
+        {...accessibility.attributes.root}
+        {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.root, unhandledProps)}
+        onClick={this.handleClick}
+      >
+        {this.state.isPlaying
+          ? Video.create(video || ElementType, {
+              defaultProps: {
+                src,
```
layershifter:
```suggestion
```

https://github.com/stardust-ui/react/pull/1108/files#r272260321

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -0,0 +1,145 @@
...
+    return (
+      <ElementType
+        {...unhandledProps}
+        className={classes.root}
+        {...accessibility.attributes.root}
+        {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.root, unhandledProps)}
+        onClick={this.handleClick}
+      >
+        {this.state.isPlaying
+          ? Video.create(video || ElementType, {
```
layershifter:
```suggestion
          ? Video.create(video, {
```

https://github.com/stardust-ui/react/pull/1108/files#r272260321

packages/react/src/components/Video/VideoGif.tsx
stuartlong: Change to auto controlled

packages/react/src/themes/teams/components/Icon/svg/icons/pause.tsx
layershifter: Icons were added in #1189.

packages/react/src/themes/teams/components/Video/videoGifStyles.ts
layershifter: Solved 👍

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+  /** Image source URL for when video isn't playing. */
+  placeholder?: ShorthandValue<ImageProps>
+
+  /** Shorthand for an embedded video. */
+  video?: ShorthandValue<VideoProps>
+}
+
+export interface EmbedState {
+  active: boolean
```
mnajdova, line 67: Let's add `onActiveChanged` prop for this component, and invoke it when we change the active prop, as it will be required when we want to use it in controlled mode (see for example the `onOpenChanged` in the `Popup` component)

packages/react/src/components/Video/Video.tsx
```diff
@@ -0,0 +1,106 @@
...
+    controls: PropTypes.bool,
+    loop: PropTypes.bool,
+    muted: PropTypes.bool,
+    poster: PropTypes.string,
+    src: PropTypes.string,
+  }
+
+  static defaultProps = {
+    as: 'video',
+    accessibility: defaultBehavior as Accessibility,
```
mnajdova: No need for this, as the `defaultBehavior` will be applied if nothing is provided here.

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+  getInitialAutoControlledState(): EmbedState {
+    return { active: false }
+  }
+
+  handleClick = e => {
+    e.stopPropagation()
+    e.preventDefault()
+
+    this.trySetState({ active: !this.state.active })
```
mnajdova, line 124: This is where we would invoke the `onActiveChanged` method.

packages/react/src/themes/teams/index.ts
```diff
@@ -47,6 +47,8 @@ const icons: ThemeIcons = {
   'stardust-arrow-up': themeIcons['triangle-up'],
   'stardust-arrow-down': themeIcons['triangle-down'],
   'stardust-arrow-end': themeIcons['triangle-right'],
+  'stardust-pause': themeIcons['pause'],
```
mnajdova, line 50: 👍

packages/react/src/themes/teams/components/Embed/embedVariables.ts
```diff
@@ -0,0 +1,13 @@
+export interface EmbedVariables {
```
mnajdova, line 1: Please, for consistency, add these new styles and variables in `themes/types.ts` in:
`ThemeComponentStylesInput`
`ThemeComponentStylesPrepared`
`ThemeComponentVariablesInput`
`ThemeComponentVariablesPrepared`

packages/react/src/themes/teams/components/Embed/embedVariables.ts
```diff
@@ -0,0 +1,13 @@
...
+  controlColor: string
+  width: number
```
mnajdova: Not sure if this is the right typing, as I would expect the user to provide something like: `10px` or `1rem` which is not a number.

packages/react/src/themes/teams/components/Video/videoVariables.ts
```diff
@@ -0,0 +1,9 @@
+export interface VideoVariables {
+  width: number
```
mnajdova: See prev comment for the typings of the `width` and `height` variables.

bmdalex (approved):
great work, no more comments on top of the ones from existing reviewers, pls address them before merge

packages/react/src/themes/teams/components/Embed/embedVariables.ts
layershifter: Great catch, updated 👍

packages/react/src/themes/teams/components/Embed/embedVariables.ts
layershifter: Also added for `Tree` component 👍

packages/react/src/components/Video/Video.tsx
layershifter: We have currently an assertion for this, going to address it in a separate PR.

packages/react/src/components/Embed/Embed.tsx
layershifter: Implemented as `onActiveChanged`, however I want address an inconsistency there separately because `Popup` has `onOpenChange` (without _d_ on end).

mnajdova: I would expect the _d_ in the end, but agree, let's address in a separate PR, to avoid unnecessary breaking changes.

packages/react/src/themes/teams/components/Embed/embedStyles.ts
```diff
@@ -0,0 +1,48 @@
...
+      ':hover': {
+        [`& .${Embed.slotClassNames.control}`]: {
+          opacity: 1,
+          zIndex: 1,
+        },
+      },
+    }
+  },
+  control: ({ props: p, variables: v }): ICSSInJSStyle => ({
+    background: `0 no-repeat rgba(0,0,0,.25)`,
```
mnajdova: Please let's use colors values in the variables, so that we may override those in different themes.

mnajdova (approved):
One last comment. LGTM 👍

Issue #896, opened 2019-02-13
Icon|Dropdown: components use an icon as indicator


layershifter:
## Feature Request

#721 introduced the `Indicator` component, it resolved issues with arrows. However, we have issues with `clearIndicator` on `Dropdown` (#885) and `Input`, it relies on the `close` icon that can be undefined in the theme.

### Proposed solution
1. Add base icons, may be even inside the indicator component
2. Find the correct unicode char for `cross`/`close` action and use it in `Indicator` component

PR #1114, opened 2019-03-27
feat(Indicator): add close indicator

mnajdova:
Fixes https://github.com/stardust-ui/react/issues/896

### DIFF
--- a/packages/react/src/components/Indicator/Indicator.tsx
+++ b/packages/react/src/components/Indicator/Indicator.tsx
@@ -23,6 +23,9 @@ export interface IndicatorProps extends UIComponentProps {
   /** The indicator can point towards different directions. */
   direction?: 'start' | 'end' | 'top' | 'bottom'
 
+  /** The indicator can point towards different directions. */
+  code?: 'arrow' | 'close'
+
   /** The indicator can show specific icon if provided. */
   icon?: ShorthandValue
 }
@@ -38,10 +41,18 @@ class Indicator extends UIComponent<ReactProps<IndicatorProps>, any> {
   static className = 'ui-indicator'
 
   static directionMap = {
-    end: { unicode: '\u25B8', rotation: -90 },
-    start: { unicode: '\u25C2', rotation: 90 },
-    top: { unicode: '\u25B4', rotation: 180 },
-    bottom: { unicode: '\u25BE', rotation: 0 },
+    arrow: {
+      end: { unicode: '\u25B8', rotation: -90 },
+      start: { unicode: '\u25C2', rotation: 90 },
+      top: { unicode: '\u25B4', rotation: 180 },
+      bottom: { unicode: '\u25BE', rotation: 0 },
+    },
+    close: {
+      end: { unicode: '\u2715', rotation: 0 },
+      start: { unicode: '\u2715', rotation: 0 },
+      top: { unicode: '\u2715', rotation: 0 },
+      bottom: { unicode: '\u2715', rotation: 0 },
+    },
   }
 
   static propTypes = {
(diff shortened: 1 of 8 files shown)


packages/react/src/components/Indicator/Indicator.tsx
```diff
@@ -23,6 +23,9 @@ export interface IndicatorProps extends UIComponentProps {
   /** The indicator can point towards different directions. */
   direction?: 'start' | 'end' | 'top' | 'bottom'
 
+  /** The indicator can point towards different directions. */
+  code?: 'arrow' | 'close'
```
layershifter, line 27: What if we will rename `direction` to `variation`/`symbol` and include there `close`?

```diff
  /** The indicator can point towards different directions. */
- direction?: 'start' | 'end' | 'top' | 'bottom'
+  variation?: 'start' | 'end' | 'top' | 'bottom' | 'close'

-   /** The indicator can point towards different directions. */
-  code?: 'arrow' | 'close'
```

mnajdova: That was my first thought but the direction is more for how the indicator/icon should be rotated. For example if you have the arrow icon, it can be pointed to different directions... In that sense what will the close mean?

mnajdova: Plus in rtl the directions for start and end are swapped...

layershifter: In the same the code below doens't have any sense:
```jsx
<Indicator code='close' direction='start' />
```

layershifter: An actually, it's more `clear`/`cross` than `close` because it doesn't _closes_ anything in context of usage. May be it will be more obvious if we will prefix existing values with `arrow-` 🤔 

```jsx
<Indicator symbol='cross' />
<Indicator symbol='arrow-start' />
```

layershifter: Also, I expect that we will need more symbols like `minus` for example...

mnajdova: I am just a bit afraid, that the `start | end | top | bottom` are one set of values, which are related to one another ('start' is replaced with 'end'), and this may be reused with other indicators as well (not sure how these character usages will grow in the future). Now we can ignore or augment the characters if different direction is used. At this moment, as the indicator is pretty simple, I am okay with either implementation, I am just thinking about future changes that may be introduced.

mnajdova: @miroslavstastny @kuzhelov @Bugaa92 let's here another opinion on this one as well.

miroslavstastny: I am sorry but I don't like the solution in general.
1. According to doc, `An indicator is suggesting additional content next to the element it is used in.` That is not true here.
2. **Theme** should be able to define its own close icon. How can that be achieved with this approach?

We should decide if this is a critical thing. And if it is not, I would rather discuss a more general approach on how theme can define content.

mnajdova: Then back to the other proposal we had (this component was introduced only because we don't have the set of icon which will be available in all themes). Can we then define in the typings of the themes, that the arrows and close icons will always be available and just use icon names directly? If we are okay with that, then we won't even need this component at all. @kuzhelov @miroslavstastny @layershifter

layershifter: Or we can define icons/symbols directly in styles of this component.

mnajdova: I agree with @miroslavstastny that this component is just covering some missing parts from our API. If we have the icons we don't need any of that.. I would rather have the icons being required as those are required by the API of the components we have...

bmdalex: +1 for @miroslavstastny 's comment and looks like we're going towards the right solution

It'd be good to define some sort of artifacts per themes in <key,value> pairs that can be reused in components styles; same as fontAwesome does for icons; not sure if we need more than icons at this point. It'd also be good to let users edit these

mnajdova: Created alternative PR: https://github.com/stardust-ui/react/pull/1120 please take a look

mnajdova:
Closing this in favor of https://github.com/stardust-ui/react/pull/1120

PR #1115, opened 2019-03-27
fix(RTL): align text to left

### LINKED ISSUE #776: RTL is not consistent between List.Item and ItemLayout
For List.Item, when using within a div with `dir="rtl"`, the header and content are right-aligned. However, ItemLayout aligns the header and content based on the language being used. The result is that displaying content from an RTL language while the app is in LTR (and vice-versa) looks incorrect with ItemLayout, but correct with List.Item:

image

Link to codesandbox:
https://codesandbox.io/s/13p3zx4rn7

miroslavstastny:
Fixes #776.

Stardust sets `dir="auto"` on all text content. To correctly handle bidirectional text.
The problem is that the attribute also changes default `text-align` to right for RTL text (which breaks LTR layout) and to left for LTR text (which breaks RTL layout).

This fix sets `text-align` on `ProviderBox` and `PopupContent` explicitly to avoid the problem.

- [x] Changelog


bmdalex (approved):

packages/react/src/themes/teams/components/Popup/popupContentStyles.ts
```diff
@@ -9,6 +9,7 @@ const popupContentStyles: ComponentSlotStylesInput<PopupContentProps, PopupConte
 
     return {
       display: 'block',
+      textAlign: 'left',
```
kuzhelov, line 12: as we are expecting this to be applied to _any_ popup content (i.e. the one that is provided not as an object prop, but as regular react element), like in the following case:

```jsx
<Popup ... content={<div>...</div>}>
```

then we need to apply this to `Popup`'s `popup` slot in general

PR #1118, opened 2019-03-27
fix(dropdown): fix tab selection and blur reset

### LINKED ISSUE #1101: Dropdown search single selection: selected option is cleared after TAB
## Bug Report

https://codesandbox.io/s/v3k81l9qz7

### Steps
Start typing a name
Select one of the options using arrow keys
Press TAB

### Expected Result
Value is selected

### Actual Result
No value is selected

### Version
0.24.0

### Testcase
https://codesandbox.io/s/v3k81l9qz7

silviuaavram:
This PR aims to fix 2 issues:

1. On multiple selection and search, using Tab (and Shift+Tab) to select an item will keep the focus on the dropdown, unless the `tabInMultipleSelection` is set to true.
2. On single search, will prevent default Downshift reset behavior when there is a blur on input text.

These 2 issues should also be addressed in Downshift but we need them now, so will patch them now like this and will follow-up in Downshift later. Downshift should make sure that selection on blur works (now it just resets on blur). Also I am planning to add multiple selection to Downshift in the future.

Please provide input about implementation and naming.

Should close https://github.com/stardust-ui/react/issues/1101.

### DIFF
--- a/packages/react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react/src/components/Dropdown/Dropdown.tsx
@@ -762,9 +762,10 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
         case keyboardKey.Tab:
           if (!_.isNil(highlightedIndex)) {
             selectItemAtIndex(highlightedIndex)
-          }
-          if (this.props.multiple && !this.props.tabInMultipleSelection) {
-            e.preventDefault()
+            // Keep focus here if condition applies.
+            if (this.props.multiple && !this.props.tabInMultipleSelection) {
+              e.preventDefault()
+            }
           }
           break
         case keyboardKey.ArrowLeft:
@@ -901,10 +902,11 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
         if (_.isNil(highlightedIndex)) {
           toggleMenu()
         } else {
+          // Keep focus here in this case.
+          if (this.props.multiple && !this.props.tabInMultipleSelection) {
+            e.preventDefault()
+          }
           selectItemAtIndex(highlightedIndex)
-        }
-        if (this.props.multiple && !this.props.tabInMultipleSelection) {
-          e.preventDefault()
         }
         return
       case keyboardKey.Escape:
--- a/packages/react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react/src/components/Dropdown/Dropdown.tsx
@@ -128,6 +128,9 @@ export interface DropdownProps extends UIComponentProps<DropdownProps, DropdownS
   /** A message to be displayed in the list when dropdown is loading. */
   loadingMessage?: ShorthandValue
 
+  /** When selecting an element with Tab, focus stays on the dropdown by default. If true, the focus will jump to next/previous element in DOM. */
+  moveFocusOnTab?: boolean
+
   /** A dropdown can perform a multiple selection. */
   multiple?: boolean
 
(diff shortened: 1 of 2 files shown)


mnajdova (approved):
The changes looks good, but I would suggest adding some unit tests for this. Also, please update the changelog as well!

jurokapsiar (approved):

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -185,6 +185,9 @@ export interface DropdownProps extends UIComponentProps<DropdownProps, DropdownS
   /** Sets search query value (controlled mode). */
   searchQuery?: string
 
+  /** When selecting an element with Tab, focus stays on the dropdown by default. If true, the focus will jump to next/previous element in DOM. */
+  tabInMultipleSelection?: boolean
```
bmdalex: can we add an example to outline the difference in behavior with/without this prop?
can be done in follow-up PR

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -185,6 +185,9 @@ export interface DropdownProps extends UIComponentProps<DropdownProps, DropdownS
   /** Sets search query value (controlled mode). */
   searchQuery?: string
 
+  /** When selecting an element with Tab, focus stays on the dropdown by default. If true, the focus will jump to next/previous element in DOM. */
```
bmdalex: >  /** When selecting an element with Tab, focus stays on the dropdown by default.

That's not true for the single selection `Dropdown`, the focus goes on the next element.

I know the intent of the PR is to introduce this behavior for multiple selection `Dropdown` but as a user I expect consistent behavior so I find it confusing that when I select items with `Tab` the default focus behavior for:
- single selection `Dropdown` is to go to next tabbable element
- on multiple selection `Dropdown` is to stay on current element

What I'm proposing is to decide on:

### 1. expected default behavior:
I'd expect pressing `Tab` / `Shift+Tab` would actually move the focus to next / previous tabbable element. That would imply reverting logic and calling our prop something like `trapFocusOnTabPress`. For me that makes more sense.

### 2. consistency for all `Dropdown` variations
If we decide that the default behavior is to keep focus on `Tab` / `Shift+Tab` press, that's fine. However, the behavior should be consistent in single/multiple selection `Dropdown`.

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: `tabInMultipleSelection` is not very descriptive; what about
- `moveFocusOnTab`
- `moveFocusOnTabKey`
- `moveFocusOnTabPress`
- `moveFocusOnTabKeyPress`

Or use other verbs instead of `move`, like `release` or `yield`.

What do you think?
@mnajdova @silviuavram

bmdalex (dismissed):

bmdalex (commented):
forgot this is something that needs to get in so I'm not blocking this PR, but pls consider the comments, especially the ones concerning changing the name of the props as we want to avoid having breaking changes

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: discussed with @silviuavram and @mnajdova and the current implementation is correct

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: will add as followup

PR #1120, opened 2019-03-27
chore(Indicator): use predefined icon set

### LINKED ISSUE #896: Icon|Dropdown: components use an icon as indicator

mnajdova:
Fixes https://github.com/stardust-ui/react/issues/896 Added default icons set for the icons used in the Stardust's component:
- stardust-close (used in Input and Dropdown)
- stardust-arrow-down (used in AccordionTitle and Dropdown)
- stardust-arrow-up (used in Dropdown)
- stardust-arrow-end (used in AccordionTitle)

It's considered theme's responsibility to implement the rtl behavior of the icons. These icons are added in the base theme as well, by using font awesome icons.

### TODO
- [x] update changelog

### BREAKING CHANGES:
Replace `Indicator` component with `Icon` and replace the `direction` prop with `name` in the following manner:
- `end` => `stardust-arrow-end`
- `bottom` => `stardust-arrow-down`
- `top` => `stardust-arrow-up`


packages/react/src/themes/teams/withProcessedIcons.ts
```diff
@@ -12,21 +12,33 @@ const declareSvg = (svgIcon: SvgIconSpec, exportedAs?: string): ThemeProcessedIc
   exportedAs,
 })
 
-const processedIcons: ThemeIcons = Object.keys(svgIconsAndStyles as {
+const getIcon = iconAndMaybeStyles => {
```
kuzhelov: given that it was decided to extract this method, it would be beneficial if we will specify types for the args and return value (actually, the return value type may be the most beneficial piece)

packages/react/src/themes/teams/withProcessedIcons.ts
```diff
@@ -12,21 +12,33 @@ const declareSvg = (svgIcon: SvgIconSpec, exportedAs?: string): ThemeProcessedIc
...
+  return (iconAndMaybeStyles as any).styles
+    ? (iconAndMaybeStyles as SvgIconSpecWithStyles).icon
+    : (iconAndMaybeStyles as SvgIconSpec)
+}
+
+const processedIcons: { [key: string]: ThemeIconSpec } = Object.keys(svgIconsAndStyles as {
```
kuzhelov: you may use `ObjectOf<ThemeIconSpec>`

packages/react/src/themes/teams/withProcessedIcons.ts
```diff
@@ -12,21 +12,33 @@ const declareSvg = (svgIcon: SvgIconSpec, exportedAs?: string): ThemeProcessedIc
...
   [iconName: string]: TeamsProcessedSvgIconSpec
-}).reduce<ThemeIcons>((accIcons, iconName) => {
+}).reduce<{ [key: string]: ThemeIconSpec }>((accIcons, iconName) => {
```
kuzhelov: we can use `ThemeIcons` at both places where we have `{ [key: string]: ThemeIconSpec }` now - there is no need to use dictionary type. I would suggest to do one of the following:
- use `ThemeIcons` type here and for `processedIcons` above
- or remove `ThemeIcons` and use `ObjectOf<ThemeIconSpec>` at both places, if you think that this would be more readable

What do you think?

packages/react/src/themes/teams/index.ts
```diff
@@ -21,22 +21,34 @@ const declareSvg = (svgIcon: SvgIconSpec): ThemeIconSpec => ({
 
 const declareFontBased = (fontIcon: FontIconSpec): ThemeIconSpec => ({ icon: fontIcon })
 
-const icons: ThemeIcons = Object.keys(svgIconsAndStyles as {
+const getIcon = iconAndMaybeStyles => {
```
kuzhelov: this logic seems to be duplicated - lets extract it and consume from the single place (`withProcessedIcons` module)

packages/react/src/themes/types.ts
```diff
@@ -505,4 +505,10 @@ export type ThemeIconSpec = {
...
 }
 
-export type ThemeIcons = { [iconName: string]: ThemeIconSpec }
+export type ThemeIcons = {
+  'stardust-close'?: ThemeIconSpec
+  'stardust-arrow-end'?: ThemeIconSpec
+  'stardust-arrow-up'?: ThemeIconSpec
+  'stardust-arrow-down'?: ThemeIconSpec
+  [iconName: string]: ThemeIconSpec
+}
```
layershifter, line 516:
```
export type RequiredIconNames =  'stardust-close' | 'stardust-arrow-end' | 'stardust-arrow-up' | 'stardust-arrow-down' 

export type ThemeIcons = Partial<Record<RequiredIconNames, ThemeIconSpec>> & {
  [iconName: string]: ThemeIconSpec
}
```

packages/react/src/themes/base/components/Icon/fontAwesomeIconStyles.ts
```diff
@@ -0,0 +1,21 @@
...
+// https://github.com/Semantic-Org/Semantic-UI-CSS/blob/master/components/icon.css
+// Corrections were made to duplicate icon names and incorrectly mapped alternates
+const fontAwesomeIcons: ThemeIcons = {
+  'stardust-close': normal('f00d'),
+  'stardust-arrow-down': normal('f0d7'),
+  'stardust-arrow-end': normal('f0da'),
+  'stardust-arrow-up': normal('f0d8'),
+}
+
+const emptyIcon = { icon: { content: '?', fontFamily: '' } }
```
layershifter: Let's get rid of `FontAwesome` there and use our existing unicode characters

packages/react/src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -130,6 +130,10 @@ const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
       ...(!rtl && {
         transform: `rotate(${rotate}deg)`,
       }),
+      // override base theme default rtl behavior
+      ...(rtl && {
```
kuzhelov, line 110: intent of this and the previous object expansion expressions could be provided in a simpler way:

```jsx
transfrom: rtl ? 'unset' : `rotate(${rotate}deg)`
```

packages/react/src/themes/base/components/Icon/iconStyles.ts
```diff
@@ -0,0 +1,136 @@
...
+
+      ...getXSpacingStyles(xSpacing, v.horizontalSpace),
+
+      ...(circular && { ...getPaddedStyle(), borderRadius: '50%' }),
+
+      ...((bordered || v.borderColor) &&
+        getBorderedStyles(v.borderColor || getIconColor(color, v))),
+
+      ...(rtl && {
+        transform: `scaleX(-1) rotate(${-1 * rotate}deg)`,
```
kuzhelov: just to be sure - seems like `rotate` prop's value won't have any effect in case if `rtl` mode is disabled, will it? miss the following style lines from the original file:

```jsx
...(!rtl && {
}),
```

packages/react/src/themes/teams/withProcessedIcons.ts
mnajdova: Yep totally agree, at the first implementation the new names were required that is why I was changing these typings, but yes as they are optional now, I can use the same type. Good catch!

packages/react/src/themes/teams/withProcessedIcons.ts
mnajdova: Will use `ThemeIcons` here as well

packages/react/src/themes/base/components/Icon/iconStyles.ts
mnajdova: yep, will add it

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -225,5 +222,9 @@ class Menu extends AutoControlledComponent<ReactProps<MenuProps>, MenuState> {
 }
 
 Menu.create = createShorthandFactory({ Component: Menu, mappedArrayProp: 'items' })
+Menu.slotClassNames = {
```
kuzhelov: lets follow the same approach that is introduced for other components (i.e. use the dedicated `static` field for now) - given that the problem with cyclic dependencies was solved in other way. We might want to reconsider the way `slotClassNames` are introduced later

packages/react/src/themes/teams/components/Icon/fontAwesomeIconStyles.ts
```diff
@@ -11,7 +11,7 @@ const brand = (content: string) => fontIcon(content, '"Font Awesome 5 Brands"')
 // Originally generated from:
 // https://github.com/Semantic-Org/Semantic-UI-CSS/blob/master/components/icon.css
 // Corrections were made to duplicate icon names and incorrectly mapped alternates
-const fontAwesomeIcons: ThemeIcons = {
+const fontAwesomeIcons: { [key: string]: ThemeIconSpec } = {
```
kuzhelov: may we use `ThemeIcons` type here as well?

kuzhelov (approved):

packages/react/src/themes/base/components/Icon/iconStyles.ts
```diff
@@ -0,0 +1,122 @@
...
+
+const getIconSize = (size: SizeValue, sizeModifier: IconSizeModifier): number => {
+  if (!sizeModifier) {
+    return sizes[size]
+  }
+
+  const modifiedSizes = {
+    large: {
+      x: 24,
+      xx: 28,
```
layershifter: I'm not sure that this is part of base theme...

packages/react/src/themes/base/components/Icon/iconStyles.ts
```diff
@@ -0,0 +1,122 @@
...
+    },
+  }
+
+  return modifiedSizes[size] && modifiedSizes[size][sizeModifier]
+}
+
+const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
+  root: ({
+    props: { name, size, bordered, circular, xSpacing, rotate },
```
layershifter: As we use `p.` instead?

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -225,5 +222,9 @@ class Menu extends AutoControlledComponent<ReactProps<MenuProps>, MenuState> {
...
+  divider: `${Menu.className}__divider`,
+  item: `${Menu.className}__item`,
```
bmdalex: fyi @layershifter @sophieH29 this will fix your issue

packages/react/src/themes/base/components/Icon/iconStyles.ts
mnajdova: Moved that back to the teams theme. The base theme now is always creating icon with fontSize of 16px

packages/react/src/themes/teams/components/Icon/fontAwesomeIconStyles.ts
mnajdova: Yep, changed

bmdalex (approved):
changelog and merge 👍

PR #1121, opened 2019-03-27
fix(Popup): styles for dark and HC Teams themes

### LINKED ISSUE #1053: Popup content doesn't honor theme styling
## Bug Report

Popup content doesn't honor theme styling

### Steps
Create a popup using a theme other than teams, for instance teamsDark

### Expected Result
Popup content should get the theme's styling

### Actual Result
Popup content doesn't get styled
image

image

### Version
0.23.0

### Testcase

kuzhelov:
## TODO
- [x] update changelog

------

Fixes #1053.

Provided changes address the problem of Popup styling for Dark and High Contrast Teams themes. In addition, variables that control `Popup`'s content styling were introduced, so that now _any custom_ popup content will be properly displayed with the right set of foreground and background colors.

#### Dark
image

#### High Contrast
image

#### Was (barely noticeable black text on dark background)
image

#### Now
image


packages/react/src/themes/teams/components/Popup/popupStyles.ts
```diff
@@ -8,6 +8,8 @@ const popupStyles: ComponentSlotStylesInput<PopupProps, PopupVariables> = {
   popup: ({ variables }): ICSSInJSStyle => ({
     zIndex: variables.zIndex,
     position: 'absolute',
+    color: variables.contentColor,
+    background: variables.contentBackgroundColor,
```
mnajdova, line 13: Is there a reason why we moved the background style from the `PopupContent` to the `Popup`? I'd prefer to see them in it's own component styles, rather then in the popup, especially because they are not anyhow dependent on some props from the `Popup` component.

packages/react/src/themes/teams/components/Popup/popupVariables.ts
```diff
@@ -2,6 +2,12 @@ export interface PopupVariables {
   [key: string]: string | number
 
   zIndex: number
+  contentColor: string
+  contentBackgroundColor: string
 }
 
-export default (siteVars: any): PopupVariables => ({ zIndex: 1000 })
+export default (siteVars: any): PopupVariables => ({
+  zIndex: 1000,
```
bmdalex, line 10: how did we come up with this value?

packages/react/src/themes/teams/components/Popup/popupStyles.ts
bmdalex, line 13: since now we have relevant styles for Dark and HC themes, can we add a few screenshot tests for these themes for  relevant Popup examples?

It'd be great to add some steps too, but that can be in different PR, something like:
- open a Popup by click and take snapshot
- open a Popup by key and take snapshot
- repeat these steps for different variations: inline Popups, Popups with positioning, etc

kuzhelov: yes - the reason is that we want to see the background and foreground colors being applied to the popup's content _irregardless of what is rendered as popup content_. Previously we have introduced proper styles only for the cases when `PopupContent` was rendered at the `content` slot - however, in case if user would render plain content there, there won't be any background/foreground styles applied at all. Thus, the following code

```jsx
<Popup ... content={<div>Hello from popup content!</div>} />
``` 

would result in the presentation that was mentioned in the PR's description:

image

However, what client really wants here is proper basic styles applied even if plain content is provided to the slot:

image

That's why the variables were moved - note that with that we are not limiting client in overriding styles for the content in case if necessary (so, if `content` provides its own styles for `background` and `color`, then those will be used)

bmdalex (approved):
great job, looks awesome, just couple of comments

miroslavstastny (approved):
We should find a way how to share styles between `ProviderBox` and `Popup`.
After this change, whatever we add to `ProviderBox`, we need to add it to `Popup` again.
Moreover, whatever a consumer overrides in ProviderBox, (s)he must override it again for each `Popup`.
See #1115 as an example of this.

packages/react/src/themes/teams/components/Popup/popupVariables.ts
kuzhelov: not sure, agreed to use the one that has worked for us back then :) Low numbers might put Popup design at risk, as there might be some client's components that rely on some elements shown on top of the others, and `z-index: 1` + `relative` positioning is always the thing that does this trick. So, admittedly, for the sake of being on a _safer_ side this value was introduced

packages/react/src/themes/teams/components/Popup/popupStyles.ts
kuzhelov: sure, will take as a follow-up

PR #1123, opened 2019-03-28
feat(Icon): export `ban` and `exclamation-circle` in Teams theme

kuzhelov:
## TODO
- [x] update changelog


packages/react/src/themes/teams/components/Icon/svg/ProcessedIcons/icons-block.tsx
```diff
@@ -17,4 +17,5 @@ export default {
     </svg>
   ),
   styles: {},
+  exportedAs: 'block',
```
codepretty: Should we export as "ban", like font awesome?
image

kuzhelov: will do that, thanks!

codepretty (approved):
Looks good - with one more suggestion... consider exporting the input-invalid to exclamation-circle as well.

PR #1125, opened 2019-03-28
feat(Menu): Make Submenu open state controlled

sophieH29:
This PR aims make submenu open  state controlled, thus to bring the flexibility of using submenu and rewrite default behavior.


packages/react/src/lib/AutoControlledComponent.tsx
```diff
@@ -187,8 +187,9 @@ export default class AutoControlledComponent<P = {}, S = {}> extends UIComponent
    * Second argument is a state object that is always passed to setState.
    * @param {object} maybeState State that corresponds to controlled props.
    * @param {object} [state] Actual state, useful when you also need to setState.
+   * @param {object} onStateChanged Callback which is called after setState applied.
    */
-  trySetState = (maybeState, state?) => {
+  trySetState = (maybeState, state?, onStateChanged?: () => void) => {
```
layershifter: https://github.com/Semantic-Org/Semantic-UI-React/pull/3452#discussion_r264972028

I support there Levi, I think that we can also remove `state` param, too. And match React's signature:

```suggestion
  trySetState = (state: Partial<State>, callback?: () => void) => {
```

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -363,12 +372,20 @@ class MenuItem extends AutoControlledComponent<ReactProps<MenuItemProps>, MenuIt
...
       e.stopPropagation()
       e.preventDefault()
     }
   }
+
+  private trySetMenuOpen(newValue: boolean, eventArgs: any, onStateChanged?: any) {
+    this.trySetState({ menuOpen: newValue }, null, onStateChanged)
+    _.invoke(this.props, 'onMenuOpenChange', eventArgs, {
+      ...this.props,
+      ...{ menuOpen: newValue },
```
layershifter:
```suggestion
      menuOpen: newValue,
```

packages/react/src/lib/AutoControlledComponent.tsx
sophieH29: I like that, actually didn't find any usage of 2nd param.
Would make this change if nobody minds

packages/react/src/lib/AutoControlledComponent.tsx
```diff
@@ -187,13 +187,14 @@ export default class AutoControlledComponent<P = {}, S = {}> extends UIComponent
    * Second argument is a state object that is always passed to setState.
    * @param {object} maybeState State that corresponds to controlled props.
    * @param {object} [state] Actual state, useful when you also need to setState.
+   * @param {object} callback Callback which is called after setState applied.
    */
-  trySetState = (maybeState, state?) => {
+  trySetState = (state: Partial<S>, callback?: () => void) => {
     const { autoControlledProps } = this.constructor as any
     if (process.env.NODE_ENV !== 'production') {
```
layershifter: This check should be removed, otherwise we will get incorrect warnings

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -362,12 +372,25 @@ class MenuItem extends AutoControlledComponent<ReactProps<MenuItemProps>, MenuIt
...
+    // The reason why post-effect is not passed as callback to trySetState method
+    // is that in 'controlled' mode the post-effect is applied before final re-rendering
+    // which cause a broken behavior: for e.g. when it is needed to focus submenu trigger on ESC.
+    // TODO: all DOM post-effects should be applied at componentDidMount & componentDidUpdated stages.
+    onStateChanged && onStateChanged()
+    _.invoke(this.props, 'onMenuOpenChange', eventArgs, {
+      ...this.props,
+      menuOpen: newValue,
+    })
+  }
```
layershifter, line 392: I don't like this method, but I don't have a better option now...

docs/src/examples/components/Menu/Usages/MenuExampleWithSubmenuControlled.shorthand.tsx
```diff
@@ -0,0 +1,50 @@
...
+            {
+              key: '2',
+              content: 'item2',
+              menu: [{ key: '1', content: 'item2.1' }, { key: '2', content: 'item2.2' }],
+            },
+            {
+              key: '3',
+              content: 'item3',
+              menu: [{ key: '1', content: 'item3.1' }, { key: '2', content: 'item3.2' }],
+            },
```
layershifter: I suggest to make it more compact:

```
menu: [
  { key: '1', content: 'item1' },
  { key: '2', content: 'item2' },
  { key: '3', content: 'item3' },
],
```

docs/src/examples/components/Menu/Usages/MenuExampleWithSubmenuControlled.shorthand.tsx
```diff
@@ -0,0 +1,50 @@
...
+    ]
+  }
+
+  render() {
+    return (
+      <>
+        <span>{`Editorials menu item requested to change its submenu open state to "${
+          this.state.editorialsMenuOpen
+        }".`}</span>
+        <Menu defaultActiveIndex={0} items={this.renderItems()} />
```
layershifter: This looks unreadable 🤔 

What about something like this:

```jsx
<Header as="h5" content="Current state:" />
<pre>{JSON.stringify(this.state, null, 2)}</pre>
<Divider />
<Menu defaultActiveIndex={0} items={this.renderItems()} />
  
```

docs/src/examples/components/Menu/Usages/MenuExampleWithSubmenuControlled.shorthand.tsx
layershifter: As it will become more compact, do we need `renderItems()` at all?

sophieH29: I would leave `renderItems()`, for better readability

docs/src/examples/components/Popup/Types/PopupExample.shorthand.tsx
```diff
@@ -1,6 +1,19 @@
...
+    content={
+      <Dialog
+        cancelButton="Cancel"
+        confirmButton="Confirm"
+        header="Action confirmation"
+        trigger={<Button content="Open a dialog" />}
+      />
+    }
+  />
+)
```
layershifter: Should we revert these changes?

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -362,12 +372,25 @@ class MenuItem extends AutoControlledComponent<ReactProps<MenuItemProps>, MenuIt
...
     if (menu && !menuOpen) {
-      this.setState({ menuOpen: true })
+      this.trySetMenuOpen(true, e)
       _.invoke(this.props, 'onActiveChanged', e, { ...this.props, active: true })
       e.stopPropagation()
       e.preventDefault()
     }
   }
+
+  private trySetMenuOpen(newValue: boolean, eventArgs: any, onStateChanged?: any) {
```
layershifter:
```suggestion
  private trySetMenuOpen(e: React.SyntethicEvent, newValue: boolean, onStateChanged?: any) {
```

nit: Can we place `e` param on the first place as in other methods?

packages/react/src/lib/AutoControlledComponent.tsx
```diff
@@ -187,37 +187,17 @@ export default class AutoControlledComponent<P = {}, S = {}> extends UIComponent
...
-      if (autoControlledProps.indexOf(prop) === -1) return acc
-
       acc[prop] = maybeState[prop]
       return acc
     }, {})
 
-    if (state) newState = { ...newState, ...state }
-
-    if (Object.keys(newState).length > 0) this.setState(newState)
+    if (Object.keys(newState).length > 0) this.setState(newState, callback)
```
layershifter, line 201: Not sure, that we really need this condition, but let's leave it as is for now

layershifter (approved):

packages/react/src/components/Menu/MenuItem.tsx
sophieH29: fyi, the order is the same as in Popup for `trySetOpen` method. 
And the other thing, if you would want to change state without event, you would have to call
`trySetMenuOpen(null, newValue)`
instead of `trySetMenuOpen(newValue)`. I would leave an order as it is as I think the `newValue` param has biggest priority than `event` object

docs/src/examples/components/Popup/Types/PopupExample.shorthand.tsx
sophieH29: oops, yeah. Did some testing, great you noticed

PR #1126, opened 2019-03-28
fix(accessibility): add missing types to `role`

### LINKED ISSUE #630: Aria roles list is missing role 'alert'
When playing with accessibility property on stardust component I realised that role="alert" is not allowed by Stardust, but is valid role: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role Can it be added? Thx.

layershifter:
Fixes #630.


miroslavstastny (approved):
Live Region Roles are ok 👍 

But going through [wai-aria-1.1](https://www.w3.org/TR/wai-aria-1.1/#roles_categorization), I see several roles missing or being in a wrong group.
Missing roles: `alertdialog`, `cell`, `feed`, `figure`, `rowgroup`, `searchbox`, `table`, `term`.
@jurokapsiar

PR #1127, opened 2019-03-28
fix(layout): remove truncate* props from Layout and ItemLayout

### LINKED ISSUE #849: TruncateEnd property in Layout component is not truncating content with ellipsis.
### Steps
The truncateEnd property in Layout component is just truncating the text without showing ellipsis.

### Expected Result
The result that you expected.
image

### Actual Result
image

### MitigationSteps
The truncateEnd property should have display:"block" instead of "flex" to truncate with ellipsis.

image
image

bmdalex:
# Steps to from BREAKING CHANGES:

#### We recommend using `Flex` instead and follow our [Layout Components recommendations](https://stardust-ui.github.io/react/layout)

However, if you still want to use `Layout`/`ItemLayout`, the steps are:

1. Wrap the element that you need truncated with `truncated` `Text` component
2. Pass that element to `Layout`/`ItemLayout`'s slot
3. Add additional CSS as in examples below

### `Layout`
```jsx
<Layout
  // ...
  end={<Text truncated>My long text that I need truncated</Text>}
  endCSS={{ overflow: 'hidden' }}
/>
```
See complete code: **https://codesandbox.io/s/pm4vn7j0mm**

### `ItemLayout`
```jsx
<ItemLayout
  // ..
  endMedia={<Text truncated>Some very loooooooooooooooooooooooooooong text</Text>}
  endMediaCSS={{ width: '100px', overflow: 'hidden', display: 'grid' }}
/>
```
See complete code: **https://codesandbox.io/s/r1mm947xp**

#### Closes #849 

`truncate*` props (`truncateStart`, `truncateMain`, `truncateEnd`) failed to handle all truncation cases for `Layout` (see #849 ) and consequently `ItemLayout` so we decided to remove them from the API. (BREAKING CHANGE)


miroslavstastny (approved):

docs/src/examples/components/ItemLayout/index.tsx
```diff
@@ -1,14 +1,15 @@
...
+import { Alert } from '@stardust-ui/react/'
+
 import Types from './Types'
 import Content from './Content'
-import Variations from './Variations'
 import Rtl from './Rtl'
 
 const ItemLayoutExamples = () => (
   <div>
+    <Alert warning>This component is deprecated</Alert>
```
hughreeling, line 10: Let's try and be as helpful as we can.  Perhaps add link to Layout guide and suggest flex as alternative?

packages/react/src/components/ItemLayout/ItemLayout.tsx
```diff
@@ -62,12 +62,10 @@ export interface ItemLayoutProps extends UIComponentProps, ContentComponentProps
   contentMediaCSS?: ICSSInJSStyle
   /** Styled applied to the end media element of the rendered component. */
   endMediaCSS?: ICSSInJSStyle
-  truncateContent?: boolean
-  truncateHeader?: boolean
 }
 
 /**
- * The Item Layout handles layout styles for menu items, list items and other similar item templates.
+ * (DEPRECATED) The Item Layout handles layout styles for menu items, list items and other similar item templates.
```
hughreeling, line 68: Can we suggest alternative approach here?  Or point to Layout guide?

packages/react/src/components/ItemLayout/ItemLayout.tsx
```diff
@@ -101,8 +99,6 @@ class ItemLayout extends UIComponent<ReactProps<ItemLayoutProps>, any> {
     contentCSS: PropTypes.object,
     contentMediaCSS: PropTypes.object,
     endMediaCSS: PropTypes.object,
-    truncateContent: PropTypes.bool,
```
hughreeling, line 104: If we just take these out, won't they just get passed through to html and not do anything.  Should we leave as warning / compile error or something for a few releases before we actually remove?

packages/react/src/components/ItemLayout/ItemLayout.tsx
kuzhelov: it should be enough to reference alternative components to use - may suggest not to include any links to the descriptions, as those are subject to potential changes

docs/src/examples/components/Layout/index.tsx
```diff
@@ -1,9 +1,12 @@
 import * as React from 'react'
+import { Alert } from '@stardust-ui/react/'
+
 import Types from './Types'
 import Variations from './Variations'
 
 const LayoutExamples = () => (
   <div>
+    <Alert warning>This component is deprecated</Alert>
```
kuzhelov, line 9: nit-nit: lets end with the dot

kuzhelov (approved):

packages/react/src/components/ItemLayout/ItemLayout.tsx
bmdalex: This is already marked as a BREAKING CHANGE in `CHANGELOG.md` and it points to the description of the PR where we have a mitigation path. This should be enough for our clients to update without too much friction

docs/src/examples/components/ItemLayout/index.tsx
bmdalex: done in #1154

packages/react/src/components/ItemLayout/ItemLayout.tsx
bmdalex: done in #1154

PR #1132, opened 2019-03-29
fix(EventListener): do not rerender on `listener` prop change

### LINKED ISSUE #1131: EventListener: incorrect updates
## Bug Report

We should not rerender `EventListener` on update of the `listener` prop as it's not sensitive and will resubscribe it.

### Steps
1. Open https://codesandbox.io/s/jz9lvowll3
2. Click on background, see: `ListenerBox: CLICK!` & `CachedListenerBox: CLICK!`
3. Click on button

### Expected Result
Counter increased, `ListenerBox: CLICK!` & `CachedListenerBox: CLICK!` in console

### Actual Result
Counter increased, `CachedListenerBox: CLICK!` in console 💣

### Version
0.25.1

layershifter:
Fixes #1131.


packages/react-component-event-listener/src/EventListener.ts
```diff
@@ -1,16 +1,21 @@
...
-class EventListener extends React.PureComponent<EventListenerProps> {
+class EventListener extends React.Component<EventListenerProps> {
   static displayName = 'EventListener'
   static propTypes = listenerPropTypes
   static defaultProps = {
     capture: false,
   }
 
+  shouldComponentUpdate(nextProps: EventListenerProps) {
+    return shouldUpdateListener(this.props, nextProps)
```
layershifter, line 16: This sounds strange, but will work because React will call actual `this.props.listener`, added UTs confirm that. This behavior is covered there: https://overreacted.io/a-complete-guide-to-useeffect/#each-render-has-its-own-everything

It will be more clear when we will migrate this logic to hooks.

packages/react-component-event-listener/src/lib/shouldUpdateListener.ts
```diff
@@ -0,0 +1,8 @@
+import { EventListenerProps } from '../types'
+
+const shouldUpdateListener = (prevProps: EventListenerProps, nextProps: EventListenerProps) =>
```
kuzhelov, line 26: actually, the real intent we would like to express here is that "do not update if only `listener` was changed". Thus, lets make this intent explicit by providing a logic like the following instead:

```jsx
const shouldUpdateListener = (prevProps: EventListenerProps, nextProps: EventListenerProps) =>
  if (shallowEqual(
       { ...prevProps, listener: nextProps.listener },
       nextProps)) 
    {
       return false 
    } 
   ...
```

This will save us from any kind of maintainability problems that might lead to bugs that would be hard to discover - like, for example, the ones that will appear in case we would need to extend the props of this component

kuzhelov:
Lets merge the fix with couple of things being addressed:

https://github.com/stardust-ui/react/pull/1132#discussion_r270777548

Also, would suggest to provide more details in the issue - specifically, the ones that would explain what _really makes it to be an issue_. As an example, I might suggest to add the following insights:
- mention explicitly that event **won't be handled (i.e. will be 'missed')** by `EventListener` in case of `listener` prop change
- explain _why_ it happens
- explain the fix strategy (at least in this PR's description)

This will make it much easier to review the code, as well as to understand the issue itself - otherwise it has taken me a while to follow the same steps you've already made to investigate the issue, as well as, what is more important, to figure out _what_ makes it to be an issue.

Thank you!

packages/react-component-event-listener/src/EventListener.ts
kuzhelov: might refer to another section from the same blog that warns against using `shouldUpdate` methods https://overreacted.io/writing-resilient-components/#dont-stop-the-data-flow-in-optimizations :) For example, now, with these changes being made, we should refrain from passing `listener` value any further down (the fact that we are not rendering anything helps, but there are no explicit constrains on this component in these regards) - as it will be another source of problems.

At the same time, agree that Hooks policy around _when_ subscription happens (and, surely, around specifying update dependencies) should help in this case.

layershifter: Fully agree with you there, however should mention that `EventListener` doesn't render any children, so actually we don't stop there nothing.

And yep, hooks will be more obvious with their _when_ 👍

packages/react-component-event-listener/src/lib/shouldUpdateListener.ts
layershifter: I support this idea, however I should mention that in this case we will add a new dependency for short term... As we are going to update to React 16.8, may be we can go with existing implementation? _However, I am fully agree that your proposal is more understandable..._

layershifter:
> * mention explicitly that event **won't be handled (i.e. will be 'missed')** by `EventListener` in case of `listener` prop change

Actually, this was mentioned in the original issue, but I got your point about PRs description 👍

packages/react-component-event-listener/src/lib/shouldUpdateListener.ts
kuzhelov: agree with your point - lets discuss in person and decide? or, if you feel strongly about merging, please, feel free to skip this comment :) in that case we will be able to handle this in a follow-up PR. The sole reason I am not completely sure to skip that is because `shallowEqual` is, essentially, a 10-lines-long func with formatting:

```js
const shallowEqual = (first, second) => {
  // or whatever strategy we would have for handling falsy values - here is the most strict one
  if (first === undefined) { return second === undefined }
  if (first === null) { return second === null; }

  if (first === second) { return true; }

  if (Object.keys(first).length !== Object.keys(second).length) { return false; }

  const isAnyDifference = Object.keys(first).some(
    key => first[key] !== second[key]
  );

  return !isAnyDifference;
};
```

kuzhelov (approved):
unblocking the fix given we've synced the vision on those points I've expressed my worries about

packages/react-component-event-listener/test/EventListener-test.tsx
```diff
@@ -4,35 +4,55 @@ import * as React from 'react'
...
-    expect(onKeyDown).toHaveBeenCalledTimes(1)
-    expect(onKeyDown).toHaveBeenCalledWith(expect.objectContaining({ type: 'keydown' }))
-  })
+      expect(onKeyDown).toHaveBeenCalledTimes(1)
+      expect(onKeyDown).toHaveBeenCalledWith(expect.objectContaining({ type: 'keydown' }))
+    })
 
-  it('unsubscribes correctly', () => {
-    const onClick = jest.fn()
+    it('will not rerender on change', () => {
```
miroslavstastny: I don't understand this test.
I would expect that this test was red before the changes and now it is green and helps us avoid any regression.
But is is green even before the change so what's its purpose?

packages/react-component-event-listener/src/lib/shouldUpdateListener.ts
layershifter: Agreed to `shallowEqual()` function to this file and remove it fully when we will migrate to hooks

packages/react-component-event-listener/test/EventListener-test.tsx
layershifter: We agreed with @kuzhelov to remove this test as it doesn't save us from regressions

PR #1135, opened 2019-03-29
chore(Test): add filename filter

miroslavstastny:
Add possibility to filter tests executed in `yarn test:watch`:

```sh
yarn test:watch --testFilePattern=Menu --testNamePattern="passes arbitrary props"
```

or

```sh
yarn test:watch -F Menu -t "passes arbitrary props"
```


build/gulp/tasks/test-unit.ts
```diff
@@ -17,10 +17,14 @@ const jest = ({ watch = false } = {}) => cb => {
     argv.runInBand && '--runInBand',
     argv.maxWorkers && `--maxWorkers=${argv.maxWorkers}`,
     argv.detectLeaks && '--detectLeaks',
+    argv.testNamePattern && `--testNamePattern="${argv.testNamePattern}"`,
+    argv.testFilePattern && `${argv.testFilePattern}`, // !!! THIS ITEM MUST GO LAST IN THE ARRAY !!!
   ]
     .filter(Boolean)
     .join(' ')
 
+  console.log(command)
```
miroslavstastny, line 33: This `console.log` is intentional.

kuzhelov:
it would be great to have shortened version of CLI API with the same functionality preserved - like it is done in jest CLI, as an example. In that way the following might be possible:

```
yarn test:watch Menu -t "passes arbitrary props"
```

The reason I would see it being a better option is solely because of the amount of typing required. We might argue that 'longhand' version is more explicit but, in fact, while it is easier to read, it is harder to memorise the exact word (like `testFilePattern` - as the next day we might try accidentally `filePattern`) that should serve as an option

kuzhelov (approved):

miroslavstastny:
> it would be great to have shortened version of CLI API

Added `-F -t` (unfortunately I cannot use lowercase `-f` as it is consumed by `gulp`) 
```sh
yarn test:watch -F Menu -t "passes arbitrary props"
```

PR #1143, opened 2019-04-01
[RFC] feat(chore): add polyfill for focus-visible

### LINKED ISSUE #749: Remove isFromKeyboard in favor of :focus-visible
### Problem description
IsFromKeyboard is not like the other states. It requires runtime style compilation. We cannot iterate over the style functions and produce stylesheets due to the presence of this runtime state value. 

We could opt to transform this state into a special class name and require consumers to use isFromKeyboard to apply the class name to their own components, but this is complicated and awkward. 

### Proposed solution
There is a :focus-visible spec that only applies styles on focus from keyboard. There is also an official polyfill. We should investigate this spec compliant future proof selector. It will allow us to iterate style functions without a runtime necessary. 

https://css-tricks.com/keyboard-only-focus-styles
https://github.com/WICG/focus-visible

mnajdova:
From the issue https://github.com/stardust-ui/react/issues/749 created by @levithomason 
# Problem description
IsFromKeyboard is not like the other states. It requires runtime style compilation. We cannot iterate over the style functions and produce stylesheets due to the presence of this runtime state value.

We could opt to transform this state into a special class name and require consumers to use isFromKeyboard to apply the class name to their own components, but this is complicated and awkward.

# Proposed solution
There is a `:focus-visible` spec that only applies styles on focus from keyboard https://css-tricks.com/keyboard-only-focus-styles
There is also an official polyfill https://github.com/WICG/focus-visible which is added in this PR. 

Without any processing from our side we can convert the styles that currently rely on the isFromKeyboard flag int he following manner:
#### Previous
```
// styles applied only on keyboard focus 
...(isFromKeyboard && {
  ':focus': {
    color: colorFocus,
    backgroundColor: backgroundColorFocus,
    borderColor: borderColorFocus,
    ':after': {
      content: '""',
      position: 'absolute',
      top: `-${pxToRem(borderWidth * 2)}`,
      right: `-${pxToRem(borderWidth * 2)}`,
      bottom: `-${pxToRem(borderWidth * 2)}`,
      left: `-${pxToRem(borderWidth * 2)}`,
      borderWidth: pxToRem(borderWidth),
      borderStyle: 'solid',
      borderColor: borderColorFocusIndicator,
      borderRadius: borderRadiusFocused,
    },
  },
}),
// styles applied only on mouse focus 
...(!isFromKeyboard && {
  ':focus': {
    ':active': {
      color: colorActive,
      backgroundColor: backgroundColorActive,
      borderColor: borderColorActive,
      boxShadow: 'none',
    },
  },
}),
```
#### Current

```
// styles applied only on keyboard focus 
'[data-focus-visible-added]': {
  color: colorFocus,
  backgroundColor: backgroundColorFocus,
  borderColor: borderColorFocus,
  ':after': {
    content: '""',
    position: 'absolute',
    top: `-${pxToRem(borderWidth * 2)}`,
    right: `-${pxToRem(borderWidth * 2)}`,
    bottom: `-${pxToRem(borderWidth * 2)}`,
    left: `-${pxToRem(borderWidth * 2)}`,
    borderWidth: pxToRem(borderWidth),
    borderStyle: 'solid',
    borderColor: borderColorFocusIndicator,
    borderRadius: borderRadiusFocused,
  },
},
// styles applied only on mouse focus 
':focus:not([data-focus-visible-added])': {
  ':active': {
    color: colorActive,
    backgroundColor: backgroundColorActive,
    borderColor: borderColorActive,
    boxShadow: 'none',
  },
},
```

We may add fela plugin, which will allow the users to use `:focus-visible` instead of `[data-focus-visible-added]`, in a way that we will just replace this strings accordingly.

The PR fixes https://github.com/stardust-ui/react/issues/749


bmdalex (approved):

notandrew:
This is awesome! Thanks for investigating!

packages/react/src/themes/teams-high-contrast/components/Menu/menuItemStyles.ts
```diff
@@ -22,10 +22,10 @@ const menuItemStyles: ComponentSlotStylesInput<MenuItemPropsAndState, MenuVariab
       }),
 
       ...((iconOnly || vertical) && {
-        ...(isFromKeyboard && {
+        '[data-focus-visible-added]': {
```
levithomason, line 25: Per your suggestion about the plugin, I think we should stick with the standard `:focus-visible` pseudo selector.  

The style files should be as vanilla as possible.  When compiled to CSS sheets, users can either use our plugin or use their own CSS processor to handle their conversions.  Eventually, the `:focus-visible` will just work without any plugins or processors since it is in the CSS spec.

levithomason:
This PR should also make it possible to remove:
- most or all of `src/lib/whatInput.ts`
- `isFromKeyboard` state in:
  - [ ] Button.tsx
  - [ ] ListItem.tsx
  - [ ] Attachment.tsx
  - [ ] Popup.tsx
  - [ ] RadioGroupItem.tsx
  - [ ] MenuItem.tsx
  - [ ] ChatMessage.tsx
  - [ ] Reaction.tsx

layershifter:
There are several issues with polyfill, also missing browser don't allow to say obviously how it should work.

Currently, we are using Fela enhancer (#1850) that uses `what-input` attributes to simulate `:focus-visible` behavior, it's also to possible to use `:focus-visible` in our styles. Closing for housekeeping.

PR #1146, opened 2019-04-01
feat(testing): adding e2e cypress tests for Dropdown

silviuaavram:
Added end to end tests using Cypress. Also added them to CircleCI as a separate job. At the moment there are just a few `Dropdown` tests.

Added cypress as a gulp task, and used that in package.json as `yarn test:2e2`.

Feedback on every part of this PR is welcomed.


kolaps33 (approved):

.circleci/config.yml
```diff
@@ -140,3 +148,41 @@ jobs:
...
+            curl -o- -L https://yarnpkg.com/install.sh | bash
+            echo 'export PATH="${PATH}:${HOME}/.yarn/bin"' >> $BASH_ENV
+      - checkout
+      # because we don't invoke npm (we use yarn) we need to add npm bin to PATH manually
+      - run:
+          name: Add npm bin to PATH
+          command: echo 'export PATH="${PATH}:$(npm bin)"' >> $BASH_ENV
+      - restore_cache:
+          keys:
+            - v1-dependencies-{{ checksum "yarn.lock" }}
```
miroslavstastny: This changed to `v1.1` in master, please update it here (and in `save_cache`) as well

miroslavstastny (commented):
The code looks good 👍 

The problem I have is that it adds the third parallel process to be run in CI. Considering we have 4 processes limit in our plan we can no longer run two builds in parallel. Does it make sense to have this as a separate process? Or as `t(screener)+t(cypress) < t(build)` can we make this a part of visual tests process? @kuzhelov what's your opinion on this?

miroslavstastny:
If we decide to merge the current approach with a separate process, we should also mark `ci/circleci: e2e` as required in branch protection rules for `master` in github config.

.circleci/config.yml
silviuaavram: thanks for pointing out, changed

Issue #1149, opened 2019-04-01
Dropdown accessibility broken when highlightFirstItemOnOpen prop is set


kohlikohl:
## Bug Report

Focus gets stuck inside the input field of the `<Dropdown>` component when `highlightFirstItemOnOpen` prop is set.

### Steps
See this [codepen](https://codesandbox.io/s/7245lw8l50) for repro.
* Press tab to navigate to buttons
* ENTER to open the popup
* When `highlightFirstItemOnOpen` is set pressing TAB does not switch to the button in the popup

### Expected Result
* Pressing TAB when focus is in the input field should switch focus to other elements in the popup.

### Version
0.25.1

silviuaavram:
fixed

PR #1150, opened 2019-04-02
fix(dropdown): fixes tab focus on closed list

silviuaavram:
Addresses https://github.com/stardust-ui/react/issues/1149


kolaps33 (approved):
Tested, from my executed scenarios it looks good  :)

miroslavstastny (approved):
please add changelog entry

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -766,14 +765,15 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     ) => {
       switch (keyboardKey.getCode(e)) {
         case keyboardKey.Tab:
-          if (!_.isNil(highlightedIndex)) {
-            selectItemAtIndex(highlightedIndex)
-            // Keep focus here if condition applies.
-            if (this.props.multiple && !this.props.moveFocusOnTab) {
-              e.preventDefault()
+          if (this.state.open) {
```
kuzhelov: lets encapsulate this logic as it is identical to the one that is provided here: https://github.com/stardust-ui/react/pull/1150/files#diff-27df3be097c43ac2046615309f24d815R912. Also, it would be great to simplify this logic, as there is already lots of branching that becomes very hard to track

silviuaavram: you're right, will do that. thanks!

PR #1153, opened 2019-04-02
fix(Popup): avoid double rendering

layershifter:
# BREAKING CHANGES

`defaultTarget` prop is removed on `Popup` component, please use directly `target` prop.

----

## Goal

This PR backports changes from SUIR. It allows us to avoid double rendering of `Popup` because we don't store `target` in state anymore. `target` is no longer autocontrolled prop because we don't need in component's state.

It uses an idea with [`VirtualReference`](https://github.com/FezVrasta/react-popper#usage-without-a-reference-htmlelement) from `react-popper`, so we can can use `createRef()` API from React and avoid double rendering as we don't need to store `target` in components state more.

-------

## Additional: fix Dropdown items' positioning

Also, these changes are needed for #1312 : to render dropdown items in the correct area based on align and placement props and Popper integration

### What

- currently everything seems to be working, but this is because of double render, as we usually open the Dropdown by mouse/keyboard (causing a double render);
- however, **if we set open flag to be always `true` on a `Dropdown`, then PopperJS does not render the items on first render**

### Note on alternative solution

The alternative would be to introduce logic to render the `Dropdown` items only when the `Dropdown` open flag is `true`; **but this breaks accessibility concerns, as the list of `Dropdown` items always needs to be open**


packages/react/src/components/Popup/createPopperReferenceProxy.ts
```diff
@@ -0,0 +1,37 @@
+import { isRefObject, toRefObject } from '@stardust-ui/react-component-ref'
+import * as _ from 'lodash'
+import * as React from 'react'
+
+class ReferenceProxy {
```
kuzhelov: is there any interface provided by PopperJS, so that we would have clear errors in case if ours violates it?

packages/react/src/components/Popup/createPopperReferenceProxy.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+/**
+ * Popper.js does not support ref objects from `createRef()` as referenceElement. If we will pass
+ * directly `ref`, `ref.current` will be `null` at the render process. We use memoize to keep the
+ * same reference between renders.
+ *
+ * @see https://popper.js.org/popper-documentation.html#referenceObject
+ */
+const createPopperReferenceProxy = _.memoize(
```
kuzhelov: how much gain this optimization introduces? Am I right that this helps to omit object being created only, or there is something else I am missing?

packages/react/src/components/Popup/createPopperReferenceProxy.ts
layershifter: Yep, add `implements PopperJS.ReferenceObject`

packages/react/src/components/Popup/createPopperReferenceProxy.ts
layershifter: > Am I right that this helps to omit object being created only

Exactly 👍

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -424,15 +423,15 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
       }),
     }
 
+    const referenceElement = createPopperReferenceProxy(target || this.triggerRef)
```
kuzhelov, line 425: just to be clear - it seems that there is a reason why this function doesn't accept just HTML element as an arg. I might guess that this is because in ref's case it is necessary to make a delayed evaluation of `current`s bounding rectangle, right?

Just trying to understand what makes guarantees to us that `triggerRef` is assigned with the corresponding DOM element at the moment when Popup is rendered with `open=true` prop?

layershifter: > I might guess that this is because in ref's case it is necessary to make a delayed evaluation of currents bounding rectangle, right?

Yep, `referenceElement` in `react-popper` can be only an HTMLElement or such proxy. We can't pass `triggerRef.current` during render phase because it will be `null` on initial render. But it will be asked by Popper for measures only in effects.

packages/react/src/components/Popup/createPopperReferenceProxy.ts
kuzhelov: why we would need it then? I see it as micro optimization that doesn't introduce much value - we might just safely omit that and simplify the logic

layershifter: https://github.com/FezVrasta/react-popper/blob/master/src/Popper.js#L166
To avoid useless calculations in Popper

packages/react/src/components/Popup/createPopperReferenceProxy.ts
```diff
@@ -0,0 +1,38 @@
...
+  (reference: any) =>
```
kuzhelov: could we be explicit here about the type of arg this function expects to handle? So that it might be `HTMLElement` or `Ref` only

kuzhelov (approved):

packages/react/src/components/Popup/createPopperReferenceProxy.ts
kuzhelov: this will create an object that will be stored for the entire app's lifetime in the `memoize's` map, for every Popup content that was once rendered (or asked to be rendered). Lets avoid this problem.

packages/react/src/components/Popup/createPopperReferenceProxy.ts
```diff
@@ -0,0 +1,48 @@
...
+const referenceProxies = new WeakMap<React.RefObject<HTMLElement>, ReferenceProxy>()
+
+/**
+ * Popper.js does not support ref objects from `createRef()` as referenceElement. If we will pass
+ * directly `ref`, `ref.current` will be `null` at the render process. We use memoize to keep the
+ * same reference between renders.
+ *
+ * @see https://popper.js.org/popper-documentation.html#referenceObject
+ */
+const createPopperReferenceProxy = (reference: HTMLElement | React.RefObject<HTMLElement>) => {
```
kuzhelov, line 33: lets introduce comment that would explain why memoization is necessary

bmdalex (approved):
👍 would be great to get this merged soon so I can reuse it for Dropdown #1294

packages/react/src/components/Popup/createPopperReferenceProxy.ts
layershifter: Updated it...

PR #1159, opened 2019-04-03
fix(Alert): imports

mnajdova:


layershifter:
I suggest to find the issue before merge why it was not captured by our project tests (Create React App with TS).

miroslavstastny (approved):

miroslavstastny:
> I suggest to find the issue before merge why it was not captured by our project tests (Create React App with TS).

Let's merge it first to have healthy master, and analyse the tests separately.

PR #1160, opened 2019-04-03
test(CRA): update tsconfig.json to disable `skipLibCheck`

layershifter:
Follow up for #1159.


layershifter:
image

https://circleci.com/gh/stardust-ui/react/10259

build/gulp/tasks/test-projects/cra/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "forceConsistentCasingInFileNames": true,
+    "module": "esnext",
+    "moduleResolution": "node",
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "noEmit": true,
+    "jsx": "preserve"
+  },
+  "include": ["src"]
+}
```
layershifter, line 19: Copied from existing config as is, changed:
```
skipLibCheck => false
strict => false // as we don't comply to strict mode
```

layershifter:
It's a short term fix to avoid regressions, we are going introduce a separate project test for it.

PR #1162, opened 2019-04-03
chore: add circular dependencies check

kuzhelov:
This PR introduces checks that are necessary to prevent uncontrolled occurrence of circular dependencies in Stardust's code. There is also a set of `benign` references that were put to the dedicated file, the ones that should have no impact on consumer scenarios.


build/gulp/tasks/test-circulars/config.ts
```diff
@@ -0,0 +1,21 @@
...
+  [reactPackageDist('components/Menu/Menu.js'), reactPackageDist('components/Menu/MenuItem.js')],
+  [
+    reactPackageDist('components/Button/Button.js'),
+    reactPackageDist('components/Button/ButtonGroup.js'),
+  ],
+  [
+    reactPackageDist('components/Reaction/Reaction.js'),
+    reactPackageDist('components/Reaction/ReactionGroup.js'),
+  ],
+  [
```
kuzhelov: looking into it..

build/gulp/tasks/test-circulars/utils.ts
```diff
@@ -0,0 +1,48 @@
+import * as path from 'path'
+import * as CircularDependencyPlugin from 'circular-dependency-plugin'
+
+import config from '../../../../config'
+
+export const isCycleToSkip = (proposedCycle, benignCycles) => {
```
layershifter:
```suggestion
export const isCycleToSkip = (proposedCycle, beginCycles) => {
```

layershifter (approved):

PR #1106, opened 2019-03-26
perf(Theme): performance measures for Context usage

layershifter:
This PR has two targets:
- test different approach to get `theme`
- improve our profiling tooling

-----

#### 1. Use `minimatch` for `--filter`

This allows to get flexible filtering and run only needed examples:
`yarn gulp perf --filter=*Context*Nesting*`
`yarn gulp perf "--filter=*(ContextUse|ContextProp|ContextConsumer|ContextFelaTheme).*"`

#### 2. Add progress bar for perf tests

perf-progress

#### 3. Add markdown table to output

You can now just copy-paste to Github 👍 

#### 4. Run perf tests with production React

As we concentrated on production performance, with should run production React and pass `process.env.NODE_ENV = production` to get optimizations for third party libraries.

#### 5. Add `perf:debug`

This will start development build of React to get measurements in flamegraph and will allow to run perf tests:

image

-----

`gulp perf --filter=*Context*Nesting* --times=100`

| Example | Avg | Avg diff | Median | Median diff |
| --- |  ---  |  ---  |  ---  |  ---  |
| ContextPropNesting.perf.tsx | 25 | +1.5% | 22.3 | **22.3** |
| ContextUseNesting.perf.tsx | 24.63 | **24.63** | 23.32 | +4.57% |
| ContextConsumerNesting.perf.tsx | 32 | +29.92% | 31.29 | +40.31% |
| ContextFelaThemeNesting.perf.tsx | 36.97 | +50.1% | 34.99 | +56.91% |

| Example | Avg | Avg diff | Median | Median diff |
| --- |  ---  |  ---  |  ---  |  ---  |
| ContextPropNesting.perf.tsx | 25.49 | +5.94% | 25.71 | +14.42% |
| ContextUseNesting.perf.tsx | 24.06 | **24.06** | 22.47 | **22.47** |
| ContextConsumerNesting.perf.tsx | 31.8 | +32.17% | 29.77 | +32.49% |
| ContextFelaThemeNesting.perf.tsx | 34.3 | +42.56% | 33.18 | +47.66% |

| Example | Avg | Avg diff | Median | Median diff |
| --- |  ---  |  ---  |  ---  |  ---  |
| ContextPropNesting.perf.tsx | 24.06 | **24.06** | 22.72 | **22.72** |
| ContextUseNesting.perf.tsx | 25.43 | +5.69% | 25.6 | +12.68% |
[...]

PR #1006, opened 2019-03-01
WIP: perf(theme): optimize theme handling

levithomason:
Merging themes on context is a hot code path. It is also heavy due to merging themes and applying variables.

This PR is experimenting with approaches to optimize this path for all components.

Initial results are promising, just memoizing `mergeThemes` produces 75-95% performance gain in all components:

```
==============================
Button.perf.tsx
BEFORE  AFTER  SAVED  
42.83   2.89   93%    
43.53   1.57   96%    
10.57   1.11   89%    
86.06   19.54  77%    

==============================
Header.perf.tsx
BEFORE  AFTER  SAVED  
1.13    1.3    -15%   
0.69    0.66   4%     
0.45    0.46   -2%    
4.71    7.63   -62%   

==============================
Icon.perf.tsx
BEFORE  AFTER  SAVED  
8.34    1.33   84%    
8.69    0.56   94%    
3.27    0.38   88%    
16.25   7.99   51%    

==============================
Loader.perf.tsx
BEFORE  AFTER  SAVED  
25.84   1.86   93%    
31.44   1.14   96%    
5.93    0.82   86%    
48.51   11.24  77%    

==============================
HeaderDescription.perf.tsx
BEFORE  AFTER  SAVED  
7.35    2.25   69%    
5.86    1.3    78%    
2.53    0.89   65%    
15.79   16.41  -4%    

==============================
Attachment.perf.tsx
BEFORE  AFTER  SAVED  
35.33   2.47   93%    
26.54   1.44   95%    
7.94    1.04   87%    
68.63   11.54  83%    

==============================
Divider.perf.tsx
BEFORE  AFTER  SAVED  
8.26    1.18   86%    
8.42    0.58   93%    
3.27    0.41   87%    
16.49   11.22  32%    

==============================
List.perf.tsx
BEFORE  AFTER  SAVED  
68.27   68.78  -1%    
67.67   65.39  3%     
54.4    56.3   -3%    
98.21   99.29  -1%    

==============================
Chat.perf.tsx
BEFORE  AFTER  SAVED  
284.64  40.03  86%    
276.29  39.44  86%    
208.04  30.11  86%    
426.62  82.98  81% 
```

PR #1163, opened 2019-04-04
perf(renderComponent): drop FelaTheme and use React.Context directly

layershifter:
This PR implements findings from #1106.

### Goal
We want to get rid of `FelaTheme` as it produces two components in the tree and slow downs rendering.

### How
We have two use cases for `renderComponent()`: `UIComponent` and `createComponent()`.

`createComponent()` allows to use only functions, so we can use `useContext()` hook.
`UIComponent` is a class component, we can't use hooks inside them, but there is [`Class.contextType`](https://reactjs.org/docs/context.html#classcontexttype) that can be adopted for this case.

### Measures

See for more details #1106. Raw measures `ContextUse` and `ContextClassField` are proposed approaches.

| Example | Avg | Avg diff | Median | Median diff |
| --- |  ---  |  ---  |  ---  |  ---  |
| ContextUse.perf.tsx | 44.76 | +2.22% | 38.98 | **38.98** |
| ContextConsumer.perf.tsx | 52.94 | +20.9% | 48.01 | +23.17% |
| ContextClassConsumer.perf.tsx | 63.73 | +45.54% | 58.06 | +48.95% |
| ContextClassField.perf.tsx | 55.79 | +27.4% | 51.22 | +31.4% |

### From real life

I used #1006 to compare.

```
⚛ App [mount]697.1149999881163
⚛ App [mount]735.6550000258721
⚛ App [mount]727.5300000328571
⚛ App [mount]706.754999991972
```

```
⚛ App [mount]652.7350000105798
⚛ App [mount]602.1899999468587
⚛ App [mount]626.8550000386313
⚛ App [mount]630.9650000184774
```

Savings are 70-100ms.

### DIFF
--- a/packages/react/src/lib/UIComponent.tsx
+++ b/packages/react/src/lib/UIComponent.tsx
@@ -1,5 +1,8 @@
 import * as React from 'react'
 import * as _ from 'lodash'
+// @ts-ignore We have this export in package, but it is not present in typings
+import { ThemeContext } from 'react-fela'
+
 import renderComponent, { RenderResultConfig } from './renderComponent'
 import { AccessibilityActionHandlers } from './accessibility/types'
 import { FocusZone } from './accessibility/FocusZone'
@@ -11,6 +14,7 @@ class UIComponent<P, S = {}> extends React.Component<P, S> {
   static displayName: string
   static className: string
 
+  static contextType = ThemeContext
   static propTypes: any
 
   /** Array of props to exclude from list of handled ones. */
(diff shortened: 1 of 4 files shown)


packages/react/src/lib/renderComponent.tsx
```diff
@@ -144,98 +146,86 @@ const renderComponent = <P extends {}>(config: RenderConfig<P>): React.ReactElem
...
+
+  const resolvedStyles: ComponentSlotStylesPrepared = Object.keys(mergedStyles).reduce(
+    (acc, next) => ({ ...acc, [next]: callable(mergedStyles[next])(styleParam) }),
+    {},
   )
+
+  const classes: ComponentSlotClasses = getClasses(renderer, mergedStyles, styleParam)
+  classes.root = cx(className, classes.root, props.className)
+
+  const resultConfig: RenderResultConfig<P> = {
```
layershifter: There is no real changes is this code except rename `config` to `resultConfig` as we already have `config` in this function.

packages/react/src/lib/UIComponent.tsx
```diff
@@ -1,5 +1,8 @@
 import * as React from 'react'
 import * as _ from 'lodash'
+// @ts-ignore We have this export in package, but it is not present in typings
```
kuzhelov, line 3: 👍

packages/react/src/lib/renderComponent.tsx
kuzhelov: nit: should we name it as `resolvedConfig` or something similar, as, otherwise, it is a bit hard to reason what `result` stands for. However, just a suggestion, feel free to decide :)

kuzhelov (approved):
please, just update the CHANGELOG before merging 👍

packages/react/src/lib/renderComponent.tsx
layershifter: Renamed 👍

PR #1164, opened 2019-04-04
test(projects): add a new check to ensure that typings are correct

layershifter:
Follow up for #1159.

Our existing test with Create React App uses `skipLibSkip` that doesn't allow us to capture errors in typings. This PR adds a new check that will enforce this.


build/gulp/tasks/test-projects/webpack/webpack.config.js
```diff
@@ -0,0 +1,27 @@
...
+  },
+  resolve: {
+    extensions: ['.tsx', '.ts', '.js'],
+  },
+  output: {
+    filename: 'bundle.js',
+    path: path.resolve(__dirname, 'dist'),
+  },
+  plugins: [new ForkTsCheckerWebpackPlugin()],
+}
```
layershifter: Config replicates simplified settings in `react-scripts`.

layershifter:
image

build/gulp/tasks/test-projects/webpack/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
+    "jsx": "react",
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "module": "esnext",
+    "moduleResolution": "node",
```
kuzhelov: this is the default value for this option (`node`), so we might skip this from being explicitly specified.

Actually, here I would rather expect that we will test client scenario that does as many overrides as possible for the TS compiler - i.e. the config that is produced by `tsc --init`, - as this mimics default configuration provided by TS team. 

Otherwise, frankly, it is hard for me to understand semantics of this test and, essentially, understand _what_ it verifies - unless we would explicitly state in the docs that this is the _recommended_ TS configuration. However, this would mean that we are imposing additional restriction s to the client (not saying that we shouldn't do that, but we need to justify them first then).

What do you think?

layershifter: > this is the default value for this option (node), so we might skip this from being explicitly specified.

Nope, `module === "AMD" or "System" or "ES6" ? "Classic" : "Node"`

image

layershifter: > Actually, here I would rather expect that we will test client scenario that does as many overrides as possible for the TS compiler - i.e. the config that is produced by tsc --init, - as this mimics default configuration provided by TS team.

1. By `tsc --init` there is `strict: true` which will not work with Stardust
2. Without `lib` section nothing will work :(

layershifter: Actually, this config is closer too official tutorial: https://www.typescriptlang.org/docs/handbook/react-&-webpack.html

layershifter: We can do:
```diff
-    "module": "esnext",
-    "moduleResolution": "node",
+    "module": "commonjs",
```

layershifter: After discussion we decided to use plain `tsc`:

image

build/gulp/tasks/test-projects.tsx
```diff
@@ -101,7 +101,10 @@ const performBrowserTest = async (publicDirectory: string, listenPort: number) =
   const server = await startServer(publicDirectory, listenPort)
 
   const browser = await puppeteer.launch({
-    args: ['--single-process'], // Workaround for newPage hang in CircleCI: https://github.com/GoogleChrome/puppeteer/issues/1409#issuecomment-453845568
+    args: [
+      // Workaround for newPage hang in CircleCI: https://github.com/GoogleChrome/puppeteer/issues/1409#issuecomment-453845568
+      process.env.CI && '--single-process',
+    ].filter(Boolean),
```
layershifter, line 107: Without this will fail locally on Windows

build/gulp/tasks/test-projects.tsx
```diff
@@ -199,7 +202,45 @@ task('test:projects:rollup', async () => {
...
+
 task(
   'test:projects',
-  series('dll', 'bundle:all-packages', parallel('test:projects:cra-ts', 'test:projects:rollup')),
+  series(
+    'dll',
+    'bundle:all-packages',
+    'test:projects:cra-ts',
+    'test:projects:rollup',
+    'test:projects:typings',
```
layershifter, line 244: We can't run in parallel now, `yarn` fails with cache errors, see previous builds

build/gulp/tasks/test-projects/typings/index.tsx
```diff
@@ -0,0 +1,35 @@
+import {
```
kuzhelov: here we might want to simplify things, just to eliminate any maintainability burden. Essentially, what we would like to check is that all the types of stardust module are consistent and have proper imports (i.e. absense of imports like `import SomeType from 'src/..'`). For that the following basic code would be just enough:

```js
import * as stardust from '@stardust-ui/react'
const Test = () => console.log(stardust)

export default Test
```

Current version of code doesn't introduce any additional problem-capturing capabilities, as I see it - am I missing something?

kuzhelov (approved):

PR #1168, opened 2019-04-04
fix(Dropdown): reset highlighted index on input change

silviuaavram:
This should reset the `highlightedIndex` to null or 0, depending whether` highlightFirstItemOnOpen` has been passed or not.

### DIFF
--- a/packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
+++ b/packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
@@ -58,7 +58,7 @@ describe('Dropdown', () => {
       )
     })
 
-    it('has the provided prop value + 1 when opened by ArrowDown', () => {
+    it('has the (provided prop value + 1) when opened by ArrowDown', () => {
       const highlightedIndex = 1
       const wrapper = mountWithProvider(
         <Dropdown highlightedIndex={highlightedIndex} onOpenChange={onOpenChange} items={items} />,
@@ -78,7 +78,7 @@ describe('Dropdown', () => {
       )
     })
 
-    it('has the provided prop value - 1 when opened by ArrowUp', () => {
+    it('has the provided (prop value - 1) when opened by ArrowUp', () => {
       const highlightedIndex = 1
       const wrapper = mountWithProvider(
         <Dropdown highlightedIndex={highlightedIndex} onOpenChange={onOpenChange} items={items} />,
@@ -138,7 +138,7 @@ describe('Dropdown', () => {
       )
     })
 
-    it('is defaultHighlightedIndex value at first opening', () => {
+    it('is defaultHighlightedIndex prop value at first opening', () => {
       const defaultHighlightedIndex = 1
       const wrapper = mountWithProvider(
         <Dropdown
@@ -198,7 +198,7 @@ describe('Dropdown', () => {
       )
     })
 
-    it('is set to 0 at searchQuery change and when highlightFirstItemOnOpen prop is provided', () => {
+    it('is set to 0 on searchQuery change and when highlightFirstItemOnOpen prop is provided', () => {
       const onSearchQueryChange = jest.fn()
       const wrapper = mountWithProvider(
         <Dropdown
--- a/packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
+++ b/packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
@@ -178,7 +178,7 @@ describe('Dropdown', () => {
       )
     })
 
-    it('is always 0 when highlightFirstItemOnOpen prop is provided', () => {
+    it('is 0 on first open when highlightFirstItemOnOpen prop is provided', () => {
       const wrapper = mountWithProvider(
         <Dropdown highlightFirstItemOnOpen onOpenChange={onOpenChange} items={items} />,
       )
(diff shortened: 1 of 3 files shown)


packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -57,7 +58,7 @@ describe('Dropdown', () => {
       )
     })
 
-    it('calls onOpen with highlightedIndex + 1 on arrow down', () => {
+    it('has the provided prop value + 1 when opened by ArrowDown', () => {
```
kuzhelov: just for the sake of better clarity, lets use braces:

> has value of (provided prop value + 1) when opened by ArrowDown

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -77,7 +78,7 @@ describe('Dropdown', () => {
       )
     })
 
-    it('calls onOpen with highlightedIndex - 1 on arrow down', () => {
+    it('has the provided prop value - 1 when opened by ArrowUp', () => {
```
kuzhelov: nit: lets try to avoid the terminology that is code specific, like `ArrowUp` - we might better use implementation-agnostic `arrow up key`

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -137,7 +138,7 @@ describe('Dropdown', () => {
       )
     })
 
-    it('calls onOpen with highlightedIndex set to defaultHighlightedIndex', () => {
+    it('is defaultHighlightedIndex value at first opening', () => {
```
kuzhelov: nit-nit: just to be on par with the naming used fo rother test cases, lets add `prop` word, to better describe origin of `defaultHighlightedIndex`:

> is defaultHighlightedIndex **prop** value at first opening

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -156,11 +157,24 @@ describe('Dropdown', () => {
           open: true,
         }),
       )
+
+      wrapper
```
kuzhelov, line 166: actually, its not easy to see the reason for the event to happen this amount of times (which is requested by test). Could you, please, suggest why `onOpenChange` fires twice for each click?

kuzhelov: also, this check, apparently, verifies that on the subsequent open events `defaultHighlightedIndex` value is not honored. I would suggest to extract this check to the dedicated test case, with the corresponding name

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -156,11 +157,24 @@ describe('Dropdown', () => {
...
+          highlightedIndex: null,
+          open: true,
+        }),
+      )
     })
 
-    it('calls onOpen with highlightedIndex always set to 0', () => {
+    it('is always 0 when highlightFirstItemOnOpen prop is provided', () => {
       const wrapper = mountWithProvider(
-        <Dropdown highlightFirstItemOnOpen={true} onOpenChange={onOpenChange} items={items} />,
```
kuzhelov, line 163: 👍

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -183,6 +197,95 @@ describe('Dropdown', () => {
         }),
       )
     })
+
+    it('is set to 0 at searchQuery change and when highlightFirstItemOnOpen prop is provided', () => {
```
kuzhelov: nit: `at` -> `on`: 

> is set to 0 **on** searchQuery change and when highlightFirstItemOnOpen prop is provided

kuzhelov (commented):
agree with the changes - could you, please, take a look on the comments for the tests part, to clarify couple of moments for me? Thank you!

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
silviuaavram: sure, good idea!

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
silviuaavram: well, it is 3 because it opened a couple of times and closed once.

I just want to make sure that on opening it the second time, it opens and there is no item highlighted.

what do you think I should change?

kuzhelov: got it - I think that it would be more apparent if you would introduce a separate test case for that, the one that would follow a plan like:
- do whatever is necessary to prepare test state (arrange + act)
- verify conditions (assert)

Specifically to this example, it could be done the following way:
- introduce test case with a name like `is not set on second and subsequent openings`
- do the arrangement once there

```jsx
wrapper
  .find(...)
  . ...
  .simulate('click')
  .expect( .. ) // and this is where assertion phase starts
```

If there will be just single arrangement phase followed by assertion, it would be much easier to infer where the numbers come from :) currently the main problem is that there are several such phases within single test, so that the side effect of the previous arrangement affects the following one, which is not intuitive and harder to maintain

silviuaavram: split the test into two.

kuzhelov (approved):

PR #1169, opened 2019-04-04
chore(package): switch distribution build to Babel

### LINKED ISSUE #508: RFC: switch to Babel

## 1. Switch docs

### Actions

### Goals

## 3. Switch to `gulp-babel`

### Use existing plugins

### LINKED ISSUE #998: RFC: Ref utils
### 1. Move `Ref` component to a separate package

This fits #433. As this component is shared between SUIR and Stardust UI.
Proposal: `@stardust-ui/react-ref`

#### 2.1. `isRefObject<T extends any>(value: T): is React.RefObject<T>`

To check that passed object is a ref created with `React.createRef()`.

```jsx
const ref = React.createRef()

isRefObject(ref)
isRefObject(document)
```

_When:_ the great example is a `context` prop in `Popup` it's very useful to be able to pass there refs and DOM elements.

#### 2.2. `toRefObject<T extends Node>(value: T): React.RefObject<T>`

To create a ref object from existing DOM node.

When: some of our internal API will support only `React.RefObject` as a prop value (#949).

```jsx
const element = document.createElement('div')
const ref = toRefObject(element)

ref === toRefObject(element) // true, is memoized
```

#### 2.3 `mergeRefs(...refs: React.Ref<any>)`

To merge refs to single assign.

_When:_ you need to pass ref to top level or third party component.

```diff
-<Ref innerRef={domNode => {
-  ref(domElement)
-  this.popupDomElement = domElement
-  handleRef(contentRef, domElement)
-}} >
+<Ref innerRef={mergeRefs(ref, this.popupRef, contentRef)} >

layershifter:
Fixes #508.

----

## Improve bundling :yay

This PR is a final step in Babel migration and aims to improve and simplify our bundling, too.

### New distribution structure

Each package will have the following structure:
- `dist-node` - single bundled file with CommonJS modules distribution
- `dist-src` - source files compiled with Babel, has the same structure as project, preserves ability to pick single files from project
- `dist-types` - contains bundle single file with types (expect `@stardust-ui/react`, _see below why_)
- `dist-web` - an ESM distribution of package via single file, built & optimized to run in most web browsers, tree shaking will work properly in this case, don't worry 📦 

Single file distribution becomes standard, even React is distributed via single file. Current distribution for ES modules includes 1,556 files, it's definitely better to have one for bundlers ❤️ 

### Uses `@pika/pack`

To get the structure above no custom code is required, everything is done using [`@pika/pack`](https://github.com/pikapkg/pack), this allow us to add custom stages to their pipeline and make complicated bundling easily 👍 

### Configured `@babel/preset-env`

We use `.browserslistrc` to target required browsers, this setting can be adjusted later, [current config](https://browserl.ist/?q=%3E0.2%25%2C+not+dead%2C+not+ie+%3C+11%2C+not+op_mini+all).

### Adds `@pika/plugin-build-types`

It's a custom plugin that will bundle definitions to a single file with [`rollup-plugin-dts`](https://github.com/Swatinem/rollup-plugin-dts), the nicest thing is that it will contain only publicly exported definitions 🐺 

### Switch to `@babel/runtime`

As we switched to Babel, we also replaced `tslib` with `@babel/runtime-corejs2`.

### `babel-plugin-lodash` 

This PR also introduces the first babel plugin that aims to optimize bundle size. Lodash is CJS and can't be treeshaked, it means that `import _ from 'lodash'` will include the whole library (70 Kb) 😭 

This plugin will transform `import _ from 'lodash'` to `import _map from 'lodash/map'` where `_map` is module required in your file.

image

### TS Project References for `@stardust-ui/react`

There is only one reason to introduce this: our `withSafeTypeForAs()` creates very complicated TS definitions and these definitions can't be bundled due some bugs in `rollup-plugin-dts` 💣 

https://cdn.jsdelivr.net/npm/@stardust-ui/react@0.30.0/dist/es/components/Chat/Chat.d.ts
https://cdn.jsdelivr.net/npm/@stardust-ui/react@0.31.0/dist/es/components/Chat/Chat.d.ts

Because of this, the PR also introduces changes to our `tsconfig.json`'s to use [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html) feature. It was designed also to handle monorepo cases (https://github.com/Microsoft/TypeScript/issues/25376).

> When you reference a project, new things happen:
> - Importing modules from a referenced project will instead load its output declaration file (.d.ts)
> - If the referenced project produces an outFile, the output file .d.ts file’s declarations will be visible in this project

https://www.typescriptlang.org/docs/handbook/project-references.html

This feature allows to properly define `rootDir` option in `tsconfig.json` and generate output proper typings by compiler:

image

Without the references feature it will fail because TypeScript doesn't know anything about modules structure:
image

### Ongoing stuff

- fix some HMR issues
- introduce `babel-transform-remove-proptypes`


layershifter:
I want to get this feature (https://github.com/pikapkg/pack/pull/50) before merge.

PR #1175, opened 2019-04-05
fix(MenuItem): click triggering logic on Space and Enter key press

### LINKED ISSUE #1155: Menu with toolbarBehavior not keyboard accessible
### Steps
When using a menu as a toolbar with `toolbarBehavior` and `toolbarButtonBehavior`, where individual items open a Popup, pressing ENTER to open the Popup does not work.

https://codesandbox.io/s/r5p9p1487n

### Version
0.25.1

kuzhelov:
Fixes #1155 .

This PR introduces fix for accessibility logic that is aimed to trigger `click` event handler on `Space` and `Enter` key events. 

The problem was that `click` was triggered unconditionally while, to match browser's behavior, it should be triggered only in case if event is not default-prevented by `e.preventDefault()`.

### DIFF
--- a/packages/react/src/components/Menu/MenuItem.tsx
+++ b/packages/react/src/components/Menu/MenuItem.tsx
@@ -285,7 +285,7 @@ class MenuItem extends AutoControlledComponent<ReactProps<MenuItemProps>, MenuIt
   }
 
   protected actionHandlers: AccessibilityActionHandlers = {
-    performClick: event => this.handleClick(event),
+    performClick: event => !event.defaultPrevented && this.handleClick(event),
     openMenu: event => this.openMenu(event),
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -285,7 +285,7 @@ class MenuItem extends AutoControlledComponent<ReactProps<MenuItemProps>, MenuIt
   }
 
   protected actionHandlers: AccessibilityActionHandlers = {
-    performClick: event => this.handleClick(event),
+    performClick: event => !event.defaultPrevented && this.handleClick(event),
```
sophieH29, line 288: nice 👍

sophieH29 (approved):

PR #1179, opened 2019-04-05
fix(Icon): rotate in rtl Teams theme

mnajdova:
Fixes https://github.com/stardust-ui/react/issues/1181

### BREAKING CHANGES
The `flipInRtl` slot inside Icon's styles was renamed to `svgFlippingInRtl`. The new slot contains all `svg` styles + the prev `flipInRtl` styles. However, this shouldn't affect any client scenarios.


packages/react/src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -150,10 +144,22 @@ const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
         fill: v.disabledColor,
       }),
 
+      transform: `rotate(${rotate}deg)`,
+
       ...getSvgStyle('svg'),
     }
   },
 
+  flipInRtl: config => {
```
kuzhelov: as we've agreed on, lets rename it to `svgFlippingInRtl`, as now this set of styles has changed its semantics by containing all the necessary ones for the icons that should flip, i.e. both ltr and rtl styles

packages/react/src/themes/teams/components/Icon/svg/icons/bullets.tsx
```diff
@@ -4,12 +4,12 @@ import { TeamsSvgIconSpec } from '../types'
...
 
 export default {
-  icon: ({ classes }) => (
+  icon: ({ classes, rtl }) => (
     <svg
       role="presentation"
       focusable="false"
       viewBox="8 8 16 16"
-      className={cx(classes.svg, classes.flipInRtl)}
+      className={rtl ? classes.flipInRtl : classes.svg}
```
kuzhelov: it should be enough just to use `classes. flipInRtl ` (or `classes.svgFlippingInRtl`) here, as this section now contains all the necessary styles both for ltr and rtl

packages/react/src/themes/teams/components/Icon/svg/icons/bullets.tsx
```diff
@@ -19,7 +19,7 @@ export default {
           <circle cx="10.5" cy="15.5" r="1" />
         </g>
         <g className={cx(teamsIconClassNames.filled, classes.filledPart)}>
-          <path d="M13.5 11.25h9a.75.75 0 0 0 0-1.5h-9a.75.75 0 0 0 0 1.5zM22.5 14.75h-9a.75.75 0 0 0 0 1.5h9a.75.75 0 0 0 0-1.5zM22.5 19.75h-9a.75.75 0 0 0 0 1.5h9a.75.75 0 0 0 0-1.5z" />
+          <path d="M13.5 11.25h9a.75.75 0 0 0 0-1.5h-9a.75.75 0 0 0 0 1.5zM22.5 14iconStyles.ts.75h-9a.75.75 0 0 0 0 1.5h9a.75.75 0 0 0 0-1.5zM22.5 19.75h-9a.75.75 0 0 0 0 1.5h9a.75.75 0 0 0 0-1.5z" />
```
kuzhelov: oops

packages/react/src/themes/teams/components/Icon/svg/icons/bullets.tsx
```diff
@@ -4,12 +4,12 @@ import { TeamsSvgIconSpec } from '../types'
 import { teamsIconClassNames } from '../teamsIconClassNames'
 
 export default {
-  icon: ({ classes }) => (
+  icon: ({ classes, rtl }) => (
```
kuzhelov: `rtl` may be safely removed

packages/react/src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -92,11 +92,16 @@ const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
       backgroundColor: v.backgroundColor,
       boxSizing: isFontBased ? 'content-box' : 'border-box',
 
+      // overriding the base theme default rtl behavior as in teams theme the svg/svgFlippingInRtl slots are used for this
+      ...((rtl || !isFontBased) && {
```
kuzhelov: the comment and the expression itself mean different things - the comment suggests that these styles should go in RTL only, but the condition used doesn't guarantee that (if icon is not a font-based one, the styles will be applied both in RTL and LTR). Lets sync the comment and the condition - probably, the intent was to do just the following:

```js
...(!isFontBased) { .. }

```

kuzhelov (approved):

PR #1180, opened 2019-04-05
fix(FocusTrapZone): do not propagate any keyboard events

### LINKED ISSUE #1156: Keyboard event propagates from Dialog component to ListItem
## Bug Report

When embedding a `<Dialog>` inside the `endMedia` of a ListItem and also adding a `keyDown` handler to the ListItem, pressing ENTER inside the `<Dialog>` to cancel/confirm will trigger the `keyDown` handler on the ListItem.

### Steps
* go to https://codesandbox.io/s/5z00lr1m1p
* press TAB to select the item
* press DEL key to open dialog
* press ENTER on the focused cancel button

### Expected Result
* the dialog closes and the `keyDown` handler on ListItem is not engaged

### Actual Result
* Keyboard event propagates to ListItem

### Version
0.25.1

sophieH29:
closes https://github.com/stardust-ui/react/issues/1156

## Issues caused:
Because of `Portal` events propagating to the parent components in React element tree, it caused post-effects: keyboard event was propagated outside `Portal`, and was handled by another component too. (Issue example https://github.com/stardust-ui/react/issues/1156)

## Why:
`FocusTrapZone` should not only trap focus but totally block user interaction outside of it - i.e. stop propagating keyboard events.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixes
 - Truncate `content` and `header` of `ListItem` when used from `DropdownSelectedItem` @silviuavram ([#1161](https://github.com/stardust-ui/react/pull/1161))
+- `FocusTrapZone` - Do not propagate any keyboard events @sophieH29 ([#1180](https://github.com/stardust-ui/react/pull/1180))
 
 <!--------------------------------[ v0.26.0 ]------------------------------- -->
 ## [v0.26.0](https://github.com/stardust-ui/react/tree/v0.26.0) (2019-04-03)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixes
 - Truncate `content` and `header` of `ListItem` when used from `DropdownSelectedItem` @silviuavram ([#1161](https://github.com/stardust-ui/react/pull/1161))
+- `FocusTrapZone` - Do not propagate any keyboard events @sophieH29 ([#1180](https://github.com/stardust-ui/react/pull/1180))
```
kuzhelov, line 27: could you, please, provide some notes about resolution strategy in this PR's description - I mean, specifically
- what was the cause of the issue
- why this was considered as a fix

This description will help a lot to better understand why provided changes are considered to be a fix for the issue, as well as to retrospect this case if we will experience similar problem in future. Thank you!

mnajdova (approved):

kuzhelov (approved):

PR #1183, opened 2019-04-07
fix: escape behavior of popup and dropdown

### LINKED ISSUE #1079: Popup|Dialog: incorrect handling of `Escape` key
### Steps
1. Open CodeSandbox, https://codesandbox.io/s/j20nlynqm5
2. Click "Open popup" button
3. Click on `Dropdown` to open it
4. Press `Escape key`

### Expected Result
`Dropdown` is closed, `Popup` is opened.

### Actual Result
`Dropdown` is closed, `Popup` is closed.

The same result will be for multiple `Popup`s.

### Version
0.23.1

kuzhelov:
Fixes #1079 .

The issue was caused by the fact that Esc key event was propagated further in the components tree, even being processed by `Popup`/`Dropdown`. Necessary instructions to prevent propagation were added to the code.


silviuaavram:
I tested in the docsite with selection and search PRs. Esc looks fixed, although I am not sure why the popup closes when you click on the search input text. or when you Tab into it and then hit Enter.

This happens in this PR, I tried to repro in that sandbox from the issue and it was ok there.

Used the code below in the Inline example in the Docsite.

```react
import React from 'react'
import { Button, Dropdown, Popup } from '@stardust-ui/react'

const inputItems = [
  'Bruce Wayne',
  'Natasha Romanoff',
  'Steven Strange',
  'Alfred Pennyworth',
  `Scarlett O'Hara`,
  'Imperator Furiosa',
  'Bruce Banner',
  'Peter Parker',
  'Selina Kyle',
]

const PopupExampleInline = () => (
  <Popup
    trigger={<Button icon="expand" />}
    content={
      <>
      <Dropdown
        items={inputItems}
        placeholder="Start typing a name"
        noResultsMessage="We couldn't find any matches."
        getA11ySelectionMessage={{
          onAdd: item => `${item} has been selected.`,
        }}
      />
      <Dropdown
        search
        items={inputItems}
        placeholder="Start typing a name"
        noResultsMessage="We couldn't find any matches."
        getA11ySelectionMessage={{
          onAdd: item => `${item} has been selected.`,
        }}
      />
      </>
    }
    inline
  />
)

export default PopupExampleInline
```

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -199,10 +199,13 @@ export default class Popup extends AutoControlledComponent<ReactProps<PopupProps
 
   protected actionHandlers: AccessibilityActionHandlers = {
     closeAndFocusTrigger: e => {
-      e.stopPropagation()
       this.close(e, () => _.invoke(this.triggerFocusableDomElement, 'focus'))
+      e.stopPropagation()
```
sophieH29, line 203: why is it needed to stop propagation in the popup? As far as understood, the problem was caused in Dropdown which was propagating keyboard event to Popup. Seems like Dropdown changes should fix it, isn't it?

sophieH29: I believe Dropdown in Popup will be usually used with focus trap behavior. This PR https://github.com/stardust-ui/react/pull/1180 stop propagation for any keyboard events within FocusTrapZone

kuzhelov: > @sophieH29  why is it needed to stop propagation in the popup? As far as understood, the problem was caused in Dropdown which was propagating keyboard event to Popup

Actually, this makes just one aspect of the problem, the one that is replicated with the codesandbox example in the issue description: #1079 . If it would be the only problem, than yes, changes to the `Dropdown` would be enough.

However, as it is mentioned in the issue, the same problem applies to Popup -> Popup case, and this behavior is not correct (from issue's description):

> The same result will be for multiple Popups.

---------------

 > @sophieH29  I believe Dropdown in Popup will be usually used with focus trap behavior. This PR #1180 stop propagation for any keyboard events within FocusTrapZone

Yes, it should - however, in the meanwhile this fix is necessary

sophieH29 (approved):

PR #1188, opened 2019-04-08
chore: introduce lint check for invalid JSX attribute names

### LINKED ISSUE #1158: Check 'Invalid DOM property' during build
### Problem description
See #1157 
Invalid attribute was used on HTML element, it was not caught by the build, it logs an error in browser:
image

### Proposed solution
This should be caught by linter or compiler.

There is no such rule in tslint: https://github.com/palantir/tslint-react/issues/135
There is a rule for eslint: https://github.com/yannickcr/eslint-plugin-react/blob/master/lib/rules/no-unknown-property.js

kuzhelov:
This PR eliminates possibility of undetected accidental usage of wrongly-named JSX attributes, as it was the case for the icons: #1157

Fixes #1158


tslint.json
```diff
@@ -4,10 +4,10 @@
     "arrow-parens": false,
     "align": [false],
     "import-name": [false],
+    "jsx-attribute-name": true,
     "linebreak-style": [true, "LF"],
     "max-line-length": [false],
     "no-increment-decrement": false,
-    "no-unused-variable": true,
```
kuzhelov, line 10: as warning message suggests, this is now checked by TS compiler

miroslavstastny (approved):

layershifter (approved):

PR #1186, opened 2019-04-08
fix(Theme): do not set HTML font size in static styles

miroslavstastny:
WIP

PR #1190, opened 2019-04-08
fix(Theme): Use global REM-base, do not set html font-size in static styles

### LINKED ISSUE #988: pxToRem sometimes returning incorrect size
pxToRem is sometimes outputting incorrect values. It appears to sometimes use the wrong remSize in its calculations.

Observed:
image

Expected:
image
The results are consistent with using 16px as the HTML font-size, but in most cases the correct value of 14px is used.

miroslavstastny:
## BREAKING CHANGE
Teams theme no longer sets `font-size` on `html`.
Set correct `font-size` on `html` manually **BEFORE** Stardust is imported.

This is a hotfix which fixes #988.

### What was broken
`pxToRem` uses `html` `font-size` as a base to correctly convert pixels to REMs. There are several problems with it:
1. As `html` `font-size` is global on a page, two themes loaded at the same time cannot use different base font sizes for the `pxToRem`, But the previous semantics with `htmlFontSize` being a theme's `siteVariable` and `Provider` resetting the cached base when a new theme was loaded looked like it is possible.
2. As `pxToRem` is called before the static styles are applied, static `html` `font-size` was cached before the themes font size was applied resulting in broken design (see #988).

### Fix
Teams theme no longer sets `html` `font-size` in its static styles and expects it to be correctly set **BEFORE** Stardust is imported not so set false expectations.

### Limitations of the fix
It's no possible to scale Stardust components to:
1. have part of the page bigger/smaller compared to the rest,
2. zoom the whole page by setting font size in browser settings.

These limitations are not new, they were present before the fix as well.

There is another PR #1186 which is an experiment on how to introduce more advanced fix without these limitations later.


packages/react/src/themes/teams/staticStyles/globalStyles.ts
```diff
@@ -1,9 +1,6 @@
 import { StaticStyleFunction } from '../../types'
 
 const globalStyles: StaticStyleFunction = siteVars => ({
-  html: {
-    fontSize: siteVars.htmlFontSize,
-  },
   body: {
```
miroslavstastny, line 4: I vote for moving all these from `body` to `ProviderBox` (and `popupStyles`) to get rid of an another piece of global styles. Opinions?

packages/react/src/lib/fontSizeUtility.ts
```diff
@@ -16,12 +16,8 @@ const getFontSizeValue = (size?: string | null): number | null => {
   return (size && parseFloat(size)) || null
 }
 
-export const updateCachedRemSize = () => {
-  _documentRemSize = null
-}
-
 /**
- * Converts the provided px size to rem based on the default font size of 10px unless
+ * Converts the provided px size to rem based on the default font size of 16px unless
```
kuzhelov, line 20: 👍

packages/react/src/themes/teams/staticStyles/globalStyles.ts
```diff
@@ -1,9 +1,6 @@
 import { StaticStyleFunction } from '../../types'
 
 const globalStyles: StaticStyleFunction = siteVars => ({
-  html: {
-    fontSize: siteVars.htmlFontSize,
```
kuzhelov, line 5: the result of this line is that for consumer now `1rem = htmlFontSize = 16px`, while previously it was `1rem = htmlFontSize = 10px`. Are we sure this doesn't provide any implications for the client (i.e. another set of static styles is used on the client side that ensure that `htmlFontSize` remains to be `10px`)?

packages/react/src/themes/teams/staticStyles/globalStyles.ts
kuzhelov: lets wait but think on this move - wait just because of emerged asks for `ProviderBox` reuse existing DOM element on render

packages/react/src/lib/fontSizeUtility.ts
```diff
@@ -35,6 +31,9 @@ export const updateCachedRemSize = () => {
  */
 export const pxToRem = (valueInPx: number, baseRemSize?: number): string => {
   if (!baseRemSize && !_documentRemSize) {
+    // there is no way how to reset the cached value, once it's cached, it's cached
```
kuzhelov: nit-nit: may suggest
- to remove `once it's cached, it's cached`, as it provides no additional value in explanation :)
- rephrase the rest in a more succinct way

> as resetting cached value won't trigger recalculation of site variables, for which originally computed values will stay unchanged

kuzhelov (commented):
just one thing I would like to refine before merge: https://github.com/stardust-ui/react/pull/1190/files#r273730374

packages/react/src/themes/teams/staticStyles/globalStyles.ts
kuzhelov: talked through and have verified client scenario - there shouldn't be any related regressions there 👍

kuzhelov (approved):

packages/react/src/themes/teams/staticStyles/globalStyles.ts
miroslavstastny: Ok, let's think about this later.

PR #1192, opened 2019-04-08
fix(docs): fix code example of theming guide

### LINKED ISSUE #1110: Provider requires a theme but Docs suggest not supplying one
## Bug Report
The docs on theme indicate that by default a theme does not have to be provided:
image

However, theme is a required input for the `Provider` component.
image

### Version
0.25.0

### Testcase
https://codesandbox.io/s/mmxpnxzmyp?fontsize=14

kuzhelov:
Fixes #1110 .


docs/src/views/ThemingExamples.tsx
```diff
@@ -44,10 +44,10 @@ export default () => (
     <ExampleSnippet
       value={`
         import React from 'react'
-        import { Button, Icon, Label, Provider } from '@stardust-ui/react'
+        import { Button, Icon, Label, Provider, themes } from '@stardust-ui/react'
 
         export default () =>
-         <Provider>
+         <Provider theme={themes.teams}>
           <>
```
layershifter: _nit:_ We don't need `Fragment` there more

layershifter (approved):

miroslavstastny (approved):

PR #1195, opened 2019-04-09
fix(grid): move focus outline inside grid items

### LINKED ISSUE #1035: Accessibility: Focus border is not rendered correctly on the grid component
## Bug Report
Focus border is not rendered correctly on the grid component

### Steps
Use the provided link, tab to the first item in the grid, use arrows to navigate through the items.

### Expected Result
The focus box should be rendered properly

### Actual Result
The focus box around the items is not rendered properly, gets cut off in some spots

### Version
0.23.0

### Testcase
https://codesandbox.io/s/9y5qylv8o4

bmdalex:
## fix(grid): move focus outline inside grid items

Fixes #1035 

### Before:

Screenshot 2019-04-09 at 17 52 40

### After:

Screenshot 2019-04-09 at 17 53 08

### DIFF
--- a/docs/src/examples/components/Grid/Variations/GridExampleKeyboardNavigable.shorthand.tsx
+++ b/docs/src/examples/components/Grid/Variations/GridExampleKeyboardNavigable.shorthand.tsx
@@ -43,7 +43,7 @@ const renderImages = () => {
 
 const renderImageButtons = () => {
   return _.map(imageNames, imageName => (
-    <Button key={imageName} text styles={imageButtonStyles}>
+    <Button key={imageName} styles={imageButtonStyles}>
       <Image fluid src={`public/images/avatar/large/${imageName}.jpg`} />
     </Button>
   ))
--- a/docs/src/examples/components/Grid/Variations/GridExampleKeyboardNavigable.shorthand.tsx
+++ b/docs/src/examples/components/Grid/Variations/GridExampleKeyboardNavigable.shorthand.tsx
@@ -28,8 +28,8 @@ const imageButtonStyles = {
   height: '72px',
   padding: '0',
   margin: '0',
-  background: '#fff',
 }
+
 const renderImages = () => {
   return _.map(imageNames, imageName => (
     <Image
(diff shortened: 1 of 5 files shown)


packages/react/src/themes/base/components/Grid/gridStyles.ts
```diff
@@ -24,6 +24,7 @@ const gridStyles: ComponentSlotStylesInput<GridProps, GridVariables> = {
       gridGap,
       display: 'grid',
       justifyContent: 'space-evenly',
+      '& > *': { outlineOffset: '-3px' },
```
kuzhelov: as we've discussed, I am suggesting to introduce this as part of the Teams theme, as this is about Teams scenarios that have introduced this issue. For other scenarios we would just introduce the same behavior (and visual appearance) as plain browser styles would provide, so I am suggesting to stick to that base browser's one in the `base` theme

bmdalex: oops...didn't notice it was moved to base; fixed, thanks, @kuzhelov

docs/src/examples/components/Grid/Variations/GridExampleKeyboardNavigable.shorthand.tsx
```diff
@@ -43,31 +43,21 @@ const renderImages = () => {
 
 const renderImageButtons = () => {
   return _.map(imageNames, imageName => (
-    <Button key={imageName} styles={imageButtonStyles}>
+    <Button key={imageName} text styles={imageButtonStyles}>
```
kuzhelov: would suggest to remove this Button's prop, however I do understand why it was introduced. Still, we shouldn't fine tune the examples because of the particular theme experiencing styling problems - in contrast, we'd better to make them visible.

Imagine what this change would be for any theme other than Teams - the semantics of the markup is not consistent (why `text`) and, most probably, will cause presentation problems to other theme, as it will style `text` Button differently than Teams theme - maybe this `text` Button will have a look drastically different from the default one.

So, based on this reasoning, suggesting to remove this prop from the docs example

bmdalex: I removed it. We still need to understand if the fact that the `Button`'s border goes outside the element on keyboard focus is a styling problem. We need to understand if placing `Button` components next to each other is a valid scenario for layouting as shown in: https://codesandbox.io/s/w2xoqpy23l

If that's true, this is a styling issue that need to be addressed in `Button` style file

fyi @bcalvery

kuzhelov (approved):

PR #1198, opened 2019-04-10
feat(Popup): add `pointing` prop

layershifter:
This PR adds a new `pointing` prop that allows to show a pointer to a trigger.

image

-----

### `pointing`

We have `pointing` prop in `Menu`, used this name for consistency.

#### DOM structure

```
WARNING: `arrow.element` must be child of its popper element!
```

A `pointer` should be a slot of `PopupContent`, otherwise it fill fail. Bootstrap uses PopperJS and the same structure, BTW.

#### `content` slot

As I have to change DOM structure, I added a new slot `content`, but it's really weird to have `PopupContent`.`content` 😮 `inner` may be?


packages/react/src/themes/teams/components/Popup/popupStyles.ts
```diff
@@ -6,11 +6,9 @@ const popupStyles: ComponentSlotStylesInput<PopupProps, PopupVariables> = {
   root: (): ICSSInJSStyle => ({}),
 
   popup: ({ variables }): ICSSInJSStyle => ({
-    zIndex: variables.zIndex,
     position: 'absolute',
     textAlign: 'left',
-    color: variables.contentColor,
-    background: variables.contentBackgroundColor,
```
layershifter, line 13: Moved to `popupContentStyles`...

packages/react/src/themes/teams/components/Popup/popupContentStyles.ts
```diff
@@ -1,20 +1,83 @@
...
+    bottom: {
+      paddingTop: pxToRem(v.pointerHeight),
+    },
+    left: {
+      paddingRight: pxToRem(v.pointerHeight),
+    },
+  }
+  const pointerStyles = {
+    top: {
+      bottom: `${pxToRem(v.pointerHeight / 2)}`,
```
bmdalex: nit: let's compute `pxToRem(v.pointerHeight / 2)` only once

packages/react/src/themes/teams/components/Popup/popupContentVariables.ts
```diff
@@ -1,15 +1,27 @@
...
-  [key: string]: string | number
-
   borderColor: string
+  borderSize: number
   padding: string
+
+  contentColor: string
+  contentBackgroundColor: string
+
+  pointerHeight: number
```
bmdalex: shouldn't we also accept strings here? same for `pointerWidth`

packages/react/src/themes/teams/components/Popup/popupContentStyles.ts
```diff
@@ -1,20 +1,83 @@
...
 
+const getPointerStyles = (
+  v: PopupContentVariables,
+  popperPlacement?: PopperChildrenProps['placement'],
+) => {
+  const placement = (popperPlacement || '').split('-', 1).pop()
+
+  const rootStyles = {
+    top: {
+      paddingBottom: pxToRem(v.pointerHeight),
```
bmdalex: nit: compute `pxToRem(v.pointerHeight)` only once

bmdalex (commented):
👍 just couple of comments

packages/react/src/themes/teams/components/Popup/popupContentStyles.ts
layershifter: 👍

packages/react/src/themes/teams/components/Popup/popupStyles.ts
miroslavstastny: Talk to @kuzhelov. In #1121 he intentionally moved the styles from `popupContentStyles` to here.

miroslavstastny (commented):
RTL is broken:
image

packages/react/src/themes/teams/components/Popup/popupContentVariables.ts
```diff
@@ -1,15 +1,33 @@
...
 
 export interface PopupContentVariables {
-  [key: string]: string | number
-
   borderColor: string
+  borderRadius: string
+  borderSize: string
   padding: string
+
+  contentColor: string
```
miroslavstastny: - there are `contentColor` and `contentBackgroundColor` left in `popupVariables` as well and are not used in `popupStyles`
- `popupVariables`in teams-dark is setting `color` which is not used
- if we move variables from `popup` to `popupContent` it is a breaking change

miroslavstastny: `contentColor` and `contentBackgroundColor` are now variables of `PopupContent`. Is the **content** in the variables' names still meaningful?

docs/src/examples/components/Popup/Types/PopupExamplePointing.shorthand.tsx
```diff
@@ -0,0 +1,48 @@
+import * as React from 'react'
+import { Button, Grid, Popup } from '@stardust-ui/react'
+
+const PopupWithButton = props => (
+  <Popup
+    align={props.align}
+    content="A popup with a pointer."
```
miroslavstastny: For the other popup example, we intentionally use multiline content to clearly show the difference between `top`/`center`/`bottom` - it would be great to have the same here as well.

miroslavstastny (approved):

packages/react/src/themes/teams/components/Popup/popupStyles.ts
kuzhelov: these styles should be left here, as we popup expects _any_ content to be rendered with this set of styles, not just `PopupContent`. Also, please, ensure that `PopupContent` just CSS-inherits this set of colors.

These two requirements are necessary for introducing correct semantics around these styles, as well as for expected effects of theme switching.

FYI #1121

kuzhelov (changes requested):
please, take a look on this comment about styles: https://github.com/stardust-ui/react/pull/1198/files#r273879017

kuzhelov (approved):

PR #1199, opened 2019-04-10
fix(Toolbar): do not close submenu and navigate next item when press Left/Right arrow keys

### LINKED ISSUE #1064: Menu/Toolbar Left and Right arrow should not activate prev/next parent when focus is in the toolbar submenu
## Bug Report

Menu/Toolbar Left and Right arrow should not activate prev/next parent when focus is in the toolbar submenu.

This is expected only in Menu, not in the toolbar

### Steps
Open Toolbar example in the Menu docs page. Activate submenu, press left arrow

### Expected Result
Nothing happens

### Actual Result
Previous toolbar button is focused

### Version
0.23.1

### Testcase
https://stardust-ui.github.io/react/components/menu#usages-toolbar

sophieH29:
fixes https://github.com/stardust-ui/react/issues/1064

Menu/Toolbar Left and Right arrow should not activate prev/next parent when the focus is in the toolbar submenu. This is expected only in Menu, not in the toolbar


packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -342,31 +347,40 @@ class MenuItem extends AutoControlledComponent<ReactProps<MenuItemProps>, MenuIt
...
+    const { menu } = this.props
     const { menuOpen } = this.state
-    if (menu && menuOpen) {
-      this.trySetMenuOpen(false, e, () => {
-        if (!inSubmenu && (!focusNextParent || this.props.vertical)) {
-          focusAsync(this.itemRef.current)
-        }
-      })
+
+    return menu && menuOpen
```
silviuaavram: maybe enforce boolean here? if menu is undefined, this will not return false.

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -342,31 +347,40 @@ class MenuItem extends AutoControlledComponent<ReactProps<MenuItemProps>, MenuIt
...
+  }
+
+  private closeAllMenus = (e, focusNextParent: boolean) => {
```
silviuaavram: nit: maybe give types `React.SyntheticEvent` to the `e` here and anywhere else since you in the area.

silviuaavram (approved):

packages/react/src/components/Menu/MenuItem.tsx
sophieH29: Unfortunately, `event` object is passed through accessibility action handlers with a type of `Event` because behaviors are targeted not only React framework

silviuaavram: then make it `Event` :D

packages/react/src/components/Menu/MenuItem.tsx
sophieH29: enforcing typing is good, but I can't come up with a case where it can turn into a problem. Do you have a concrete example on your mind? Just to make sure if I really need to rewrite it to
`return menu && menuOpen ? true : false`

packages/react/src/components/Menu/MenuItem.tsx
bmdalex: cannot find the place where `closeAllMenus` is used with `focusNextParent` set to false. can we remove that branch from code?

packages/react/src/components/Menu/MenuItem.tsx
bmdalex: I'd say it's fine, we care about truthy / nontruthy values only in this case

bmdalex (approved):

packages/react/src/components/Menu/MenuItem.tsx
silviuaavram: I did not look at the workflow, just at the signature saying that it returns a `boolean`. If it's not relevant and it can return the `undefined` safely then leave it, otherwise `return !!(menu && menuOpen)`.

sophieH29: it is safe for that case to have `undefined` however would agree with you on consistency of the returning type. Will apply your suggestion. Thank you!

miroslavstastny (commented):
Technically, it works.

I don't like it much from architecture POV:
Every time I hit left/right arrow, acc behavior calls `doNotNavigateNextParentItem` and it `MenuItem` what decides whether to stop propagation of the event based on its submenu state.

If I consider accessibility a separate encapsulated package, this does not work. How should I, as a component developer, know when not to navigate? We can rename the handler to something like `doNotNavigateNextParentItemIfYourSubmenuIsOpen` but not sure it makes the API contract much better.

sophieH29:
> 
> Every time I hit left/right arrow, acc behavior calls `doNotNavigateNextParentItem` and it `MenuItem` what decides whether to stop propagation of the event based on its submenu state.
> 

Thank you for the comment.
`doNotNavigateNextParentItem` is only called if specified in behavior. Currently specified in `toolbarButtonBehaavior`. Basically, you say "I don't want to navigate next parent item when I hit right/left arrow on submenu toolbar button"
Agree, that it looks messy and in an unnatural way. Would be happy to hear any ideas on it.

Also, what can be changed is to move check if submenu is opened to behavior
```
wrapper: {
doNotNavigateNextParentItem: props.menu && props.menuOpen && [
            { keyCode: keyboardKey.ArrowRight },
            { keyCode: keyboardKey.ArrowLeft },
          ]
}
```

miroslavstastny:
> Also, what can be changed is to move check if submenu is opened to behavior
> 
> ```
> wrapper: {
> doNotNavigateNextParentItem: props.menu && props.menuOpen && [
>             { keyCode: keyboardKey.ArrowRight },
>             { keyCode: keyboardKey.ArrowLeft },
>           ]
> }
> ```

If it is possible, that would be better.

But even the fact that item behavior knows the implementation details of the components (that item component must do something (`stopPropagation`) not to break container component) is strange.
But I don't have any better proposal right now, sorry :-/

miroslavstastny (approved):

PR #65, opened 2018-08-08
fix(Fela): prevent styles crash on invalid CSS property provided

kuzhelov:
# Problem

### TL;DR
When, for some reason, for CSS property provided a value that ends with opening bracket (or, in general, a value with invalid braces sequence), this results in shadowing all the subsequent Fela styles added to the Fela's rules `<style />` DOM element.

This problem shows up quite often while playing with docs, as it is typical situation when not fully-provided variable values are applied.

---------------

Specifically, suppose that we have the following style object defined for our component:

```jsx
const style = (props) => {
   ...
   color: 'rgba('
}
```

When component is rendered, its styles are transformed into class declarations and added to Fela's `<style>` DOM element on the page:

```html
<style>
   .a{ display: block; color: rgb( } .b{ display: inline; color: red } ...
</ style>
```

The problem is that invalid sequence of brackets (that was provided as part of the component's CSS property value) results in browser engine ignoring the rest of the `<style />` element content - so that, essentially, all subsequent classes are like not declared.

```html
<style>
    .a{ display: block; color: rgb(       // everything that goes next is ignored by browser
</ style>
```

This results in absence of styling for components that were relying on these classes:

```html
<!-- oops, styles WON'T be applied -->
<div class="b">
   ...
</div>
```

# Solution

Proposed solution is to introduce plugin that will be responsible for ignoring CSS properties that have improper braces sequence as their value. We can extend this functionality to handle other cases if necessary - till we won't find a better solution (problem will be fixed on the Fela's side, new widely-used plugin will be introduced for handling these cases, etc).

PR #1201, opened 2019-04-10
test(Dropdown): add more unit tests

silviuaavram:
I did not follow any test plan, we don't have any. There are tests for basic things, like `highlightedIndex`, `open`, `value` and some props like `highlightFirstItemOnOpen` or `moveFocusOnTab`.

There are many more to be added, but this one is already a biggie. Will put it for review like this and will add others in a separate PR.

Did not manage to test:
1. open dropdown by Space or Enter. Dropdown does not open.
2. test `moveFocusOnTab`. Focus is not handled when using Tab. We can check if selection is performed but not where the focus is.


kolaps33 (approved):

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -34,22 +34,207 @@ describe('Dropdown', () => {
...
+      onOpenChange.mockReset()
+    })
+
+    it('is false when closed by trigger button click', () => {
+      const wrapper = mountWithProvider(<Dropdown onOpenChange={onOpenChange} items={items} />)
+
+      wrapper
+        .find(`button.${Dropdown.slotClassNames.triggerButton}`)
+        .simulate('click')
+        .simulate('click')
```
layershifter: nit: _Personally I_ don't like chains with `find` because they confusing:

```
wrapper
  .some
  .some
  .simulate() // it was simulated on wrapper???
```

```jsx
const wrapper = mountWithProvider(<Dropdown onOpenChange={onOpenChange} items={items} />)
const button = wrapper.find(`button.${Dropdown.slotClassNames.triggerButton}`)

button.simulate('click')
button.simulate('click')
```

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -316,22 +516,330 @@ describe('Dropdown', () => {
...
+      )
+    })
+
+    it('is changed correctly on arrow down navigation', () => {
+      const wrapper = mountWithProvider(<Dropdown items={items} />)
+      const dropdown = wrapper.find(Dropdown)
+      const itemsList = wrapper.find(`ul.${Dropdown.slotClassNames.itemsList}`)
+
+      wrapper.find(`button.${Dropdown.slotClassNames.triggerButton}`).simulate('click')
+      for (let index = 0; index < 2; index++) {
```
layershifter: Nit: `_.times(2, index) => {}`

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
kuzhelov: support this opinion as well - it would be better to have a dedicated variable with semantically correct name that would aid the readability:

```jsx
wrapper
  ...
  .find('.button-ok')
  .simulate() // what this button is responsible for?
```

vs 

```jsx
const confirmButton = wrapper
  ...
  .find('.button-ok')
  
confirmButton.simulate('click')
```

Just a thought - however, please, feel free to decide by yourself, as this is really a _nit_ :)

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -316,22 +516,330 @@ describe('Dropdown', () => {
...
+      wrapper.find(`button.${Dropdown.slotClassNames.triggerButton}`).simulate('click')
+      wrapper
+        .find(`li.${Dropdown.slotClassNames.item}`)
+        .at(itemSelectedIndex)
+        .simulate('click')
+
+      expect(onSelectedChange).toHaveBeenLastCalledWith(
+        null,
+        expect.objectContaining({
+          value: items[itemSelectedIndex],
```
layershifter, line 637:
```suggestion
          value: 'item2',
```

Nit: It's more human readable

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -316,22 +516,330 @@ describe('Dropdown', () => {
...
+      toggleIndicator.simulate('click')
+      wrapper
+        .find(`li.${Dropdown.slotClassNames.item}`)
+        .at(0)
+        .simulate('click')
+      toggleIndicator.simulate('click')
+      wrapper
+        .find(`li.${Dropdown.slotClassNames.item}`)
+        .at(0)
+        .simulate('click')
```
layershifter:
```diff
-      wrapper
-        .find(`li.${Dropdown.slotClassNames.item}`)
-        .at(0)
-        .simulate('click')
      const firstItem = wrapper.find(`li.${Dropdown.slotClassNames.item}`).at(0)
 
      firstItem.simulate('click')
      toggleIndicator.simulate('click')
      firstItem.simulate('click')
```

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -316,22 +516,330 @@ describe('Dropdown', () => {
...
+        <Dropdown onSelectedChange={onSelectedChange} multiple items={items} search />,
+      )
+      const input = wrapper.find(`input.${DropdownSearchInput.slotClassNames.input}`)
+      const toggleIndicator = wrapper.find(`span.${Dropdown.slotClassNames.toggleIndicator}`)
+
+      toggleIndicator.simulate('click')
+      wrapper
+        .find(`li.${Dropdown.slotClassNames.item}`)
+        .at(0)
+        .simulate('click')
```
layershifter: https://github.com/stardust-ui/react/pull/1201/files#r274853940

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -357,5 +865,160 @@ describe('Dropdown', () => {
...
+        wrapper.find(`input.${DropdownSearchInput.slotClassNames.input}`).getDOMNode(),
+      )
+    })
+  })
+
+  describe('moveFocusOnTab', () => {
+    it('keeps focus on trigger button when not passed', () => {
+      const wrapper = mountWithProvider(
+        <>
+          <Dropdown multiple items={items} />,<div tabIndex={0} id="give-me-focus" />
```
layershifter:
```suggestion
          <Dropdown multiple items={items} />
          <div tabIndex={0} id="give-me-focus" />
```

nit: comma looks very confusing there, is it necessary?

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -34,22 +34,207 @@ describe('Dropdown', () => {
...
+      )
+    })
+
+    it.skip('is null when opened by Enter', () => {
+      const wrapper = mountWithProvider(<Dropdown onOpenChange={onOpenChange} items={items} />)
+
+      wrapper
+        .find(`button.${Dropdown.slotClassNames.triggerButton}`)
+        .simulate('focus')
+        .simulate('keydown', { keyCode: keyboardKey.Enter, key: 'Enter' })
```
layershifter: I was not able to simulate this even in browser, consider to remove these assertions. Should notice that they assert for existing browser behavior...

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -21,7 +21,7 @@ describe('Dropdown', () => {
 
       wrapper.find({ className: Dropdown.slotClassNames.clearIndicator }).simulate('click')
       expect(onSelectedChange).toBeCalledTimes(1)
-      expect(onSelectedChange).toHaveBeenCalledWith(
+      expect(onSelectedChange).toHaveBeenLastCalledWith(
```
layershifter: Nit: I propose to use proper semantics there.
When we have multiple calls, let's use `toHaveBeenLastCalledWith()`, for calls let's use `toHaveBeenCalledWith()`

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
silviuaavram: just a mistake. I will remove this whole suite, since Tab is broken and I cannot consider these tests to be of any value.

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
silviuaavram: will change thanks!

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -34,20 +35,169 @@ describe('Dropdown', () => {
...
+    })
+
+    it('is false when closed by hitting Escape in items list', () => {
+      const wrapper = mountWithProvider(
+        <Dropdown onOpenChange={onOpenChange} multiple items={items} />,
+      )
+      const triggerButton = wrapper.find(`button.${Dropdown.slotClassNames.triggerButton}`)
+
+      triggerButton
+        .simulate('click')
```
mnajdova, line 87: Shouldn't we first navigate to the list items by arrow down and then press escape? Otherwise we are laying about what the test is checking..

silviuaavram: we are already checking that on trigger click, the focus goes to the list, it's a test called `'opens the menu and moves focus to list in selection mode'`

so assuming this works, the focus is already in the list, and we hit Escape to close it.

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -34,20 +35,169 @@ describe('Dropdown', () => {
...
   })
 
+  describe('open', () => {
+    const onOpenChange = jest.fn()
+
+    afterEach(() => {
+      onOpenChange.mockReset()
+    })
+
+    it('is false when closed by trigger button click', () => {
```
layershifter:
```suggestion
    it('is "false" when closed by trigger button click', () => {
```

_NIT_

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -34,20 +35,169 @@ describe('Dropdown', () => {
...
+      triggerButton.simulate('click').simulate('click')
+
+      expect(onOpenChange).toBeCalledTimes(2)
+      expect(onOpenChange).toHaveBeenLastCalledWith(
+        null,
+        expect.objectContaining({
+          open: false,
+        }),
+      )
+    })
```
layershifter, line 58:
```
expect(onOpenChange).toHaveBeenLastCalledWithMatch(null, { open: false })
```

Not related to this PR, but I suggest to introduce a custom matcher for these cases (6 lines vs 1)...

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -323,15 +486,302 @@ describe('Dropdown', () => {
...
+      expect(onSelectedChange).toHaveBeenCalledTimes(2)
+      expect(onSelectedChange).toHaveBeenLastCalledWith(
+        null,
+        expect.objectContaining({
+          value: items[itemSelectedIndex],
+        }),
+      )
+    })
+
+    it('has multiple items at multiple selections if the multiple prop is supplied', () => {
```
layershifter: I don't have a suggestion there, but there are definitely too many `multiple` 🐱

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -323,15 +486,302 @@ describe('Dropdown', () => {
...
+      const itemAtIndex2 = wrapper
+        .find(`li.${Dropdown.slotClassNames.item}`)
+        .at(indexesOfItemsSelected[1])
+      itemAtIndex2.simulate('click')
+
+      expect(onSelectedChange).toHaveBeenCalledTimes(2)
+      expect(onSelectedChange).toHaveBeenLastCalledWith(
+        null,
+        expect.objectContaining({
+          value: [items[indexesOfItemsSelected[0]], items[indexesOfItemsSelected[1] + 1]],
```
layershifter:
```suggestion
          value: [items[1], items[4]],
```

Can we use constant values there? It's very hard to read...

layershifter: https://github.com/stardust-ui/react/pull/65#discussion_r208722857

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -34,20 +35,169 @@ describe('Dropdown', () => {
...
+    it('is false when closed by hitting Escape in search input', () => {
+      const wrapper = mountWithProvider(
+        <Dropdown onOpenChange={onOpenChange} search items={items} />,
+      )
+      const searchInput = wrapper.find(`input.${DropdownSearchInput.slotClassNames.input}`)
+
+      searchInput
+        .simulate('click')
+        .simulate('change', { target: { value: 'test' } })
+        .simulate('keydown', { keyCode: keyboardKey.Escape, key: 'Escape' })
```
layershifter, line 69: _Not related to this PR:_
It's strange to me that we should pass `key` always for Downshift: https://github.com/downshift-js/downshift/blob/8f987928e6a0ff4d0d9e6259748bcd243532f0ec/src/utils.js#L252

It will be great to reuse `keyboard-key` there an simplify it...

layershifter (approved):
Please take a look on comments before merge 💃

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
silviuaavram: will change here and for next tests as well thanks!

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
silviuaavram: yeah ...

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
silviuaavram: I will try to take a look in a future PR.

PR #1202, opened 2019-04-10
chore: disable bundle stat collection/save for dev branches

kuzhelov:
This PR removes `Bundle Size` and `Perf Tests` steps from the build process that is carried for dev branches (i.e. all other than `master`). 

This allows to reduce build time by almost 50% (from average of 14-16 mins to 7-8 mins), without compromising any checks


.circleci/config.yml
```diff
@@ -77,17 +77,22 @@ jobs:
...
+              echo 'skip bundle statistics collection'
+            fi
       - run:
           name: Performance Tests
           command: yarn perf
       - run:
           name: Save Statistics to DB
           command: |
-            if [ -n "${STATS_URI}" ]; then
+            if [ -n "${STATS_URI}" ] && [ $CIRCLE_BRANCH == 'master' ]; then
```
miroslavstastny: This step saves both bundle statistics and perf results to the DB (because the currently supports insert operation only, not update)
We should rather update `stats:save` gulp task to make `bundleSize` optional and always save statistics here.

miroslavstastny (changes requested):

.circleci/config.yml
kuzhelov: yes, but should we collect perf statistics for the dev changes that haven't reached master (the reason for these doubts is the same as for bundle stat)? This was the main reason I've introduced the changes this way - and, actually, was thinking about removing perf tests for dev branches as well. What do you think?

miroslavstastny: I can imagine that you can use branch perf charts when working on a perf-related PR.
But you are correct we are not doing that right now and we can revisit this when needed.

But in this case, as I wrote, why do we run perf tests on branch at all? If we decide not to store, then we should not run it.

That being said, I would be afraid of master and PR builds diverging too much for one reason - with the same build, green build on a PR guarantees green master build after the PR is merged. I don't want to get into a situation where PR is green, you merge it and master explodes.

kuzhelov: yes, agree - but currently there are no reasons to be afraid about divergence of _checks_. We can return perf step back to CI chain once there will be some benchmarks introduced, while for now, when it is only about collecting some stat data and store it (without any associated checks being made), we could safely avoid it for dev branches. 

More than that, even with checks being introduced, I would still say that we shouldn't store all the "intermediate" results to stats DB, as it is intended to reflect the story for `master` branch).

Thus, I would suggest to eliminate perf stat collection step in addition to changes that are already introduced.

What are your thoughts on that?

miroslavstastny: Agree

kuzhelov: great - will introduce the necessary changes then, thank you!

miroslavstastny (approved):

PR #1208, opened 2019-04-11
fix(docs): Fix routing for accessibility documentation

sophieH29:
When heading navigation within the article was implemented initially for Accessibility documentation, there were "/" sign added to each link. Since deployed version has a different route `/react/`, these links were rewriting  `/react/` part what caused 404 error when navigating.


mnajdova (approved):
Please add changelog entry

docs/src/views/Accessibility.tsx
```diff
@@ -10,56 +12,56 @@ export default () => (
...
         <ul>
-          <li>{link('Semantic HTML', '/accessibility#semantic-html')}</li>
-          <li>{link('Design Considerations', '/accessibility#design-considerations')}</li>
+          <li>{link('Semantic HTML', 'accessibility#semantic-html')}</li>
+          <li>{link('Design Considerations', 'accessibility#design-considerations')}</li>
         </ul>
       </li>
       <li>
-        {link('Keyboard Navigation', '/accessibility#keyboard-navigation')}
+        {link('Keyboard Navigation', 'accessibility#keyboard-navigation')}
```
kuzhelov, line 28: given that those are the links to the same page, suggest to use just id selector as their content - this would solve all the maintainability headache:

```jsx
{link('Keyboard Navigation', '#keyboard-navigation')}
```

miroslavstastny:
There are anchor links next to each header on the page which are broken as well:
image

sophieH29:
> There are anchor links next to each header on the page which are broken as well:
> image

They're broken across all pages. I think a separate PR should be addressed to fix that issue. Filed a bug https://github.com/stardust-ui/react/issues/1210

docs/src/views/Accessibility.tsx
sophieH29: I've tried that approach, unfortunately, it does not work that way. That query part is not added to the end but rewrites the path

PR #1221, opened 2019-04-15
fix(RadioGroup): Set label font size to smaller in Teams Theme

### LINKED ISSUE #1220: [Teams Theme] Radio button label should be 12px
### Steps
```jsx
const RadioGroupItemExample = () => (
  <RadioGroup
    items={[
      {
        key: '1',
        label: 'Make your choice',
        value: '1',
      },
      {
        key: '2',
        label: 'Another option',
        value: '2',
      },
    ]}
  />
)
```

### Expected Result
Radio labels are 12px

### Actual Result
Radio labels are 14px 💣

### Version
0.27.0

miroslavstastny:
Set font size of Radio button label to `small` in Teams Theme.

Fixes #1220.

- [ ] changelog

### DIFF
--- a/packages/react/src/themes/teams/components/RadioGroup/radioGroupItemVariables.ts
+++ b/packages/react/src/themes/teams/components/RadioGroup/radioGroupItemVariables.ts
@@ -26,7 +26,7 @@ export default (siteVars: any): RadioGroupItemVariables => ({
   focusInnerBorderColor: siteVars.colors.white,
   focusOuterBorderColor: siteVars.colors.black,
 
-  textFontSize: siteVars.fontSizes.medium,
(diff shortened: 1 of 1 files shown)


miroslavstastny:
@notandrew, @codepretty please review

codepretty (commented):
By default radio buttons should be 14px, but for the people picker scenario we want 12px. It is limited to those 2 sizes, but wondering if we should have some sort of prop to determine which size to use?

miroslavstastny:
> By default radio buttons should be 14px, but for the people picker scenario we want 12px

In that case the label should remain `medium` size and be overriden to `small` in people picker by using variables, closing this one.

PR #1223, opened 2019-04-15
feat(Provider): set scrollbar styles

layershifter:
This adds styles for scrollbar to Teams theme. I defined styles in `ProviderBox` instead of `staticStyles` to avoid collisions.

### DIFF
--- a/packages/react/src/lib/handleRef.ts
+++ b/packages/react/src/lib/handleRef.ts
@@ -21,9 +21,7 @@ const handleRef = <N>(ref: React.Ref<N>, node: N) => {
   }
 
   if (ref !== null && typeof ref === 'object') {
-    // @ts-ignore The `current` property is defined as readonly, however it's a valid way because
-    // `ref` is a mutable object
-    ref.current = node
+    ;(ref as React.MutableRefObject<N>).current = node
   }
 }
 
--- a/packages/react/src/lib/handleRef.ts
+++ b/packages/react/src/lib/handleRef.ts
@@ -21,6 +21,7 @@ const handleRef = <N>(ref: React.Ref<N>, node: N) => {
   }
 
   if (ref !== null && typeof ref === 'object') {
+    // The `current` property is defined as readonly, however it's a valid way because `ref` is a mutable object
     ;(ref as React.MutableRefObject<N>).current = node
   }
 }
--- a/packages/react/src/themes/base/components/Provider/providerBoxVariables.ts
+++ b/packages/react/src/themes/base/components/Provider/providerBoxVariables.ts
@@ -1,31 +1,9 @@
-import { pxToRem } from '../../../../lib'
-
 export interface ProviderBoxVariables {
   background: string
   color: string
-
-  scrollbarHeight: string
-  scrollbarWidth: string
-
-  scrollbarThumbBackgroundColor: string
-  scrollbarThumbBorderRadius: string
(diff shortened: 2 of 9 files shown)


packages/react/src/lib/handleRef.ts
```diff
@@ -21,9 +21,8 @@ const handleRef = <N>(ref: React.Ref<N>, node: N) => {
   }
 
   if (ref !== null && typeof ref === 'object') {
-    // @ts-ignore The `current` property is defined as readonly, however it's a valid way because
-    // `ref` is a mutable object
-    ref.current = node
+    // The `current` property is defined as readonly, however it's a valid way because `ref` is a mutable object
+    ;(ref as React.MutableRefObject<N>).current = node
```
mnajdova, line 25: Is the `;` cruft?

packages/react/src/themes/base/components/Provider/providerBoxVariables.ts
```diff
@@ -1,9 +1,31 @@
...
 }
 
 export default (siteVariables): ProviderBoxVariables => ({
   background: siteVariables.bodyBackground,
   color: siteVariables.bodyColor,
+
+  scrollbarHeight: pxToRem(16),
+  scrollbarWidth: pxToRem(16),
+
+  scrollbarThumbBackgroundColor: siteVariables.gray06,
```
mnajdova: Oops, base theme doesn't have `gray06` and `gray04` :S

packages/react/src/lib/handleRef.ts
layershifter: Nope and this is sad: https://github.com/prettier/prettier/issues/4630 😭

packages/react/src/themes/base/components/Provider/providerBoxVariables.ts
layershifter: Definitely, I initially overlooked the path of styles, now they are in correct place.

P.S. Motivation to add them only to Teams theme:
- one of the frequent questions in SUI: "How I can disable custom scrollbars?"
- this feature is not cross-browser

packages/react/src/lib/handleRef.ts
layershifter: image
https://stardust-ui.github.io/react/shorthand-props
The same case, actually 😨

mnajdova: Ok, just to mention I hate this, especially that we are not using `;` at all... Sad day for me :D

packages/react/src/themes/base/components/Provider/providerBoxVariables.ts
mnajdova: I would really love to see this as part of base theme, but I see your points. Let's think about this in the future. The base theme's provider should have custom scrollbar by default, at least in my opinion. Agreed that we should have way of enabling or disabling this behavior as well. The main reason for this is that, I don't see much different ways of defining the scrollbar, more over we can customize it trough it's variables.

packages/react/src/lib/handleRef.ts
mnajdova: Even worse with the provided example, it is used in our guides... :\

mnajdova (approved):

miroslavstastny (commented):
Changes look good, but are not covered by any test. Would it make sense to add a screener test for that?

PR #1224, opened 2019-04-15
 docs(Knobs): use new knobs

layershifter:
Follow up from #1082.

image

Replaces old knobs via new context-based. Also, fixes broken examples in `ExternalLayout`.


packages/react/src/components/List/ListItem.tsx
```diff
@@ -178,15 +178,15 @@ class ListItem extends UIComponent<ReactProps<ListItemProps>, ListItemState> {
       },
     })
 
-    const hasHeaderPart = headerElement || headerMediaElement
+    const hasHeaderPart = !!(headerElement || headerMediaElement)
```
layershifter: Fixes:
```
Warning: Failed prop type: Invalid prop `column` of type `object` supplied to `Flex`, expected `boolean`.
    in Flex (created by ListItem)
    in ListItem (created by List)
    in ul (created by List)
    in List (created by ListExampleSelectable)
```

packages/docs-components/src/knobs/defaultComponents.tsx
```diff
@@ -49,6 +50,34 @@ const KnobSelect: React.FunctionComponent<KnobComponentProps> = props => (
...
+    const hasDecimal = /\.\d/.test(parseValue)
+
+    return hasDecimal ? parseFloat(parseValue) : parseInt(parseValue, 10)
+  }
+  const { defaultValue, unit } = React.useMemo(() => {
+    return {
+      defaultValue: parseValue(props.value),
+      unit: `${props.value}`.replace(`${parseValue(props.value)}`, ''),
+    }
+  }, [])
```
layershifter: Was taken from `KnobScalar` 🦅

packages/react/src/components/List/ListItem.tsx
miroslavstastny: But why is it in this PR? Shouldn't it go to its own with a changelog entry?

packages/docs-components/src/knobs/defaultComponents.tsx
miroslavstastny: Cool, but looks like an unnecessary premature optimisation.

miroslavstastny (commented):
These are thing you haven't touched in this PR but maybe you can think about them as well:

https://github.com/stardust-ui/react/blob/c5f2477c0388816c0fbc11d1d4c9879716322ac7/packages/docs-components/src/knobs/KnobProvider.tsx#L32
-> `_.omit()`

https://github.com/stardust-ui/react/blob/c5f2477c0388816c0fbc11d1d4c9879716322ac7/packages/docs-components/src/knobs/KnobContext.ts#L16
get rid of `any`

docs/src/examples/components/Loader/Usage/LoaderExampleDelay.tsx
```diff
@@ -1,8 +1,10 @@
...
 import * as React from 'react'
+import { useBooleanKnob } from '@stardust-ui/docs-components'
 
-const LoaderExampleDelay: React.FC<{ knobs: { mounted: boolean } }> = ({ knobs }) => (
-  <div style={{ minHeight: 24 }}>{knobs.mounted && <Loader delay={500} />}</div>
-)
+const LoaderExampleDelay = () => {
+  const [mounted] = useBooleanKnob({ name: 'mounted', initialValue: true })
 
+  return <div style={{ minHeight: 24 }}>{mounted && <Loader delay={500} />}</div>
```
miroslavstastny: why `delay` is not a knob?

packages/docs-components/src/knobs/KnobInspector.tsx
```diff
@@ -46,30 +47,37 @@ const getKnobComponents = (
   return components
 }
 
-const KnobInspector: React.FunctionComponent = () => {
+type KnobInspectorProps = {
+  children?: (children: React.ReactElement) => React.ReactElement
+}
+
+const KnobInspector: React.FunctionComponent<KnobInspectorProps> = props => {
```
miroslavstastny, line 54: We should decide whether we prefer `FC` or `FunctionComponent` and be consistent with it

packages/docs-components/src/knobs/KnobInspector.tsx
```diff
@@ -46,30 +47,37 @@ const getKnobComponents = (
...
-            <Label {...knobProps} />
-            <Control {...knobProps}>
-              {React.createElement(knobComponents[knob.type], knobProps)}
-            </Control>
-          </Field>
-        )
-      })}
-    </>
-  )
+  return props.children ? props.children(children) : children
```
miroslavstastny, line 80: `children()` confuses me. Can we use `render` prop?

packages/docs-components/src/knobs/defaultComponents.tsx
layershifter: Actually, no. It computes `defaultValue` & `unit` once because `defaultValue` depends on `props.value` which will be changed during renders

packages/react/src/components/List/ListItem.tsx
layershifter: Will create a separate PR for it

docs/src/examples/components/Loader/Usage/LoaderExampleDelay.tsx
layershifter: Add it there as knob 👍

packages/docs-components/src/knobs/KnobInspector.tsx
layershifter: As I remember we agreed on `FunctionComponent`, @kuzhelov ?

packages/docs-components/src/knobs/KnobInspector.tsx
layershifter: I want to follow React API, they use `children`: https://reactjs.org/docs/context.html#contextconsumer

layershifter:
`_.omit()` will be removed in next release: https://github.com/lodash/lodash/wiki/Roadmap#v500-2019
It also has performance issues

And this package doesn't have `lodash` in deps 😺

layershifter:
> get rid of `any`

I don't want to fill default context with something, it should be filled by `KnobProvider` to avoid direct usages of `KnobContext`. I can do:
```diff
- components: {} as any,
+  components: {
+    KnobControl: () => null,
+    KnobField: () => null,
+    KnobLabel: () => null,
+
+    KnobBoolean: () => null,
+    KnobRange: () => null,
+    KnobSelect: () => null,
+    KnobString: () => null,
+  },
```

But will it better? 🤔

miroslavstastny (approved):

packages/react/src/components/List/ListItem.tsx
layershifter: See #1266.

docs/src/examples/components/Dropdown/State/DropdownExampleLoading.shorthand.tsx
```diff
@@ -1,17 +1,22 @@
...
-    loading={knobs.loading}
-    loadingMessage="Loading..."
-    multiple
-    search
-    items={inputItems}
-    placeholder="Start typing a name"
-  />
-)
+const DropdownExampleLoading = () => {
+  const [loading] = useBooleanKnob({ name: 'loading', initialValue: true })
```
kuzhelov, line 8: for this specific example, just to make it user friendly in terms of user being able to notice effect, I might suggest to introduce `open` knob in addition that will be triggered by default. This will make an effect of `loading` apparent to the client

kuzhelov (approved):

PR #1226, opened 2019-04-16
fix(Alert): themeing updates for the alert component

codepretty:
1. Updated the themeing colors for Alert component.
2. Added new variations for OOF and Urgent banner types and added examples in the docs for these new types.

**Themeing update for Alert component**
image


docs/src/examples/components/Alert/Variations/AlertExampleOOF.shorthand.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+import { Alert } from '@stardust-ui/react'
+
+const AlertExampleInfo = () => (
+  <Alert
+    oof
```
mnajdova: We don't have the` oof` as a prop in the component, so this will end up as a unhandled prop on the generated HTML dom element. This is not something that we want. I agree that these are Teams specific, so as a first step would recommend using boolean variable for this with the same name, instead of property.

packages/react/src/themes/teams/components/Alert/alertStyles.ts
```diff
@@ -18,17 +20,33 @@ const getIntentColorsFromProps = (
 
   if (p.danger) {
     return {
-      color: siteVars.red,
-      backgroundColor: siteVars.red10,
-      borderColor: siteVars.red08,
+      color: v.dangerColor,
```
mnajdova, line 23: +1 for adding variables for all different colors

packages/react/src/themes/teams/components/Alert/alertVariables.ts
```diff
@@ -34,5 +53,24 @@ export default (siteVars: SiteVariablesPrepared): AlertVariables => {
...
     actionColor: undefined,
+
+    focusInnerBorderColor: siteVars.colors.white,
+    focusOuterBorderColor: siteVars.colors.black,
+
+    dangerColor: siteVars.red,
+    dangerBackgroundColor: siteVars.red10,
+    dangerBorderColor: siteVars.red08,
+
+    oofColor: siteVars.orchid, // pink[600] when new color palette PR goes in
```
mnajdova, line 67: Thanks for the comments! :)

docs/src/examples/components/Alert/Variations/AlertExampleOOF.shorthand.tsx
mnajdova: Not sure if these should be publicly visible as an examples though if we move the props to variables, as they are becoming theme specific. Anyway, we can add the examples and show them on the docs, and eve add steps for testing them in different themes. @Bugaa92 @kuzhelov can I have your input on this as well.

codepretty: I updated `oof` and `urgent` to use a variable instead

packages/react/src/themes/teams-high-contrast/components/Alert/alertVariables.ts
```diff
@@ -0,0 +1,47 @@
+import { SiteVariablesPrepared } from '../../../types'
+
+export interface AlertVariables {
```
miroslavstastny: import `AlertVariables` from `teams` theme instead of declaring a new one.

packages/react/src/themes/teams-dark/components/Alert/alertVariables.ts
```diff
@@ -0,0 +1,47 @@
+import { SiteVariablesPrepared } from '../../../types'
+
+export interface AlertVariables {
```
miroslavstastny: import `AlertVariables` from `teams` theme instead of declaring a new one.

packages/react/src/themes/teams-high-contrast/components/Alert/alertVariables.ts
codepretty: Good catch!

miroslavstastny:
> *****Does anyone know why the popout page feature returns a 404 for the OOF variation?**

Maximized view breaks if example filename contains multiple consecutive capital letters (OOF in this case) as a workaround please rename the example files to `AlertExampleOof`.

docs/src/examples/components/Alert/Variations/AlertExampleAction.shorthand.tsx
```diff
@@ -9,7 +9,13 @@ class AlertExampleClosable extends React.Component<{}, { open: boolean }> {
       this.state.open && (
         <Alert
           content="This is a closable alert"
-          action={{ icon: 'close', onClick: () => this.setState({ open: false }) }}
+          action={{
+            icon: {
+              name: 'close',
+              outline: true,
```
bmdalex: @codepretty 
looks like outline always needs to be present, can we make this part of the default theme behavior?

packages/react/src/themes/teams/components/Alert/alertStyles.ts
```diff
@@ -83,12 +101,42 @@ const alertStyles: ComponentSlotStylesInput<AlertProps, AlertVariables> = {
     flexGrow: 1,
   }),
 
-  action: ({ variables: v }): ICSSInJSStyle => ({
+  action: ({ variables: v, props }): ICSSInJSStyle => ({
```
bmdalex: nit: unwritten convention to create `props: p` alias

packages/react/src/themes/teams/components/Alert/alertStyles.ts
```diff
@@ -83,12 +101,42 @@ const alertStyles: ComponentSlotStylesInput<AlertProps, AlertVariables> = {
...
+
+      [`& .${teamsIconClassNames.outline}`]: {
+        display: 'none',
+      },
+    },
+
+    ...(props.isFromKeyboard && {
+      ':focus': {
+        borderColor: v.focusInnerBorderColor,
+        boxShadow: `0 0 0 ${pxToRem(1)} ${v.focusOuterBorderColor}`,
```
bmdalex: shouldn't boxShadow be a design term, hence part of variables or siteVariables?

docs/src/examples/components/Alert/Variations/AlertExampleOOF.shorthand.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+import { Alert } from '@stardust-ui/react'
+
+const AlertExampleInfo = () => (
+  <Alert
+    variables={{ oof: true }}
```
bmdalex: @codepretty I'm not sure you're aware about the Alerts prototype we have at http://localhost:8080/prototype-alerts

It'd be great to update it by:
- deleting code for `BannerAlert` component from `docs/src/prototypes/alerts/AnimatedBannerAlert.tsx` as it's already covered by work in your PR
- using `Alert` directly in `AnimatedBannerAlert` in the same file
- testing the prototypes works as expected

I can help with it and we can do it as follow-up, let me know what you prefer.

docs/src/examples/components/Alert/Variations/AlertExampleOOF.shorthand.tsx
bmdalex: agreed with @mnajdova , `oof` and `urgent` should be variables for now
@codepretty what happened with `critical` variation?

bmdalex (approved):
looks good, few comments 👍

docs/src/examples/components/Alert/Variations/AlertExampleAction.shorthand.tsx
codepretty: ah, good point. yes, i can update the styles to the show the outline version and then swap to filled on hover/focus.

packages/react/src/themes/teams/components/Alert/alertStyles.ts
codepretty: i see that in a lot of places in the code, but i've also heard @levithomason  mention differently. lol. i can update to be consistent though.

docs/src/examples/components/Alert/Variations/AlertExampleOOF.shorthand.tsx
codepretty: oh, i didn't realize we had a prototype for it. do we still need the prototype if the scenarios are covered in the component examples?

bmdalex: @codepretty it still works, you can try it out at http://localhost:8080/prototype-alerts 
I'd say that we don't really need it anymore; the only value it has it that it's integrated with our mocked compose box so you can see how it's "attaching" itself to it to achieve nice styling

packages/react/src/themes/teams/components/Alert/alertStyles.ts
bmdalex: not sure what @levithomason 's thought's on this one are, hmm.. ok, I don't want to bring more confusion, we can definitely leave it like this

packages/react/src/themes/teams/components/Alert/alertStyles.ts
codepretty: Yes, but I will remove these styles for now until I can test the focus styles with `isFromKeyboard` working

docs/src/examples/components/Alert/Variations/AlertExampleOOF.shorthand.tsx
codepretty: I updated the prototype - we can decide to remove later if we want

codepretty:
> agreed with @mnajdova , oof and urgent should be variables for now
@codepretty what happened with critical variation?

@Bugaa92 `critical` merged into `urgent`

packages/react/src/themes/teams/components/Alert/alertStyles.ts
bmdalex: makes sense, I'll update you once I have a PR

packages/react/src/themes/teams/components/Alert/alertVariables.ts
```diff
@@ -34,5 +55,26 @@ export default (siteVars: SiteVariablesPrepared): AlertVariables => {
 
     actionSize: minHeight,
     actionColor: undefined,
+
+    focusInnerBorderColor: siteVars.colors.white,
+    focusOuterBorderColor: siteVars.colors.black,
+
+    dangerColor: siteVars.red,
```
mnajdova: Can we add comments here about what should this color be with the new palette?

mnajdova (approved):

PR #1227, opened 2019-04-16
feat(docs): add FAQ page

hughreeling:
Adds basic FAQ - Only start - but wanted this with code.

I know contains Teams Specific items (under header 'Teams Specific').  But in no way secret, so ok.


docs/src/views/FAQ.tsx
```diff
@@ -0,0 +1,55 @@
...
+    <Header as="h3" content="Does Stardust support mobile?" />
+    <p>No. Scope is limited to Web / Desktop at present.</p>
+
+    <Header as="h3" content="How can @mixins be used in the Stardust's styles" />
+    <p>
+      As the styles in Stardust are defined using CSS in JS, mixins can be defined as simple
+      functions that can be reused on multiple places.
+    </p>
+
+    <Header as="h3" content="Links and React-Router" />
```
layershifter: I think that this should be extracted to a separate page, you can use SUIR docs: https://github.com/Semantic-Org/Semantic-UI-React/blob/master/docs/src/pages/Augmentation.mdx

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -295,6 +295,13 @@ class Sidebar extends React.Component<any, any> {
         to: '/quick-start',
         styles: menuItemStyles,
       },
+      {
+        key: 'faw',
```
alinais: nit: `key: 'faq'`

docs/src/views/FAQ.tsx
hughreeling: You mean as a page describing the 'as' property?  If so, I agree we should have that. But I had this specific question - so we could duplicate this information in the FAQ.

layershifter: > You mean as a page describing the 'as' property? 

Exactly

docs/src/views/FAQ.tsx
```diff
@@ -0,0 +1,55 @@
+import * as React from 'react'
+import DocPage from '../components/DocPage/DocPage'
+import GuidesNavigationFooter from '../components/GuidesNavigationFooter'
+import CodeSnippet from '../components/CodeSnippet'
+
+import { Header } from '@stardust-ui/react'
+
+export default () => (
+  <DocPage title="FAQ">
```
sophieH29, line 14: would be great to add content at the beginning with links to a specific heading "How can ...".
As a FAQ will grow, it is an efficient way for users to go through the questions and jump to the one they're interested in. An example from Accessibility page:

image

bmdalex, line 14: It's a bit hard to read the raw data in this page so maybe we should try to implement a better structure in the page by defining small stateless components such as:
```jsx
const Category = props => <Header as="h2" {...props} />
const Question = props => <Header as="h3" {...props} />
const Answer = props => <p {...props} />
```

This page will become:

```jsx
<DocPage title="FAQ">
  <Category content="General">
    <Question content="Does Stardust support mobile?" />
    <Answer>No. Scope is limited to Web / Desktop at present.</Answer>
    // ...
  </Category>
  // ...
</DocPage>
```

Just a suggestion, can be done in another iteration

docs/src/views/FAQ.tsx
```diff
@@ -0,0 +1,55 @@
...
+      />
+    </p>
+
+    <Header as="h2" content="Teams Specific" />
+    <Header as="h3" content="Adding Icons to Stardust" />
+
+    <p>
+      Generally Teams should be using SVG icons only. If you need FontAwesome or other icons,
+      process for adding them is described in{' '}
+      <a href="https://github.com/stardust-ui/react/pull/585">Issue 585</a>
```
sophieH29: as it is an external link, better to make it open in another tab, so it won't annoy a user to lose a docs page and then go back.
you can import `link`

```tsx
import { link } from '../utils/helpers'

...
{link(
 'Issue 585',
 'https://github.com/stardust-ui/react/pull/585',
  true, // if 'true', opens in a new tab
)}

```

docs/src/views/FAQ.tsx
bmdalex: Should we phrase this as a question?
```What is the recommended approach for using `Link` components to achieve routing? (React-Router)?```

File: docs/src/views/FAQ.tsx [hunk 5]
```diff
@@ -0,0 +1,55 @@
...
+            primary
+            to={previous.url}
+          />
+      </>
+      `}
+      />
+    </p>
+
+    <Header as="h2" content="Teams Specific" />
+    <Header as="h3" content="Adding Icons to Stardust" />
```
bmdalex: Again, would be better to phrase as question

bmdalex (approved):
looks good 👍  approved with some suggestions

docs/src/views/FAQ.tsx
bmdalex: 💯 for this

bmdalex: Also the visual presentation will need some improvement

### Current State:

Screenshot 2019-04-24 at 12 11 19

### Proposed:

Something still very simple, similar to prototypes:
Screenshot 2019-04-24 at 12 13 42

hughreeling: let's do later when we have enough content to get past one screen ;)

PR #1231, opened 2019-04-16
proto(dropdown mentions): improvements and refactoring

bmdalex:
### Description

This PR sets up the ground for a smooth integration of the customized `Dropdown` we use for @mentions scenario and a text editor by:
- refactoring code to decouple **`Dropdown` logic** from **editor logic**
- adding the ability to emphasize suggestions in `Dropdown` items list by increasing fontWeight (part of #1234)

### DEMO
Screen Recording 2019-04-17 at 19 45 26


docs/src/prototypes/Prototypes.tsx
```diff
@@ -3,27 +3,34 @@ import { Box, Header, Segment } from '@stardust-ui/react'
...
-export const PrototypeSection: React.FC<ComponentPrototypeProps> = props => (
-  <Box style={{ margin: 20 }}>
-    {props.title && <Header as="h1">{props.title}</Header>}
-    {props.children}
-  </Box>
-)
+export const PrototypeSection: React.FC<ComponentPrototypeProps> = props => {
+  const { title: title, children, style, ...rest } = props
+  return (
+    <Box style={{ margin: 20, ...style }} {...rest}>
```
mnajdova: Just curious, why we use style here instead of styles?

docs/src/prototypes/mentions/MentionsDropdown.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import { MentionsContainerProps } from './MentionsEditor'
+
+const MentionsDropdown: React.FunctionComponent<MentionsContainerProps> = props => {
+  const { searchQuery, items, onOpenChange, onSearchQueryChange, onInputKeyDown } = props
+
+  const renderItem = React.useCallback(
+    (Item: React.ReactType, props: DropdownItemProps, boldFont: number) => {
+      const { header = '', ...rest } = props
+      const headerStr = String(header)
```
mnajdova: The header is `ShorthandValue`. Please add check here if it is an object, we are interested in the content form the header's object.

docs/src/prototypes/mentions/MentionsEditor.tsx
```diff
@@ -0,0 +1,130 @@
...
+      e: React.SyntheticEvent,
+      { searchQuery }: MentionsContainerProps,
+    ) => {
+      this.setState({ searchQuery })
+    }
+
+    private handleInputKeyDown = (e: React.KeyboardEvent) => {
+      const keyCode = keyboardKey.getCode(e)
+      switch (keyCode) {
+        case keyboardKey.Backspace: // 8
```
mnajdova: Not for this PR, but it will be great if we can introduce behavior for these checks of the keys, it will be good for people to see how to use behaviors as well.

docs/src/prototypes/mentions/MentionsDropdown.tsx
```diff
@@ -0,0 +1,66 @@
...
+            ),
+          }}
+        />
+      )
+    },
+    [searchQuery],
+  )
+
+  return (
+    <Provider.Consumer
```
layershifter, line 23: As we require `react@16.8` I suggest to reexport this context as hook:
```
const siteVariables = useStardustTheme()
```

docs/src/prototypes/mentions/MentionsDropdown.tsx
```diff
@@ -0,0 +1,66 @@
...
+          {...rest}
+          header={{
+            content: (
+              <span>
+                {headerStr.substring(0, queryStartIndex)}
+                <span style={{ fontWeight: boldFont }}>
+                  {headerStr.substring(queryStartIndex, queryEndIndex)}
+                </span>
+                {headerStr.substring(queryEndIndex)}
+              </span>
```
layershifter:
```jsx
const DropdownItemFilteredHeader = (props) => {
  const { header, queryStartIndex } = props

  // logic

  return condition ? <ListItemHeader {...} /> : <span>{...}</span>
}
```

Why we can't create a custom component there instead of `renderMethod()`?

docs/src/prototypes/mentions/MentionsEditor.tsx
```diff
@@ -0,0 +1,130 @@
...
+const withMentionsEditor = <P extends MentionsContainerProps = MentionsContainerProps>(
+  MentionsContainerComponent: React.ComponentType<MentionsContainerProps>,
+) =>
+  class extends React.Component<P, MentionsEditorState> {
+    private readonly initialState: MentionsEditorState = {
+      open: false,
+      searchQuery: '',
+    }
+
+    private contendEditableRef = React.createRef<HTMLDivElement>()
```
layershifter: nit: remove modifiers and reorder methods, please

docs/src/prototypes/mentions/MentionsEditor.tsx
```diff
@@ -0,0 +1,130 @@
...
+const editorStyle: React.CSSProperties = {
+  backgroundColor: '#eee',
+  borderRadius: '5px',
+  border: '1px dashed grey',
+  padding: '5px',
+  minHeight: '100px',
+  outline: 0,
+}
+
+const withMentionsEditor = <P extends MentionsContainerProps = MentionsContainerProps>(
```
layershifter: Please consider to avoid the HOC pattern and especially combine it with hooks.

You can just use `MentionsDropdown` as component, currently I don't see how it can be reused.

docs/src/prototypes/mentions/MentionsEditor.tsx
bmdalex: I will, just FYI I'm just moving this code and not actually adding the modifiers (they were already there)

docs/src/prototypes/mentions/MentionsEditor.tsx
bmdalex: It can be reused for instance to create the same prototype with Popup: https://github.com/stardust-ui/react/compare/proto/mentions-with-popup?expand=1

@layershifter What is the alternative to HOC pattern if I still want to avoid coupling the editor and the @mentions dropdown?

docs/src/prototypes/mentions/MentionsDropdown.tsx
bmdalex: agreed, do we have any established pattern on where to create hooks?
where would you place `useStardustTheme`?

also, I'd say this should be a separate PR that will address all usages of `Provider.Consumer` in our codebase. @layershifter what do you think?

layershifter: Yep, let's do it in a separate PR

docs/src/prototypes/mentions/MentionsEditor.tsx
layershifter: You can use renderProps pattern in this case:
```
<MentionsEditor>
  {editorProps => <Component {...editorProps} />
</MentionsEditor>
```

docs/src/prototypes/mentions/MentionsDropdown.tsx
bmdalex: @layershifter - plenty of reasons:

- more difficult to customize because we need to create new component with dependency on `searchQuery` and styles (`fontWeight`)

```jsx
const DropdownItemFilteredHeader = (
  props: DropdownItemProps & { searchQuery?: string; fontWeight?: FontWeightProperty },
) => {
  const { fontWeight, searchQuery, header, ...rest } = props

  // logic
  
  return (
    <DropdownItem
      {...rest}
      header={{
        content: (
          <span>
            {header.substring(0, queryStartIndex)}
            <span style={{ fontWeight }}>{header.substring(queryStartIndex, queryEndIndex)}</span>
            {header.substring(queryEndIndex)}
          </span>
        ),
      }}
    />
  )
}
```

- we cannot reuse the prop overrides for Dropdown items (`onRemove`, `onClick`, `onKeydown`) => broken functionality

- broken accessibility

- need to define `key` prop in items:

```jsx
items={items.map((item, index) => (
  <DropdownItemFilteredHeader key={index} {...item} />
))}
```

docs/src/prototypes/Prototypes.tsx
bmdalex: we implemented this component when `styles` was not a popular prop at all (we were advising users to avoid it). Changed to `styles`, thanks!

docs/src/prototypes/mentions/MentionsDropdown.tsx
bmdalex: done, thanks

docs/src/prototypes/mentions/MentionsEditor.tsx
bmdalex: done, thanks

docs/src/prototypes/mentions/MentionsEditor.tsx
bmdalex: done, thanks

docs/src/prototypes/mentions/MentionsEditor.tsx
bmdalex: would be an improvement but I wouldn't spend too much time on it since it's a prototype

mnajdova (approved):

docs/src/prototypes/Prototypes.tsx
```diff
@@ -3,27 +3,34 @@ import { Box, Header, Segment } from '@stardust-ui/react'
...
 }
 
-export const PrototypeSection: React.FC<ComponentPrototypeProps> = props => (
-  <Box style={{ margin: 20 }}>
-    {props.title && <Header as="h1">{props.title}</Header>}
-    {props.children}
-  </Box>
-)
+export const PrototypeSection: React.FC<ComponentPrototypeProps> = props => {
+  const { title: title, children, style, ...rest } = props
```
layershifter:
```suggestion
  const { title, children, styles, ...rest } = props
```

docs/src/prototypes/Prototypes.tsx
```diff
@@ -3,27 +3,34 @@ import { Box, Header, Segment } from '@stardust-ui/react'
...
+  return (
+    <Box styles={{ margin: '20px', ...style }} {...rest}>
```
layershifter:
```suggestion
    <Box styles={{ margin: '20px', ...styles }} {...rest}>
```

docs/src/prototypes/Prototypes.tsx
```diff
@@ -3,27 +3,34 @@ import { Box, Header, Segment } from '@stardust-ui/react'
 
 interface PrototypeSectionProps {
   title?: React.ReactNode
+  style?: React.CSSProperties
```
layershifter:
```suggestion
  styles?: React.CSSProperties
```

docs/src/prototypes/mentions/MentionsDropdown.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { FontWeightProperty } from 'csstype'
+import { Dropdown, DropdownItemProps, Provider } from '@stardust-ui/react'
+
+import { MentionsContainerProps } from './MentionsEditor'
+
+const MentionsDropdown: React.FunctionComponent<MentionsContainerProps> = props => {
+  const { searchQuery, items, onOpenChange, onSearchQueryChange, onInputKeyDown } = props
+
+  const renderItem = React.useCallback(
+    (args: { Item: React.ReactType; props: DropdownItemProps; fontWeight: FontWeightProperty }) =>
```
layershifter: Extract this to type?

docs/src/prototypes/mentions/MentionsEditor.tsx
```diff
@@ -0,0 +1,160 @@
...
+    }
+  }
+
+  const resetStateAndUpdateEditor = (searchQueryArg?: string) => {
+    if (state.open) {
+      dispatch({ type: 'UPDATE_AND_CLOSE', args: searchQueryArg || state.searchQuery })
+    }
+  }
+
+  const tryFocusEditor = () => _.invoke(contendEditableRef.current, 'focus')
```
layershifter: This is used single time, we can omit it...

docs/src/prototypes/mentions/MentionsEditor.tsx
```diff
@@ -0,0 +1,160 @@
...
+  )
+
+  const handleEditorKeyUp = (e: React.KeyboardEvent) => {
+    if (!state.open && e.shiftKey && keyboardKey.getCode(e) === keyboardKey.AtSign) {
+      dispatch({ type: 'OPEN' })
+    }
+  }
+
+  const handleOpenChange = (e: React.SyntheticEvent, { open }: MentionsContainerProps) => {
+    if (!open) {
```
layershifter, line 95: `resetStateAndUpdateEditor()` contains opposite condition...

docs/src/prototypes/mentions/MentionsEditor.tsx
```diff
@@ -0,0 +1,160 @@
...
+      return { ...state, searchQuery: action.args }
+    default:
+      throw new Error(`Action ${action.type} is not supported`)
+  }
+}
+
+const MentionsEditor: React.FunctionComponent<
+  MentionsContainerProps & { render: (props: MentionsContainerProps) => React.ReactNode }
+> = props => {
+  const { render, ...rest } = props
```
layershifter: Please consider use `children` instead of `render` as it done in native React APIs:
https://reactjs.org/docs/render-props.html#using-props-other-than-render
https://reactjs.org/docs/context.html#contextconsumer

docs/src/prototypes/mentions/MentionsEditor.tsx
```diff
@@ -0,0 +1,160 @@
...
+  const contendEditableRef = React.useRef(null)
+  const [state, dispatch] = React.useReducer(stateReducer, {
```
layershifter, line 66:
```suggestion
  const [state, dispatch] = React.useReducer<MentionsEditorState>(stateReducer, {
```

This should work as I remember

docs/src/prototypes/mentions/MentionsEditor.tsx
```diff
@@ -0,0 +1,160 @@
...
+  searchQuery: string
+  selectedItem: string
+}
+
+interface MentionsEditorAction {
+  type: 'OPEN' | 'UPDATE_AND_CLOSE' | 'RESET_UPDATE_FLAG' | 'SET_SEARCH_QUERY'
+  args?: string
+}
+
+const editorStyle: React.CSSProperties = {
```
layershifter, line 31: _Nit_: we can use `createComponent()`

docs/src/prototypes/Prototypes.tsx
```diff
@@ -3,27 +3,34 @@ import { Box, Header, Segment } from '@stardust-ui/react'
...
-    {(props.title || props.description) && (
-      <Segment>
-        {props.title && <Header as="h3">{props.title}</Header>}
-        {props.description && <p>{props.description}</p>}
-      </Segment>
-    )}
-    <Segment>{props.children}</Segment>
-  </Box>
-)
+export const ComponentPrototype: React.FC<ComponentPrototypeProps> = props => {
```
layershifter:
```suggestion
export const ComponentPrototype: React.React.FunctionComponent<ComponentPrototypeProps> = props => {
```
nit

docs/src/prototypes/Prototypes.tsx
```diff
@@ -3,27 +3,34 @@ import { Box, Header, Segment } from '@stardust-ui/react'
...
   description?: React.ReactNode
 }
 
-export const PrototypeSection: React.FC<ComponentPrototypeProps> = props => (
-  <Box style={{ margin: 20 }}>
-    {props.title && <Header as="h1">{props.title}</Header>}
-    {props.children}
-  </Box>
-)
+export const PrototypeSection: React.FC<ComponentPrototypeProps> = props => {
```
layershifter:
```suggestion
export const PrototypeSection: React.React.FunctionComponent<ComponentPrototypeProps> = props => {
```
nit

docs/src/prototypes/mentions/MentionsDropdown.tsx
layershifter: If we will use `items.map(item => { ...item, as: DropdownItemFilteredHeader })` everything should work.

But it will break custom `as`...

docs/src/prototypes/mentions/MentionsEditor.tsx
bmdalex: not really, `resetStateAndUpdateEditor` is checking current `open` flag from `state` object while `handleOpenChange` is looking at what will be the state of `open` flag coming from `onOpenChange` prop

I agree that the check inside `resetStateAndUpdateEditor` is redundant because `onOpenChange` is fired when `open` flag changes (obviously), so we can just call `dispatch` directly, something like:

```typescript
const handleOpenChange = (e: React.SyntheticEvent, { open }: MentionsContainerProps) => {
  if (!open) {
    dispatch({ type: 'UPDATE_AND_CLOSE', args: state.searchQuery })
  }
}
```

is this better?

layershifter: Yep, I vote for the direct call 👍

docs/src/prototypes/mentions/MentionsEditor.tsx
bmdalex: It didn't work and I have no idea how to fix the typing there, but it looks like type inference worked quite well anyway

docs/src/prototypes/mentions/MentionsEditor.tsx
bmdalex: I don't think I understand the reason behind it or the benefits 🤔

docs/src/prototypes/mentions/MentionsDropdown.tsx
bmdalex: @layershifter - can we agree to leave as is then? I still don't see much benefit to using map vs `renderItem`

docs/src/prototypes/Prototypes.tsx
bmdalex: done, thx

docs/src/prototypes/Prototypes.tsx
bmdalex: done, thx

docs/src/prototypes/mentions/MentionsDropdown.tsx
bmdalex: done, thx

docs/src/prototypes/mentions/MentionsEditor.tsx
bmdalex: done, thx

docs/src/prototypes/Prototypes.tsx
bmdalex: done, thx

docs/src/prototypes/Prototypes.tsx
bmdalex: done, thx

docs/src/prototypes/mentions/MentionsEditor.tsx
bmdalex: done, thx

docs/src/prototypes/Prototypes.tsx
bmdalex: done, thx

docs/src/prototypes/mentions/MentionsEditor.tsx
bmdalex: done

PR #1233, opened 2019-04-17
Adding keyboard enter and spacebar functionality for the Tree component

priyankar205:
1) Added open & close subtree functionality (for accessibility) from keyboard.

### DIFF
--- a/packages/react/src/components/Tree/TreeItem.tsx
+++ b/packages/react/src/components/Tree/TreeItem.tsx
@@ -103,7 +103,6 @@ class TreeItem extends UIComponent<ReactProps<TreeItemProps>> {
 
   handleTitleOverrides = (predefinedProps: TreeTitleProps) => ({
     onClick: (e, titleProps) => {
-      e.preventDefault()
       _.invoke(this.props, 'onTitleClick', e, this.props)
       _.invoke(predefinedProps, 'onClick', e, titleProps)
     },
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -103,7 +104,10 @@ class TreeItem extends UIComponent<ReactProps<TreeItemProps>> {
 
   handleTitleOverrides = (predefinedProps: TreeTitleProps) => ({
     onClick: (e, titleProps) => {
-      e.preventDefault()
+      // Only prevent default if the key is spacebar
+      if (keyboardKey.getCode(e) === keyboardKey.Spacebar) {
+        e.preventDefault()
+      }
```
sophieH29: we try to omit using keys explicitly in Stardust components. 
Instead, put `e.preventDefault` in actionHandlers. See comment below

packages/react/src/components/Tree/TreeTitle.tsx
```diff
@@ -56,11 +57,14 @@ class TreeTitle extends UIComponent<ReactProps<TreeTitleProps>> {
...
 
   public static defaultProps = {
-    as: 'a',
-    href: '#',
+    as: 'span',
     accessibility: treeTitleBehavior,
   }
 
+  protected actionHandlers: AccessibilityActionHandlers = {
+    performClick: event => !event.defaultPrevented && this.handleClick(event),
```
sophieH29: You can preventDefault for both `Enter` and `Space` here
```
 performClick: event => {
      event.preventDefault()
      this.handleClick(event)
    },
```

No need to check for `event.dafaultPrevented` here. 
It is done in menu for cases when `MenuItem` is a toolbar button

sophieH29 (approved):

sophieH29 (approved):

PR #1235, opened 2019-04-17
fix(Tree): Using anchor for tree title

priyankar205:


packages/react/src/components/Tree/TreeTitle.tsx
```diff
@@ -57,15 +57,13 @@ class TreeTitle extends UIComponent<ReactProps<TreeTitleProps>> {
...
+    href: '#',
     accessibility: treeTitleBehavior,
   }
 
   protected actionHandlers: AccessibilityActionHandlers = {
-    performClick: e => {
-      e.preventDefault()
-      this.handleClick(e)
-    },
+    performClick: e => this.handleClick(e),
```
layershifter: @jurokapsiar Changes in this PR looks very confusing to me.

```
<Title as='h1' /> => <h1 href='#' />
```

We also will always call `preventDefault` even for other elements.

```
<TreeTitle as={Link} />
```

layershifter: And it actually reverts changes from #1233 as I see.

@priyankar205 please clarify this.

jurokapsiar: We had a long discussion couple of months ago with accessibility testers about the structure of the tree. The challenge is that we cannot fully follow the ARIA pattern as it does not work the way how we need for all screen readers. The structure that was agreed upon was implemented and it was changed in #1233 without sufficient testing. @priyankar205 created this PR to revert and I promised him to take a look at the event propagation.

packages/react/src/components/Tree/TreeTitle.tsx
```diff
@@ -57,15 +57,13 @@ class TreeTitle extends UIComponent<ReactProps<TreeTitleProps>> {
   }
 
   public static defaultProps = {
-    as: 'span',
+    as: 'a',
+    href: '#',
```
jurokapsiar: @priyankar205 please keep { as: 'a' } but remove href. This is similar to our Menu strucutre. I have also tested with screen readers and it works ok

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -103,6 +103,7 @@ class TreeItem extends UIComponent<ReactProps<TreeItemProps>> {
 
   handleTitleOverrides = (predefinedProps: TreeTitleProps) => ({
     onClick: (e, titleProps) => {
+      e.preventDefault()
```
jurokapsiar: please remove this, preventDefault only in the performClick handler

packages/react/src/components/Tree/TreeTitle.tsx
jurokapsiar: @priyankar205 preventDefault should be called here, inside of performClick

jurokapsiar (approved):

jurokapsiar:
changelog not necessary as the information is already provided in the previous push (which has not yet been released)

PR #1237, opened 2019-04-18
fix(dropdown): cleanup add and remove a11y messages

silviuaavram:
Clear the text added in the aria live container, so that virtual cursor will not read it afterwards.

Works only for the text that is added on add and on remove, in multiple selection cases. Should fix our issue for now. The other status text is generated by Downshift, will need to fix it as well in the future.

### DIFF
--- a/packages/react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react/src/components/Dropdown/Dropdown.tsx
@@ -317,6 +317,12 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     }
   }
 
+  a11yStatusTimeout: any
+
+  componentWillUnmount() {
+    clearTimeout(this.a11yStatusTimeout)
+  }
+
   public renderComponent({
     ElementType,
     classes,
@@ -952,9 +958,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
 
     if (getA11ySelectionMessage && getA11ySelectionMessage.onAdd) {
       this.setState({ a11ySelectionStatus: getA11ySelectionMessage.onAdd(item) })
-      setTimeout(() => {
-        this.setState({ a11ySelectionStatus: '' })
-      }, Dropdown.a11yStatusCleanupTime)
+      this.setA11ySelectionMessage()
     }
 
     if (multiple) {
@@ -1040,9 +1044,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
 
     if (getA11ySelectionMessage && getA11ySelectionMessage.onRemove) {
       this.setState({ a11ySelectionStatus: getA11ySelectionMessage.onRemove(poppedItem) })
-      setTimeout(() => {
-        this.setState({ a11ySelectionStatus: '' })
-      }, Dropdown.a11yStatusCleanupTime)
+      this.setA11ySelectionMessage()
     }
 
     this.trySetStateAndInvokeHandler('onSelectedChange', null, { value })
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1035,6 +1040,9 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
 
     if (getA11ySelectionMessage && getA11ySelectionMessage.onRemove) {
       this.setState({ a11ySelectionStatus: getA11ySelectionMessage.onRemove(poppedItem) })
+      setTimeout(() => {
+        this.setState({ a11ySelectionStatus: '' })
+      }, Dropdown.a11yStatusCleanupTime)
```
layershifter: Just note: we should properly cleanup these timeouts.

silviuaavram: like with a debounce or what do you mean?

layershifter:
```
      clearTimeout(this.a11yStatusTimeout)
      this.a11yStatusTimeout = setTimeout(() => {
        this.setState({ a11ySelectionStatus: '' })
      }, Dropdown.a11yStatusCleanupTime)

//

componentWillUnmount() {
  clearTimeout(this.a11yStatusTimeout)
}
```

Or you will randomly have this warning:
```
Warning: Can’t call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
```

Or overlapped state...

https://medium.freecodecamp.org/how-to-work-with-react-the-right-way-to-avoid-some-common-pitfalls-fc9eb5e34d9e

See `Clear Subscriptions` part 👍

silviuaavram: oh cool. will integrate this. thank you!

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -315,6 +317,12 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     }
   }
 
+  a11yStatusTimeout: any
```
layershifter, line 320:
```suggestion
  a11yStatusTimeout: number
```

We have conflicting definitions with `node.setTimeout`, https://github.com/Microsoft/TypeScript/issues/842
But we can easily solve it with usage `window.setTimeout`. Opened for better solutions

/cc @kuzhelov

kuzhelov: actually, would rather leave it as `any`, given that this is used as an opaque ref. Other options have more downsides than this, to my mind.

Speaking of `window`-based one, in particular - this would require us to make refactoring for the entire code base of the app. While it is a step to express the actual state of things we have at the moment, I would rather wait a bit with that, as in the upcoming future we might rework our codebase towards consuming `window` abstraction instead of direct DOM `window` access. Also, think that this is a pretty minor thing to worry about.

Please, folks, share your thoughts on that. @layershifter, @silviuavram

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -788,6 +794,82 @@ describe('Dropdown', () => {
...
+      )
+      const dropdown = wrapper.find(Dropdown)
+      const triggerButton = wrapper.find(`button.${Dropdown.slotClassNames.triggerButton}`)
+
+      triggerButton.simulate('click')
+      const firstItem = wrapper.find(`li.${Dropdown.slotClassNames.item}`).at(0)
+      firstItem.simulate('click')
+
+      jest.runAllTimers()
+      expect(dropdown.state('a11ySelectionStatus')).toBe('')
```
layershifter, line 817: Can we merge these suites with previous?

```js
firstItem.simulate('click')
expect(dropdown.state('a11ySelectionStatus')).toBe('bla bla added')

jest.runAllTimers()
expect(dropdown.state('a11ySelectionStatus')).toBe('')
```

silviuaavram: did, thanks !

layershifter (approved):

PR #1238, opened 2019-04-18
fix(Alert): add `isFromKeyboard` to state

### LINKED ISSUE #1230: Unable to style action button in Alert component
### Steps
I am unable to add focus styles to the Alert action button, because `isFromKeyboard` is not recognized

### Extra info
@jurokapsiar: The problem is that `isFromKeyboard` is not in the state of `Alert`, it is in the state of `Button`. How can she style the button based on its state? Do we need to add the state to Alert or is there a better way?

Sounds like we need to add `isFromKeyboard` to the state to be able to detect and style appropriately.

layershifter:
Fixes #1230.

### DIFF
--- a/packages/react/src/components/Alert/Alert.tsx
+++ b/packages/react/src/components/Alert/Alert.tsx
@@ -1,4 +1,5 @@
 import * as customPropTypes from '@stardust-ui/react-proptypes'
+import * as _ from 'lodash'
 import * as PropTypes from 'prop-types'
 import * as React from 'react'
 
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Alert/Alert.tsx
```diff
@@ -108,6 +130,7 @@ class Alert extends UIComponent<ReactProps<AlertProps>> {
     return (
       <ElementType
         className={classes.root}
+        onFocus={this.handleFocus}
```
mnajdova, line 133: Shouldn't this be added on the action, instead of the root element?

mnajdova: The Alert itself as far as I understand is not focusable

codepretty (approved):
Yay! Thanks so much for fixing this! :)

packages/react/src/components/Alert/Alert.tsx
layershifter: Current points:
- as I see the whole `Alert` is not focusable by [spec](https://www.w3.org/TR/wai-aria-practices/examples/alert/alert.html), so currently only `action` slot can be focused
- `focus` event will be propogated from the `action` slot

If we will have issues with this in future, i.e. the whole `Alert` can be focused we can separate values to `isFromKeyboard` & `isAlertFromKeyboard`.

mnajdova: Make sense, I understand that it will be propagated, but my point was more in a sense that, if the content of the alert has focusable event, it would in the same way behave. But we can anyway use the `':focus'` selector together with the `isFromKeyboard` state value, so it should be ok.

mnajdova (approved):

PR #1239, opened 2019-04-18
feat(doc): Add component perf charts

miroslavstastny:
Adds performance charts for all performance examples in docsite:
image
For now, the charts are available only in dev docsite.

There are some missing features, to be worked on in follow-up PRs:
-  cache data in Local Storage
- optionally fetch more data
- refresh support
- serve static data in public builds
- render perf example only when displayed


docs/src/components/ComponentDoc/PerfChart/PerfChart.tsx
```diff
@@ -0,0 +1,92 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+import { FlexibleXYPlot, XAxis, YAxis, HorizontalGridLines, LineSeries, Crosshair } from 'react-vis'
+
+const ChartTooltip = ({ x, data, ...rest }: { x: number; data: any }) => {
+  return (
+    <Crosshair {...rest} values={[{ x, y: 20 }]}>
+      <div style={{ background: '#555', color: 'white', padding: '.5em' }}>
+        <div>Build:&nbsp;{x}</div>
+        <div>Date:&nbsp;{data.ts}</div>
```
layershifter: image
`Date` is missing :(

docs/src/components/ComponentDoc/ComponentPerfExample/ComponentPerfExample.tsx
```diff
@@ -0,0 +1,88 @@
...
+      <Segment variables={{ padding: 0 }}>
+        <Segment variables={{ boxShadowColor: undefined }}>
+          <ComponentExampleTitle title={props.title} description={props.description} />
+          <ComponentPerfChart perfTestName={perfTestName} />
+        </Segment>
+        <Accordion
+          panels={
+            [
+              {
+                key: '0',
```
layershifter: This `key` is cruft as I remember

docs/src/components/ComponentDoc/ComponentPerfExample/ComponentPerfExample.tsx
```diff
@@ -0,0 +1,88 @@
...
+                  content: 'Show example',
+                  styles: ({ theme }) => {
+                    return {
+                      fontSize: theme.siteVariables.fontSizes.smaller,
+                    }
+                  },
+                },
+                content: {
+                  key: 'c',
+                  content: <ComponentExample {..._.omit(props, 'title', 'description')} />, // FIXME: defer rendering until opened
```
layershifter, line 43: nit, We can use destruction there:

```js
const { title, description, ...rest } = props
```

docs/src/components/ComponentDoc/ComponentPerfExample/ComponentPerfExample.tsx
```diff
@@ -0,0 +1,88 @@
...
+const ComponentPerfExample = props => {
+  // FIXME: find a better way
+  // "components/Divider/Performance/Divider.perf" -> dividerPerfTsx
+  const perfTestName = `${_.camelCase(_.last(props.examplePath.split('/')))}Tsx`
+
+  return (
+    <>
+      <Segment variables={{ padding: 0 }}>
+        <Segment variables={{ boxShadowColor: undefined }}>
+          <ComponentExampleTitle title={props.title} description={props.description} />
```
layershifter:
```suggestion
          <ComponentExampleTitle title={title} description={description} />
```

https://github.com/stardust-ui/react/pull/1239/files#r277560039

docs/src/components/ComponentDoc/ComponentPerfExample/index.ts
```diff
@@ -0,0 +1 @@
+export { default as ComponentPerfExample } from './ComponentPerfExample'
```
layershifter:
```suggestion
export { default } from './ComponentPerfExample'
```

docs/src/components/ComponentDoc/PerfChart/PerfChart.tsx
```diff
@@ -0,0 +1,92 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+import { FlexibleXYPlot, XAxis, YAxis, HorizontalGridLines, LineSeries, Crosshair } from 'react-vis'
+
+const ChartTooltip = ({ x, data, ...rest }: { x: number; data: any }) => {
```
layershifter: We have a rule in ESLint that disallows to have multiple components in one file

docs/src/components/ComponentDoc/PerfChart/PerfChart.tsx
```diff
@@ -0,0 +1,92 @@
...
+  )
+}
+
+/**
+ * Draws a performance chart for all items in perfData.performance.
+ * Shows tooltip with details for selected build on mouse hover.
+ * x-axis is a build number
+ * y-axis is a render time
+ */
+const PerfChart: React.FC<any> = ({ perfData }) => {
```
layershifter: nit: `any` is not cool 🐱

docs/src/components/ComponentDoc/PerfChart/PerfChart.tsx
```diff
@@ -0,0 +1,92 @@
...
+  const availableCharts: string[] = perfData
+    .reduce(
+      (acc, next) => {
+        return Array.from(new Set([...acc, ...Object.keys(next.performance)]))
+      },
+      [] as string[],
+    )
+    .sort()
```
layershifter, line 21: What it does? Can we normalize data in `Provider` that fetches data?

docs/src/components/ComponentDoc/PerfChart/PerfData.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+interface PerfDataProps {
+  render: ({ loading: boolean, error: Error, data: any }) => React.ReactElement
+  filter?: string
+}
+
+/**
+ * Reads data from context, filters them and passes them to render function
+ * @param render
+ * @param filter
```
layershifter:
```suggestion
- * @param filter
```

docs/src/components/ComponentDoc/PerfChart/PerfData.tsx
```diff
@@ -0,0 +1,42 @@
...
+import PerfDataContext from './PerfDataContext'
+
+interface PerfDataProps {
+  render: ({ loading: boolean, error: Error, data: any }) => React.ReactElement
+  filter?: string
+}
+
+/**
+ * Reads data from context, filters them and passes them to render function
+ * @param render
```
layershifter:
```suggestion
- * @param render
```

docs/src/components/ComponentDoc/PerfChart/PerfData.tsx
```diff
@@ -0,0 +1,42 @@
...
+ * @param filter
+ */
+const PerfData: React.FC<PerfDataProps> = ({ render, filter }) => {
```
layershifter: Actually, this component can be a hook:

```
const { loading, error, data } = usePerfData(perfExampleName)
```

docs/src/config.ts
```diff
@@ -0,0 +1,5 @@
+const config = {
+  getStatsUri: 'https://stardust.azurewebsites.net/api/GetPerfStats',
```
layershifter, line 2: We already have a `config.ts` in root, do we really need a separate config?

docs/src/components/ComponentDoc/PerfChart/PerfChart.tsx
layershifter: image

It will be also great to avoid this block in a separate PR...

miroslavstastny: Fixed on service side

docs/src/components/ComponentDoc/PerfChart/PerfChart.tsx
miroslavstastny: Each component is now in its own file.

docs/src/components/ComponentDoc/PerfChart/PerfChart.tsx
miroslavstastny: This creates a list of all charts which have any entry in any of the samples (=all available charts). I think we should not normalize the data anywhere else because the structure which we currently have is useful when walking through the data.

docs/src/config.ts
miroslavstastny: Not sure with this one. I know we have a global config, but this one is docs-specific. Should it really go to the global one?

docs/src/components/ComponentDoc/ComponentPerfExample/ComponentPerfChart.tsx
```diff
@@ -0,0 +1,40 @@
+import { Box, Flex, Text } from 'src/index'
```
layershifter:
```suggestion
import { Box, Flex, Text } from '@stardust-ui/react'
```

Lint rule will be introduced with ESLint

docs/src/components/ComponentDoc/ComponentPerfExample/ComponentPerfExample.tsx
```diff
@@ -0,0 +1,48 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+
+import ComponentExampleTitle from '../ComponentExample/ComponentExampleTitle'
+import { Accordion, Segment } from '@stardust-ui/react'
+import ComponentExample from '../ComponentExample'
+import { ComponentPerfChart } from './ComponentPerfChart'
+
+const ComponentPerfExample = props => {
```
layershifter: Typed props? You can reuse existing from `ComponentExample`...

docs/src/components/ComponentDoc/PerfChart/PerfDataContext.ts
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react'
+
+export type PerfSample = {
+  build: number
+  ts: string
+  performance: {
```
layershifter: _**nit**_
```suggestion
  performance: Record<string,  { actualTime: { median: number }}>
```

docs/src/config.ts
layershifter: Don't have a strong opinion there, but I don't reasonable need for it... You can go with your decision 👍

layershifter:
_nit:_ I am not sure that we need to load data for all builds, can we limit for last 200/300 builds?

layershifter (approved):

miroslavstastny:
> _nit:_ I am not sure that we need to load data for all builds, can we limit for last 200/300 builds?

Server returns last 50 builds and we do not have pagination implemented on client side so this is ok.

layershifter:
Checked, you're right 👍 Build numbers confused me...

PR #1242, opened 2019-04-19
chore(dropdown): add more unit tests

silviuaavram:
PR means to finish the unit tests that at least cover the scenarios with the state object properties.

Also fixed the focused style on list focus. Before, when focus was moved to the items list (in selection mode, when dropdown is open). Now the focus is true also when focus is on the list.


packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -59,6 +59,23 @@ describe('Dropdown', () => {
       )
     })
 
+    it('is "false" when closed by toggle indicator click', () => {
+      const wrapper = mountWithProvider(<Dropdown onOpenChange={onOpenChange} items={items} />)
+      const triggerButton = wrapper.find(`button.${Dropdown.slotClassNames.triggerButton}`)
```
kuzhelov: is it necessary to specify exact DOM elements in the selector? This makes these tests being brittle - for example, if `toggleIndicator` will change its element type from `span` to, say, `i`, it will result in failed test while, in fact, nothing will be broken for this scenario.

This is a general comment to all the tests provided by this PR - can we remove element types from selectors, so that our tests would be more resilient in this aspect?

silviuaavram: From what I remember, it was, as otherwise it would also find the actual React component, not only my DOM element. I can double check, but I remember adding element type specifically for this reason.

silviuaavram: Capture

kuzhelov: you can use `getDOMNode` for that: https://airbnb.io/enzyme/docs/api/ReactWrapper/getDOMNode.html

kuzhelov (approved):
approving to unblock, but it will be necessary for us to create follow-up for this. We need to ensure no DOM element types will be used in selectors

kuzhelov (changes requested):
@silviuavram, please, consider to try this - worked for me. This will allow us to avoid DOM element types being hardcoded

```
// this filters only DOM element components from the selected ones.
// may be introduced as a wrapper
wrapper
  .find(`.${...}`)
  .filterWhere(n => typeof n.type() === 'string')
```

kuzhelov (approved):
agreed to merge after explicit DOM elements' names will be removed from selector expressions in tests

PR #1246, opened 2019-04-22
fix(Chat): update ChatMessage styles in Teams themes

layershifter:
This PR:
- updates styles for `ChatMessage` in Teams, Teams Dark and Teams HC themes
- fixes propTypes warnings in examples

---

### Teams theme

| Before | After |
| - | - |
| image | image |

### Teams Dark theme

| Before | After |
| - | - |
| image | image |

### Teams HC theme 

| Before | After |
| - | - |
| image | image |

### DIFF
--- a/packages/react/src/components/Chat/ChatMessage.tsx
+++ b/packages/react/src/components/Chat/ChatMessage.tsx
@@ -121,7 +121,10 @@ class ChatMessage extends UIComponent<ReactProps<ChatMessageProps>, ChatMessageS
     timestamp: customPropTypes.itemShorthand,
     onBlur: PropTypes.func,
     onFocus: PropTypes.func,
-    reactionGroup: customPropTypes.itemShorthand,
+    reactionGroup: PropTypes.oneOfType([
+      customPropTypes.collectionShorthand,
+      customPropTypes.itemShorthand,
+    ]),
     reactionGroupPosition: PropTypes.oneOf(['start', 'end']),
   }
 
--- a/packages/react/src/themes/teams-dark/components/Chat/chatMessageVariables.ts
+++ b/packages/react/src/themes/teams-dark/components/Chat/chatMessageVariables.ts
@@ -4,7 +4,7 @@ export default (siteVars: any): Partial<ChatMessageVariables> => {
   return {
     backgroundColor: siteVars.gray10,
     backgroundColorMine: '#3B3C54',
-    authorColor: siteVars.gray02,
+    authorColor: siteVars.gray02, // will be gray[250] with new palette
     contentColor: siteVars.colors.white,
     color: siteVars.colors.white,
     contentFocusOutlineColor: siteVars.brand,
--- a/packages/react/src/themes/teams-dark/components/Chat/chatMessageVariables.ts
+++ b/packages/react/src/themes/teams-dark/components/Chat/chatMessageVariables.ts
@@ -4,6 +4,8 @@ export default (siteVars: any): Partial<ChatMessageVariables> => {
   return {
     backgroundColor: siteVars.gray10,
     backgroundColorMine: '#3B3C54',
+    authorColor: siteVars.gray02,
+    contentColor: siteVars.colors.white,
     color: siteVars.colors.white,
     contentFocusOutlineColor: siteVars.brand,
     hasMentionNubbinColor: siteVars.naturalColors.orange[900], // orange[300] when the new palette pr is checked in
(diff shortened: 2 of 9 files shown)


packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
```diff
@@ -37,8 +39,10 @@ export default (siteVars): ChatMessageVariables => ({
   offset: pxToRem(100),
   padding: pxToRem(16),
   authorMarginRight: pxToRem(12),
-  authorFontWeight: siteVars.fontWeightBold,
+  authorColor: siteVars.gray02,
+  authorFontWeight: siteVars.fontWeightRegular,
```
codepretty, line 43: Is this even needed if it's regular?

packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
```diff
@@ -37,8 +39,10 @@ export default (siteVars): ChatMessageVariables => ({
...
   headerMarginBottom: pxToRem(2),
+  contentColor: siteVars.colors.grey[900],
```
codepretty: This hex value is actually slightly wrong, but it should be fixed with the new color palette updates (`#252423` instead of `#252424`. Can you add a comment here so we know that when the color palette updates go in we should be using `gray[750]`?

packages/react/src/themes/teams-dark/components/Chat/chatMessageVariables.ts
```diff
@@ -4,6 +4,8 @@ export default (siteVars: any): Partial<ChatMessageVariables> => {
   return {
     backgroundColor: siteVars.gray10,
     backgroundColorMine: '#3B3C54',
+    authorColor: siteVars.gray02,
```
codepretty: Not sure if it is helpful to @mnajdova , but this will be `gray[250]` with new color palette

packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
```diff
@@ -37,8 +39,10 @@ export default (siteVars): ChatMessageVariables => ({
   offset: pxToRem(100),
   padding: pxToRem(16),
   authorMarginRight: pxToRem(12),
-  authorFontWeight: siteVars.fontWeightBold,
+  authorColor: siteVars.gray02,
```
codepretty: Not sure if it is helpful to @mnajdova , but will be `gray[500]` with new color palette

packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
layershifter: You're right, we don't need it. But we already have this variable and I don't want to remove it because it will be a breaking change...

packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
layershifter: I added comments for these values

packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
layershifter: Replaced with direct value and added a note 👍

codepretty (approved):

packages/react/src/themes/teams/components/Chat/chatMessageVariables.ts
```diff
@@ -37,8 +39,10 @@ export default (siteVars): ChatMessageVariables => ({
   offset: pxToRem(100),
   padding: pxToRem(16),
   authorMarginRight: pxToRem(12),
-  authorFontWeight: siteVars.fontWeightBold,
+  authorColor: siteVars.gray02, // will be gray[500] with new palette
+  authorFontWeight: siteVars.fontWeightRegular,
   headerMarginBottom: pxToRem(2),
+  contentColor: '#252423', // will be gray[750] with new palette
```
codepretty, line 45: There is a `color` variable here too. I wonder if we don't need this variable if the color prop above is set to this value.

layershifter: But `color` is `rgb(64, 64, 64)` that points to `#404040`...

packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -121,7 +121,10 @@ class ChatMessage extends UIComponent<ReactProps<ChatMessageProps>, ChatMessageS
     timestamp: customPropTypes.itemShorthand,
     onBlur: PropTypes.func,
     onFocus: PropTypes.func,
-    reactionGroup: customPropTypes.itemShorthand,
+    reactionGroup: PropTypes.oneOfType([
```
mnajdova, line 124: 👍 My bad for not updating this change. We should update this on other places (Menu is the first one that pops out of my mind), but I will take care of those changes.

mnajdova (approved):

PR #1247, opened 2019-04-22
feat(Input): move styles to Base theme

layershifter:
This PR moves `Input` styles to Base theme.

# BREAKING CHANGES

In `InputVariables` the `inputFocusBorderBottomColor` variables was renamed to `inputFocusBorderColor`. It allows to customize all borders instead of only `bottom`.

### Before

```
inputFocusBorderBottomColor: 'red',
```

### After
```
inputFocusBorderColor: 'transparent transparent red transparent`,
```

### DIFF
--- a/packages/react/src/themes/teams/components/Input/inputVariables.ts
+++ b/packages/react/src/themes/teams/components/Input/inputVariables.ts
@@ -1,41 +1,15 @@
+import { InputVariables } from '../../../base/components/Input/inputVariables'
 import { pxToRem } from '../../../../lib'
 
-export interface InputVariables {
-  backgroundColor: string
-  borderRadius: string
-  borderWidth: string
-  fontColor: string
-  fontSize: string
-  iconColor: string
-  iconPosition: string
-  iconRight: string
-  iconLeft: string
-  inputPaddingWithIconAtStart: string
-  inputPaddingWithIconAtEnd: string
-  inputPadding: string
-  inputFocusBorderBottomColor: string
-  inputFocusBorderRadius: string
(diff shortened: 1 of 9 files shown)


packages/react/src/themes/teams/components/Input/inputVariables.ts
```diff
@@ -1,41 +1,17 @@
...
-  iconPosition: 'absolute',
-  iconRight: pxToRem(10),
   iconColor: siteVars.bodyColor,
-  iconLeft: pxToRem(6),
-  inputPaddingWithIconAtStart: `${pxToRem(7)} ${pxToRem(12)} ${pxToRem(7)} ${pxToRem(24)}`,
-  inputPaddingWithIconAtEnd: `${pxToRem(7)} ${pxToRem(24)} ${pxToRem(7)} ${pxToRem(12)}`,
-
-  inputPadding: `${pxToRem(7)} ${pxToRem(12)}`,
-  inputFocusBorderBottomColor: siteVars.colors.primary[500],
+  inputFocusBorderColor: `transparent transparent ${siteVars.colors.primary[500]} transparent`,
```
mnajdova, line 14: I hate that we are using `input` as part of the variables name, but the `input` is a shorthand of the `Input` component, so I don't see any better way... :(

mnajdova (approved):
LGTM 👍

packages/react/src/themes/teams/components/Input/inputVariables.ts
layershifter: I don't have better ideas there, too :(

PR #1260, opened 2019-04-24
chore(deployment): add dynamic deployment with now.sh

layershifter:
This PR aims to add automatic deployment our doc site on temporary domains.

Details: https://zeit.co/blog/every-push-now
Sample website: https://stardust-react-p6b8lvin2.now.sh

----

# Plans

Full details: https://zeit.co/account/plan
Bandwidth: 100 GB / mo

We don't use serverless, so I assume that bandwidth is single valid limitation. I also don't see any limitations for build count.

# Workflow

On each commit a new build will be created with Github hooks. If a new commit will be pushed before a build will end, it will be cancelled. I.e. we will have only one build per PR.

Each PR will have an URL with published docsite, it will be available in PR comments.

More details: https://zeit.co/github#features


.nowignore
```diff
@@ -0,0 +1,31 @@
...
+.snyk
+codecov.yml
+dangerfile.ts
+jest.config.js
+lerna-debug.log
+CHANGELOG.md
+LICENSE.md
+MANIFESTO.md
+README.md
+yarn-error.log
```
layershifter: This config allow to ignore files for sync.
General process:
- upload files to Now
- build them with Now
- publish

build/gulp/plugins/gulp-doctoc.ts
```diff
@@ -1,10 +1,15 @@
...
+import config from '../../../config'
 import sh from '../sh'
 
+const insideGitRepo = fs.existsSync(config.paths.base('.git'))
+
 export default () =>
   through2.obj((file, enc, done) => {
     sh(`doctoc ${file.path} --github --maxlevel 4`)
-      .then(() => sh(`git add ${file.path}`))
+      .then(() => insideGitRepo && sh(`git add ${file.path}`))
```
layershifter, line 12: We don't have Git repo in Now deployment

layershifter:
We should also install [Now App](https://github.com/apps/now) to get it working in PR builds.

kuzhelov:
@layershifter, could you, please, add information about usage plans that are provided by Zeit team for Now, as well as generally describe the workflow with this tool being introduced - directly to the PR description? Thank you!

bmdalex (approved):
👍 let's do this, try get another approval

mnajdova:
@layershifter can we see as part of the PR how will the comment with the link to the served docs looks like? I think that is the only thing missing in the PR. Other then that looks good.

.gitignore
```diff
@@ -1,4 +1,3 @@
-.awcache
```
layershifter, line 1: We don't use `awesome-typescript-loader` anymore...

PR #1261, opened 2019-04-24
style(eslint): enable `spaced-comment`, `@typescript-eslint/no-unused-vars`

layershifter:
This PR enables following rules:
- `spaced-comment`
- `@typescript-eslint/no-unused-vars`

Please see #600 for more context.


packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1018,20 +1018,20 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
-    _.invoke(predefinedProps, 'onKeyDown', e, DropdownSelectedItemProps)
+    _.invoke(predefinedProps, 'onKeyDown', e, dropdownSelectedItemProps)
   }
 
   private handleSelectedItemRemove(
     e: React.SyntheticEvent,
     item: ShorthandValue,
     predefinedProps: DropdownSelectedItemProps,
-    DropdownSelectedItemProps: DropdownSelectedItemProps,
+    dropdownSelectedItemProps: DropdownSelectedItemProps,
```
miroslavstastny, line 1042: 👍

miroslavstastny (commented):
I don't agree with enabling `no-shadow`.
One of the purposes of scopes is to give you some level of encapsulation. Although it can sometimes lead to a better naming it often results in using misleading or artificial names and makes code more difficult to read.
One example:
image

layershifter:
> I don't agree with enabling `no-shadow`.

I can't agree with you there: yep, sometimes it will require changes in names, but in the most cases you will clearly understand what happens and it will save you from accidental bugs.

Actually, this case is noop and it will be removed in a future PR: https://github.com/stardust-ui/react/pull/1268/files#diff-ad7603568c43fa756c43f87b68b69161R82. However, if I will remove `active` (`isActive`) from [the destruction](https://github.com/stardust-ui/react/pull/1261/files#diff-ad7603568c43fa756c43f87b68b69161R89) the code will still work, but now it has completely different sense (`active` is still boolean, but comes from another scope).

kuzhelov:
speaking of `no-shadow` - I agree with both points expressed, however, I would rather vote for _defer_ to introduce this lint rule for now.

The reason is because its value is not clearly seen at this moment. Ideally, I would want to see real issue (say, in app's logic) that this rule helps to prevent, before introducing it. Also, as a point towards deferring it is that it introduces some friction points in code, without clear value _for our code_ seen at the moment

miroslavstastny (approved):

Issue #1039, opened 2019-03-08
Should be able to easily understand what props each Icon supports


codepretty:
### Problem description
We want the ability to specify what each icon allows. Some icons will only support outline and not filled, some will be have brand colors and others will allow you to change the colors, etc.

### Proposed solution
Each icon should be a separate export so we can explicitly define the supported props, like outline, color, sizes.

PR #1264, opened 2019-04-24
feat(Icons): Adding new callings & meetings icons

codepretty:
New icons added for calling scenarios

image

---

Addresses this issue #1278


alinais (approved):

layershifter:
- `call-park` - nit, `call-parking` because _park is large public garden or area of land used for recreation_
- `call-control-present` - I want to mention that we also have `call-control-present-new`
- `cast` - I suggest `video-cast` or `screen-cast` becase FA has `podcast`
- `pause-thick` should be a variation of `pause`, but we don't have such option yet
- `user-phone` - not, sure, we have there `control`/`screen`, `user` & `phone`
- `close-share-tray` - `call-control-present-new` is opposite to this icon?
- `user-friends` - just `users`?

layershifter (approved):

codepretty:
> * `cast` - I suggest `video-cast` or `screen-cast` becase FA has `podcast`


I like `screen-cast`

> * `call-control-present` - I want to mention that we also have `call-control-present-new`
> * `close-share-tray` - `call-control-present-new` is opposite to this icon?


what i'd really like to do is rename all of these to something more generic. these are all very non-intuitive. I think these would be better named to something like `screen-arrow-up` and `screen-x`, but i'm not sure we should rename icons and cause potential regressions?
image

> * `pause-thick` should be a variation of `pause`, but we don't have such option yet


should we delay adding this?

> * `user-phone` - not, sure, we have there `control`/`screen`, `user` & `phone`


any ideas?? @levithomason 

> * `user-friends` - just `users`?


we have a `teams` icon that i think should really be `users`. i agree that `user-friends` is confusing, it's just what i found FA doing, but not sure that justifies here

layershifter:
> I like `screen-cast`

Only one thing, as I see `screencast` is a single word :) https://en.wikipedia.org/wiki/Screencast

> what i'd really like to do is rename all of these to something more generic.

Let's leave it as it it now.

> should we delay adding this?

Nope, because we don't have any ETA for #1039 :(

Issue #1259, opened 2019-04-24
Extract a common hover/focus style for icons


codepretty:
In multiple components, there are similar hover/focus styles being applied to icons within a hovered/focused container.

### Attachment
![image](https://user-images.githubusercontent.com/828692/56669945-ca67ac00-6666-11e9-93ee-566a465b2f46.png)

### Alert
![image](https://user-images.githubusercontent.com/828692/56669956-cf2c6000-6666-11e9-8356-5b6c84d7cdf5.png)

### Menu
![image](https://user-images.githubusercontent.com/828692/56669975-d6536e00-6666-11e9-9dc8-38c099b86471.png)

---

The code all looks like this 
```
[`& .${teamsIconClassNames.filled}`]: {
  display: 'none',
},

[`& .${teamsIconClassNames.outline}`]: {
  display: 'block',
},

':hover': {
  [`& .${teamsIconClassNames.filled}`]: {
    display: 'block',
  },

  [`& .${teamsIconClassNames.outline}`]: {
    display: 'none',
  },
},

...(p.isFromKeyboard && {
  ':focus': {
    [`& .${teamsIconClassNames.filled}`]: {
      display: 'block',
    },

    [`& .${teamsIconClassNames.outline}`]: {
      display: 'none',
    },
  }
}),
```

PR #1269, opened 2019-04-25
feat(border): focus styles mechanism

bmdalex:
### BREAKING CHANGES MIGRATION STEPS

The following component variables have been removed from:
- **`Attachment`**:
`focusInnerBorderColor`, `focusOuterBorderColor`

- **`Button`**:
`circularRadius`, `colorActive`, `borderColorActive`, `borderColorFocus`, `borderColorFocusIndicator`, `borderColorDisabled`, `borderWidth`, `primaryColorActive`, `primaryBorderColorActive`, `primaryBorderColorHover`, `primaryBorderColorFocus`, `primaryBorderColorFocusIndicator`, `primaryBorderWidth`, `primaryCircularBorderColorFocusIndicator `, `circularBorderColorActive`, `circularBorderColorFocusIndicator`, `borderRadiusFocused`

- **`RadioGroupItem`**:
`focusInnerBorderColor`, `focusOuterBorderColor`

**These variables are deprecated and should not be used.**

1. For the variables related to borders, Teams now have a unified way of displaying it that is consistent with the whole Teams theme => border styles should be unified, no custom border appearances.

2. For the other variables, they were either used in dead style objects or they do not match redline requirements => NO IMPACT

### Description

This PR covers:

1. implementation of utility function that will be used for styling components on focus state (from keyboard): **`getBorderFocusStyles`** from **`getBorderFocusStyles.ts`**

2. integration of **`getBorderFocusStyles`** for border focus styles for following components:
  - `Button`
  - `Attachment`
  - `RadioGroupItem`

#### Regular Theme
Screenshot 2019-05-06 at 11 40 42

#### Dark Theme
Screenshot 2019-05-06 at 11 41 01

#### High Contrast Theme
Screenshot 2019-05-06 at 11 41 23

**We will adopt this mechanism for all other components in separate PR.**

3. simplifying component variables and writing styles for border by:
  - moving generic border variables from component variables to `siteVariables.ts`
  - leveraging **`getBorderFocusStyles`** and generic border styles from `siteVariables.ts`

4. fix for `Button` outline styles overlapping issue in `Grid`:

### BEFORE

Screenshot 2019-04-25 at 15 07 36

### AFTER

Screenshot 2019-04-25 at 15 09 35

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,11 +16,11 @@
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-
 ### BREAKING CHANGES
 - Rename `context` prop to `mountNode` in `PortalInner` @layershifter ([#1288](https://github.com/stardust-ui/react/pull/1288))
 - Updated Teams' theme color palette values, removed color related site variables @mnajdova ([#1069](https://github.com/stardust-ui/react/pull/1069))
 - Remove `defaultTarget` prop in `Popup` component @layershifter ([#1153](https://github.com/stardust-ui/react/pull/1153))
+- Add focus border styling mechanism @Bugaa92 in Teams theme ([#1269](https://github.com/stardust-ui/react/pull/1269))
 
 ### Features
 - Add default child a11y behavior to `Menu` related behaviors @silviuavram ([#1282](https://github.com/stardust-ui/react/pull/1282))
--- a/packages/react/src/components/Attachment/Attachment.tsx
+++ b/packages/react/src/components/Attachment/Attachment.tsx
@@ -104,13 +104,10 @@ class Attachment extends UIComponent<ReactProps<AttachmentProps>, AttachmentStat
         {...unhandledProps}
         {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.root, unhandledProps)}
       >
-        {icon && (
-          <div className={classes.icon}>
-            {Icon.create(icon, {
-              defaultProps: { size: 'larger' },
-            })}
-          </div>
-        )}
+        {icon &&
+          Icon.create(icon, {
+            defaultProps: { size: 'larger', styles: styles.icon },
+          })}
         {(header || description) && (
           <div className={classes.content}>
             {Text.create(header, {
@@ -122,16 +119,13 @@ class Attachment extends UIComponent<ReactProps<AttachmentProps>, AttachmentStat
             })}
           </div>
         )}
-        {action && (
-          <div className={classes.action}>
-            {Button.create(action, {
-              defaultProps: { iconOnly: true, text: true },
-            })}
-          </div>
-        )}
+        {action &&
+          Button.create(action, {
+            defaultProps: { iconOnly: true, text: true, styles: styles.action },
+          })}
         {!_.isNil(progress) &&
           Box.create('', {
-            defaultProps: { className: classes.progress },
+            defaultProps: { styles: styles.progress },
           })}
       </ElementType>
     )
--- a/packages/react/src/themes/teams-dark/siteVariables.ts
+++ b/packages/react/src/themes/teams-dark/siteVariables.ts
@@ -2,6 +2,12 @@ import { colors } from '../teams/siteVariables'
 export { colorScheme } from './colors'
 
 //
+// BORDER STYLES
+//
+export const focusInnerBorderColor = colors.black
+export const focusOuterBorderColor = colors.white
+
+//
 // SHADOW LEVELS
 //
 export const shadowLevel1 = '0 .2rem .4rem -.075rem rgba(0, 0, 0, .25)'
--- a/packages/react/src/themes/teams-high-contrast/components/Alert/alertStyles.ts
+++ b/packages/react/src/themes/teams-high-contrast/components/Alert/alertStyles.ts
@@ -1,10 +1,10 @@
 import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
 import { AlertProps } from '../../../../components/Alert/Alert'
 import { AlertHighContrastVariables } from './alertVariables'
-import { pxToRem } from '../../../../lib'
+import getBorderFocusStyles from 'src/themes/teams/getBorderFocusStyles'
 
 const alertStyles: ComponentSlotStylesInput<AlertProps, AlertHighContrastVariables> = {
-  action: ({ variables: v, props: p }): ICSSInJSStyle => ({
+  action: ({ variables: v, props: p, theme: { siteVariables } }): ICSSInJSStyle => ({
     ...(p.isFromKeyboard && {
       ':focus': {
         backgroundColor: v.focusBackgroundColor,
@@ -18,16 +18,8 @@ const alertStyles: ComponentSlotStylesInput<AlertProps, AlertHighContrastVariabl
     ':hover': {
       backgroundColor: v.hoverBackgroundColor,
 
-      ':before': {
-        content: '""',
-        position: 'absolute',
-        top: 0,
-        right: 0,
-        bottom: 0,
-        left: 0,
-        border: `${pxToRem(2)} solid ${v.hoverBorderColor}`,
-        borderRadius: pxToRem(2),
-      },
+      // TODO: consider creating dedicated method for border styles on hover
+      ...getBorderFocusStyles({ siteVariables, isFromKeyboard: true })[':focus'],
     },
   }),
 }
--- a/packages/react/src/themes/teams-high-contrast/components/Alert/alertStyles.ts
+++ b/packages/react/src/themes/teams-high-contrast/components/Alert/alertStyles.ts
@@ -1,7 +1,7 @@
 import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
 import { AlertProps } from '../../../../components/Alert/Alert'
 import { AlertHighContrastVariables } from './alertVariables'
-import getBorderFocusStyles from 'src/themes/teams/getBorderFocusStyles'
+import getBorderFocusStyles from '../../../teams/getBorderFocusStyles'
 
 const alertStyles: ComponentSlotStylesInput<AlertProps, AlertHighContrastVariables> = {
   action: ({ variables: v, props: p, theme: { siteVariables } }): ICSSInJSStyle => ({
--- a/packages/react/src/themes/teams-high-contrast/components/Attachment/attachmentStyles.ts
+++ b/packages/react/src/themes/teams-high-contrast/components/Attachment/attachmentStyles.ts
@@ -33,12 +33,6 @@ const attachmentStyles: ComponentSlotStylesInput<AttachmentProps, AttachmentVari
       },
     }),
   }),
-
-  action: ({ variables: v }): ICSSInJSStyle => ({
-    ':hover': {
-      outline: `1px solid ${v.textColorHover}`,
-    },
-  }),
 }
 
 export default attachmentStyles
--- a/packages/react/src/themes/teams-high-contrast/components/Button/buttonVariables.ts
+++ b/packages/react/src/themes/teams-high-contrast/components/Button/buttonVariables.ts
@@ -10,13 +10,9 @@ export default (siteVars: any): Partial<ButtonVariables> => {
     backgroundColorFocus: siteVars.accessibleYellow,
     backgroundColorDisabled: siteVars.accessibleGreen,
     borderColor: siteVars.colors.white,
-    borderColorActive: 'transparent',
     borderColorHover: 'transparent',
-    borderColorFocus: siteVars.colors.black,
-    borderColorFocusIndicator: siteVars.colors.white,
 
     primaryColor: siteVars.colors.white,
-    primaryColorActive: siteVars.colors.black,
     primaryColorHover: siteVars.colors.black,
     primaryColorFocus: siteVars.colors.black,
     primaryBackgroundColor: siteVars.colors.black,
@@ -24,11 +20,6 @@ export default (siteVars: any): Partial<ButtonVariables> => {
     primaryBackgroundColorHover: siteVars.accessibleYellow,
     primaryBackgroundColorFocus: siteVars.accessibleYellow,
     primaryBorderColor: siteVars.colors.white,
-    primaryBorderColorFocus: siteVars.colors.black,
-    primaryBorderColorFocusIndicator: siteVars.colors.white,
-    primaryBorderWidth: 2,
-
-    primaryCircularBorderColorFocusIndicator: siteVars.colors.black,
 
     circularColor: siteVars.colors.white,
     circularColorActive: siteVars.colors.black,
@@ -37,10 +28,8 @@ export default (siteVars: any): Partial<ButtonVariables> => {
     circularBackgroundColorHover: siteVars.accessibleYellow,
     circularBackgroundColorFocus: siteVars.accessibleYellow,
     circularBorderColor: siteVars.colors.white,
-    circularBorderColorActive: siteVars.colors.white,
     circularBorderColorHover: siteVars.colors.white,
     circularBorderColorFocus: siteVars.colors.white,
-    circularBorderColorFocusIndicator: siteVars.colors.black,
 
     textColor: siteVars.accessibleYellow,
     textColorHover: siteVars.accessibleYellow,
--- a/packages/react/src/themes/teams-high-contrast/siteVariables.ts
+++ b/packages/react/src/themes/teams-high-contrast/siteVariables.ts
@@ -2,6 +2,12 @@ export * from './colors'
 import { colors } from '../teams/siteVariables'
 
 //
+// BORDER STYLES
+//
+export const focusInnerBorderColor = colors.black
+export const focusOuterBorderColor = accessibleCyan
+
+//
 // SEMANTIC ASSIGNMENTS
 //
 export const bodyBackground = colors.black
--- a/packages/react/src/themes/teams-high-contrast/siteVariables.ts
+++ b/packages/react/src/themes/teams-high-contrast/siteVariables.ts
@@ -1,5 +1,6 @@
-export * from './colors'
 import { colors } from '../teams/siteVariables'
+import { accessibleCyan } from './colors'
+export * from './colors'
 
 //
 // BORDER STYLES
--- a/packages/react/src/themes/teams/components/Alert/alertStyles.ts
+++ b/packages/react/src/themes/teams/components/Alert/alertStyles.ts
@@ -2,8 +2,8 @@ import * as React from 'react'
 import { ComponentSlotStylesInput, ICSSInJSStyle, SiteVariablesPrepared } from '../../../types'
 import { AlertProps } from '../../../../components/Alert/Alert'
 import { AlertVariables } from './alertVariables'
-import { teamsIconClassNames } from '../Icon/svg'
-import { pxToRem } from '../../../../lib'
+import getBorderFocusStyles from '../../getBorderFocusStyles'
+import getIconFillOrOutlineStyles from '../../getIconFillOrOutlineStyles'
 
 const getIntentColorsFromProps = (
   p: AlertProps,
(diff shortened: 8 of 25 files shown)


packages/react/src/themes/teams/siteVariables.ts
```diff
@@ -25,6 +25,14 @@ export const red = colors.red[900] // no mapping color $app-red
 export const red08 = '#F3D6DB' // no mapping color $app-red-08
 export const red10 = '#FCF4F6' // no mapping color $app-red-10
 
+//
+// BORDER STYLES
+//
+export const borderWidth = '1px'
```
mnajdova, line 12: Can we be more specific with the name, by replacing the `borderWidth` with `focusBorderWidth` or `borderFocusWidth` and the `borderRadius` with `focusBorderRadius` or `borderFocusRadius`?

mnajdova: Or if these represent the values for all borders, not just focused one, then it is ok :)

packages/react/src/themes/teams-high-contrast/siteVariables.ts
```diff
@@ -28,6 +28,12 @@ export const errorStatusTextColor = black
 export const unknownStatusBackgroundColor = colors.white
 export const unknownStatusTextColor = black
 
+//
+// BORDER STYLES
+//
+export const focusInnerBorderColor = colors.black
```
mnajdova, line 8: Hmm, I thought the focused border looks exactly the same in light and dark contrast. BTW why we don't have mechanism for high contrast theme? I think it is suppose to be accessibleCyan in that case. @codepretty ?

codepretty: High contrast has different requirements. we're supposed to be mapping to what Windows does by default which isn't always clear. Depending on the context, some things will have a background. i'm not sure how to standardize high contrast, but maybe a pattern will emerge as we go along.

packages/react/src/themes/teams-high-contrast/components/Alert/alertStyles.ts
```diff
@@ -18,16 +18,8 @@ const alertStyles: ComponentSlotStylesInput<AlertProps, AlertHighContrastVariabl
...
-        content: '""',
-        position: 'absolute',
-        top: 0,
-        right: 0,
-        bottom: 0,
-        left: 0,
-        border: `${pxToRem(2)} solid ${v.hoverBorderColor}`,
-        borderRadius: pxToRem(2),
-      },
+      // TODO: consider creating dedicated method for border styles on hover
```
bmdalex, line 21: fyi @codepretty do we expect to have similar behavior for `:hover` border styles so that we can create something similar to `getBorderFocusStyles`?

packages/react/src/themes/teams/siteVariables.ts
bmdalex: they are values for all borders, indeed, resolving

packages/react/src/themes/teams-high-contrast/siteVariables.ts
bmdalex: I think the current design is:
- Regular theme: `focusInnerBorderColor: white`, `focusOuterBorderColor: black`
- Dark theme: `focusInnerBorderColor: black`, `focusOuterBorderColor: white` (inverted basically)
- HC theme: same as Dark

@codepretty to confirm

packages/react/src/themes/teams-dark/siteVariables.ts
```diff
@@ -32,6 +32,12 @@ export const red = '#d74654' // no mapping color
 export const red08 = '#4f232b' // no mapping color
 export const green04 = naturalColors.lightGreen[900]
 
+//
+// BORDER STYLES
+//
+export const focusInnerBorderColor = colors.black
```
codepretty, line 7: these don't need to swap for dark theme. the reason we do a double border of white and black is so we can keep the exact same style for both themes

bcalvery:
The focus border as originally implemented appears around the item, but in this PR it looks like it is within the item. Has the design changed to be more easily implemented?

Before:
image
After:
image

This is pretty subtle in normal themes, but in contrast the "focus border" was much thicker and would look distinctly different if drawn "inside" the button.

packages/react/src/themes/teams-high-contrast/siteVariables.ts
codepretty: Confirmed today that current design is:

Regular theme: focusInnerBorderColor: white, focusOuterBorderColor: black
Dark theme: focusInnerBorderColor: white, focusOuterBorderColor: black
HC theme: focusInnerBorderColor: cyan, focusOuterBorderColor: black

packages/react/src/themes/teams-dark/siteVariables.ts
bmdalex: done

bmdalex:
@bcalvery these changes are all part of the new design indeed

packages/react/src/themes/teams-high-contrast/siteVariables.ts
bmdalex: Made the changes. Here are screenshots for `Button`:

#### Regular Theme
Screenshot 2019-05-06 at 11 40 42

#### Dark Theme
Screenshot 2019-05-06 at 11 41 01

#### High Contrast Theme
Screenshot 2019-05-06 at 11 41 23

However, as you can see, items seem smaller in Dark and HC theme because the white border is more visible:
#### in Regular Theme:
we have **light background -> dark border -> light border -> content**

#### in Dark and HC Theme:
we have **dark background -> dark border\* -> light border -> content**
\* - dark background next to dark border => component seems smaller 
Screen Recording 2019-05-06 at 11 52 54


I believe it'd be better if we swapped colors for Dark and HC theme and have it like:
1. Regular theme: `focusInnerBorderColor: white`, `focusOuterBorderColor: black`
2. Dark theme: `focusInnerBorderColor: black`, `focusOuterBorderColor: white`
3. HC theme: `focusInnerBorderColor: black`, `focusOuterBorderColor: cyan`

@codepretty what do you think?

packages/react/src/themes/teams-high-contrast/components/Alert/alertStyles.ts
codepretty: i believe so. there are a lot of hoverable elements that in hc will get a yellow background. some will need a border/outline too (eg icon button in attachment)

packages/react/src/themes/teams-high-contrast/components/Button/buttonVariables.ts
```diff
@@ -10,25 +10,16 @@ export default (siteVars: any): Partial<ButtonVariables> => {
...
-    borderColorActive: 'transparent',
     borderColorHover: 'transparent',
-    borderColorFocus: siteVars.black,
-    borderColorFocusIndicator: siteVars.white,
 
     primaryColor: siteVars.white,
-    primaryColorActive: siteVars.black,
     primaryColorHover: siteVars.black,
     primaryColorFocus: siteVars.black,
     primaryBackgroundColor: siteVars.black,
```
codepretty: I'm not sure if this is the right value to change, but I believe the primary rest background color for buttons in Teams hc should be white
image

packages/react/src/themes/teams-high-contrast/components/Button/buttonVariables.ts
```diff
@@ -10,25 +10,16 @@ export default (siteVars: any): Partial<ButtonVariables> => {
...
     primaryBackgroundColorActive: siteVars.accessibleYellow,
     primaryBackgroundColorHover: siteVars.accessibleYellow,
     primaryBackgroundColorFocus: siteVars.accessibleYellow,
```
codepretty, line 21: `primaryBackgroundColorFocus` should be cyan. In general to keep things consistent for hc, elements should get

- Focused background - cyan
- Hover background - yellow
- 	Focus/hover foreground - black
- 	Focus border - 1px cyan border, 1px black border

packages/react/src/themes/teams-high-contrast/siteVariables.ts
bmdalex: Had a conversation with @codepretty and we decided to invert colors

packages/react/src/themes/teams-high-contrast/components/Alert/alertStyles.ts
bmdalex: @codepretty would be great to create a list of all of these in #1259 , already started

packages/react/src/themes/teams-high-contrast/components/Button/buttonVariables.ts
bmdalex: resolving this and tracking separately in #1300

packages/react/src/themes/teams-high-contrast/components/Button/buttonVariables.ts
bmdalex: once again, tracked by #1300 as this is not directly related to the PR

packages/react/src/themes/teams-high-contrast/components/Attachment/attachmentStyles.ts
```diff
@@ -33,12 +33,6 @@ const attachmentStyles: ComponentSlotStylesInput<AttachmentProps, AttachmentVari
       },
     }),
   }),
-
-  action: ({ variables: v }): ICSSInJSStyle => ({
-    ':hover': {
```
kuzhelov, line 38: where these styles have migrated to? am I right that those are not needed now?

packages/react/src/themes/teams/components/Alert/alertStyles.ts
```diff
@@ -130,41 +128,19 @@ const alertStyles: ComponentSlotStylesInput<AlertProps, AlertVariables> = {
       },
     },
 
-    ...(p.isFromKeyboard && {
-      ':focus': {
-        outline: 0,
-
+    ':focus': {
+      ...(p.isFromKeyboard && {
```
kuzhelov: 👍

packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -135,41 +117,26 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps & ButtonState, any> = {
...
-          ...(!isFromKeyboard && {
-            ':focus': {
-              ':active': {
+
+          ':focus': isFromKeyboard
+            ? {
                 color: circularColorActive,
-                backgroundColor: circularBackgroundColorActive,
-                borderColor: circularBorderColorActive,
-                boxShadow: 'none',
```
kuzhelov, line 169: shouldn't `boxShadow: none` be preserved, the same way it is done for regular one here?

https://github.com/stardust-ui/react/pull/1269/files?file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-62f5a77adef0b99461fbb8c43b4c5b04R101

packages/react/src/themes/teams/getBorderFocusStyles.ts
```diff
@@ -0,0 +1,75 @@
...
+    top: borderEdgeValue,
+    right: borderEdgeValue,
+    bottom: borderEdgeValue,
+    left: borderEdgeValue,
+    ...styles,
+  }
+}
+
+/**
+ * Returns style object that can be used for styling components on focus state
```
kuzhelov: nit: sentence needs dot

packages/react/src/themes/teams/getBorderFocusStyles.ts
```diff
@@ -0,0 +1,75 @@
...
+ * NOTE: the element where this is used needs to have relative positioning so that the
+ * pseudo elements created on focus can be properly positioned
```
kuzhelov: nit: dot is missed

packages/react/src/themes/teams/getBorderFocusStyles.ts
```diff
@@ -0,0 +1,75 @@
...
+ */
+const getBorderFocusStyles = (args: BorderFocusStyles): ICSSInJSStyle => {
+  const sv = args.siteVariables || ({} as BorderFocusStyles)
```
kuzhelov: cast to `BorderFocusStyles` is semantically incorrect in this context. Lets cast to the type of `siteVariables`, i.e. to `SiteVariablesPrepared`. Also, braces may be safely omitted

packages/react/src/components/Attachment/Attachment.tsx
```diff
@@ -122,13 +122,10 @@ class Attachment extends UIComponent<ReactProps<AttachmentProps>, AttachmentStat
...
         )}
-        {action && (
-          <div className={classes.action}>
-            {Button.create(action, {
-              defaultProps: { iconOnly: true, text: true },
-            })}
-          </div>
-        )}
+        {action &&
+          Button.create(action, {
```
kuzhelov, line 123: not sure about these changes in the elements structure provided to Attachment's layout, as those making things inconsistent - would expect either every slot being wrapped with div, or all not being wrapped. 

Also, given that styles for `action` slot are removed (https://github.com/stardust-ui/react/pull/1269/files?file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-8bdbadd856993b56fb3e32cd80f44335L37), it seems questionable that `classes.action` originally applied to wrapper now is applied as `styles` to Button. 

Could you, please, suggest the reasons for these changes?

packages/react/src/themes/teams-high-contrast/components/Attachment/attachmentStyles.ts
bmdalex: indeed they are not needed and they were applied to the wrong element anyway ( wrapping div instead of stardust `Button` )

packages/react/src/themes/teams/components/Button/buttonStyles.ts
bmdalex: great catch, thanks, fixed

packages/react/src/themes/teams/getBorderFocusStyles.ts
bmdalex: fixed, thanks

packages/react/src/themes/teams/getBorderFocusStyles.ts
bmdalex: fixed, thanks

packages/react/src/themes/teams/getBorderFocusStyles.ts
bmdalex: fixed, thanks

packages/react/src/components/Attachment/Attachment.tsx
bmdalex: > not sure about these changes in the elements structure provided to Attachment's layout, as those making things inconsistent - would expect either every slot being wrapped with div, or all not being wrapped.

good point, I was able to bring consistency by removing unnecessary wrapping divs from `icon` and `action` slots with no regressions

> given that styles for action slot are removed (https://github.com/stardust-ui/react/pull/1269/files?file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-8bdbadd856993b56fb3e32cd80f44335L37), it seems questionable that classes.action originally applied to wrapper now is applied as styles to Button.
Could you, please, suggest the reasons for these changes?

styles for `action` slot are removed only for HC theme because they were redundant, we still apply styles for `action` slots in `packages/react/src/themes/teams/components/Attachment/attachmentStyles.ts`

And yes, it was probably wrong to apply them on the wrapper in initial implementation or the implementation is just outdated. Should be fine now.

The changes are done in order to optimize number of DOM nodes and apply styles to the correct element, hope it makes sense

packages/react/src/themes/teams-high-contrast/components/Button/buttonVariables.ts
mnajdova: Once the color scheme is in, we can move the Button styles to use the color scheme, and these things should be resolved. I agree with @Bugaa92 that these are not changes in scope of this PR.

packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -87,44 +83,27 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps & ButtonState, any> = {
...
+
+        ':focus': {
+          boxShadow: 'none',
+          ...(isFromKeyboard
+            ? {
+                color: colorFocus,
+                backgroundColor: backgroundColorFocus,
+                ...borderFocusStyles,
+              }
+            : { ':active': { backgroundColor: backgroundColorActive } }),
```
mnajdova, line 105: Why we decided to still use the background active color, but not the foreground (the colorActive variable). When we press the button now, we may have some regressions..

CHANGELOG.md
```diff
@@ -16,9 +16,9 @@
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-
 ### BREAKING CHANGES
 - Rename `context` prop to `mountNode` in `PortalInner` @layershifter ([#1288](https://github.com/stardust-ui/react/pull/1288))
+- Add focus border styling mechanism @Bugaa92 ([#1269](https://github.com/stardust-ui/react/pull/1269))
```
mnajdova: Add `in Teams theme` in the changelog entry

packages/react/src/themes/teams/components/Alert/alertStyles.ts
```diff
@@ -101,71 +101,28 @@ const alertStyles: ComponentSlotStylesInput<AlertProps, AlertVariables> = {
...
-
-    [`& .${teamsIconClassNames.filled}`]: {
-      display: 'none',
-    },
-
-    [`& .${teamsIconClassNames.outline}`]: {
-      display: 'block',
-    },
+  action: ({ variables: v, props: p, theme: { siteVariables } }): ICSSInJSStyle => {
+    const iconFilledStyles = getIconFillOrOutlineStyles(false)
```
mnajdova: Can we change the signature to something like this: `getIconFillOrOutlineStyles({outline: false})` so that it will be more obvious what the false is referring to?

packages/react/src/themes/teams/components/Attachment/attachmentStyles.ts
```diff
@@ -64,23 +62,28 @@ const attachmentStyles: ComponentSlotStylesInput<AttachmentProps, AttachmentVari
...
-    },
+  action: ({ props: p, variables: v, theme: { siteVariables } }): ICSSInJSStyle => {
+    const iconFilledStyles = getIconFillOrOutlineStyles(false)
 
-    ':hover': {
-      [`& .${teamsIconClassNames.filled}`]: {
-        display: 'block',
+    return {
+      [`& .${Icon.className}`]: {
+        color: v.textColor, // this breaks the color change on hover
```
mnajdova, line 70: Should we use different color on regular/hover icons? Should we add condition about this? What this actually break?

packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -193,64 +155,23 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps & ButtonState, any> = {
...
+
+          ':focus': {
+            boxShadow: 'none',
+            ...(isFromKeyboard
+              ? {
+                  color: primaryColorFocus,
+                  backgroundColor: primaryBackgroundColorFocus,
+                  ...borderFocusStyles,
+                }
+              : { ':active': { backgroundColor: primaryBackgroundColorActive } }),
```
mnajdova, line 173: What about the text color: `primaryColorActive`?

packages/react/src/themes/teams/components/Button/buttonStyles.ts
bmdalex: we don't foreground colors were the same as regular (not focused) view, no point in having so many variables that actually bring nothing

packages/react/src/themes/teams/components/Attachment/attachmentStyles.ts
bmdalex: I'm just moving existing code around, I did not actually make that change; you can see that if you follow the history carefully :)

packages/react/src/themes/teams/components/Button/buttonStyles.ts
bmdalex: same response as in: https://github.com/stardust-ui/react/pull/1269#discussion_r282614613

it was not used, it was the same as regular color so no need for it

CHANGELOG.md
bmdalex: done, thanks

packages/react/src/themes/teams/components/Alert/alertStyles.ts
bmdalex: sounds good, applied, thanks

kuzhelov (approved):

bmdalex:
I had to squash my commits into one in order to be able to merge more easily the big amount of changes that got to master recently via #1069

mnajdova (approved):

PR #1270, opened 2019-04-25
feat(Dropdown): highlight by character key in non-search

silviuaavram:
Behaves as in [aria spec](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html)

When open, sending a character key will focus on the item that begins with that key.

If there is already an option highlighted, the search will begin from the next item.

If it reaches the end of the array and does not find any match, it will check from the start (circular movement).

If there is no option starting with that character, it will not do anything.

If characters are typed in rapid succession (one in under 500ms) it will search for the option starting with those keys. After 500ms after last key type, the 'starting with' string is cleared.

Also includes adding to the state the array of already filtered files (by value and searchQuery) to be used where needed. These strings will be updated only when the items are changed, value is changed in multiple selection or searchQuery is changed in search.


packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -941,6 +942,16 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
         e.stopPropagation()
         return
       default:
+        const keyString = String.fromCharCode(keyCode)
+        if (/[a-zA-Z]/.test(keyString)) {
```
bmdalex: what about numbers or any other printable characters? they can be part of the dropdown items as well.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -941,6 +942,16 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+          const highlightedIndex = this.getHighlightedIndexOnCharKeyPress(keyString)
+          if (highlightedIndex >= 0) {
+            this.setState({
```
bmdalex: shouldn't we do `trySetState`? the user might still want to control `highlightedIndex` and opt out from this behavior

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1035,6 +1046,30 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     }
   }
 
+  private getHighlightedIndexOnCharKeyPress = (keyPressed: string): number => {
+    const { highlightedIndex } = this.state
+    const filteredItemLowercasedStrings = this.getItemsFilteredBySearchQuery().map(item =>
```
bmdalex: `getItemsFilteredBySearchQuery` can prove to be expensive for large `items` array and we're already calling it pretty often; can we add a `state` property, let's call it `itemsAsString` that keeps a copy of the items as string and update it only when `items` array changes?

```typescript
componentDidUpdate(prevProps: DropdownProps) {
  const items = this.props.items
  const prevItems = prevProps.items

  if (
    items.length !== prevItems.length ||
    _.difference(items, prevItems).length > 0 ||
    _.difference(prevItems, items).length > 0
  ) {
    this.trySetState({ itemsAsString: this.getItemsAsString() })
  }
}
```

changes should be minimal as `getItemsFilteredBySearchQuery` will use `itemsAsString` directly

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1035,6 +1046,30 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+      newHighlightedIndex = _.findIndex(
+        filteredItemLowercasedStrings,
+        item => item.startsWith(keyPressed.toLowerCase()),
+        highlightedIndex + 1,
+      )
+    }
+
+    if (newHighlightedIndex < 0) {
+      return _.findIndex(filteredItemLowercasedStrings, item =>
+        item.startsWith(keyPressed.toLowerCase()),
```
bmdalex: super nit: compute `keyPressed.toLowerCase()` only once

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1035,6 +1046,30 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
   }
 
+  private getHighlightedIndexOnCharKeyPress = (keyPressed: string): number => {
+    const { highlightedIndex } = this.state
+    const filteredItemLowercasedStrings = this.getItemsFilteredBySearchQuery().map(item =>
+      this.props.itemToString(item).toLocaleLowerCase(),
+    )
+    let newHighlightedIndex = -1
+
+    if (_.isNumber(highlightedIndex)) {
```
bmdalex, line 1123: 👍

bmdalex (commented):
looks good, nice feature, just few comments

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: I also want to have the best performance, but only having these values in state won't help that much in my opinion. After you take a closer look in `getItemsFilteredBySearchQuery` you will see that the current implementation will not benefit much from the items as strings.

But maybe we can look into this a bit more and find a strategy that is better overall. Thanks!

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: good catch, thanks!

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: I will try to find another regex.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -318,9 +325,36 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+      _.difference(prevItems, items).length > 0
+    const valueHasBeenAddedOrRemoved =
+      (this.props.multiple && value.length !== prevValue.length) ||
+      _.difference(value, prevValue).length > 0 ||
+      _.difference(prevValue, value).length > 0
+    const searchQueryHasBeenUpdated =
+      this.props.search && this.state.searchQuery !== prevState.searchQuery
+
+    if (itemsHaveBeenUpdated || valueHasBeenAddedOrRemoved || searchQueryHasBeenUpdated) {
+      this.setFilteredItemsAndItemStrings()
```
layershifter: This will produce an additional render 😭 Please consider to use `static getDerivedStateFromProps()`, in that case you don't need `itemsHaveBeenUpdated` and other complicated logic, just always filter them

silviuaavram: will look into it, thank you!

silviuaavram: refactored with `getDerivedStateFromProps` and it works, it's just that we will get the error from React about the deprecated `componentWillReceiveProps` from the controlled component implementation.

leave it like this and create a task to fix that ?

silviuaavram: actually we need to fix this, it fails the build.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -317,9 +324,62 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
   }
 
   a11yStatusTimeout: any
+  charKeysPressedTimeout: any
```
sophieH29, line 327: it can be `number` type as `setTimeout` returns number

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -732,6 +733,72 @@ describe('Dropdown', () => {
...
+      const wrapper = mountWithProvider(<Dropdown items={items} />)
+      const dropdown = wrapper.find(Dropdown)
+      const itemsList = wrapper.find(`ul.${Dropdown.slotClassNames.itemsList}`)
+      const triggerButton = wrapper.find(`button.${Dropdown.slotClassNames.triggerButton}`)
+
+      triggerButton.simulate('click')
+      itemsList.simulate('keydown', { keyCode: keyboardKey.A, key: 'A' })
+      expect(dropdown.state('highlightedIndex')).toBe(0)
+
+      jest.advanceTimersByTime(200)
```
miroslavstastny: nit: `Dropdown.charKeyPressedCleanupTime / 2`

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1161,12 +1230,21 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+  private setA11ySelectionMessage = (a11ySelectionStatus: string): void => {
     clearTimeout(this.a11yStatusTimeout)
+    this.setState({ a11ySelectionStatus })
     this.a11yStatusTimeout = setTimeout(() => {
       this.setState({ a11ySelectionStatus: '' })
     }, Dropdown.a11yStatusCleanupTime)
   }
+
+  private setStartingString = (startingString: string): void => {
+    clearTimeout(this.charKeysPressedTimeout)
```
miroslavstastny, line 1279: I would replace `clearTimeout` + `setTimeout` with `_.debounce`

miroslavstastny (approved):

packages/react/src/components/Dropdown/Dropdown.tsx
layershifter: It can't because it conflicts with type for `NodeJS.Timer`: https://github.com/Microsoft/TypeScript/issues/842
Obvious usage of `window.setTimeout()` solves the issue

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -317,9 +324,62 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+   * their string equivalents, in order to be used throughout the component.
+   * @param props
+   * @param state
+   */
+  static getDerivedStateFromProps(props: DropdownProps, state: DropdownState) {
+    const { items, itemToString, multiple, search } = props
+    const { value } = state
+
+    if (!items) {
+      return state
```
```suggestion
      return null
```

> getDerivedStateFromProps is invoked right before calling the render method, both on the initial mount and on subsequent updates. It should return an object to update the state, or null to update nothing.

https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -317,9 +324,62 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+    let filteredResults
+
+    if (search) {
+      const { itemToString } = props
+      const { searchQuery } = state
+
+      if (_.isFunction(search)) {
+        filteredResults = {
+          filteredItems: search(filteredItemsByValue, searchQuery),
+        }
```
layershifter, line 360: As we have already used early return, I suggest to use it there, too

```tsx
if (_.isFunction(search)) {
  return { filteredItems: search(filteredItemsByValue, searchQuery) }
}
```

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1034,6 +1082,28 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
     }
   }
 
+  private getHighlightedIndexOnCharKeyDown = (keyString: string): number => {
```
layershifter: I am lost with semantics of this method, it's called `getHighlightedIndexOnCharKeyDown()`, but it contains `setStartingString`

Should it be `setHighlightedIndexOnCharKeyDown()` and contain `this.setState({ highlightedIndex })`?

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: I thought it's better to have it assigned to that variable in branch, and return at the end `{...state, ...filteredResults }`. Now with return we need, on each branch, to return `{...state, filteredItems: whateverResultOnTheBranch}`.
Do you think it's better?

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: good point. I will refactor it as a setter. Let's see if I also need to refactor the setStartingString one with debounce or not.

PR #1275, opened 2019-04-29
feat: styling approach updates

kuzhelov:
# Review notes
- several files are introduced as an examples that demonstrate features provided by these PR (all of them are introduced to the `docs` site) - _those are not seen to be merged, for now_. It was necessary to make review scenarios of 'fetch branch -> play' smoother

# What is introduced
- Reusable Styles
- styled() function
- Component Variants
- Examples

**Please, take a look into Examples section, where common use-cases to justify these features are provided.**

-----------------

# Reusable Styles

Current options for styling layout are limited, and usually rely on markup of the layout elements.  This is undesirable, as they are then non-controllable by the theme.

Additionally, the use of arbitrary one-off styles allows unconstrained variance of look and feel - again this is undesirable as we attempt to move to a common design language. 

**Reusable styles feature:**
- allows to reuse styles defined in theme
  - now layout styles can be reused in component's styles (e.g., Chat styles may reuse Flex styles, see `Examples` section), and both are controlled by theme
- have limited set of variations defined by style props
  - style props are seen to match Design Spec's language 
  - e.g. `styles.Flex.root({ column: true })`

### API
```jsx
// -----------------------------------------
// GENERAL SYNTAX
// styles.ComponentName.SlotName(props?)
// -----------------------------------------

// -----------------------------------------
// chatStyles.ts
// -----------------------------------------
const chatStyles = {
  root: ({ .., styles }) => ({
     ...styles.Flex.root({ column: true })
     ...otherCssStyles
  }),
  item: ({ styles }) => ({
    ...styles.FlexItem.root({ ... }),
    ...otherCssStyles
  })
}
```

# styled() function
- allows to apply theme styles to intrinsic/third-party React components on client side
- important piece to provide smooth transition to Stardust for the client
  - while preserving Stardust's structured and semantic approach to styles (and design, in general)

### API
```jsx
// original client's code
const MyPage = ( ... ) => (
  <div className='third-party-css-component'>
    <section className='third-party-css-component__section--important' >...</section>
    <aside className='third-party-css-component__aside' >...</aside>
  </div>
)

// --------------------------------------------------------------
// SUGGESTED approach - with 'classes' consumed from styled()
// --------------------------------------------------------------
import { styled } from '@stardust-ui/react'

const MyPage = ( ... ) => styled(({ classes }) => (
  <div className={classes.MyPage.root()}>

    // indicates that client's design language implies different variations of 'section'
    <section className={classes.MyPage.section({ important: true })}>...</section>

    <aside className={classes.MyPage.aside()}>...</aside>
  </div>
))

// --------------------------------------------------------------
// ALTERNATIVE approach - with 'styles' consumed from styled() 
// - note that immediate migration to Stardust components (in this case, Box) is necessary
// --------------------------------------------------------------
import { styled } from '@stardust-ui/react'

const MyPage = ( ... ) => styled(({ styles }) => (
  <Box styles={styles.MyPage.root()}>
    <Box as='section' styles={styles.MyPage.section({ important: true })}>...</Box>
    <Box as='aside' styles={styles.MyPage.aside()}>...</Box>
  </Box>
))
```
### Notes
Optimised implementation - no noticeable overhead introduced to rendering phases, due to
- minimal amount of additional objects created on rendering phase (just 2, not dependent on the amount of styles in theme)
- `styles`/`classes` are evaluated 'on demand'

# Component Variants

Current Stardust options are limited to address the following situation on the client side.

Suppose that there is one category of buttons that should appear with icon's borders being rounded, and another one - with icon's border being rectangular. 

This, essentially, results in need to define two variants of Button, with `icon` slot being styled differently, but **with all the behavioural (and accessibility) aspects of `Button` being preserved**. 

Also, we would like to ensure that these `Button` variants won't spawn in uncontrolled way - i.e. all of them will be registered in theme, discoverable and, in general, will match `Button` variants of the Design Spec.

**Component Variants feature:**

- adjusts component's appearance, while preserving behavior aspects of original component
- allows client to extend set of props of original component, in controlled way
  - this should be seen as an extensibility point to introduce design terms that are specific to client's domain
- provides 'is a' semantics - props, behavior and accessibility of the original component (or, in general, all the _scenarios_ of original component) will work for its variant

### API
```js
// --------------------------------
// DEFINE THEME BITS (styles)
// theme/index.ts
// --------------------------------
const theme = {
  componentStyles: { .. }
  ...
  componentVariants: {
    Input: { // <------------- which component variant defined for
      Primitive: {     // <--------------- variant name
         // 'variables' and 'styles' props are optional,
         // will be applied on top of corresponding ones of 'Input'
         variables: { 
            background: transparent,
            underlineOnFocus: null
         }
         styles: {
             ....
         }
      }
    }
  }
}

// --------------------------------
// DEFINE COMPONENT FOR VARIANT
// components/index.ts
// --------------------------------
import { createVariant } from '@stardust-ui/react'

// propTypes and defaultProps are optional
export const PrimitiveInput = createVariant(Input, 'Primitive', propTypes = { }, defaultProps = { })

// --------------------------------
// USE VARIANT
// client.ts
// --------------------------------
<PrimitiveInput       // <----- all props of Input are supported
  placeholder='Your name' 
  icon={..} ... />
```

# Examples
### Problem
Often it is necessary to reuse layout styles (`Flex`, `Grid`) in component's style definitions, as otherwise styling information will be hardcoded into markup with corresponding `<Flex/ >` and `<Grid />` elements, and this will steal control over this aspect from theme. 

Consider an example of `Chat` and `ChatItem` components:

```jsx
// chat.ts
// -----------------------------------------
// BAD 
// layout styling is now hardcoded into component's markup (with <Flex.Item />), 
// and, thus, cannot be controlled by theme styles for ChatItem component - it will be flex layout always
// -----------------------------------------
const ChatItem = ({ props }) => <Flex.Item ... >...</Flex.Item>

// -----------------------------------------
// BAD 
// layout styling is now hardcoded into component's markup (with <Flex/>), 
// thus, cannot be controlled by theme styles for Chat component- it will be always layout of flex item
// -----------------------------------------
const Chat = ({ props }) = (
  <Flex column>  
    <ChatItem ... />
  </Flex>
)
```

### Solution

This is how this problem will be addressed with layout styles being reused:

```jsx
// -----------------------------------------
// chatStyles.ts
// -----------------------------------------
const chatStyles = {
  root: ({ .., styles }) => ({
     ...styles.Flex.root({ column: true })
     ...otherCssStyles
  }),
  item: ({ styles }) => ({
    ...styles.FlexItem.root({ ... }),
    ...otherCssStyles
  })
}

// -----------------------------------------
// chat.ts (now has no styles hardcoded into markup)
// -----------------------------------------
const Chat = ({ props, ElementType, classes, styles }) = (
  <ElementType className={classes.root}>  // GOOD, no layout styles hardcoded into markup
    <ChatItem styles={styles.item} ... />  // GOOD, no layout styles hardcoded into markup
  </ElementType>
)
```

## Example: client-defined styles (_style terms_)

Lets suppose that, as a client, we want to define styles for `master-details` layout that is reused in our app (and thus, most probably, is defined by app's Design Spec).

### Problem

Previously, to address this case, it was necessary to the client to either:
- define set of custom CSS styles and apply them as class names at places where necessary
  - obvious downside is that this introduces another styling approach which is different from the one Stardust provides, and it results in client needs to maintain both
- define Stardust component for that. However, it it may be problematic because of the following:
  - the styles may be not a direct attribute of specific component (such as `scrollbar` or `border-on-focus` styles which are designed to be reused in many style definitions for different components)
  - Stardust component may be an overkill in case if client needs just to apply styles

### Solution

This is how this case can be addressed with reusable styles feature.

```jsx
// --------------------------
// DEFINE
// --------------------------
// masterDetailsStyles.ts
const masterDetailsLayoutStyles = {
    root: ({ ..., styles }) => ({ 
       ...styles.Flex.root({ gap: ... }), 
       ... 
    }),
    masterColumn: ({ ... }) => ({ ... })
    detailsColumn:  ({ ..., styles }) => ({ 
       ...styles.FlexItem.root({ grow: true }), 
       ... 
    })
}

// theme/index.ts
const theme = {
  componentStyles: {
    ...
    MasterDetailsLayout: masterDetailsLayoutStyles
  }
}
```

And now, to use these styles:

```jsx
// --------------------------
// USE (see next section for 'styled()' description)
// --------------------------
// client.ts
import { styled } from '@stardust-ui/react'

const Foo = ({ .. }) => styled(({ classes }) => (
  <div className={classes.MasterDetailsLayout.root()} >
     <aside className={classes.MasterDetailsLayout.masterColumn()}>...</aside>
     <section className={classes.MasterDetailsLayout.detailsColumn()}>...</section>
  </div>
))
```

### Additional notes

From there this layout styles could naturally evolve to React component, or even Stardust component. These is the only code necessary for that (no changes to theme's style definitions):

```jsx
type MasterDetailsLayoutProps = {
  master: React.Node // content to be rendered to 'master' slot
  details: React.Node // content to be rendered to 'details' slot
}

// ------------------------------
// REGULAR COMPONENT
// - integrate only styles from Stardust
// ------------------------------
import { styled } from '@stardust-ui/react'

const MasterDetailsLayout = (props) => styled(({ classes }) => (
  <div className={classes.MasterDetailsLayout.root()} >
     <aside className={classes.MasterDetailsLayout.masterColumn()}>...</aside>
     <section className={classes.MasterDetailsLayout.detailsColumn()}>...</section>
  </div>
))

// ------------------------------
// STARDUST COMPONENT
// - if further integration with Stardust is needed (such as accessibility, 'as' concept, etc)
// ------------------------------
import { createComponent } from '@stardust-ui/react'

const render = ({ props, classes }) => (
  <div className={classes.MasterDetailsLayout.root} >
     <aside className={classes.MasterDetailsLayout.masterColumn}>...</aside>
     <section className={classes.MasterDetailsLayout.detailsColumn}>...</section>
  </div>
))

const MasterDetailsLayout = createComponent(
  displayName: 'MasterDetailsLayout',
  render: render
)
```

## Example: <Text /> variant with client-specific `timestamp` and `userMention` props
Lets imagine that we have `Text` component (with general-use props API), and we want to introduce `ChatText` variant which would support `timestamp` and `userMention` terms - the ones that are _specific to our chat app_.

### Problem
Previously the options were
- introduce props as part of Stardust API, which is problematic due to
   - client cannot rely on that
   - Stardust spec is not able to absorb terms that are not general enough
- introduce variables to support this case, and this had resulted in awkward API and, what is more important, broken styles encapsulation on the client side:

```jsx
// mix of props and variables to express the same thing
<Text important variables={{ userMention: true }} /> 

// textStyles.ts
const styles = {
  root: ({ ..., variables }) => ({
     /* .. MIXED list of other variants styles - BROKEN ENCAPSULATION .. */
     ...(variables.userMention && { ... })
  })
}
```
### Solution

Now this scenario may be addressed with variants - note that styles encapsulation, semantics and design language's expressiveness is preserved in the JSX markdown

```jsx
// --------------------------------
// DEFINE
// --------------------------------

// types/index.ts
type ChatTextProps = {
  userMention: boolean,
  timestamp: boolean
}

// theme/index.ts
const theme = {
  componentStyles: ...,
  componentVariants: {
    Text: {
       ChatText: {
          styles: ({ props: TextProps & ChatTextProps }) => ({
             ...(props.userMention && { /* .. userMention styles  .. */ })
             ...(props.timestamp && { /* .. timestamp styles  .. */ })
          )}
       }
    }
}

// components/index.ts
export const ChatText = createVariantOf<TextProps, ChatTextProps>(
  Text,  // of component
  'ChatText', // variant name
  { atMention: PropTypes.bool, timestamp: PropTypes.bool } // propTypes
)

// --------------------------------
// USE
// --------------------------------
<ChatText timestamp>11:03 PM</ChatText>
```


packages/react/src/lib/mergeThemes.ts
```diff
@@ -212,6 +216,8 @@ const mergeThemes = (...themes: ThemeInput[]): ThemePrepared => {
 
     acc.componentVariables = mergeThemeVariables(acc.componentVariables, next.componentVariables)
 
+    acc.componentVariants = mergeComponentVariants(acc.componentVariants, next.componentVariants)
```
kuzhelov, line 219: policy of merging variants' sets; in case of name collisions (`Button` -> `EmojiButton` in source theme, `Button`-> `EmojiButton` in theme to merge) the last one will win

packages/react/src/lib/styled.tsx
```diff
@@ -0,0 +1,120 @@
+import * as React from 'react'
```
kuzhelov, line 1: optimise

packages/react/src/themes/teams/componentVariants.ts
```diff
@@ -0,0 +1 @@
+export { default as Input } from './components/Input/variants'
```
kuzhelov, line 1: these temporary changes are introduced for the sake of example to reviewer - won't be merged

docs/src/examples/components/Button/Types/ButtonExample.shorthand.tsx
```diff
@@ -1,8 +1,11 @@
 import * as React from 'react'
```
kuzhelov, line 1: these temporary changes are introduced for the sake of example to reviewer - won't be merged

docs/src/examples/components/Input/Variants/Variants.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from 'react'
```
kuzhelov, line 1: these temporary changes suggest the way how variants could be rendered to docs in automated way - won't be merged

docs/src/examples/components/Input/Variants/index.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
```
kuzhelov, line 1: these temporary changes are introduced for the sake of example to reviewer - won't be merged

docs/src/examples/components/Input/index.tsx
```diff
@@ -3,9 +3,11 @@ import * as React from 'react'
 import Rtl from './Rtl'
```
kuzhelov, line 3: these temporary changes are introduced for the sake of example to reviewer - won't be merged

packages/react/src/themes/teams/components/Input/variants/Custom.ts
```diff
@@ -0,0 +1,16 @@
+export default {
```
kuzhelov, line 1: these temporary changes are introduced for the sake of example to reviewer - won't be merged

packages/react/src/themes/teams/components/Input/variants/Minimalistic.ts
```diff
@@ -0,0 +1,5 @@
+export default {
```
kuzhelov, line 1: these temporary changes are introduced for the sake of example to reviewer - won't be merged

packages/react/src/themes/teams/components/Input/variants/index.ts
```diff
@@ -0,0 +1,7 @@
+import Minimalistic from './Minimalistic'
```
kuzhelov, line 1: these temporary changes are introduced for the sake of example to reviewer - won't be merged

packages/react/src/themes/types.ts
```diff
@@ -261,6 +263,7 @@ export interface ThemeInput {
   siteVariables?: SiteVariablesInput
   componentVariables?: ThemeComponentVariablesInput
   componentStyles?: ThemeComponentStylesInput
+  componentVariants?: any
```
kuzhelov, line 331: TODO: address `any`

packages/react/src/lib/createVariant.ts
```diff
@@ -0,0 +1,38 @@
+import hoistNonReactStatic from 'hoist-non-react-statics'
+
+function createVariant<TProps>(
```
kuzhelov, line 3: make API consistent with the one provided by `createComponent` function - at the end it should match the following (prop names may be different):

```jsx
const Variant = createVariant({
  forComponent: Input,
  variantName: 'Minimalistic',
  propTypes: ...,
  defaultProps: ...
})
```

and preserve the same type safety checks it has now.

packages/react/src/themes/teams/components/Menu/menuItemStyles.ts
```diff
@@ -119,8 +119,8 @@ const pointingBeak: ComponentSlotStyleFunction<MenuItemPropsAndState, MenuVariab
       left: '50%',
       transform: 'translateX(-50%) translateY(-50%) rotate(45deg)',
       margin: '.5px 0 0',
-      width: pxToRem(10),
-      height: pxToRem(10),
+      width: v.pointingBeakSize,
```
kuzhelov, line 122: the intent was to use the same token here

packages/react/src/themes/types.ts
```diff
@@ -356,6 +378,8 @@ export interface ThemeComponentStylesPrepared {
   DropdownItem?: ComponentSlotStylesPrepared
   DropdownSearchInput?: ComponentSlotStylesPrepared
   Embed?: ComponentSlotStylesPrepared
+  Flex?: ComponentSlotStylesPrepared
```
kuzhelov: this is a clear indication that these types were useless in a sense of declaring each component individually, as nothing was checked/guaranteed with them. This entire declaration was identical, in a sense of its effect, to
```js
type ThemePrepared = { [K: string]: ComponentSlotStylesPrepared }
```

PR #1280, opened 2019-05-02
chore(perf): Normalize perf tests results POC

miroslavstastny:
Our perf test numbers are not quite stable - this can be caused by unstable environment (shared CI machine) the tests are running in.
As an attempt to stabilize the numbers, this PR adds normalization of the results:
1. before each example is rendered, we render Baseline example (react component with no stardust in it)
2. we use that baseline result to normalize the time of the real perf example.

Let's try this for some time and then evaluate the result.

Another options for further investigation:
- as suggested by @hughreeling, instead of rendering nonstardust example, render stardust example with fixed version of stardust?
- as suggested by @kuzhelov, check what part of HW config (other than CPU load) affects the numbers (GPU, mem...)


perf/src/PerfBaseline.tsx
```diff
@@ -0,0 +1,12 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+
+const PerfBaseline = () => (
+  <>
+    {_.times(1000, i => (
+      <div>perf baseline {i}</div>
```
layershifter:
```suggestion
      <div key={i}>perf baseline {i}</div>
```

build/gulp/tasks/perf.ts
```diff
@@ -82,39 +105,36 @@ const getPercentDiff = (minValue: number, actualValue: number): number =>
   _.round((actualValue / minValue) * 100 - 100, 2)
 
 const createMarkdownTable = (
-  normalizedMeasures: NormalizedMeasures,
-  metricName: string = 'actualTime',
-  fields: string[] = ['avg', 'median'],
+  perExamplePerfMeasures: PerExamplePerfMeasures,
+  metricName: MeasuredValues = 'actualTime',
+  fields: ('avg' | 'avgNormalized' | 'median' | 'medianNormalized')[] = [
```
layershifter:
```suggestion
  fields: (key of Omit<ReducedMeasures, value>)[] = [
```

It looks like a valid change

layershifter (approved):

PR #1281, opened 2019-05-02
feat(Ref): extract to a separate package, add utils

### LINKED ISSUE #998: RFC: Ref utils

layershifter:
Fixes #998.

----

This PR:
- moves `Ref` component to `@stardust-ui/react-component-ref` package, however `Ref` can be still accessed from `@stardust-ui/react`
- adds `isRefObject()`, `toRefObject()` utils


packages/react-component-ref/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "license": "MIT",
+  "main": "dist/commonjs/index.js",
+  "module": "dist/es/index.js",
+  "peerDependencies": {
+    "react": "^16.8.0",
+    "react-dom": "^16.8.0"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
```
layershifter, line 30: FYI @miroslavstastny `publishConfig` is defined

packages/react-component-ref/src/Ref.tsx
```diff
@@ -1,14 +1,14 @@
 import * as customPropTypes from '@stardust-ui/react-proptypes'
 import * as PropTypes from 'prop-types'
 import * as React from 'react'
-import { isForwardRef } from 'react-is'
+import * as ReactIs from 'react-is'
```
layershifter, line 4: **nit:** To be common with other imports

packages/react-component-ref/src/handleRef.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (typeof ref === 'function') {
+    ref(node)
+    return
+  }
+
+  if (ref !== null && typeof ref === 'object') {
+    // The `current` property is defined as readonly, however it's a valid way because
+    // `ref` is a mutable object
+    ;(ref as React.MutableRefObject<N>).current = node
```
layershifter, line 26: Only this line was changed: we don't need `@ts-ignore` more because `MutableRefObject` was introduced to typings

packages/react-component-ref/src/RefFindNode.tsx
```diff
@@ -20,10 +21,10 @@ export default class RefFindNode extends React.Component<RefFindNodeProps> {
 
   static propTypes = {
     children: PropTypes.element.isRequired,
-    innerRef: customPropTypes.ref,
+    innerRef: customPropTypes.ref.isRequired as PropTypes.Validator<React.Ref<any>>,
```
mnajdova, line 24: 👍

packages/react-component-ref/test/RefForward-test.tsx
```diff
@@ -1,7 +1,7 @@
+import { RefForward } from '@stardust-ui/react-component-ref'
```
mnajdova, line 1: I don't like that we have inconsistency between the imports in these tests vs the imports in the components' tests. Can we change this import or is there a reason why it is like this?

packages/react-component-ref/test/handleRef-test.ts
```diff
@@ -1,5 +1,5 @@
+import { handleRef } from '@stardust-ui/react-component-ref'
```
mnajdova, line 1: See the prev comment for the import

packages/react-component-ref/src/toRefObject.ts
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react'
+
+const nullRefObject: React.RefObject<null> = { current: null }
```
mnajdova, line 3: What if the user uses reference like this in the following way: ref.current.focus()? How is it safe to have a `nullRefObject`?

layershifter: It can be `null` by typings:
```tsx
    interface RefObject<T> {
        readonly current: T | null;
    }
    // Mutable it used in hooks an initial value is obviously defined
    // React.useRef(null)
    interface MutableRefObject<T> {
        current: T;
    }
```

packages/react-component-ref/test/RefForward-test.tsx
layershifter: We want to test public APIs and get rid of `src` alias: it creates issues with IDE and works only in Jest/Webpack.

/cc @kuzhelov

CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Features
+- `Ref` component extracted to a `@stardust-ui/react-component-ref` @layershifter ([#1281](https://github.com/stardust-ui/react/pull/1281))
+- added `isRefObject()`, `toRefObject()` and `unstable_mergeRefs()` utils for React refs @layershifter ([#1281](https://github.com/stardust-ui/react/pull/1281))
```
mnajdova: If we are not sure about this method, why even adding it with the `unstable_` prefix? Is this blocker on client's side? I really don't see the things mention in the PR description as a valid points for adding it...

mnajdova (approved):
Approving the changes. As agreed offline, we will not add the `unstable_mergeRefs()` for now, until we are certain of it's API and usage.

PR #1282, opened 2019-05-02
feat(a11y): add default children behavior to behaviors

### LINKED ISSUE #1008: RFC: Accessibility children behavior
## Feature Request
The parent accessibility behavior should define behaviors for children. 

### Problem description

https://stardust-ui.github.io/react/components/menu#usages-toolbar

```jsx
const items = [
  {
    key: 'format',
    accessibility: toolbarButtonBehavior,
  },
  {
    key: 'paperclip',
    accessibility: toolbarButtonBehavior,
  },

// ...

<Menu accessibility={toolbarBehavior} />
```

I defined `toolbarBehavior` for a `Menu`, why I should _always_ define `toolbarButtonBehavior`?

### Proposed solution

Extend the `Accessibility` to support `childrenBehaviors` (???):

```tsx
const toolbarBehavior: Accessibility = (props: any) => ({
  attributes: {},
  focusZone: {},
  childrenBehaviors: {
    item: toolbarButtonBehavior, // here is magic
  },
})
```

silviuaavram:
Fixes https://github.com/stardust-ui/react/issues/1008

It adds child behaviors by default to Menu related behaviors:
- menuBehavior/submenuBehavior - menuItemBehavior
- tabListBehavior - tabBehavior
- toolbarBehavior - toolbarButtonBehavior


packages/react/src/components/Menu/Menu.tsx
```diff
@@ -203,6 +207,7 @@ class Menu extends AutoControlledComponent<ReactProps<MenuProps>, MenuState> {
           active,
           inSubmenu: submenu,
           indicator,
+          accessibility: (accessibility as any).childrenBehavior,
```
sophieH29: what if `accessibility` is undefined?

packages/react/src/lib/accessibility/types.ts
```diff
@@ -160,6 +160,7 @@ export interface AccessibilityDefinition {
   focusZone?: FocusZoneDefinition
   focusTrap?: FocusTrapDefinition
   autoFocus?: AutoFocusZoneDefinition
+  childrenBehavior?: Accessibility
```
sophieH29: I would name it `childBehavior` as it is only one.

packages/react/src/lib/accessibility/Behaviors/Toolbar/toolbarBehavior.ts
```diff
@@ -22,6 +23,7 @@ const toolbarBehavior: Accessibility = (props: any) => ({
       shouldFocusInnerElementWhenReceivedFocus: true,
     },
   },
+  childrenBehavior: props => toolbarButtonBehavior(props),
```
sophieH29: I think the `toolbarButton` props should be passed here as it is not valid to pass toolbar props.
`childBehavior: toolbarButtonBehavior`

And then pass to `toolbarButton` prop `accessibility: accessibility.childBehavior`. Props should be passed automatically there, like we do it from client side `accessibility={toolbarButtonBehavior}`

silviuaavram: I think I tried this or something similar, and I would get an error in `renderComponent` because it will look for a function to be called with props and I am passing that function already called.

or am I missing something?

packages/react/src/lib/accessibility/types.ts
layershifter: But, why it should be single? We also have `MenuDivider` that can be iterated in the same scope...

sophieH29: I the sense of AccessibilityBehavior type, it can have only one child behavior. For e.g.

```
MenuBehavior= (props) => {
...
childBehavior: menuItemBehavior
}
```

packages/react/src/lib/accessibility/Behaviors/Toolbar/toolbarBehavior.ts
sophieH29: hmm, let me try this

packages/react/src/components/Menu/Menu.tsx
sophieH29: btw, since you're making changes in `Menu`, makes sense to add `childBehavior` for `menuBehavior`

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -148,7 +148,11 @@ class Menu extends AutoControlledComponent<ReactProps<MenuProps>, MenuState> {
     },
   })
 
-  renderItems = (styles: ComponentSlotStylesPrepared, variables: ComponentVariablesObject) => {
+  renderItems = (
+    styles: ComponentSlotStylesPrepared,
+    variables: ComponentVariablesObject,
+    accessibility: Accessibility,
```
layershifter: It should be `AccessibilityBehavior`:

```suggestion
    accessibility: AccessibilityBehavior,
```

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -203,6 +207,7 @@ class Menu extends AutoControlledComponent<ReactProps<MenuProps>, MenuState> {
           active,
           inSubmenu: submenu,
           indicator,
+          accessibility: accessibility ? (accessibility as any).childBehavior : undefined,
```
layershifter:
```suggestion
          accessibility: accessibility ? accessibility.childBehavior : undefined,
```

And actually, we will always have `accessibility` because we have there a check in `getAccessibility()`:

```suggestion
          accessibility: accessibility.childBehavior,
```

packages/react/test/specs/components/Menu/Menu-test.tsx
```diff
@@ -122,43 +122,59 @@ describe('Menu', () => {
...
 
       describe('as a TabList', () => {
         test('root role should be tablist', () => {
-          const menuItemComponent = mountWithProviderAndGetComponent(
+          const menuComponent = mountWithProviderAndGetComponent(
             Menu,
             <Menu accessibility={tabListBehavior} />,
           )
-          expect(getRenderedAttribute(menuItemComponent, 'role', '')).toBe('tablist')
+          expect(getRenderedAttribute(menuComponent, 'role', '')).toBe('tablist')
```
sophieH29, line 177: can you please add one more test, to check if the user would be still able to override behavior for one of the items, for case like
```

getItems = () => [
{id: '1'},
{id: '2', accessibility: menuItemBehavior}
]

...
<Menu accessibility={tabListBehavior} items={getItems()}
```
So the last item would be rendered with role 'menuitem'

sophieH29 (approved):

packages/react/src/components/Menu/Menu.tsx
silviuaavram: will add both the type change and this change. thanks!

packages/react/test/specs/components/Menu/Menu-test.tsx
silviuaavram: good idea. will add it!

packages/react/src/lib/accessibility/types.ts
layershifter: _It's not true_. `menuBehavior` has multiple siblings: `menuItemBehavior`, `menuDividerBehavior` & `submenuBehavior`. I had an offline conversation with @miroslavstastny and we agreed on the same thing that was proposed originally:
```js
menuBehavior= (props) => {
  // ...
  childBehaviors: {
     item: menuItemBehavior,
     divider: menuDividerBehavior,
  }
}
```

It allows to override behaviors for all child items and it's aligned with existing structure on `attributes` and `keyHandlers`. Please take one more look on this change.

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -180,6 +184,7 @@ class Menu extends AutoControlledComponent<ReactProps<MenuProps>, MenuState> {
             variables,
             styles: styles.divider,
             inSubmenu: submenu,
+            accessibility: accessibility.childBehaviors.divider,
```
jurokapsiar: `childBehaviors` is optional, check its existence before accessing `divider`

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -203,6 +208,7 @@ class Menu extends AutoControlledComponent<ReactProps<MenuProps>, MenuState> {
           active,
           inSubmenu: submenu,
           indicator,
+          accessibility: accessibility.childBehaviors.item,
```
jurokapsiar: `childBehaviors` is optional, check its existence before accessing `item`

layershifter (approved):

jurokapsiar (approved):

PR #1286, opened 2019-05-03
[DO NOT REVIEW] feat: add strict type checks for Stardust components' props

kuzhelov:
This PR demos the approach we could use for introducing strict type checks for Stardust components' props. Lets consider it a starting point for discussing the approach to types we might want to introduce.

## Important note
Unfortunately, precommit checks had stripping out some of the generic types used, need to address this issue.

## Type files to review
- types.ts

## Problems discovered (files list)
- ComponentDocSee
- ComponentControls
- ComponentExample
- ComponentProps
- ComponentSidebar
- ComponentSidebarSection
- ItemLayoutExampleSelection
- PortalExample
- PortalExampleControlled
- PortalExampleFocusTrapped
- AsyncShorthand
- IconViewerExample/index
- ChatPaneContent
- PageNotFound
- QuickStart
- DropdownItem

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentProps/ComponentProps.tsx
+++ b/docs/src/components/ComponentDoc/ComponentProps/ComponentProps.tsx
@@ -59,8 +59,10 @@ export default class ComponentProps extends React.Component<any, any> {
             <Flex.Item styles={{ display: 'inline-block' }}>
               <>
                 <Input
-                  type="checkbox"
-                  checked={!!activeDisplayName}
+                  input={{
+                    type: 'checkbox',
+                    checked: !!activeDisplayName,
+                  }}
                   onClick={this.handleToggle}
                   inline
                 />
--- a/docs/src/components/GuidesNavigationFooter.tsx
+++ b/docs/src/components/GuidesNavigationFooter.tsx
@@ -19,7 +19,7 @@ const GuidesNavigationFooter: React.FC<GuidesNavigationFooterProps> = ({ previou
     <br />
     <Flex gap="gap.small">
       {previous && (
-        <Button
+        <Button<Link>
           as={Link}
           content={previous.name}
           icon="arrow left"
@@ -29,7 +29,7 @@ const GuidesNavigationFooter: React.FC<GuidesNavigationFooterProps> = ({ previou
         />
       )}
       {next && (
-        <Button
+        <Button<Link>
           as={Link}
           content={next.name}
           icon="arrow right"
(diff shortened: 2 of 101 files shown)


docs/src/components/ComponentDoc/ComponentProps/ComponentProps.tsx
```diff
@@ -59,8 +59,10 @@ export default class ComponentProps extends React.Component<any, any> {
             <Flex.Item styles={{ display: 'inline-block' }}>
               <>
                 <Input
-                  type="checkbox"
-                  checked={!!activeDisplayName}
+                  input={{
+                    type: 'checkbox',
+                    checked: !!activeDisplayName,
+                  }}
```
layershifter, line 65: This change is invalid...

docs/src/components/GuidesNavigationFooter.tsx
```diff
@@ -19,7 +19,7 @@ const GuidesNavigationFooter: React.FC<GuidesNavigationFooterProps> = ({ previou
     <br />
     <Flex gap="gap.small">
       {previous && (
-        <Button
+        <Button<Link>
```
layershifter, line 22: This is unacceptable solution...

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -127,6 +127,9 @@ class Avatar extends UIComponent<ReactProps<AvatarProps>, any> {
   }
 }
 
-Avatar.create = createShorthandFactory({ Component: Avatar, mappedProp: 'name' })
+Avatar.create = createShorthandFactory<AvatarProps>({
```
layershifter, line 130: Can we grab props from `Component` instead of using generic there?

packages/react/src/lib/factories.ts
```diff
@@ -135,8 +118,8 @@ function createShorthandFromValue({
   options,
 }: {
   Component: React.ReactType
-  mappedProp?: string
-  mappedArrayProp?: string
+  mappedProp?: string | number | symbol
```
layershifter, line 121: Why it can be `number` or `symbol`?

kuzhelov: @layershifter this is WIP - please, do not review yet.

For the question you've asked - this was necessary to match indexer definitions for the typing approach previously used, when it was simplified this has been removed.

But, once again, please, refrain from review now, because there will be some changes in addition, so these comments may be not relevant

docs/src/components/GuidesNavigationFooter.tsx
kuzhelov: @layershifter why, specifically? Generics is an acceptable thing from TS point of view - please, be specific about your criterias when using words like 'unacceptable'

layershifter: - `<Button as={Link} />` - fine
- `<Button<Link> />` - fine
- `<Button<Link> as={Link} />` - too noisy and as I understand without this generic it will be broken

I want to find a way where we can go with the first option

PR #1289, opened 2019-05-03
Update comments in Focus Zones

sophieH29:
Update comments in FocusZone, FocusTrapZone, AutoFocusZone

### DIFF
--- a/packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.types.tsx
+++ b/packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.types.tsx
@@ -7,8 +7,8 @@ export interface AutoFocusZoneProps extends React.HTMLAttributes<HTMLDivElement>
   as?: React.ReactType
 
   /**
-   * Indicates the selector for first focusable item.
-   * @default the first tabbable element will get focus.
+   * Indicates the selector for first focusable item. By default, the first tabbable element will get focus.
+   * @default undefined
    */
   firstFocusableSelector?: string | (() => string)
 }
--- a/packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.types.tsx
+++ b/packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.types.tsx
@@ -8,6 +8,7 @@ export interface AutoFocusZoneProps extends React.HTMLAttributes<HTMLDivElement>
 
   /**
    * Indicates the selector for first focusable item.
+   * @default the first tabbable element will get focus.
    */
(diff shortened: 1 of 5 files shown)


packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.types.tsx
```diff
@@ -8,6 +8,7 @@ export interface AutoFocusZoneProps extends React.HTMLAttributes<HTMLDivElement>
 
   /**
    * Indicates the selector for first focusable item.
+   * @default the first tabbable element will get focus.
```
layershifter: It should `undefined`?

```diff
-   * Indicates the selector for first focusable item.
-   * @default the first tabbable element will get focus.
+  Indicates the selector for first focusable item, will use the first tabbable element will get focus.
```

packages/react/src/lib/accessibility/FocusZone/FocusZone.types.ts
```diff
@@ -32,111 +32,137 @@ export interface IFocusZone {
...
-   * Function which uses root element to select and return the intial tabbable element.
+   * Function which uses root element as parameter to return the intial tabbable element.
+   * For example, when there is a chat with a bottom-up approach, it is expected that the last chat message is tabbable (active), not the first default one.
+   * @default undefined
    */
   defaultTabbableElement?: (root: HTMLElement) => HTMLElement
 
   /**
    * If a default tabbable element should be force focused on FocusZone mount.
+   * @default false
```
layershifter: I suggest to patch our plugin instead, something like `if (type === 'boolean' && default === undefined) default = false`

packages/react/src/lib/accessibility/FocusZone/FocusTrapZone.types.tsx
```diff
@@ -13,7 +13,7 @@ export interface FocusTrapZoneProps extends React.HTMLAttributes<HTMLDivElement>
 
   /**
    * Sets the HTMLElement to focus on when exiting the FocusTrapZone.
-   * @default The element.target that triggered the FTZ.
+   * @default The target which triggered the FocusTrapZone.
```
layershifter: The similar to https://github.com/stardust-ui/react/pull/1289/files#r280799691

packages/react/src/lib/accessibility/FocusZone/FocusZone.types.ts
sophieH29: sorry, didn't get 🙃

layershifter (approved):

Issue #201, opened 2018-09-06
[RFC] Naming of <Label> and <label> relative to an Input component


brianespinosa:
### Problem description
Inside of an Input component we have the concept of a Label component which attaches to the front or back of an input. This is different than the native `<label>` which appears above an input (or to the left if the input is inline). This leads to confusion for the user, especially when using the shorthand API. When I set something as the `label` prop, am I setting the html lowercase `<label>` or am I passing attributes to a `Label` component.

### Proposed solution
I think we should rename the Label component itself, relative to Inputs. This is going to reduce confusion in both discussion and semantics, but also simplify how be define these elements in the shorthand API.

Bootstrap calls these elements addons. That might be a good place to start. They also have the concept of prepend/append for positioning. I think we could just use `addonPosition` and `"before"||"after"`

Any other naming suggestions welcomed. I feel like it is early enough on that any challenges to be addressed in this name change can be handled fairly easily.

kuzhelov:
My absolute support on this proposal - also, we should make it clear that provided Stardust `Input` component has one-to-one correspondence to `input` DOM element in general (so that any input attributes supplied to `Input` component should be reflected by its look and behavior) - and, at the same time we should provide these extensibility mechanism to enhance visual look of `Input` for certain scenarios (see https://github.com/stardust-ui/react/pull/120#issuecomment-417130800). 


![image](https://user-images.githubusercontent.com/5067638/44819501-eb443400-aba1-11e8-9a7b-8ac44c2493fa.png)

![image](https://user-images.githubusercontent.com/5067638/44819507-f008e800-aba1-11e8-98f1-b43f8eb2cac8.png)

![image](https://user-images.githubusercontent.com/5067638/44819517-f5663280-aba1-11e8-88bb-3bd83f956edd.png)

![image](https://user-images.githubusercontent.com/5067638/44819524-fb5c1380-aba1-11e8-8d4d-2ae36068fa7f.png)

![image](https://user-images.githubusercontent.com/5067638/44819537-01ea8b00-aba2-11e8-917c-682482c280c7.png)

![image](https://user-images.githubusercontent.com/5067638/44819540-07e06c00-aba2-11e8-8f9d-92993401d6f9.png)

![image](https://user-images.githubusercontent.com/5067638/44819548-0f077a00-aba2-11e8-922e-f5766050c302.png)

[...]

PR #1290, opened 2019-05-06
feat: add strict type checks for components' props

### LINKED ISSUE #1263: Props type safety
### Problem description
Props are not type safe, for example this is a valid code that compiles using Typescript:
<Button title={{test: true}} ariaLabel={{test: true}} />

This has some implications to how we are doing translations (outside of Stardust scope) where there is one function that returns a string and one that returns React.Node intended for rendering the subtree for RTL purposes.

With no type checking on props like title or aria-label, we do not have any realiable way to enforce usage of the correct functions

### Proposed solution
Provide typings at least for the common html props similar to the types provided by "@types/react"

### MVP
https://codesandbox.io/s/rl9068mvro

kuzhelov:
Fixes #1263.

# BREAKING CHANGES

Restricted type checks introduced by this PR may introduce breaking changes on the client side - however, **most certain this will be an indication of problem in client's logic**. 

However, to mitigate potential migration problems, the following fallback option was introduced, to loose type checks to the same state they were working before.

## Fallback option to loose type checks
**In the exceptional cases**, to mitigate migration risks on the client side, the fallback type was introduced as part of this PR. This type corresponds to the one that clients originally had for Stardust components before this update.

```js
// re-export on client side
import { 
  Stardust_Button, 
  ButtonProps, 
  UNSAFE_TypedComponent, 
  UNSAFE_typed 
} from '@stardust-ui/react'

// either explicit cast
const Button = Stardust_Button as any as UNSAFE_TypedComponent<
  typeof Stardust_Button, 
  ButtonProps
>

// or simpler
const Button = UNSAFE_typed(Stardust_Button).withProps<ButtonProps>()

export Button
```

-------------

# Description

Provides strict type checks for Stardust components' props. 

## Type checks introduced

Examples of type checks introduced are provided with `Button` component taken, which supports
- `circular: boolean` prop
- `formEncType: string` prop from the interface of intrinsic `button` element, which is rendered by default as `Button`'s element type.

#### No 'as' prop defined (regular case)
```jsx
<Button />

<Button circular />
<Button circular={13} />  // BAD, type mismatch (number instead of boolean)

<Button formEncType='...' />  // fine, as 'formEncType: string' is a prop of intrinsic 'button'
<Button formEncType />  // BAD, type mismatch (boolean instead of string)

<Button asdasdasd />  // BAD, unknown prop 'asdasdasd'
```

#### With 'as' prop defined
```jsx
<Button as={'h1'} />

// fine, ANY prop is allowed in addition to Button's ones. 
// this is what we had before for all the cases of Stardust components' usage
<Button as={'h1'} asdasdasd />   

// checks for ButtonProps are preserved
<Button as={'h1'} circular={13} />   // BAD, type mismatch (number instead of boolean)
```

Type checks for `as` case were intentionally loosened, to avoid `heap out-of-memory` problem of TS compiler (more on that below). However, even with that **provided type system was sufficient enough to capture all the same type problems in Stardust codebase that would be discovered by the strictest type-check solution**, and generally is expected to prevent most of type problems on the client side.

## Stardust problems discovered (just files list, for brevity)
 - docs/src/components/ComponentDoc/ComponentControls/ComponentControls.tsx
 - docs/src/components/ComponentDoc/ComponentDocSee.tsx
 - docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
 - docs/src/components/ComponentDoc/ComponentSidebar/ComponentSidebar.tsx
 - docs/src/components/ComponentDoc/ComponentSidebar/ComponentSidebarSection.tsx
 - docs/src/examples/components/ItemLayout/Types/ItemLayoutExampleSelection.shorthand.tsx
 - docs/src/examples/components/Portal/Types/PortalExample.shorthand.tsx
 - docs/src/examples/components/Portal/Types/PortalExample.tsx
 - docs/src/examples/components/Portal/Types/PortalExampleControlled.shorthand.tsx
 - docs/src/examples/components/Portal/Types/PortalExampleControlled.tsx
 - docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.shorthand.tsx
 - docs/src/examples/components/Portal/Types/PortalExampleFocusTrapped.tsx
 - docs/src/prototypes/AsyncShorthand/AsyncShorthand.tsx
 - docs/src/prototypes/IconViewer/index.tsx
 - docs/src/prototypes/chatPane/chatPaneContent.tsx
 - docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
 - docs/src/views/PageNotFound.tsx
 - docs/src/views/QuickStart.tsx
 - packages/react/src/components/Attachment/Attachment.tsx
 - packages/react/src/components/Chat/ChatMessage.tsx
 - packages/react/src/components/Dropdown/DropdownItem.tsx
 - packages/react/src/components/Dropdown/DropdownSearchInput.tsx
- packages/react/src/components/Provider/Provider.tsx
- packages/react/src/components/Provider/ProviderBox.tsx
 - build/gulp/tasks/test-projects/cra/App.tsx
   - scaffolding file for `create-react-app` test project

## Additional cleanup
- obsolete `...WithDefault` types were removed from: `Divider`, `ItemLayout`, `Layout`
- types for components that doesn't support `as` prop by design were updated: `Provider`, `Popup`, `Portal`, `PortalInner`

## Main type files changed
 - packages/react/src/lib/commonPropInterfaces.ts
 - packages/react/src/types.ts

## Notes on the approach

The following things were taken into account in the solution provided.

### Avoid 'heap out of memory' problem from TS Compiler

In order to prevent cause of this TS compiler's problem (which repros even for the latest to date, v3.4.5 of TS compiler), it was necessary to loose type checks for the cases when `as` prop is explicitly specified by client's code. 

However, even with that provided type system was sufficient enough to capture all the same type problems in Stardust codebase that would be discovered by the strictest type-check solution, and generally is expected to prevent majority of type problems on the client side.

Introduced types don't cause 'heap out of memory' problem for TS compiler - as component type overloads do not cause deep recursion like the following one, while processed by TS engine:

image

### Explicit prop type declarations in component types
Exported components are wrapped in `withSafeTypeForAs` type that requires type of component props to be explicitly provided in generic args:

```jsx
export default withSafeTypeForAs<typeof Button, ButtonProps, 'button'>(Button)
```

While component prop types could be inferred, this approach was taken because of its huge performance benefits introduced to TS Server, as it helps to avoid unnecessary inference and significantly reduce load on TS server. With that, we are making things explicit only once while exporting component, and reap performance benefits of rapid TS checks every time it is used.

These are comparison metrics between 'solution with inference' and 'solution with no inference'

| Metric | With Props Inference | With Explicit Props |
|---|---|---|
| **reaction from IDE (its TS server) on key press while typing prop name** | 8-10s | 0.5-1s |
| `yarn build` time | 4.82 - 5 min | 2.8-2.98 min

------------------

## Non-related problems spotted
- `Embed` has description for GIF (#1330)


build/gulp/tasks/test-projects/cra/App.tsx
```diff
@@ -25,7 +25,7 @@ class App extends React.Component {
             <Icon name="umbrella" circular bordered />
           </Animation>
           <Attachment header="Document.docx" />
-          <Avatar src="//placehold.it" />
+          <Avatar image={{ src: '//placehold.it' }} />
```
layershifter:
```suggestion
          <Avatar image='//placehold.it' />
```

It seems that it can simplified...

docs/src/app.tsx
```diff
@@ -19,7 +19,6 @@ class App extends React.Component<any, ThemeContextData> {
     return (
       <ThemeContext.Provider value={this.state}>
         <Provider
-          as={React.Fragment}
```
layershifter: Why this change is required? It will create an additional `div` element...

docs/src/components/ComponentDoc/ComponentControls/ComponentControls.tsx
```diff
@@ -64,12 +64,10 @@ const ComponentControls: React.FC<ComponentControlsProps> = props => {
   } = props
 
   return (
-    <Provider theme={controlsTheme} as={React.Fragment}>
```
layershifter: Same thing, https://github.com/stardust-ui/react/pull/1290/files#r282797120

docs/src/components/ComponentDoc/ComponentProps/ComponentProps.tsx
```diff
@@ -59,8 +59,10 @@ export default class ComponentProps extends React.Component<any, any> {
             <Flex.Item styles={{ display: 'inline-block' }}>
               <>
                 <Input
+                  input={{
```
layershifter: I can't agree with this change because it currently works, we use `partitionHTMLProps` to pass down correct props to `input`.

docs/src/components/Logo/Logo.tsx
```diff
@@ -3,6 +3,6 @@ import { Image } from '@stardust-ui/react'
 
 const Logo: any = props => <Image {...props} src="logo.png" />
 
-Logo.propTypes = Image.propTypes
+Logo.propTypes = (Image as any).propTypes
```
layershifter: I suggest to remove this line at all

docs/src/components/ComponentDoc/ComponentControls/ComponentControls.tsx
layershifter: This change breaks styling, BTW

image

layershifter:
It's not critical, but it seems that autocomplete is not working 😭 

not-work

docs/src/app.tsx
kuzhelov: because provider hasn't support for `as` - this has been detected by type checks:

```jsx
Provider extends React.Component<...>
```

docs/src/components/ComponentDoc/ComponentProps/ComponentProps.tsx
kuzhelov: types wouldn't agree with the original implementation, as default element of `Input` is `div` which is rendered, and it hasn't `checked` attribute

kuzhelov:
> It's not critical, but it seems that autocomplete is not working 😭

This doesn't work for VSCode at all for `aria-` attributes, thus would skip this problem as the one that is IDE-related. At the same time, type checks work correctly, and this is the most important thing there.

docs/src/components/ComponentDoc/ComponentProps/ComponentProps.tsx
kuzhelov: will augment InputProps with the supported props from `htmlPropsUtils.ts`

however, `partitionHTMLProps ` seems a serious complication, could we avoid it? What prevents us from rendering `Input` as `input`, with no `div` wrapper? Even from client's perspective it would be the anticipated thing

docs/src/app.tsx
layershifter: But it's supported by `ProviderBox` and unhandled props will be passed there

kuzhelov: this should be resolved differently, as this case only will require us to open gate for any unhandled prop being passed to `Provider`. Again, it seems to be an example of some implicit assumptions (aka magic) that we'd better avoid - if `as` prop is supported by `Provider`, then it should be part of its props interface

docs/src/components/ComponentDoc/ComponentProps/ComponentProps.tsx
layershifter: > should we define input as default element type for Input? 

Nope, because default is `div` and it will be used for the `wrapper` slot.

> partitionHTMLProps seems a serious complication, could we avoid it?

This will mean that all attributes should be passed to `input` slot. So we it will be something like this: `<Input input={{ readOnly: true }} />`. At least it looks weird, from other side it means that `readOnly` will be absent in styles and accessibility.

kuzhelov: if `Input` would be rendered as `input`, why `<Input input='..' />` (i.e. `input` slot) is necessary then?

layershifter: Can you please clarify how [this example](https://stardust-ui.github.io/react/components/input#variations-icon) can be rendered with `<Input as=input />`?

kuzhelov: there is no way to render this with plain Input, yes, but we could create higher-level `DecoratedInput` component (or the one with other name) which would work for this scenario.

Key thing is that, as a client, I would expect Stardust Component to be equivalent to DOM Element in case if name matches - i.e. I may pass all the same props, expect all the same attributes to provide, etc. Real use-cases and problems from Stardust clients suggest the same expectations: https://github.com/stardust-ui/react/issues/201#issuecomment-419484862

This is what drives me towards this thinking - in either case, original issue is now addressed, this should be taken as a separate issue.

layershifter (approved):
I don't want to block these changes because I see benefits from them.

However I am thinking that changes in run-time code to satisfy TypeScript is something evil 💣

kuzhelov:
> However I am thinking that changes in run-time code to satisfy TypeScript is something evil

would want to avoid them as well, but for now, given that we have class components and there the only option are generics, as well as that it is impossible to introduce function overload semantics with regular TS types, I don't see any better approach there - and noop runtime tricks is the only way to deal with that.

Also, in case if solution will be found (introduced by TS), it will be quite easy for us to adopt it, as this type is applied consistently to all components, so we might be able to introduce necessary fixes to it.

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -198,13 +198,7 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
...
     }
 
-    const providerVariables = showTransparent ? { background: 'initial' } : undefined
-
-    return (
-      <Provider theme={newTheme} variables={providerVariables}>
-        {element}
-      </Provider>
-    )
+    return <Provider theme={newTheme}>{element}</Provider>
```
miroslavstastny: This breaks the possibility to render examples on transparent background:
image

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -236,4 +231,9 @@ class Menu extends AutoControlledComponent<ReactProps<MenuProps>, MenuState> {
 
 Menu.create = createShorthandFactory({ Component: Menu, mappedArrayProp: 'items' })
 
-export default Menu
+/**
+ * A menu displays grouped navigation actions.
+ * @accessibility
+ * Implements ARIA Menu, Toolbar or Tabs design pattern, depending on the behavior used.
+ */
+export default withSafeTypeForAs<typeof Menu, MenuProps, 'ul'>(Menu)
```
miroslavstastny, line 239: The following line should fail but it compiles without any error:
```jsx
const MenuExample = () => <Menu as="ul" defaultActiveIndex={0} items={items} invalidAttribute />
```

kuzhelov: this 'loose types for as'  is intentional - as this is necessary to avoid 'JS Heap overflow' problem. More details are provided in the description in the 'Notes on the approach' section

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
kuzhelov: got it - the thing is that this `variables` prop was missing from the `Provider`'s API, and that has led me to confusion. Will address it

miroslavstastny (approved):

PR #1291, opened 2019-05-06
docs(faq): FAQ update

hughreeling:
Adding FAQ

image

Adding Default value for Form.Field

image


docs/src/views/FAQ.tsx
```diff
@@ -12,6 +12,35 @@ const Answer = props => <p>{props.content}</p>
...
+    <ul>
+      <li>
+        {link('General', '#General')}
+        <ul>
+          <li>{link('Does Stardust support mobile?', '#does-stardust-support-mobile')}</li>
+          <li>
+            {link(
+              "How can @mixins be used in the Stardust's styles?",
+              '#how-can-mixins-be-used-in-the-stardusts-styles',
+            )}
```
alinais, line 25: Maybe we can add here an example, to be more clear for users.

hughreeling: Can you suggest?

mnajdova: We can reference the `getBorderFocusStyles` function that we are already using in the button styles in Teams theme.

docs/src/views/FAQ.tsx
```diff
@@ -40,6 +69,20 @@ export default () => (
       }
     />
 
+    <Question content="How can I set default value of Form.Field?" />
+    <Answer
+      content={
+        <p>
+          You need to set the default value on the input slot.
```
mnajdova: I would rephrase this to:
`You need to set the default value on the component used in the control slot.`
The reason for this is that, in the control slot we may add input, radio, checkbox components that may implement this props in a different way.

layershifter (approved):

PR #1292, opened 2019-05-06
docs(Slots): Clearly identify Slots in DocSite

hughreeling:
Identify slot properties explicitly and mark with blue 'S'

image

Add ref to Shorthand page.

image

Identified explicitly in code:
image


build/gulp/plugins/util/docgen.ts
```diff
@@ -359,17 +360,22 @@ export class Parser {
...
       }
 
+      if (jsDocComment.tags.slot) {
+        slot = true
+      }
       const parent = getParentType(prop)
 
       result[propName] = {
         defaultValue,
+        slot,
```
layershifter: I want to omit any changes in `docgen.ts` because it was copy-pasted from `react-typescript-docgen`.

---

All of our shorthand use `ShorthandValue` or `ShorthandCollection` types, so I suggest to update our `parseType()` function to return `slot`.

hughreeling: Reverted changes to docgen - not needed, as tags are already handled.

Do we see any value in identifying slots explicitly in comments in code?  The relation between ShorthandValue  / ShorthandCollection and slots seems 'fragile' - I can image a property that takes in shorthandValue but doesn't behave fully as slot, and likewise, I could see a slot that has some restrictions on what can be passed to it....

PR #1297, opened 2019-05-07
fix(dialogs): themeing support for Teams dark and hc

### LINKED ISSUE #1313: Dialog Component breaks in Dark mode
## Bug Report

Dark Mode doesn't respect Dark Mode selection in theme.

### Steps

Go to Stardust docs site, switch to Dark Mode theme, open a Dialog. You see the following:

image

This is using a white background, and also the Buttons do not appear to be correctly themed to the dark theme.

### Expected Result
The Dark Theme is applied correctly to the Dialog.

### Actual Result
The Dark Theme is not applied correctly to the Dialog.

### Version
latest

### Testcase
Go to the Stardust docs site.

codepretty:
Fixes #1313.

This **_does not_** include all the redline work for Dialogs. This is just to fix the Dialog's color support in Teams dark and Teams high contrast themes.

_Screenshots below (Before | After)_
**Teams theme**
image

**Teams dark theme**
image

**Teams high contrast theme**
image


packages/react/src/themes/teams/components/Dialog/dialogStyles.ts
```diff
@@ -0,0 +1,47 @@
+import { DialogProps } from '../../../../components/Dialog/Dialog'
+import { ComponentStyleFunctionParam, ICSSInJSStyle } from '../../../types'
+import { DialogVariables } from './dialogVariables'
+
+type DialogStyleParams = ComponentStyleFunctionParam<DialogProps, DialogVariables>
+
+export default {
+  root: ({ props: p, variables: v }: DialogStyleParams): ICSSInJSStyle => ({
```
layershifter, line 8: As I see we have the same stuff in `base` theme, why we can't reuse it and add only missing things?
I.e. skip `actions`, `overlay` and `content` slots at all...

codepretty: I was originally copied styles to update for redlining for the component. But this PR is just going to be a color/themeing update and the redlining can follow in a separate PR. Will update the styles to not duplicate here.

packages/react/src/themes/teams-high-contrast/components/Dialog/dialogStyles.ts
```diff
@@ -0,0 +1,7 @@
+import { ICSSInJSStyle } from '../../../types'
+
+export default {
+  root: ({ variables: v }): ICSSInJSStyle => ({
+    border: `1px solid ${v.foregroundColor}`,
```
notandrew: should this have pxtorem?

notandrew, line 5: Should this be pxtorem?

packages/react/src/themes/teams-dark/components/Dialog/dialogVariables.ts
```diff
@@ -0,0 +1,11 @@
+import { DialogVariables } from '../../../teams/components/Dialog/dialogVariables'
+
+export default (siteVars: any): Partial<DialogVariables> => {
+  return {
+    boxShadow: siteVars.shadowLevel4,
+    rootBackground: '#2d2c2c', // gray[650] with new color palette
```
layershifter: #1069 was merged, can we update this?

packages/react/src/themes/teams/components/Dialog/dialogVariables.ts
```diff
@@ -0,0 +1,39 @@
+import { pxToRem } from '../../../../lib'
+
+export interface DialogVariables {
```
layershifter: I suggest to reuse type for base theme and remove duplicate values:
```
import { DialogVariables as BaseDialogVariables } from '../../'

export interface DialogVariables extends BaseDialogVariables {
```

packages/react/src/themes/teams-high-contrast/components/Dialog/dialogStyles.ts
```diff
@@ -0,0 +1,8 @@
+import { pxToRem } from '../../../../lib'
+import { ICSSInJSStyle } from '../../../types'
+
+export default {
+  root: ({ variables: v }): ICSSInJSStyle => ({
+    border: `${pxToRem(1)} solid ${v.foregroundColor}`,
```
miroslavstastny: This is the second place throughout the whole theme where we use `pxToRem` for borders. Anywhere else we intentionally used `px` for border not to scale.
So what's the correct approach from design perspective?

codepretty: i think pixels were used for borders in hc theme originally because in some browsers the border will disappear when you scale down to 50%. but the more interesting question is if we care about support of scaling down that far?

cc/ @levithomason

codepretty: Chatted with @levithomason today about this topic. For this case (and most borders) we will want to use pixels since we don't want them to scale.

packages/react/src/themes/teams/components/Dialog/dialogVariables.ts
```diff
@@ -0,0 +1,40 @@
+import { DialogVariables as BaseDialogVariables } from '../../../base/components/Dialog/dialogVariables'
+import { pxToRem } from '../../../../lib'
+
+export interface DialogVariables extends BaseDialogVariables {
```
miroslavstastny, line 4: As this extends `BaseDialogVariables` it should contain only newly added vars, not the ones inherited from Base:
```jsx
  boxShadow: string
  foregroundColor: string
  headerFontSize: string
  headerFontWeight: number
```

miroslavstastny (approved):
Just one comment added, other than that 👍

PR #1298, opened 2019-05-07
feat(Dialog): add default aria-labelledby if not provided

### LINKED ISSUE #812: Dialog: add aria-attributes
## Feature Request

Also, would it be possible to allow adding aria-labelledby (which needs an ID reference to the header)?
The logic should be:
- by default, generate `aria-labelledby`
- if `aria-labelledby` provided by the user, use that instead
- if `aria-label` provided, do not generate `aria-labelledby`

For `aria-describedby` I would leave it fully on the consumer to provide correct referecne as not all dialogs have clear description.

https://github.com/stardust-ui/react/pull/790#discussion_r252560457

### Proposed solution
```jsx
const getAriaLabeledBy = (props) => {
  if (props['aria-labelledby']) return props['aria-labelledby']
  if (typeof props.header === 'string' && props.header.id) return props.header.id
  
  return _.uniqid('dialog') // PREFIX
}
```

But, I'm not happy with it, to complex for one prop.

silviuaavram:
Fixes https://github.com/stardust-ui/react/issues/812.

If Dialog has `aria-label`, do nothing.
If Dialog has `aria-labelledbby`, do nothing.
If Dialog does not have `aria-label` or `aria-labelledby`, get `id` from `header` prop and use it as `aria-labelledby` in Dialog.
If Dialog does not have `aria-label` or `aria-labelledby`, and `header` is string or simply has no `id`, generate `id` and add it as default prop to header and as `aria-labelledby` to Dialog.


packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -177,6 +177,21 @@ class Dialog extends AutoControlledComponent<ReactProps<DialogProps>, DialogStat
     },
   })
 
+  /**
```
sophieH29: What do you think if to move this logic to `dialogBehavior`?
Since it is about adding accessibility attributes and `getDefaultAriaLabelBy` is based on `props` which are passed to behavior.
Also, to add `id` to header slot, can be done in behavior

```tsx
atrributes: {
  header: {
   id: ...
}
}
```

And use in component like
```tsx
{Header.create(header, {
            defaultProps: {
              as: 'h2',
              styles: styles.header,
          {...accessibility.attributes.header}
            },
          })}
```

Please let me know what do you think.

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -177,6 +177,21 @@ class Dialog extends AutoControlledComponent<ReactProps<DialogProps>, DialogStat
...
+   * In case it generates an id, its uses will ensure that the id on the header
+   * and the value of aria-labelledBy on the dialog will be the same.
+   */
+  private getDefaultAriaLabelledBy() {
+    const { header } = this.props
+    if (this.props['aria-label'] || this.props['aria-labelledby'] || !header) {
+      return undefined
+    }
+    if (typeof header === 'string' || !header['id']) {
```
sophieH29: if the header is a string type there is never gonna be `header[id]`, isn't it? or I might misunderstand smth

packages/react/src/components/Dialog/Dialog.tsx
silviuaavram: It's how I started it, but I did not know if we should handle the header ID in the behavior, in the case we had to generate it. If we think it's acceptable, I will move all of it back to the dialog behavior, it's cleaner :)

packages/react/src/components/Dialog/Dialog.tsx
silviuaavram: does it cover the case when header is object but does not have `id` ?

silviuaavram: oh probably you are right. `!header['id']` should cover both. great!

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -177,6 +177,21 @@ class Dialog extends AutoControlledComponent<ReactProps<DialogProps>, DialogStat
...
+      return _.uniqueId('dialog-header-')
```
jurokapsiar: this will generate a different id on each render. it is not necessarily bad but maybe we can come up with a solution that would just use one for each render.

Slightly contradicting to this but more important - can we move the logic to the behavior and do similar logic for describedby? I just noticed in the issue that it explicitly says that we should not touch describedby - but that goes against what ARIA specifies: https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

If we want to omit describedby, we can have an alternative behavior that just adds labelledby.

layershifter: > this will generate a different id on each render.

Hooks implementation: https://github.com/reach/reach-ui/blob/master/packages/auto-id/src/index.js

silviuaavram: I am confused. we want `same` id for each render or `different` id on each render?

packages/react/src/lib/accessibility/Behaviors/Dialog/dialogBehavior.ts
```diff
@@ -15,12 +16,50 @@ import popupFocusTrapBehavior from '../Popup/popupFocusTrapBehavior'
...
+const getDefaultAriaLabelledBy = (props: any) => {
+  const { header } = props
+  if (props['aria-label'] || props['aria-labelledby'] || !header) {
+    return undefined
+  }
+  return header['id'] || _.uniqueId('dialog-header-')
+}
+
+/**
+ * Returns the element id of the header or generates a default one. Is
```
sophieH29: - `id of the content`
- `provide aria-describedby`
- `content id`

packages/react/src/lib/accessibility/Behaviors/Dialog/dialogBehavior.ts
```diff
@@ -15,12 +16,50 @@ import popupFocusTrapBehavior from '../Popup/popupFocusTrapBehavior'
...
+ * used when user does not provide aria-label or aria-labelledby as
```
sophieH29: `provide aria-describeby`

packages/react/src/lib/accessibility/Behaviors/Dialog/dialogBehavior.ts
silviuaavram: that's what happens when you copy paste without thinking, kids.

sophieH29 (approved):

packages/react/src/lib/accessibility/Behaviors/Dialog/dialogBehavior.ts
```diff
@@ -1,5 +1,6 @@
 import { Accessibility } from '../../types'
 import popupFocusTrapBehavior from '../Popup/popupFocusTrapBehavior'
+import * as _ from 'lodash'
 
 /**
  * @description
```
jurokapsiar, line 6: specification with unit tests would be preferable, but if not, at least add the description

silviuaavram: I tried to add specs and another definition. Please take a look.

jurokapsiar (approved):

PR #1299, opened 2019-05-07
feat(Dropdown): update styles for dark and hc Teams themes

### LINKED ISSUE #1215: Theming of Dropdown component
## Bug Report
The dropdown component has following theming issues:

**Dark Theme**
* Results have white background and text
image

**High Contrast Theme**
* Input field has white text on light background
* Results are shown as white text on white background
image

### Version
0.27 - reproducible on doc site

### LINKED ISSUE #1254: Dropdown handler overrides should be refactored
At the moment there are probably duplicated. 

The handler should be `user input` + `stardust logic` + `downshift logic` and the resulting event should be added once.

As this logic was never required before, it was implemented as is, but should be refactored in order to make sure issues will not occur and performance is not impacted.

### LINKED ISSUE #1329: Adding onClick event on the triggerButton of the Dropdown is breaking the behavior of it
### Steps
Add toggleButton prop on the Dropdown with custom onClick event.

### Expected Result
The toggleButton should still open the Dropdown when clicked, and should invoke the user's onClick event

### Actual Result
Clicking on the trigger button is not opening the Dropdown, and the user's onClick handler is never invoked.

### Version
0.30.0

### Proposed solution
when creating the toggleButton, the events should be written in the overrideProps section. There we can invoke the Downshift's handlers, and then the user's specified handlers.

```
  private renderTriggerButton(
    styles: ComponentSlotStylesInput,
    rtl: boolean,
    getToggleButtonProps: (options?: GetToggleButtonPropsOptions) => any,
  ): JSX.Element {
    const { triggerButton } = this.props
    const content = this.getSelectedItemAsString(this.state.value)

    const toggleButtonProps = getToggleButtonProps({
        onKeyDown: e => {
        },
        'aria-label': content,
      })

    const { onClick, onFocus, onBlur, onKeyDown, ...restToggleButtonProps } = toggleButtonProps

    return (
      <Ref innerRef={this.buttonRef}>
        {Button.create(triggerButton, {
          defaultProps: {
            className: Dropdown.slotClassNames.triggerButton,
            content,
            fluid: true,
            styles: styles.triggerButton,
            ...restToggleButtonProps,
          },
          overrideProps: (predefinedProps: IconProps) => ({
            onClick: e => {
              onClick(e)
              _.invoke(predefinedProps, 'onClick', e, predefinedProps)
            },
            onFocus: e => {
              onFocus(e)
              _.invoke(predefinedProps, 'onFocus', e, predefinedProps)
            },
            onBlur: e => {
              onBlur(e)
              _.invoke(predefinedProps, 'onBlur', e, predefinedProps)
            },
            onKeyDown: e => {
              onKeyDown(e)
              _.invoke(predefinedProps, 'onKeyDown', e, predefinedProps)
            },
          }),
        })}
      </Ref>
    )
  }
```

### Testcase
https://codesandbox.io/s/10k3m855j

mnajdova:
Fixes https://github.com/stardust-ui/react/issues/1215 - Dropdown theming.
Fixes https://github.com/stardust-ui/react/issues/1329 - onClick is breaking the triggerButton behavior.
Fixes https://github.com/stardust-ui/react/issues/1254 - Dropdown event handlers should be refactored.

In order for the styles to applied correctly, I've added the following things:
- `selected` prop for the `DropdownItem`, which is set to true from the `Dropdown`, if the selection is single, and the item is the selected one
- `isFromKeyboard` prop for the `DropdownItem`, which is set from the `Dropdown`, based on the type of event received from Downshift, when the item is set to active
- `isFromKeyboard` state for the `Dropdown`

Thins that will be addressed in separte PRs:
- adding `inverted` prop to the Dropdown
- adding `selected` icon for the selected list item (may be implemented differently)


packages/react/src/themes/teams/components/Dropdown/dropdownItemStyles.ts
```diff
@@ -3,12 +3,22 @@ import { DropdownVariables } from './dropdownVariables'
...
+    border: `1px solid transparent`,
     backgroundColor: v.listItemBackgroundColor,
+    ...(p.selected && {
+      fontWeight: v.listItemSelectedFontWeight,
+      color: v.listItemSelectedColor,
+    }),
     ...(p.active && {
-      color: v.listItemColorActive,
-      backgroundColor: v.listItemBackgroundColorActive,
+      ...(p.isFromKeyboard && {
```
codepretty: Does the new focus solution @Bugaa92 setup in #1269 not work here?

codepretty:
Click anywhere on the dropdown doesn't seem to work anymore

dropdown-not-working

packages/react/src/themes/teams/components/Dropdown/dropdownVariables.ts
```diff
@@ -1,46 +1,74 @@
...
+  selectedItemColor: siteVars.bodyColor,
   comboboxPaddingButton: `0 ${_12px_asRem}`,
   comboboxFlexBasis: pxToRem(50),
   listBackgroundColor: siteVars.colors.white,
-  listBorderRadius: _3px_asRem,
+  listBorderRadius: `0 0 ${_3px_asRem} ${_3px_asRem}`,
+  listBorderColor: 'transparent',
+  listBorderWidth: '0px',
   listPadding: `${pxToRem(8)} 0`,
   listBoxShadow: `0 .2rem .6rem 0 ${siteVars.colors.grey[250]}`,
```
codepretty: The `listBoxShadow` looks wrong
image

packages/react/src/themes/teams/components/Dropdown/dropdownItemStyles.ts
mnajdova: Unfortunately not, will sync today again with @Bugaa92 on this.

mnajdova:
> Click anywhere on the dropdown doesn't seem to work anymore

@codepretty acknowledge the problem, the worse part is that the tests didn't caught this :(

packages/react/src/themes/teams/components/Dropdown/dropdownVariables.ts
mnajdova: This was there before, so I wasn't sure whether it should be changed. Will check the designs again.

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -60,19 +69,33 @@ const dropdownStyles: ComponentSlotStylesInput<DropdownPropsAndState, DropdownVa
...
     color: v.color,
     backgroundColor: v.backgroundColor,
-    ...(p.focused && { borderBottomColor: v.borderColorFocus }),
+    ':hover': {
+      backgroundColor: v.backgroundColorHover,
+      [`& .${Dropdown.slotClassNames.triggerButton}`]: {
+        // reset all styles
+      },
+    },
+    ...(p.focused && {
```
mnajdova, line 90: Found out what is causing the issue with the trigger button, but no idea why :( It's the styles applied when the container of the Dropdown is focused:
```
  ...(p.focused && {
     ...(p.search && { borderBottomColor: v.borderColorFocus }),
     ...(!p.search &&
        !p.open &&
        getBorderFocusStyles({
          siteVariables,
          isFromKeyboard: true,
        })[':focus']),
    }),
```
@Bugaa92 am I using this function as intended? In the Dropdown container, we are using the `focused` prop as an indicator whether it is focused. FYI @silviuavram

packages/react/src/themes/teams/components/Dropdown/dropdownItemStyles.ts
mnajdova: This is how the new focus solution looks: 
image
More important that it is breaking the accessibility (on keys up and down, this border is not moving to the right list element). @Bugaa92 FYI

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
mnajdova: Maybe the reason for breaking this is the added pseudo elements: `::before` and `::after`

packages/react/src/themes/teams/components/Dropdown/dropdownItemStyles.ts
mnajdova: Manage to fix this, the `position: 'relative'` is necessary for the element where the border focus styles are applied. Should we add this somehow in the border focus styles?

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -466,15 +482,26 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
-              onFocus: this.handleTriggerButtonOrListFocus,
-              onBlur: this.handleTriggerButtonBlur,
-              onKeyDown: e => {
-                this.handleTriggerButtonKeyDown(e, rtl)
-              },
-              'aria-label': content,
-            }),
+            ...restToggleButtonProps,
           },
+          overrideProps: (predefinedProps: IconProps) => ({
```
mnajdova, line 483: @silviuavram can you take a look on this changes? They should fix the issue: https://github.com/stardust-ui/react/issues/1329

packages/react/src/themes/teams/components/Dropdown/dropdownVariables.ts
mnajdova: This is fixed now, please take another look.

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
mnajdova: I fixed this, by adding isFromKeyboard in the Dropdown state. Let's see if this will turn around again as an issue.

mnajdova:
The issue with the trigger button is resolved, everything works as expected now.

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: I think there is an issue created by me to refactor these handlers, probably this is the way to go. It was confusing for me at the time. We have user handler, stardust behavior, Downshift handler.

https://github.com/stardust-ui/react/issues/1254

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -587,6 +623,8 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
         defaultProps: {
           className: Dropdown.slotClassNames.item,
           active: highlightedIndex === index,
+          selected: !this.props.multiple ? selectedIndex === index : false,
```
silviuaavram: do we really need this logic? can't we just say `selected: !this.props.multiple && value === item` ? Even the first check is not needed, as when it's multiple it will never be equality between the item and the array.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -765,6 +808,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
       searchInputProps: DropdownSearchInputProps,
     ) => {
       this.setState({ focused: false })
+      this.setState({ isFromKeyboard: isFromKeyboard() })
```
silviuaavram: don't forget to batch these up

packages/react/src/components/Dropdown/Dropdown.tsx
mnajdova: Yep, maybe they are duplicated. At least those should be resolved with these changes.

packages/react/src/components/Dropdown/Dropdown.tsx
mnajdova: I am using the `changes.selectedItem` which is always the last selected item (if I use the value I will always have to do checks whether it is array or not, and do casting - which in my opinion is more prone for bugs in the future). I can rewrite this to: `selected: !this.props.multiple && selectedIndex === index`, I agree that it is more readable. Are you ok with this?

silviuaavram: If it works I am ok, I just thought that simply doing `value === item` will be enough. If it's multiple, it's false. If it's single, it will compare objects and if those are equal then we are good.

Why would you need to enforce array checking?

mnajdova: Changed

packages/react/src/themes/teams-dark/components/Dropdown/dropdownVariables.ts
```diff
@@ -0,0 +1,22 @@
+import { DropdownVariables } from '../../../teams/components/Dropdown/dropdownVariables'
+
+export default (siteVars): Partial<DropdownVariables> => ({
+  backgroundColor: siteVars.colors.grey[650],
+  backgroundColorHover: siteVars.colors.grey[550],
+  borderColor: 'transparent',
+  borderColorFocus: siteVars.colors.brand[400],
+  color: siteVars.colors.grey[250],
+  selectedItemColor: siteVars.colors.grey[250],
+  listBoxShadow: `0 .2rem .6rem 0 rgba(0, 0, 0, 0.25)`,
```
codepretty: Can we add the box shadow value to siteVariables?

mnajdova: Used siteVariables.shadowLevel3, which is added in dark theme as well: `export const shadowLevel3 = '0 .8rem 1rem -0.2rem rgba(0, 0, 0, .25)'` In high contrast there is no shadow on the list

CHANGELOG.md
```diff
@@ -21,7 +21,11 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Restrict Typescript checks for component props @kuzhelov ([#1290](https://github.com/stardust-ui/react/pull/1290))
 
 ### Fix
-- Fixed Flex.Item children prop type @mnajdova ([#1320](https://github.com/stardust-ui/react/pull/1320))
+- Fixed `Flex.Item` children prop type @mnajdova ([#1320](https://github.com/stardust-ui/react/pull/1320))
+
+### Features
+- Add `DropdownItem` `selected`, `isFromKeyboard` props @mnajdova ([#1299](https://github.com/stardust-ui/react/pull/1299))
```
kuzhelov: nit: just to make it more clear, may be rephrased as:

> Add  `selected`, `isFromKeyboard` props to `DropdownItem`

Otherwise it reads like `DropdownItem` is an added prop as well

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -308,6 +311,8 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
       highlightedIndex: this.props.highlightFirstItemOnOpen ? 0 : null,
       searchQuery: search ? '' : undefined,
       value: multiple ? [] : null,
+      isFromKeyboardItem: false,
```
kuzhelov: nit: may suggest to rename it to something like `itemIsFromKeyboard`, to better reflect semantics of this field - mainly due to `keyboardItem` is quite confusing

packages/react/src/themes/teams-high-contrast/components/Dropdown/dropdownStyles.ts
```diff
@@ -0,0 +1,44 @@
+import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
+import {
+  default as Dropdown,
+  DropdownProps,
+  DropdownState,
+} from '../../../../components/Dropdown/Dropdown'
+import { DropdownVariablesHOC } from 'src/themes/teams-high-contrast/components/Dropdown/dropdownVariables'
```
bmdalex: this should be a relative import

packages/react/src/themes/teams/components/Dropdown/dropdownItemStyles.ts
```diff
@@ -1,14 +1,29 @@
...
     backgroundColor: v.listItemBackgroundColor,
+    ...(p.selected && {
+      fontWeight: v.listItemSelectedFontWeight,
+      color: v.listItemSelectedColor,
+    }),
+    position: 'relative',
     ...(p.active && {
-      color: v.listItemColorActive,
-      backgroundColor: v.listItemBackgroundColorActive,
+      ...(p.isFromKeyboard && {
```
bmdalex: can we pass `p.isFromKeyboard ` as argument to `getBorderFocusStyles`?

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -453,6 +459,17 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
     const { triggerButton } = this.props
     const content = this.getSelectedItemAsString(this.state.value)
 
+    const toggleButtonProps = getToggleButtonProps({
```
kuzhelov: is `toggle` and `trigger` button the same thing? Seems that yes, given where props of `toggleButton` are applied to. Would suggest to maintain naming consistency in this case

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -582,6 +614,8 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
         defaultProps: {
           className: Dropdown.slotClassNames.item,
           active: highlightedIndex === index,
+          selected: !this.props.multiple && value === item,
```
kuzhelov, line 617: this `value === item` is based on potentially dangerous assumption that

```js
const filteredItems = this.getItemsFilteredBySearchQuery()
```

identity of each `item` in `this.props.items` won't be changed. However, given that client's code is providing this logic, it is possible that this won't be true, and chances are higher in case if any library based on immutability principles is used (such as Redux). 

Simplest example is:

```jsx
<Dropdown search={items => [...items]} />
```

bmdalex (approved):
👍 awesome changes, GG @mnajdova 
just couple of comments

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -680,7 +714,9 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
     }
 
     if (this.state.open && _.isNumber(changes.highlightedIndex)) {
+      const isFromKeyboardItem = changes.type !== Downshift.stateChangeTypes.itemMouseEnter
```
kuzhelov: could you, please, just clarify our assumptions here? Am I right that

- if we have state change for `Dropdown`, which is change for its `open` value
- and this `open` value is `true`
- and, if also we have `highlightedIndex` being number in changes (does it imply that this has been changed as well - I mean, otherwise it would be non-defined, right?)
- as well as changes are not caused by `MouseEnter`

then we providing `itemIsFromKeyboard` value to item, which is an indicator of item being selected from keyboard?

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -453,6 +459,17 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
+        this.handleTriggerButtonKeyDown(e, rtl)
+      },
+      'aria-label': content,
+    })
+
+    const { onClick, onFocus, onBlur, onKeyDown, ...restToggleButtonProps } = toggleButtonProps
+
     return (
       <Ref innerRef={this.buttonRef}>
         {Button.create(triggerButton, {
```
kuzhelov, line 475: this is another good example of why we  should refrain from blindly passing all calculated props to _any_ slot's content - we might imagine what it would be if regular `<button />`, or even any other Stardust component type which is different from `Button` will be provided here

packages/react/src/components/Dropdown/Dropdown.tsx
mnajdova: `getToggleButtonProps` as a method comes from Downshift, and we are using this method on the `triggerButton` - main area of the single selection Dropdown, and the `toggleIndicator`, the arrow on the right side of the `Dropdown`. Can you please let me know what should be renamed in this case?

packages/react/src/components/Dropdown/Dropdown.tsx
mnajdova: I had this prev implemented with using state's selectedIndex to track this. Do you think this will be better option?

packages/react/src/components/Dropdown/Dropdown.tsx
mnajdova: It goes like this:
- if we have state change for Dropdown and the dropdown is open
- if the highlighted index was changed
then we are providing the itemIsFromKeyboard to the DropdownItem (if it is from mouse, this will be false, otherwise it will be true)

packages/react/src/components/Dropdown/Dropdown.tsx
kuzhelov: I think that better one would be to provide some unique key to each item (if those are objects), like `_someKey` prop, the one that will be used for identity check. Something like the following (pseudo):

```jsx
const getItemsFilteredBySearchQuery = () => { 
  const itemsToFilter = augmentWithSomeKey(this.props.items)
  ...
  const filteredItems = search(itemsToFilter)
  return filteredItems
}
```

and then, before rendering, strip this helper `_someKey` prop:

```jsx
const renderItems = () => {
  const filteredItemsWithNoAugmentedProps = stripAugmentedProps(filteredItems)
  .. render filteredItemsWithNoAugmentedProps ..
}
```

This would be the safest option.

**However, have made another thought on that - given that we are comparing items of the array, the chance is pretty low that _item itself_ will be changed. Thus, I think that the strategy you've provided is sufficient enough, at least till the moment when we would spot real problem with that.**

So, lets proceed with your solution, thanks!

packages/react/src/components/Dropdown/Dropdown.tsx
kuzhelov: given that method is named `renderTriggerButton`, as well as that props contained in _scoped_ variable `toggleButtonProps` are applied to trigger button only in the context of this method, I would rename it as `triggerButtonProps` - yes, we have `getToggleButtonProps` as method from `Downshift`, but in the context of our method, where everything is scoped to this method, I would expect to see everything related to `triggerButton`, accordingly

kuzhelov (approved):

packages/react/src/components/Dropdown/Dropdown.tsx
mnajdova: Got it, will rename the const to triggerButtonProps.

PR #1306, opened 2019-05-09
fix(Input): clear by Escape keydown when Input is clearable

### LINKED ISSUE #1024: [Accessibility] Input component "clear" button is not keyboard accessible
### Steps
```jsx
<Input clearable placeholder="Search..." />
```

### Expected Result
Be able to clear the input content with the keyboard

### Actual Result
"Clear" button is not accessible with the keyboard

### Version
0.23.0

silviuaavram:
Fixes https://github.com/stardust-ui/react/issues/1024


packages/react/src/components/Input/Input.tsx
```diff
@@ -179,7 +181,15 @@ class Input extends AutoControlledComponent<ReactProps<InputProps>, InputState>
       this.inputRef.current.focus()
       _.invoke(predefinedProps, 'onClick', e, this.props)
     },
-    ...(predefinedProps.onClick && { tabIndex: '0' }),
+  })
+
+  private handleInputOverrides = predefinedProps => ({
+    onKeyDown: (e: React.SyntheticEvent) => {
```
jurokapsiar: can we put this to a behavior?

jurokapsiar: also, if value was not equal to '' before, we should stop propagation. because if you put input to a popup, popup will get closed (instead of clearing the input).  if the user really wants to close the popup, he can press esc twice.

silviuaavram: I've moved it, take a look again.

About the stopping of the propagation, I tried to do it but the popup still closes. No idea why, the Popup behavior function will still run even with the event that had the stopPropagation applied to it. I added a breakpoint in `closeAndFocusTrigger` from popup behavior, and the event `isPropagationStopped` returns true.

jurokapsiar (approved):

PR #1309, opened 2019-05-09
docs(Menu): add playground for component

layershifter:
This PR adds a playground for `Menu` component and adds `useAccessibilityKnob` to docs.

---

image


docs/src/components/ExampleSnippet/renderElementToJSX.ts
```diff
@@ -14,6 +14,7 @@ const renderElementToJSX = (element: React.ReactNode, triggerErrorOnRenderFn: bo
       (typeof element.type === 'function' ? 'NoDisplayName' : element.type),
     showDefaultProps: false,
     showFunctions: true,
+    filterProps: ['accessibility'],
```
layershifter, line 17: As it function we can't display it properly in markup...

jurokapsiar (approved):

PR #1312, opened 2019-05-09
feat(dropdown): align, position, offset props + automatic position

bmdalex:
## BREAKING CHANGES MITIGATION
Replace `Dropdown` variables:

- ~~`borderRadius`~~ :arrow_right: `containerBorderRadius`
- ~~`openBorderRadius`~~ :arrow_right: `openAboveContainerBorderRadius` and `openBelowContainerBorderRadius`
- ~~`listBorderRadius`~~ :arrow_right: `aboveListBorderRadius` and `belowListBorderRadius` 

#### This PR tackles:
- introducing `align`, `position`, `offset` props for `Dropdown` component
- automatic positioning for `Dropdown` results
- reuses code from #1153 to avoid double render and simplify code

Screen Recording 2019-05-10 at 18 05 46

#### TODO:
- [x] Refactor method names and clean code
- [x] Extract common code for positioning elements to `src/lib/positioner`
- [x] Integrate new `Positioner` helper component to `Popup` and `Dropdown`
- [x] Refactor unit tests for `positionerHelper` (move from `Popup-test.tsx` to dedicated file)
- [x] Figure out way how to add some extra padding between `Dropdown` trigger and list of results
**any ideas?**


packages/react-component-ref/src/Ref.tsx
```diff
@@ -14,7 +14,7 @@ export interface RefProps {
    *
    * @param {HTMLElement} node - Referred node.
    */
-  innerRef: React.Ref<any>
```
layershifter: This change is not correct. `innerRef` will be passed to `RefForwardRef` and there are no guarantees that it will be `HTMLElement`

docs/src/examples/components/Dropdown/Variations/DropdownExamplePosition.shorthand.tsx
```diff
@@ -0,0 +1,38 @@
+import * as React from 'react'
+import { Grid, Dropdown } from '@stardust-ui/react'
+
+const inputItems = ['Bruce Wayne', 'Natasha Romanoff', 'Steven Strange', 'Alfred Pennyworth']
+
+const DropdownArrowExample = props => {
+  const { position, align } = props
```
layershifter: Just wonder, will knobs more usable for this example?

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -236,6 +245,7 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
       content: false,
     }),
     activeSelectedIndex: PropTypes.number,
+    align: PropTypes.oneOf(_.without(ALIGNMENTS)),
```
```suggestion
    align: PropTypes.oneOf(ALIGNMENTS),
```

Cruft?

packages/react/src/lib/positioner/UpdatableComponent.tsx
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * Array of conditions to be met in order to trigger a subsequent render.
+   */
+  updateDependencies: any[]
+}
+
+const UpdatableComponent: React.FunctionComponent<Extendable<UpdatableListProps>> = props => {
+  const { Component, innerRef, scheduleUpdate, updateDependencies, ...rest } = props
+
+  React.useEffect(() => scheduleUpdate && scheduleUpdate(), updateDependencies)
```
layershifter:
```suggestion
  React.useEffect(() => scheduleUpdate(), updateDependencies)
```

It defined as required in `UpdatableListProps`...

packages/react/src/lib/positioner/createPopperReferenceProxy.ts
```diff
@@ -4,11 +4,7 @@ import * as React from 'react'
 import * as PopperJS from 'popper.js'
 
 class ReferenceProxy implements PopperJS.ReferenceObject {
-  ref: React.RefObject<HTMLElement>
-
-  constructor(refObject) {
-    this.ref = refObject
-  }
+  constructor(private ref: React.RefObject<HTMLElement>) {}
```
layershifter: Nit, but I am not sure about this change as I see it's TS specific stuff

[Babel REPL](https://babeljs.io/repl/#?babili=false&browsers=last%208%20versions%2C%20safari%20%3E%208%2C%20firefox%20%3E%2023%2C%20chrome%20%3E%2024%2C%20opera%20%3E%2015%2C%20not%20ie%20%3C%2011%2C%20not%20ie_mob%20%3C%3D%2011&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAECC0G8BQwD2A7CAXATgV2C1RwAocBTAMwEpEBfJBlDbaMaAXmgHIxulmmVCHIA6EKgDmJMKLDUgA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Ctypescript&prettier=true&targets=Node-8.9&version=7.4.4&externalPlugins=)

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -559,20 +574,47 @@ class Dropdown extends AutoControlledComponent<Extendable<DropdownProps>, Dropdo
...
+
+    return (
+      <Positioner
+        align={align}
+        position={position}
+        offset={offset}
+        rtl={rtl}
+        target={this.selectedItemsRef}
+        children={popperChildrenProps => (
+          <UpdatableComponent
```
layershifter: I just wondered, why `UpdatableComponent` is not part of `Positioner`? For me it has much more sence

packages/react/src/lib/positioner/createPopperReferenceProxy.ts
bmdalex: well, aren't we writing TS code? I don't understand the Babel reference it contains JS code..

packages/react-component-ref/src/Ref.tsx
bmdalex: oops, reverted, but the doc comment above is misleading 😕

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: oops, reverted, thanks

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: I was thinking the same, but wanted to allow `Positioner` to receive components that don't necessarily update based on condition.

Alternative would be to:
- make `UpdatableComponent` HOC so that we don't need to pass `Component` and rest of props as props of `UpdatableComponent`
- move  `UpdatableComponent` to `Positioner` and add `updateDependencies` prop to `Positioner` as well and make it optional; then, if `updateDependencies` is:
  - `falsy`: we don't use `UpdatableComponent` at all
  - `[]`: component is rendered only when mounted (hooks behavior)
  - `[condition1, condition2, ..., conditionN]`: when at least a `conditionX` is true component rerenders (hooks behavior)

I think it makes sense. What do you think, @layershifter ?

packages/react/src/lib/positioner/UpdatableComponent.tsx
bmdalex: good point, applied suggestion, thanks

docs/src/examples/components/Dropdown/Variations/DropdownExamplePosition.shorthand.tsx
bmdalex: hmm.. we can but for Popup we don't do that for very similar example. I can make the change for both for consistency and use knobs , but which one do you recommend?

@layershifter

packages/react/src/lib/positioner/createPopperReferenceProxy.ts
layershifter: We're going to use Babel as compiler, if we can't compile this code with Babel - it's a 💣 I didn't check it with local compiler, but was not able to make it working in REPL

packages/react/src/lib/positioner/Positioner.tsx
```diff
@@ -0,0 +1,64 @@
...
+   * and 'start' | 'end' respectively), then provided value for 'align' will be ignored and 'center' will be used instead.
+   */
+  position?: Position
+}
+
+interface PositionerProps extends PopperProps, PositionCommonProps {
+  /**
+   * Content for children using render props API
+   */
+  children: (props: PopperChildrenProps) => React.ReactNode
```
layershifter: It's defined in `PopperProps`, do we need it?

packages/react/src/lib/positioner/Positioner.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * rtl attribute for the component
+   */
+  rtl?: boolean
+
+  target?: HTMLElement | React.RefObject<HTMLElement>
```
layershifter: Missing description?

packages/react/src/components/Dropdown/Dropdown.tsx
layershifter: I suggest to remove `UpdatableComponent` at because it doesn't contain anything meaningful and place update logic in `Positioner`:

```tsx
  const scheduleUpdate = React.useRef<PopperChildrenProps['scheduleUpdate']>(null)
  
  React.useEffect(() => {
    if (scheduleUpdate.current) scheduleUpdate.current()
  }, updateDependencies) // `updateDependencies` is [] by default

  return (
    <Popper
      children={(childrenProps) => {
        scheduleUpdate.current = childrenProps.scheduleUpdate
        
        return children(childrenProps)
      }}
   />
```

layershifter: And may be `positioningDependencies` will be more accurate, not sure

docs/src/examples/components/Dropdown/Variations/DropdownExamplePosition.shorthand.tsx
layershifter: I think that we can both, it will make them more understandable and compact I think

packages/react/src/lib/positioner/createPopperReferenceProxy.ts
bmdalex: even if that's the case the solution is not to avoid TS features but to:
- either make compilation work with whatever compiler we use (e.g.: for babel there are plugins such as: https://babeljs.io/docs/en/babel-plugin-transform-typescript)
- or mark somehow (tslint config, lint rules, etc) what TS features we cannot use in the project (if there is strong reason).

I already see this TS feature being used in:
https://github.com/stardust-ui/react/blob/5e76653d940cd037ef9fc17931da9f4e2ac8327a/packages/react/src/lib/accessibility/FocusHandling/FocusContainer.ts#L4 

### tl;dr
I think preventing devs from using language features because our tooling might not work is wrong approach (unless that language feature has perf issues, increases bundle size or introduces problems).

docs/src/examples/components/Dropdown/Variations/DropdownExamplePosition.shorthand.tsx
bmdalex: that took longer than expected; reimplemented Dropdown and Popup examples related to position using select knobs

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: makes sense, the only thing missing was assigning the ref, added that following your suggestions, thanks!

packages/react/src/lib/positioner/Positioner.tsx
bmdalex: cruft, removed, thanks

packages/react/src/lib/positioner/Positioner.tsx
bmdalex: yup, added, thanks

packages/docs-components/src/knobs/useSelectKnob.ts
```diff
@@ -1,9 +1,9 @@
 import { UseKnobOptions } from './types'
 import useKnob from './useKnob'
 
-const useSelectKnob = (options: UseKnobOptions<string>) =>
-  useKnob<string>({
-    initialValue: '',
+const useSelectKnob = <T = string>(options: UseKnobOptions<T>) =>
+  useKnob<T>({
+    initialValue: undefined,
```
bmdalex: @layershifter this should be safe, right? is this still needed or can I omit?

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
```diff
@@ -1,54 +1,91 @@
...
+
+  React.useEffect(() => setPositionBeforeOrAfter(isPositionBeforeOrAfter(position)), [position])
+
+  const buttonStyles = { padding: paddings[position][align], height: '38px', minWidth: '64px' }
+
+  return (
+    <Grid columns="1" variables={{ padding: '100px 0' }} styles={{ justifyItems: 'center' }}>
+      <Popup
+        pointing
+        open={open}
```
miroslavstastny: This means that if the checkbox is not checked, it is not possible to open the popup by clicking the trigger.
I would use something like `open={open || undefined}`

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
```diff
@@ -1,54 +1,91 @@
...
+          values: ['start', 'center', 'end'],
+        },
+  )
+
+  React.useEffect(() => setPositionBeforeOrAfter(isPositionBeforeOrAfter(position)), [position])
+
+  const buttonStyles = { padding: paddings[position][align], height: '38px', minWidth: '64px' }
+
+  return (
+    <Grid columns="1" variables={{ padding: '100px 0' }} styles={{ justifyItems: 'center' }}>
```
miroslavstastny, line 19: Is `Grid` the best component for 0D layout?

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
```diff
@@ -1,54 +1,91 @@
...
-          <p>
-            The popup is rendered {position} the trigger
-            <br />
-            aligned to the {align}.
-          </p>
-        ),
-      }}
-    />
+  const [position] = useSelectKnob<Position>({
+    name: 'position shorthand',
```
miroslavstastny: What does the word 'shorthand' mean in knob titles?

miroslavstastny (commented):
The dropdown is jumping during page scroll:

dropdown_scroll

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
bmdalex: I used it just to have a different name from the one in `PopupExamplePosition.tsx`. For knobs the name acts as an id so if I use the same name I get an error:

`Knob with name "${knob.name}" has been already registered`

Open to new name ideas that are more relevant

@miroslavstastny @layershifter

layershifter: Will check it...

packages/docs-components/src/knobs/useSelectKnob.ts
layershifter: As I know it was changed in #1290, so you can just pick this change on merge

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
layershifter: We can also use `onOpenChange`, each knob contains a setter in its tuple: `const [open, setOpen] = useBooleanKnob()`

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
layershifter: `packages/docs-components/src/knobs/KnobProvider.tsx` should be updated with this:

```jsx
  const registerKnob = (knob: KnobDefinition) => {
    setKnobs(prevKnobs => {
        if (prevKnobs[knob.name]) {
          throw new Error(`Knob with name "${knob.name}" has been already registered`)
        }
      }

      return { ...prevKnobs, [knob.name]: knob }
    })
  }
```

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
bmdalex: I don't know :( I used it because it was used in https://github.com/stardust-ui/react/blob/5c2688a9715b3b5e3925cc8da59420eada40a5c7/docs/src/examples/components/Popup/Variations/PopupExampleOffset.shorthand.tsx#L15

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
bmdalex: It'd be a nice thing to add. It worked for the `Popup` but didn't for the `Dropdown`. When I deselect the `open` checkbox, Downshift starts updating the current state of `open` flag before the `open` prop is changed from `true` to `false`.

Order of events is strange.. I debugged this with @layershifter all day and didn't reach a conclusion :(

Any suggestions @miroslavstastny ? I'd keep this as is until we have issues impacting users.

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
bmdalex: @layershifter sent me a fix for this so I was able to use just the same name, fixed

bmdalex:
> The dropdown is jumping during page scroll:

@miroslavstastny I am aware of that. I was debugging popper code and noticed they recreate the popper instance (responsible for calculating styles that need to be applied for positioning elements) almost (they use debounce) every time there is a scroll/resize event: https://github.com/FezVrasta/popper.js/blob/36cc6c66e55f716232166c9a5725d8d08c4ec338/packages/popper/src/utils/setupEventListeners.js#L38 . This is where the jump comes from.

I think this can cause perf issues especially when trying to:
- position complex elements (such as the list of results in `Dropdown`);
- position multiple elements in the same view.

The jump is quite visible in our docs because there is a popper instance for **every Dropdown example** even if they are closed. This is due to the fact that for `Dropdown` we render the list of results even if it's empty (accessibility reasons, correct me if I'm wrong @silviuavram @jurokapsiar ).

### MITIGATION:

I think I was able to mitigate this by adding **`eventsEnabled={open}`** prop to `Popper`, which means the scroll/resize events are attached only when the list of results is open for a `Dropdown`. **@miroslavstastny can you please test again?**

### ALTERNATIVE:

I'm starting to think the simpler solution for `Dropdown` would be just to use `position: 'absolute'` and `top/bottom` CSS properties to position the the list of results above or below the trigger.

#### PROS:
- no jump;
- no perf issues because we won't recalculate anything on subsequent renders/scroll events/resize events;
- no need to integrate with popper at all.

#### CONS:
- we'd be limiting the `Dropdown` positioning to `above` or `below` the trigger; @jurokapsiar @silviuavram  @layershifter would that be enough according to our specs? I didn't really see too many libraries aiming for more than this;
- no auto positioning: we will need a function to tell us if the list of results is in the viewport to achieve that.

miroslavstastny:
> * no auto positioning: we will need a function to tell us if the list of results is in the viewport to achieve that.

https://react.semantic-ui.com/behaviors/visibility/

packages/docs-components/src/knobs/KnobProvider.tsx
```diff
@@ -12,13 +12,14 @@ const KnobProvider: React.FunctionComponent<KnobProviderProps> = props => {
   const [knobs, setKnobs] = React.useState<KnobSet>({})
 
   const registerKnob = (knob: KnobDefinition) => {
-    if (process.env.NODE_ENV !== 'production') {
-      if (knobs[knob.name]) {
-        throw new Error(`Knob with name "${knob.name}" has been already registered`)
+    setKnobs(prevKnobs => {
```
bmdalex: fix for the problem with using a different names for the same knob, thx @layershifter

packages/react/src/lib/positioner/createPopperReferenceProxy.ts
bmdalex: NA anymore, code was removed

bmdalex:
> The dropdown is jumping during page scroll:
> 
> dropdown_scroll

this perf issue was fixed by #1358

packages/docs-components/src/knobs/KnobProvider.tsx
miroslavstastny: This breaks existing docsite pages with knobs (`Button`, `Menu`):
image

bmdalex: reverted so that I get unblocked, thanks, @layershifter let's tackle this fix in separate PR as the one proposed introduces the issue @miroslavstastny is mentioning

docs/src/examples/components/Dropdown/Variations/DropdownExamplePosition.shorthand.tsx
```diff
@@ -0,0 +1,49 @@
...
+        align={align}
+        position={position}
+      />
+    </Grid>
+  )
+}
+
+export default DropdownExamplePosition
+
+type PositionAndAlign =
```
layershifter: Just wonder, do we really need this type?

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
```diff
@@ -1,54 +1,90 @@
...
-        position={position}
-        align={align}
-        icon={icon}
-        padding={padding}
-        key={`${position}-${align}`}
-      />
-    ))}
-  </Grid>
-)
+const icons: { [key in Position]: string } = {
```
layershifter:
```suggestion
const icons: Record<Position, string> = {
```

 _nit_

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1273,6 +1296,15 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
   }
+
+  private getToggleIndicatorIconName = (): string => {
+    const { open } = this.state
+    const positionedAbove = this.props.position === 'above'
+
+    return (open && !positionedAbove) || (!open && positionedAbove)
+      ? 'stardust-arrow-up'
+      : 'stardust-arrow-down'
+  }
```
layershifter: This behavior is very strange, native `select` doesn't do the same thing:

image

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -49,6 +49,7 @@ const Popper: React.FunctionComponent<PopperProps> = props => {
       const popperHasScrollableParent = getScrollParent(contentRef.current) !== document.body
 
       const modifiers: PopperJS.Modifiers = _.merge(
+        { preventOverflow: { padding: 0 } },
```
layershifter, line 52: Should be this in modifiers in `Dropdown` component?

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -144,17 +147,14 @@ const dropdownStyles: ComponentSlotStylesInput<DropdownPropsAndState, DropdownVa
 
   list: ({ props: p, variables: v }): ICSSInJSStyle => ({
     outline: 0,
-    position: 'absolute',
-    borderRadius: v.listBorderRadius,
+    borderRadius: getOpenBorderStyles(p.position === 'below' ? 'top' : 'bottom', v),
```
layershifter: Can we omit this function and write it more flat, but more obvious?

```js
...p.open && p.position === 'top' && ({ borderBottomRaduis: v.borderBottomRadius }),
...p.open && p.position === 'bottom' && ({ borderBottomRaduis: v.borderBottomRadius }),
// etc
```

It's much more readable, see ##1301.

docs/src/examples/components/Dropdown/Variations/DropdownExamplePosition.shorthand.tsx
bmdalex: Don't needed, just for safety. Placed it at the end so that it doesn't distract user. Can be removed if you feel strongly about it @layershifter

layershifter: As it was mentioned before, example should extra simple and contain less code. So if we don't have reasons to have this type, let's remove it.

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: removed logic for now, will create issue

packages/react/src/lib/positioner/Popper.tsx
bmdalex: I think it's something we want for all our components

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: One of the aria examples is broken

packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
bmdalex: done, thanks

docs/src/examples/components/Dropdown/Variations/DropdownExamplePosition.shorthand.tsx
bmdalex: done, thanks

docs/src/examples/components/Popup/Variations/PopupExamplePosition.tsx
```diff
@@ -1,57 +1,77 @@
...
-        position={position}
-        align={align}
-        icon={icon}
-        padding={padding}
-        key={`${position}-${align}`}
-      />
-    ))}
-  </Grid>
-)
+const icons: { [key in Position]: string } = {
```
layershifter:
```suggestion
const icons: Record<Position, string> = {
```
_nit_

bmdalex: done, thanks

layershifter (approved):

miroslavstastny (approved):

PR #1315, opened 2019-05-10
fix(docs): fixed styles on the editor menus

mnajdova:
Fixed regressions on the menus used in the editor, after merging the color updates.

Prev:
image

Now:
image


layershifter (approved):

layershifter:
This regression came from #1069, so we don't need to add anything to changelog. Thanks for fixing 👍

PR #1318, opened 2019-05-10
docs(sources): use ESLint to format generated JS output

layershifter:
#### Uses ESLint to format generated JS output

It allows to apply ESLint's rules with automatic fixes to get better better autogenerated output.

##### Before

image

##### After

image

#### Enables `import/newline-after-import` rule


miroslavstastny (approved):
I am confused by the screenshots in the description - this PR enables `newline-after-import` but after has no new line.

layershifter:
Correct the order of screenshots 👍

PR #1019, opened 2019-03-06
docs(layouts): more layout fixes

layershifter:
Continues work from #1014.

PR #1319, opened 2019-05-10
fix(docs): Set maximum width for examples

### LINKED ISSUE #1317: Docs: wide example breaks docsite layout
### Steps
Have a wide example

### Expected Result
Example pane does not grow out of browser width.

### Actual Result
image

### Version
master

### Testcase

miroslavstastny:
Fixes #1317.

Before:
image

After:
image

The visible overflow is intentional.

### DIFF
--- a/docs/src/examples/components/Text/States/TextExampleTruncated.shorthand.tsx
+++ b/docs/src/examples/components/Text/States/TextExampleTruncated.shorthand.tsx
@@ -11,7 +11,7 @@ const [notTruncatedText, truncatedText] = [
 )
 
 const TextExampleTruncatedShorthand = () => (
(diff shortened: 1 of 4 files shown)


docs/src/examples/components/Text/States/TextExampleTruncated.shorthand.tsx
```diff
@@ -11,7 +11,7 @@ const [notTruncatedText, truncatedText] = [
 )
 
 const TextExampleTruncatedShorthand = () => (
-  <div style={{ overflow: 'hidden', textOverflow: 'ellipsis', width: 500 }}>
+  <div style={{ overflow: 'hidden', textOverflow: 'ellipsis' }}>
```
miroslavstastny, line 14: Reverted suboptimal fix from #1019

layershifter (approved):

PR #1320, opened 2019-05-10
fix(FlexItem): fix children type

mnajdova:
FIxing this console warning:

```
react.development.js:2726 Warning: Failed prop type: Invalid prop `children` supplied to `FlexItem`, expected a ReactNode.
    in FlexItem (created by ComponentExampleKnobs)
    in ComponentExampleKnobs (created by KnobInspector)
    in KnobInspector (created by ComponentExample)
    in div (created by Segment)
    in Segment (created by ComponentExample)
    in KnobProvider (created by ComponentExample)
    in FlexItem (created by ComponentExample)
    in div (created by Flex)
    in Flex (created by ComponentExample)
    in ComponentExample (created by ComponentSourceManager)
   ...
```


packages/react/src/components/Flex/FlexItem.tsx
```diff
@@ -49,9 +50,12 @@ class FlexItem extends UIComponent<ReactProps<FlexItemProps>> {
 
   static propTypes = {
     ...commonPropTypes.createCommon({
+      children: false,
```
miroslavstastny, line 52: nit: `children` argument in createCommon can be boolean or string. Would it make sense to pass direct value (like you do on line 57) as well (or instead of the string)?

miroslavstastny (approved):

packages/react/src/components/Flex/FlexItem.tsx
mnajdova: That's an interesting question. I can add as a separate PR options for passing direct value for the children and content options.

PR #1322, opened 2019-05-10
feat(Accordion): add a11y behavior for kb navigation and screen reader support

silviuaavram:
Work that aims to implement the Accordion design pattern from aria. https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/#accordion and example https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/accordion/accordion.html

Changes:
- title has new structure. It has a focusable `Box` with a `button` role between root and `Layout`.
- changed element types to a description list from HTML5. Used `dl`, `dh` and `dd` for accordion, title and content.
- added behaviors for accordion, title and content.
- added default IDs for title and content. Did it from Accordion, could not figure out a better way through behaviors.
- implemented `expanded` Accordion type. With it truthy, it will always keep a panel open.
- added keyboard actions support, on Arrow up/down, home/end and Enter/Space. Navigation is circular, so updated the `ContainerFocusHandler` to support this.
-added unit tests for Accordion, Title, Content and behaviors.

Work in progress:
- Ctrl + PageUp/PageDown from Content to jump focus to previous/next Title. - **Chrome uses this for tab navigation. I don't think we should override it, and this shortcut is only useful for big panels with many focusable elements in the Content.**


packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -115,22 +122,66 @@ class Accordion extends AutoControlledComponent<WithAsProp<AccordionProps>, any>
...
+
+          targetDomNode && targetDomNode.focus()
+        })
+      },
+      true,
+    )
+  }
+
+  getInitialAutoControlledState({ alwaysActive, exclusive }: AccordionProps) {
+    return { activeIndex: exclusive ? (alwaysActive ? 0 : -1) : alwaysActive ? [0] : [-1] }
```
sophieH29: nit: I would rewrite it for a better readability
```
const alwaysActiveIndex = alwaysActive ? 0 : -1
return { activeIndex: exclusive ? alwaysActiveIndex : [alwaysActiveIndex] }
```

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -23,6 +25,7 @@ import {
   ShorthandRenderFunction,
   withSafeTypeForAs,
 } from '../../types'
+import { ContainerFocusHandler } from 'src/lib/accessibility/FocusHandling/FocusContainer'
```
mnajdova: Convert this to relative import: `'../../lib/accessibility/FocusHandling/FocusContainer'`

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -146,33 +197,73 @@ class Accordion extends AutoControlledComponent<WithAsProp<AccordionProps>, any>
...
 
-  isIndexActive = (index): boolean => {
+  private isIndexActive = (index: number): boolean => {
     const { exclusive } = this.props
     const { activeIndex } = this.state
 
     return exclusive ? activeIndex === index : _.includes(activeIndex, index)
   }
 
+  private isIndexClosingPrevented = (index: number): boolean => {
```
sophieH29: would be good to add comments about what this method does, not very self-explanatory method name

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -33,6 +36,9 @@ export interface AccordionProps extends UIComponentProps, ChildrenComponentProps
   /** Index of the currently active panel. */
   activeIndex?: number[] | number
 
+  /** At least one panel should be active at any time. */
+  alwaysActive?: boolean
```
mnajdova: `expanded` as a name? I really don't like that we are using active, but not your fault (even in the description of the active prop we have that it means open...)

packages/react/src/components/Accordion/Accordion.tsx
silviuaavram: I had a hard time to think about a proper name for it. I will probably name it `canBeCollapsed` as per Marija's suggestion below :D. Will also add description thanks!

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -115,22 +122,66 @@ class Accordion extends AutoControlledComponent<WithAsProp<AccordionProps>, any>
...
+    },
+  }
+
+  constructor(props, context) {
+    super(props, context)
+
+    this.focusHandler = new ContainerFocusHandler(
+      () => this.props.panels.length,
+      index => {
+        this.setState({ focusedIndex: index }, () => {
```
mnajdova: Can we extract this to a method? At first sight it looks like we are invoking setState in the constructor, but this is just a param to the constructor of the ContainerFocusHandler

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -146,33 +197,73 @@ class Accordion extends AutoControlledComponent<WithAsProp<AccordionProps>, any>
...
 
+  private isIndexClosingPrevented = (index: number): boolean => {
+    const { activeIndex } = this.state
+    const { alwaysActive, exclusive } = this.props
+
+    if (alwaysActive) {
+      if (exclusive) {
+        return activeIndex === index
+      }
+      return activeIndex.length === 1 && activeIndex[0] === index
```
mnajdova: What should happen if the `alwaysActive` is used with `exclusive = false`, and the index is not the first element?

packages/react/src/components/Accordion/Accordion.tsx
sophieH29: I like `canBeCollapsed` more 😄

packages/react/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -24,9 +29,18 @@ export interface AccordionTitleProps
   /** Whether or not the title is in the open state. */
   active?: boolean
 
+  /** If at least one panel needs to stay active and this title does not correspond to the last active one. */
+  cannotBeClosed?: boolean
```
sophieH29: to be consistent with expand/collapse terminology, maybe just `cannotBeCollapsed`

packages/react/src/lib/accessibility/Behaviors/Accordion/accordionTitleBehavior.ts
```diff
@@ -0,0 +1,31 @@
...
+  return {
+    attributes: {
+      root: {
+        role: isHeading ? undefined : 'heading',
+        'aria-level': isHeading ? undefined : 3,
+      },
+      button: {
+        'aria-expanded': !!props.active,
+        'aria-disabled': !!(props.active && props.cannotBeClosed),
+        'aria-owns': props.contentId,
```
sophieH29: I see from spec and example they suggest using `aria-controls`, not `aria-owns` or I might missing something
`The accordion header button element has aria-controls set to the ID of the element containing the accordion panel content.`

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -146,33 +197,73 @@ class Accordion extends AutoControlledComponent<WithAsProp<AccordionProps>, any>
...
+    this.focusHandler.syncFocusedIndex(focusedIndex)
 
     _.each(panels, (panel, index) => {
       const { content, title } = panel
       const active = this.isIndexActive(index)
+      const cannotBeClosed = this.isIndexClosingPrevented(index)
+      const buttonRef = React.createRef<HTMLElement>()
+      const titleId = title['id'] || _.uniqueId('accordion-title-')
+      const contentId = content['id'] || _.uniqueId('accordion-content-')
+      this.itemRefs[index] = buttonRef
```
sophieH29: do you need a `buttonRef` prop in `AccordionTitle` component?
I mean, to do it in the same way how it's done in List.

```jsx
const ref = React.createRef()
this.itemRefs[index] = ref
...
const accordionTitleProps = {
   ...
   ref,
   ...
}

AccordionTitle.create(title, {
    defaultProps: accordionTitleProps,
    ...
})
```

packages/react/src/components/Accordion/Accordion.tsx
silviuaavram: actually my logic here is stupid. I will refactor here and hopefully it will not be stupid. again.

packages/react/src/lib/accessibility/Behaviors/Accordion/accordionTitleBehavior.ts
silviuaavram: no, you are right.

packages/react/src/components/Accordion/Accordion.tsx
silviuaavram: isn't that going to give me the `ref` to the `<ElementType>` ? I need the one from `<Button>` which is the child of `<ElementType>`

I am applying navigation on the buttons, not on their parents.

packages/react/src/components/Accordion/AccordionTitle.tsx
silviuaavram: added `canBeCollapsed` and inverted the logic

packages/react/src/components/Accordion/Accordion.tsx
silviuaavram: I just copy-pasted it from the List. Will extract it.

packages/react/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -49,37 +70,67 @@ class AccordionTitle extends UIComponent<WithAsProp<AccordionTitleProps>, any> {
...
-        main={rtlTextContainer.createFor({ element: content })}
-      />
+      <Ref innerRef={buttonRef}>
+        <Button
+          onFocus={this.handleFocus}
+          {...accessibility.attributes.button}
+          {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.button, unhandledProps)}
+        >
+          <Layout
+            start={Icon.create(indicatorWithDefaults, {
```
mnajdova: Why aren't we simple using the icon and content slots on the `Button`?

silviuaavram: I don't know, that's how it was implemented before. Now I am usign Box instead of Button. Should I refactor it better?

jurokapsiar: I think we would need a plain text variant of a button to make this look neutral. Both button and textOnly button have centered caption for example. I think a div is better

packages/react/test/specs/components/Accordion/AccordionContent-test.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { isConformant, handlesAccessibility, getRenderedAttribute } from 'test/specs/commonTests'
+import { mountWithProviderAndGetComponent } from 'test/utils'
+
+describe('AccordionContent', () => {
+  isConformant(AccordionContent)
+
+  describe('accessiblity', () => {
+    handlesAccessibility(AccordionContent)
+
+    describe('aria-labelledby', () => {
```
mnajdova, line 13: This should be tested by the behavior tests, not the component's test.

packages/react/test/specs/components/Accordion/AccordionTitle-test.tsx
```diff
@@ -0,0 +1,143 @@
...
+    describe('div header', () => {
+      handlesAccessibility(AccordionTitle, {
+        requiredProps: {
+          buttonRef: _.noop,
+        },
+        defaultRootRole: 'heading',
+      })
+    })
+
+    describe('aria-disabled', () => {
```
mnajdova: Again, why is this tests in the component's test?

packages/react/test/specs/components/Accordion/AccordionTitle-test.tsx
```diff
@@ -0,0 +1,143 @@
...
+import * as _ from 'lodash'
+
+import AccordionTitle from 'src/components/Accordion/AccordionTitle'
+import { isConformant, handlesAccessibility, getRenderedAttribute } from 'test/specs/commonTests'
+import { mountWithProviderAndGetComponent } from 'test/utils'
+
+describe('AccordionTitle', () => {
+  isConformant(AccordionTitle, {
+    eventTargets: {
+      onClick: `.${AccordionTitle.slotClassNames.button}`,
```
mnajdova: Why is `onClick` a string?

packages/react/test/specs/components/Accordion/AccordionTitle-test.tsx
```diff
@@ -0,0 +1,143 @@
...
+          AccordionTitle,
+          <AccordionTitle buttonRef={_.noop} />,
+        )
+        expect(
+          getRenderedAttribute(
+            renderedComponent,
+            'aria-expanded',
+            `.${AccordionTitle.slotClassNames.button}`,
+          ),
+        ).toBe('false')
```
mnajdova: Test description is wrong here. Again, let's move all behavior tests from the component's test.

packages/react/test/specs/components/Accordion/Accordion-test.tsx
```diff
@@ -0,0 +1,227 @@
...
+        },
+      ]
+      const wrapper = mountWithProvider(<Accordion panels={panels} onTitleClick={onTitleClick} />)
+      getTitleButtonAtIndex(wrapper, 0).simulate('click')
+
+      expect(onTitleClick).toBeCalledTimes(1)
+      expect(onClick).toBeCalledTimes(1)
+    })
+
+    it('when focused call onFocus if provided by the user', () => {
```
mnajdova: Isn't this tested already by the conformance's handlers events transparently? Let's just check, before removing this test.

packages/react/test/specs/components/Accordion/Accordion-test.tsx
```diff
@@ -0,0 +1,227 @@
...
+      const wrapper = mountWithProvider(<Accordion panels={panels} />)
+      const title = getTitleButtonAtIndex(wrapper, 1)
+      title.simulate('click')
+      title.simulate('keydown', { keyCode: keyboardKey.ArrowUp, key: 'ArrowUp' })
+      expect(document.activeElement).toEqual(getTitleButtonAtIndex(wrapper, 0).getDOMNode())
+    })
+  })
+
+  describe('panels', () => {
+    it('when clicked call onClick and onTitleClick if provided by the user', () => {
```
mnajdova, line 192: onClick should be tested by the isConformant - handles events transparently. The test for the onTitleClick is okay (although not sure why we have dedicated prop for that, but as it was not added by this changes, would open the discussion separately).

packages/react/test/specs/components/Accordion/Accordion-test.tsx
```diff
@@ -0,0 +1,227 @@
...
+    it('contains the indexes clicked by the user if the panels were closed', () => {
+      const wrapper = mountWithProvider(<Accordion panels={panels} />)
+      const accordion = wrapper.find(Accordion)
+      getTitleButtonAtIndex(wrapper, 0).simulate('click')
+      getTitleButtonAtIndex(wrapper, 2).simulate('click')
+
+      expect(accordion.state('activeIndex')).toEqual(expect.arrayContaining([0, 2]))
+    })
+
+    it('contains the only one index clicked by the user if exclusive prop is passed', () => {
```
mnajdova, line 86:
```
      const wrapper = mountWithProvider(<Accordion panels={panels} exclusive />)
      const accordion = wrapper.find(Accordion)
      getTitleButtonAtIndex(wrapper, 0).simulate('click')
      expect(accordion.state('activeIndex')).toEqual(0) // let's add this line as well

      getTitleButtonAtIndex(wrapper, 2).simulate('click')
       expect(accordion.state('activeIndex')).toEqual(2)
```

packages/react/test/specs/components/Accordion/Accordion-test.tsx
```diff
@@ -0,0 +1,227 @@
...
+  return wrapper
+    .find(`.${AccordionTitle.slotClassNames.button}`)
+    .filterWhere(n => typeof n.type() === 'string')
+    .at(index)
+}
+
+describe('Accordion', () => {
+  isConformant(Accordion)
+
+  describe('activeIndex', () => {
```
mnajdova, line 38: Very good tests, all expected behavior is nicely describe trough the tests. Very good job! 🥇

packages/react/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -46,40 +71,73 @@ class AccordionTitle extends UIComponent<WithAsProp<AccordionTitleProps>, any> {
...
-            styles: styles.indicator,
-          },
-        })}
-        main={rtlTextContainer.createFor({ element: content })}
-      />
+      <Ref innerRef={buttonRef}>
+        <Box
+          onFocus={this.handleFocus}
+          onClick={this.handleClick}
+          className={AccordionTitle.slotClassNames.button}
```
mnajdova: This is not button anymore, let's change the name of the slotClassNames

mnajdova: One more question, do we really need the box around the Layout? The Layout itself is a div, can't we just use the Layout component here?

packages/react/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -46,40 +71,73 @@ class AccordionTitle extends UIComponent<WithAsProp<AccordionTitleProps>, any> {
...
-            name: active ? 'stardust-arrow-down' : 'stardust-arrow-end',
-            styles: styles.indicator,
-          },
-        })}
-        main={rtlTextContainer.createFor({ element: content })}
-      />
+      <Ref innerRef={buttonRef}>
+        <Box
+          onFocus={this.handleFocus}
+          onClick={this.handleClick}
```
mnajdova, line 115: I am a bit confused with the handleClick and performClick concepts. Can you explain why we use here the handleClick instead of the performClick?

packages/react/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -46,40 +71,73 @@ class AccordionTitle extends UIComponent<WithAsProp<AccordionTitleProps>, any> {
 
   static className = 'ui-accordion__title'
 
+  static slotClassNames: AccordionTitleSlotClassNames
+
   static propTypes = {
     ...commonPropTypes.createCommon(),
     active: PropTypes.bool,
-    index: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+    buttonRef: customPropTypes.ref,
```
mnajdova: let's rename this to contentRef, as is no longer a Button

packages/react/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -12,10 +13,18 @@ import {
...
 } from '../../lib'
 import { WithAsProp, ComponentEventHandler, ShorthandValue, withSafeTypeForAs } from '../../types'
 import Icon from '../Icon/Icon'
 import Layout from '../Layout/Layout'
+import Box from '../Box/Box'
+import { accordionTitleBehavior } from '../../lib/accessibility'
+import { AccessibilityActionHandlers } from 'src/lib/accessibility/types'
+
+export interface AccordionTitleSlotClassNames {
+  button: string
```
mnajdova: Rename this to content.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -115,64 +127,162 @@ class Accordion extends AutoControlledComponent<WithAsProp<AccordionProps>, any>
...
+    const { focusedIndex } = this.state
+
+    this.itemRefs = []
+    this.focusHandler.syncFocusedIndex(focusedIndex)
 
     _.each(panels, (panel, index) => {
       const { content, title } = panel
       const active = this.isIndexActive(index)
+      const canBeCollapsed = this.isIndexActionable(index)
+      const buttonRef = React.createRef<HTMLElement>()
```
mnajdova: Be careful when renaming this in the other component.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -115,64 +127,162 @@ class Accordion extends AutoControlledComponent<WithAsProp<AccordionProps>, any>
...
+      true,
+    )
+  }
+
+  private handleNavigationFocus = (index: number) => {
+    this.setState({ focusedIndex: index }, () => {
+      const targetComponent = this.itemRefs[index] && this.itemRefs[index].current
+      const targetDomNode = ReactDOM.findDOMNode(targetComponent) as any
+
+      targetDomNode && targetDomNode.focus()
```
mnajdova: No need for using ReactDOM here. You can directly use the targerComponent.focus(). Is there a case where this doesn't work? I've tested this and it seems to work, but maybe I am missing something...

```
 private handleNavigationFocus = (index: number) => {
    this.setState({ focusedIndex: index }, () => {

      targetComponent && targetComponent.focus()
    })
  }
```

mnajdova:
We can use `marginInlineStart: 0`, in the accordionContentStyles.ts root elements, in order to avoid the browser's added spacing before the accordion content.

CHANGELOG.md
```diff
@@ -28,6 +28,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fixed positioing of `Popup` in scrollable container @layershifter ([#1341](https://github.com/stardust-ui/react/pull/1341))
 
 ### Features
+- Add `accordionBehavior`, `accordionTitleBehavior` and `accordionContentBehavior` @silviuavram ([#1322](https://github.com/stardust-ui/react/pull/1322)) 
+- Add `expanded` prop to `Accordion` @silviuavram ([#1322](https://github.com/stardust-ui/react/pull/1322))
```
sophieH29: why not make it as one feature the same as a PR's name - Add keyboard navigation and screen reader support for Accordion? It might make more sense for the end users

packages/react/test/specs/components/Accordion/AccordionTitle-test.tsx
silviuaavram: it says that the handler will go to that slot and not on the component. otherwise the isConformant will fail and tell that onClick is not applied to ElementType.

mnajdova: Thanks for the clarification, missed that this is used as a config for the conformat test. Will resolve this one :)

packages/react/test/specs/components/Accordion/Accordion-test.tsx
silviuaavram: we agreed on the fix

packages/react/test/specs/components/Accordion/Accordion-test.tsx
silviuaavram: same here

packages/react/src/components/Accordion/AccordionTitle.tsx
silviuaavram: performClick is used throughout our components to be grabbed by the behavior, for components that need to have click functionality on Enter, Space etc.

I used the same performClick and handleClick for the sake of consistency

packages/react/src/components/Accordion/Accordion.tsx
silviuaavram: copy-pasted this from List. Will change.

CHANGELOG.md
silviuaavram: It's a good idea, will change. But shouldn't we emphasize that we added some props as well? As a separate entry apart from kb navigation and screen reader support?

packages/react/src/components/Accordion/AccordionTitle.tsx
silviuaavram: let's think about it more. I also added contentId to point to the AccordionContent. I think contentRef will create some confusion. What do you think?

packages/react/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -12,20 +13,37 @@ import {
...
   commonPropTypes,
   rtlTextContainer,
+  applyAccessibilityKeyHandlers,
 } from '../../lib'
 import { WithAsProp, ComponentEventHandler, ShorthandValue, withSafeTypeForAs } from '../../types'
 import Icon from '../Icon/Icon'
 import Layout from '../Layout/Layout'
+import Box from '../Box/Box'
+import { accordionTitleBehavior } from '../../lib/accessibility'
+import { AccessibilityActionHandlers } from 'src/lib/accessibility/types'
```
mnajdova: Ooops, change this to relative import.

mnajdova (approved):

PR #1327, opened 2019-05-13
fix(IconBehavior): changing icon behavior as for some cases icon could be visible

kolaps33:
During table examples implementation, we figure out icon could be visible inside the table cell. In order not narrate empty cell for user. 
And if icon should behave correctly it should have role image.


sophieH29 (approved):

mnajdova:
The changes looks good, but let's fist fix the failing test before final approval.

mnajdova (approved):

PR #1331, opened 2019-05-13
fix(Ref): handle refs on updates of `innerRef` prop

### LINKED ISSUE #1328: Ref: refs should be update on their change
## Bug Report

Newly passed objects or functions are not behave properly: on React elements it behave differently. The each passed not shallow equal object or a function should produce an assignment or a function call.

# Case 1

### Steps
1. Open https://codesandbox.io/s/xrvvzyjw3p
2. Open console, see that `nativeRef.current` and `ourRef.current` are defined
3. Click on `CLICK` button

### Expected Result
`nativeRef.current` and `ourRef.current` are defined

### Actual Result
`nativeRef.current` is defined, but `ourRef.current` is `null`

# Case 2

### Steps
1. Open https://codesandbox.io/s/nw063xxz8j
2. Open console, see that `nativeFnRef` and `ourFnRef` in output
3. Click on `CLICK` button

### Expected Result
 `nativeFnRef` and `ourFnRef ` in output

### Actual Result
only `nativeFnRef` in output

layershifter:
Fixes #1328.

Align existing behavior with React's, see the issue for more context and explanation.

### DIFF
--- a/packages/react-component-ref/src/RefFindNode.tsx
+++ b/packages/react-component-ref/src/RefFindNode.tsx
@@ -32,11 +32,15 @@ export default class RefFindNode extends React.Component<RefFindNodeProps> {
     handleRef(this.props.innerRef, this.prevNode)
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps: RefFindNodeProps) {
     const currentNode = ReactDOM.findDOMNode(this)
 
     if (this.prevNode !== currentNode) {
       this.prevNode = currentNode
+      handleRef(this.props.innerRef, currentNode)
+    }
+
+    if (prevProps.innerRef !== this.props.innerRef) {
       handleRef(this.props.innerRef, currentNode)
     }
   }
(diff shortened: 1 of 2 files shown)


packages/react-component-ref/src/RefFindNode.tsx
```diff
@@ -32,13 +32,17 @@ export default class RefFindNode extends React.Component<RefFindNodeProps> {
     handleRef(this.props.innerRef, this.prevNode)
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps: RefFindNodeProps) {
```
kuzhelov, line 35: think that it might be beneficial to implement this component with Hooks, for couple of reasons
- in that case we will have explicit update dependencies specified in `useEffect`
- explicitness on `unsubscribe` actions for effects
- this component is fairly small in terms of amount of code

kuzhelov (approved):

PR #1337, opened 2019-05-14
feat(themes): add FontAwesome theme

layershifter:
# BREAKING CHANGES

FontAwesome icons is not a part of Teams theme more consider to use SVG icons.

----

FontAwesome styles were extracted to a separate theme which can be merged with the base for example.

```jsx
import { mergeThemes, themes } from '@stardust-ui/react'

const myTheme = mergeThemes(themes.base, themes.fontAwesome)
```

----

Adds separate roots for font and SVG icons, so styles can be configured independently now without conditions in styles:

```
root: () => {} // will be applied to both types
fontRoot: () => {} // will be applied only for font icons
svgRoot: () => {} // will be applied only for SVG icons
```


docs/src/app.tsx
```diff
@@ -20,7 +20,7 @@ class App extends React.Component<any, ThemeContextData> {
       <ThemeContext.Provider value={this.state}>
         <Provider
           as={React.Fragment}
-          theme={mergeThemes(themes[themeName], {
+          theme={mergeThemes(themes.fontAwesome, themes[themeName], {
```
layershifter, line 23: We're using FA on docs, but we don't to have this theme in any other even in `base`

docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -540,6 +540,7 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
               render={this.renderElement}
               renderHtml={showCode}
               resolver={importResolver}
+              themeName={themeName}
```
layershifter, line 543: This will allow us to access `themeName` in example components

docs/src/examples/components/Avatar/Variations/AvatarExampleImageCustomization.shorthand.tsx
```diff
@@ -18,7 +18,7 @@ const AvatarExampleImageCustomizationShorthand = () => (
         // This example does not react to the avatar size variable
         // and otherwise produces bad results when border is applied compared to "normal" image
         <Icon
-          name="user"
+          name="lock"
```
layershifter, line 21: Teams theme doesn't have this icon and font icons will not be supported inside of it

packages/react/src/components/Icon/Icon.tsx
```diff
@@ -80,46 +79,28 @@ class Icon extends UIComponent<WithAsProp<IconProps>, any> {
...
+      {
+        defaultProps: {
+          as: ElementType,
+          className: cx(Icon.className, className),
+          ...accessibility.attributes.root,
+          ...unhandledProps,
+          styles: {
+            ...(isSvgIcon ? styles.svgRoot : styles.fontRoot),
+            ...styles.root,
+          },
```
layershifter, line 100: I have to use `styles` to avoid collisions in generated classes names... The other option is to use `generateCSS()` directly, but I will need to get `renderer` from somewhere

packages/react/src/themes/base/components/Icon/iconStyles.ts
```diff
@@ -61,27 +31,39 @@ const getPaddedStyle = (): ICSSInJSStyle => ({
...
+
+    ...getXSpacingStyles(p.xSpacing, v.horizontalSpace),
+
+    ...(p.bordered && getBorderedStyles(v.borderColor)),
+    ...(p.circular && { ...getPaddedStyle(), borderRadius: '50%' }),
+    ...(p.disabled && {
+      color: v.disabledColor,
+    }),
+  }),
+  fontRoot: ({ props: p, variables: v, theme: t }): ICSSInJSStyle => {
```
layershifter, line 46: Now we have separate roots for SVG and font icons that allows to apply overrides without conditions

packages/react/src/themes/base/components/Icon/iconVariables.ts
```diff
@@ -1,19 +1,34 @@
...
+  disabledColor: string
   horizontalSpace: string
+
+  smallestSize: string
+  smallerSize: string
+  smallSize: string
+  mediumSize: string
+  largeSize: string
+  largerSize: string
+  largestSize: string
```
layershifter, line 16: Size values where moved to variables

packages/react/src/themes/font-awesome/index.ts
```diff
@@ -0,0 +1,8 @@
+import { ThemeInput } from '../types'
+import * as componentStyles from './componentStyles'
+import { default as icons } from './components/Icon/iconNames'
+
+export default {
+  icons,
+  componentStyles,
+} as ThemeInput
```
mnajdova, line 8: Can we not expose this theme to the users? Everything is breaking if I choose this one on the Icon's docs. I don't see any benefit for the clients to explore this theme..

docs/src/examples/components/Avatar/Variations/AvatarExampleImageCustomization.shorthand.tsx
mnajdova: I am a bit confuse here. We are changing only this icon because Teams' theme doesn't support font icons? What about all other examples like: 'chess rook', 'book' etc. Are we planning to move all examples for now to reflect the Teams' theme icon names?

docs/src/examples/components/Icon/Usage/IconSetExample.shorthand.tsx
```diff
@@ -1,46 +1,46 @@
...
-          </Grid>
-        </div>
+const IconSetExampleShorthand: React.FunctionComponent<{ themeName: string }> = ({ themeName }) => (
+  <>
+    <div>
+      <Divider>
+        <Header as="h3" content="Regular" />
+      </Divider>
+      <Grid columns={4} styles={{ textAlign: 'center' }}>
+        {Object.keys(themes[themeName].icons).map(name => (
```
mnajdova, line 15: I understand the changes here and the reason why they were introduced, but if client's is testing adding new icon on this example, these example won't show it. However, I don't have any better proposal for now :(

packages/react/src/components/Icon/Icon.tsx
```diff
@@ -80,46 +79,28 @@ class Icon extends UIComponent<WithAsProp<IconProps>, any> {
...
+      { content: isSvgIcon && callable(maybeIcon.icon)({ classes, rtl }) },
+      {
+        defaultProps: {
+          as: ElementType,
+          className: cx(Icon.className, className),
+          ...accessibility.attributes.root,
+          ...unhandledProps,
+          styles: {
+            ...(isSvgIcon ? styles.svgRoot : styles.fontRoot),
+            ...styles.root,
```
mnajdova, line 98: Let's change the order of these and try to fix possible regressions. The svgRoot and fontRoot are more specific so I would expect that will be applied on top of styles.root. Btw, don't we have method for merging styles (`mergeStyles`)?

packages/react/src/themes/base/components/Icon/iconStyles.ts
```diff
@@ -61,27 +31,39 @@ const getPaddedStyle = (): ICSSInJSStyle => ({
...
+    const iconSpec = t.icons[p.name] || emptyIcon
```
mnajdova, line 47: Does the base theme have font awesome icons?? And if it provides styles for font awesome, why not svg as well? We have support for font and svg icons, but base theme implements only font. This seems awkward to me..

packages/react/src/themes/base/components/Icon/iconStyles.ts
mnajdova: I like that we have two things separated 👍

packages/react/src/themes/base/components/Icon/iconVariables.ts
mnajdova: Not sure that these size values are general enough for base theme...

packages/react/src/themes/font-awesome/components/Icon/iconStyles.ts
```diff
@@ -0,0 +1,11 @@
+import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
+import { IconProps } from '../../../../components/Icon/Icon'
+import { IconVariables } from '../../../teams/components/Icon/iconVariables'
+
+const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
+  fontRoot: (): ICSSInJSStyle => ({
+    fontWeight: 900, // required for the fontAwesome to render
```
mnajdova, line 7: It's great that the only required override is actually related to the font awesome icons. I would really like the same thing to be applicable for svgs :)

packages/react/src/themes/teams/components/RadioGroup/radioGroupItemStyles.ts
```diff
@@ -64,6 +64,7 @@ const radioStyles: ComponentSlotStylesInput<
     // overrides from icon styles
     backgroundColor: 'transparent',
     boxShadow: 'none',
+    boxSizing: 'border-box',
```
mnajdova: Shouldn't this be applied by default?? We are using only svg icons and they already have border-box

packages/react/src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -72,47 +30,30 @@ const getIconSize = (size: SizeValue, sizeModifier: IconSizeModifier): number =>
...
-      ...((bordered || v.borderColor) &&
+      ...((p.bordered || v.borderColor) &&
         getBorderedStyles(v.borderColor || getIconColor(v, colors))),
     }
   },
+  svgRoot: ({ props: p, variables: v }): ICSSInJSStyle => {
+    return {
+      display: 'inline-block',
+      backgroundColor: v.backgroundColor,
+      boxSizing: 'border-box',
```
mnajdova: No need for specifying boxSizing: 'border-box'. Teams theme users only svg icons, and there is no prev definition of 'content-box' (it is for font icons in base theme, not svg)

docs/src/examples/components/Avatar/Variations/AvatarExampleImageCustomization.shorthand.tsx
layershifter: > We are changing only this icon because Teams' theme doesn't support font icons?

This slot in attachment requires an additional styling for font icons that is not present, so yes.

packages/react/src/themes/base/components/Icon/iconVariables.ts
layershifter: Base theme should support `size` prop properly and we should avoid nesting in variables. Do you have better proposal? 🐱

docs/src/examples/components/Icon/Usage/IconSetExample.shorthand.tsx
layershifter: > if client's is testing adding new icon on this example

Why? If it will be added directly to a theme it will work as expected...

packages/react/src/components/Icon/Icon.tsx
layershifter: > Let's change the order of these and try to fix possible regressions. The svgRoot and fontRoot are more specific so I would expect that will be applied on top of styles.root. 

👍 

> Btw, don't we have method for merging styles (mergeStyles)?

It's plain objects so I don't see any need in an additional function.

packages/react/src/themes/base/components/Icon/iconStyles.ts
layershifter: > Does the base theme have font awesome icons?

Nope.

> And if it provides styles for font awesome, why not svg as well? 

It provides styles for our required icons like `stardust-arrow-up` which are characters in Base theme.

packages/react/src/themes/base/components/Icon/iconVariables.ts
mnajdova: I agree that we should have variables and use this, I was saying about the size values (which now reflect the Teams theme) If you think these values are general and should be defined as this, then I am okay

packages/react/src/themes/teams/components/RadioGroup/radioGroupItemStyles.ts
layershifter: Based on offline discussion we decided to add a new icon and use it.

mnajdova (approved):

PR #1340, opened 2019-05-15
chore(colors): Improve colors typings per component variables

mnajdova:
Things addressed in this PR:

- In teams theme, we want to scope the color used to the ones actually defined in the theme (introduced generic for this)

- We want from component's variables to export the parts of the color scheme only the areas actually used in the component. For example for the Label component, we are interested only in the background and foreground colors. (introduced two more generics for this, as well as stricter version of ones of the existing ones' introduced helper method: `pickValuesFromColorScheme` for making this easier for developers)

- removed `getColorScheme`, and used `getColorSchemeKey` in order not to complicate too much the functions with generics that will have to be added by devs.

- renamed `emphasisColors` to `contextualColors` in Teams theme, as we have only the brand value there (we don't have primary and secondary colors) - open for other possible names here.

- added transparent colors in the color palette, those were added later in the process

- added Not defined information on the Color Box for being more clear, rather then just showing transparent box


packages/react/src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -1,9 +1,16 @@
 import { pxToRem } from '../../../../lib'
-import { SiteVariablesPrepared, ColorSchemeMapping } from '../../../types'
-import { extendColorScheme } from '../../../colorUtils'
+import { SiteVariablesPrepared, ColorSchemeMapping, ColorScheme } from '../../../types'
+import { extendColorScheme, pickValuesFromColorScheme } from '../../../colorUtils'
+
+// TODO extract this to a util
+const tuple = <T extends string[]>(...args: T) => args
```
kuzhelov: why generic and not just `const tuple = (...args: string[]) => args` ?

packages/react/src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -1,9 +1,21 @@
...
+  SiteVariablesPrepared,
+  ColorSchemeMapping,
+  ColorScheme,
+  ComponentAreaName,
+} from '../../../types'
+import { extendColorScheme, pickValuesFromColorScheme } from '../../../colorUtils'
+
+// TODO extract this to a util
+const tuple = <T extends string[]>(...args: T) => args
+const labelColorComponentAreasTuple = tuple<ComponentAreaName[]>('foreground', 'background')
```
mnajdova: @kuzhelov we want to use here the `ComponentAreaName` values only, not just any string, that's why I've added the generic, but forgot to specify it. Sorry my bad. Is it more clear now?

kuzhelov: actually, got the intent - essentially, you would like to have strict string literal type in TS _being projected to run-time array_, now got it. Couple of things that we could simplify here:

```js
const tuple = <T extends string[]>(...args: T) => args
// to
const tuple = <T extends ComponentAreaName>(...args: T[]) => args
```

```js
const labelColorComponentAreasTuple = tuple<ComponentAreaName[]>('foreground', 'background')
// to
const labelColorComponentAreasTuple = tuple('foreground', 'background') // no need for explicit generic param
```

kuzhelov: also, we could rename `labelColorComponentAreasTuple ` to just `labelColorComponentAreas` - I think there is no need to specify type in the name, it will make things easier to read

packages/react/src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -15,16 +27,21 @@ export interface LabelVariables {
...
-    colorScheme: extendColorScheme(siteVars.colorScheme, {
-      default: {
-        background: color,
-        foreground: 'rgb(232, 232, 232)',
-      },
-      brand: {
-        background: siteVars.colorScheme.brand.foreground4,
-      },
-    }),
+    colorScheme: pickValuesFromColorScheme<LabelColorComponentAreas>(
```
kuzhelov: no need to specify generic parameter here if for `colorUtils.ts` the following changes will be introduced:

```js
export function pickValuesFromColorScheme<T extends Partial<ComponentAreaName>>(
  colorScheme: ColorSchemeMapping,
  componentAreaSubset: T[],    // <--------------------- use generic arg here
):
```

packages/react/src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -15,16 +27,21 @@ export interface LabelVariables {
 export default (siteVars: SiteVariablesPrepared): LabelVariables => {
   const color = 'rgba(0, 0, 0, 0.6)'
 
+  const colorScheme = extendColorScheme(siteVars.colorScheme, {
+    default: {
+      background: color,
+      foreground: 'rgb(232, 232, 232)',
+    },
+    brand: {
```
kuzhelov, line 24: not sure if you were going to address this by the set of changes provided, but as a client I would expect that `brand` would appear under props of `LabelColorSchemeMapping`, so that at least types would suggest me that this is a safe thing to use:

image

packages/react/src/themes/colorUtils.ts
```diff
@@ -16,3 +21,23 @@ export const extendColorScheme = (
   })
   return result
 }
+
+export function pickValuesFromColorScheme<T extends Partial<ComponentAreaName>>(
```
kuzhelov: the intent of this method is to take specified areas from _each_ color scheme - in that case I would expect that even for the `default` one we would have only `foreground` and `background` for `Label`. However, types suggests that for `default` we still have all of the areas originally defined:

image

It is necessary that we would marry types and actual implementation, in a way that real code will guarantee type correctness

packages/react/src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -15,16 +27,21 @@ export interface LabelVariables {
...
+      background: siteVars.colorScheme.brand.foreground4,
```
kuzhelov, line 25: not sure if this PR is going to address this, but as a client might see another issue with types. Here you might see that types suggest that `foreground` prop is guaranteed to be defined for each `color` of `colorSheme`. 

image

However, given this code, it is clear that value might be omitted - this is the case of `brand` color for Label's color scheme, where `foreground` is absent.

It would be great if our types will suggest the fact that value of area props (in our example, `foreground`) is, in fact, optional

mnajdova: The color scheme defined for the teams theme, already have all these values defined. In the labelVariables, we are just overriding some of those. Basically there is no need to override the brand foreground hover, that's why it is not defined.

packages/react/src/themes/colorUtils.ts
mnajdova: This was fixed: 
image

I had to introduce `StrictColorSchemeMapping` as a new type, plus I had prev generic that I was not using... It's behaving as expected now. Thanks for the catch!

packages/react/src/themes/teams/components/Label/labelVariables.ts
mnajdova: The problem is that now, we are using one ColorNames type, but this should be redefine for teams theme (we don't have all colors that the typings suggest, and we have few more. Will introduce this changes too.

packages/react/src/themes/teams/components/Label/labelVariables.ts
mnajdova: Changed, thanks for the suggestions

packages/react/src/themes/teams/components/Label/labelVariables.ts
mnajdova: Introduced TeamsColorNames, a bit scared as how these typings with generics are looking now :S I am open for ides for restructuring them. Will try on my own to see how can I improve these.

packages/react/src/themes/teams/colors.ts
```diff
@@ -1,16 +1,26 @@
...
-  ColorVariants,
-  ContextualColors,
-} from '../types'
-import { Extendable } from '../../types'
-
-export const emphasisColors: ContextualColors = {
+import { ColorPalette, PrimitiveColors, ColorSchemeMapping, ColorVariants } from '../types'
+import { ColorScheme } from 'src/themes/types'
+
+export type TeamsContextualColors = {
```
mnajdova: Interfaces added for the Team's specific colors

packages/react/src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -1,9 +1,25 @@
...
+
+// TODO extract this to a util
+const tuple = <T extends ComponentAreaName>(...args: T[]) => args
+const labelColorComponentAreas = tuple('background', 'foreground')
+type LabelColorComponentAreas = typeof labelColorComponentAreas[number]
+
+type LabelColorSchemeMapping = StrictColorSchemeMapping<
+  StrictColorScheme<LabelColorComponentAreas>,
+  TeamsColorNames
+>
```
mnajdova: I want us to decide if this is the right approach for defining these, and I will extract this to a separate utility and reuse it in other component variables files.

packages/react/src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -1,9 +1,25 @@
...
 
 export interface LabelVariables {
-  colorScheme: ColorSchemeMapping
+  colorScheme: LabelColorSchemeMapping
```
mnajdova, line 10: This may be a problem when we are using here some base theme interface and want to extend this, as this prop should have a different type..

docs/src/components/ColorBox.tsx
```diff
@@ -86,7 +86,7 @@ const ColorBox = createComponent<ColorBoxProps>({
           <div className={classes.value}>
             <span onClick={onClick}>
               {value && <Icon name={active ? 'checkmark' : 'copy outline'} size="small" />}
-              {value}
+              {value || 'Not defined'}
```
kuzhelov: 👍

docs/src/views/ColorPalette.tsx
```diff
@@ -59,7 +59,9 @@ const ColorPalette = () => (
     }}
   >
     <ProviderConsumer
-      render={({ siteVariables: { colors, emphasisColors, naturalColors } }) => (
+      render={({
+        siteVariables: { colors, contextualColors, naturalColors, transparentColors },
```
kuzhelov, line 63: probably it is better to introduce transparent colors in separate PR, given that this one is focused on color types

packages/react/src/themes/colorUtils.ts
```diff
@@ -16,3 +21,23 @@ export const extendColorScheme = (
...
+  colorScheme: ColorSchemeMapping,
+  componentAreaSubset: T[],
```
kuzhelov: maybe just `componentAreas`? I mean, this would be sufficient to understand what this method is about, as well as to read its code

packages/react/src/themes/colorUtils.ts
kuzhelov: maybe simpler `pickColors` name, given that arguments are descriptive enough?

packages/react/src/themes/teams/colors.ts
```diff
@@ -1,16 +1,26 @@
...
+  brand: ColorVariants
```
kuzhelov: on `ColorVariants` - currently this type is defined as type that any Stardust theme could use. However, we don't have any piece of Stardust code that relies on this structure (actually, good thing, as different themes might use different schemes to organise colors), and Teams is the only theme which uses it. 

I think that it is logical to move this type from common to the types of Teams theme

packages/react/src/themes/teams/colors.ts
```diff
@@ -122,7 +132,7 @@ export const primitiveColors: PrimitiveColors = {
   white: '#fff',
 }
 
-export const transparentColors = {
+export const transparentColors: TeamsTransparentColors = {
   silver: {
     100: undefined,
```
kuzhelov, line 131: just for the curiosity sake: these values that are `undefined` technically may be omitted from declarations, so that instead of 

```js
 yellow: {
    50: undefined,
    100: '#FBF6D9', // old message highlight color
    200: undefined,
    300: '#F9EC02', // old acc critical ufd icon color
    400: '#F8D22A', // old siteVariables.yellow, $app-yellow, yellow[900]
    500: '#FFB900', // old $bcast pre live color
    600: '#FFAA44', // new away presence
    700: undefined,
    800: undefined,
    900: undefined,
  },
```

we might define just what actually defined:

```js
 yellow: {
    100: '#FBF6D9', // old message highlight color
    300: '#F9EC02', // old acc critical ufd icon color
    400: '#F8D22A', // old siteVariables.yellow, $app-yellow, yellow[900]
    500: '#FFB900', // old $bcast pre live color
    600: '#FFAA44', // new away presence
  },
```

However, I might assume there is a reason why explicit `undefined` values are present - is it done for the sake of explicitness?

packages/react/src/themes/teams/components/Label/labelVariables.ts
kuzhelov: are we expecting this `TeamsColorNames` set to change based on the component type? I think, we might start from exporting simpler `StrictTeamsColorSchemeMapping<TComponentAreas>` type that would do this scaffolding in the background. This will make type interface being easier to client.

Also, we might want to rename this type to something like `TeamsReducedSchemeMapping<TComponentAreas>`

packages/react/src/themes/teams/components/Label/labelVariables.ts
kuzhelov: just to be sure this is not overlooked - could you, please, suggest an example of this (like, some pseudo-code)? Because at the moment I think that it would be normal to extend this Teams interface in that case - and semantically that would mean that now this colors' set is dependent on colors object (or, as equivalent, its type) provided by base theme.

mnajdova: Sure, let me provide an example. The text component in the base themes, specifies some color names, that may won't exists in the Teams theme (like blue for example). That means, that the interface that we use here will be different. For example:

```
//base theme:
interface TextVariables {
  colorScheme: ColorSchemeMapping<ColorScheme, ColorNames>
}

//Teams theme:
interface TextVariables {
  colorScheme: ColorSchemeMapping<ColorScheme, TeamsColorNames>
}
```

packages/react/src/themes/colorUtils.ts
mnajdova: Hmm, it's not about picking colors is more about picking areas (like foreground, background)... So maybe `pickColorAreas` would be better?

packages/react/src/themes/teams/colors.ts
mnajdova: The base theme is using these typings as well. Yes is not required by any react specific file, but still is used in base theme..

packages/react/src/themes/teams/colors.ts
mnajdova: Yes, exactly, this are defined as not defined in the designs as well, because we said that it will be more obvious that those are not defined for this color, instead of just missed. Also in the docs, we want to emphasize the same point, when showing the color palette.

docs/src/views/ColorPalette.tsx
mnajdova: Okay, will revert these changes from here.

packages/react/src/themes/teams/components/Label/labelVariables.ts
mnajdova: I would expect that those will be different across different component, as the text have much more color values then the menu for example. (The menu supports only default and brand colors and we want to use that as typings in the styles.)

packages/react/src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -1,9 +1,25 @@
...
+  SiteVariablesPrepared,
+  StrictColorSchemeMapping,
+  StrictColorScheme,
+  ComponentAreaName,
+} from '../../../types'
+import { extendColorScheme, pickValuesFromColorScheme } from '../../../colorUtils'
+import { TeamsColorNames } from '../../colors'
+
+// TODO extract this to a util
+const tuple = <T extends ComponentAreaName>(...args: T[]) => args
```
bmdalex: `tuple` is a bit misleading, not sure what this does..

bmdalex: ah, I understand now..by looking on next line. pls add JSdoc and example of usage in comments, would be helpful

packages/react/src/themes/teams/components/Menu/menuDividerStyles.ts
```diff
@@ -1,11 +1,11 @@
 import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
 import { MenuDividerProps } from '../../../../components/Menu/MenuDivider'
 import { MenuVariables } from './menuVariables'
-import { getColorScheme } from '../../colors'
+import { getColorSchemeKey } from '../../colors'
 
 const menuDividerStyles: ComponentSlotStylesInput<MenuDividerProps, MenuVariables> = {
   root: ({ props: p, variables: v }): ICSSInJSStyle => {
-    const colors = getColorScheme(v.colorScheme, null, p.primary)
+    const colors = v.colorScheme[getColorSchemeKey(null, p.primary)]
```
bmdalex: ouch..this doesn't look too nice, I will need to pull the branch to understand why we need to make this change; the previous `getColorScheme` was much friendlier for the theme developer

bmdalex: no suggestion for now as lacking a bit of context, sorry

bmdalex (commented):
I'd need to deep dive to understand all the changes, let's chat in person

packages/react/src/themes/teams/components/Label/labelVariables.ts
mnajdova: Yep, will add this, thank you!

mnajdova: I refactored this to use an util function. This is the current code:
```
const labelColorComponentAreas = generateComponentAreas('foreground', 'background')
type LabelColorComponentAreas = typeof labelColorComponentAreas[number]

type LabelColorSchemeMapping = StrictColorSchemeMapping<
  StrictColorScheme<LabelColorComponentAreas>,
  TeamsColorNames
>
```
Does this make more sense now?

packages/react/src/themes/teams/components/Menu/menuDividerStyles.ts
mnajdova: I will add Teams specific interfaces to this method and will revert to using it, as this is seen clearly as something used only in the Teams theme.

packages/react/src/themes/teams/components/Text/textVariables.ts
```diff
@@ -8,6 +19,7 @@ export interface TeamsTextVariables extends TextVariables {
 
 export default (siteVariables): Partial<TeamsTextVariables> => {
   return {
+    // TODO: how can we specify these typings here, when this is already defined and used in the base theme!?
```
mnajdova: @kuzhelov this is the problem I was talking about (having different typings for the color schemes in base vs teams theme. How can we address this?

mnajdova: We may use generic for the color scheme typing, not sure how will it go with the overriding, as we are using this in the styles...

kuzhelov: discussed with @mnajdova and agreed to introduce generic type for variables in the base theme, with default type value of `ColorSchemeMapping`:

```js
// base theme
export interface TextVariables<TColorSchemeMapping = ColorSchemeMapping> { ... }
```

This will allow client to restrict/modify declarations of schema types in those cases where it is necessary, and `ColorSchemeMapping` is a good type to capture general cases.

packages/react/src/themes/teams/components/Text/textVariables.ts
```diff
@@ -1,4 +1,15 @@
 import { TextVariables } from '../../../base/components/Text/textVariables'
+import { StrictColorSchemeMapping, StrictColorScheme } from '../../../../themes/types'
+import { TeamsColorNames } from '../../colors'
+import { generateComponentAreas } from '../../../colorUtils'
+
+const textColorComponentAreas = generateComponentAreas('foreground')
+export type TextColorComponentAreas = typeof textColorComponentAreas[number]
+
+export type TextColorSchemeMapping = StrictColorSchemeMapping<
```
kuzhelov: agreed to simplify the set of typings that are provided as part of client's API, to ensure that client needs to specify only those type aspects that actually change (names of types are subject to change):

```jsx
// this is the simplest type that could be used on the client side
type TeamsSchemeMapping<TAreas extends string> =  StrictColorSchemeMapping<
  StrictColorScheme<TAreas>,
  TeamsColorNames
>

type ItemType<T> = T extends (infer TItem)[] ? TItem : never
```

Then, to simply define scheme mapping for Teams, on the client side it will be sufficient to use the following type declaration:

```jsx
type TextColorSchemeMapping = TeamsSchemeMapping<'foreground'> 

// or, with more explicit naming strategy for the same type
type TextColorSchemeMapping = TeamsSchemesWithAreas<'foreground'>
```

and, in case if item type of runtime array is intended to be used, then

```jsx
const textColorAreas = stringLiteralsArray('foreground')
type TextColorSchemeMapping = TeamsSchemesWithAreas<ItemType<typeof textColorAreas>>
```

packages/react/src/themes/teams/components/Text/textVariables.ts
```diff
@@ -1,4 +1,15 @@
 import { TextVariables } from '../../../base/components/Text/textVariables'
+import { StrictColorSchemeMapping, StrictColorScheme } from '../../../../themes/types'
+import { TeamsColorNames } from '../../colors'
+import { generateComponentAreas } from '../../../colorUtils'
+
+const textColorComponentAreas = generateComponentAreas('foreground')
```
kuzhelov: actually, `generateComponentAreas ` encapsulates quite generic logic - essentially, the intent of this method is to create array of string literals _with items type being inferred from these string literals_, so it won't be just generic `string` type in case of `['foreground']`.

Thus, what if we would create generic method for this, something like 
```jsx
// stringLiteralsArray() method ensures 
// that textColorComponentAreas will be of type ('foreground' | ... )[], instead of string[]
const textColorComponentAreas = stringLiteralsArray('foreground',  ...)`
```

bmdalex (approved):
👍 looks good to me, feel free to merge once comments from other reviewers are resolved

docs/src/components/ColorBox.tsx
```diff
@@ -14,6 +14,7 @@ import CopyToClipboard from './CopyToClipboard'
 type ColorBoxProps = {
   children?: React.ReactNode
   name?: string
+  header?: boolean
```
kuzhelov: the name of the prop is a bit misleading when this is used on the client side:

```jsx
  <ColorBox
        header={true}  // what is the effect of that? and it is tempting to write like 'header="my header"'
        name={token}
        key={`${token}schema`}
        size="small"
```

Maybe we could come up with some better option? What are the cases this prop should differentiate? According to implementation, it is about  rendering copy to clipboard icon or not. Maybe we should name it accordingly, like `copyToClipboardIcon={true}`?

packages/react/src/themes/types.ts
```diff
@@ -131,15 +134,71 @@ export type ComponentAreaName =
...
+  Scheme = ColorScheme,
+  Colors extends string | number | symbol = ColorNames
+> = ColorValues<Partial<Extendable<Scheme, string>>, Colors> & {
   default?: Partial<Extendable<ColorScheme, string>>
 }
 
+/**
+ * Teams theme specific interfaces
+ */
+export type TeamsContextualColors = {
```
kuzhelov: lets move them `teams/types.ts`, as these are specific to Teams theme only

kuzhelov (approved):

packages/react/src/themes/types.ts
mnajdova: Moved, thanks for the suggestion!

PR #1341, opened 2019-05-15
fix(Popup): fix positioning in scrollable container

### LINKED ISSUE #1335: Popup: positioning is broken inside overflow containers
### Steps
1. Open https://codesandbox.io/s/7k7lpm1761
2. Scroll in inner container to a button
3. Click on a button to open Popup

### Expected Result

sd-popup-fix

Popup is attached to a point inside container

### Actual Result

sd-popup-bug

Popup is attached to a point inside window 💣

### Version
0.30.0

layershifter:
Fixes #1335.

`Popper.JS` relies on [`getScrollParent()`](https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/utils/getScrollParent.js) function which recursively uses [`getParentNode()`](https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/utils/getParentNode.js) under hood. `getParentNode()` uses `parentNode` field and our `ReferenceProxy` doesn't have it before 😢


bmdalex (approved):
👍

PR #1350, opened 2019-05-17
feat: component variants

kuzhelov:
# Component Variants

Current Stardust options are limited to address the following situation on the client side.

Suppose that there is one category of buttons that should appear with icon's borders being rounded, and another one - with icon's border being rectangular. 

This, essentially, results in need to define two variants of Button, with `icon` slot being styled differently, but **with all the behavioural (and accessibility) aspects of `Button` being preserved**. 

Also, we would like to ensure that these `Button` variants won't spawn in uncontrolled way - i.e. all of them will be registered in theme, discoverable and, in general, will match `Button` variants of the Design Spec.

**Component Variants feature:**

- adjusts component's appearance, while preserving behavior aspects of original component
- allows client to extend set of props of original component, in controlled way
  - this should be seen as an extensibility point to introduce design terms that are specific to client's domain
- provides 'is a' semantics - props, behavior and accessibility of the original component (or, in general, all the _scenarios_ of original component) will work for its variant

### API
```js
// --------------------------------
// DEFINE THEME BITS (styles)
// theme/index.ts
// --------------------------------
const theme = {
  componentStyles: { .. }
  ...
  componentVariants: {
    Input: { // <------------- which component variant defined for
      Primitive: {     // <--------------- variant name
         // 'variables' and 'styles' props are optional,
         // will be applied on top of corresponding ones of 'Input'
         variables: { 
            background: transparent,
            underlineOnFocus: null
         }
         styles: {
             ....
         }
      }
    }
  }
}

// --------------------------------
// DEFINE COMPONENT FOR VARIANT
// components/index.ts
// --------------------------------
import { createVariant } from '@stardust-ui/react'

// propTypes and defaultProps are optional
export const PrimitiveInput = createVariant(Input, 'Primitive', propTypes = { }, defaultProps = { })

// --------------------------------
// USE VARIANT
// client.ts
// --------------------------------
<PrimitiveInput       // <----- all props of Input are supported
  placeholder='Your name' 
  icon={..} ... />
```

## Example: <Text /> variant with client-specific `timestamp` and `userMention` props
Lets imagine that we have `Text` component (with general-use props API), and we want to introduce `ChatText` variant which would support `timestamp` and `userMention` terms - the ones that are _specific to our chat app_.

### Problem
Previously the options were
- introduce props as part of Stardust API, which is problematic due to
   - client cannot rely on that
   - Stardust spec is not able to absorb terms that are not general enough
- introduce variables to support this case, and this had resulted in awkward API and, what is more important, broken styles encapsulation on the client side:

```jsx
// mix of props and variables to express the same thing
<Text important variables={{ userMention: true }} /> 

// textStyles.ts
const styles = {
  root: ({ ..., variables }) => ({
     /* .. MIXED list of other variants styles - BROKEN ENCAPSULATION .. */
     ...(variables.userMention && { ... })
  })
}
```
### Solution

Now this scenario may be addressed with variants - note that styles encapsulation, semantics and design language's expressiveness is preserved in the JSX markdown

```jsx
// --------------------------------
// DEFINE
// --------------------------------

// types/index.ts
type ChatTextProps = {
  userMention: boolean,
  timestamp: boolean
}

// theme/index.ts
const theme = {
  componentStyles: ...,
  componentVariants: {
    Text: {
       ChatText: {
          styles: ({ props: TextProps & ChatTextProps }) => ({
             ...(props.userMention && { /* .. userMention styles  .. */ })
             ...(props.timestamp && { /* .. timestamp styles  .. */ })
          )}
       }
    }
}

// components/index.ts
export const ChatText = createVariantOf<TextProps, ChatTextProps>(
  Text,  // of component
  'ChatText', // variant name
  { atMention: PropTypes.bool, timestamp: PropTypes.bool } // propTypes
)

// --------------------------------
// USE
// --------------------------------
<ChatText timestamp>11:03 PM</ChatText>
```


packages/react/src/themes/types.ts
```diff
@@ -285,6 +285,7 @@ export interface ThemeInput {
   siteVariables?: SiteVariablesInput
   componentVariables?: ThemeComponentVariablesInput
   componentStyles?: ThemeComponentStylesInput
+  componentVariants?: any
```
kuzhelov: will be addressed

jdhuntington:

PR #1352, opened 2019-05-17
fix(flex): fill styles

### LINKED ISSUE #1348: Nested Flex components don't fill correctly
### Steps
When rendering a Flex within another Flex, the inner flex does not fill the parent when the fill property is passed.

In my use case, I have a component passing a JSX.Element (A) as a prop to another component (B) to render somewhere in its tree. B renders A inside of a Flex, since it's part of a complex layout. However, A is itself a Flex, since it has its own complex layout. When I try to make A fill the available space in B by setting the fill property, it does not correctly fill the available space.

Because A is being created elsewhere and passed into B, I cannot make clever assumption about the parent and replace the Flex in A with Flex.Items. It's possible that A may need to be rendered somewhere else where it is not wrapped in a Flex, and so if it was just a collection of Flex.Item elements, then it would render incorrectly in those cases.

### Expected Result
The inner Flex fills the available space in the parent Flex

### Actual Result
The inner Flex doe not fill the available space in the outer Flex. 💣

### Version
0.27.0

### Testcase
https://codesandbox.io/s/2g40z

kuzhelov:
Fixes #1348 

This PR addresses the issue of expected behavior of `<Flex fill />`, and makes it aligned with the behavior suggested by spec:

> fill - Orders container to fill all parent's space available.

## BREAKING CHANGES

Effect of this fix is changed behavior of `fill` prop of `Flex` component in case if it is a child of other `Flex`. 

```jsx
<Flex>
  <Flex fill>
  </Flex>
</Flex>
```

image

**If client would like to preserve this look**, i.e. inner `Flex` to take only the minimal width necessary (and not the entire space provided by outer `Flex`), then 
- `fill` prop should be removed for inner `Flex`
- provide `vAlign=stretch` prop to outer `Flex`

```jsx
<Flex vAlign='stretch'>
  <Flex></Flex>
</Flex>
```

------------------------

## Description

From https://codesandbox.io/s/ulrbd?module=/example.js :
```jsx
// outer
<Flex debug .. >
  <Flex fill .. >
     Inner Flex
  </Flex>  
</Flex>
```

### Was
image

### Now
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### BREAKING CHANGES
 - Restrict Typescript checks for component props @kuzhelov ([#1290](https://github.com/stardust-ui/react/pull/1290))
 - Aligned focus styles for `Chat.Message` component with latest Teams theme design @Bugaa92 ([#1269](https://github.com/stardust-ui/react/pull/1269))
+- Fix styles for `fill` prop of `Flex` @kuzhelov ([#1352](https://github.com/stardust-ui/react/pull/1352))
 
 ### Fixes
 - Fixed `Flex.Item` children prop type @mnajdova ([#1320](https://github.com/stardust-ui/react/pull/1320))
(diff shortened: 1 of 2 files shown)


mnajdova (approved):
Would suggest adding a screenshot test for catching possible regressions. Other than that looks good :)

kuzhelov:
@mnajdova, just verified that there is no valid use case for the nested single `Flex` in `Flex` - case where, actually, issue was discovered. Would refrain from providing very contrived example in docs for now

PR #1353, opened 2019-05-17
feat(colors): add red color scheme in Teams theme

mnajdova:
- added red color scheme specified by the design theme
- fixed some color schemes for the colors still not red-lined with design, in order to improve the usage showcase in the mean time


packages/react/src/themes/teams-dark/colors.ts
```diff
@@ -40,13 +40,16 @@ export const colorScheme: ColorSchemeMapping = {
   default: {
     foreground: colors.white,
     foreground1: colors.grey[250],
-    foreground3: colors.grey[300],
-    foreground4: colors.white,
+    foreground2: colors.grey[300],
```
mnajdova, line 43: Fixed wrong name of color scheme token, added two new approved by design.

packages/react/src/themes/teams-dark/colors.ts
```diff
@@ -40,13 +40,16 @@ export const colorScheme: ColorSchemeMapping = {
...
+    foreground3: colors.white,
+    foreground4: colors.grey[650],
+    foreground5: colors.grey[500],
 
     background: colors.grey[650],
     background1: colors.grey[700],
     background2: colors.grey[800],
     background3: colors.grey[550],
+    background4: colors.grey[250],
```
mnajdova: Added new color scheme token approved by design.

packages/react/src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -9,29 +9,44 @@ export interface LabelVariables {
...
+  const invertedColorScheme = invertColorScheme(siteVars.colorScheme)
 
   return {
-    colorScheme: extendColorScheme(siteVars.colorScheme, {
+    colorScheme: extendColorScheme(invertedColorScheme, {
+      // TODO use here some color scheme values, after the designs are delivered
       default: {
-        background: color,
-        foreground: 'rgb(232, 232, 232)',
+        foreground: 'rgba(0, 0, 0, 0.6)',
```
mnajdova: Would really like to use here some color scheme values, because these overrides now goes to all themes (this is why we have to add labelVariables in teams and HC theme), which are actually partially copies from the light theme :(

packages/react/src/themes/teams/components/Label/labelStyles.ts
```diff
@@ -14,8 +14,8 @@ const labelStyles: ComponentSlotStylesInput<LabelProps, LabelVariables> = {
       overflow: 'hidden',
       height: v.height,
       lineHeight: v.height,
-      color: colors.background,
-      backgroundColor: colors.foreground,
+      color: colors.foreground,
+      backgroundColor: colors.background,
```
mnajdova: We are using now the foreground for color and the background for background.

packages/react/src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -9,29 +9,44 @@ export interface LabelVariables {
   startPaddingLeft: string
   endPaddingRight: string
   height: string
-  iconColor: string
+}
+
+export const invertColorScheme = colorScheme => {
```
layershifter:
```suggestion
export const invertColorScheme = (colorScheme: ColorScheme) => {
```

Please define input's type

packages/react/src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -9,29 +9,44 @@ export interface LabelVariables {
...
+  return Object.keys(colorScheme).reduce((accumulator, key) => {
```
layershifter:
```suggestion
  return Object.keys(colorScheme).reduce((accumulator, colorName: string) => {
```

miroslavstastny (approved):

packages/react/src/themes/teams/components/Label/labelVariables.ts
```diff
@@ -9,29 +9,44 @@ export interface LabelVariables {
...
+        foreground: background,
+        background: foreground,
+      },
+    }
+  }, {})
 }
 
 export default (siteVars: SiteVariablesPrepared): LabelVariables => {
-  const color = 'rgba(0, 0, 0, 0.6)'
+  const invertedColorScheme = invertColorScheme(siteVars.colorScheme)
```
layershifter: So, we are using `invertColorScheme` only to flip `background` and `foreground`? I am little confused with it because when I see `inverted` I expect this: https://semantic-ui.com/elements/button.html#inverted

The another problem that I will see `foreground` color in `styles`, will try to find it in `colorScheme` and will fail 😕

mnajdova: This inverted color scheme is only for the Label component, not in general, and the Label implements the inverted only by flippin the foreground and background - this are the only colors used. The label does not support inverted prop, it's just that it inverts the default color scheme. For the foreground, I don't understand what you mean it is not defined. All colors in the color scheme have the foreground and background colors. Can you provide more details on the second part?

More clarification on the inverted is that, design is considering whether we should add inverted color scheme as part of teams, or whether we should add more foreground and background values in the original one. Until that is decided, we just want to remove the confusion of using foreground on the backgroundColor css prop, and vise versa.

layershifter: > For the foreground, I don't understand what you mean it is not defined. All colors in the color scheme have the foreground and background colors. Can you provide more details on the second part?

https://github.com/stardust-ui/react/pull/1353#discussion-diff-285065169R17

I see there `color: colors.foreground,`, but when I will go to `packages/react/src/themes/teams/colors.ts` there will another value for `foreground` because we flipped them, correct?

mnajdova: Yes, but the colors are generated from the variable's color scheme, not the site variables.
```
 const colors = getColorScheme(v.colorScheme, p.color)
```

layershifter: Agree, but as I see in the most cases we simply passing values:
```
colorScheme: siteVars.colorScheme,
```

mnajdova: @miroslavstastny can we hear your opinion on this one?

PR #1354, opened 2019-05-17
chore(docs): use babel-loader

### LINKED ISSUE #508: RFC: switch to Babel

layershifter:
Related #508.

### Enable `isolatedModules` in TS `compilerOptions`

Babel behaves as if the `--isolatedModules` option was passed to the TypeScript Compiler and this can't be worked around because Babel doesn't support cross-file analysis.

How does Babel handle TypeScript code? **It removes it**, it strips out all the TypeScript, turns it into "regular" JavaScript. Any bundler don't know anything about types because don't exists in runtime and that's why types can't be reexported.

https://babeljs.io/docs/en/babel-plugin-transform-typescript

### Updates in exports

As mentioned before we can't reexport types, that's why exports were transformed:

#### Before
```ts
export { default as Accordion, AccordionProps } from './Accordion'
```
#### After
```ts
export { default as Accordion } from './Accordion'
export * from './Accordion' // will AccordionProps and everything except default
```

#### `types.internal.ts`

These files will contain shared and reusable types in package, but these types shouldn't be part of a public API and can't be exported.

```js
export * from './types' // contains only types that are part of public API
```

### Remove `awesome-typescript-loader` & `ts-loader`

Was replaced with `babel-loader` and `fork-ts-checker-webpack-plugin` to keep type checks.

### Add `fork-ts-checker-webpack-plugin`

Webpack plugin that runs TypeScript type checker on a separate process. It allows very fast compile code by `babel-loader` and run type checks parallel, so by switching to Babel we will not loose TS checks.

---

_This PR do not resolve issues with HMR._


packages/react/src/components/Loader/Loader.tsx
```diff
@@ -142,6 +142,7 @@ class Loader extends UIComponent<WithAsProp<LoaderProps>, LoaderState> {
           {Box.create(label, {
             defaultProps: { className: Loader.slotClassNames.label, styles: styles.label },
           })}
+          222
```
miroslavstastny: ?

layershifter: Cruft, removed it :)

build/tsconfig.commonjs.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "./tsconfig.common.json",
   "compilerOptions": {
+    "isolatedModules": false,
```
kuzhelov, line 4: this config and the next one are used for `yarn build` scenarios, where Babel is not used.

This is necessary to ensure that type declarations will be generated.

docs/src/components/ComponentDoc/ComponentExample/index.tsx
```diff
@@ -1 +1,2 @@
-export { default, ComponentExampleProps } from './ComponentExample'
+export * from './ComponentExample'
```
kuzhelov, line 1: these changes in exports are necessary to make TS with `--isolatedModules` flag work - this flag ensures that TS compiler uses the same 'file-by-file' processing strategy as Babel

packages/react/src/types.ts
```diff
@@ -8,13 +8,8 @@ export type Extendable<T, V = any> = T & {
   [key: string]: V
 }
 
-export type Nullable<T> = T | null
```
kuzhelov, line 11: all these are unused, right?

perf/src/index.tsx
```diff
@@ -1,3 +1,5 @@
+import '@babel/polyfill'
```
kuzhelov, line 1: could you, please, suggest why this change was necessary? where it is used?

kuzhelov:
just couple of moments that we might add to the description (all those are mentioned in the comments):
- brief explanation on how transpilation logic has changed - from recursive modules resolution used by Typescript to isolated files/modules used by Babel. This will serve as concise reasoning behind `--isolatedModules` flag that is introduced to TS compiler's config
- provide brief explanation on why export statements need rework
- provide explanation on the strategy with `types.private` (mb `types.internal`, just to better reflect semantics?) modules - whey these are necessary and which problem they help to address

kuzhelov (approved):

perf/src/index.tsx
layershifter: It contains `regeneratorRuntime` that it's required by `async/await` transform.

packages/react/src/types.ts
layershifter: Exactly 😇

PR #1358, opened 2019-05-20
feat(popper): custom react wrapper

### LINKED ISSUE #1342: Custom PopperJS React wrapper
### Problem description
We need to build a custom React wrapper around [`popper.js`](https://github.com/FezVrasta/popper.js) library (pure JS). The one offered by `popper.js`, called `react-popper` has performance issues:

#### 1. [Performance issues on a large number of poppers #185](https://github.com/FezVrasta/react-popper/issues/185):

React Bootstrap and Material UI are using `popper.js` as well and have decided to write their own React wrappers, https://material-ui.com/utils/popper/
_(see: https://github.com/FezVrasta/react-popper/issues/185#issuecomment-403241821)_

#### 2. Popper perf issues when integrating `react-popper` in `Dropdown` component

As you can see here:
- https://github.com/stardust-ui/react/pull/1312#pullrequestreview-236511503 
- https://github.com/stardust-ui/react/pull/1312#issuecomment-491957594

There are obvious perf issues caused by:
1. recreating the `PopperJS` instance (responsible for calculating styles that need to be applied for positioning elements) every time there is a scroll/resize event: https://github.com/FezVrasta/popper.js/blob/36cc6c66e55f716232166c9a5725d8d08c4ec338/packages/popper/src/utils/setupEventListeners.js#L38.

2. recreating the `PopperJS` instance causes the `modifiers` to be run on every scroll/resize event

3. one of these default modifiers, called `updateStateModifier` causes a rerender on every scroll/resize event: https://github.com/FezVrasta/react-popper/blob/ff9dde9ca94c93be30098b28cf0361ca65a766ce/src/Popper.js#L94

### Proposed solution

Create a basic custom `Popper` component (similar to the one provided by `react-popper`) that:
- applies styles directly
- causes rerenders on small set of conditions (mostly props changes or calculated styles) but not on side effects coming from browser events
- reference: https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Popper/Popper.js

bmdalex:
#### Fixes #1342 

We need to build a custom React wrapper around popper.js library (pure JS). The one offered by popper.js, called react-popper has performance issues: [Performance issues on a large number of poppers #185](https://github.com/FezVrasta/react-popper/issues/185):

React Bootstrap and Material UI are using `popper.js` as well and have decided to write their own React wrappers, https://material-ui.com/utils/popper/
_(see: https://github.com/FezVrasta/react-popper/issues/185#issuecomment-403241821)_

### Proposed solution

Create a basic custom `Popper` component (similar to the one provided by `react-popper`) that:
- applies styles directly
- causes rerenders on small set of conditions (mostly props changes or calculated styles) but not on side effects coming from browser events
- reference: https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Popper/Popper.js


layershifter:
@Bugaa92 can you please link issue in `react-popper` repo with discussions of there problems?

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -140,7 +114,6 @@ export interface PopupProps
 
 export interface PopupState {
   open: boolean
-  target: HTMLElement
```
layershifter, line 143: 👍

packages/react/src/components/Popup/PopupContent.tsx
```diff
@@ -51,10 +52,7 @@ export interface PopupContentProps
   pointing?: boolean
 
   /** A ref to a pointer element. */
-  pointerRef?: PopperChildrenProps['arrowProps']['ref']
-
-  /** An object with positioning styles fof a pointer. */
-  pointerStyle?: PopperChildrenProps['arrowProps']['style']
+  pointerRef?: React.RefObject<Element>
```
layershifter:
```suggestion
  pointerRef?: React.Ref<Element>
```

As we accept functions there, too

packages/react/src/components/Popup/PopupContent.tsx
```diff
@@ -51,10 +52,7 @@ export interface PopupContentProps
...
 }
 
 class PopupContent extends UIComponent<WithAsProp<PopupContentProps>, any> {
```
layershifter:
```suggestion
class PopupContent extends UIComponent<WithAsProp<PopupContentProps>> {
```

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -0,0 +1,171 @@
+import * as React from 'react'
+import PopperJS from 'popper.js'
+import { Ref } from '@stardust-ui/react-component-ref'
+
+import { getPlacement, applyRtlToOffset } from './positioningHelper'
+import { Alignment, Position } from './index'
```
layershifter: Can we move types to a separate file?

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -0,0 +1,171 @@
...
+   * Position for the component. Position has higher priority than align. If position is vertical ('above' | 'below')
+   * and align is also vertical ('top' | 'bottom') or if both position and align are horizontal ('before' | 'after'
+   * and 'start' | 'end' respectively), then provided value for 'align' will be ignored and 'center' will be used instead.
+   */
+  position?: Position
+
+  /**
+   * Array of conditions to be met in order to trigger a subsequent render to reposition the elements.
+   */
+  positioningDependencies?: any[]
```
layershifter:
```suggestion
  positioningDependencies?: React.DependencyList
```

nit: We can use `DependencyList` type to comply

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -0,0 +1,171 @@
...
+  const contentRef = React.useRef<HTMLElement>(null)
+  const latestPlacement = React.useRef<PopperJS.Placement>(proposedPlacement)
+  const [computedPlacement, setComputedPlacement] = React.useState<PopperJS.Placement>(
+    proposedPlacement,
+  )
+
+  React.useEffect(
+    () => {
+      const handleUpdate = (data: PopperJS.Data) => {
+        if (data.placement !== latestPlacement.current) {
```
layershifter, line 49: Can we add a meaningful comment why we need this?

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -0,0 +1,171 @@
...
+      }
+
+      const pointerRefElement = pointerRef && pointerRef.current
+      const options: PopperJS.PopperOptions = {
+        placement: proposedPlacement,
+        eventsEnabled,
+        positionFixed,
+        modifiers: {
+          ...(offset && {
+            offset: { offset: rtl ? applyRtlToOffset(offset, position) : offset },
```
layershifter: I suggest to move out this to a root of function and call it `offsetModifier`, it will allow us to reduce dependency list

```ts
const offsetModifier = React.useMemo(() => {
  return { offset: rtl ? applyRtlToOffset(offset, position) : offset }
}, [rtl, offset, position])
```

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -0,0 +1,171 @@
...
+      return () => popperRef.current.destroy()
+    },
+    [eventsEnabled, modifiers, offset, position, positionFixed, rtl, proposedPlacement],
+  )
+
+  React.useEffect(
+    () => {
+      popperRef.current.scheduleUpdate()
+    },
+    [...(positioningDependencies || []), computedPlacement],
```
layershifter: We can `positioningDependencies: []` in props, so we can simplify this:

```suggestion
    [...positioningDependencies, computedPlacement],
```

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -0,0 +1,171 @@
...
+  )
+
+  let processedChildren = children
+  if (typeof children === 'function') {
+    processedChildren = (children as PopperChildrenFn)({ placement: computedPlacement })
+  }
```
layershifter: It seems that we need to rewrite this:

```tsx
const child = typeof children === 'function'
  ? (children as PopperChildrenFn)({ placement: computedPlacement })
  : React.Children.only(children)
```

layershifter:
We should:
- remove `react-popper` from `@stardust-ui/react` dependencies
- add `popper.js` to `@stardust-ui/react` dependencies
- add `react-popper` to root package dependencies as we use it in prototypes
- ensure that we don't use any types from `react-popper` in `lib/positioner`

packages/react/src/components/Popup/PopupContent.tsx
bmdalex: done, thanks

packages/react/src/lib/positioner/Popper.tsx
bmdalex: done, thanks

packages/react/src/components/Popup/PopupContent.tsx
bmdalex: done, thanks

packages/react/src/lib/positioner/Popper.tsx
bmdalex: done, thanks

packages/react/src/lib/positioner/Popper.tsx
bmdalex: done, thanks

packages/react/src/lib/positioner/Popper.tsx
bmdalex: done, thanks

packages/react/src/lib/positioner/Popper.tsx
bmdalex: done a bit differently, thanks

packages/react/src/lib/positioner/Popper.tsx
bmdalex: done, thanks

bmdalex:
@layershifter 
> * remove `react-popper` from `@stardust-ui/react` dependencies

agreed

> * add `popper.js` to `@stardust-ui/react` dependencies

agreed

> * add `react-popper` to root package dependencies as we use it in prototypes

shouldn't we rather avoid using `react-popper` at all?

> * ensure that we don't use any types from `react-popper` in `lib/positioner`

I checked, but how can we ensure that programmatically?

layershifter:
> shouldn't we rather avoid using react-popper at all?

We can, but we don't export `Popper` component now...

> I checked, but how can we ensure that programmatically?

If we will remove the package 😼

layershifter (approved):

bmdalex:
> > shouldn't we rather avoid using react-popper at all?
> 
> We can, but we don't export `Popper` component now...

Should we try that? I'd do it, but not sure how. Do we want it?

> > I checked, but how can we ensure that programmatically?
> 
> If we will remove the package 😼

exactly my point :)

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -0,0 +1,174 @@
...
+  /**
+   * Rtl attribute for the component.
+   */
+  rtl?: boolean
+}
+
+/**
+ * Popper relies on the 3rd party library [Popper.js](https://github.com/FezVrasta/popper.js) for positioning.
+ */
+const Popper: React.FunctionComponent<PopperProps> = props => {
```
kuzhelov, line 11: no tests for this guy?

packages/react/src/themes/teams/components/Popup/popupContentStyles.ts
```diff
@@ -1,4 +1,4 @@
-import { PopperChildrenProps } from 'react-popper'
+import { PopperChildrenProps } from '../../../../lib/positioner'
```
kuzhelov: Popup styles shouldn't depend on this, they should depend just on Popup's prop types - otherwise we are breaking encapsulation. It should be sufficient to use just PopupContentProps['placement'] at place where this type was necessary

kuzhelov:
not sure why breaking change in the `Popup's` API was necessary? I may understand that in case if it would provide only beneficial effect in return - however, this solution seems to work nicely for Function Component + Hooks world. In case if it will be class component as consumer of Popup, it would require some tricky turns from the client to establish the variable to watch for - as, in contrast, callback-based approach would be the most intuitive thing.

I would suggest to leave this change aside from the PR, as it is not necessary to solve the PR's goal

packages/react/src/lib/positioner/Popper.tsx
bmdalex: TBH I didn't really now HOW to test it. Reasons:
- `PopperJS` is setting styles directly on element so we have no control and no ability to predict anything there
- We could test returned `placement` but that's basically the result of `getPlacement` method that is already tested

Unless you want to test the options arguments we pass to `PopperJS` and mock the constructor? Open to suggestions.

bmdalex:
> not sure why breaking change in the `Popup's` API was necessary? I may understand that in case if it would provide only beneficial effect in return - however, this solution seems to work nicely for Function Component + Hooks world. In case if it will be class component as consumer of Popup, it would require some tricky turns from the client to establish the variable to watch for - as, in contrast, callback-based approach would be the most intuitive thing.

> not sure why breaking change in the `Popup's` API was necessary?

It's not really necessary, but I find the existing `renderContent` API more confusing and this seemed like a good time to try another one and see how it goes.

I believe the new API is more discoverable by users and more clear in what it's trying to accomplish. `positioningDependencies` prop makes it obvious and explicit that this is the place where users can handle the positional updates.

> this solution seems to work nicely for Function Component + Hooks world. In case if it will be class component as consumer of Popup, it would require some tricky turns from the client to establish the variable to watch for

Here's working code achieving the same scenario with function and class components. Doesn't really look like tricky stuff to me:

```typescript
import { Button, Popup, Segment, Loader, Grid, Header } from '@stardust-ui/react'

const CustomSegment = () => (
  <Segment styles={{ minHeight: '300px' }}>Hello from loaded data!</Segment>
)

const AsyncDataLoader = props => {
  const [data, setData] = React.useState(<Loader />)

  React.useEffect(() => {
    setTimeout(() => {
      setData(<CustomSegment />)
      props.onLoaded()
    }, 1000)
  }, [])

  return data
}

const PopupExampleAsyncFn = () => {
  const [loading, setLoading] = React.useState(true)

  return (
    <Popup
      onOpenChange={(e, data) => {
        if (!data.open) setLoading(true)
      }}
      trigger={<Button icon="expand" content="Click me!" />}
      positioningDependencies={[loading]}
      content={{ content: <AsyncDataLoader onLoaded={() => setLoading(false)} /> }}
    />
  )
}

class PopupExampleAsyncClass extends React.Component {
  state = { loading: true }

  render() {
    return (
      <Popup
        onOpenChange={(e, data) => {
          if (!data.open) this.setState({ loading: true })
        }}
        trigger={<Button icon="expand" content="Click me!" />}
        positioningDependencies={[this.state.loading]}
        content={{
          content: <AsyncDataLoader onLoaded={() => this.setState({ loading: false })} />,
        }}
      />
    )
  }
}

const PopupExampleAsync = () => (
  <Grid columns={2}>
    <Header as="h3" content="PopupExampleAsyncFn" />
    <PopupExampleAsyncFn />
    <Header as="h3" content="PopupExampleAsyncClass" />
    <PopupExampleAsyncClass />
  </Grid>
)

export default PopupExampleAsync
```

> in contrast, callback-based approach would be the most intuitive thing.

Hard to claim that without any user feedback. I checked some of our user's code and this API is not used at all..

Personally I had difficulties understanding how the `renderContent` approach worked and I don't like the fact that the users will need to switch from a prop to another (`content` -> `renderContent`). With current API it means just adding a prop to explicitly handle the fact that data is changed and we need to reposition content (`positioningDependencies`)

However, I don't feel too strongly about this and I am willing to hear other thoughts. I can agree to having this discussion on another PR. @layershifter , what do you think?

packages/react/src/themes/teams/components/Popup/popupContentStyles.ts
bmdalex: good point, done thanks

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -59,17 +57,15 @@ export interface PopupSlotClassNames {
 export interface PopupProps
   extends StyledComponentProps<PopupProps>,
     ChildrenComponentProps,
-    ContentComponentProps<ShorthandValue> {
+    ContentComponentProps<ShorthandValue>,
+    PositionCommonProps {
```
kuzhelov: nit: we may make it shorter and rename to `PositioningProps`

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -438,17 +409,9 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
...
-      arrowProps,
-      placement,
-      ref,
-      scheduleUpdate,
-      style: popupPlacementStylesRaw,
-    }: PopperChildrenProps,
+    { placement, scheduleUpdate }: PopperChildrenProps,
   ) => {
     const { content: propsContent, renderContent, contentRef, mountDocument, pointing } = this.props
-    const popupPlacementStyles = _.omitBy(popupPlacementStylesRaw, _.isNaN)
```
kuzhelov, line 449: this expression that filters NaN values from styles should stay as part of the logic. Please, consider to review description of the following PR for more details: https://github.com/stardust-ui/react/pull/1365

packages/react/src/components/Popup/Popup.tsx
bmdalex: done, let's try to get all comments on the initial review, this interface was there from beginning. of course I understand it might have come to your attention on the 2nd round. thanks for reviewing 👍

yarn.lock
```diff
@@ -11138,10 +11125,10 @@ pn@^1.1.0:
   resolved "https://registry.yarnpkg.com/pn/-/pn-1.1.0.tgz#e2f4cef0e219f463c179ab37463e4e1ecdccbafb"
   integrity sha512-2qHaIQr2VLRFoxe2nASzsV6ef4yOOH+Fi9FBOVH6cqeSgUnoyySPZkxzLuzd+RYOQTRpROA0ztTMqxROKSb/nA==
 
-popper.js@^1.14.4:
-  version "1.14.6"
-  resolved "https://registry.yarnpkg.com/popper.js/-/popper.js-1.14.6.tgz#ab20dd4edf9288b8b3b6531c47c361107b60b4b0"
-  integrity sha512-AGwHGQBKumlk/MDfrSOf0JHhJCImdDMcGNoqKmKkU+68GFazv3CQ6q9r7Ja1sKDZmYWTckY/uLyEznheTDycnA==
+popper.js@^1.15.0:
```
kuzhelov: just to not forget - we need to initiate checks for this updated version

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -0,0 +1,102 @@
...
+      const options: PopperJS.PopperOptions = {
+        placement: proposedPlacement,
+        eventsEnabled,
+        positionFixed,
+        modifiers: {
+          ...computedModifiers,
+          ...modifiers,
+          arrow: {
+            enabled: !!pointerRefElement,
+            element: pointerRefElement,
```
kuzhelov: nit: lets rename it to `pointerTarget` (and corresponding `pointingTargetRefElement`). Otherwise from the code it was unclear whether this implies element that is pointer _itself_ (like arrow), or its target, or something else

packages/react/src/components/Popup/Popup.tsx
bmdalex: no need for this anymore because styles will be set directly to the DOM element by Popper js. Only valid CSS will be set and it will not leak to props. Wish I saw that PR earlier as there was no need for it

yarn.lock
bmdalex: no need to anymore, we downgraded the min version to `1.14.6`

packages/react/src/lib/positioner/Popper.tsx
bmdalex: but `pointerRef` is more aligned to current `targetRef`:

```typescript
/**
 * Ref object containing the pointer node.
 */
pointerRef?: React.RefObject<Element>

/**
 * Ref object containing the target node (the element that we're using as reference for Popper box).
 */
targetRef?: React.RefObject<Element>
```

To me it seems that having `pointerTarget` and `targetRef` is more confusing

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -0,0 +1,102 @@
+import * as React from 'react'
+import PopperJS from 'popper.js'
```
kuzhelov, line 2: `PopperJS` is improperly imported here - given that namespace is imported, it should be 

```
import * as PopperJS from 'popper.js'
```

Currently the code transpiled to CommonJS will result in the following client's error:

image

bmdalex: I tried updating to:
```typescript
import * as PopperJS from 'popper.js'
```

and ended up with this error:
Screenshot 2019-05-23 at 10 12 33

If I change the constructor to `new PopperJS.default` I get a lint error that constructor cannot be lowercase and the code transpiled to CommonJS will result in the same error.

Are the types wrong? I'm blocked with this right now...
@layershifter @kuzhelov

kuzhelov: have checked `popper.js` module used in Stardust, as well as tried to repro the issue in Stardust environment - seems that everything is working fine, as well as types are properly defined for PopperJS module. Maybe this is a problem with client environment setup, where there is a stable repro for that - lets unblock this PR and see.

kuzhelov: @layershifter, what are your thoughts on that?

PR #1362, opened 2019-05-21
chore: tolerate build problems in DEV

kuzhelov:
This change is necessary to alleviate DEV scenarios.

## Problem

Currently when one is running `yarn start` it might end up with the error or warning - and quite often this warning might be result of some leftover in the code which is easy to fix. Simple example - unused type declaration:

image

The annoying thing is that, after fixing this problem, **developer will again need to wait for `yarn start` to launch docs**.

image

## Solution

This problem might be alleviated be tolerating build problems in DEV environment, so that it will be possible for developer to make the necessary fixes and immediately see the result of that. It is also possible to even skip some warning messages while making experiments 👍 

image

-------

**At the same time, all the build checks for any environment other than DEV are preserved.**
- [x] check this behavior


miroslavstastny (approved):
Wasn't this the main reason why we wanted babel?

kuzhelov:
@miroslavstastny, 

> Wasn't this the main reason why we wanted babel?

as I might expect, we will still have similar problem even with Babel, as this will come with `ForkTsChecker` plugin working in parallel with Babel - and this plugin will be responsible for making type checks. Thus, still, we will have warnings/errors produced by this plugin - and while it is an inevitable thing for PROD env - for DEV, in contrast, we are safe to tolerate them and have quick process to fix :)

PR #1363, opened 2019-05-21
fix(Dropdown): adds more info to a11y status message in multiple examples

### LINKED ISSUE #1005: Acc > improve user experience in Dropdown
### Problem description
When user navigate into the "Multiple Selection" dropdown, then user is not aware about possibility of navigate to selected users by right and left arrows. 

### Proposed solution
wrote by @silviuavram 
we should create / edit an example in Dropdowns with another overloaded getA11yMessage
We should at least have an example that will overload the standard Downshift message and say Press up and down to navigate, press Enter to select, press left and right to navigate through the already selected items.

### MVP

silviuaavram:
Fixes https://github.com/stardust-ui/react/issues/1005


layershifter:
Can we create a separate example that describes how to to use `getA11yStatusMessage()` in single/multiple cases instead?

silviuaavram:
> Can we create a separate example that describes how to to use `getA11yStatusMessage()` in single/multiple cases instead?

Well, it's the same use, and we already have the example with different language that shows how to use both to change the narration. Why create another example just to emphasise the single/multiple use of this prop?

docs/src/examples/components/Dropdown/Types/DropdownExampleMultiple.shorthand.tsx
```diff
@@ -28,4 +29,25 @@ const getA11ySelectionMessage = {
   onRemove: item => `${item} has been removed.`,
 }
 
+const getA11yStatusMessage = ({
```
bmdalex, line 32: is this the right thing to do?

I'd suggest encapsulate this logic as an accessibility behavior /  internal implementation for `Dropdown` that will just receive the relevant strings / methods as props:

```typescript
const getStatusMessageFromCount = (resultCount: number) =>
  `${resultCount} result${ resultCount === 1 ? ' is' : 's are'} available, use up and down
arrow keys to navigate. Press Enter key to select. Press left and right arrow keys to
navigate through the selected options.` 

getA11yStatusMessage={{
  default: getStatusMessageFromCount,
  noResult: 'No results are available.',
}}

bmdalex: same comment to the other instances

silviuaavram: I don't think we want to do that. It's a message we don't want to save in our code due to localisation considerations.

PR #1364, opened 2019-05-21
chore(dialogBehavior): add more tests

silviuaavram:
Added tabIndex tests for dialogBehavior and popupBehavior.


packages/react/test/specs/behaviors/dialogBehavior-test.tsx
```diff
@@ -0,0 +1,20 @@
+import { dialogBehavior } from 'src/lib/accessibility'
+import Button from 'src/components/Button/Button'
```
bmdalex: nit: no need to take a dependency on stardust for this test, you can use `<button />`

bmdalex (approved):
👍

PR #1368, opened 2019-05-22
docs(Accessibility): improve introduction section

jurokapsiar:
This PR improves introduction section to accessibility, mainly summarizing what needs to be done in preparation for correct component usage to achieve accessible experiences

image


docs/src/views/Accessibility.tsx
```diff
@@ -56,31 +56,70 @@ export default () => (
...
-      {link('ARIA best practises', 'https://www.w3.org/TR/wai-aria-practices-1.1/')}
-      and validates them with subject matter experts.
+      correct ARIA roles, testing on multiple screen reader / os combinations. This allows spending
+      more time on the usability aspects of accessibility.
+    </p>
+    <p>
+      Following steps help to design an accessible user experience:
+      <ol>
+        <li>
+          decompose UI to parts and <b>identify components, variants and behaviors to use</b>
```
kolaps33, line 78: do we want to mention "behaviors to use" ?

I mean if we have already default behaviors. What do you think about something like this?: 
"Decompose UI to parts and <b>identify components, variants. Default behavior will apply for component. "

docs/src/views/Accessibility.tsx
```diff
@@ -56,31 +56,70 @@ export default () => (
...
+            {link('color and contrast', 'https://accessibility.umn.edu/core-skills/color-contrast')}
+          </b>{' '}
+          to convey information
+        </li>
+        <li>
+          define <b>tab order and arrow key navigation</b>
+        </li>
+        <li>
+          specify <b>labels</b>, especially for components without textual information (icon only
+          buttons)
```
kolaps33: Maybe I would add if you agree something like.

"Label the menu, list, toolbars component, to give user more context. "

docs/src/views/Accessibility.tsx
```diff
@@ -56,31 +56,70 @@ export default () => (
...
+        <li>
+          identify UI parts that appear on <b>hover or focus</b> and specify keyboard and screen
+          reader interaction with them
+        </li>
+        <li>
+          list cases when <b>focus</b> needs to be <b>moved programatically</b> (if parts of the UI
+          are appearing/disappearing or other cases){' '}
+        </li>
+        <li>
+          list cases when <b>focus</b> needs to be <b>trapped</b> in sections of the UI
```
kolaps33: if we want to add note, that focus is usually trapped in dialog(s) ?

docs/src/views/Accessibility.tsx
jurokapsiar: for example for menu or popup there are multiple behaviors - they will need to choose which one to use

CHANGELOG.md
```diff
@@ -16,6 +16,10 @@
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Documentation
+- Accessibility: improve introduction section @jurokapsiar ([#1368](https://github.com/stardust-ui/react/pull/1368))
```
miroslavstastny, line 30: Please follow the order of changelog sectionsas described in INSTRUCTIONS above.

miroslavstastny (approved):

sophieH29 (approved):

PR #1370, opened 2019-05-22
 chore(docs): make HMR working properly

layershifter:
Related #508.

This PR updates:
- usages of `react-hot-loader` to use their moden `hot()` API
- updates `react-source-render`, I rewrote it to omit React Context (use only render props) and be compatible with HMR (`unstable_hot` prop)

---

### Works properly
- reload of components everywhere
- reload of example sources in maximized view

### Still broken
- example's source reload on component's pages

image

- warning about missing `Provider`
image

Occurred because there is an issue with lost context: https://github.com/gaearon/react-hot-loader/issues/1207


docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -537,39 +507,53 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
...
-              <Segment styles={{ padding: 0 }}>
-                {this.renderSourceCode()}
-                {this.renderError()}
-                {this.renderHTML()}
-                {this.renderVariables()}
-              </Segment>
-              <div style={{ paddingBottom: '10px' }} />
+              {({ element, error, markup }) => (
+                <>
+                  <Provider.Consumer
```
layershifter, line 518: We don't need `Provider.Consumer` more there

docs/src/index.tsx
```diff
@@ -3,46 +3,12 @@ import * as ReactDOM from 'react-dom'
...
-      try {
-        render(require('./app').default)
-        document.scrollingElement!.scrollTop = scrollTop
-        document.scrollingElement!.scrollLeft = scrollLeft
-      } catch (e) {
-        console.error(e)
-      }
-    })
-  }
-}
```
layershifter, line 42: This stuff will be done by `react-hot-loader`

docs/src/routes.tsx
```diff
@@ -21,6 +21,18 @@ import AccessibilityBehaviors from './views/AccessibilityBehaviors'
...
+import AsyncShorthandPrototype from './prototypes/AsyncShorthand'
+import EmployeeCardPrototype from './prototypes/employeeCard'
+import MeetingOptionsPrototype from './prototypes/meetingOptions'
+import SearchPagePrototype from './prototypes/SearchPage'
+import MentionsPrototype from './prototypes/mentions'
+import DropdownsPrototype from './prototypes/dropdowns'
+import PopupsPrototype from './prototypes/popups'
+import IconViewerPrototype from './prototypes/IconViewer'
+import MenuButtonPrototype from './prototypes/MenuButton'
+import AlertsPrototype from './prototypes/alerts'
```
layershifter, line 35: We're using Webpack 4, these imports will be tree-shaken in production

This change is **required**: to prevent errors related to HMR like this:
image

docs/src/utils/exampleContexts.ts
```diff
@@ -15,8 +15,25 @@ export const examplePlaygroundContext = require.context(
...
+// ----------------------------------------
+
+if (__DEV__) {
+  // When the application source code changes, re-render the whole thing.
+  if (module.hot) {
+    // We need this to catch cases unhandled by RHL
+    // https://github.com/webpack/webpack/issues/834#issuecomment-76590576
+    module.hot.accept(exampleSourcesContext.id, () => {
+      exampleSourcesContext = require.context('docs/src/exampleSources/', true, /.source.json$/)
+    })
```
layershifter, line 37: Context should be updated separately.

bmdalex (approved):
👍 should we organize a session to test the docs to be sure we didn't break something?

layershifter:
@Bugaa92 at least it doesn't make anything worst 😸 It will affect only local development...

I will merge it and will be able to detect and fix more cases, so I am opened for feedback when more people will be able to perform live test

PR #1371, opened 2019-05-22
docs(Accessibility): improve acc descriptions for components

jurokapsiar:
Improve acc descriptions for components


packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -275,5 +275,7 @@ Dialog.slotClassNames = {
 
 /**
  * A Dialog indicates a possible user action.
```
hughreeling: I don't think that's good description of dialog.  We can look at other frameworks for better description, but perhaps something like, a dialog is used to interrupt the standard flow to get some particular input from the user.

jurokapsiar: it is not - I have missed this one :)

package.json
```diff
@@ -112,6 +112,7 @@
     "react-dom": "^16.8.6",
     "react-element-to-jsx-string": "^14.0.2",
     "react-hot-loader": "^4.8.2",
+    "react-markdown": "^4.0.8",
```
bmdalex, line 115: @layershifter @kuzhelov @miroslavstastny 
- are we ok with adding this dependency?
- can we suggest some existing tooling to achieve this?

bmdalex (commented):
pls check the new dependency with the people I mentioned

package.json
jurokapsiar: @layershifter added this

layershifter: > can we suggest some existing tooling to achieve this?

Nope

> are we ok with adding this dependency?

Yep, it's only for docs

bmdalex (approved):

PR #1372, opened 2019-05-22
feat(Dropdown): add label prop

silviuaavram:
Adds the `label` prop to a Dropdown. Label content is used to label the search input, if a search, or the trigger button, if not a search.

Uses getLabelProps from Downshift that adds to it the same ID as the values used in other parts of the Dropdown, via `getMenuProps`,` getInputProps` etc.

Changes include:
- label prop in the Dropdown component and changes required for its rendering.
- unit tests.
- example in the docsite.


packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -506,17 +526,18 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
-
+    const labelProps = getLabelProps()
     const triggerButtonProps = getToggleButtonProps({
       onFocus: this.handleTriggerButtonOrListFocus,
       onBlur: this.handleTriggerButtonBlur,
       onKeyDown: e => {
         this.handleTriggerButtonKeyDown(e, rtl)
       },
-      'aria-label': content,
+      'aria-label': undefined,
```
silviuaavram, line 541: no need for label here. it will narrate either the label text or if missing the placeholder/value in the button content.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -720,6 +742,18 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
   }
 
+  private renderLabel(getLabelProps: (options?: GetLabelPropsOptions) => any) {
+    const { label, search } = this.props
+    return Text.create(label, {
+      defaultProps: {
+        as: 'label',
+        className: Dropdown.slotClassNames.label,
+        ...getLabelProps(),
+        ...(!search && { htmlFor: undefined }),
```
silviuaavram: Downshift adds the `for` as `input` id. Will remove this if there isn't any input. Also this is not really needed, as both input and trigger button have labelledby pointing to this.

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -1459,4 +1461,22 @@ describe('Dropdown', () => {
...
     })
   })
+
+  describe('label', () => {
+    it('is rendered with the text passed as prop', () => {
+      const labelText = 'Add People'
+      const wrapper = mountWithProvider(<Dropdown items={items} label={labelText} />)
+      const label = getLabelWrapper(wrapper)
+
+      expect(label.exists()).toBe(true)
```
bmdalex, line 1471: hmm why is `findIntrinsicElement` needed? maybe I'm missing something. Can you try
```typescript
const label = wrapper.find(`.${Dropdown.slotClassNames.label}`)
expect(label.exists()).toBe(true)
```

bmdalex (dismissed):
TODO

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -720,6 +742,18 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
     )
   }
 
+  private renderLabel(getLabelProps: (options?: GetLabelPropsOptions) => any) {
```
bmdalex: can we move this in a separate file?

this one is getting very big; let's try be proactive about the upcoming refactoring (splitting logic to separate files)

see also: https://github.com/stardust-ui/react/pull/1372#discussion_r287319327

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: can you please rephrase this a bit and put as comment in code? looks a bit strange that we need to do it; I'd also encapsulate this in a dedicated method together with spreading `...getLabelProps(),`

see: https://github.com/stardust-ui/react/pull/1372#discussion_r287319327

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: not sure I understand.. are we overriding existing `aria-label` or why are we setting it to undefined?

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -529,6 +550,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
             content,
             fluid: true,
             styles: styles.triggerButton,
+            ...(label && { 'aria-labelledby': labelProps['id'] }),
```
bmdalex, line 554: I guess it'd be difficult to separate this to an accessibility behavior at this point, but can we compute the `labelProps` (using Downshift function) in separate method (and preferably file) and spread the result here; 

the point I'm trying to make is that we should decouple these concerns as much as we can. the area where we apply props should work as an aggregator:

```typescript
{Button.create(triggerButton, {
  defaultProps: {
    ...stardustProps,
    ...accessibilityProps, // we should not care it's coming from Downshift 
                           // so that we can easily refactor if needed
    ...restProps,
  },
  // ...
})

File: packages/react/src/components/Dropdown/Dropdown.tsx (hunk 1 above)
Comment (silviuaavram) [reply packages/react/src/components/Dropdown/Dropdown.tsx:541]: We are overriding the existing one. The existing one comes from Downshift and we don't want that one. We were overriding it before with the content, but having an `aria-label` on a button with content is enough for screen readers to read it.

File: packages/react/test/specs/components/Dropdown/Dropdown-test.tsx (hunk 3 above)
Comment (silviuaavram) [reply packages/react/test/specs/components/Dropdown/Dropdown-test.tsx:1471]: No, it does not. That selector finds the component as well.

File: packages/react/src/components/Dropdown/Dropdown.tsx (hunk 5 above)
Comment (silviuaavram) [reply packages/react/src/components/Dropdown/Dropdown.tsx:554]: added them to a separate method.

**Review by layershifter — APPROVED** [review]
---

**Review by layershifter — CHANGES_REQUESTED** [review]
I am not fully understand why `label` should be part of `Dropdown`. I expect that `label` is a prop of `Form.Field` and `Dropdown` can be used there as `control`:

```jsx
<Form
    fields={[
      {
        label: 'Heroes',
        control: {
          as: Dropdown,
        },
        items: ['Bruce Wayne', 'Natasha Romanoff', 'Steven Strange', 'Selina Kyle'],
        key: 'heroes',
      },
    ]}
  />
```

How it can be composed with these changes? 🤔 

In Material UI it's composed in the similar way:
```jsx
 <FormControl className={classes.formControl}>
  <InputLabel htmlFor="age-helper">Age</InputLabel>
  <Select>
    <MenuItem value={10}>Ten</MenuItem>
  </Select>
</FormControl>
```

PR #1375, opened 2019-05-23
chore(Accordion): generate default ids only once

### LINKED ISSUE #1369: Create better element ID generator
## Bug Report
Replace the use of `_.uniqueId` with a better implementation, as these are re-generated at every render, and is confusing.

### Steps
Replace the lodash function in `dialogBehavior.ts` and `Accordion.tsx` with a better implementation, that will generate a unique ID but only once.

silviuaavram:
Fixes https://github.com/stardust-ui/react/issues/1369


miroslavstastny (approved):

miroslavstastny (dismissed):
Sorry, this is actually broken - you have multiple panels, you need unique ids per panel.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -137,6 +137,8 @@ class Accordion extends AutoControlledComponent<WithAsProp<AccordionProps>, Acco
 
   private focusHandler: ContainerFocusHandler = null
   private itemRefs = []
+  private defaultAccordionTitleId = _.uniqueId('accordion-title-')
```
bmdalex, line 140: nit `private readonly`, although I'm aware how @layershifter feels about modifiers 😆

bmdalex (approved):
👍

packages/react/src/components/Accordion/Accordion.tsx
miroslavstastny: Now imagine you have 10 accordions on a page each having 10 sections :-)

jurokapsiar (approved):

packages/react/src/components/Accordion/Accordion.tsx
layershifter: Visibility modifiers will be removed by lint rule on next week, no more discussions and comments around them

PR #1377, opened 2019-05-23
feat(Provider): add option for disabling animations

### LINKED ISSUE #1273: Allow disabling all animations (Animation and Loader components)
### Problem description

For testing purposes it should be possible to disable animations so that the screenshots would be stable

### Proposed solution
Add a prop to the provider (or theme) to disable animations.

mnajdova:
Changes introduced in the API with this PR in order to fix https://github.com/stardust-ui/react/issues/1273:
- The `theme` object does not contain the `rtl` and `renderer` props anymore. These props, together with the new added `disableAnimations` prop are hoisted at the top level of the `Provider`'s props API: 

Prev:
```jsx
<Provider theme={{siteVariables: {...}, rtl: true}}>
  <Avatar ... />
</Provider>
```
Now:
```jsx
<Provider theme={{siteVariables: {...}}} rtl>
  <Avatar ... />
</Provider>
```
- `ComponentStyleFunctionParam` interface is extended with the `rtl` prop, which is not defined in the `theme` anymore, as well as with the new `disableAnimations` prop.
- Example of how the disabling of the animations works
```jsx
import React from 'react'
import { Animation, Icon, Provider, Loader } from '@stardust-ui/react'

// simple spinner animation
const spinner = {
  keyframe: {
    from: {
      transform: 'rotate(0deg)',
    },
    to: {
      transform: 'rotate(360deg)',
    },
  },
  duration: '5s',
  iterationCount: 'infinite',
}

const AnimationExample = () => (<>
  {/* This tree should not have animations */}
  <Provider
    theme={{ animations: { spinner }}}
    disableAnimations
  >
    {/* Animation provided trough the Animation component */}
    <Animation name="spinner">
      <Icon name="umbrella" circular bordered />
    </Animation>
    {/* Animation provided trough the animation prop */}
    <Icon name='book' animation='spinner' />
    {/* Animation provided trough the styles of the component */}
    <Loader />

    {/* This subtree should have animations. */}
    <Provider disableAnimations={false}>
      <Animation name="spinner">
        <Icon name="umbrella" circular bordered />
      </Animation>
      <Icon name='book' animation='spinner' />
      <Loader />
    </Provider>
  </Provider>
</>)

export default AnimationExample
```

TODO
- [x] add tests


packages/react/src/components/Provider/ProviderConsumer.tsx
```diff
@@ -2,15 +2,15 @@ import * as PropTypes from 'prop-types'
...
+import { ProviderContextPrepared } from '../../themes/types'
 
 export interface ProviderConsumerProps {
   /**
    * Uses the function children pattern to access theme.
    * @param {object} theme
    * @param {object} theme.siteVariables - The siteVariables passed from the nearest Provider.
    */
-  render: (theme: ThemePrepared) => React.ReactNode
+  render: (context: ProviderContextPrepared) => React.ReactNode
```
kuzhelov: would suggest to leave the `render` prop's function signature the same: essentially, we would like client to use only the `theme` attributes here. This will allow us to avoid lots of related changes that had resulted in decomposition of even more deeply nested objects:

```jsx
 <ProviderConsumer
      render={({
        theme: {
          siteVariables: { colors, emphasisColors, naturalColors, transparentColors },
        },
      }) => (
```

mnajdova: Do you suggest that we should have the rtl, renderer and disableAnimations props here as part of the team object?

kuzhelov: no, the structure of `context` object is good now 👍 It is just about changing the argument provided for this particular `render` method of `ProviderConsumer`

packages/react/src/themes/types.ts
```diff
@@ -285,14 +288,19 @@ export interface ThemeInput {
   siteVariables?: SiteVariablesInput
   componentVariables?: ThemeComponentVariablesInput
   componentStyles?: ThemeComponentStylesInput
-  rtl?: boolean
-  renderer?: Renderer
```
mnajdova: Should we add `disableAnimations` here?

packages/react/test/specs/lib/createAnimationStyles-test.ts
```diff
@@ -1,12 +1,11 @@
 import { createAnimationStyles, felaRenderer } from 'src/lib'
+import { ProviderContextPrepared } from 'src/themes/types'
 
 const theme = {
   siteVariables: { fontSizes: {} },
   componentVariables: {},
   componentStyles: {},
   icons: {},
-  rtl: false,
-  renderer: felaRenderer,
```
mnajdova: TODO: add `disableAnimations` here

packages/react/src/components/Provider/ProviderConsumer.tsx
mnajdova: It's resolved. Please take a look

packages/react/test/specs/lib/mergeThemes/mergeThemes-test.ts
```diff
@@ -307,48 +306,6 @@ describe('mergeThemes', () => {
...
-    test('is NOT set if no boolean was provided', () => {
-      expect(mergeThemes({ rtl: null }, { rtl: null })).not.toHaveProperty('rtl')
-      expect(mergeThemes({ rtl: null }, { rtl: undefined })).not.toHaveProperty('rtl')
-
-      expect(mergeThemes({ rtl: undefined }, { rtl: null })).not.toHaveProperty('rtl')
-      expect(mergeThemes({ rtl: undefined }, { rtl: undefined })).not.toHaveProperty('rtl')
-    })
-  })
-
-  describe('renderer', () => {
```
mnajdova, line 341: TODO: add renderer tests in the mergeContexts-test.ts

packages/react/src/components/Provider/Provider.tsx
```diff
@@ -3,33 +3,40 @@ import { render } from 'fela-dom'
...
+  ProviderContextPrepared,
+  Renderer,
+  ThemeInput,
 } from '../../themes/types'
 
 import ProviderConsumer from './ProviderConsumer'
 import { mergeSiteVariables } from '../../lib/mergeThemes'
 import ProviderBox from './ProviderBox'
 import { WithAsProp } from '../../types'
+import mergeContexts from '../../lib/mergeContexts'
```
kuzhelov: may we call the file as `mergeProviderContexts`, so it will be clear what this is about by just looking at its name? Otherwise it may be not obvious from the first look. Something like:

```js
// not necessary to change name of the imported function, 
// as `Provider`'s file context (pan not intended:) is enough here
import mergeContexts from '../../lib/mergeProviderContexts'
```

packages/react/src/components/Provider/Provider.tsx
```diff
@@ -146,41 +160,49 @@ class Provider extends React.Component<WithAsProp<ProviderProps>> {
...
+      as,
+      theme,
+      rtl,
+      disableAnimations,
+      renderer,
+      variables,
+      children,
+      ...unhandledProps
+    } = this.props
+    const context: ProviderContextInput = {
```
kuzhelov: nit: just for the sake of better explicitness, may suggest to rename to `inputContext` or `contextFromProps`

packages/react/src/components/Provider/Provider.tsx
```diff
@@ -146,41 +160,49 @@ class Provider extends React.Component<WithAsProp<ProviderProps>> {
...
+      renderer,
+    }
     // rehydration disabled to avoid leaking styles between renderers
     // https://github.com/rofrischmann/fela/blob/master/docs/api/fela-dom/rehydrate.md
+    const outgoingContext: ProviderContextPrepared = mergeContexts(this.context, context)
+
+    // Heads up!
+    // We should call render() to ensure that a subscription for DOM updates was created
+    // https://github.com/stardust-ui/react/issues/581
+    if (isBrowser()) render(outgoingContext.renderer)
```
kuzhelov: lets import this function as something like `felaDomRender`, just to disambiguate from lots of other renders defined here

packages/react/src/lib/felaDisableAnimationsPlugin.ts
```diff
@@ -0,0 +1,26 @@
+import * as _ from 'lodash'
+
+export default () => {
+  const disableAnimations = (styles: Object, type?, renderer?, props?) => {
```
kuzhelov, line 22: note, infinite recursion is possible in case if object has links to self in the props tree. At the very least, lets mention this limitation in the comments - it shouldn't be a problem of correct styles object, but given that this object is something that _client provides us with, it may be an accidental mistake in styles that would result in whole app hanging

packages/react/src/lib/felaDisableAnimationsPlugin.ts
```diff
@@ -0,0 +1,26 @@
+import * as _ from 'lodash'
+
+export default () => {
```
kuzhelov, line 21: lets document behavior of this plugin in comment - specifically, the following facts
- it is triggered by the styles' prop value (in contrast to many other Fela plugins)
- what is the name of the prop value

packages/react/src/lib/renderComponent.tsx
```diff
@@ -16,6 +16,8 @@ import {
   PropsWithVarsAndStyles,
   State,
   ThemePrepared,
+  ProviderContextPrepared,
```
kuzhelov: would expect these two types to be declared in `../types`, not in theme types, as those are not dependent on Stardust's theme semantics

packages/react/src/lib/renderComponent.tsx
```diff
@@ -40,6 +42,8 @@ export interface RenderResultConfig<P> {
   styles: ComponentSlotStylesPrepared
   accessibility: AccessibilityBehavior
   rtl: boolean
+  renderer: Renderer
```
kuzhelov: actually, we shouldn't pass `renderer` to component's `render` method - could you, please, suggest where this is necessary thing to do (sorry if I am missing something)? Also, the same most probably true for `disableAnimations` flag - we need to pass only those bits that are necessary for the component to finish rendering: classes (styles), ElementType, unhandledProps ...

packages/react/src/themes/teams/components/Animation/animationStyles.ts
```diff
@@ -1,5 +1,23 @@
...
+      duration: props.duration,
+      delay: props.delay,
+      iterationCount: props.iterationCount,
+      direction: props.direction,
+      fillMode: props.fillMode,
+      playState: props.playState,
+      timingFunction: props.timingFunction,
+    }
+
+    return createAnimationStyles(animation, context as ProviderContextPrepared)
```
kuzhelov: lets discuss this - it seems that we could simplify this a bit and avoid leaking `ProviderContextPrepared` to styles API _for all components_ - this is, essentially, the thing I would like to prevent. Essentially, we would like to pass (for now) only theme object, with nothing else being leaked

packages/react/src/lib/felaDisableAnimationsPlugin.ts
mnajdova: Added, please take a look.

packages/react/src/lib/felaDisableAnimationsPlugin.ts
mnajdova: Added, please take a look. We should add this for the other plugins that potentially can cause infinite recursion.

packages/react/src/lib/renderComponent.tsx
mnajdova: For the `renderer`, currently we are using it to render the keyframes in the animations. If we use the https://github.com/rofrischmann/fela/tree/master/packages/fela-plugin-embedded - @layershifter has confirmed that the animations are working as expected in rtl now with this plugin, then we can skip the `renderer` as well. For the `disableAnimations`, not sure if we should skip it, as for some components, if the animation are disabled they may want to show different slots, or style the component differently (imagine Loader with `disaledAnimations` - they may want to provide some circle etc)

packages/react/src/themes/base/components/Loader/loaderStyles.ts
```diff
@@ -27,8 +27,8 @@ export default {
     theme: t,
     variables: v,
   }: ComponentStyleFunctionParam<LoaderProps, LoaderVariables>): ICSSInJSStyle => {
-    const animationName = t.renderer.renderKeyframe(
-      () =>
+    const animationName = {
```
kuzhelov, line 30: 👍 for obtained declarativeness

packages/react/src/themes/teams/components/Loader/loaderStyles.ts
```diff
@@ -48,15 +47,18 @@ export default {
...
     }
     const svgAnimation: ICSSInJSStyle = {
-      animationName: t.renderer.renderKeyframe(
-        () =>
+      animationName: {
+        keyframe: ({ from, to }) =>
           ({
             to: {
-              transform: `translate3d(0, ${v.svgTranslatePosition[p.size]}, 0)`,
+              transform: `translate3d(, ${v.svgTranslatePosition[p.size]}, 0)`,
```
kuzhelov: just to be sure - is `0` removed intentionally?

packages/react/src/themes/teams/components/Loader/loaderStyles.ts
```diff
@@ -48,15 +47,18 @@ export default {
       position: 'relative',
     }
     const svgAnimation: ICSSInJSStyle = {
-      animationName: t.renderer.renderKeyframe(
-        () =>
+      animationName: {
+        keyframe: ({ from, to }) =>
```
kuzhelov: where `from` parameter is used?

packages/react/src/types.ts
```diff
@@ -3,6 +3,7 @@
 // ========================================================
 
 import * as React from 'react'
+import { ThemeInput, Renderer, ThemePrepared } from './themes/types'
```
kuzhelov, line 6: 👍

packages/react/test/specs/components/Provider/Provider-test.tsx
```diff
@@ -57,7 +57,7 @@ describe('Provider', () => {
   describe('RTL', () => {
     test('Sets dir="rtl" on the div for RTL theme', () => {
       const component = mount(
-        <Provider id="top-level-provider" theme={{ rtl: true }}>
+        <Provider id="top-level-provider" rtl={true}>
```
kuzhelov: we may simplify these expressions and pass just `rtl`:

```jsx
 <Provider id="top-level-provider" rtl />
```

packages/react/src/themes/teams/components/Loader/loaderStyles.ts
mnajdova: No, I reverted this change.

packages/react/src/themes/teams/components/Loader/loaderStyles.ts
mnajdova: Now they are used :)

packages/react/src/components/ItemLayout/ItemLayout.tsx
```diff
@@ -49,19 +49,19 @@ export interface ItemLayoutProps extends UIComponentProps, ContentComponentProps
     classes: ComponentSlotClasses,
   ) => React.ReactNode
   /** Styled applied to the root element of the rendered component. */
-  rootCSS?: ICSSInJSStyle
+  rootCSS?: React.CSSProperties
```
mnajdova, line 52: As these are applied on div elements, we need the React.CSSProperties as a type

packages/react/src/themes/types.ts
```diff
@@ -209,7 +209,18 @@ export interface ICSSPseudoElementStyle extends ICSSInJSStyle {
...
+  params?: object
+}
+
+export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
+
+export type CSSProperties = Omit<React.CSSProperties, 'animationName'> & {
+  animationName?: StardustAnimationName | string | 'none'
+}
+
+export interface ICSSInJSStyle extends CSSProperties {
```
mnajdova, line 223: This makes the ICSSInJSStyle not being able to accept something typed as React.CSSProperties. Not sure if this will be a problem, but if we don't add this, we will have to introduce new names for the animations defined as stardust animations (as `{keyframe: function, params: {...}}`)

packages/react/test/specs/lib/felaRenderKeyframesPlugin-test.ts
```diff
@@ -0,0 +1,59 @@
...
+      animationName: {
+        keyframe: () => ({ from: { rotate: '0deg' }, to: { rotate: '360deg' } }),
+        params,
+      },
+      animationDuration: '2s',
+    }
+
+    const renderer = { renderKeyframe: jest.fn() }
+    renderKeyframesPlugin(style, 'RULE', renderer)
+    expect(renderer.renderKeyframe).toHaveBeenCalledWith(expect.any(Function), params)
```
mnajdova, line 57: Not sure how to check here that the keyframe was actually the first param, as the animationName is replaced with string after the plugin is invoked, so I cannot use the same reference. I tried extracting the keyframe function in a const, but then again jest reports it as a different function.

docs/src/components/ExternalExampleLayout.tsx
```diff
@@ -80,10 +84,7 @@ class ExternalExampleLayout extends React.Component<
 
   private getTheme = (): ThemeInput => {
```
layershifter: nit: It looks we can remove this method at all

packages/react/src/lib/felaDisableAnimationsPlugin.ts
```diff
@@ -0,0 +1,34 @@
...
+        const cssPropertyValue = styles[cssPropertyName]
+
+        if (typeof cssPropertyValue === 'object') {
+          return {
+            ...acc,
+            [cssPropertyName]: disableAnimations(cssPropertyValue, type, renderer, props),
+          }
+        }
+
+        if (_.startsWith(cssPropertyName, 'animation')) {
```
layershifter: Let's go with an array of CSS props:
- it's more obvious which props will be really affected
- it's much faster in FF: https://jsperf.com/startswith-vs-array-vs-polyfill
- we will omit Lodash there

packages/react/src/lib/felaRenderKeyframesPlugin.ts
```diff
@@ -0,0 +1,50 @@
+import callable from './callable'
+
+/**
+ * Fela plugin for disabling animations. The animations are disabled or not based on the
+ * props' disableAnimations param. If the value of the prop is true, all animation related
+ * styles are removed.
```
layershifter: Description is invalid...

packages/react/src/lib/felaDisableAnimationsPlugin.ts
```diff
@@ -0,0 +1,34 @@
...
+ * Fela plugin for disabling animations. The animations are disabled or not based on the
+ * props' disableAnimations param. If the value of the prop is true, all animation related
+ * styles are removed.
+ *
+ * Caution! Infinite recursion is possible in case if style object has links to self in the props
+ * tree.
+ */
+export default () => {
+  const disableAnimations = (styles: Object, type?, renderer?, props?) => {
+    if (props && props.disableAnimations && type && type === 'RULE') {
```
layershifter: `type && type === 'RULE'` is too strict, `type === 'RULE'` is enough

```suggestion
    if (props && props.disableAnimations && type === 'RULE') {
```

packages/react/src/lib/renderComponent.tsx
```diff
@@ -168,7 +170,9 @@ const renderComponent = <P extends {}>(
     props.variables,
   )(siteVariables)
 
-  const animationCSSProp = props.animation ? createAnimationStyles(props.animation, theme) : {}
+  const animationCSSProp = props.animation
+    ? createAnimationStyles(props.animation, context.theme)
```
layershifter, line 168:
```suggestion
    ? createAnimationStyles(props.animation, theme)
```

 nit: We already destcructed it :)

packages/react/src/themes/teams/components/Animation/animationStyles.ts
```diff
@@ -1,5 +1,23 @@
+import { AnimationProp } from '../../../types'
+import createAnimationStyles from '../../../../lib/createAnimationStyles'
+
 export default {
   root: () => ({
     display: 'inline-block',
   }),
+  children: ({ props, variables, theme }) => {
+    const animation: AnimationProp = {
+      name: props.name,
```
layershifter:
```suggestion
      name: p.name,
```

nit

packages/react/src/themes/teams/components/Animation/animationStyles.ts
```diff
@@ -1,5 +1,23 @@
+import { AnimationProp } from '../../../types'
+import createAnimationStyles from '../../../../lib/createAnimationStyles'
+
 export default {
   root: () => ({
     display: 'inline-block',
   }),
+  children: ({ props, variables, theme }) => {
```
layershifter: It's strange to have `children` slot, but I don't have better idea

packages/react/src/themes/types.ts
```diff
@@ -209,7 +209,18 @@ export interface ICSSPseudoElementStyle extends ICSSInJSStyle {
   content?: string
 }
 
-export interface ICSSInJSStyle extends React.CSSProperties {
+export interface StardustAnimationName {
+  keyframe?: any
+  params?: object
+}
+
+export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
```
layershifter:
```suggestion
type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
```

This type is never used expect this file

packages/react/test/specs/lib/felaDisableAnimationsPlugin-test.ts
```diff
@@ -0,0 +1,60 @@
...
+      disableAnimationsPlugin(style, 'RULE', undefined, { disableAnimations: false }),
+    ).toMatchObject(style)
+  })
+
+  test('disables animations if the disableAnimations flag is true', () => {
+    const style = {
+      animationName: 'k1',
+      animationDuration: '1s',
+      margin: '0px 10px',
+    }
```
layershifter: nit: We have the same object defined 5 times, can we move it out of `describe()` scope?

packages/react/test/specs/lib/felaRenderKeyframesPlugin-test.ts
```diff
@@ -0,0 +1,59 @@
...
+      animationDuration: '2s',
+    }
+
+    expect(renderKeyframesPlugin(style, 'RULE', felaRenderer)).toMatchObject(style)
+  })
+
+  test('transforms the animationName prop if it contains keyframe in the definition', () => {
+    const style = {
+      animationName: {
+        keyframe: () => ({ from: { rotate: '0deg' }, to: { rotate: '360deg' } }),
```
layershifter, line 19: I am confused with this stuff. Do we really want to implement something that is different from Fela APIs?

https://github.com/rofrischmann/fela/blob/master/packages/fela-plugin-embedded/src/index.js#L45

mnajdova: I'm open to suggestions, we need to still use the keyframe params object, we don't want to restrict the API with this changes.

packages/react/src/lib/renderComponent.tsx
mnajdova: The destructuring was not correct, not sure how to properly destructure it with the correct defaults. Will use only context.theme in the file.

packages/react/test/specs/lib/felaDisableAnimationsPlugin-test.ts
mnajdova: Will restructure it a bit. We need few scenarios to test, but will make sure those are reused across the tests.

packages/react/test/specs/lib/felaRenderKeyframesPlugin-test.ts
layershifter: Let's keep existing for this PR, but we definitely should move to `fela-plugin-embedded` and align APIs.

packages/react/src/lib/renderComponent.tsx
layershifter: Got your point, let's try to refactor it somehow. Now it's super confusing:

image

mnajdova: I am using the context.theme everywhere now, removed the theme const. Please take another look.

packages/react/test/specs/lib/felaDisableAnimationsPlugin-test.ts
mnajdova: Extracted two consts before the tests, which are used across all tests.

packages/react/test/specs/lib/felaRenderKeyframesPlugin-test.ts
mnajdova: We can adjust the plugin to convert the Stardust's animation API to the one required by the fela embedded plugin and preserve our functionality as well as use the embedded plugin.

packages/react/src/themes/teams/components/Animation/animationStyles.ts
mnajdova: I was struggling a bit with this one as well, but this is the most correct..

layershifter (approved):

layershifter:
Can we add a new dedicated example to docs for the `disableAnimations` prop?

mnajdova:
@layershifter yes, will add some examples. Do you agree with adding examples in the Provider's page for the `rtl`, `disableAnimations` and `renderer` props as those are now in it's API?

layershifter:
@mnajdova  Not sure about `rtl` and `renderer`, but we definitely should have an example for animations.

mnajdova:
I added for rtl and disableAnimations. As the renderer is now something we are calculating, will not specify example for it. I don't see why we should not have an example of how rtl is used, as we have it as an option on each example.
image
@layershifter let me know if you have anything else.

PR #1380, opened 2019-05-23
chore(behaviors): some a11y behavior improvements

### LINKED ISSUE #1346: Review behavior testDefinitions.ts and improve rules
## Feature Request
Some rules can be merged better into fewer, more flexible ones. Having too many rules may be confusing while developing a11y behavior related features.
### Problem description
For instance, create a single rule that checks:
- if an attribute with a name has been added.
- optional value for that attribute.
- optional part of the component where it has been added.
- optional based on what prop of the component or a component's shorthand.
- optional value of that prop.

Right now, to cover cases above, there are multiple rules with repeated code.

v 0.30.0

silviuaavram:
Closes https://github.com/stardust-ui/react/issues/1346.

After going through the rules I decided not to change anything. Instead I removed a couple unused ones and went through the behaviors and added more tests to some of them.

### DIFF
--- a/packages/react/src/lib/accessibility/Behaviors/Alert/alertWarningBehavior.ts
+++ b/packages/react/src/lib/accessibility/Behaviors/Alert/alertWarningBehavior.ts
@@ -3,7 +3,7 @@ import { Accessibility } from '../../types'
 /**
  * @specification
  * Adds role 'alert' to 'content' component's part.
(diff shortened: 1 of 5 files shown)


sophieH29 (approved):

bmdalex (approved):
👍

jurokapsiar (approved):

PR #1383, opened 2019-05-23
chore: tolerate build errors only for 'yarn start'

kuzhelov:
These changes are necessary for now, given that Circle CI steps are running with settings of DEV environment. 

This is a separate issue to address (as simple NODE_ENV being set to 'production' leads to another issue - the fact that lint warnings are not tolerated anymore), so for now errors will be tolerated only for single whitelisted `yarn start` scenario.

- [x] check that provided changes detect problems during the build: https://github.com/stardust-ui/react/pull/1381

### DIFF
--- a/build/gulp/plugins/gulp-webpack.ts
+++ b/build/gulp/plugins/gulp-webpack.ts
@@ -1,8 +1,10 @@
 import * as webpack from 'webpack'
 import config from '../../../config'
 
-const { __DEV__ } = config.compiler_globals
+const { __DEV__, __SKIP_ERRORS__ } = config.compiler_globals
 const { log, PluginError } = require('gulp-load-plugins')().util
+
+const DEV_SKIP_ERRORS = __DEV__ && __SKIP_ERRORS__
 
 const webpackPlugin = (webpackConfig, cb, onComplete = (err: any, stats: any) => {}) => {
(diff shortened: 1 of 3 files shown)


build/gulp/plugins/gulp-webpack.ts
```diff
@@ -15,11 +17,11 @@ const webpackPlugin = (webpackConfig, cb, onComplete = (err: any, stats: any) =>
       log('Webpack compiler encountered a fatal error.')
       throw new PluginError('webpack', err.toString())
     }
-    if (!__DEV__ && errors.length > 0) {
+    if (!DEV_SKIP_ERRORS && errors.length > 0) {
```
kuzhelov, line 20: this is where condition for tolerating errors is restricted

package.json
```diff
@@ -24,7 +24,7 @@
     "release:minor": "yarn prerelease && lerna publish minor && yarn postrelease",
     "release:patch": "yarn prerelease && lerna publish patch && yarn postrelease",
     "prestart": "yarn satisfied --fix yarn",
-    "start": "gulp --series dll docs",
+    "start": "cross-env SKIP_ERRORS=true gulp --series dll docs",
```
kuzhelov, line 27: this is where corresponding flag is set, only for `start` script

mnajdova (approved):

PR #1384, opened 2019-05-23
feat(menu): automatic positioning inside viewport

### LINKED ISSUE #1360: Menus with submenus are cut-off
### Steps
Open Menus and submenus at lower part of screen
image

### Expected Result
menus and submenus are automatically rendered inside viewport (on top)

### Actual Result
menus and submenus are cut off when there is not enough space in the viewport 💣

### Version
v0.30.0

bmdalex:
# feat(menu): automatic positioning inside viewport

Fixes #1360

## Description
Open Menus and submenus at lower part of screen
image

### Expected Result
menus and submenus are automatically rendered inside viewport (on top)

### Actual Result
menus and submenus are cut off when there is not enough space in the viewport 💣

## DEMO with current behavior
Screen Recording 2019-05-23 at 19 13 31

### DIFF
--- a/packages/react/src/components/Menu/MenuItem.tsx
+++ b/packages/react/src/components/Menu/MenuItem.tsx
@@ -247,6 +247,7 @@ class MenuItem extends AutoControlledComponent<WithAsProp<MenuItemProps>, MenuIt
               align={vertical ? 'top' : 'start'}
               position={vertical ? 'after' : 'below'}
               targetRef={this.itemRef}
+              modifiers={{ flip: { flipVariationsByContent: true } }}
             >
               {Menu.create(menu, {
                 defaultProps: {
--- a/packages/react/src/components/Menu/MenuItem.tsx
+++ b/packages/react/src/components/Menu/MenuItem.tsx
@@ -32,6 +32,7 @@ import {
   withSafeTypeForAs,
 } from '../../types'
 import { focusAsync } from '../../lib/accessibility/FocusZone'
+import { Popper } from '../../lib/positioner'
 
 export interface MenuItemSlotClassNames {
   wrapper: string
@@ -242,18 +243,24 @@ class MenuItem extends AutoControlledComponent<WithAsProp<MenuItemProps>, MenuIt
       menu && active && menuOpen ? (
         <>
           <Ref innerRef={this.menuRef}>
-            {Menu.create(menu, {
-              defaultProps: {
-                accessibility: submenuBehavior,
-                className: MenuItem.slotClassNames.submenu,
-                vertical: true,
-                primary,
-                secondary,
-                styles: styles.menu,
-                submenu: true,
-                indicator,
-              },
-            })}
+            <Popper
+              align={vertical ? 'top' : 'start'}
+              position={vertical ? 'after' : 'below'}
+              targetRef={this.itemRef}
+            >
+              {Menu.create(menu, {
+                defaultProps: {
+                  accessibility: submenuBehavior,
+                  className: MenuItem.slotClassNames.submenu,
+                  vertical: true,
+                  primary,
+                  secondary,
+                  styles: styles.menu,
+                  submenu: true,
+                  indicator,
+                },
+              })}
+            </Popper>
           </Ref>
           <EventListener listener={this.outsideClickHandler} targetRef={documentRef} type="click" />
         </>
--- a/packages/react/src/lib/getScrollParent.ts
+++ b/packages/react/src/lib/getScrollParent.ts
@@ -1,5 +1,3 @@
-import isBrowser from './isBrowser'
-
 /**
  * Returns the parentNode or the host of the element
  * @argument {Node} node
@@ -34,8 +32,6 @@ const getStyleComputedProperty = (node: Node, property?: string) => {
  * @returns {Node} scroll parent
  */
 const getScrollParent = (node: Node): Node => {
-  if (!isBrowser()) return null
-
   // Return body, `getScroll` will take care to get the correct `scrollTop` from it
   if (!node) return document.body
 
--- a/packages/react/src/lib/getScrollParent.ts
+++ b/packages/react/src/lib/getScrollParent.ts
@@ -1,0 +1,21 @@
+import isBrowser from './isBrowser'
+
+const SCROLL_CSS_VALUE_REGEX = /(auto|scroll)/
+
+const isScrollable = (element: Element) => {
+  try {
+    const { overflow, overflowY, overflowX } = window.getComputedStyle(element)
+    return SCROLL_CSS_VALUE_REGEX.test(overflow + overflowX + overflowY)
+  } catch (e) {
+    return false
+  }
+}
+
+const getScrollParent = (element: Element) => {
+  if (!isBrowser()) return null
+  if (!element || element === document.body) return document.body
+  if (isScrollable(element)) return element
+  return getScrollParent(element.parentNode as Element)
+}
+
+export default getScrollParent
--- a/packages/react/src/lib/getScrollParent.ts
+++ b/packages/react/src/lib/getScrollParent.ts
@@ -1,21 +1,57 @@
 import isBrowser from './isBrowser'
 
-const SCROLL_CSS_VALUE_REGEX = /(auto|scroll)/
-
-const isScrollable = (element: Element) => {
-  try {
-    const { overflow, overflowY, overflowX } = window.getComputedStyle(element)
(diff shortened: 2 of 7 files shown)


kuzhelov (approved):

miroslavstastny (approved):

mnajdova (approved):

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -242,18 +243,25 @@ class MenuItem extends AutoControlledComponent<WithAsProp<MenuItemProps>, MenuIt
...
-                styles: styles.menu,
-                submenu: true,
-                indicator,
-              },
-            })}
+            <Popper
+              align={vertical ? 'top' : 'start'}
+              position={vertical ? 'after' : 'below'}
+              targetRef={this.itemRef}
+              modifiers={{ flip: { flipVariationsByContent: true } }}
```
layershifter, line 250: This modifier was introduced in `1.15.0`, we should increase required PopperJS version
https://github.com/FezVrasta/popper.js/releases/tag/v1.15.0

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -43,6 +50,21 @@ const Popper: React.FunctionComponent<PopperProps> = props => {
 
   React.useEffect(
     () => {
+      const pointerTargetRefElement = pointerTargetRef && pointerTargetRef.current
+      const popperHasScrollableParent = getScrollParent(contentRef.current) !== document.body
+
+      const modifiers: PopperJS.Modifiers = _.merge(
+        popperHasScrollableParent && flipForScrollParentModifiers,
```
layershifter: Let's inline them, it's not obvious what `flipForScrollParentModifiers` does

packages/react/src/lib/getScrollParent.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+const getScrollParent = (element: Element) => {
+  if (!isBrowser()) return null
+  if (!element || element === document.body) return document.body
+  if (isScrollable(element)) return element
+  return getScrollParent(element.parentNode as Element)
+}
+
+export default getScrollParent
```
layershifter: Can we use original function from Popper?
https://github.com/FezVrasta/popper.js/blob/master/packages/popper/src/utils/getScrollParent.js

packages/react/src/components/Menu/MenuItem.tsx
bmdalex: This is not going to break anything since we still download `1.15.0` (see yarn.lock); will do after OSS approval: #1403 for tracking

packages/react/src/lib/positioner/Popper.tsx
bmdalex: done, thanks

packages/react/src/lib/getScrollParent.ts
bmdalex: done, thanks

layershifter (approved):

packages/react/src/lib/getScrollParent.ts
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  const window = node.ownerDocument.defaultView
+  const css = window.getComputedStyle(node as Element, null)
+  return property ? css[property] : css
+}
+
+/**
+ * Returns the scrolling parent of the given element
+ * @argument {Node} node
```
kuzhelov: nit: `@argument {Node} node - DOM node.` - to disambiguate

packages/react/src/lib/getScrollParent.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * Returns the parentNode or the host of the element
+ * @argument {Node} node
```
kuzhelov: lets provide more comments here on what these nodes are: 

```js
 @argument {Node} node - DOM node.
@returns {Node} parent DOM node.
```

packages/react/src/lib/getScrollParent.ts
```diff
@@ -0,0 +1,53 @@
...
+    return node
+  }
+
+  return node.parentNode || (node as any).host
+}
+
+/**
+ * Get CSS computed property of the given element
+ * @argument {Node} node
+ * @argument {string} property
```
kuzhelov: lets add short comments for the arguments, what they stand for.

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -43,6 +45,24 @@ const Popper: React.FunctionComponent<PopperProps> = props => {
 
   React.useEffect(
     () => {
+      const pointerTargetRefElement = pointerTargetRef && pointerTargetRef.current
+      const popperHasScrollableParent = getScrollParent(contentRef.current) !== document.body
+
+      const modifiers: PopperJS.Modifiers = _.merge(
+        popperHasScrollableParent && {
```
kuzhelov, line 57: may we leave comments for each of the modifiers used here - like short explanation that suggests the case it is aimed to address? See this being especially useful for the first and the last one in the list

packages/react/src/lib/getScrollParent.ts
```diff
@@ -0,0 +1,53 @@
...
+  switch (node.nodeName) {
+    case 'HTML':
+    case 'BODY':
+      return node.ownerDocument.body
+    case '#document':
+      return (node as Document).body
+  }
+
+  // Firefox wants us to check `-x` and `-y` variations as well
+  const { overflow, overflowX, overflowY } = getStyleComputedProperty(node)
```
kuzhelov: sorry, seems that this is not completely clear to me what we are checking for here - could you, please, suggest more details on that? Thank you!

packages/react/src/lib/getScrollParent.ts
```diff
@@ -0,0 +1,53 @@
...
+    return []
+  }
+
+  const window = node.ownerDocument.defaultView
+  const css = window.getComputedStyle(node as Element, null)
+  return property ? css[property] : css
+}
+
+/**
+ * Returns the scrolling parent of the given element
```
kuzhelov: nit: lets end comments with dots, to be consistent

packages/react/src/lib/getScrollParent.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @argument {Node} node
+ * @returns {Node} scroll parent
+ */
+const getScrollParent = (node: Node): Node => {
```
kuzhelov: actually, very much worrying that all this `Popper`'s functionality is not covered by tests. Especially given that now we are introducing modifier to Popper's logic based on some pretty involving condition

packages/react/src/lib/getScrollParent.ts
bmdalex: done, thanks

packages/react/src/lib/getScrollParent.ts
bmdalex: done, thanks

packages/react/src/lib/getScrollParent.ts
bmdalex: done, thanks

packages/react/src/lib/positioner/Popper.tsx
bmdalex: done, thanks

packages/react/src/lib/getScrollParent.ts
bmdalex: done, thanks

packages/react/src/lib/getScrollParent.ts
bmdalex: done, thanks

packages/react/test/specs/lib/positioner/getScrollParent-test.ts
```diff
@@ -0,0 +1,99 @@
...
+}
+
+const setStylesForElements = (elements: HTMLElement[], style: Partial<CSSStyleDeclaration>) =>
+  elements.forEach(element => setStylesForElement(element, style))
+
+const resetOverflowStyles = (element: HTMLElement) =>
+  ['overflow', 'overflowX', 'overflowY'].map(prop => (element.style[prop] = ''))
+
+const testsSetupFactory = () => {
+  const treeElements = Array(4)
```
kuzhelov: `_.range(4).map(...)`

bmdalex: done. thanks

packages/react/src/lib/getScrollParent.ts
bmdalex: added unit tests

PR #1386, opened 2019-05-24
fix(menuItemBehavior): Remove redundant aria attributes

sophieH29:
This PR removes 'aria-posinset' and 'aria-setsize' attributes from `menuitemBehavior` as they're no more needed.

### DIFF
--- a/packages/react/src/lib/accessibility/Behaviors/Menu/menuItemBehavior.ts
+++ b/packages/react/src/lib/accessibility/Behaviors/Menu/menuItemBehavior.ts
@@ -45,8 +45,6 @@ const menuItemBehavior: Accessibility = (props: any) => ({
         ? true
         : undefined,
       [IS_FOCUSABLE_ATTRIBUTE]: !props['disabled'],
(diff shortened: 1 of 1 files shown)


packages/react/src/lib/accessibility/Behaviors/Menu/menuItemBehavior.ts
```diff
@@ -45,8 +45,6 @@ const menuItemBehavior: Accessibility = (props: any) => ({
         ? true
         : undefined,
       [IS_FOCUSABLE_ATTRIBUTE]: !props['disabled'],
-      'aria-posinset': props.itemPosition,
```
layershifter, line 48: Should we remove this props in component too? I.e. fully revert #1066?

jurokapsiar: no, we would like still to have the possibility for the users to pick them up if they want

jurokapsiar (approved):

bmdalex (approved):

PR #1387, opened 2019-05-24
chore(package): use Project References in TS

layershifter:
Related to #508.

----

This PR introduces changes to our `tsconfig.json`'s to use [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html) feature. It was designed also to handle monorepo cases (https://github.com/Microsoft/TypeScript/issues/25376).

> When you reference a project, new things happen:
> - Importing modules from a referenced project will instead load its output declaration file (.d.ts)
> - If the referenced project produces an outFile, the output file .d.ts file’s declarations will be visible in this project

https://www.typescriptlang.org/docs/handbook/project-references.html

### Generated types structure

This feature allows to properly define `rootDir` option in `tsconfig.json` and output proper typings by compiler:

image

Without references feature it will fail because TypeScript doesn't know anything about modules structure:
image


layershifter:
image

mnajdova (approved):

PR #1388, opened 2019-05-24
chore(ts): rearrange configs to disallow use aliases

layershifter:
Picked from #1387.

This PR updates `paths` option in our project structure and disallows to use `src`, `docs` and `test` aliases in package directories.

### Motivation

- creates issues with typings
- one of the most common reasons of failed CI builds
- editors use it for auto import and it is easy to miss them

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "release:major": "yarn prerelease && lerna publish major && yarn postrelease",
     "release:minor": "yarn prerelease && lerna publish minor && yarn postrelease",
     "release:patch": "yarn prerelease && lerna publish patch && yarn postrelease",
-    "prestart": "yarn satisfied --fix yarn",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "release:major": "yarn prerelease && lerna publish major && yarn postrelease",
     "release:minor": "yarn prerelease && lerna publish minor && yarn postrelease",
     "release:patch": "yarn prerelease && lerna publish patch && yarn postrelease",
-    "prestart": "yarn satisfied --fix yarn",
+    "prestart": "yarn satisfied",
```
layershifter, line 26: https://github.com/stardust-ui/react/pull/1301#discussion_r281915689

bmdalex (approved):
yaay 👍

PR #1393, opened 2019-05-26
chore(behavior): rename toolbarBehavior to menuAsToolbarBehavior

miroslavstastny:
### BREAKING CHANGE
- rename `toolbarBehavior` to `menuAsToolbarBehavior`
- rename `toolbarButtonBehavior` to `menuItemAsToolbarButtonBehavior`

### Details
We plan to add `toolbarBehavior` to be used by `Toolbar` component, therefore we need to rename existing `toolbarBehavior`.
Current `toolbarBehavior` which uses `toolbarButtonBehavior` expect children items to have `wrapper` and `root` slots - this works for `Menu` component (which renders as `ul > li > a`) but is not suitable for `Toolbar` component (which is expected to render as `div > button`).
Discussed offline with @jurokapsiar


docs/src/components/ComponentDoc/ComponentControls/ComponentControls.tsx
```diff
@@ -69,13 +63,12 @@ const ComponentControls: React.FC<ComponentControlsProps> = props => {
...
         fluid
         pills
-        accessibility={toolbarBehavior}
+        accessibility={menuAsToolbarBehavior}
         items={[
           {
             key: 'show-code',
             content: <ComponentButton iconName="code" label="Try it" />,
             onClick: onShowCode,
-            accessibility: toolbarButtonBehavior,
```
miroslavstastny, line 78: This is no longer necessary with #1282

sophieH29:
@miroslavstastny shouldn't it be `menuItemAsToolbarButtonBehavior`?
I know it's longer, but `menuAsToolbarButtonBehavior` might not give any sense

layershifter (approved):

PR #46, opened 2018-08-03
fix: remove allowSyntheticDefaultImports and allow it on docs

aniknafs:
allowSyntheticDefaultImports is not allowed on neither of teams-modular-packages/WebClient. This PR will disable it for /src but will allow it on the docs site.

As a result, all import statements from packages with no default import is updated: 
e.g. `import React from 'react'` --> `import * as React from 'react'`

There are a few instances where a 3rd party dependency has an incomplete type def, which are skipped by type assertion:
e.g. `const inputClasses = cx(classes.input)` --> `const inputClasses = (cx as any)(classes.input)`

PR #1395, opened 2019-05-27
chore(package): use @babel/register instead of ts-node

layershifter:
### Motivation

We are using `ts-node` to compile our TS code on the fly during Gulp tasks and Screener run. As we we're switching to use Babel (#508), we can use `@babel/register` as it is faster than `ts-node` (there are no type checks during run).

This change is required for #1387, otherwise TS configs will become complicated and hard to maintain: _we're calling Gulp with different configs_ that in package's root which results us to strange build commands:
`"build": "cross-env TS_NODE_PROJECT=../../tsconfig.json gulp bundle:package:no-umd --package react-component-ref"`

#### Remove custom TSLint rule

We don't need it more as these cases are covered by [`react/no-unknown-property`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unknown-property.md) 🚀 

#### `allowSyntheticDefaultImports` in `/build`

> This rule allows you to use ES style default imports even when the code you’re importing doesn’t have an ES default export. This happens when you import, for example, React, whose code is not written in ES (the source code is, but React ships a built version). This means that it technically doesn’t have an ES default export, so TypeScript will tell you so when you import it.

We **can't** enable this option in our main project (`react`) because we can produce broken typings (#46) for our consumers who don't enable this option. But it's safe to enable it for our tools, otherwise it becomes endless war with Babel because it handles CommonJS modules differently:

```js
// Module
module.exports = () => {}

// Babel
import cx from 'module' // correct
import * as cx from 'module' // TypeError: express is not a function

// Typescript
import cx from 'module' // Typings error: no default export
import * as cx from 'module'
```

More in [Babel REPL](https://babeljs.io/repl/#?babili=false&browsers=last%208%20versions%2C%20safari%20%3E%208%2C%20firefox%20%3E%2023%2C%20chrome%20%3E%2024%2C%20opera%20%3E%2015%2C%20not%20ie%20%3C%2011%2C%20not%20ie_mob%20%3C%3D%2011&build=&builtIns=false&spec=false&loose=false&code_lz=JYWwDg9gTgLgBAKjgQwM5wMYEY4DMoQhwDkIEAJgK4A2ApgLRbFwBQoksmATHgUaRRoMuzFi1oAPDvHK1cyGvADemLABpucAL5A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Ctypescript&prettier=true&targets=Node-8.9&version=7.4.5&externalPlugins=).

We had the same issue before even in distributed code (`classnames` #570, https://github.com/stardust-ui/react/blob/v0.31.0/packages/react/src/lib/felaExpandCssShorthandsPlugin.ts#L7).

`create-react-app` enables `allowSyntheticDefaultImports: true` and `skipLibCheck: true` by default: https://github.com/facebook/create-react-app/blob/v3.0.1/packages/react-scripts/scripts/utils/verifyTypeScriptSetup.js#L106


build/tslint/jsxAttributeNameRule.ts
```diff
@@ -1,35 +0,0 @@
-import * as Lint from 'tslint'
```
layershifter, line 1: Don't need this more, is catched by `react/no-unknown-property`

```
16:28  error  Unknown property 'stop-color' found, use 'stopColor' instead  react/no-unknow
n-property
```

PR #1396, opened 2019-05-27
chore(getElementType): remove unused logic branches

kuzhelov:
This PR removes couple of branches from logic of `getElementType` method that infers element type for rendered Stardust component.

One of these branches - `getDefault`, - wasn't used anywhere, while another one provides unnecessary complication to element type's inference.

## BREAKING CHANGES
_**These changes are NOT breaking for clients that are using Typescript, as well as highly likely to introduce no breaking changes to JS clients, given that they are affecting edge-case usages.**_

The following elements that were originally rendered as `a` now will be rendered as `as` or default element of the corresponding component:

```jsx
// - was rendered as <a href>{content}</a> 
// - now is rendered as <span href>{content}</span> 
<Text href='..' content={...} />  
```
### Migration path
It is necessary to provide value for `as` prop explicitly:
```jsx
<Text as='a' href='..' content={...} /> 
```


miroslavstastny (approved):
👍 on code changes.
But as this functionality goes back to SUIR, it might be useful to hear @levithomason opinion.

PR #1398, opened 2019-05-27
chore(Perf): describe current workflow

miroslavstastny:
Describe current workflow or automatic performance measuring.


perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
+# Performance measuring
+
+*You can't improve what you can't measure.*
+
+As part of Stardust build pipeline we run a set of performance tests in order to avoid any regressions. This document describes the workflow.
+
+## Measures
+
+### Performance examples
+For individual Stardust components there are usage examples in public docsite. Beside these *usage* examples there are also *performance* examples (not currently available in the public docsite). These examples render components with more data (ie list of 30 chat messages) so it takes more time (~100ms) to render them and makes any changes in render times visible.
```
kuzhelov: nit: 'Besides', 'currently not available'

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+
+An example measures a performance of a single component. We currently do not have any compound perf examples, like a whole page built using several Stardust components.
```
kuzhelov: Maybe rephrase this a bit, to emphasize the main point provided by this paragraph:

> Current perf examples measure performance of a single component - at the moment there are no any compound perf examples, like a whole page built using several Stardust components.

perf/doc/perf.md
kuzhelov: > These examples render components with more data (ie list of 30 chat messages) so it takes more time (~100ms) to render them and makes any changes in render times visible.

Frankly, am a bit lost while trying to get idea of this sentence. May you, please, suggest what are the main points you'd want to share here?

Is it, essentially, about saying that we would like to have running times that surpass the margin of potential measuring errors?

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Measures
+
+### Performance examples
+For individual Stardust components there are usage examples in public docsite. Beside these *usage* examples there are also *performance* examples (not currently available in the public docsite). These examples render components with more data (ie list of 30 chat messages) so it takes more time (~100ms) to render them and makes any changes in render times visible.
+
+An example measures a performance of a single component. We currently do not have any compound perf examples, like a whole page built using several Stardust components. 
+
+### Getting the numbers
+To get a performance number we use [Puppeteer](https://github.com/GoogleChrome/puppeteer) to render a performance example in headless Chrome and measure the **first render time**.
```
kuzhelov: nit: 
> To get performance numbers, 
> to render ~a performance example~ in headless Chrome and measure the **first render time**.

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+
+To stabilize the numbers, the example is rendered 50 times and median value is used.
+
+## Workflow
+1. With each GIT commit, build pipeline is executed in CircleCI. As part of the build, we run the performance tests for all performance examples.
+2. Performance results are then stored to an external storage.
+3. In Stardust docsite (not currently available in public docsite) for each measured component a performance chart is displayed, showing measures for the component from last 50 test runs.
```
kuzhelov: it is unclear how Stardust docsite is different from the 'public' docsite :) Maybe rephrase to something like

> In Stardust docsite **running locally** (currently not available in public dociste https://stardust-ui.github.io/react/)

nit: just to simplify reading:

> showing measures ~for the component~ from last 50 test runs

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+
+The following picture shows a chart for `Chat` component performance example:
```
kuzhelov: > shows an **example** chart for `Chat` component ~performance example~:

This will hint that this chart is not 'live', and it won't update with more perf data being accumulated. Also, given that you've already explained how perf data is collected, it should be fine to omit words about perf examples here.

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+![Chart example](perf_chart.png)
+ - x-axis: build number (newer builds on right)
+ - y-axis: first render time in milliseconds (the lower the better)
+ - tooltip: details about highlighted build
+ 
+The example chart shows that something wrong happened around build 12847 and performance decreased by ~50%.
```
kuzhelov: nit-nit: we may simplify it a bit and leave only pure facts there (maybe this 'something' wasn't, actually, wrong:)

> This example chart suggests that something around build 12847 had decreased performance by ~50%.

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+The following picture shows a chart for `Chat` component performance example:
+![Chart example](perf_chart.png)
+ - x-axis: build number (newer builds on right)
+ - y-axis: first render time in milliseconds (the lower the better)
+ - tooltip: details about highlighted build
+ 
+The example chart shows that something wrong happened around build 12847 and performance decreased by ~50%. 
+
+## Automatic PR gates
+There was a plan to introduce an automated check to all PRs - to run the same tests as part of a PR build and compare them with relevant master build - and if the performance decreases then block that PR.
```
kuzhelov: nit: 

> There **is** a plan to introduce an automated check for PR builds: measure performance and block the PR if performance had significantly decreased compared to master.

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+
+Unfortunately, the numbers are currently not stable enough to be used for automatic gates. In the example above you can see that even without any relevant change in code, the results differ by ~20%.
```
kuzhelov: nit: remove personality

> However, current numbers are not stable enough to be used for automatic gates. From the example above it may be seen that even without any relevant changes in code, the results still vary in range of ~20%.

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+
+One of the reasons can be that the tests are run on a shared CI machine with non-deterministic performance and load. To mitigate this issue we tried to normalize the measures (see [#1280](https://github.com/stardust-ui/react/pull/1280) for details). Before rendering each example, we measure a baseline example which is supposed to have constant render time and then normalize the real example measure by this. However the normalized results are quite similar to the plain ones.
```
kuzhelov: would suggest to skip the part related to actual speculations, and just leave the one which describes that measured times are normalised - and would suggest to move this to `Getting the numbers` section

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Automatic PR gates
+There was a plan to introduce an automated check to all PRs - to run the same tests as part of a PR build and compare them with relevant master build - and if the performance decreases then block that PR.
+
+Unfortunately, the numbers are currently not stable enough to be used for automatic gates. In the example above you can see that even without any relevant change in code, the results differ by ~20%.
+
+One of the reasons can be that the tests are run on a shared CI machine with non-deterministic performance and load. To mitigate this issue we tried to normalize the measures (see [#1280](https://github.com/stardust-ui/react/pull/1280) for details). Before rendering each example, we measure a baseline example which is supposed to have constant render time and then normalize the real example measure by this. However the normalized results are quite similar to the plain ones. 
+
+So **currently, it is not possible to enable automatic PR gates for performance, but still if a real person looks at the charts performance gains/problems are noticable so we can react accordingly.**
```
kuzhelov: I think that we could simplify this statement and make it more bold (in figural sense), as currently it has some sort of justification sentiment:)

> This makes it to be not possible to enable reliable automatic PR gates for performance. At the same time, still, there are performance charts that make all serious performance degradations being discoverable.

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+
+Unfortunately, the numbers are currently not stable enough to be used for automatic gates. In the example above you can see that even without any relevant change in code, the results differ by ~20%.
+
+One of the reasons can be that the tests are run on a shared CI machine with non-deterministic performance and load. To mitigate this issue we tried to normalize the measures (see [#1280](https://github.com/stardust-ui/react/pull/1280) for details). Before rendering each example, we measure a baseline example which is supposed to have constant render time and then normalize the real example measure by this. However the normalized results are quite similar to the plain ones. 
+
+So **currently, it is not possible to enable automatic PR gates for performance, but still if a real person looks at the charts performance gains/problems are noticable so we can react accordingly.** 
+
+## Bundle size
+Not only the rendering speed is affecting perceived performance. The size of the package (which affects application launch time) is equally important. For that reason we are also measuring generated bundle size of whole package as well as the size of individual themes and components.
```
kuzhelov: nit: could be simplified

>  For that reason bundle size measures are also taken - for the whole `react` package, as well as for individual themes and components.

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+
+The numbers are currently being measured and saved by the same infrastructure as the performance but are not displayed anywhere yet.
```
kuzhelov: > Bundle size numbers are stored at the same infrastructure used for storing performance measures - however, they are not displayed anywhere yet.

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+- explore possible ways to make the measurements more stable,
+- add perf examples (charts) for more components and more variants,
+- add perf examples for compound examples,
+- make perf charts available in production docsite,
+- make bundle size charts available in docsite,
+
+## FAQ
+
+### Are all components and all their variants covered?
+No. We add examples as needed. If we expect a problem somewhere or a perf issue is reported, we add a new perf example which covers that area.
```
kuzhelov: nit: remove personality, simplify

> No. Examples are added as needed. With a perf issue discovered, a new perf example is added to cover related area.

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+
+So **currently, it is not possible to enable automatic PR gates for performance, but still if a real person looks at the charts performance gains/problems are noticable so we can react accordingly.** 
+
+## Bundle size
+Not only the rendering speed is affecting perceived performance. The size of the package (which affects application launch time) is equally important. For that reason we are also measuring generated bundle size of whole package as well as the size of individual themes and components.
+
+The numbers are currently being measured and saved by the same infrastructure as the performance but are not displayed anywhere yet.
+
+## Future plans
+- explore possible ways to make the measurements more stable,
```
kuzhelov: nit: do we need commas? May suggest to remove them

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+- add perf examples (charts) for more components and more variants,
+- add perf examples for compound examples,
```
kuzhelov: lets, again, explain by example what is meant by `compound` example:

> - add perf examples for compound examples (app's page example built with Stardust components)

perf/doc/perf.md
```diff
@@ -0,0 +1,56 @@
...
+- make perf charts available in production docsite,
```
kuzhelov: again, not quite clear what `production` docsite means. May suggest to rephrase:

> make perf charts available in public docsite (https://stardust-ui.github.io/react/)

kuzhelov (approved):

perf/doc/perf.md
miroslavstastny: This is how **usage** examples and **perf** examples differ.
And the time - exactly as you wrote. I'll try to rephrase.

PR #1399, opened 2019-05-27
refactor(customPropTypes): remove usage of _.chain

layershifter:
Picked from #1169.

This PR removes the usage of Lodash's `_.chain()`:
https://medium.com/making-internets/why-using-chain-is-a-mistake-9bc1f80d51ba

The main problem: with `_.chain()` you will include the whole Lodash and you will not be able to pick single modules. `babel-plugin-lodash` will also stuck in this case.

### DIFF
--- a/packages/react/test/specs/components/Box/Box-test.ts
+++ b/packages/react/test/specs/components/Box/Box-test.ts
@@ -1,8 +1,0 @@
-import Box from 'src/components/Box/Box'
-import { isConformant } from 'test/specs/commonTests'
-
-describe('Box', () => {
-  xdescribe('is conformant', () => {
-    isConformant(Box)
-  })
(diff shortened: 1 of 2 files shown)


packages/react/test/specs/components/Box/Box-test.ts
```diff
@@ -1,8 +0,0 @@
-import Box from 'src/components/Box/Box'
-import { isConformant } from 'test/specs/commonTests'
-
-describe('Box', () => {
-  xdescribe('is conformant', () => {
-    isConformant(Box)
-  })
-})
```
layershifter, line 8: This file does nothing, let's remove dead code.

bmdalex (approved):

PR #1400, opened 2019-05-27
refactor(Ref): use only one type for props

layershifter:
Picked from #1169.

It's strange that we have separated types for these components, this PR removes `RefFindNodeProps` and `RefForwardProps` because they are useless.


bmdalex (approved):
👍

PR #1401, opened 2019-05-27
fix(eslint): address warnings

kuzhelov:
This PR addresses the problem of eslint warnings for our repo: there were 55 warnings previously reported:

image

These warnings are turning into errors in production environment (`NODE_ENV` being equal to `production`), and **this has prevented us from running CI scripts agains PROD**. Now the issue is fixed.

image

### DIFF
--- a/build/.eslintrc.json
+++ b/build/.eslintrc.json
@@ -1,0 +1,7 @@
+{
+  "extends": ["../packages/internal-tooling/eslint/index.js"],
+  "rules": {
+    "no-console": "off"
+  },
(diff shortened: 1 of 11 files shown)


layershifter (changes requested):
Please use `eslint-disable-line`/`eslint-disable-next-line` because `eslint-disable` disables rules until `eslint-enable` will be called

https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments

layershifter (approved):

PR #1402, opened 2019-05-28
feat(Tree): implement Left and Right Arrow navigation

silviuaavram:
Removes the circular navigation in focus zone.
`TreeItem` generates 2 `Refs` that are filled with the anchor in the `TreeTitle`.

1. `titleRef` is used in `focusParent` function that will be passed to the children `TreeTitle` in order for them to be able to focus their parent by Left Arrow Key.
2. `firstTitleChildRef` is used in `focusFirstChild` function that will be passed to the `TreeTitle` in order for it to be able to focus on the first child title by Right Arrow Key. The `Ref` itself is passed to the first child `TreeTitle` so that it can be filled with the element.

The other changes involve the actual 'travelling' of these `Refs` and focus functions throughout the Tree-TreeItem-TreeTitle structure so that they can achieve their purpose.


bmdalex:
@silviuavram - there's a bad merge happening there, difficult to see the actual changes, can you take a look?

silviuaavram:
bad merge and taking another approach. will link another PR soon

PR #1404, opened 2019-05-28
chore(babel): switch to Babel in distribution

### LINKED ISSUE #508: RFC: switch to Babel

layershifter:
Fixes #508.

This PR is a final step in Babel migration and aims to improve and simplify our bundling, too.

### Use `gulp-babel`

Now we use `gulp-babel` to transpile files.

### Switch to `@babel/runtime`

As we switched to Babel, we also replaced `tslib` with `@babel/runtime`.

#### why `@babel/runtime` over `@babel/runtime-corejs*`?

It's commonly used by other libraries:

- https://github.com/mui-org/material-ui/blob/v4.0.2/packages/material-ui/package.json#L40
- https://github.com/react-bootstrap/react-bootstrap/blob/v1.0.0-beta.9/package.json#L115
- https://github.com/segmentio/evergreen/blob/master/package.json#L45
- https://github.com/reactstrap/reactstrap/blob/master/package.json#L101

### `babel-plugin-lodash` 

This PR also introduces the first babel plugin that aims to optimize bundle size. Lodash is CJS and can't be treeshaked, it means that `import _ from 'lodash'` will include the whole library (70 Kb) 😭 

This plugin will transform `import _ from 'lodash'` to `import _map from 'lodash/map'` where `_map` is module required in your file.

image

### Bundle size

Benefits in bundle size are received from decreasing `lodash`: from `97Kb` to `51Kb`.

#### `@stardust-ui/react-proptypes`
- **76.3 kB** | https://bundlephobia.com/result?p=@stardust-ui/react-proptypes@0.32.0 
- **35.8 kB** | https://bundlephobia.com/result?p=@stardust-ui/react-proptypes@0.33.0-alpha.16

#### `@stardust-ui/react-component-ref`
- **75.6 kB**  | https://bundlephobia.com/result?p=@stardust-ui/react-component-ref@0.32.0
- **37.7 kB** | https://bundlephobia.com/result?p=@stardust-ui/react-component-ref@0.33.0-alpha.16

Includes `@stardust-ui/react-proptypes`: we definitely need to perform advanced optimizations like this https://github.com/airbnb/prop-types/blob/master/index.js.

#### `@stardust-ui/react`
- **692.9kB** | https://bundlephobia.com/result?p=@stardust-ui/react@0.32.0
- **733.7kB** | https://bundlephobia.com/result?p=@stardust-ui/react@0.33.0-alpha.16

About increase of main package: `self@0.32` - 448KB vs `self@0.33` - 536KB


packages/internal-tooling/babel/index.js
```diff
@@ -7,11 +7,11 @@ module.exports = {
   plugins: [
     '@babel/plugin-proposal-class-properties',
     '@babel/plugin-syntax-dynamic-import',
-    '@babel/plugin-transform-runtime',
+    ['@babel/plugin-transform-runtime', { corejs: 2 }],
   ],
   env: {
-    development: {
-      plugins: ['react-hot-loader/babel'],
```
layershifter: Moved to Webpack's config directly, should fix issues with perf tests
FYI: @miroslavstastny

layershifter:
### Bundle size

#### `@stardust-ui/react-proptypes`
- **76.3 kB** | https://bundlephobia.com/result?p=@stardust-ui/react-proptypes@0.32.0 
- **35.8 kB** | https://bundlephobia.com/result?p=@stardust-ui/react-proptypes@0.33.0-alpha.16

#### `@stardust-ui/react-component-ref`
- **75.6 kB**  | https://bundlephobia.com/result?p=@stardust-ui/react-component-ref@0.32.0
- **37.7 kB** | https://bundlephobia.com/result?p=@stardust-ui/react-component-ref@0.33.0-alpha.16

#### `@stardust-ui/react`
- **692.9kB** | https://bundlephobia.com/result?p=@stardust-ui/react@0.32.0
- **733.7kB** | https://bundlephobia.com/result?p=@stardust-ui/react@0.33.0-alpha.16

About increase of main package: `self@0.32` - 448KB vs `self@0.33` - 536KB

kuzhelov:
nit: for bundle size stat in description, it will be more representative to share decrease in `lodash` size - (from 97Kb to 51Kb), as this is exactly the thing changes were aimed on

build/gulp/tasks/bundle.ts
```diff
@@ -43,30 +43,29 @@ const componentsSrc = [
...
+)
 
-  const typescript = g.typescript.createProject(tsConfig, settings)
-  const { dts, js } = src(componentsSrc).pipe(typescript())
+task('bundle:package:types', () => {
+  const tsConfig = paths.base('build/tsconfig.json')
+  const typescript = g.typescript.createProject(tsConfig, {
+    declaration: true,
+    isolatedModules: false,
+    module: 'esnext',
```
kuzhelov: check this flag

packages/internal-tooling/babel/index.js
```diff
@@ -23,8 +28,10 @@ module.exports = api => {
   const plugins = [
     '@babel/plugin-proposal-class-properties',
     '@babel/plugin-syntax-dynamic-import',
-    '@babel/plugin-transform-runtime',
-  ]
+    ['@babel/plugin-transform-runtime', { corejs: false, useESModules }],
```
kuzhelov: `false` is the default value for `corejs` option, seems that we can omit it

packages/internal-tooling/babel/index.js
```diff
@@ -1,18 +1,23 @@
...
+const isDistCaller = caller => {
+  return !!(caller && caller.name === 'babel-gulp')
+}
+const supportsESM = caller => {
+  return !!((caller && caller.name === 'babel-loader') || caller.useESModules)
 }
 
 module.exports = api => {
-  const isNode = api.caller(isBabelRegister) || api.caller(isJest)
+  const isDistBuild = api.caller(isDistCaller)
```
kuzhelov: nit: probably, `isDistBundle` name will better reflect semantics and related optimizations applied for this scenario

kuzhelov (approved):
As this is announced as a concluding step for Babel transition, would like to point out that there is a regression we currently have in comparison to the original TS-based setup: currently our `.ts` config files are not checked for type safety before executed.

This may introduce problems for rename refactorings, especially, when it is about changing the variable names that contain path values. We should ensure that our config code will be checked for type correctness _before_ being executed - behavior we've had before.

Lets file an issue for that, to keep this piece of work tracked.

PR #1405, opened 2019-05-28
feat(Checkbox): add component

### LINKED ISSUE #634: I am missing a checkbox component.
This feature request is to have a checkbox component which is having nicer styling than basic HTML checkbox element. It can look like as a toggle button. 

You can check how we implement it in Boolean renderer.

layershifter:
Fixes #634.

This PR adds a base implementation of `Checkbox` component with basic`checkboxBehavior`. Also adds `stardust-checkmark` icon that is used for `Checkbox` component.

### `<Checkbox checked />`

image

### `<Checkbox checked toogle />`

image

### Playground

checkbox


packages/react/src/themes/base/components/Checkbox/checkboxVariables.ts
```diff
@@ -0,0 +1,77 @@
...
+  disabledToggleColor: string
+  disabledToggleBackground: string
+  disabledToggleBorderColor: string
+
+  padding: string
+}
+
+export default (siteVars: any): CheckboxVariables => ({
+  checkboxBackground: 'transparent',
+  checkboxBorderColor: siteVars.colors.grey[300],
```
mnajdova, line 43: Let's try to use the color scheme here, as this is a fresh new component ;)

mnajdova: I meant in teams theme :)

packages/react/src/themes/teams/index.tsx
```diff
@@ -36,6 +36,7 @@ const themeIcons: ThemeIcons = Object.keys(svgIconsAndStyles as {
 
 const icons: ThemeIcons = {
   ...themeIcons,
+  'stardust-checkmark': themeIcons['stardust-checkmark'],
```
mnajdova: Can't we just use the `accept` icon from teams theme?

layershifter: Overlooked it, definitely it's better to use `accept` 👍

docs/src/components/ExampleSnippet/renderElementToJSX.ts
```diff
@@ -14,7 +14,7 @@ const renderElementToJSX = (element: React.ReactNode, triggerErrorOnRenderFn: bo
       (typeof element.type === 'function' ? 'NoDisplayName' : element.type),
     showDefaultProps: false,
     showFunctions: true,
-    filterProps: ['accessibility'],
+    filterProps: ['accessibility', 'onClick', 'onChange'],
```
mnajdova, line 17: 👍  maybe in the future we can make this filtering smarter, but definitely now we need them

packages/react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,153 @@
...
+    },
+  }
+
+  getInitialAutoControlledState(): CheckboxState {
+    return { checked: false, isFromKeyboard: false }
+  }
+
+  handleClick = (e: React.MouseEvent | React.KeyboardEvent) => {
+    const { disabled } = this.props
+    const checked = !this.state.checked
```
mnajdova, line 119: nit, could be moved in the if statement

packages/react/src/themes/base/components/Checkbox/checkboxVariables.ts
mnajdova: As we decided not to add the color scheme now and make the component red-lined, let's be sure to track this work

packages/react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,153 @@
...
+  isFromKeyboard,
+  UIComponentProps,
+} from '../../lib'
+import { ComponentEventHandler, WithAsProp, ShorthandValue, withSafeTypeForAs } from '../../types'
+import Icon from '../Icon/Icon'
+import Text from '../Text/Text'
+import { Accessibility } from '../../lib/accessibility/types'
+import { checkboxBehavior } from '../../lib/accessibility'
+
+export interface CheckboxProps extends UIComponentProps, ChildrenComponentProps {
```
mnajdova, line 21: Let's add conformat test for the new component.

packages/react/src/lib/accessibility/Behaviors/Checkbox/checkboxBehavior.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Adds role='checkbox'. This allows screen readers to handle the component as a checkbox button.
+ * Adds attribute 'aria-checked=true' based on the property 'checked'.
+ * Adds attribute 'tabIndex=0' to 'root' component's part.
+ */
+const checkboxBehavior: Accessibility = props => ({
+  attributes: {
+    root: {
+      'aria-checked': props.checked,
+      role: 'checkbox',
+      tabIndex: 0,
```
mnajdova, line 20: Just to be sure, we should not add any attribute if the checkbox is disabled?

packages/react/src/themes/teams/components/Checkbox/checkboxStyles.ts
```diff
@@ -0,0 +1,15 @@
+import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
+import { CheckboxProps, CheckboxState } from '../../../../components/Checkbox/Checkbox'
+import { CheckboxVariables } from './checkboxVariables'
+import getBorderFocusStyles from '../../getBorderFocusStyles'
+
+const checkboxStyles: ComponentSlotStylesInput<CheckboxProps & CheckboxState, CheckboxVariables> = {
+  root: ({ props: p, variables: v, theme: t }): ICSSInJSStyle => ({
+    ...getBorderFocusStyles({
+      siteVariables: t.siteVariables,
```
mnajdova, line 9: It seems that no focus styles are visible. The component looks exactly the same when focused:
image

packages/react/src/themes/base/components/Checkbox/checkboxVariables.ts
```diff
@@ -0,0 +1,77 @@
...
+  checkboxBorderRadius: pxToRem(4),
+  checkboxBorderWidth: pxToRem(1),
+  checkboxColor: 'transparent',
+  checkboxMargin: `0 ${pxToRem(12)} 0 0`,
+  checkboxPadding: '0',
+
+  toggleBackground: 'transparent',
+  toggleBorderColor: siteVars.colors.grey[300],
+  toggleBorderStyle: `solid`,
+  toggleBorderRadius: pxToRem(8),
```
mnajdova: nit: can we make this value bigger, it looks better that way:
before pxToRem(8)
image

after pxToRem(999)
image

docs/src/components/ExampleSnippet/renderElementToJSX.ts
layershifter: Yep, I thinking about it and how we can avoid this filtering at all

packages/react/src/lib/accessibility/Behaviors/Checkbox/checkboxBehavior.ts
layershifter: https://www.w3.org/TR/wai-aria-practices/examples/checkbox/checkbox-1/checkbox-1.html

I don't see this in the spec, so decided to skip for the initial implementation

packages/react/src/components/Checkbox/Checkbox.tsx
layershifter: Will keep as it is for now :)

packages/react/src/components/Checkbox/Checkbox.tsx
layershifter: Added, have to add an additional handler for `onChange`. Decided to keep it flat to be more obvious

packages/react/src/themes/teams/components/Checkbox/checkboxStyles.ts
layershifter: Fixed 👍

packages/react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,168 @@
...
+
+  renderComponent({ ElementType, classes, unhandledProps, styles, accessibility }) {
+    const { label, icon, toggle } = this.props
+
+    return (
+      <ElementType
+        className={classes.root}
+        onClick={this.handleClick}
+        onChange={this.handleChange}
+        onFocus={this.handleFocus}
```
mnajdova, line 143: Was this the thing missing for the focus styles? :)

layershifter: Yep :) No IDE highlight 😭

mnajdova (approved):

Issue #1392, opened 2019-05-25
Menu items as object array with variables loses correct styling


dslane:
### Steps
When rendering providing the `items` property of a Menu as an array of MenuItemProps objects, if you specify `variables` in those objects, the menu items get incorrect styling. This is most obvious in teamsDark theme.

The kind of issues varies between different Stardust versions. In 0.27.0, for example, the "selected" background color is the same as the normal background color, so you cannot tell which item is selected (using `pointing` helps, but the background color is still the same). Also, the text color of the hovered-on menu item is too dark, and is hard to see agains the hover background color.

In 0.31.0, the text color of the selected is the same as the background color of the selected item, so you cannot see the text of the selected item in the menu.

### Expected Result
The colors of the text and background should not be impacted by passing in a variable that does not alter those CSS properties.

### Actual Result
Passing any variables causes colors that make the menu much less usable 💣

### Version
x.y.z

### Testcase
[Fork, update, and replace this pen to show the bug]:
https://codesandbox.io/s/stardustuiexample-ihpcn

There are two side-by-side menus, with the only difference being that variables passed to in the MenuItem props on the first one.

PR #1408, opened 2019-05-29
feat(Toolbar): add component

miroslavstastny:
### Basic `Toolbar` implementation
image

Toolbar currently supports three kinds of items:
- `ToolbarItem` - the icon button in the toolbar,
- `ToolbarDivider` - visual separator between items,
- `ToolbarRadiogroup` - logical grouping of `ToolbarItem` (can also contain `ToolbarDivider`).

See https://www.w3.org/TR/wai-aria-practices/examples/toolbar/toolbar.html for details on accessibility.

- [x] unit tests
- [x] docsite example
- [x] teams, teamsDark, teamsHighContrast themes
- [ ] changelog

To be handled in separate tasks:
- menu - #1411
- popup - #1410
- accessibility - #1413
- overflow menu - #1409


packages/react/src/themes/teams-dark/components/Toolbar/toolbarVariables.ts
```diff
@@ -0,0 +1,11 @@
+import { ToolbarVariables } from '../../../teams/components/Toolbar/toolbarVariables'
+import { colors } from '../../../teams/colors'
+
+export default (siteVars: any): Partial<ToolbarVariables> => ({
+  foreground: siteVars.colors.grey[250],
+  dividerBorder: siteVars.colors.grey[450],
+  foregroundHover: siteVars.colors.brand[400],
+  foregroundActive: siteVars.colors.brand[400],
+  foregroundDisabled: colors.grey[450],
```
kuzhelov: shouldn't we consume all colors from `siteVariables`?

miroslavstastny: Yup, I plan to discuss the usage of color palette with @mnajdova and update the variables accordingly.

packages/react/src/themes/teams-high-contrast/components/Toolbar/toolbarVariables.ts
```diff
@@ -0,0 +1,16 @@
+import { ToolbarVariables } from '../../../teams/components/Toolbar/toolbarVariables'
+
+export default (siteVars: any): Partial<ToolbarVariables> => ({
+  foreground: siteVars.colors.white,
```
kuzhelov: just the thought unrelated to this PR - really think that we need to introduce `siteVariables`'s types for Teams theme - otherwise it might become to be pretty complicated story to maintain all this, especially given that name of the variable is not fixed on the client side

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  static defaultProps = {
+    as: 'button',
+    accessibility: defaultBehavior as Accessibility,
+  }
+
+  renderComponent({ ElementType, classes, unhandledProps, accessibility }) {
+    const { icon, children } = this.props
+    return (
+      <ElementType
```
miroslavstastny, line 92: Intentionally NOT setting `disabled` attribute - according to WAI ARIA disabled items should be focusable to make screen readers happy.

packages/react/src/themes/teams-dark/components/Toolbar/toolbarVariables.ts
mnajdova: As this is a new component, we can start with the using the color scheme, and do only the necessary overrides in the different theme, if there is any 👍

miroslavstastny: Switched to color scheme

packages/react/src/components/Toolbar/ToolbarItem.tsx
miroslavstastny: Setting `disabled` for now.

packages/react/src/themes/teams/componentVariables.ts
```diff
@@ -63,6 +63,8 @@ export { default as Status } from './components/Status/statusVariables'
 
 export { default as Text } from './components/Text/textVariables'
 
+export { default as Toolbar } from './components/Toolbar/toolbarVariables'
```
mnajdova, line 68: Can we export these for the other Toolbar components as well (as we are exporting the menuVariables for the MenuItem too)?

packages/react/src/themes/teams/components/Menu/menuVariables.ts
```diff
@@ -12,7 +12,6 @@ export const menuColorAreas = stringLiteralsArray(
   'backgroundHover',
   'backgroundActive',
   'backgroundFocus',
-  'foregroundFocus',
```
mnajdova, line 15: This is used in the `menuItemStyles.ts` why is it removed? Btw we should have typescript error in the menuItemStyles.ts after this change

packages/react/src/themes/teams/components/Toolbar/toolbarVariables.ts
```diff
@@ -0,0 +1,86 @@
...
+  'foregroundActive',
+
+  'foregroundHover',
+  'backgroundHover',
+
+  'foregroundFocus',
+  'backgroundFocus',
+  'borderFocus',
+
+  'foregroundDisabled1',
```
mnajdova, line 19: nit: we may want to extend the color scheme, and use the `foregroundDisabled1` for the `foregroundDisabled`, so it would be more user friendly, but then it may raise a problem because the user will think this one is acttually the `foregroundDisabled` from the color scheme... Let's just think about this and decide together whether we want that.

packages/react/src/components/Toolbar/Toolbar.tsx
```diff
@@ -0,0 +1,105 @@
...
+    accessibility: defaultBehavior,
+  }
+
+  static Item = ToolbarItem
+  static Divider = ToolbarDivider
+  static RadioGroup = ToolbarRadioGroup
+
+  handleItemOverrides = variables => predefinedProps => ({
+    variables: {
+      ...variables,
```
mnajdova, line 62: A bit confused why we don't have this in the MenuItem as well...

packages/react/src/themes/teams/components/Menu/menuVariables.ts
miroslavstastny: This one was in the list twice (line 11)

packages/react/src/components/Toolbar/Toolbar.tsx
miroslavstastny: We will - #1392

mnajdova (approved):

alinais (approved):

PR #1416, opened 2019-05-29
fix(Dropdown): call getDerivedStateFromProps() from AutoControlledComponent

### LINKED ISSUE #1412: Dropdown - Text not displayed in Input in Controlled mode
## Bug Report
When using the Dropdown component in controlled mode (passing `searchQuery` from the container), the text that the user types is not displayed in the input box.

### Steps
1. Install `v0.31` of Stardust.
2. In your component, keep the searchQuery in state.
3. Pass searchQuery state down to Dropdown component's `searchQuery` prop.
4. Set the searchQuery in state in the `onSearchQueryChange` callback.
5. Type in the Dropdown input box.

### Expected Result
Input should display the keystrokes typed by the user.

### Actual Result
Input not displaying the keystrokes typed by the user.

### Version
0.31.0

### Testcase
https://codesandbox.io/s/stardustdropdowncontrolledbug-q0fpp

layershifter:
Fixes #1412.

This issues is caused by inheritance of `AutocontrolledComponent` by `Dropdown`, we should its `getDerivedStateFromProps()` to properly compute props.


mnajdova:
Based on the screenshot differences it seems that we broke the multiple Dropdown variations. When something is picked in the multiple search, it should not appear in the list.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -338,25 +338,30 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
+    // state we should merge it with existing
+    const actualState: DropdownState = { ...state, ...autoState }
+
     const { items, itemToString, multiple, search } = props
-    const { value } = state
+    const { searchQuery, value } = actualState
 
     if (!items) {
-      return null
+      return autoState
```
mnajdova: But confused, why we are not returning `actualState`?

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -338,25 +338,30 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
       ? _.difference(items, value as ShorthandCollection)
       : items
 
     if (search) {
-      const { itemToString } = props
-      const { searchQuery } = state
-
       if (_.isFunction(search)) {
-        return { filteredItems: search(filteredItemsByValue, searchQuery) }
+        return { ...autoState, filteredItems: search(filteredItemsByValue, searchQuery) }
```
mnajdova: Same here, why not `actualState`?

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -1070,6 +1070,15 @@ describe('Dropdown', () => {
   })
 
   describe('searchQuery', () => {
+    it("updates component's state", () => {
```
kuzhelov: nit: would suggest to add `.. on props updates`, just to make the intent of test case more clear

kuzhelov (approved):

bmdalex (approved):

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -343,29 +343,33 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
   /**
    * Used to compute the filtered items (by value and search query) and, if needed,
    * their string equivalents, in order to be used throughout the component.
-   * @param props
-   * @param state
    */
-  static getDerivedStateFromProps(props: DropdownProps, state: DropdownState) {
+  static getDerivedStateFromProps(props: DropdownProps, currentState: DropdownState) {
+    // Due inheritance of AutoControlled component we should call its getDerivedStateFromProps(),
+    // to properly compute next state we should merge it with existing
```
mnajdova: >  // Due to the inheritance of the AutoControlledComponent we should call its getDerivedStateFromProps() and merge it with the existing state

mnajdova (approved):

PR #1418, opened 2019-05-29
fix(Accordion.Title): make contentRef prop optional

### LINKED ISSUE #1417: Accordion.Title contentRef prop is required
### Steps
Try use `Accordion` component with `Accordion.Title` subcomponent without specifying `contentRef` prop.

### Expected Result
It should be possible and should work without errors. 

### Actual Result
`contentRef` is a required prop but there is no reason for it. Stardust strives to create components that work out of the box with minimal props ➡️  `Accordion.Title` should work fine without specifying required props  💣

https://github.com/stardust-ui/react/blob/c69313cd008cf380d9ac2913342d0d403a398a05/packages/react/src/components/Accordion/AccordionTitle.tsx#L45

### Version
UNRELEASED

bmdalex:
### fix(Accordion.Title): make contentRef prop optional

Fixes #1417

### DIFF
--- a/packages/react/src/components/Accordion/AccordionTitle.tsx
+++ b/packages/react/src/components/Accordion/AccordionTitle.tsx
@@ -42,7 +42,7 @@ export interface AccordionTitleProps
   index?: number
 
   /** Ref to the clickable element that contains the title. */
-  contentRef: React.Ref<HTMLElement>
+  contentRef?: React.Ref<HTMLElement>
 
   /**
    * Called on click.
--- a/packages/react/src/components/Accordion/AccordionTitle.tsx
+++ b/packages/react/src/components/Accordion/AccordionTitle.tsx
@@ -86,6 +86,7 @@ class AccordionTitle extends UIComponent<WithAsProp<AccordionTitleProps>, any> {
   static defaultProps = {
     accessibility: accordionTitleBehavior,
     as: 'dt',
+    contentRef: _.noop,
   }
 
   actionHandlers: AccessibilityActionHandlers = {
--- a/packages/react/test/specs/components/Accordion/AccordionTitle-test.tsx
+++ b/packages/react/test/specs/components/Accordion/AccordionTitle-test.tsx
@@ -1,5 +1,3 @@
-import * as _ from 'lodash'
-
 import AccordionTitle from 'src/components/Accordion/AccordionTitle'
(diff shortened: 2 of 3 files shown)


kuzhelov (approved):

packages/react/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -42,7 +42,7 @@ export interface AccordionTitleProps
   index?: number
 
   /** Ref to the clickable element that contains the title. */
-  contentRef: React.Ref<HTMLElement>
+  contentRef?: React.Ref<HTMLElement>
```
silviuaavram, line 45: is it possible to also remove the `_.noop` from the tests?

silviuaavram (changes requested):

packages/react/test/specs/components/Accordion/AccordionTitle-test.tsx
```diff
@@ -8,26 +6,17 @@ describe('AccordionTitle', () => {
...
   })
 
   describe('accessiblity', () => {
     describe('header', () => {
       handlesAccessibility(AccordionTitle, {
-        requiredProps: {
-          as: 'h3',
-          contentRef: _.noop,
-        },
+        requiredProps: { as: 'h3' },
```
bmdalex, line 14: @silviuavram are these changes correct? also why is `requiredProps: { as: 'h3' },` needed?

packages/react/src/components/Accordion/AccordionTitle.tsx
bmdalex: done, pls check https://github.com/stardust-ui/react/pull/1418/files#r288565651

packages/react/test/specs/components/Accordion/AccordionTitle-test.tsx
silviuaavram: if it's a header it will not receive a role. you can see that the example below does not have the as prop as required, but the `handlesAccessibility` test will check that the role of `header` has been added.

silviuaavram (approved):

packages/react/test/specs/components/Accordion/AccordionTitle-test.tsx
bmdalex: got it, so the changes are good, thanks

layershifter (approved):

PR #93, opened 2018-08-15
feat(grid): base implementation

bmdalex:
# Grid

This PR will introduce the **Grid** component with the following props: 
- `columns=[number|string]`
- `rows=[number|string]`

The Grid will specify the layout of the elements it encapsulates as `children` or in the `content` prop.

See also:
- #11 - for description of what we're trying to accomplish
- https://learncssgrid.com/ and
https://css-tricks.com/snippets/css/complete-guide-grid/ - for CSS grid layout ramp-up

### TODO

- [x] Conformance test
- [x] Minimal doc site example
- [ ] Stardust base theme
- [x] Teams Light theme
- [ ] Teams Dark theme
- [ ] Teams Contrast theme
- [ ] Confirm RTL usage
- [ ] [W3 accessibility](https://www.w3.org/standards/webdesign/accessibility) check
- [ ] [Stardust accessibility](https://github.com/stardust-ui/accessibility) check
- [ ] Update glossary props table
- [x] Update the CHANGELOG.md

# API Proposal

## `columns`

We can specify a certain amount of columns:

columncount

or the explicit columns for a grid:

columnexplicit

```jsx
<Grid columns="7">
  <Image fluid src="img1.png" />
  <Image fluid src="img2.png" />
  <Image fluid src="img3.png" />
  ...
</Grid>
```

and

```jsx
<Grid columns="repeat(3, 1fr) 2fr 2fr 110px 14rem 50px 20%">
  <Image fluid src="img1.png" />
  <Image fluid src="img2.png" />
  <Image fluid src="img3.png" />
  ...
</Grid>
```

generate:

```html
<div class="ui-grid a b c d e f k">
  <img src="img1.png" class="ui-image h i b j" />
  <img src="img2.png" class="ui-image h i b j" />
  <img src="img3.png" class="ui-image h i b j" />
  ...
</div>
```

## `rows`

We can specify a certain amount of rows:

rowcount

or the explicit rows for a grid:

rowexplicit

```jsx
<Grid rows="2">
  <Image fluid src="img1.png" />
  <Image fluid src="img2.png" />
  <Image fluid src="img3.png" />
  ...
</Grid>
```

and

```jsx
<Grid rows="2fr repeat(2, 1fr)">
  <Image fluid src="img1.png" />
  <Image fluid src="img2.png" />
  <Image fluid src="img3.png" />
  ...
</Grid>
```

generate:

```html
<div class="ui-grid a b c d e f k">
  <img src="img1.png" class="ui-image h i b j" />
[...]

PR #1428, opened 2019-05-29
docs(Examples): fix Grid variables to fix gaps in Firefox

notandrew:
after:
image

before:
image


layershifter (commented):
Actually, this change is more than an update for because it affects also `base` theme and `teams` theme. These values were introduced in the initial PR #93, however I am not sure that we need them at all.

@Bugaa92 @kuzhelov can we safely remove them? See #1440 for Screener test run.

bmdalex:
> Actually, this change is more than an update for because it affects also `base` theme and `teams` theme. These values were introduced in the initial PR #93, however I am not sure that we need them at all.
> 
> @Bugaa92 @kuzhelov can we safely remove them? See #1440 for Screener test run.

@layershifter I cannot see the screener test results... I remember it was not safe to remove them because of issues with IE and Edge, did you test there?

layershifter:
> I cannot see the screener test results... 
https://screener.io/v2/dashboard/5b6d198a6b2944400b00fb25/try%2Fempty-values

bmdalex:
> > I cannot see the screener test results...
> > https://screener.io/v2/dashboard/5b6d198a6b2944400b00fb25/try%2Fempty-values

what about IE and Edge? @layershifter

notandrew:
@layershifter @Bugaa92  - not sure what to do with this PR...

codepretty:
@layershifter @levithomason do you know if we plan to have screener support for other browsers?

layershifter (approved):
Tested it in FF & Edge, works fine.
There are issues in IE11, but they related to CSS Grid module at all.

PR #1420, opened 2019-05-29
fix(Popup): proptypes for content and trigger

miroslavstastny:
Fixes #1419.

### Expected behavior (provided by PR's changes)
It is possible to use `content` + `children`.
It is not possible to use `trigger` + `children`.

PR #1430, opened 2019-05-30
fix(popup): revert proptypes for content and trigger fix

bmdalex:
Revert "fix(Popup): proptypes for content and trigger (#1420)"

This reverts commit bd714f42da3e8ccbba20f0a53b5d22a02ded88c9.


bmdalex:
not needed anymore

PR #1431, opened 2019-05-30
fix(popper): fix popper.js imports

bmdalex:
### Description

This PR fixes the following issue:

The `commonjs` bundle file generated for `packages/react/src/lib/positioner/Popper.tsx` is incorrect; it produces the following error:

`TypeError: popper_js_1.default is not a constructor`

### Fix

`popper.js` has a UMD build without `.default`, it breaks CJS builds. Therefore we need to conditionally check `.default` prop on `PopperJS` object when we create the popper instance:
```typescript
new ((_PopperJS as any).default || _PopperJS)(reference, popper, options)
```

### DIFF
--- a/packages/react/src/lib/positioner/Popper.tsx
+++ b/packages/react/src/lib/positioner/Popper.tsx
@@ -7,6 +7,8 @@ import { getPlacement, applyRtlToOffset } from './positioningHelper'
 import { PopperProps, PopperChildrenFn } from './types'
 import getScrollParent from './getScrollParent'
 
+// `popper.js` has a UMD build without `.default`, it breaks CJS builds:
+// https://github.com/rollup/rollup/issues/1267#issuecomment-446681320
 const createPopper = (
   reference: Element,
   popper: Element,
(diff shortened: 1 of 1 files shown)


packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -1,12 +1,18 @@
...
 import * as _ from 'lodash'
-import PopperJS from 'popper.js'
+import PopperJS, * as _PopperJS from 'popper.js'
 import { Ref } from '@stardust-ui/react-component-ref'
 
 import { getPlacement, applyRtlToOffset } from './positioningHelper'
 import { PopperProps, PopperChildrenFn } from './types'
 import getScrollParent from './getScrollParent'
 
+const createPopper = (
```
layershifter, line 12:
```suggestion
// `popper.js` has a UMD build without `.default`, it breaks CJS builds:
// https://github.com/rollup/rollup/issues/1267#issuecomment-446681320
const createPopper = (
```

layershifter (approved):

PR #1435, opened 2019-05-31
fix(Icon): revert changes with different roots

### LINKED ISSUE #1407: Icon backgroundColor not working
### Steps

```jsx
<Icon
   name="accept"
   circular
   styles={{ color: 'red', backgroundColor: 'green' }}
   size="smallest"
/>
```

### Expected Result

Background color is changed according to provided styling. This worked in 0.29.1

image

### Actual Result

Background color style did not propagate

image

### Version
0.31.0

### Testcase

layershifter:
Fixes #1407.

Reverts changes with multiple roots from #1337 (`svgRoot` & `fontRoot`). This idea breaks ability to override styles via `styles` prop on `Icon` component: `svgRoot` & `fontRoot` had more priority on merge. Also we used `Box` component there, it's redundant.

We definitely will want to separate these styles somehow to avoid conditions, but `svgRoot`/`fontRoot` is definitely bad stuff.

### DIFF
--- a/packages/react/src/themes/font-awesome/components/Icon/iconStyles.ts
+++ b/packages/react/src/themes/font-awesome/components/Icon/iconStyles.ts
@@ -1,11 +1,19 @@
-import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
+import { ComponentSlotStylesInput, ICSSInJSStyle, ThemeIconSpec } from '../../../types'
 import { IconProps } from '../../../../components/Icon/Icon'
 import { IconVariables } from '../../../teams/components/Icon/iconVariables'
+import { emptyIcon } from '../../../base/components/Icon/iconNames'
 
 const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
-  root: (): ICSSInJSStyle => ({
-    fontWeight: 900, // required for the fontAwesome to render
-  }),
+  root: ({ props: p, theme: t }): ICSSInJSStyle => {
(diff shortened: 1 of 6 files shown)


packages/react/src/themes/font-awesome/components/Icon/iconStyles.ts
```diff
@@ -3,7 +3,7 @@ import { IconProps } from '../../../../components/Icon/Icon'
 import { IconVariables } from '../../../teams/components/Icon/iconVariables'
 
 const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
-  fontRoot: (): ICSSInJSStyle => ({
+  root: (): ICSSInJSStyle => ({
```
mnajdova: Shuold we add this conditionally if the icon is font icon? Won't this break icons on docs?

mnajdova: I check that it doesn't break anything, so not sure if you should implement the comment...

bmdalex (approved):
👍

mnajdova (approved):

packages/react/src/themes/font-awesome/components/Icon/iconStyles.ts
kuzhelov: agree that it is better to introduce explicit condition check here, as we are expecting this rule to be applied exclusively to font-based icons

kuzhelov (approved):

PR #1438, opened 2019-06-03
feat: e2e tests

kuzhelov:
This PR introduces support for in-browser tests to the project.

## Command to run in-browser tests
`yarn test:e2e`

This command is also introduced as part of CI checks chain.

### Flags
`--skip-build` - disables prior (re)build of Docs site which end-to-end tests are running against.

## How things work
All the test files reside under `e2e/tests` directory of the project. These test files are identical to the ones already written for unit tests, with only difference is that `e2e` global object is provided as an entry point for browser API. 

**This global `e2e` object serves as a thin wrapper on top of `puppeteer's` API, and should be used as a single client's entry point to simulate browser stuff: clicks, focus and keyboard events.**

## Test cases introduced
Couple of `Dropdown` test cases that were not possible to cover with unit tests (from https://github.com/stardust-ui/react/pull/1201 - both are related to focus behavior scenarios) were introduced.

## Implementation considerations
Tests are optimised at maximum degree - thin wrapper (`puppeteer`) on top of Chromium API is used, as well as performant Jest runner. In comparison to Cypress-based implementation: there is a 7-10 times improvement of tests' running times, as well as full set of API capabilities provided by `puppeteer` to simulate browser scenarios.

## Other improvements
Duplicated logic used for gulp tasks has been fetched and refactored:
- `serve` module is introduced to encapsulate imperative code of scaffolding `express` server
- `gulp-jest` plugin is introduced to easily embed Jest tests into Gulp chains


package.json
```diff
@@ -130,6 +131,7 @@
     "syncpack": "^4.3.1",
     "through2": "^2.0.3",
     "tmp": "^0.0.33",
+    "ts-jest": "^24.0.2",
```
kuzhelov: used to transpile code of jest tests - should be removed after fixes for Babel config will be merged @layershifter

e2e/e2eApi.ts
```diff
@@ -0,0 +1,74 @@
...
+import config from '../config'
+
+const serverUrl = `http://${config.server_host}:${config.e2e_port}`
+const launchOptions = {
+  dumpio: false,
+  headless: true,
+  slowMo: 10,
+
+  // Workaround for newPage hang in CircleCI: https://github.com/GoogleChrome/puppeteer/issues/1409#issuecomment-453845568
+  args: [process.env.CI && '--single-process'].filter(Boolean),
```
layershifter: We already have `'--single-process'` in two files, can we move this to a shared config?

e2e/e2eApi.ts
```diff
@@ -0,0 +1,74 @@
...
+  private browser: Browser
+  private page: Page
+
+  public beforeAll = async () => {
+    this.browser = await launch(launchOptions)
+  }
+
+  public beforeEach = async () => {
+    this.page = await this.browser.newPage()
+  }
```
layershifter: As I see we use `browser` only twice: to create/to clsoe pages.

My proposal is to remove `before*` & `after*` logic and pass `page` directly to class constructor. Logic of creation `browser` and `page` will be moved to `setup.test.ts`. As bonus: we will not need to separate one API from other via types (`type E2EApi = Pick<>`).

kuzhelov: I would rather like to follow Jest semantics with the methods exposed - so that this API will have the same set of methods provided, and it will be straightforward to see where the logic should be put

e2e/e2eApi.ts
kuzhelov: yes - actually, now it is three places (so it was two before), and this is the reason I've followed the pattern :) will address it

e2e/e2eApi.ts
kuzhelov: but I see your point - let me try this..

kuzhelov: yeah - the problem is that logic of initializing `browser` is async - thus we cannot put it to `setup.test.ts` module's initialiser code. And we cannot put it `beforeAll` either, as that will mean that we have to create `global` `e2e` object there - and this is not correct to assign globals in these methods. Of course, we might involve crazy reference-based logic, but I see that it will be far much less obvious and maintainable code.

Thus, lets just have `e2e` object being synchronously created, and make all initialization/uninitialization logic (async one) happening in `beforeAll`/`afterAll` methods - the ones that are designed to handle these aspects.

layershifter: https://jestjs.io/docs/en/setup-teardown#repeating-setup-for-many-tests
Why we can't create a `browser` in async?

> and this is not correct to assign globals in these methods.

Why? Can you point me what is wrong with code below?

```jsx
let browser: Browser
let page: Page

beforeAll(async () => {
  browser = await launch(launchOptions)
})
beforeEach(async () => {
  page = await browser.newPage()
  global['e2e'] = new E2EApiClass(page)
})

afterEach(async () => {
  await page.close()
  global['e2e'] = null
})
afterAll(async () => {
  await browser.close()
})
```

Currently `E2EApiClass` breaks single responsibility and it's obvious because you have `E2EApi` that has another signature. I expect that this class will only provide API, currently it does everything and knows about Jest.

kuzhelov: done

layershifter (approved):

PR #1444, opened 2019-06-03
fix(attachment): action click

### LINKED ISSUE #1334: Attachment's action icon button onclick isn't working correctly when attachment is focused
### Steps
1. Click on the action icon in the attachment and notice that the action's onClick event is fired.
2. Focus the Attachment, then click on the action icon and notice that the attachment's onClick event gets fired. 

### Expected Result
I expect that the action's onClick event is fired whenever it is clicked on, regardless of it the Attachment is focused or not.

### Actual Result
The Attachment onClick event is fired when clicking the action icon when the attachment is focused.

attachment-onClick-bug

bmdalex:
### Description

This PR fixes #1334 

Steps to reproduce issue:

1. Click on the action icon in the attachment and notice that the action's onClick event is fired.
2. Focus the Attachment, then click on the action icon and notice that the attachment's onClick event gets fired.

### Fix

Add `pointerEvents: 'none'` to `getBorderFocusStyles.ts` style helper in order ignore pointer events for the pseudo elements.

### BEFORE
The Attachment onClick event is fired when clicking the action icon when the attachment is focused.

attachment-onClick-bug

### AFTER
Screen Recording 2019-06-03 at 14 43 59

### DIFF
--- a/packages/react/src/themes/teams/getBorderFocusStyles.ts
+++ b/packages/react/src/themes/teams/getBorderFocusStyles.ts
@@ -22,6 +22,7 @@ const getPseudoElementStyles = (args: BorderPseudoElementStyles): ICSSInJSStyle 
     content: '""',
     position: 'absolute',
     borderStyle: 'solid',
+    pointerEvents: 'none',
     top: borderEdgeValue,
     right: borderEdgeValue,
     bottom: borderEdgeValue,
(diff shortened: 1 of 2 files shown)


layershifter (approved):

packages/react/src/themes/teams/getBorderFocusStyles.ts
```diff
@@ -22,6 +22,7 @@ const getPseudoElementStyles = (args: BorderPseudoElementStyles): ICSSInJSStyle
     content: '""',
     position: 'absolute',
     borderStyle: 'solid',
+    pointerEvents: 'none',
```
jurokapsiar, line 25: @kolaps33 please test this with screen readers

kolaps33 (commented):
tested with VoiceOver looks ok :)

PR #1445, opened 2019-06-03
feat(attachmentBehavior): add "data-is-focusable" for attachment

### LINKED ISSUE #1397: Accessibility - add "data-is-focusable" for attachment behavior
### Problem description
when attachment component is used in bigger scope, as in chat message, then "focus zone" doesn't navigate into attachment

### Proposed solution
add attribute "data-is-focusable" for Attachment behavior 
or investigate which "attachment" component is not focusable when is part of message

sophieH29:
closes https://github.com/stardust-ui/react/issues/1397

When attachment component is used in a bigger scope, as in chat message, then "focus zone" doesn't navigate into attachment


packages/react/src/lib/accessibility/Behaviors/Attachment/attachmentBehavior.ts
```diff
@@ -10,6 +10,7 @@ const attachmentBehavior: Accessibility = (props: any) => ({
   attributes: {
     root: {
       tabIndex: 0,
+      'data-is-focusable': true,
```
layershifter: As I remember we're using `IS_FOCUSABLE_ATTRIBUTE`:

```suggestion
      [IS_FOCUSABLE_ATTRIBUTE]: true,
```

sophieH29: truee, thanks! 💃

layershifter (approved):

bmdalex (approved):

PR #1446, opened 2019-06-03
feat(Label): integrate with Dropdown

silviuaavram:
Replacing https://github.com/stardust-ui/react/pull/1372 with this approach.

Since both the triggerButton/searchInput need to be labelled, I passed the id of the label to the Dropdown via `labelId` and handled it there.

Added example of usage.


docs/src/examples/components/Form/Types/FormExampleDropdown.tsx
```diff
@@ -0,0 +1,31 @@
...
+    label: { content: `Your best friend's name is:`, id: labelId },
+    name: 'chooseFriend',
+    key: 'choose-friend',
+    required: true,
+    control: {
+      as: Dropdown,
+      items: ['John Doe', 'Dohn Joe', 'John Joe', 'Dohn Doe'],
+      labelId,
+      search: true,
+      placeholder: 'Choose friend',
```
sophieH29: nit: choose a friend

sophieH29 (approved):

docs/src/examples/components/Form/Types/FormExampleDropdown.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react'
```
bmdalex, line 1: pls change extension to `.shorthand.tsx` and also add the regular Children API example file that will use the current name of the file

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -524,16 +531,18 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
 
     const triggerButtonProps = getToggleButtonProps({
       onFocus: this.handleTriggerButtonOrListFocus,
       onBlur: this.handleTriggerButtonBlur,
       onKeyDown: e => {
         this.handleTriggerButtonKeyDown(e, rtl)
       },
-      'aria-label': content,
+      'aria-label': undefined,
+      'aria-labelledby': `${labelId} ${triggerButtonId}`,
```
bmdalex: so the trigger button is labeled by the label id and himself? 😕 
Also, why wasn't `'aria-labelledby': triggerButtonId` even before this change?

bmdalex: Also, we don't handle this case for the trigger input at all? Inputs don't have `aria-labelledby`?

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -125,6 +125,8 @@ export interface DropdownProps
    */
   itemToString?: (item: ShorthandValue) => string
 
+  labelId?: string
```
bmdalex: do we need this prop? can't we just use 'aria-labelledby' attribute directly?

It also misses documentation

docs/src/examples/components/Form/Types/FormExampleDropdown.tsx
```diff
@@ -0,0 +1,31 @@
...
+      placeholder: 'Choose friend',
+    },
+  },
+  { control: { as: Button, content: 'Submit' }, key: 'submit' },
+]
+
+const FormExample = () => (
+  <Form
+    onSubmit={() => {
+      alert('Form submitted')
```
bmdalex, line 9: nit: inline function to save 2 lines

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: 1. That's what ARIA says. https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html
1.1 Because we used `aria-label` instead. Now we've corrected it.
2. Downshift handles that by default. That is why it is passed to it as prop.

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: Oups, will add documentation!

Maybe because we need to compute `aria-labelledby` from that (in case of the trigger button, we also add its own id). Also because it is used in two places: button/input and ul. I think it's better this way. We will have it computed in the behavior in the future. Will ask for more opinions on this.

bmdalex: ok, let's see what others say
@layershifter @kuzhelov @mnajdova

packages/react/src/components/Dropdown/Dropdown.tsx
bmdalex: thanks for clarifying

docs/src/examples/components/Form/Types/FormExampleDropdown.shorthand.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react'
+import { Form, Dropdown, Button } from '@stardust-ui/react'
+
+const labelId = 'choose-friend-label'
+const fields = [
+  {
+    label: { content: `Your best friend's name is:`, id: labelId },
+    name: 'chooseFriend',
+    key: 'choose-friend',
+    required: true,
```
layershifter: Please remove it, it passes `required` to `div`:
```html
<div class="ui-dropdown ui-box" required="" name="chooseFriend" />
```

docs/src/examples/components/Form/Types/FormExampleDropdown.shorthand.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react'
+import { Form, Dropdown, Button } from '@stardust-ui/react'
+
+const labelId = 'choose-friend-label'
+const fields = [
+  {
+    label: { content: `Your best friend's name is:`, id: labelId },
```
layershifter, line 7: I am surprised with generated markup, is it valid? I expect:
```html
<label for="favoritefood">Favorite food</label> <!-- <<== `for` is used -->
<select name="favoritefood" id="favoritefood">
  <option>Cheese</option>
</select>
```

We have:

```html
<label class="ui-text" dir="auto" id="choose-friend-label">Your best friend&#x27;s name is:</label>
<div class="ui-dropdown ui-box" required="" name="chooseFriend" />
```

And if we will pass `id` on root:

```html
 <label class="ui-text" dir="auto" for="choose-friend-label" id="choose-friend-label">
```

layershifter (commented):
I don't have better ideas than `labelId` right now. Please also update PR description: it integrates `Dropdown` and `Form.Field`.

docs/src/examples/components/Form/Types/FormExampleDropdown.shorthand.tsx
layershifter: About `label` & `htmlFor`: https://codesandbox.io/s/dark-microservice-47wft

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -526,14 +528,16 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
+    const triggerButtonId = triggerButton['id'] || this.defaultTriggerButtonId
 
     const triggerButtonProps = getToggleButtonProps({
       onFocus: this.handleTriggerButtonOrListFocus,
       onBlur: this.handleTriggerButtonBlur,
       onKeyDown: e => {
         this.handleTriggerButtonKeyDown(e, rtl)
       },
-      'aria-label': content,
+      'aria-label': undefined,
```
jurokapsiar, line 539: if somebody provides `aria-label` should not we apply it to the trigger or to the input as well? the same for aria-describedby

silviuaavram: I don't see it as being suggested by ARIA, and if the user wants to do his own magic with label and describedby he can always use the slots.

But if you think we should handle it by default then we can create a new task.

PR #1447, opened 2019-06-03
fix(MenuItem): Merge Menu and MenuItem variables

### LINKED ISSUE #1392: Menu items as object array with variables loses correct styling

miroslavstastny:
## Problem description
`Menu` and all its `MenuItem`s share the same variables.
`Menu` passes its variables down to `MenuItem` as `defaultProps`. That means that if a consumer defines a variable directly on a `MenuItem`, it overwrites the variables passed from the `Menu`:

```jsx
<Menu items={[
  {content: 'valid item'}, // ok, variables passed from Menu to the item
  {content: 'item', variables: {special: true}} // BUG, item variables replaces Menu variables, this menu item has broken styles
]} />
```

## Solution
Instead of Menu passing its variables via `defaultProps`, it now does it through `overrideProps` and shallow-merges Menu variables with MenuItem variables.

- [ ] changelog

Fixes #1392.


packages/react/test/specs/components/Menu/Menu-test.tsx
```diff
@@ -118,6 +118,38 @@ describe('Menu', () => {
       })
     })
 
+    describe('variables', () => {
+      it('are passed from Menu to MenuItem and correctly merged', () => {
```
mnajdova: Let's mention the MenuDivider in the test as well: 'are passed from Menu to MenuItem and MenuDivider and correctly merged'

mnajdova (approved):

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -141,6 +141,17 @@ class Menu extends AutoControlledComponent<WithAsProp<MenuProps>, MenuState> {
...
+    variables: {
+      ...variables,
+      ...predefinedProps.variables,
+    },
+  })
+
+  handleDividerOverrides = variables => predefinedProps => ({
+    variables: {
+      ...variables,
+      ...predefinedProps.variables,
```
kuzhelov: note that `predefinedProps.variables` may be a function (for example, if client will provide value for `variables` as a function) - in that case shallow merge logic will fail. Would suggest to reuse `mergeComponentVariables` function here - this will ensure that all `variables` formats that Stardust is/will be able to handle will be handled properly here as well, without any code changes necessary

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -164,6 +175,9 @@ class Menu extends AutoControlledComponent<WithAsProp<MenuProps>, MenuState> {
     const itemsCount = _.filter(items, item => getKindProp(item, 'item') !== 'divider').length
     let itemPosition = 0
 
+    const itemOverridesFn = this.handleItemOverrides(variables)
```
kuzhelov: nit: may suggest to rename these to something like `overrideItemProps` and `overrideDividerProps` - think that it will make code easier to read and deduce that it is actually _props_ that are meant to be overriden

packages/react/src/components/Menu/Menu.tsx
```diff
@@ -124,7 +124,7 @@ class Menu extends AutoControlledComponent<WithAsProp<MenuProps>, MenuState> {
   static Item = MenuItem
   static Divider = MenuDivider
 
-  handleItemOverrides = predefinedProps => ({
+  handleItemOverrides = variables => predefinedProps => ({
```
kuzhelov, line 130: may suggest to remove `functionFactory -> function` relationship, as well as potential struggling points for the code reader while trying to understand what each function does given its name only.

We may remove currying altogether, and leave just two functions with `override` semantics - one for `menuItem`, another for `divider`. Will take `menuItem` as an example:

```jsx
// was: handleItemOverrides = variables => predefinedProps => ({
overrideItemProps = (predefinedProps, variables) => ({ ... })
```

And then, at the place where is used:
```jsx
// renderItems() method, we have 'variables' in scope here
MenuItem.create(item, {
    defaultProps: {...}
    overrideProps: props => this.overrideItemProps(props, variables),
})
```

layershifter: Can't agree with this, `Dropdown` component does the same thing and it looks obscure when `props` are passed in different order. Benefit of the factory: it returns the function with the same signature as `overrideProps` are defined.

kuzhelov: Lets focus on this specific example - does suggestion make things obscure? I think that not.

I would say it is much more obscure when we have several abstraction/indirection layers for, essentially, the same thing - and with the proposed approach we would have only one function, which is explicit on the input data necessary and _is not dependent on the `overrideProps` signature_ - which, in fact, it shouldn't be dependent on. 

I would see benefits of currying if this function will be reused in multiple places, but it is not - thus, with current example cannot see any reason for this complication.

miroslavstastny: Discussed offline. Let's leave it as it is (for now) to be consistent with other components (Dropdown)

PR #1448, opened 2019-06-04
fix(ListItem): end media in Teams theme breaks ButtonGroup styles

### LINKED ISSUE #1287: ButtonGroup style breaking when inside endMedia of ListItem
## Bug Report
ButtonGroup style breaking when inside endMedia of ListItem.

### Expected Result
The buttons inside ButtonGroup should stay inline

### Actual Result
The buttons inside ButtonGroups are positioned in a column
onHover

### Version
x.y.z

### Testcase
https://codesandbox.io/embed/0qp5yvj4ql

mnajdova:
Fix https://github.com/stardust-ui/react/issues/1287

ButtonGroup style were broken when inside endMedia of ListItem.

**Before:**
image

**After:**
image

### DIFF
--- a/packages/react/src/themes/teams/components/List/listItemStyles.ts
+++ b/packages/react/src/themes/teams/components/List/listItemStyles.ts
@@ -100,6 +100,7 @@ const listItemStyles: ComponentSlotStylesInput<ListItemPropsAndState, any> = {
   }),
   endMedia: ({ props: p }) => ({
     ...(p.selectable && { display: 'none' }),
+    flexShrink: 0,
   }),
   main: () => ({
     minWidth: 0, // needed for the truncate styles to work
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixes
 - Fix click handling on focus for `action` slot in `Attachment` component @Bugaa92 ([#1444](https://github.com/stardust-ui/react/pull/1444))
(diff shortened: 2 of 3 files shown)


packages/react/src/themes/teams/components/List/listItemStyles.ts
```diff
@@ -100,6 +100,7 @@ const listItemStyles: ComponentSlotStylesInput<ListItemPropsAndState, any> = {
   }),
   endMedia: ({ props: p }) => ({
     ...(p.selectable && { display: 'none' }),
+    flexShrink: 0,
```
bmdalex, line 103: I'm still not sure if the List behaves correctly... Pls. take a look at this screenshot:

Screen Recording 2019-06-04 at 14 21 49

here is the example I used:

```jsx
import * as _ from 'lodash'
import { List, Button } from '@stardust-ui/react'

const items = [
  {
    key: 'ftp',
    endMedia: (
      <Button.Group
        buttons={_.times(20).map(i => ({
          key: 'accept' + i,
          icon: 'accept',
          iconOnly: true,
          text: true,
        }))}
      />
    ),
    content:
      'Use the online FTP application to input the multi-byte application!Use the online FTP application to input the multi-byte application!',
  },
]

const ListExample = () => <List items={items} selectable />

export default ListExample
```

mnajdova: The `media` and `endMedia` should have width as the content inside them. The `content` between them is suppose to grow in order to fill the list item. If we follow this, I think your example is still correct.

kuzhelov (approved):

PR #1449, opened 2019-06-04
fix(dialogBehavior): move id generation to component

layershifter:
### Problem 1

IDs in `dialogBehavior` are generated on each render. If we will try to have some state there to persist there IDs it will break the idea that all accessibility behaviors are pure functions.

### Problem 2

Abstraction with slots leaked to components because we accessed there `header['id']` and `content['id']`.

### Solution

Move IDs generation to component and keep them in state, IDs will be stored for each slot separately (`headerId`, `contentId`) so Stardust slots will not leak.


packages/react/src/lib/AutoControlledComponent.tsx
```diff
@@ -169,14 +178,19 @@ export default class AutoControlledComponent<P = {}, S = {}> extends UIComponent
...
+      ...state,
+      ...initialAutoControlledState,
+      autoControlledProps,
+      getAutoControlledStateFromProps,
+    }
   }
 
   static getDerivedStateFromProps(props, state) {
-    const { autoControlledProps } = state
+    const { autoControlledProps, getAutoControlledStateFromProps } = state
```
layershifter: The single downside that I can't access actual `getAutoControlledStateFromProps()` by referencing `this` because it's static. Has anyone better idea than keep it in the state?

packages/react/src/lib/accessibility/Behaviors/Dialog/dialogBehavior.ts
```diff
@@ -15,8 +14,6 @@ import * as _ from 'lodash'
  * Adds attribute 'aria-labelledby' based on the property 'aria-labelledby' to 'popup' component's part.
  * Adds attribute 'aria-describedby' based on the property 'aria-describedby' to 'popup' component's part.
  * Adds attribute 'role=dialog' to 'popup' component's part.
- * Generates unique ID and adds it as attribute 'id' to the 'header' component's part if it has not been provided by the user.
- * Generates unique ID and adds it as attribute 'id' to the 'content' component's part if it has not been provided by the user.
```
layershifter, line 18: This definitely should be replaced. @silviuavram @sophieH29 can you help me with rules?

CHANGELOG.md
```diff
@@ -29,6 +29,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Fixes
 - Fix click handling on focus for `action` slot in `Attachment` component @Bugaa92 ([#1444](https://github.com/stardust-ui/react/pull/1444))
 
+### Fixes
+- Generate IDs for `header` and `content` slots once ([#1449](https://github.com/stardust-ui/react/pull/1449))
```
mnajdova: Let's add for which component this refers to :)

packages/react/src/lib/AutoControlledComponent.tsx
```diff
@@ -169,14 +178,19 @@ export default class AutoControlledComponent<P = {}, S = {}> extends UIComponent
       return acc
     }, {})
 
-    this.state = { ...state, ...initialAutoControlledState, autoControlledProps }
+    this.state = {
+      ...state,
+      ...initialAutoControlledState,
+      autoControlledProps,
+      getAutoControlledStateFromProps,
```
mnajdova: Isn't this a method? Should we invoke it to get the object containing props?

packages/react/src/lib/AutoControlledComponent.tsx
mnajdova: I got confused, where this method comes from? Why are we extracting it form the state?

packages/react/src/lib/AutoControlledComponent.tsx
layershifter: https://github.com/stardust-ui/react/pull/1449#discussion_r290287824

It's an initial state 🤔 We store it in state and will be able to access it inside of `static`.

packages/react/src/lib/AutoControlledComponent.tsx
layershifter: Let me check this part again: https://codesandbox.io/s/elastic-framework-w3kfk

I expect that this should work properly...

packages/react/src/lib/AutoControlledComponent.tsx
mnajdova: Let's describe from user perspective how these methods should be defined and used in the PR description. I am a bit confused as why we are introducing this method, and how it should be used.

packages/react/src/lib/AutoControlledComponent.tsx
layershifter: React calls it without `this`, that's why it's not working:
https://github.com/facebook/react/blob/2b93d686e359c7afa299e2ec5cf63160a32a1155/packages/react-test-renderer/src/ReactShallowRenderer.js#L560

packages/react/src/lib/accessibility/Behaviors/Dialog/dialogBehavior.ts
silviuaavram: Maybe you can refactor the `Dialog-test.tsx` to check that IDs are generated there if they are not provided. I think you can remove the other tests (label, labelledby etc) since they should be checked in the behavior.

In `dialogBehavior-test.tsx` you can add individual tests for the generation. Pass `header` as a shorthand with `id` and check it's added as `aria-labelledby`. Similar to `content`. The other case when they are applied directly from the `aria-labelledby` / `aria-describedby` should already be tested by a rule.

After looking at this file at `'aria-labelledby': defaultAriaLabelledBy || props['aria-labelledby'],` probably you can remove the `|| props['aria-labelledby']` and do the whole logic in the method, which you can rename without `Default`: `getAriaDescribedBy` or something. Where you return it as the prop value, if passed, as header['id] if header exists, or nothing.

silviuaavram: what do you think @sophieH29 ?

packages/react/src/lib/AutoControlledComponent.tsx
layershifter: Update PR description

packages/react/src/lib/accessibility/Behaviors/Dialog/dialogBehavior.ts
layershifter: Updated UTs 🔍

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -142,7 +144,21 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
   triggerRef = React.createRef<HTMLElement>()
 
   getInitialAutoControlledState(): DialogState {
-    return { open: false }
+    return {
+      contentId: _.uniqueId('dialog-content-'),
```
mnajdova: Can we store this strings to some statics, or reuse the classnames if they exists?

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -88,6 +88,8 @@ export interface DialogProps
 }
 
 export interface DialogState {
+  contentId?: string
```
mnajdova, line 111: Shouldn't we use these as an id of the elements, or they are used only for the aria-* attributes?

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -142,7 +144,21 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
...
+  }
+
+  static getAutoControlledStateFromProps(
+    props: DialogProps,
+    state: DialogState,
+  ): Partial<DialogState> {
+    return {
+      contentId: (props.content && props.content['id']) || state.contentId,
+      headerId: (props.header && props.header['id']) || state.headerId,
+    }
```
mnajdova, line 179: I really like the changes. It would be great if we can extract this logic, as I think it may be reused in other components as well, of course not as part of this PR.

mnajdova (approved):

packages/react/src/components/Dialog/Dialog.tsx
layershifter: I don't there a big benefit because we don't want to expose them to clients

packages/react/src/components/Dialog/Dialog.tsx
layershifter: I decided to keep responsibility for this on accessibility behaviors: https://github.com/stardust-ui/react/pull/1449/files#diff-7d8c4d117a2402a4e3fbb997e3c6693fL32

packages/react/src/components/Dialog/Dialog.tsx
layershifter: Yep, we should inspect other cases with leaked React abstractions

packages/react/src/lib/accessibility/Behaviors/Dialog/dialogBehavior.ts
```diff
@@ -39,41 +39,33 @@ const dialogBehavior: Accessibility<DialogBehaviorProps> = props => {
...
 export default dialogBehavior
 
 type DialogBehaviorProps = {
-  header?: {
-    id?: string
-  }
-  content?: {
-    id?: string
-  }
+  header?: string | object
```
kuzhelov: are `header` and `content` props used anywhere in the behavior's logic? I mean, even from semantical point of view as a client I wouldn't expect them to be necessary to calculate accessibility attributes - am I missing something?

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -142,7 +144,21 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
...
+      open: false,
+    }
+  }
+
+  static getAutoControlledStateFromProps(
+    props: DialogProps,
+    state: DialogState,
+  ): Partial<DialogState> {
+    return {
+      contentId: (props.content && props.content['id']) || state.contentId,
```
kuzhelov:
```
props.content && props.content['id']
```

what if React element is passed to any of these slots, with `id` provided? As I see, in that case we will silently override client's ID value - this is definitely what client expects in this situation. 

As we've committed to handle the props calculation for React elements provided to slots, we should handle this case as well.

packages/react/src/lib/accessibility/Behaviors/Dialog/dialogBehavior.ts
layershifter: `header` & `content` are used in `getDefaultAriaLabelledBy` & `getDefaultAriaDescribedBy` logic there is quite complicated.

I will introduce `hasContent` & `hasHeader` instead, it has more sense.

packages/react/src/components/Dialog/Dialog.tsx
layershifter: Good catch, will rework it

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -23,6 +23,26 @@ import Header from '../Header/Header'
 import Portal from '../Portal/Portal'
 import Flex from '../Flex/Flex'
 
+const getOrGenerateIdFromShorthand = (
```
kuzhelov, line 26: 👍 probably, in future we will need to move this function to `lib`, as the logic of `peek if client had specified some prop for shorthand value` seems to be a quite common case

kuzhelov: may just suggest to cover React element's case for shorthand slot in tests

kuzhelov (approved):

packages/react/src/components/Dialog/Dialog.tsx
layershifter: Was done in https://github.com/stardust-ui/react/pull/1449/commits/e816e3c8f5cc45f1cb62cea070c93ea9d7dcdd9c

silviuaavram (approved):

PR #1451, opened 2019-06-04
feat(Tree): left and right arrow keyboard navigation

silviuaavram:
Follows the ARIA design pattern https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-1/treeview-1a.html

On Left Arrow, it should collapse tree if it is open. If it is closed, it should move focus to parent.
On Right Arrow, it should expand tree if it is closed and has subtree. If it is open, it should move focus to first child in subtree.

Also removed circular navigation provided by the FocusZone.

WIP: Unit Tests.


packages/react/src/components/Tree/Tree.tsx
```diff
@@ -151,6 +158,18 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
         {childrenExist(children) ? children : this.renderContent()}
       </ElementType>
     )
+
+    return contentRef ? (
+      <Ref
```
sophieH29: what if you wrap the Tree comp with a Ref in TreeItem? So then 
- you render <Ref> only for subtree
- no need to make this check here
- you don't have to pass a ref prop. You get a ref in TreeItem

packages/react/src/lib/accessibility/Behaviors/Tree/treeBehavior.ts
```diff
@@ -14,7 +13,7 @@ const treeBehavior: Accessibility = (props: any) => ({
   focusZone: {
     mode: FocusZoneMode.Embed,
     props: {
-      isCircularNavigation: true,
+      isCircularNavigation: false,
```
sophieH29: it is false by default. Just remove it

packages/react/src/lib/accessibility/Behaviors/Tree/treeTitleBehavior.ts
```diff
@@ -21,6 +23,12 @@ const treeTitleBehavior: Accessibility = (props: any) => ({
       performClick: {
         keyCombinations: [{ keyCode: keyboardKey.Enter }, { keyCode: keyboardKey.Spacebar }],
       },
+      expandOrFocusFirstChild: {
+        keyCombinations: [{ keyCode: keyboardKey.ArrowRight }],
+      },
+      collapseOrFocusParent: {
+        keyCombinations: [{ keyCode: keyboardKey.ArrowLeft }],
+      },
```
sophieH29, line 31: do you need to handle these in tree title? I think you should be able to handle these cases in 1 place -  TreeItem

silviuaavram: On left/right when I expand/collapse, I am invoking the `onClick` handler. The only way to get the whole event chain (`Tree`, `TreeItem`, `TreeTitle`) is from `TreeTitle`, otherwise I may need to do another props-callback magic.

I would rather do the collapse/expand here and let `TreeItem` to handle the focus. What do you think?

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -102,10 +102,14 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
   }
 
   computeNewIndex = (index: number) => {
-    const { exclusive } = this.props
+    const activeIndexes = this.getActiveIndexes()
+    if (!this.props.items[index]['items']) {
```
bmdalex: nit: let's have consistency in destructuring (line 106 vs line 110)

bmdalex: can an item be undefined?
also, can you cast `items[index]` to `TreeItemProps`? In this way we're protected in case the API for `TreeItem` changes when we access `items[index].items`

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -99,7 +102,29 @@ class TreeItem extends UIComponent<WithAsProp<TreeItemProps>> {
...
+  actionHandlers = {
+    getFocusFromParent: e => {
+      const { open } = this.props
+      if (open) {
+        e.stopPropagation()
+        this.titleRef.current.focus()
+      }
+    },
+    setFocusToFirstChild: e => {
+      if (open) {
```
bmdalex: nit: nested ifs are a bit ugly

```typescript
if (!open) {
  return
}

e.stopPropagation()

const element = getFirstFocusable(this.treeRef.current, this.treeRef.current, true)
if (element) {
  element.focus()
}
```

packages/react/test/specs/components/Tree/Tree-test.tsx
```diff
@@ -1,7 +1,127 @@
-import { isConformant } from 'test/specs/commonTests'
+import * as React from 'react'
```
bmdalex: nice test suite 👍

packages/react/test/specs/components/Tree/Tree-test.tsx
```diff
@@ -1,7 +1,127 @@
...
+  describe('activeIndex', () => {
+    it('should contain index of item open at click', () => {
+      const wrapper = mountWithProvider(<Tree items={items} />)
+      const tree = wrapper.find(Tree).at(0)
+
+      wrapper
+        .find(`.${TreeTitle.className}`)
+        .at(0) // title 1
+        .simulate('click')
+      expect(tree.state('activeIndex')).toEqual([0])
```
bmdalex: if the component implementation changes (let's say we change name of `activeIndex` state property, reimplement with hooks or use state manager - which will happen soon) then the tests will need to change too;

can you reimplement the expectations just by searching through the DOM elements the open / displayed `TreeItem` elements?

packages/react/src/components/Tree/Tree.tsx
silviuaavram: no, it shouldn't be undefined, since this is called by an item click handler.
will refactor this a bit, you can take a look.

packages/react/test/specs/components/Tree/Tree-test.tsx
silviuaavram: refactored the expect

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -99,7 +102,32 @@ class TreeItem extends UIComponent<WithAsProp<TreeItemProps>> {
...
+  actionHandlers = {
+    getFocusFromParent: e => {
+      const { open } = this.props
+      if (open) {
+        e.stopPropagation()
+        this.titleRef.current.focus()
+      }
+    },
+    setFocusToFirstChild: e => {
+      if (!open) {
```
sophieH29, line 122: you forgot to read `open` from props 😄

packages/react/src/components/Tree/TreeTitle.tsx
```diff
@@ -66,6 +66,22 @@ class TreeTitle extends UIComponent<WithAsProp<TreeTitleProps>> {
       e.preventDefault()
       this.handleClick(e)
     },
+    expandOrFocusFirstChild: e => {
+      const { open } = this.props
+      e.preventDefault()
```
sophieH29, line 71: just a thought, something we can discuss: these action handlers just call handle click based on the open prop. I would say they have nothing to do with expand/collapse, but tree item does as it is responsible for rendering subtree. I would rather rename these action handlers

packages/react/src/components/Tree/TreeItem.tsx
silviuaavram: It's @Bugaa92 's fault. He gave me a code to copy paste at review, and he forgot to read `open` from props. Not my fault that I pasted his code mindlessly.

Will fix, thanks!

bmdalex (approved):
👍

Issue #1450, opened 2019-06-04
"Enter" key on "More option" button in attachment, invoke click on whole attachment


kolaps33:
### Steps
1. Go to https://codesandbox.io/s/pfo2n?module=/example.js
2. Navigate into the more (..) button in the `Attachment`
3. press Enter / Space key

### Expected

1. Enter key pressed: action is invoked only on the button
2. Space key pressed: action is invoked only on the button

### Actual
1. Enter key pressed: click action is invoked on whole Attachment and then on the button 
2. Space key pressed: click action is invoked on whole Attachment and NOT on the button at all

### Fix:

1. compare `e.currentTarget === e.target` to find out if the event coming from `Attachment` element or not... If yes then invoke click. 
2. (currently commented out in the code) add `keyActions` on `action` slot in `attachmentBehavior` and then stop propagation of it in `Attachment` component code.
3. other ideas?

kolaps33:
ticket is fixed :)

PR #1452, opened 2019-06-04
fix(attachment) - more option button invoke correctly action on space/enter

kolaps33:
Please see more details:

["Enter" key on "More option" button in attachment, invoke click on whole attachment #1450](https://github.com/stardust-ui/react/issues/1450)


packages/react/src/components/Attachment/Attachment.tsx
```diff
@@ -136,8 +136,10 @@ class Attachment extends UIComponent<WithAsProp<AttachmentProps>, AttachmentStat
   }
 
   performClick = e => {
-    e.stopPropagation()
-    this.handleClick(e)
+    if (e.currentTarget === e.target) {
```
jurokapsiar, line 139: based on agreement with @layershifter and @miroslavstastny this is ok for now, but it shows a bigger issue we have related to click handling of components that have parent/child clickable elements, especially if the elements trigger click on keyboard events (button, a)

This might help move the logic to behaviors: https://github.com/stardust-ui/react/issues/1502

jurokapsiar (approved):

Issue #1366, opened 2019-05-21
[Popup]: Improve async rendering API 


bmdalex:
### Problem description
TODO

### Proposed solution
Replace `renderContent` with `positioningDependencies`

## BREAKING CHANGE

```typescript
/**
 * Function to render popup content.
 * @param {Function} updatePosition - function to request popup position update.
 */
renderContent?: (updatePosition: Function) => ShorthandValue
```
will be replaced by
```typescript
/**
 * Array of conditions to be met in order to trigger a subsequent render
 * to reposition the elements.
 */
positioningDependencies?: any[]
```

jurokapsiar:
this can be done using the new children callback

PR #1455, opened 2019-06-05
feat(Tooltip): base implementation

### LINKED ISSUE #1453: [RFC] Tooltip - base component
### Problem description
We need a Tooltip component as part of Stardust, which is basically a popup that displays information related to an element when the element receives keyboard focus or the mouse hovers over it. It should follow the [ WAI-ARIA Tooltip Pattern](https://www.w3.org/TR/wai-aria-practices/#tooltip)

#### Material Design
The tooltip is implemented as a component which wraps the element associated with it. For example:
```jsx
<Tooltip title="Delete">
  <IconButton aria-label="Delete">
    <DeleteIcon />
  </IconButton>
</Tooltip>
```
#### Semantic UI React
Do not implement dedicated Tooltip component, they are reusing the Popup on hover for the Tooltip scenarios.
### Reakit
The recommended implementation is again in a form of component wrapping the element associated with the tooltip. Example:
```jsx
import React from "react";
import { Button } from "reakit/Button";
import {
  Tooltip as ReakitTooltip,
  TooltipReference,
  useTooltipState
} from "reakit/Tooltip";

function Tooltip({ children, title, ...props }) {
  const tooltip = useTooltipState();
  return (
    <>
      <TooltipReference {...tooltip}>
        {referenceProps =>
          React.cloneElement(React.Children.only(children), referenceProps)
        }
      </TooltipReference>
      <ReakitTooltip {...tooltip} {...props}>
        {title}
      </ReakitTooltip>
    </>
  );
}

function Example() {
  return (
    <Tooltip title="Tooltip">
      <Button>Reference</Button>
    </Tooltip>
  );
}
```

#### Fabric UI
Again the Tooltip is implemented as a wrapper component (TooltipHost) which is appearing on hover/focus. Example:
```jsx
import * as React from 'react';
import { DefaultButton } from 'office-ui-fabric-react/lib/Button';
import { TooltipHost, TooltipDelay, DirectionalHint } from 'office-ui-fabric-react/lib/Tooltip';
import { getId } from 'office-ui-fabric-react/lib/Utilities';

export class TooltipCustomExample extends React.Component<any, any> {
  // Use getId() to ensure that the ID is unique on the page.
  // (It's also okay to use a plain string without getId() and manually ensure uniqueness.)
  private _hostId: string = getId('tooltipHost');

  public render(): JSX.Element {
    return (
      <TooltipHost
        calloutProps={{ gapSpace: 20 }}
        tooltipProps={{
          onRenderContent: () => {
            return (
              <div>
                <ul style={{ margin: 0, padding: 0 }}>
                  <li>1. One</li>
                  <li>2. Two</li>
                </ul>
              </div>
            );
          }
        }}
        delay={TooltipDelay.zero}
        id={this._hostId}
        directionalHint={DirectionalHint.bottomCenter}
      >
        <DefaultButton aria-labelledby={this._hostId} text="Hover Over Me" />
      </TooltipHost>
    );
  }
}
```

#### React bootstrap
Here, there is an Overlay component, which is associated with the element via the target prop defined on it (they use ref for this).
```jsx
class Example extends React.Component {
  constructor(...args) {
    super(...args);

    this.attachRef = target => this.setState({ target });
    this.state = {
      show: false,
    };
  }

  render() {
    const { show, target } = this.state;
    return (
      <>
        <Button
          variant="danger"
          ref={this.attachRef}
          onClick={() => this.setState({ show: !show })}
        >
          Click me to see
        </Button>
        <Overlay target={target} show={show} placement="right">
          {({
            placement,
            scheduleUpdate,
            arrowProps,
            outOfBoundaries,
            show: _show,
            ...props
          }) => (
            <div
              {...props}
              style={{
                backgroundColor: 'rgba(255, 100, 100, 0.85)',
                padding: '2px 10px',
                color: 'white',
                borderRadius: 3,
                ...props.style,
              }}
            >
              Simple tooltip
            </div>
          )}
        </Overlay>
      </>
    );
  }
}

render(<Example />);
```

## Proposed solution
Following the common practices, we will add new Tooltip component, that will have the following API:

#### Common Stardust props
- accessibility
- styles
- variables
- className
- content - the content that will appear inside the tooltip
- children
#### Custom component props
- target - the element to which the tooltip is associated with
- mouseLeaveDelay - delay in ms for the mouse leave event, before the tooltip will be closed.
- pointing - whether the tooltip should show pointer towards the target
- align - alignment for the tooltip. Can have one of the following values: 'top' | 'bottom' | 'start' | 'end' | 'center'
- mountNode - Existing element the tooltip should be bound to.
- offset - offset value to apply to rendered component. Accepts the following units:
  - px or unit-less, interpreted as pixels
  - %, percentage relative to the length of the trigger element
  - %p, percentage relative to the length of the component element
  - vw, CSS viewport width unit
  - vh, CSS viewport height unit
- position - has one of the following value: 'above' | 'below' | 'before' | 'after'. It has higher priority than align. If position is vertical ('above' | 'below') and align is also vertical ('top' | 'bottom') or if both position and align are horizontal ('before' | 'after' and 'start' | 'end' respectively), then provided value for 'align' will be ignored and 'center' will be used instead.
With this API the component will be similar to the Popup component, and it will be easy for the users to learn the new component's API.

With this API the component will be similar to the Popup component, and it will be easy for the users to learn the new component's API.

mnajdova:
Fix https://github.com/stardust-ui/react/issues/1453 - add Tooltip component.
A tooltip displays information related to an element when the element receives keyboard focus, or the mouse hovers over it. The Tooltip component has the following API:
#### Common Stardust props
- `accessibility`
- `styles`
- `variables`
- `className`
- `content` - the content that will appear inside the tooltip
- `children`
#### Custom component props
- `target` - the element to which the tooltip is associated with
- `mouseLeaveDelay` - delay in ms for the mouse leave event, before the tooltip will be closed.
- `pointing` - whether the tooltip should show pointer towards the target
- `align` - alignment for the tooltip. Can have one of the following values: `'top' | 'bottom' | 'start' | 'end' | 'center'`
- `mountNode` - Existing element the tooltip should be bound to.
- `position` - has one of the following value: `'above' | 'below' | 'before' | 'after'`. It has higher priority than align. If position is vertical (`'above' | 'below'`) and align is also vertical (`'top' | 'bottom'`) or if both position and align are horizontal (`'before' | 'after' and 'start' | 'end'` respectively), then provided value for 'align' will be ignored and `'center'` will be used instead.
With this API the component will be similar to the Popup component, and it will be easy for the users to learn the new component's API.

TODO:
- [x] add Tooltip tests


codepretty:
@mnajdova thanks for starting on this!! There are just a couple things I noticed..

1. the pointing tooltip needs to have an svg beak for the pointer - @notandrew has the asset and can give to you
2. should `pointing` be the default? it seems to be the most common way i see tooltips
3. i don't think tooltips should ever render off page even if an offset is set?
image
4. i might just not have noticed this, but is there a way to specify one of the common color schemes? It might be nice to have those as example usages if possible :)
5. are nested tooltips a common paradigm?

jurokapsiar:
Similarly to [reakit tooltip](https://reakit.io/docs/tooltip/) or Fabric, we will need to have a reference between the tooltip between the tooltip and trigger, most probably using aria-describedby. When rendered it should look something like this:
```
<button aria-describedby="tooltip1">I am a button</button>
<div id="tooltip1">I am a tooltip</div>
```

Let's discuss how the API should look like, I would propose to make the `id` prop of `Tooltip` required so that the user would need to provide the ID. In the jsdoc comment for the id prop, we can say something like "put the tooltip id in the aria-describedby prop"

We can later (out of this PR) have some helper hook:
```
let tooltip = useTooltip();
<Tooltip
    {...tooltip.tooltipProps}
    trigger={<Button content="I am a button" {...tooltip.triggerProps} />}
    content={{
      content: "I am a tooltip"
    }}
  />
```

And for the actionable components:
```
<Button content="I am a button" tooltip="I am a tooltip" />
```
which would automatically render the Tooltip and Button with appropriate pros

mnajdova:
@jurokapsiar if the id is necessary, we can make it in required and hoist it in the Tooltip component. Then in the behavior we can specify this prop as `id` in the content and `aria-describedby` in the trigger. Wouldn't this be sufficient for all cases? We can make the adding of the required props in the behavior, I don't see why we need from the client to make this correlation. Am I missing something?

I see the tooltipBehavior as something like this:
```
const tooltipBehavior: Accessibility = (props: any) => {
  const { id } = props
  return {
    attributes: {
      trigger: {
        'aria-describedby': id,
      },
      content: {
        id,
      },
    },
  }
}
```

mnajdova:
@codepretty answers to your points

> @mnajdova thanks for starting on this!! There are just a couple things I noticed..
>
>the pointing tooltip needs to have an svg beak for the pointer - @notandrew has the asset and can give to you 

**Currently the beak is implemented exactlly as it is in the Popup. Are these two different? Let's align them both in a separate PR for red-lining the component, as this is base implementation**
>should pointing be the default? it seems to be the most common way i see tooltips 

**I was thinking the same, but as in the Popup this prop was by default false, I added that as a default. I will change this to be true by default in the Tooltip. If people are agains this, we can export the Tooltip on the client's side to have this to be true by default**
>i don't think tooltips should ever render off page even if an offset is set?

**Let me try to remove this option in the alignment of the Tooltip, I agree is not something typical for the Tooltip.**
>i might just not have noticed this, but is there a way to specify one of the common color schemes? It might be nice to have those as example usages if possible :) 

**Would rather wait for designs for something like that before adding examples in the docs, but yes if there are use cases like this, would be happy to add more examples**
>are nested tooltips a common paradigm? 

**NO! Very valid point, will remove that example :)**

jurokapsiar:
we did some testing on our prototypes and it looks like we will need to change the approach: the div containing tooltip text will always need to be present in the DOM and the trigger elements will need to point to it using aria-describedby. focus/hover should then only change the visibility. let's discuss offline how this can be achieved.

Regarding your response - yes, it can work, if we are ok with modifying the trigger.
There is one more case I need to think about - if the user provides element/component with aria-describedby - then the tooltip id should probably be added to the beginning of trigger's aria-describedby.

docs/src/examples/components/Tooltip/Types/TooltipContentWrapperExample.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const TooltipContentWrapperExample = () => {
+  const plainContentStyle = {
+    zIndex: 1000,
+    padding: 5,
+  }
+
+  return (
+    <Flex gap="gap.smaller">
+      <Tooltip content={<p style={plainContentStyle}>Plain tooltip content rendered 'as is'.</p>}>
```
layershifter: Do we really want to pay attention to this behavior? We previously had discussion and decision was to use `Popper` content for positioning purposes

docs/src/examples/components/Tooltip/Types/TooltipControlledExample.shorthand.tsx
```diff
@@ -0,0 +1,24 @@
...
+    return (
+      <Tooltip
+        open={this.state.open}
+        onOpenChange={this.handleOpenChange}
+        trigger={<Button icon="expand" />}
+        content={{ content: 'This is a controlled Tooltip' }}
+      />
+    )
+  }
+}
```
layershifter: Can be reworked with knobs...

docs/src/examples/components/Tooltip/Types/TooltipCustomTargetExample.tsx
```diff
@@ -0,0 +1,38 @@
...
+            "To the lascivious looking-glass I, that love's majesty to strut before a want love's
+            majesto, to the souls of York."
+          </Text>
+        </div>
+      </Grid>
+    )
+  }
+}
+
+export default PopupExample
```
layershifter: Cruft?

docs/src/examples/components/Tooltip/Types/TooltipCustomTargetExample.shorthand.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react'
+import { Tooltip, Button, Divider, Text, Grid } from '@stardust-ui/react'
+import { findDOMNode } from 'react-dom'
```
layershifter: Can we use `Ref` component?

docs/src/examples/components/Tooltip/Types/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+      description="A default tooltip."
+      examplePath="components/Tooltip/Types/TooltipExample"
+    />
+    <ComponentExample
+      title="Pointing"
+      description="A tooltip can have a pointer."
+      examplePath="components/Tooltip/Types/TooltipExamplePointing"
+    />
+    <ComponentExample
+      title="Controlled"
```
layershifter: This example should be in state section and present `open` prop

docs/src/examples/components/Tooltip/Types/index.tsx
```diff
@@ -0,0 +1,36 @@
...
+      description="Note that if Tooltip is controlled, then its 'open' prop value could be changed either by parent component, or by user actions (e.g. key press) - thus it is necessary to handle 'onOpenChange' event. Try to type some text into tooltip's input field and press ESC to see the effect."
+      examplePath="components/Tooltip/Types/TooltipControlledExample"
+    />
+    <ComponentExample
+      title="Content Wrapper"
+      description="Use 'content' prop of the Tooltip to set whether Tooltip content should be rendered with the default wrapper."
+      examplePath="components/Tooltip/Types/TooltipContentWrapperExample"
+    />
+    <ComponentExample
```
layershifter: Usage section...

docs/src/examples/components/Tooltip/Usage/TooltipExampleAsync.shorthand.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+const TooltipExampleAsync = () => (
+  <Tooltip
+    trigger={<Button icon="expand" content="Hover me!" />}
+    renderContent={updatePosition => ({ content: <AsyncDataLoader onLoaded={updatePosition} /> })}
+  />
+)
+
+export default TooltipExampleAsync
```
layershifter: Is it required for initial implementation, if so can we reuse `positionDependencies` prop and omit this `renderContent` stuff?

docs/src/examples/components/Tooltip/Types/TooltipExamplePointing.shorthand.tsx
```diff
@@ -0,0 +1,57 @@
...
+  { position: 'below', align: 'start', icon: 'arrow circle down', padding: '18px 42px 5px 5px' },
+  { position: 'below', align: 'center', icon: 'arrow circle down', padding: '18px 5px 5px 5px' },
+  { position: 'below', align: 'end', icon: 'arrow circle down', padding: '18px 5px 5px 42px' },
+  { position: 'before', align: 'top', icon: 'arrow circle left', padding: '5px 42px 18px 5px' },
+  { position: 'before', align: 'center', icon: 'arrow circle left', padding: '5px 42px 5px 5px' },
+  { position: 'before', align: 'bottom', icon: 'arrow circle left', padding: '18px 42px 5px 5px' },
+  { position: 'after', align: 'top', icon: 'arrow circle right', padding: '5px 5px 18px 42px' },
+  { position: 'after', align: 'center', icon: 'arrow circle right', padding: '5px 5px 5px 42px' },
+  { position: 'after', align: 'bottom', icon: 'arrow circle right', padding: '18px 5px 5px 42px' },
+]
```
layershifter: Can we add `pointing` prop into position example instead and keep only simple example there?

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,290 @@
...
+    inline: PropTypes.bool,
+    mountNode: customPropTypes.domNode,
+    mouseLeaveDelay: PropTypes.number,
+    open: PropTypes.bool,
+    onOpenChange: PropTypes.func,
+    pointing: PropTypes.bool,
+    position: PropTypes.oneOf(POSITIONS),
+    renderContent: PropTypes.func,
+    target: PropTypes.any,
+    trigger: customPropTypes.every([customPropTypes.disallow(['children']), PropTypes.any]),
```
layershifter: Is it really `any`?  I expect `element` at least

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,290 @@
...
+    return (
+      <>
+        {this.renderTrigger(accessibility)}
+        {mountNode && ReactDOM.createPortal(tooltipContent, mountNode)}
+      </>
+    )
+  }
+
+  getTriggerProps(triggerElement) {
+    const triggerProps: any = {}
```
layershifter: Valid type?

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,290 @@
...
+    position: 'above',
+    mouseLeaveDelay: 500,
+    pointing: true,
+  }
+
+  static autoControlledProps = ['open']
+
+  pointerTargetRef = React.createRef<HTMLElement>()
+  triggerRef = React.createRef<HTMLElement>() as React.MutableRefObject<HTMLElement>
+  tooltipDomElement = null
```
layershifter: I see the usage, but I don't see an assignment. Can you please check?

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,290 @@
...
+        align={align}
+        position={position}
+        rtl={rtl}
+        targetRef={target ? toRefObject(target) : this.triggerRef}
+        children={this.renderPopperChildren.bind(this, tooltipPositionClasses, rtl, accessibility)}
+      />
+    )
+  }
+
+  renderPopperChildren = (
```
layershifter, line 219: It looks that we can simplify code if we will move this method directly to `renderTooltipContent`

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,290 @@
...
+        <Ref innerRef={this.triggerRef}>
+          {React.cloneElement(triggerElement, {
+            ...accessibility.attributes.trigger,
+            ...triggerProps,
+            ...applyAccessibilityKeyHandlers(accessibility.keyHandlers.trigger, triggerProps),
+          })}
+        </Ref>
+      )
+    )
+  }
```
layershifter: Do we really need `renderTrigger` method?

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,290 @@
...
+
+  renderTrigger(accessibility) {
+    const { children, trigger } = this.props
+    const triggerElement = childrenExist(children) ? children : (trigger as any)
+    const triggerProps = this.getTriggerProps(triggerElement)
+    return (
+      triggerElement && (
+        <Ref innerRef={this.triggerRef}>
+          {React.cloneElement(triggerElement, {
```
layershifter: This stuff is unsafe because we don't use `React.Children.only()` that grants that only single child will be passed

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,290 @@
...
+  getContentProps = (predefinedProps?) => {
+    const contentHandlerProps: any = {}
+
+    contentHandlerProps.onMouseEnter = (e, contentProps) => {
+      this.setTooltipOpen(true, e)
+      predefinedProps && _.invoke(predefinedProps, 'onMouseEnter', e, contentProps)
+    }
+    contentHandlerProps.onMouseLeave = (e, contentProps) => {
+      this.setTooltipOpen(false, e)
+      predefinedProps && _.invoke(predefinedProps, 'onMouseLeave', e, contentProps)
```
layershifter:
```suggestion
      _.invoke(predefinedProps, 'onMouseLeave', e, contentProps)
```

Do we really need condition?

docs/src/examples/components/Tooltip/Types/TooltipControlledExample.shorthand.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react'
+import { Button, Tooltip } from '@stardust-ui/react'
+
+class TooltipControlledExample extends React.Component {
+  state = { open: false }
+
+  handleOpenChange = (e, { open }) => {
+    console.log(`Tooltip requested to change its open state to "${open}".`)
```
bmdalex: we usually use alert or we display something on the screen; I don't see this pattern anywhere else

docs/src/examples/components/Tooltip/Types/TooltipControlledExample.shorthand.tsx
bmdalex: https://github.com/stardust-ui/react/blob/727e9c2915ca8450e1e030ab37d1b7bc12e65615/docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx#L7

docs/src/examples/components/Tooltip/Usage/TooltipExampleAsync.shorthand.tsx
bmdalex: +1 for `positionDependencies` or omitting this for now as it's gonna need reworking (see #1366)

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,290 @@
...
+    }
+
+    return contentHandlerProps
+  }
+
+  shouldBlurClose = e => {
```
bmdalex: NIT: since the body of this function does not suggest any relationship with the concept of blur specifically I'd recommend:
~~`shouldBlurClose`~~ --> `shouldClose` or `shouldCloseOnEvent`

packages/react/src/lib/positioner/types.ts
```diff
@@ -6,12 +6,21 @@ export type Alignment = 'top' | 'bottom' | 'start' | 'end' | 'center'
 
 export type PopperChildrenFn = (props: PopperChildrenProps) => React.ReactNode
 
-export interface PositioningProps {
+export interface SimplifiedPositioningProps {
```
bmdalex: NIT: another recommendation would be `BasicPositioningProps` but I'm ok with any of them

packages/react/src/themes/teams/getPointerStyles.ts
```diff
@@ -0,0 +1,56 @@
+import { PopperChildrenProps } from '../../lib/positioner'
+
+const rtlMapping = {
+  left: 'right',
+  right: 'left',
+}
+
+const getPointerStyles = (
```
bmdalex, line 8: 👍

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,290 @@
...
+            ...accessibility.attributes.trigger,
+            ...triggerProps,
+            ...applyAccessibilityKeyHandlers(accessibility.keyHandlers.trigger, triggerProps),
+          })}
+        </Ref>
+      )
+    )
+  }
+
+  renderTooltipContent(
```
bmdalex, line 200: `renderTooltipContent` and `renderPopperChildren` have a bit of duplicated code with `Popup.tsx` - anything worth reusing? treat this comment as NIT

bmdalex (commented):
👍 good stuff, pls take a look at the comments

docs/src/examples/components/Tooltip/Types/TooltipContentWrapperExample.tsx
mnajdova: I just copy the examples we had in the Popup component, but you are right, we don't want to advertise this. Will remove it!

docs/src/examples/components/Tooltip/Types/TooltipControlledExample.shorthand.tsx
mnajdova: I'll rework the example, but I am a bit worried, that we are not showing how the onOpenChange handler should be used in this case.. :\

layershifter: Knobs expose `set` callback as second argument:

```
const [open, setOpen] = useBoolean()
//
onOpenChange={(e, data) => setOpen(data.open)}
```

docs/src/examples/components/Tooltip/Usage/TooltipExampleAsync.shorthand.tsx
mnajdova: Ok, will revert this changes

packages/react/src/components/Tooltip/Tooltip.tsx
mnajdova: this method is actually the children of Popper, so it is invoked with some more parameters.

docs/src/examples/components/Tooltip/Types/TooltipControlledExample.shorthand.tsx
mnajdova: Will change this to display something on the screen, the Alert doesn't make sense for the Tooltip, so I changed this to console.log, but it make sense to be visible for the user

docs/src/examples/components/Tooltip/Types/TooltipExamplePointing.shorthand.tsx
mnajdova: Yep, added.

packages/react/src/components/Tooltip/Tooltip.tsx
mnajdova: Yep, I removed the ref around the tooltip element. Added, please check.

packages/react/src/components/Tooltip/Tooltip.tsx
mnajdova: removed

docs/src/examples/components/Tooltip/Types/TooltipControlledExample.shorthand.tsx
mnajdova: Reworked this example with knobs.

packages/react/src/components/Tooltip/Tooltip.tsx
mnajdova: Changed, please check.

docs/src/examples/components/Tooltip/Types/TooltipControlledExample.shorthand.tsx
mnajdova: Ok, I'll reworked the example like this, as I see this as a rather important thing for the clients to be aware of.

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -120,7 +120,7 @@ const Popper: React.FunctionComponent<PopperProps> = props => {
       popperRef.current = createPopper(targetRef.current, contentRef.current, options)
     },
     // TODO review dependencies for popperHasScrollableParent
-    [computedModifiers, enabled, userModifiers, positionFixed, proposedPlacement],
+    [computedModifiers, enabled, userModifiers, positionFixed, proposedPlacement, targetRef],
```
mnajdova: @Bugaa92 @layershifter please take a look on this change. I needed this in order for the Tooltip with custom target to work. Should we add here targetRef.current instead (or together with this)?

docs/src/examples/components/Tooltip/Variations/TooltipExamplePosition.shorthand.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  return (
+    <Grid columns="1" variables={{ padding: '100px 0' }} styles={{ justifyItems: 'center' }}>
+      <Tooltip
+        open={open || undefined}
+        align={align}
+        position={position}
+        pointing={pointing}
+        trigger={<Button icon={icons[position]} styles={buttonStyles} />}
+        content={{
```
layershifter: Can be simplified:
```suggestion
content={`The tooltip is rendered ${position} the trigger, aligned to the ${align}.`}
```

docs/src/examples/components/Tooltip/Variations/TooltipExamplePosition.shorthand.tsx
```diff
@@ -0,0 +1,76 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+import { Button, Grid, Tooltip, Alignment, Position } from '@stardust-ui/react'
+import { useBooleanKnob, useSelectKnob } from '@stardust-ui/docs-components'
+
+const TooltipExamplePosition = () => {
+  const [open] = useBooleanKnob({ name: 'open-s', initialValue: true })
```
layershifter: Looks that `Tooltip` in this example should be always opened

```suggestion
```

docs/src/examples/components/Tooltip/Variations/index.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react'
+
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
+
+const Variations = () => (
+  <ExampleSection title="Variations">
+    <ComponentExample
+      title="Alignment, Position and Pointing"
```
layershifter:
```suggestion
      title="Alignment & Position"
```

I suggest to also add a separate example to `pointing` in `Types` section because it's a type

Very simple, with only single `Tooltip` without `alignment` and `position`...

docs/src/examples/components/Tooltip/States/TooltipOpenControlledExample.shorthand.tsx
```diff
@@ -0,0 +1,18 @@
...
+    <Tooltip
+      open={open}
+      onOpenChange={(e, data) => setOpen(data.open)}
+      trigger={<Button icon="expand" />}
+      content={{ content: 'This is a controlled Tooltip' }}
+    />
+  )
+}
+
+export default TooltipControlledExample
```
layershifter:
```suggestion
export default TooltipOpenExample
```

It's an example for `open` state

docs/src/examples/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import Variations from './Variations'
+import Usage from './Usage'
+import States from './States'
+
+const TooltipExamples = () => (
+  <>
+    <Types />
+    <Variations />
+    <Usage />
+    <States />
```
layershifter, line 13: To be aligned with other pages: `States` are before `Usage`

packages/react/src/components/Tooltip/TooltipContent.tsx
```diff
@@ -0,0 +1,130 @@
...
+    accessibility: defaultBehavior,
+  }
+
+  private handleMouseEnter = e => {
+    _.invoke(this.props, 'onMouseEnter', e, this.props)
+  }
+
+  private handleMouseLeave = e => {
+    _.invoke(this.props, 'onMouseLeave', e, this.props)
+  }
```
layershifter: As I see these handlers do nothing and `contentProps` from are used nowhere. Let's remove them

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,286 @@
...
+      defaultProps: {
+        ...tooltipContentAttributes,
+        placement,
+        pointing,
+        pointerRef: this.pointerTargetRef,
+      },
+      overrideProps: this.getContentProps,
+    })
+
+    return (
```
layershifter:
```tsx
    return <Ref innerRef={this.contentRef}>{tooltipContent}</Ref>
```

packages/react/src/components/Tooltip/Tooltip.tsx
layershifter: In this case let's use `contentRef`:

```suggestion
  contentRef = React.createRef<HTMLElement>
```

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,286 @@
...
+    }
+    contentHandlerProps.onMouseLeave = (e, contentProps) => {
+      this.setTooltipOpen(false, e)
+      _.invoke(predefinedProps, 'onMouseLeave', e, contentProps)
+    }
+
+    return contentHandlerProps
+  }
+
+  shouldClose = e => {
```
layershifter:
```tsx
shouldClose = e => {
    const containsTarget = _.invoke(e, 'currentTarget.contains', e.relatedTarget) && _.invoke(this.contentRef.current, 'contains', e.relatedTarget)
    
    return !containsTarget
```

Can be simplied with `Lodash`, may be it makes sense to rename it:

```tsx
shouldStayOpen = e => _.invoke(e, 'currentTarget.contains', e.relatedTarget) && _.invoke(this.contentRef.current, 'contains', e.relatedTarget)
```

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,286 @@
...
+    align: PropTypes.oneOf(ALIGNMENTS),
+    defaultOpen: PropTypes.bool,
+    inline: PropTypes.bool,
+    mountNode: customPropTypes.domNode,
+    mouseLeaveDelay: PropTypes.number,
+    open: PropTypes.bool,
+    onOpenChange: PropTypes.func,
+    pointing: PropTypes.bool,
+    position: PropTypes.oneOf(POSITIONS),
+    target: PropTypes.any,
```
layershifter:
```suggestion
    target: customPropTypes.domNode,
```

Should be more accurate

packages/react/src/components/Tooltip/Tooltip.tsx
mnajdova: Agreed, is more simpler, just `&&` should be replaced with `||`:
```
shouldStayOpen = e => _.invoke(e, 'currentTarget.contains', e.relatedTarget) || _.invoke(this.contentRef.current, 'contains', e.relatedTarget)

```

docs/src/examples/components/Tooltip/Variations/TooltipExamplePosition.shorthand.tsx
mnajdova: Agreed, we have a separate example for the open prop.

packages/react/src/lib/positioner/Popper.tsx
layershifter: It's correct, @Bugaa92 will attach an updated example's code

docs/src/examples/components/Tooltip/Usage/TooltipCustomTargetExample.tsx
```diff
@@ -0,0 +1,34 @@
...
+              majesto, to the souls of York."
+            </Text>
+          </Ref>
+        </div>
+      </Grid>
+    )
+  }
+}
+
+export default TooltipExample
```
layershifter:
```suggestion
export default TooltipTargetExample
```

And rename file please to match other examples

docs/src/examples/components/Tooltip/Usage/TooltipCustomTargetExample.shorthand.tsx
```diff
@@ -0,0 +1,33 @@
+import * as React from 'react'
+import { Tooltip, Button, Divider, Text, Grid, Ref } from '@stardust-ui/react'
+
+class TooltipExample extends React.Component {
```
bmdalex: Investigated with @layershifter and this example does not work correctly (try it in maximized view and you'll see that);

**the reason**: unfortunately the refs are not triggering a re-render so we need to use state to do that ourselves as clients.. here is the correct example that also needs the change you're mentioning [here](https://github.com/stardust-ui/react/pull/1455#discussion_r292454499), which is correct and needed BTW:

```jsx
import { Button, Divider, Grid, Ref, Text, Tooltip } from '@stardust-ui/react'

const TooltipExample = () => {
  const [target, setTarget] = React.useState<HTMLElement>(null)

  return (
    <Grid columns="auto 1fr">
      {/* CUSTOM DOM ELEMENT is used as target for Tooltip */}
      <Tooltip
        target={target}
        trigger={<Button icon="question" circular styles={{ cursor: 'pointer' }}/>}
        content="well, yes, I am just a garbish text ¯\_(ツ)_/¯"
        position="below"
      />
      <div style={{ marginLeft: 10 }}>
        <Text>Could you guess what does this text means? :)</Text>
        <Divider/>
        <Ref innerRef={setTarget}>
          <Text>
            "To the lascivious looking-glass I, that love's majesty to strut before a want love's
            majesto, to the souls of York."
          </Text>
        </Ref>
      </div>
    </Grid>
  )
}

export default TooltipExample
```

packages/react/src/lib/positioner/Popper.tsx
bmdalex: I investigated this with @layershifter and this change is correct but the examples need to be updated, pls take a look at https://github.com/stardust-ui/react/pull/1455/files#r292519287

packages/react/src/themes/teams/components/Tooltip/tooltipContentStyles.ts
```diff
@@ -0,0 +1,37 @@
...
+const tooltipContentStyles: ComponentSlotStylesInput<
+  TooltipContentProps,
+  TooltipContentVariables
+> = {
+  root: ({ props: p, variables: v, rtl }): ICSSInJSStyle => ({
+    borderRadius: v.borderRadius,
+    display: 'block',
+    maxWidth: v.maxWidth,
+
+    ...(p.pointing && getPointerStyles(v.pointerOffset, v.pointerMargin, rtl, p.placement).root),
```
layershifter, line 15: May be two separate functions will be better, for `root` and `pointer`?

packages/react/src/themes/teams/components/Tooltip/tooltipContentStyles.ts
```diff
@@ -0,0 +1,37 @@
...
+    width: v.pointerSize,
+
+    ...getPointerStyles(v.pointerOffset, v.pointerMargin, rtl, p.placement).pointer,
+  }),
+  content: ({ props: p, variables: v }): ICSSInJSStyle => ({
+    display: 'block',
+    padding: v.padding,
+
+    borderRadius: 'inherit',
+    boxShadow: `0 2px 4px 0 ${v.boxShadowColor}, 0 2px 10px 0 ${v.boxShadowColor}`,
```
layershifter: Should it be a variable?

packages/react/test/specs/components/Tooltip/Tooltip-test.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react'
+
+import Tooltip from 'src/components/Tooltip/Tooltip'
+
+import { mountWithProvider } from '../../../utils'
+
+describe('Tooltip', () => {
+  describe('onOpenChange', () => {
+    test('is called on hover', () => {
+      const spy = jest.fn()
```
layershifter:
```suggestion
      const onOpenChange = jest.fn()
```

packages/react/src/themes/teams/components/Tooltip/tooltipContentStyles.ts
mnajdova: This is how getBorderFocusStyles is implemented, so I wouldn't change this one. Btw this is just extracted as a separate function, so wouldn't like to change it as part of this PR.

packages/react/test/specs/components/Tooltip/Tooltip-test.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+      mountWithProvider(<Tooltip trigger={<button />} content="Hi" onOpenChange={spy} />)
+        .find('button')
+        .simulate('mouseEnter')
+
+      expect(spy).toHaveBeenCalledTimes(1)
+      expect(spy.mock.calls[0][1]).toMatchObject({ open: true })
```
layershifter:
```js
expect(onOpenChange).toHaveBeenCalledWith(
  expect.objectContaining({ type: 'mouseEnter' }),
  expect.objectContaining({ open: true }),
)
```

There is a dedicated assertion

packages/react/src/themes/teams/components/Tooltip/tooltipContentStyles.ts
mnajdova: The only thing that is different between themes is the color, otherwise I would have to redefine everything..

packages/react/test/specs/components/Tooltip/Tooltip-test.tsx
mnajdova: Unfortunately, with these changes the tests are time-outing, not sure why.. Let's investigate this tomorrow.

layershifter:
```js
expect(onOpenChange).toHaveBeenCalledTimes(1)
expect(onOpenChange).toHaveBeenCalledWith(
  expect.objectContaining({ type: 'mouseenter' }),
  expect.objectContaining({ open: true }),
)
```

There was typo in `type`. Tests timeouts because Jest tries to pretty print giant object with an event

packages/react/src/themes/teams/components/Tooltip/tooltipContentStyles.ts
layershifter: What about separate variables?

```
boxShadow: `${v.boxSomeName} ${v.boxShadowColor}, ${v.boxSomeName} ${v.boxShadowColor}`,
```

packages/react/src/themes/teams-high-contrast/components/Tooltip/tooltipContentVariables.ts
```diff
@@ -0,0 +1,3 @@
+export default {
```
layershifter: Type it 😈 

```
export default (): Partial<TooltipContentVariables> => ({
  boxShadowColor: undefined,
})
```

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,273 @@
...
+    position: 'above',
+    mouseLeaveDelay: 500,
+    pointing: true,
+  }
+
+  static autoControlledProps = ['open']
+
+  pointerTargetRef = React.createRef<HTMLElement>()
+  triggerRef = React.createRef<HTMLElement>() as React.MutableRefObject<HTMLElement>
+  contentRef = React.createRef<HTMLElement>() as React.MutableRefObject<HTMLElement>
```
layershifter:
```suggestion
  contentRef = React.createRef<HTMLElement>()
```

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,273 @@
...
+    mountNode: isBrowser() ? document.body : null,
+    position: 'above',
+    mouseLeaveDelay: 500,
+    pointing: true,
+  }
+
+  static autoControlledProps = ['open']
+
+  pointerTargetRef = React.createRef<HTMLElement>()
+  triggerRef = React.createRef<HTMLElement>() as React.MutableRefObject<HTMLElement>
```
layershifter:
```suggestion
  triggerRef = React.createRef<HTMLElement>()
```

We need that `as` in `Popup` because we manually update `.current` there

docs/src/examples/components/Tooltip/Types/TooltipExamplePointing.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
+import { Button, Tooltip } from '@stardust-ui/react'
+
+const TooltipExamplePointing = () => (
+  <Tooltip open pointing content={`The tooltip is pointing.`}>
```
layershifter:
```suggestion
  <Tooltip open pointing content="The tooltip is pointing.">
```

I will activate lint rule for this soon 😺

docs/src/examples/components/Tooltip/Types/TooltipExamplePointing.shorthand.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react'
+import { Button, Tooltip } from '@stardust-ui/react'
+
+const TooltipExamplePointing = () => (
+  <Tooltip open pointing trigger={<Button icon="expand" />} content={`The tooltip is pointing.`} />
```
layershifter:
```suggestion
  <Tooltip open pointing trigger={<Button icon="expand" />} content="The tooltip is pointing." />
```

docs/src/examples/components/Tooltip/Types/TooltipExample.shorthand.steps.ts
```diff
@@ -0,0 +1,11 @@
+import { Button } from '@stardust-ui/react'
+
+const config: ScreenerTestsConfig = {
+  themes: ['teams', 'teamsDark', 'teamsHighContrast'],
+  steps: [
+    builder => builder.hover(`.${Button.className}`).snapshot('Shows tooltip'),
+    builder => builder.hover('.ui-provider__box').snapshot('Shows tooltip'),
```
layershifter: What this step does? Naming is wrong? 🤔

mnajdova: It's for resetting, we don't even need screenshot, will refactor it.

layershifter (approved):

bmdalex (approved):
👍 great job!

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -127,7 +127,15 @@ const Popper: React.FunctionComponent<PopperProps> = props => {
       popperRef.current = createPopper(targetRef.current, contentRef.current, options)
     },
     // TODO review dependencies for popperHasScrollableParent
```
kuzhelov, line 129: is this comment still necessary - I mean, if these are reviewed, we may remove `TODO` comment :) or there are some checks that you still expect to be made?

mnajdova: Not sure, this was here before. I found this issue while working on the tooltip, so I added the triggerRef there as well. I think is @Bugaa92 's call about whether it is safe to remove it

kuzhelov (approved):

docs/src/examples/components/Tooltip/States/TooltipOpenControlledExample.shorthand.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react'
+import { Button, Tooltip } from '@stardust-ui/react'
+import { useBooleanKnob } from '@stardust-ui/docs-components'
+
+const TooltipOpenExample = () => {
+  const [open, setOpen] = useBooleanKnob({ name: 'open-c', initialValue: true })
```
layershifter, line 6:
```suggestion
  const [open, setOpen] = useBooleanKnob({ name: 'open', initialValue: true })
```

After #1489 will be merged

docs/src/examples/components/Tooltip/States/TooltipOpenControlledExample.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react'
+import { Button, Tooltip } from '@stardust-ui/react'
+import { useBooleanKnob } from '@stardust-ui/docs-components'
+
+const TooltipOpenExample = () => {
+  const [open, setOpen] = useBooleanKnob({ name: 'open-c', initialValue: true })
```
layershifter, line 6:
```suggestion
  const [open, setOpen] = useBooleanKnob({ name: 'open', initialValue: true })
```

After #1489 will be merged

docs/src/examples/components/Tooltip/Variations/TooltipExamplePosition.tsx
```diff
@@ -0,0 +1,67 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+import { Button, Grid, Tooltip, Alignment, Position } from '@stardust-ui/react'
+import { useSelectKnob } from '@stardust-ui/docs-components'
+
+const TooltipExamplePosition = () => {
+  const [positionAndAlign] = useSelectKnob({
+    name: 'position-align-c',
```
layershifter, line 8:
```suggestion
    name: 'position-align',
```

After #1489 will be merged

docs/src/examples/components/Tooltip/Variations/TooltipExamplePosition.shorthand.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+import { Button, Grid, Tooltip, Alignment, Position } from '@stardust-ui/react'
+import { useSelectKnob } from '@stardust-ui/docs-components'
+
+const TooltipExamplePosition = () => {
+  const [positionAndAlign] = useSelectKnob({
+    name: 'position-align-s',
```
layershifter, line 8:
```suggestion
    name: 'position-align',
```

After #1489 will be merged

PR #1457, opened 2019-06-05
feat(Tree): expand all siblings on asterisk key

silviuaavram:
Follows the pattern at https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-1/treeview-1a.html in the keyboard support 'asterisk' section.

On asterisk key, the tree and all its siblings are expanded.


packages/react/src/components/Tree/Tree.tsx
```diff
@@ -90,6 +91,22 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
 
   static autoControlledProps = ['activeIndex']
 
+  protected actionHandlers = {
+    expandSiblings: e => {
+      const { items, exclusive } = this.props
+      e.preventDefault()
+      e.stopPropagation()
+
+      if (exclusive) {
```
sophieH29, line 100: move this check above `e.stopPropagation` & `e.preventDefault`. If we don't handle this keyboard event, we should let it propagate

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -90,6 +91,22 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
+        return
+      }
+
+      this.setState({
+        activeIndex: Array.from(Array(items.length).keys()),
```
```suggestion
        activeIndex: _.range(0, items.length),
```

packages/react/src/components/Tree/Tree.tsx
silviuaavram: I don't think we should. The event will come from this Tree's title. If we leave it propagate to the parent tree, that one may expand all its siblings, which is not something we want.

This check with `exclusive` is stupid in my opinion, but since we still have this `exclusive` prop for the Tree I had to put it. Maybe we should push to remove it.

bmdalex (approved):

sophieH29 (approved):

PR #1458, opened 2019-06-05
chore(eslint): add lint rule to disable access modifiers

layershifter:
Related to #936.

This PR adds a new rule that disallows to use access modifiers (`private`, `public`, `protected`) for `.tsx` files.

It also adds a new package `@stardust-ui/eslint-plugin` because it's required by ESLint, i.e. you can't simply place a file with rules somewhere, only in the package that has correct naming:
https://eslint.org/docs/developer-guide/working-with-plugins


packages/eslint-plugin/rules/no-visibility-modifiers/index.js
```diff
@@ -0,0 +1,117 @@
...
+      })
+    }
+
+    /**
+     * Checks if a method declaration has an accessibility modifier.
+     * @param methodDefinition The node representing a MethodDefinition.
+     */
+    function checkMethodAccessibilityModifier(methodDefinition) {
+      let nodeType = 'method definition'
+      switch (methodDefinition.kind) {
```
miroslavstastny:
```js
if (['get', 'set'].includes(methodDefinition.kind)
```

miroslavstastny (approved):

PR #1459, opened 2019-06-06
perf(render): use single Fela renderer

### LINKED ISSUE #1311: Fela memory improvements
Ran a test in Teams after loading 400 messages
- The fela cache size is ~270KB
- Fela is also adding a lot of strings in memory due to bloated class names and reference to normalize.css:
image

image

layershifter:
Fixes #1311.

### Problem

Previously we had multiple Fela renderers for LTR and RTL.

#### Why it is a problem?
- complicated logic of switching
- double usage of memory: each renderer has own cache
- tricks to properly create subscriptions: https://github.com/stardust-ui/react/blob/master/packages/react/src/components/Provider/Provider.tsx#L184

#### Why we need this previously?
- conditional style transforms with [`fela-plugin-rtl`](https://github.com/rofrischmann/fela/tree/master/packages/fela-plugin-rtl), were added in `10.2.1`
- proper support for RTL in animations - fixed with usage [`fela-plugin-embedded`](https://github.com/rofrischmann/fela/tree/master/packages/fela-plugin-embedded), required fix for its usage was release in `10.4.1`

With latest changes in Fela we can remove this tricks 🚀


packages/react/src/components/Provider/Provider.tsx
```diff
@@ -1,18 +1,12 @@
...
 import { RendererProvider, ThemeProvider, ThemeContext } from 'react-fela'
 import * as customPropTypes from '@stardust-ui/react-proptypes'
 
-import {
-  felaRenderer as felaLtrRenderer,
-  felaRtlRenderer,
-  isBrowser,
-  ChildrenComponentProps,
-} from '../../lib'
+import { felaRenderer, ChildrenComponentProps } from '../../lib'
```
mnajdova, line 9: I love the simplification you've done here 👍

packages/react/src/lib/felaRenderer.tsx
```diff
@@ -62,21 +65,15 @@ const createRendererConfig = (options: any = {}) => ({
     // Heads up!
     // This is required after fela-plugin-prefixer to resolve the array of fallback values prefixer produces.
     felaPluginFallbackValue(),
+
```
mnajdova, line 67: `felaExpandCssShorthandsPlugin` should go after `felaPluginFallbackValue`. Watch out for the next merge with master.

packages/react/src/lib/mergeProviderContexts.ts
```diff
@@ -36,8 +36,8 @@ const mergeProviderContexts = (...contexts: ProviderContextInput[]): ProviderCon
         acc.rtl = mergedRTL
       }
 
-      // Use the correct renderer for RTL
-      acc.renderer = acc.rtl ? felaRtlRenderer : felaRenderer
+      // Use the correct renderer
+      acc.renderer = felaRenderer
```
mnajdova: I think now we can safely use the renderer of the outgoing context if defined when merging. This can be our fallback value if that's not defined.

packages/react/src/themes/teams/components/Loader/loaderStyles.ts
```diff
@@ -48,12 +45,9 @@ export default {
     }
     const svgAnimation: ICSSInJSStyle = {
       animationName: {
-        keyframe: () =>
-          ({
-            to: {
-              transform: `translate3d(0, ${v.svgTranslatePosition[p.size]}, 0)`,
-            },
-          } as any),
+        to: {
```
mnajdova, line 48: I like that we have the two options when defining animations - directly defining keyframes, as well as `keyframe` + `params`.

packages/react/src/themes/types.ts
```diff
@@ -209,15 +209,18 @@ export interface ICSSPseudoElementStyle extends ICSSInJSStyle {
   content?: string
 }
 
-export interface StardustAnimationName {
-  keyframe?: any
-  params?: object
+type StardustAnimationKeyFrame = Record<string, React.CSSProperties>
```
mnajdova: Aren't there typings for keyframes coming from CssInJS/fela?

packages/react/test/specs/commonTests/isConformant.tsx
```diff
@@ -408,7 +408,7 @@ export default (Component, options: Conformant = {}) => {
           'data-simulate-event-here': true,
         }
 
-        const component = mount(<Component {...props} />).childAt(0)
+        const component = mount(<Component {...props} />)
```
mnajdova: Why is this not necessary anymore?

packages/react/test/specs/lib/felaInvokeKeyframesPlugin-test.ts
```diff
@@ -0,0 +1,59 @@
...
+        keyframe: ({ from }) => ({ from: { rotate: from }, to: { rotate: '360deg' } }),
+        params: { from: '100deg' },
+      },
+      animationDuration: '2s',
+    }
+
+    expect(renderInvokeKeyframes(style, 'RULE', felaRenderer)).toMatchObject({
+      animationName: expect.objectContaining({
+        from: { rotate: '100deg' },
+        to: expect.any(Object),
```
mnajdova, line 54: Why `expect.any(Object)` here?

mnajdova: Aah I see we have only the from param.

packages/react/test/specs/commonTests/isConformant.tsx
layershifter: See #1469 🐱

packages/react/src/themes/types.ts
layershifter: Fela exports `TKeyFrame`, but it's a type for a function 😢 Reworked these type to this one:

```jsx
type AnimationKeyFrame = Record<'from' | 'to' | string, ICSSInJSStyle>

export interface StardustAnimationName<P = Record<string, any>> {
  keyframe?: AnimationKeyFrame | ((params: P) => AnimationKeyFrame)
  params?: P
}
```

packages/react/src/lib/mergeProviderContexts.ts
layershifter: Updated:
```js
acc.renderer = next.renderer || acc.renderer
```

packages/react/src/lib/felaRenderer.tsx
layershifter: Merged this update

packages/react/src/components/Provider/Provider.tsx
```diff
@@ -87,15 +81,6 @@ class Provider extends React.Component<WithAsProp<ProviderProps>> {
...
-  get topLevelFelaRenderer() {
-    if (!Provider._topLevelFelaRenderer) {
-      Provider._topLevelFelaRenderer = this.props.rtl ? felaRtlRenderer : felaLtrRenderer
-    }
-    return Provider._topLevelFelaRenderer
-  }
-
   renderStaticStyles = (mergedTheme: ThemePrepared) => {
     // RTL WARNING
     // This function sets static styles which are global and renderer agnostic.
```
miroslavstastny: Remove the mentions of renderer in the comment, these make no sense now.

packages/react/src/lib/felaInvokeKeyframesPlugin.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Fela plugin for invoking keyframes with params. The keyframes, defined in the animationName prop,
+ * are called with the params object, if defined in the animationName prop.
+ *
+ * Caution! Infinite recursion is possible in case if style object has links to self in the props
+ * tree.
+ */
+export default () => {
+  const invokeKeyframes = (styles: Object, type?, renderer?, props?) => {
```
miroslavstastny: Remove unused params: `type`, `renderer`, `props`

miroslavstastny (approved):

PR #1461, opened 2019-06-06
fix(felaPluginFallbackValue): broken because of order of plugins

### LINKED ISSUE #1460: CSS Fallback values do not work (again)
### Steps
Specify style as an array:
```jsx
styles: {
  border: ["1px solid red", "1px solid blue"],
},
```

### Expected Result
This is considered valid, both values are applied in CSS:
image

### Actual Result
No value applied
image

### Version
0.32.0, it broke between 0.29.0 and 0.29.1

### Testcase
https://codesandbox.io/s/css-fallback-value-wmg81

mnajdova:
Fix https://github.com/stardust-ui/react/issues/1460 After introducing the `felaExpandCssShorthandsPlugin` the `felaPluginFallbackValue` was broken, because of the wrong order of the plugins. The `felaExpandCssShorthandsPlugin` should run after the `felaPluginFallbackValue`. As part of the PR there is a screenshot test added that will ensure the fallback value plugin is working as expected.

### DIFF
--- a/docs/src/examples/components/Provider/Types/ProviderExampleRendererFelaPluginFallbackValue.shorthand.tsx
+++ b/docs/src/examples/components/Provider/Types/ProviderExampleRendererFelaPluginFallbackValue.shorthand.tsx
@@ -2,10 +2,16 @@ import * as React from 'react'
 import { Segment } from '@stardust-ui/react'
 
 const ProviderExampleShorthand = () => (
-  <Segment
-    content="Segment with blue border"
-    styles={{ border: ['1px solid red', '1px solid blue'] as any }}
-  />
+  <>
+    <Segment
+      content="Segment with red border"
+      styles={{ border: ['1px solid red', '1px solid invalid'] as any }}
+    />
+    <Segment
+      content="Segment with blue border"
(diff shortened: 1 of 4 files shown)


docs/src/examples/components/Provider/Types/ProviderExampleRendererFelaPluginFallbackValue.shorthand.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react'
+import { Segment } from '@stardust-ui/react'
+
+const ProviderExampleShorthand = () => (
+  <Segment
+    content="Segment with blue border"
+    styles={{ border: ['1px solid red', '1px solid blue'] as any }}
```
mnajdova: We should update the typings for this, currently an array is not a valid `border` value...

miroslavstastny: If we want to use visual tests for this I would propose to add one red segment with `['1px solid red', '1px solid invalid']` to verify that both values from the array are written to the styles.

miroslavstastny (approved):

kuzhelov (approved):

Issue #1324, opened 2019-05-10
Popup is closed incorrectly when click inside element


xiaolanli:
### Steps
1. click on the button, a Popup with an inner button "click me" shows up
2. click on "click me", the popup disappeared

### Expected Result
click on "click me", the popup should get smaller. 
*Notice that we put "e.preventDefault();" to the onClick handler, but it didn't trap the event. 

### Actual Result
the popup disappears

### Version
latest

### Testcase
https://codesandbox.io/s/031rl98lzw?fontsize=14

johannao76:
I would like to add that this is also causing an issue with Enter key press on the toolbar buttons in compose, so we should address both issues in the same change. Keyboard events also have the same issue on the stickers picker. The deeper issue here seems to be that stardust is using native events as opposed to react events. These don't share the same pipeline and should not be mixed. We should only use react events.

PR #1463, opened 2019-06-06
chore: add Popup issue-example-based e2e tests

kuzhelov:
This PR makes browser tests to execute against encapsulated React examples. 

With these changes it will be possible to easily introduce coverage for those problems that are reported in issues, given that issue contains a repro example - and, by that, ensure that each previously problematic scenario remains to be a working one after each new change to the code.

## Commands (yarn scripts)
- `test:e2e` (`yarn test:e2e`) - independent task: builds, serves and runs tests
- `test:e2e:serve` (`yarn test:e2e:serve`) - builds and serves site with E2E examples. Index page of this site contains link to all E2E examples

## HOW TO - add browser test with custom example
Lets suppose that one would like to introduce tests for some scenario with `<scenarioName>`. This is set of steps to follow:
- all test files should be added  to `e2e/tests` directory of the repo
- create React example file named as `<scenarioName>-example.tsx`
  - provide component to render, export it as the default one
  - to check how it renders, run `yarn test:e2e:serve` and go to `/scenario-name` URL (kebab-cased name of scenario) - or go to the index page (`/`) and search for the link to your example there.
- create test file named `<scenarioName>-test.ts`
  - to render corresponding example, use `await e2e.gotoTestCase('<scenarioName>')`. This call is usually made in `beforeEach()` test setup function
  - for safe rename refactoring, it is recommended to use `__filename` variable of NodeJS as an argument:
     - `e2e.gotoTestCase(__filename)`
     - this `gotoTestCase` function will trim `-(example|test).tsx?` postfix from file name, to leave only `<scenarioName>` part

## EXAMPLE - Popup tests
As a working example, this PR contains couple of `Popup` tests for previously discovered and addressed issues:
- https://github.com/stardust-ui/react/issues/557 (menu item's popup is not closing when another menu item is clicked)
- https://github.com/stardust-ui/react/issues/1079 (incorrect handling of ESC key)
- https://github.com/stardust-ui/react/issues/1324 (incorrect click handling when popup content is clicked)
  - ~~is disabled for now, as this test case doesn't pass~~
  - ~~will be enabled with #1482, where fix is introduced~~


e2e/tests/popupClickHandling-test.ts
```diff
@@ -0,0 +1,23 @@
+import { selectors } from './popupClickHandling-example'
+
+const popupTrigger = `#${selectors.triggerButtonId}`
+const popupContent = `.${selectors.popupContentClass}`
+const popupContentButton = `#${selectors.popupContentButtonId}`
+
+// https://github.com/stardust-ui/react/issues/1324
+xdescribe('Popup - on content click', () => {
```
kuzhelov: disabled for now - will be enabled once corresponding fix (#1482) will arrive

docs/src/routes.tsx
```diff
@@ -33,11 +33,25 @@ import PopupsPrototype from './prototypes/popups'
 import IconViewerPrototype from './prototypes/IconViewer'
 import MenuButtonPrototype from './prototypes/MenuButton'
 import AlertsPrototype from './prototypes/alerts'
+import { Routes as BrowserTestRoutes } from '../../e2e'
+import { match } from 'react-router'
+
+type E2EProps = {
```
layershifter: Can we use `RouteComponentProps` instead?

```suggestion
type E2EProps = RouteComponentProps<{ exampleName: string }>
```

e2e/tests/popupClickHandling-example.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const selectors = {
+  triggerButtonId: 'trigger',
+  popupContentClass: Popup.slotClassNames.content,
+  popupContentButtonId: 'content-button',
+}
+
+const PopupExample = () => {
+  const [show, setShow] = React.useState(true)
+  return (
+    <React.Fragment>
```
layershifter: It seems that `Fragment` is cruft there

e2e/tests/popupEscHandling-example.tsx
```diff
@@ -0,0 +1,33 @@
...
+              onAdd: item => `${item} has been selected.`,
+            }}
+          />
+        </React.Fragment>
+      ),
+    }}
+  />
+)
+
+export default PopupExample
```
layershifter: Should name somehow match filename?

docs/src/routes.tsx
```diff
@@ -33,11 +33,25 @@ import PopupsPrototype from './prototypes/popups'
...
+  match: match<{
+    exampleName: string
+  }>
+}
+
+const E2EExample: React.FC<E2EProps> = ({ match }) => {
+  const Component = BrowserTestRoutes[match.params.exampleName]
```
layershifter:
```suggestion
  return React.createElement(BrowserTestRoutes[match.params.exampleName])
```
nit

docs/src/routes.tsx
```diff
@@ -33,11 +33,25 @@ import PopupsPrototype from './prototypes/popups'
...
+  return <Component />
+}
 
 const Router = () => (
   <BrowserRouter basename={__BASENAME__}>
     <Switch>
       <Route exact path="/maximize/:exampleName/:rtl?" component={ExternalExampleLayout} />
+      <Route exact path="/e2e/:exampleName" component={E2EExample} />
```
layershifter: I am not fully support the idea that we injecting this to existing docsite because it's clearly separated, similar case to perf examples. Can we use a separate config to bundle this?

kuzhelov: currently some of the tests are running against doc site examples - while some of them will run against custom ones, thus. it is necessary to run doc site, as we are doing it for other visual/perf tests. 

Lets defer it to future PR - currently we are not making things worse, as there are some artifacts in docs that we are using for perf tests, as an example. I've just made these e2e tests being non-distractive to the client.

e2e/tests/popupEscHandling-example.tsx
kuzhelov: actually, think that it will be way to easier if we will name example component types as just `Example` - as, essentially, there are no scenarios where specific name will be helpful, while just `Example` as name will save headache from name refactoring

docs/src/components/E2EExample.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
+import { Routes as BrowserTestRoutes } from '../../../e2e'
```
layershifter:
```suggestion
import { routes as browserTestRoutes } from '../../../e2e'
```

nit: it's just variables, so it can be camelcased

e2e/index.ts
```diff
@@ -0,0 +1,12 @@
+import { exampleUrlTokenFromFilePath } from './e2eApi'
+
+const e2eExampleContext = require.context('./tests/', false, /-example.tsx$/)
+
+export const Routes = e2eExampleContext.keys().reduce((acc, key) => {
```
layershifter:
```suggestion
export const routes = e2eExampleContext.keys().reduce((acc, key) => {
```

nit: it's just variable, so it can be camelcased

e2e/tests/popupClickHandling-example.tsx
```diff
@@ -0,0 +1,35 @@
...
+export const selectors = {
+  triggerButtonId: 'trigger',
+  popupContentClass: Popup.slotClassNames.content,
+  popupContentButtonId: 'content-button',
+}
+
+const PopupClickHandlingExample = () => {
+  const [show, setShow] = React.useState(true)
+  return (
+    <>
```
layershifter: Sorry, I was not clear there: we don't need a fragment there at all because there is a single child

kuzhelov: got your point, thanks - yes, was just copied from the issue's example. Thanks!

docs/src/routes.tsx
layershifter: We definitely should rethink it, too many responsibilities for docs

However, as I see it will be included to production build? Can we use `ENV` condition?

kuzhelov: let me take a look on that..

kuzhelov: for prod we can remove this route, but it _might_ still be included (depends on config). I will add corresponding condition, like we are doing for prototype routes, and conclude on that

e2e/server/app.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <Flex column hAlign="center">
+      <Header as="h1" content="E2E Test Examples" />
+      <List
+        items={_.keys(routes).map(testExampleUrl => ({
+          key: testExampleUrl,
+          header: <Link to={testExampleUrl}>{_.startCase(testExampleUrl)}</Link>,
+        }))}
+      />
+    </Flex>
+  </>
```
layershifter, line 20: image

Suggest to change to:

```
  <Flex column hAlign="center">
    <Header as="h1" content="E2E Test Examples" />
    <List
      items={_.keys(routes).map(testExampleUrl => ({
        key: testExampleUrl,
        header: { as: Link, to: testExampleUrl, content: _.startCase(testExampleUrl) },
      }))}
    />
  </Flex>
```

e2e/tests/popupClickHandling-example.tsx
```diff
@@ -0,0 +1,33 @@
...
+  popupContentClass: Popup.slotClassNames.content,
+  popupContentButtonId: 'content-button',
+}
+
+const PopupClickHandlingExample = () => {
+  const [show, setShow] = React.useState(true)
+  return (
+    <Popup
+      trigger={
+        <Button id={selectors.triggerButtonId} icon="expand" onClick={() => setShow(true)} />
```
layershifter: image

```suggestion
        <Button id={selectors.triggerButtonId} content='trigger' icon="expand" onClick={() => setShow(true)} />
```

nit: Can we add `content`? Otherwise this example looks like broken

e2e/tests/popupInMenu-example.tsx
```diff
@@ -0,0 +1,36 @@
...
+    items={items.map((item, index) => render =>
+      render(item, (MenuItem, props) => renderItem(MenuItem, props, index)),
+    )}
+  />
+)
+
+const renderItem = (MenuItem, props, index) => {
+  return (
+    <Popup
+      trigger={<MenuItem id={selectors.menuItemId(index)} {...props} />}
```
layershifter, line 31:
```suggestion
      key={props.key}
      trigger={<MenuItem id={selectors.menuItemId(index)} {...props} />}
```

Missing `key`, it's a second case with propTypes error. Can we add a check somehow that there are no any errors in console?

You can take a look on `performBrowserTest` in `build/gulp/tasks/test-projects.ts`

e2e/server/app.tsx
kuzhelov: another proof of our unsafe API for shorthand as React element..

e2e/tests/popupInMenu-example.tsx
kuzhelov: yes, will add this check, thanks!

layershifter (approved):
🚀

PR #1464, opened 2019-06-07
fix(MenuItem): icon only active state in Teams theme

### LINKED ISSUE #1465: Teams theme: <MenuItem iconOnly active> not highlighted
### Steps
Have iconOnly active menu item

### Expected Result
it should be highlighted

### Actual Result
no highlight 💣

### Version
0.30.0

### Testcase
https://codesandbox.io/embed/activemenuitem-pv1xm

mnajdova:
Fix https://github.com/stardust-ui/react/issues/1465

Menu with icon only, didn't show any active state.

## Teams default theme
Before:
image
After:
image

## Teams dark theme
Before:
image
After:
image

## Teams high-contrast theme
Before:
image
After:
image

# Question
**Not sure whether the icon in hc should be outline/filled. On focus it is filled, so I added the same logic here.  I may revert this changes.**


packages/react/src/themes/teams/components/Menu/menuItemStyles.ts
```diff
@@ -325,11 +325,14 @@ const menuItemStyles: ComponentSlotStylesInput<MenuItemPropsAndState, MenuVariab
       }),
 
       // active styles
-      ...(active &&
-        (primary
-          ? {
-              ...(iconOnly && { color: 'inherit' }),
+      ...(active && {
+        ...(iconOnly && {
+          color: v.iconOnlyColorActive,
```
kuzhelov, line 330: just to be sure - could you, please, suggest where original `inherit` value had resulted in suboptimal styles?

docs/src/examples/components/Menu/commonScreenerSteps.ts
```diff
@@ -2,23 +2,27 @@ import { Menu } from '@stardust-ui/react'
 
 interface StepsOptions {
   vertical?: boolean
+  startIdx?: number
+  endIdx?: number
```
layershifter: `startItem` & `endItem`?

packages/react/src/themes/teams/components/Menu/menuItemStyles.ts
mnajdova: In team's theme the `iconOnly` `Menu` should always have the brand color for the active items. The inherit color is black in default theme.

kuzhelov:
@codepretty, could you, please, suggest what is prescribed by Teams design spec for the case @mnajdova had mentioned in the description? Thank you!

> Not sure whether the icon in HC should be outline/filled. On focus it is filled, so I added the same logic here. I may revert this changes.

kuzhelov (approved):

layershifter (approved):

PR #1469, opened 2019-06-07
chore(package): update Enzyme

layershifter:
Cut from #1459.

### Problem

`ThemeProvider` was part of produced tree by Enzyme, it means that we should always should to call `.find()` method to get actual `Button` component. Also, with `fela` upgrade `ThemeProvider` become stateless component and it was impossible to use `.state()` method of Enzyme anymore.

### Solution

Use `wrappingComponent` component option of `shallow`/`mount` that was introduced in new versions of Enzyme.

https://github.com/airbnb/enzyme/blob/master/docs/api/shallow.md#shallownode-options--shallowwrapper

#### Before update

```jsx
    <ThemeProvider theme={{...}} overwrite={false}>
      <Button disabled={true} onClick={[Function: mockConstructor]} as="button" accessibility={[Function: buttonB
ehavior]}>
        <button className="ui-button" disabled={true} onClick={[Function]} onFocus={[Function]} role={[undefined]
} aria-disabled={true} />
      </Button>
    </ThemeProvider>
```

#### After update

```jsx
    <Button disabled={true} onClick={[Function: mockConstructor]} as="button" accessibility={[Function: buttonBeh
avior]}>
      <button className="ui-button" disabled={true} onClick={[Function]} onFocus={[Function]} role={[undefined]}
aria-disabled={true} />
    </Button>
```


packages/react/test/specs/commonTests/isConformant.tsx
```diff
@@ -56,7 +54,7 @@ export default (Component, options: Conformant = {}) => {
   const componentType = typeof Component
 
   const helperComponentNames = [
-    ...[ThemeProvider, FelaTheme, Ref, RefFindNode],
```
layershifter, line 59: - `ThemeProvider` is not needed anymore as we use `wrappingComponent`
- `FelaTheme` is not needed because we use `ThemeContext` directly #1163

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -1075,7 +1075,7 @@ describe('Dropdown', () => {
 
       expect(wrapper.find(Dropdown).state('searchQuery')).toBe('foo')
 
-      wrapper.setProps({ children: <Dropdown items={items} search searchQuery="bar" /> })
```
layershifter, line 1078: This update also eliminates this craziness: previously we should update whole tree...

packages/react/test/specs/components/Portal/Portal-test.tsx
```diff
@@ -1,23 +1,10 @@
...
-    const TestPortal = props => withProvider(<Portal content={<p />} {...props} />)
-
-    const wrapperWrapper = mount(<TestPortal {...props} />, options)
-    const wrapper = wrapperWrapper.find(Portal)
-
-    return {
-      wrapper: wrapperWrapper,
-      portal: wrapper,
-    }
-  }
```
layershifter, line 19: This stuff was removed to make tests more clear

bmdalex (approved):

packages/react/test/specs/commonTests/isConformant.tsx
```diff
@@ -408,7 +405,7 @@ export default (Component, options: Conformant = {}) => {
           'data-simulate-event-here': true,
         }
 
-        const component = mount(<Component {...props} />).childAt(0)
```
layershifter, line 411: We don't need this because `ThemeProvider` is now outside the tree

Issue #1391, opened 2019-05-24
Popups don't render properly when popup is larger than window size


stuartlong:
## Bug Report
When popups that are rendered as dialogs, i.e. rendered as a top level element underneath the document body rather than rendered underneath their trigger elements, are larger than the window size it can cause some weird rendering issues. 

### Steps
1. Create a popup whose content is rather tall.
2. Anchor its trigger to the bottom of the window
3. Shrink the window so that it's 

### Expected Result
I'd expect as much of the popup to be visible as possible and the overflow to be handled gracefully.

### Actual Result
The bottom dissappears underneath the window

### Version
0.31.0 

### Testcase
[Fork, update, and replace this pen to show the bug]:

Easiest to repro if you pop-out the sandbox.
https://codesandbox.io/s/stardustuiexample-qeku8

bmdalex:
![Screen Recording 2019-06-03 at 14 21 15](https://user-images.githubusercontent.com/5442794/58801519-02212680-860b-11e9-8d2d-1f4129274683.gif)
@stuartlong - is this the correct repro?

> I'd expect as much of the popup to be visible as possible and the overflow to be handled gracefully.

Can you offer more details on the expectation? What does your Design specification say for this?

For the case where the Popup content does not fully fit in the viewport, would you expect it to stay on the part of the trigger with more room and:

1. be cut off suggesting the user to enlarge the size of the window?
**more likely to implement in Stardust**

2. shrink and display its content in a scrollable?
**less likely to implement in Stardust**

fyi @codepretty

stuartlong:
Yup, that's exactly the issue.

Let me get more details on the expected design. But for now, I'd at least expect option 1, so that the user isn't totally confused why when they clicked on a button that no popup appeared.

bmdalex:
### Short term fix - boolean prop to pin the content using `align` and `position` props

The short term fix will be to add a property that will turn off the auto positioning; we had requests where clients prefer to have the content positioned no matter if there is enough room in the viewport.

The prop could be called: `pinned`, `positionFixed` (conflicting with CSS props a bit)

e.g.
```jsx
<Popup align="start" position="above" pinned ... />
```

will create a `Popup` that will always be placed on top of the trigger aligned to the start (auto positioning will be turned off).

### Midterm - improve auto positioning and flipping algorithm

[...]

PR #1471, opened 2019-06-07
feat(popup): unstable_pinned prop

bmdalex:
### Description

There are certain use cases where our users need to position `Popup` content or `Dropdown` items using a specific alignment and position (via `align` and `position` props).

By default, `Popup` content or `Dropdown` items reposition themselves automatically on both axis if the content / items do not fit in the viewport. However, the algorithm has a few issues, outlined in #1391 so this PR addresses that with a short term fix presented [here](https://github.com/stardust-ui/react/issues/1391#issuecomment-498592986):

The short term fix is to add the `unstable_pinned` prop that **disables automatic repositioning of the component**; it will always be placed according to the values of `align` and `position` props, regardless of the size of the component, the reference element or the viewport.

e.g.
```jsx
<Popup align="start" position="above" pinned ... />
```

will create a `Popup` that will always be placed on top of the trigger aligned to the start (auto positioning will be turned off). Same applies for `Dropdown`.

### Screenshot outlining the difference in `Popup`:
Screen Recording 2019-06-07 at 14 23 45

### DIFF
--- a/packages/react/src/components/Popup/Popup.tsx
+++ b/packages/react/src/components/Popup/Popup.tsx
@@ -386,7 +386,7 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
     rtl: boolean,
     accessibility: ReactAccessibilityBehavior,
   ): JSX.Element {
-    const { align, position, offset, target } = this.props
+    const { align, position, offset, target, unstable_pinned } = this.props
 
     return (
       <Popper
@@ -395,6 +395,7 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
         position={position}
         offset={offset}
         rtl={rtl}
+        unstable_pinned={unstable_pinned}
         targetRef={target ? toRefObject(target) : this.triggerRef}
         children={this.renderPopperChildren.bind(this, popupPositionClasses, rtl, accessibility)}
       />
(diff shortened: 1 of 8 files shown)


bmdalex:
fyi @kuzhelov @layershifter @stuartlong

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -386,7 +386,7 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
     rtl: boolean,
     accessibility: ReactAccessibilityBehavior,
   ): JSX.Element {
-    const { align, position, offset, target } = this.props
+    const { align, position, offset, target, unstable_pinned } = this.props
```
mnajdova, line 390: Why is this not added in the propTypes? Isn't thing going to end up as unhandled prop?

mnajdova: We are not using the unhandledProps in the Popup component so it won't be a problem, but still, let's add this prop in the propTypes as well

bmdalex:
closing for now as @layershifter will introduce a better solution

packages/react/src/components/Popup/Popup.tsx
bmdalex: closing for now as @layershifter will introduce a better solution, if we reopen I will add the Proptype, I missed it for sure

thanks for reviewing @mnajdova

bmdalex: and we reopened it 😆 added the proptype

bmdalex:
in the end we need this PR, added few more changes

layershifter (approved):

PR #1476, opened 2019-06-09
perf(themes): use minified normalize.css in Teams theme

layershifter:
Related to #1311.

### Problem

Currently `normalizeCSS` is a string that can't be properly minified by bundlers:

image

### Solution

Use minified version.

image

#### Props
- less bundle size
- less memory usage (comments are removed)

### ~~Im~~possible solution

Convert `normalizeCSS` to an object, however the main problem is a different API in Fela's `renderStatic()`:
```jsx
// object type style
renderer.renderStatic({
  boxSizing: 'border-box',
  margin: 0
}, 'html,body')
```

I don't think that effort that we will spent there will have much gain for us.


CHANGELOG.md
```diff
@@ -34,6 +34,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix Teams' theme list item end media styles @mnajdova ([#1448](https://github.com/stardust-ui/react/pull/1448))
 - Fix the order of the fela plugin @mnajdova ([#1461](https://github.com/stardust-ui/react/pull/1461))
 
+### Performance
+- Use minified version of `normalize.css` @layershifter  ([#1476](https://github.com/stardust-ui/react/pull/1476))
```
miroslavstastny: You also upgraded normalize from 8.0.0 to 8.0.1. Should it be mentioned in the changelog?

miroslavstastny (approved):

CHANGELOG.md
layershifter: Updated the entry 👍

PR #1477, opened 2019-06-09
docs(CodeSnippet): add copy, support SSR

layershifter:
### Add copy

codesnippet-copy

`label` on `CodeSnippet` is now clickable and implements copy action.

### `CopyToClipboard` moved to `@stardust-ui/docs-components`

This component is quite common thing and exists in SUIR codebase, too. `CopyToClipboard` was rewritten to hooks, `useCopyToClipboard` is also provided.

---

Fixes also issues with SSR that blocked usage in SUIR.


miroslavstastny (commented):
> I also implemented selection for whole code by one click, it's expected behavior in the most cases when you want just copy code.

I find this extremely unfriendly:
- I want to select just a part, when I do it (sometimes) the whole file is selected
- I want to select a single word bu double click, sometimes the whole file is selected
- I want to deselect text, so I single click but the whole file is selected...

miroslavstastny (approved):
review+ for clipboard and SSR, I do not like single-click-to-select-the-whole-file

PR #1479, opened 2019-06-10
feat(Checkbox): improve accessibility, tests and use in docs

jurokapsiar:
Adds aria-disabled
Adds accessibility handling and conformance unit tests
Replaces the use of input checkbox with the Checkbox component in the docs

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentProps/ComponentProps.tsx
+++ b/docs/src/components/ComponentDoc/ComponentProps/ComponentProps.tsx
@@ -55,7 +55,7 @@ export default class ComponentProps extends React.Component<any, any> {
       <Flex column gap="gap.small">
         <Flex.Item styles={{ display: 'block', verticalAlign: 'middle' }}>
           <Flex gap="gap.medium">
-            <Checkbox checked={!!activeDisplayName} onClick={this.handleToggle} label="Props" />
+            <Checkbox checked={!!activeDisplayName} onChange={this.handleToggle} label="Props" />
             <ComponentPropsComponents
               activeDisplayName={activeDisplayName}
               displayNames={displayNames}
--- a/docs/src/components/ComponentDoc/ComponentProps/ComponentProps.tsx
+++ b/docs/src/components/ComponentDoc/ComponentProps/ComponentProps.tsx
@@ -6,7 +6,7 @@ import { getComponentGroup } from 'docs/src/utils'
 import ComponentTableProps from '../ComponentPropsTable'
 import ComponentPropsComponents from './ComponentPropsComponents'
 import ComponentPropsDescription from './ComponentPropsDescription'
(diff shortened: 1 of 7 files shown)


sophieH29 (approved):

packages/react/src/lib/accessibility/Behaviors/Checkbox/checkboxBehavior.ts
```diff
@@ -3,26 +3,29 @@ import { Accessibility } from '../../types'
...
  * Adds role='checkbox'. This allows screen readers to handle the component as a checkbox button.
  * Adds attribute 'aria-checked=true' based on the property 'checked'.
+ * Adds attribute 'aria-disabled=true' based on the property 'disabled'.
  * Adds attribute 'tabIndex=0' to 'root' component's part.
  */
 const checkboxBehavior: Accessibility<CheckboxBehaviorProps> = props => ({
   attributes: {
     root: {
       'aria-checked': props.checked,
+      'aria-disabled': props.disabled,
```
miroslavstastny, line 20: I am wondering why we have different logic implemented here vs `buttonBehavior` vs `menuItemBehavior`. Should we use just this simple mapping or anything else and reuse it everywhere?

docs/src/components/ComponentDoc/ComponentProps/ComponentProps.tsx
```diff
@@ -54,27 +54,14 @@ export default class ComponentProps extends React.Component<any, any> {
...
-            <Flex.Item>
-              <ComponentPropsComponents
-                activeDisplayName={activeDisplayName}
-                displayNames={displayNames}
-                onItemClick={this.handleComponentClick}
-                parentDisplayName={displayName}
-              />
-            </Flex.Item>
+          <Flex gap="gap.medium">
+            <Checkbox checked={!!activeDisplayName} onClick={this.handleToggle} label="Props" />
```
layershifter:
```suggestion
            <Checkbox checked={!!activeDisplayName} onChange={this.handleToggle} label="Props" />
```

packages/react/src/lib/accessibility/Behaviors/Checkbox/checkboxBehavior.ts
jurokapsiar: yes, this is overcomplicated in the case of button and menu item. we will simplify those.

packages/react/src/lib/accessibility/Behaviors/Checkbox/checkboxBehavior.ts
```diff
@@ -3,26 +3,29 @@ import { Accessibility } from '../../types'
 
 type CheckboxBehaviorProps = {
   checked: boolean
+  disabled: boolean
```
sophieH29: shouldn't be disabled an optional prop? If it is not passed then we don't render `aria-diabled`

PR #1480, opened 2019-06-10
feat(Toolbar): add popup prop

### LINKED ISSUE #1410: Toolbar - popup
Folloup of #1408.

A toolbar item can open a popup:
image

miroslavstastny:
Closes #1410.

A Toolbar item can now have a popup:
image

### API
`popup` prop was added to the `ToolbarItem`. It is a shorthand for `Popup` component, accepting all config valid for Popup with the following exceptions:
- `accessibility` defaults to `popupFocusTrapBehavior` (but can be overriden)
- `children` prop cannot be used, it is always reset by the `ToolbarItem`
- `trigger` prop cannot be used, it is force-set to the `ToolbarItem`'s button

`Toolbar` is a controlled component, it does not internally handle items' state. It is consumer's responsibility to handle `MenuItem:active` prop based on the `Popup` state.


packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -58,6 +59,8 @@ export interface ToolbarItemProps
    * @param {object} data - All props.
    */
   onBlur?: ComponentEventHandler<ToolbarItemProps>
+
+  popup?: ShorthandValue
```
sophieH29: missed comment

docs/src/examples/components/Toolbar/Types/ToolbarExampleEditor.shorthand.tsx
```diff
@@ -44,8 +86,37 @@ const ToolbarExampleShorthand = () => {
...
         { key: 'divider1', kind: 'divider' },
-        { key: 'highlight', icon: { name: 'highlight', outline: true } },
-        { key: 'font-color', icon: { name: 'font-color', outline: true } },
+        {
+          key: 'highlight',
+          icon: { name: 'highlight', outline: true },
+          active: highlightOpen,
+          popup: {
+            content: {
+              content: (
```
bmdalex, line 95: I understand why we do this but we desperately need to simplify this API :(

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -101,6 +105,20 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>, ToolbarItemS
...
     )
+
+    if (popup) {
+      return Popup.create(popup, {
+        defaultProps: {
+          accessibility: popupFocusTrapBehavior,
+        },
+        overrideProps: {
+          trigger: renderedItem,
+          children: undefined,
```
bmdalex: this is a bit unexpected, why is it needed? conflict with `trigger` prop that cannot be defined at the same time?

bmdalex (approved):
looks good generally, couple of comments

packages/react/src/components/Toolbar/ToolbarItem.tsx
mnajdova: Are we trying to override this if something is specified?

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -101,6 +105,20 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>, ToolbarItemS
         {childrenExist(children) ? children : Icon.create(icon)}
       </ElementType>
     )
+
+    if (popup) {
```
mnajdova, line 128: Somehow I feel that the `popup` should not be treated as a regular shorthand prop, just because it is tightly in correlation with the element it is used in. At least we can introduce correct typings for allowing users to specify only the things that make sense, and allow it only to be defined as object/string. I really think it's confusing now that we will allow the user to specify some jsx element there - but it won't be used as `content` of the `Popup`, but rather will be treated as `Popup` and it will end up being modified if the `trigger` or `children` props are used. Let me know what are your thoughts.

mnajdova: We can use the `Omit` generic to omit the children and trigger props from the `Popup`

packages/react/src/components/Toolbar/ToolbarItem.tsx
miroslavstastny: Yup, `children` collides with `trigger`, so we need to force-reset them.
Changed typings for `popup` prop to make it clearer and added comment.

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -101,6 +120,20 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>, ToolbarItemS
...
+      return Popup.create(popup, {
+        defaultProps: {
+          accessibility: popupFocusTrapBehavior,
+        },
+        overrideProps: {
+          trigger: renderedItem,
+          children: undefined, // force-reset `children` defined for `Popup` as it collides with the `trigger
```
mnajdova, line 135:
```suggestion
          children: undefined, // force-reset `children` defined for `Popup` as it collides with the `trigger`
```

mnajdova: Still not sure why we need this explicitly, let me fetch the branch and play with it a bit.

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -58,6 +65,13 @@ export interface ToolbarItemProps
    * @param {object} data - All props.
    */
   onBlur?: ComponentEventHandler<ToolbarItemProps>
+
+  /**
+   * Attaches a `Popup` component to the ToolbarItem.
+   * Accepts all props as a `Popup`, except `trigger` and `children`.
+   * Sets `accessibility` to `popupFocusTrapBehavior` by default.
+   */
+  popup?: Omit<PopupProps, 'trigger' | 'children'>
```
mnajdova:
```suggestion
  popup?: Omit<PopupProps, 'trigger' | 'children'> | string
```
String can still be valid type here, we can map it to the content of the Popup.

packages/react/src/components/Toolbar/ToolbarItem.tsx
mnajdova: I don't see any issue without this being applied, as the user cannot specify children, and we are not passing it. Probably I am missing something..

miroslavstastny: - Without this, user can define the item as follows:
```jsx
{
  key: 'font-color',
  icon: { name: 'font-color', outline: true },
  popup: {
    content: { content: <Input icon="search" placeholder="Search..." /> },
    children: <Button content="invalid" />,
  },
},
```
- `ToolbarItem` uses `Popup.trigger` to 'point to itself'.
- But `children` has priority over `trigger` in `Popup` implementation.
- That means that **invalid** button would be rendered instead of the **font-color** icon.

That's the reason for the force-reset.

mnajdova: Got it, I thought the typings would be enough, but I see your point.

docs/src/examples/components/Toolbar/Types/index.tsx
```diff
@@ -4,6 +4,11 @@ import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
 
 const Types = () => (
   <ExampleSection title="Types">
+    <ComponentExample
+      title="Toolbar can contain a popup"
+      description="Toolbar item can open a popup. See `Popup` component for more details."
```
mnajdova: Would be nice if the Popup can be a link here.

mnajdova (approved):
Looks good 👍 ! Don't forget to add changelog entry.

PR #1481, opened 2019-06-10
feat(AccessibilityBehaviors): Define types for accessibility behaviors

### LINKED ISSUE #1472: Define types for accessibility behavior props
### Problem description
A clear and concise description of what the problem is. Ex. I have an issue when [...]

### Proposed solution
A clear and concise description of what you want to happen. Add any considered drawbacks.

### MVP

sophieH29:
Closes https://github.com/stardust-ui/react/issues/1472

Some improvements:

```ts
'aria-disabled': !_.isNil(props['aria-disabled'])
        ? props['aria-disabled']
        : !!props['disabled']	
        ? true	
        : undefined,
```

was simplified just to this 
```ts
'aria-disabled': props.disabled,
```
Since `unhandledProps` go after accessibility attributes in components and can override it in any case, no need to check if the attribute was passed.


packages/react/src/lib/accessibility/Behaviors/Image/imageBehavior.ts
```diff
@@ -8,12 +8,18 @@ import { Accessibility } from '../../types'
...
     },
   },
 })
 
 export default imageBehavior
+
+type ImageBehaviorProps = {
+  /** Alternative text. */
+  alt?: string
+  'aria-label'?: string
```
layershifter, line 24: Can we reuse `AccessibilityAttributes`?

```suggestion
  'aria-label'?: AccessibilityAttributes['aria-label'],
```

packages/react/src/lib/accessibility/Behaviors/Popup/popupBehavior.ts
```diff
@@ -85,3 +82,21 @@ const getAriaAttributeFromProps = (
...
+  /** Indicates if popup's trigger is disabled. */
+  disabled?: boolean
+  /** Element which triggers popup */
+  trigger?: {
+    props?: {
+      /** Element type. */
+      as?: string
+      href?: string
+      [attributeName: string]: string
+    }
```
layershifter, line 98: image

packages/react/src/lib/accessibility/Behaviors/Menu/menuItemBehavior.ts
```diff
@@ -39,12 +37,8 @@ const menuItemBehavior: Accessibility = (props: any) => ({
       'aria-label': props['aria-label'],
       'aria-labelledby': props['aria-labelledby'],
       'aria-describedby': props['aria-describedby'],
-      'aria-disabled': !_.isNil(props['aria-disabled'])
-        ? props['aria-disabled']
-        : !!props['disabled']
-        ? true
-        : undefined,
```
layershifter, line 46: Please mention this simplification in PR description

layershifter (commented):
❤️ 

Please fix CI and reuse types 👍

packages/react/src/lib/accessibility/Behaviors/Image/imageBehavior.ts
layershifter: or `& Pick<AccessibilityAttributes, 'aria-label'>`

layershifter (approved):

layershifter:
Do not forget about `CHANGELOG.md`: we definitely should mention changes for `aria-disabled`

PR #1483, opened 2019-06-10
feat(AutoControlledComponent): add getAutoControlledStateFromProps()

layershifter:
Extracted from #1449.

### Problem

We can't use `getDerivedStateFromProps()` without hacks in components that extend `AutoControlledComponent`, see #1416 for more details.

This issue is caused by inheritance of `AutocontrolledComponent` by another component that uses `getDerivedStateFromProps()`. In this case `getDerivedStateFromProps()` method from `AutocontrolledComponent` is silently overridden and logic for state changes will be broken.

### Solution

Provide another method similar to `getInitialAutoControlledState()`, it will be `static getAutoControlledStateFromProps()` that will behave like `getDerivedStateFromProps()`. Validations were added to `AutoControlledComponent`, so you will receive an error with wrong usage.

## How and when to use `getAutoControlledStateFromProps`?

When you need `getDerivedStateFromProps()` in component that extends `AutoControlledComponent`: you should use `getAutoControlledStateFromProps()` instead  of `getDerivedStateFromProps()`.

API of `getAutoControlledStateFromProps(props: P, state: S): Partial<S> | null` is same as in `getDerivedStateFromProps()`: https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops


packages/react/src/lib/AutoControlledComponent.tsx
```diff
@@ -169,14 +178,19 @@ export default class AutoControlledComponent<P = {}, S = {}> extends UIComponent
...
+      ...state,
+      ...initialAutoControlledState,
+      autoControlledProps,
+      getAutoControlledStateFromProps,
+    }
   }
 
   static getDerivedStateFromProps(props, state) {
-    const { autoControlledProps } = state
+    const { autoControlledProps, getAutoControlledStateFromProps } = state
```
layershifter, line 190: React calls `getDerivedStateFromProps()` without `this`: https://github.com/facebook/react/blob/2b93d686e359c7afa299e2ec5cf63160a32a1155/packages/react-test-renderer/src/ReactShallowRenderer.js#L560

So we can't use `this` to get `getAutoControlledStateFromProps()` and should keep it in the state:
 https://codesandbox.io/s/elastic-framework-w3kfk

packages/react/src/lib/AutoControlledComponent.tsx
```diff
@@ -75,11 +75,20 @@ export default class AutoControlledComponent<P = {}, S = {}> extends UIComponent
...
     const state = _.invoke(this, 'getInitialAutoControlledState', this.props) || {}
 
     if (process.env.NODE_ENV !== 'production') {
-      const { defaultProps, name, propTypes } = this.constructor as any
+      const { defaultProps, name, propTypes, getDerivedStateFromProps } = this.constructor as any
+
+      // require usage of getAutoControlledStateFromProps()
+      if (getDerivedStateFromProps !== AutoControlledComponent.getDerivedStateFromProps) {
+        /* eslint-disable-next-line no-console */
+        console.log(
```
kuzhelov: lets make it a error log - because this scenario will make a real issue to the client

layershifter: Overlooked this, great catch 👍

packages/react/src/lib/AutoControlledComponent.tsx
```diff
@@ -169,14 +178,19 @@ export default class AutoControlledComponent<P = {}, S = {}> extends UIComponent
       return acc
     }, {})
 
-    this.state = { ...state, ...initialAutoControlledState, autoControlledProps }
+    this.state = {
+      ...state,
+      ...initialAutoControlledState,
+      autoControlledProps,
+      getAutoControlledStateFromProps,
```
kuzhelov, line 185: why do we need to store it in component's state, given that this is a static? we may just use field for that

kuzhelov: oh, I see: https://github.com/stardust-ui/react/pull/1483/files#r292032236

packages/react/src/lib/AutoControlledComponent.tsx
```diff
@@ -185,7 +199,27 @@ export default class AutoControlledComponent<P = {}, S = {}> extends UIComponent
...
-    return newState
+    // Due to the inheritance of the AutoControlledComponent we should call its
+    // getAutoControlledStateFromProps() and merge it with the existing state
+    if (getAutoControlledStateFromProps) {
+      const computedState = getAutoControlledStateFromProps(props, {
+        ...state,
+        ...newStateFromProps,
+      })
+
+      // However, we should follow the idea of getDerivedStateFromProps() and return only modified
```
kuzhelov: nit: we may safely remove `However` from the comment

packages/react/src/lib/AutoControlledComponent.tsx
```diff
@@ -185,7 +199,27 @@ export default class AutoControlledComponent<P = {}, S = {}> extends UIComponent
       return acc
     }, {})
 
-    return newState
+    // Due to the inheritance of the AutoControlledComponent we should call its
+    // getAutoControlledStateFromProps() and merge it with the existing state
+    if (getAutoControlledStateFromProps) {
+      const computedState = getAutoControlledStateFromProps(props, {
```
kuzhelov, line 205: not tested? Am I missing something?

kuzhelov (approved):

mnajdova (approved):

PR #1487, opened 2019-06-11
fix(styles): update definition of `FontFace` and rename `FontFaceStyle`

### LINKED ISSUE #1454: Not possible to define fontFace with local()
## Bug Report

I am trying to define `@font-face` with `local()`:

```css
  @font-face { 
    font-family: MyHelvetica; 
    src: local("Helvetica Neue Bold"), local("HelveticaNeue-Bold"),
        url(MgOpenModernaBold.ttf);
    font-weight: bold; 
  } 
```

It will not be properly handled if I will try to use `paths` and will cause infinite loop:

```js
{
  name: "Segoe UI Local",
  paths: [
     'Segoe UI Light',
     'SegoeUI-Light',
     'Segoe WP Light',
   ],
}
```

The infinite loop will be caused because Fela in `renderer.renderFont()` expects `files`:
> It must have one of the following file extensions: .woff, .woff2, .eof, .ttf or .svg

#### Workaround

We are passing `font.style` in `this.topLevelFelaRenderer.renderFont(font.name, font.paths, font.style)` so I can do this:

```jsx
  {
    name: "Segoe UI Local",
    paths: [],
    style: {
      localAlias: [
        'Segoe UI Light',
        'SegoeUI-Light',
        'Segoe WP Light',
      ],
    }
  },
```

However, it will cause TS error `Error:(8, 7) TS2322: Type 'string[]' is not assignable to type 'string'.` because we expect plain strings in `style`.

### Solution

Expose `localAliases` in `FontFace` type and pass it down Fela renderer.

### Version
0.32.0

layershifter:
Fixes #1454.

# BREAKING CHANGES

Our `FontFace` definition haven't matched Fela [`renderFont()`](http://fela.js.org/docs/api/fela/Renderer.html#renderfont):

```js
{ name: "Segoe UI Local", style: {} },
// vs
{ name: "Segoe UI Local", props: {} },
```

So when are passing `localAlias` it looks that we are passing it to `style` what it is completely weird:

```js
{ name: "Segoe UI Local", style: { localAlias: 'Segoe UI' } },
```

## Migration

This PR contains:
- rename of `FontFaceStyle` to `FontFaceProps`
- rename of `style` to `props` on `FontFace` type

```diff
-import { FontFaceStyle } from '@stardust-ui/react'
+import { FontFaceProps } from '@stardust-ui/react'
```

```diff
const fontFaces: FontFaces = [
  {
    name: 'Segoe UI',
    paths: [],
-   style: { fontWeight: 400 },
+   props: { fontWeight: 400 },
  },
]
```

### DIFF
--- a/packages/react/src/themes/types.ts
+++ b/packages/react/src/themes/types.ts
@@ -547,19 +547,19 @@ export interface Renderer extends FelaRenderer {}
 // Fonts
 // ========================================================
 
-export interface FontFaceStyle {
+export interface FontFaceProps {
   fontStretch?: string
   fontStyle?: string
   fontVariant?: string
   fontWeight?: number
+  localAlias?: string | string[]
   unicodeRange?: string
 }
 
 export interface FontFace {
   name: string
-  localAliases?: string | string[]
   paths: string[]
-  style: FontFaceStyle
+  props: FontFaceProps
 }
 
 export type FontFaces = FontFace[]
--- a/packages/react/src/themes/types.ts
+++ b/packages/react/src/themes/types.ts
@@ -552,13 +552,13 @@ export interface FontFaceStyle {
   fontStyle?: string
   fontVariant?: string
   fontWeight?: number
-  localAlias?: string
+  localAlias?: string | string[]
   unicodeRange?: string
 }
 
 export interface FontFace {
   name: string
-  localAliases?: []
+  localAliases?: string[]
   paths: string[]
   style: FontFaceStyle
 }
--- a/packages/react/src/themes/types.ts
+++ b/packages/react/src/themes/types.ts
@@ -552,13 +552,12 @@ export interface FontFaceStyle {
   fontStyle?: string
   fontVariant?: string
   fontWeight?: number
-  localAlias?: string | string[]
   unicodeRange?: string
 }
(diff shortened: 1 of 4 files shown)


packages/react/src/themes/types.ts
```diff
@@ -558,6 +558,7 @@ export interface FontFaceStyle {
 
 export interface FontFace {
   name: string
+  localAliases?: []
```
layershifter: Fela uses `localAlias`, should I use it?

miroslavstastny (approved):

bmdalex (approved):

packages/react/src/themes/types.ts
```diff
@@ -547,19 +547,19 @@ export interface Renderer extends FelaRenderer {}
...
 // ========================================================
 
-export interface FontFaceStyle {
+export interface FontFaceProps {
   fontStretch?: string
   fontStyle?: string
   fontVariant?: string
   fontWeight?: number
-  localAlias?: string
+  localAlias?: string | string[]
```
layershifter, line 555: According to Fela definitions: `localAlias?: string | Array<string>`

PR #1488, opened 2019-06-11
feat(Tree): integrate ARIA HTML design pattern

silviuaavram:
https://www.w3.org/TR/wai-aria-practices-1.1/examples/treeview/treeview-2/treeview-2b.html

Added ARIA attributes to tree, treeitem and titles depending on their role. Items that are leaves will get `treeitem` role on the `<a>` (and role `none` on `<li>`) tag while expandable items will get `treeitem` role on the parent `<li>`.

Focus is handled differently: if leaf, focus will be on the `<a>`. If not, focus will be on the parent `<li>`.

Adapted UTs to match new specification. Also behavior UTs.

Not possible: to add focus tests for Left-Right behavior. I cannot focus anything on the tree with Enzyme.

WIP: adjust the behavior prop types.


CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add 'data-is-focusable' attribute to `attachmentBehavior` @sophieH29 ([#1445](https://github.com/stardust-ui/react/pull/1445))
 - Improve accessibility for `Checkbox` @jurokapsiar ([1479](https://github.com/stardust-ui/react/pull/1479))
 - Add `unstable_pinned` prop to `Popup` and `Dropdown` components @Bugaa92 ([#1471](https://github.com/stardust-ui/react/pull/1471))
+- Implement `Tree` ARIA design pattern from HTML point of view @silviuavram ([#1488](https://github.com/stardust-ui/react/pull/1488))
```
sophieH29: I would use the name of PR, "HTML point of view" 😄

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -106,28 +105,42 @@ class TreeItem extends UIComponent<WithAsProp<TreeItemProps>> {
...
+      e.stopPropagation()
+
+      _.invoke(this.props, 'onTitleClick', e, this.props)
+    },
+    collapseOrReceiveFocus: e => {
+      const { items, open } = this.props
+
+      e.preventDefault()
+
+      if (e.currentTarget !== e.target && items && items.length) {
```
sophieH29, line 124: can you please add a comment to explain this condition, so it would be easier to get through in future

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -106,28 +105,42 @@ class TreeItem extends UIComponent<WithAsProp<TreeItemProps>> {
...
 
   actionHandlers = {
-    getFocusFromParent: e => {
-      const { open } = this.props
-      if (open) {
+    performClick: e => {
+      e.preventDefault()
+      e.stopPropagation()
+
+      _.invoke(this.props, 'onTitleClick', e, this.props)
```
sophieH29, line 116: shouldn't `performClick` be the same as `handleClick` function?

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -176,15 +188,18 @@ class TreeItem extends UIComponent<WithAsProp<TreeItemProps>> {
...
-        {childrenExist(children) ? children : this.renderContent()}
-      </ElementType>
+      <Ref innerRef={this.itemRef}>
+        <ElementType
+          className={classes.root}
+          {...accessibility.attributes.root}
+          {...rtlTextContainer.getAttributes({ forElements: [children] })}
+          {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.root, unhandledProps)}
+          {...unhandledProps}
+          {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.root, unhandledProps)}
```
sophieH29: you've added 2 lines of key handlers, remove 198 line

packages/react/src/lib/accessibility/Behaviors/Tree/treeBehavior.ts
```diff
@@ -30,4 +34,6 @@ const treeBehavior: Accessibility = () => {
   }
 }
 
+export type TreeBehaviorProps = {}
```
sophieH29: why do you need this empty type if `props` aren't used here.

packages/react/src/lib/accessibility/Behaviors/Tree/treeItemBehavior.ts
```diff
@@ -1,25 +1,48 @@
...
-      setFocusToFirstChild: {
+      expandOrPassFocus: {
         keyCombinations: [{ keyCode: keyboardKey.ArrowRight }],
       },
     },
   },
 })
 
+export type TreeItemBehaviorProps = {
+  items?: object[]
```
sophieH29, line 45: please add comments here, saying, `items` - items which are rendered in subtree

packages/react/src/components/Tree/TreeItem.tsx
silviuaavram: sure, will add one.

packages/react/src/lib/accessibility/Behaviors/Tree/treeItemBehavior.ts
silviuaavram: do we need a specific format for the message?

sophieH29: no, this is fine

packages/react/src/lib/accessibility/Behaviors/Tree/treeItemBehavior.ts
```diff
@@ -1,25 +1,48 @@
...
         keyCombinations: [{ keyCode: keyboardKey.ArrowRight }],
       },
     },
   },
 })
 
+export type TreeItemBehaviorProps = {
+  /** The items which are rendered in subtree. */
+  items?: object[]
+  open?: boolean
```
sophieH29, line 47: and comment here too `if a subtree is open` :)

packages/react/src/lib/accessibility/Behaviors/Tree/treeTitleBehavior.ts
```diff
@@ -3,32 +3,28 @@ import * as keyboardKey from 'keyboard-key'
 import { IS_FOCUSABLE_ATTRIBUTE } from '../../FocusZone/focusUtilities'
 
 /**
+ * @description
+ * Adds role 'treeitem' if the title is a leaf node inside the tree.
+ * Adds 'tabIndex' as '0' or '-1' if the title is a leaf node inside the tree.
```
sophieH29: I don't see where is set to `-1`

packages/react/src/lib/accessibility/types.ts
```diff
@@ -30,6 +30,7 @@ export type AriaWidgetRole =
   | 'tooltip'
   | 'treeitem'
   | 'switch'
+  | 'none'
```
sophieH29, line 33: is role `none` supported by all screen readers?
from what I found in one discussion
```
JAWS supports role = "none", but NVDA does not.
Both JAWS and NVDA support role = "presentation".
Only JAWS works with role = "none presentation", even though its usage is flawed.
```

But it might be outdated, just making sure that it works

silviuaavram: I tested and it works. Thanks for pointing it out!

packages/react/src/lib/accessibility/Behaviors/Tree/treeBehavior.ts
```diff
@@ -4,16 +4,20 @@ import subtreeBehavior from './subtreeBehavior'
...
  */
-const treeBehavior: Accessibility = () => {
+const treeBehavior: Accessibility = props => {
   const subtreeBehaviorData = subtreeBehavior({})
   return {
     attributes: {
       root: {
         ...subtreeBehaviorData.attributes.root,
+        role: 'tree',
+        'aria-labelledby': props['aria-labelledby'],
```
sophieH29, line 20: since you're using props here, please add `treeBehaviorProps` type

```
type treeBehaviorProps = Pick<AccessibilityAttributes, 'aria-labelledby>
```

packages/react/src/lib/accessibility/Behaviors/Tree/treeItemBehavior.ts
silviuaavram: changed the comments, you can check.

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -106,28 +105,43 @@ class TreeItem extends UIComponent<WithAsProp<TreeItemProps>> {
...
+        e.stopPropagation()
+        this.itemRef.current.focus()
+      } else if (open) {
         e.stopPropagation()
-        this.titleRef.current.focus()
+        _.invoke(this.props, 'onTitleClick', e, this.props)
       }
     },
-    setFocusToFirstChild: e => {
+    expandOrPassFocus: e => {
```
sophieH29, line 132: nit: `expandOrPassFocusToSubtree` - wouldn't be more meanigful?

kuzhelov: same sort of comment - as a client that is consuming this API, I don't understand what `passFocus` means. Would suggest shorter form of what @sophieH29 has suggested: `expandOrFocusSubtree`

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -106,28 +105,43 @@ class TreeItem extends UIComponent<WithAsProp<TreeItemProps>> {
...
-    getFocusFromParent: e => {
-      const { open } = this.props
-      if (open) {
+    performClick: e => {
+      e.preventDefault()
+      e.stopPropagation()
+
+      _.invoke(this.props, 'onTitleClick', e, this.props)
+    },
+    collapseOrReceiveFocus: e => {
```
kuzhelov, line 118: may be even shorter: `collapseOrFocus`

kuzhelov: general thoughts on current `actionHandlers` API. As a client, frankly, it is completely unobvious for me how I should respond on `collapseOrFocus` action - specifically, what drives my confusion is this `or` part. What is the principle based on which I should select one or the other? It turns out that it is based on `open` prop value - but how I would know that? 

Ok, and then the question for me, as a client, is - I have provided all the props to accessibility behavior, so I might expect that it has access to all of them. Then, why it doesn't just expose `collapse` or `focus` actions, because I expect that inner implementation might just check `open` prop I've provided to it - and then I, as a client, won't be required to solve this puzzle on my side.

Really think that we should adjust implementation of action handlers a bit, to make this decision for the client (not this PR - just general thoughts). This will result in much more convenient API

packages/react/test/specs/behaviors/treeItemBehavior-test.tsx
```diff
@@ -0,0 +1,49 @@
+import { treeItemBehavior } from 'src/lib/accessibility'
+
+describe('TreeItemBehavior.ts', () => {
```
kuzhelov: lets omit extension from the test name

packages/react/test/specs/behaviors/treeItemBehavior-test.tsx
```diff
@@ -0,0 +1,49 @@
+import { treeItemBehavior } from 'src/lib/accessibility'
```
kuzhelov, line 1: suggest to introduce e2e tests for Tree's _focus_ behavior - will be glad to assist if necessary.

packages/react/src/components/Tree/TreeItem.tsx
silviuaavram: @jurokapsiar @sophieH29 @kolaps33 We can follow up.

packages/react/test/specs/behaviors/treeItemBehavior-test.tsx
silviuaavram: already created task for that, thanks!

packages/react/src/components/Tree/TreeItem.tsx
jurokapsiar: agree, action handlers can be extended to reflect this kind of situation - let's discuss how the API would look like

sophieH29 (approved):

packages/react/src/components/Tree/TreeItem.tsx
jurokapsiar: https://github.com/stardust-ui/react/issues/1502

PR #1489, opened 2019-06-11
docs(Knobs): fix registerKnob()

layershifter:
## Problem

There was a bug in implementation of `registerKnob()` and usage of `setKnobs()`, we used unactual value from state to check that knobs with same name will not be registered.

It caused an issue that you can't to use the same name for knob in Children and Shorthand examples, for example: https://github.com/stardust-ui/react/pull/1489/files#diff-50c810d56072655fb0b3d6c8703d58a3L7


docs/src/components/ComponentPlayground.tsx
```diff
@@ -12,6 +12,17 @@ type ComponentPlaygroundProps = {
   componentName: string
 }
 
+const NoopKnobProvider: React.FunctionComponent = props => {
+  const knobContext = React.useContext(unstable_KnobContext)
+  const noopContext = { ...knobContext, registerKnob: _.noop, unregisterKnob: _.noop }
+
+  return (
+    <unstable_KnobContext.Provider value={noopContext}>
```
layershifter, line 20: I don't have a better idea now, but I definitely want to remove this from knobs:

image

docs/src/components/ComponentDoc/ComponentExample/ComponentExampleKnobs.tsx
```diff
@@ -23,6 +23,7 @@ const ComponentExampleKnobs = createComponent({
           {({ classes }) => (
             <CodeSnippet
               className={classes}
+              copyable={false}
```
layershifter, line 26: With #1477 was introduced an ability to copy contents of snippet, but there are no sence to copy values from knobs

docs/src/components/ComponentPlayground.tsx
mnajdova: The Tooltip component was merged, please consider those changes as well.

bmdalex (approved):

docs/src/components/ComponentPlayground.tsx
layershifter: Done 👍

miroslavstastny (approved):

PR #1491, opened 2019-06-12
chore(dangerjs): Check changelog entries' format

miroslavstastny:
Some entries in CHANGELOG break the format.

This PR prevents that by adding a check to danger.js, which checks all added/modified lines which start with '-' to be of the following format:
```
- description @githubname ([#DDDD](https://github.com/stardust-ui/react/pull/DDDD))
```

### DIFF
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -24,6 +24,25 @@ const hasAddedLinesAfterVersionInChangelog = async (): Promise<boolean> => {
   return addedLines.some(line => line.ln > versionLineNumber)
 }
 
+const checkChangelogFormat = async () => {
+  const changelogDiff = await danger.git.structuredDiffForFile(CHANGELOG_FILE)
+  const addedLines = changelogDiff.chunks.reduce((acc, chunk) => {
+    const filteredLines = chunk.changes.filter(change => change.type === 'add')
+    return acc.concat(filteredLines)
+  }, [])
+
+  // +- description @githubname ([#DDDD](https://github.com/stardust-ui/react/pull/DDDD))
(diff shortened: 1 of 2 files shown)


dangerfile.ts
```diff
@@ -24,6 +24,25 @@ const hasAddedLinesAfterVersionInChangelog = async (): Promise<boolean> => {
   return addedLines.some(line => line.ln > versionLineNumber)
 }
 
+const checkChangelogFormat = async () => {
+  const changelogDiff = await danger.git.structuredDiffForFile(CHANGELOG_FILE)
+  const addedLines = changelogDiff.chunks.reduce((acc, chunk) => {
+    const filteredLines = chunk.changes.filter(change => change.type === 'add')
+    return acc.concat(filteredLines)
+  }, [])
```
layershifter: We have the same code in `hasAddedLinesAfterVersionInChangelog()`, can we extract it to `getAddedLinesInChangelog()`?

dangerfile.ts
```diff
@@ -93,6 +112,7 @@ export default async () => {
       'There are no updates provided to CHANGELOG. Ensure there are no publicly visible changes introduced by this PR.',
     )
   } else {
+    await checkChangelogFormat()
```
layershifter:
```js
hasImproperlyFormattedLinesInChangelog().then(lines => {
  lines.forEach(line => fail(`Invalid entry format in ${CHANGELOG_FILE}: >${line.content}<`))
})
```

Can we be consistent and keep all warnings and failures inside default function?

layershifter (approved):
❤️

PR #1492, opened 2019-06-12
fix(Popup): fix styling with `pointer` when it is wrapped by FocusZones

### LINKED ISSUE #1433: Teams theme: Popup pointing broken styles with focus trap
### Steps
Have a **pointing** popup with `popupFocusTrapBehavior`.
```jsx
<Popup
  pointing
  trigger={<Button icon="expand" content="Open" />}
  content="content"
/>
```

### Actual Result
Pointing beak is broken (wrong position, transparent bg):
image

### Version
0.31.0

### Testcase
https://codesandbox.io/embed/stardustuiexample-7g7wo

layershifter:
Fixes #1433.

---

## Problem

When we use behaviors for `Popup` that have FocusZones, it results in new `div` and changed DOM structure, which causes broken styles.

## Solution

Provide `wrapped` prop to `PopupContent`, so we will know if it is wrapped by FocusZones and provide correct styles. This solution is short-term and aims only to fix bug - as any refactoring that will make it simpler and will allow to avoid `wrapped` prop will contain breaking changes, which we want to allow for now.

I decided to prefer prop instead of passing down the `variables` because we don't have a proper merging of them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixes
 - Fix prop types of `Tooltip` component @kuzhelov ([#1499](https://github.com/stardust-ui/react/pull/1499))
+- Fix `Popup` styling with `pointer` when it is wrapped by FocusZones @layershifter ([#1492](https://github.com/stardust-ui/react/pull/1492))
 
 <!--------------------------------[ v0.33.0 ]------------------------------- -->
 ## [v0.33.0](https://github.com/stardust-ui/react/tree/v0.33.0) (2019-06-13)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixes
 - Fix prop types of `Tooltip` component @kuzhelov ([#1499](https://github.com/stardust-ui/react/pull/1499))
-- Fix `Popup` styling with `pointer` when it is wrapped by FocusZones @layershifter ([#1492](https://github.com/stardust-ui/react/pull/1492))
+- Fix `Popup` styling with `pointer` when it is wrapped by `FocusZones` @layershifter ([#1492](https://github.com/stardust-ui/react/pull/1492))
 
 <!--------------------------------[ v0.33.0 ]------------------------------- -->
 ## [v0.33.0](https://github.com/stardust-ui/react/tree/v0.33.0) (2019-06-13)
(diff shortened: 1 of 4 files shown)


mnajdova (approved):
Don't have better idea for now. We may actually add the wrapper prop, as it is in the menu, but we really need just a boolean here for now, so as a short term fix, I am okay with this.

miroslavstastny (approved):

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixes
 - Fix prop types of `Tooltip` component @kuzhelov ([#1499](https://github.com/stardust-ui/react/pull/1499))
+- Fix `Popup` styling with `pointer` when it is wrapped by FocusZones @layershifter ([#1492](https://github.com/stardust-ui/react/pull/1492))
```
kuzhelov: nit: lets use code formatting for `FocusZones`

packages/react/src/components/Popup/PopupContent.tsx
```diff
@@ -53,6 +53,12 @@ export interface PopupContentProps
 
   /** A ref to a pointer element. */
   pointerRef?: React.Ref<Element>
+
+  /**
+   * Indicates that PopupContent is wrapped with FocusZone. Do not use it, it used only for internal implementation and
```
kuzhelov, line 59: nit: lets use @deprecated pragma to make proper accent on that - as this pragma is commonly supported and visualised (at least) by IDEs

kuzhelov (approved):

packages/react/src/themes/teams/components/Popup/popupContentStyles.ts
```diff
@@ -8,6 +8,10 @@ const popupContentStyles: ComponentSlotStylesInput<PopupContentProps, PopupConte
     borderRadius: v.borderRadius,
     display: 'block',
 
+    ...(p.unstable_wrapped && {
+      backgroundColor: 'inherit',
```
bmdalex, line 12: Wouldn't it be enough to add the styles all the time? without the condition of `p.unstable_wrapped`? I mean maybe we can avoid this prop altogether... Or will adding these styles cause different issue?

layershifter: If we will always add these styles we will inherit `backgroundColor` from `body` when there are no wrapper...

packages/react/src/components/Popup/PopupContent.tsx
layershifter: Will do, but you accidentally tagged someone 😸

PR #1494, opened 2019-06-13
feat(docs): improve color guide

mnajdova:
This PR introduced improvements on the guides regarding colors. Things implemented:
- added new page for exploring the color scheme available
- restructure the color guide, by providing the code sample before the color scheme
- restricted the color scheme example on the colors guide, to show few colors with fading ending, and added `see more` button that will navigate to the new page
- removed `copy` functionality from the color box


kuzhelov:
see that 'copy color' functionality is still preserved for tint colors - is it intentional?

image

mnajdova:
@kuzhelov not sure if we want to remove it from everywhere, as for color palette, for designers they may actually want to copy this... On the other hand, if developer is taking a look, they may think it's okay to copy them, so maybe we should actually remove the copy to clipboard functionality from here as well. What do you think?

kuzhelov:
yes, couple of thoughts on that
- the copy-paste functionality for Design team should be preserved,
- at the same time, it is better to refrain from using copy-paste buttons in the _developers_ guide - as this will provide false intuition that colors should be just copy-pasted to styles.

Given all that, ideally I see it as two separate pages:
- developers guide page, the one we have now: no copy-paste buttons there for colors, but there are links to 
- Teams color palette page, for Design team. There is an introduction text that explicitly says that _color values should be copy pasted by devs, this is just designer's tool_ - and this text is followed by actual Teams color palette.

In that case we will also be able to save additional space on the Devs guide page (as we won't need to reference all color gradients in `All colors` section, as an example - we will be able just to omit it).

Please, let me know what do you think?

kuzhelov (approved):

docs/src/views/Colors.tsx
```diff
@@ -0,0 +1,352 @@
...
+    <ProviderConsumer
+      render={({
+        siteVariables: { colors, contextualColors, naturalColors, transparentColors },
+      }) => (
+        <DocPage title="Colors">
+          <Header as="h2">Introduction</Header>
+          <p>
+            The organizing of the colors for an application has many requirements and constraints.
+            There is a need to be intentional and functional with color use. We analyzed existing
+            frameworks and picked the best ideas from them. In Stardust, the colors mechanisms are
```
layershifter: > We analyzed existing frameworks and picked the best ideas from them.

We are great, but let's remove it

---

<pre>
The organizing of the colors for an application has many requirements and constraints. In Stardust, the colors mechanisms are completely based on the <code>siteVariables</code>. We have two concepts to work transparently when the theme switching is in play:
</pre>

My proposal to keep this in single paragraph

mnajdova: 👍

docs/src/views/Colors.tsx
```diff
@@ -0,0 +1,352 @@
...
+                {link('Color scheme', '#color-scheme')}
+              </Text>{' '}
+              - design tokens for all colors used in the application that should be appropriately
+              mapped in all themes.
+            </li>
+          </ul>
+          <p>
+            Be aware that everything that follows is our recommendation, not the requirement
+            (everything will work even if you'll decide to structure the palette and schemas
+            differently in your theme).
```
layershifter:
```jsx
<Alert info>
It's our recommendation and not the requirement, you can make own decision on the structure.
</Alert>
```

docs/src/views/Colors.tsx
```diff
@@ -0,0 +1,352 @@
...
+            This part of palette includes colors from those that are the most commonly used among
+            popular frameworks ({code('blue')}, {code('green')}, {code('grey')},{code('orange')},{' '}
+            {code('pink')}, {code('purple')}, {code('teal')}, {code('red')}, {code('yellow')}). Each
+            color includes at least ten gradients, this allows us to satisfy most common needs.
+          </p>
+          <p>
+            This decision is experienced from Material UI and allows to define more variants than by
+            using semantical naming ({code('lightest')}, {code('lighter')}, etc.). However, there is
+            no requirement for client to define all the gradient values for each color - it is just
+            enough to define those that are actually used in the app.
```
layershifter, line 143:
```diff
-  - it is just enough to define those that are actually used in the app
```

Do we really need this phrase?

docs/src/views/Colors.tsx
```diff
@@ -0,0 +1,352 @@
...
+                <ColorBox
+                  name={color}
+                  rounded
+                  size="big"
+                  value={colors[color]}
+                  copyToClipboardIcon={false}
+                />
+              </div>
+            ))}
+          </Grid>
```
layershifter, line 130: ### padding: 4rem

image

### padding: .75rem

image

I vote for second

docs/src/views/Colors.tsx
mnajdova: I wanted to emphasize on the fact that in the color palette you should have only values that are used, and not add values just so that you will have 10 values for each color.

docs/src/views/Colors.tsx
```diff
@@ -0,0 +1,352 @@
...
+            using semantical naming ({code('lightest')}, {code('lighter')}, etc.). However, there is
+            no requirement for client to define all the gradient values for each color - it is just
+            enough to define those that are actually used in the app.
+          </p>
+          <p>
+            Below there is an example of base tints of Teams' natural colors provided. Full list of
+            all gradient values for them can be found on{' '}
+            {link('Teams color palette page', '/color-palette')}.
+          </p>
+          <Grid columns={2} variables={{ gridGap: '2rem' }}>
```
layershifter: ### grid-gap: .5rem, padding: .75rem

image

docs/src/views/Colors.tsx
mnajdova: All Headers for the color are the same, would not like to add special case for these two colors.

docs/src/views/Colors.tsx
mnajdova: Ah I see, ok will make everything smaller then, to save on unnecessary used space.

docs/src/views/Colors.tsx
mnajdova: Will make all of them smaller. 👍

docs/src/views/Colors.tsx
```diff
@@ -0,0 +1,352 @@
...
+              <Provider theme={mergeThemes(themes.teams, theme)}>
+                <Text content={'Box in light theme'} />
+                <Box content={'LIGHT THEME - HOVER ME'} />
+                <Text content={'Box in high contrast theme'} />
+                <Provider theme={themes.teamsHighContrast}>
+                  <Box content={'HIGH CONTRAST THEME - HOVER ME'} />
+                </Provider>
+              </Provider>
+            )}
+          />
```
layershifter, line 281: I don't understand what happens in this example. Do we need `:hover` there? How we can make it really simple?

docs/src/views/Colors.tsx
```diff
@@ -0,0 +1,352 @@
...
+          <p>
+            To see all colors variants in the palette, follow this{' '}
+            <Text as={Link} weight="bold" content="link" color="brand" to="color-palette" />.
+          </p>
+
+          <Header as="h2" content="Color scheme" />
+          <p>
+            Now that we know how the color palette is defined, let's see how we can use these
+            values, so that they will work correctly when different themes will be applied. As part
+            of each theme, we are defining <b>color scheme</b>, which will define the design tokens
```
layershifter:
```diff
-Now that we know how the color palette is defined, let's see how we can use these
-values, so that they will work correctly when different themes will be applied. As part
-each theme, we are defining <b>color scheme</b>, which will define the design tokens
+Each our theme defines <b>color scheme</b>
```

Red text in diff looks useless...

docs/src/views/Colors.tsx
mnajdova: I wanted to show at least two states, hover seems like the most simple for example.

layershifter: We share in this examples too many concepts. 

1. '::hover' selector, is it a part related to color scheme? 
2. 
``` 
     root: ({ theme: { siteVariables } }) => {
        const colorScheme = siteVariables.colorScheme
``` 

We showing in our example direct usage of `componentStyles` (💣 ), nesting is 🙈 

```jsx
componentVariables: (siteVariables) => ({
  Box: {
    color: siteVariables.colorScheme.brand.foreground3,
    backgroundColor: siteVariables.colorScheme.brand.background,
  })
},
componentStyles: ({ variables }) => ({
  color: variables.color,
  backgroundColor: variables.backgroundColor,
})
```

Can we try to utilize real pattern?

docs/src/views/Colors.tsx
```diff
@@ -0,0 +1,352 @@
...
+            )}
+          />
+
+          <p>
+            Here is how one color scheme looks like, defined for brand in Teams light, high contrast
+            and dark themes.
+          </p>
+
+          <Fader url={'color-schemes'}>
+            <ColorSchemes
```
layershifter, line 308: image

Not sure, but it seems that this block should be last

mnajdova: Not sure I understand the question. Is there text where?

docs/src/views/Colors.tsx
mnajdova: image
Is it better now?

alinais (approved):

docs/src/views/Colors.tsx
alinais: Meaning the text is not vertically aligned. It looks like there is some space under each color text, maybe because of the first column where the text goes on 2 lines.

docs/src/views/Colors.tsx
mnajdova: image
 Final look :)

PR #1495, opened 2019-06-13
docs(ComponentExample): allow to disable html preview

layershifter:
### Problem

image

Some examples are broken.

### Why it happens?

HTML preview is generated using `react-dom` which doesn't support `createPortal()` from `react` 💣 

### Solution

Disable preview for examples that uses `React.createPortal()` in initial state.

### DIFF
--- a/docs/src/examples/components/Popup/Variations/index.tsx
+++ b/docs/src/examples/components/Popup/Variations/index.tsx
@@ -9,6 +9,7 @@ const Variations = () => (
       title="Alignment and Position"
       description="A popup can be positioned around its trigger and aligned relative to the trigger's margins. Click on a button to open a popup on a specific position and alignment."
       examplePath="components/Popup/Variations/PopupExamplePosition"
+      previewHtml={false}
     />
     <ComponentExample
       title="Offset"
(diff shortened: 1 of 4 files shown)


docs/src/examples/components/Popup/Variations/index.tsx
```diff
@@ -9,6 +9,7 @@ const Variations = () => (
       title="Alignment and Position"
       description="A popup can be positioned around its trigger and aligned relative to the trigger's margins. Click on a button to open a popup on a specific position and alignment."
       examplePath="components/Popup/Variations/PopupExamplePosition"
+      previewHtml={false}
```
mnajdova, line 12: We should also add this on the Tooltip examples as well... A bit hard to keep track of where this should be added. Can't we catch this error and do something afterwards?

layershifter: React doesn't export error codes in errors 👶 

I will try to rethink it 🤔

kuzhelov:
not completely agree that we should disable the HTML previews for these - is there any other way to solve this problem? Can we ignore this warning somehow?

I clearly recall that it was quite beneficial to see the dropdown's HTML structure during development of this component, as well as while providing bug fixes to it.

docs/src/examples/components/Popup/Variations/index.tsx
bmdalex: +1 for omitting `previewHtml` and catching error if possible

layershifter:
#1529

PR #1499, opened 2019-06-13
fix(Tooltip): prop types

kuzhelov:
Adds `open` and `defaultOpen` props that were originally missed from Tooltip's prop types.


mnajdova (approved):
Thanks for catching this!

PR #1500, opened 2019-06-14
feat(Provider): add `target` prop

layershifter:
This PR is a part of #1252. Only to unblock usage, no changelog. Will be updated with real implementation later.


kuzhelov:
why we will introduce this prop without real implementation? I mean, if this prop is necessary for the client, for some emulation scenarios, client may just wrap our `Provider` component with the additional prop (`target`) provided. This should `unblock usage`, which is the goal of this PR

PR #1505, opened 2019-06-17
fix(FocusTrapZone): Set focus into zone when inner content is lazy loaded

### LINKED ISSUE #1429: Accessibility: Focus trap zone is not properly setting focus for lazy loaded components
## Bug Report
Focus trap zone is not properly setting focus for lazy loaded components. It looks like the focus trap zone currently only sets focus on componentDidMount, should it also do this in componentDidUpdate?

### Steps
The attached codepen opens a popup with a spinner and simulates lazy loading the popup content. Note that even though we are using the popupFocusTrapBehavior, the first time you click on the "Click Me!" button, the button inside the popup will not have focus. It works the second time once the content is loaded.

Repro steps (in codepen):
- Click on the "Click Me!" button
- Hit Enter key

### Expected Result
- Should get an alert with the message "I was clicked"

### Actual Result
The popup closes

### Version
0.31.1

### Testcase
https://codesandbox.io/s/stardustuiexample-gfzhd

sophieH29:
Fixes https://github.com/stardust-ui/react/issues/1429

# BREAKING CHANGES

Renamed prop `forceFocusInsideTrap` to `forceFocusInsideTrapOnOutsideFocus`

### Improvements

- Added props `forceFocusInsideTrapOnComponentUpdate` - `false` by default
       - once `componentDidUpdate` called in FocusTrapZone and current focused element is outside the FTZ's container, it will grab focus and put it on the inner element 
- To make it work for lazy loaded content there are 2 options:
       1. trigger `componetDidUpdate` by using state change when new content is loaded:

```jsx
   const ButtonExampleEmphasis = () => {
  const loader = <Loader />;
  const [content, setContent] = React.useState(loader);

  setTimeout(() => {
    if (content === loader) {
      setContent(
        <Button content="has focus?" onClick={() => alert("I was clicked")} />
      );
    }
  }, 3000);

  return (
    <Popup
      trigger={<Button content="Click Me!" primary />}
      content={{ content }}
    />
  );
};
```
2. put the focus inside the Popup manually after content is loaded


packages/react/src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -58,18 +58,16 @@ export class FocusTrapZone extends React.Component<FocusTrapZoneProps, {}> {
...
-    if (
-      !this._root.current.contains(this._previouslyFocusedElementOutsideTrapZone) &&
-      !disableFirstFocus
-    ) {
-      this._findElementAndFocusAsync()
+  componentDidUpdate(): void {
+    const activeElement = document.activeElement as HTMLElement
+    // if after componentDidUpdate focus is not inside the focus trap, bring it back
+    if (!this._root.current.contains(activeElement)) {
+      this._bringFocusIntoZone()
```
bmdalex, line 74: I'm not sure this is a good idea. A component can update for many reasons and I am pretty sure it's not always expected behavior to bring focus back to the zone after every update.

Is see this behavior as risky because it can cause unexpected focus jumps so I'd recommend allowing the users to configure the focus zone to add / disable this behavior. Let's wait for other opinions.

sophieH29: @Bugaa92 this will only work if the active element (currently focused element) is not inside the focus trap container (for any reason) - focus should also be there once it is rendered.

jurokapsiar: as discusse doffline - one scenario where this might not work is if the popup opens another popup with focus trap. Then if the first popup updates, it might steal focus from the second popup. Can you please test this scenario?

sophieH29: @jurokapsiar I tested this scenario. Since inner popup is rendered on button click in the first popup without relation to the outside state, it doesn't trigger componentDidUpdate.
Maybe there could be another case that I am missing, I see fabric-ui is doing the same thing. Folks, let me know what you think.

layershifter: @sophieH29 can we make this behavior configurable via prop?

sophieH29: I've tried this case and it works
```jsx
<Popup> -- with lazy load state change    
      <InnerPopup> -- with lazy load state change
              <InnerInnerPopup> -- autocontrolled Popup
                      <Input>
```
fyi @Bugaa92 @jurokapsiar

sophieH29: @layershifter agree, updated

packages/react/src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -58,22 +59,24 @@ export class FocusTrapZone extends React.Component<FocusTrapZoneProps, {}> {
...
+    }
   }
 
   render(): JSX.Element {
-    const { className, forceFocusInsideTrap, ariaLabelledBy } = this.props
+    const { className, forceFocusInsideTrapOnOutsideFocus, ariaLabelledBy } = this.props
```
jurokapsiar, line 79: should we use these new props in popupFocusTrapBehavior?

sophieH29: if we want it by default then probably set this prop to "true" in FocusTrapZone. But I am not sure if want this by default, maybe better to allow a user to do it through customizing the behavior

packages/react/test/specs/lib/FocusTrapZone-test.tsx
```diff
@@ -30,7 +30,10 @@ class FocusTrapZoneTestComponent extends React.Component<
   render() {
     return (
       <div>
-        <FocusTrapZone forceFocusInsideTrap={true} isClickableOutsideFocusTrap={false}>
+        <FocusTrapZone
+          forceFocusInsideTrapOnOutsideFocus={true}
```
layershifter:
```suggestion
          forceFocusInsideTrapOnOutsideFocus
```

layershifter (approved):

PR #1507, opened 2019-06-17
fix(Accessibility): Action handlers should respect modifiers

### LINKED ISSUE #1332: Accessibility: key actions don't account modifiers
## Bug Report
Key actions in behaviors will not account modifiers along the keyCodes.
### Steps
In a behavior
```
  keyActions: {
    root: {
      f1: {
        keyCombinations: [{ keyCode: keyboardKey.Tab, shiftKey: false }],
      },
      f2: {
        keyCombinations: [{ keyCode: keyboardKey.Tab, shiftKey: true }],
      },
    }.
   }
```
And in a component
```
  actionHandlers: AccessibilityActionHandlers = {
    f1: e => {
      console.log('F1')
    },
    f2: e => {
      console.log('F2')
    },
}
```

### Expected Result
When hitting Tab, expect log F1.
When hitting Tab + Shift, expect log F2

### Actual Result
On Tab + Shift, both F2 and F1 are logged.

### Version
0.29.1

sophieH29:
Fixes https://github.com/stardust-ui/react/issues/1332

### DIFF
--- a/packages/react/src/lib/shouldHandleOnKeys.ts
+++ b/packages/react/src/lib/shouldHandleOnKeys.ts
@@ -3,18 +3,26 @@ import * as keyboardKey from 'keyboard-key'
 import * as _ from 'lodash'
 import * as React from 'react'
 
+const shouldHandleModifier = (modifierValue: boolean, combinationValue?: boolean) => {
+  if (typeof combinationValue === 'undefined') {
+    return true
+  }
+
+  return modifierValue === combinationValue
+}
+
 const shouldHandleOnKeys = (
   event: React.KeyboardEvent,
   keysCombinations: KeyCombinations[],
 ): boolean =>
   _.some(
     keysCombinations,
-    keysCombination =>
+    (keysCombination: KeyCombinations) =>
       keysCombination.keyCode === keyboardKey.getCode(event) &&
-      (!keysCombination.altKey || event.altKey) &&
-      (!keysCombination.shiftKey || event.shiftKey) &&
-      (!keysCombination.metaKey || event.metaKey) &&
-      (!keysCombination.ctrlKey || event.ctrlKey),
+      shouldHandleModifier(event.altKey, keysCombination.altKey) &&
+      shouldHandleModifier(event.shiftKey, keysCombination.shiftKey) &&
+      shouldHandleModifier(event.metaKey, keysCombination.metaKey) &&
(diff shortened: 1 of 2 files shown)


layershifter (approved):

packages/react/src/lib/shouldHandleOnKeys.ts
```diff
@@ -3,18 +3,26 @@ import * as keyboardKey from 'keyboard-key'
 import * as _ from 'lodash'
 import * as React from 'react'
 
+const shouldHandleModifier = (modifierValue: boolean, combinationValue?: boolean) => {
+  if (typeof combinationValue === 'undefined') {
```
kuzhelov: `combinationValue === undefined` should be enough here, as `combinationValue` is guaranteed to be declared (local one)

packages/react/src/lib/shouldHandleOnKeys.ts
```diff
@@ -3,18 +3,26 @@ import * as keyboardKey from 'keyboard-key'
...
+    return true
+  }
+
+  return modifierValue === combinationValue
+}
+
 const shouldHandleOnKeys = (
```
kuzhelov, line 14: nit: may suggest to rename this method to `matchesAny`/`matchesAnyKeyCombination` - this will better reflect its semantics

packages/react/src/lib/shouldHandleOnKeys.ts
```diff
@@ -3,18 +3,26 @@ import * as keyboardKey from 'keyboard-key'
 import * as _ from 'lodash'
 import * as React from 'react'
 
+const shouldHandleModifier = (modifierValue: boolean, combinationValue?: boolean) => {
```
kuzhelov: nit: `isKeyModifiersMatch` will provide better idea on what this method is designed to check

kuzhelov (approved):

packages/react/src/lib/shouldHandleOnKeys.ts
sophieH29: I was renamed to `shouldHandleOnKeys` here https://github.com/stardust-ui/react/pull/1421 and most of the people approved. Frankly, I like it more as we know what is the expected answer we want to receive - if we need to handle keydown events. `matchesAnyKeyCombination ` - doesn't answer this question for me.

PR #1510, opened 2019-06-18
chore(e2e): add focus tests for Tree and Dropdown

silviuaavram:
2 tests for `Dropdown`: with `moveFocusOnTab` passed it checks that selection with Tab/Shift Tab will move focus to next/previous element.

2 tests for `Tree`: checks that ArrowRight moves focus from parent to first child and ArrowLeft moves focus from child to parent.


e2e/tests/dropdownMoveFocusOnTab-test.ts
```diff
@@ -0,0 +1,33 @@
+import { Dropdown } from '@stardust-ui/react'
+
+describe('Dropdown', () => {
+  const triggerButton = `.${Dropdown.slotClassNames.triggerButton}`
+  const nextButton = '#next-button'
```
kuzhelov: lets declare these ids in the `example` file, reuse it them both for `example` and `test` - this will ensure that this selector IDs are defined in the single place

e2e/tests/dropdownMoveFocusOnTab-example.tsx
```diff
@@ -0,0 +1,36 @@
...
+  `Scarlett O'Hara`,
+  'Imperator Furiosa',
+  'Bruce Banner',
+  'Peter Parker',
+  'Selina Kyle',
+]
+
+const DropdownExampleMultiple = () => (
+  <>
+    <button id="previous-button" />
```
kuzhelov: names like `previous-focusable-sibling` and `next-focusable-sibling` will help a lot to understand the test's intent (as, in fact, we don't care if it is a button or anything else, only `focusable` trait matters to us). 

As an example, this is how test's code will look like with these names:

```jsx
 await e2e.pressKey('ArrowDown') // open dropdown list
 await e2e.pressKey('Tab') // TAB from opened dropdown list

 expect(await e2e.isFocused(triggerButton)).toBe(false)

 expect(await e2e.isFocused(nextFocusableSibling)).toBe(true)
// vs  expect(await e2e.isFocused(nextButton)).toBe(true)
```

e2e/tests/dropdownMoveFocusOnTab-example.tsx
```diff
@@ -0,0 +1,36 @@
...
+  'Steven Strange',
+  'Alfred Pennyworth',
+  `Scarlett O'Hara`,
+  'Imperator Furiosa',
+  'Bruce Banner',
+  'Peter Parker',
+  'Selina Kyle',
+]
+
+const DropdownExampleMultiple = () => (
```
kuzhelov: lets name example the same way as test

e2e/tests/dropdownMoveFocusOnTab-example.tsx
```diff
@@ -0,0 +1,36 @@
...
+      items={inputItems}
+      placeholder="Select your heroes"
+      getA11ySelectionMessage={getA11ySelectionMessage}
+      noResultsMessage="We couldn't find any matches."
+    />
+    <button id="next-button" />
+  </>
+)
+
+const getA11ySelectionMessage = {
```
kuzhelov: nit: these lines are not necessary for the test, may we omit them?

e2e/tests/tree-test.ts
```diff
@@ -0,0 +1,28 @@
...
+    it('moves focus to first child from expanded parent', async () => {
+      await e2e.clickOn(treeTitle(1)) // expands first subtree and focuses it.
+      await e2e.pressKey('ArrowRight') // moves focus to first child.
+
+      expect(await e2e.isFocused(`${treeItem(1)} ${treeItem(1)}`)).toBe(true)
+    })
+
+    it('moves focus to parent from one of its children', async () => {
+      await e2e.clickOn(treeTitle(1)) // expands first subtree and focuses it.
+      await e2e.focusOn(`${treeItem(1)} ${treeItem(1)}`) // move focus to first child.
```
kuzhelov: lets add the following to comment, this will explain the CSS selector used

> // move focus to first child **of first subtree**.

e2e/tests/tree-test.ts
```diff
@@ -0,0 +1,28 @@
...
+      await e2e.pressKey('ArrowLeft') // moves focus back to parent.
```
kuzhelov: > // moves focus back to parent (first subtree).

packages/react/src/index.ts
```diff
@@ -157,8 +157,12 @@ export { default as ToolbarItem } from './components/Toolbar/ToolbarItem'
 export * from './components/Toolbar/ToolbarRadioGroup'
 export { default as ToolbarRadioGroup } from './components/Toolbar/ToolbarRadioGroup'
 
-export * from './components/Tree'
-export { default as Tree } from './components/Tree'
+export * from './components/Tree/Tree'
```
kuzhelov, line 160: 👍

kuzhelov (approved):

PR #1511, opened 2019-06-18
fix(Accordion): expanded view for content

### LINKED ISSUE #1272: Accordion not working when content is specified as an element
### Steps
Using element in the Accordion's panel's content breaks the behavior of the Accordion. Example:
```jsx
const panels = [
  {
    key: "p",
    title: "What is a point?",
    content: (
      <span>
        Use the haptic SDD circuit, then you can index the redundant pixel!
      </span>
    )
  },
  {
    key: "d",
    title: "What is a dimension of a point?",
    content: <span>We need to copy the primary USB firewall!</span>
  }
];

const AccordionExample = () => (
  <Accordion defaultActiveIndex={[0]} panels={panels} />
);
```

### Expected Result
The Accordion should still keep it's behavior (clicking title should open the panel and close the other), or at least note in the documentation that this is not supported.

### Actual Result
All panels are always open.

### Version
0.29.0

### Testcase
https://codesandbox.io/s/14mqj908n4

kuzhelov:
# TL;DR: in this PR
- fixes #1272 
- `styles` in shorthand factory's `defaultProps` now are reliably applied to **_any_** slot's element: no unexpected effects of broken styles for intrinsic elements provided to slots (e.g. `span`, 'em', etc)
- **no changes to slots' DOM tree are introduced** 
  - no regressions in screener tests (those also check for DOM structure)
  - no regressions in unit/e2e tests (which, amongst other, check accessibility behaviors)

# Accordion issue

Issue #1272.

The cause of the problem is that `Accordion` is not able to handle expanded/collapsed visual appearance for its panels in case if custom element is provided to corresponding slot, such as string content wrapped in `span`. This happens because `AccordionContent` component (default component type of the solt) was in charge of handling `expanded/collapsed` state - and with this being replaced to `span`, desired effect of `expanded/collapsed` state disappeared.

# Fix strategy
- define styles for expanded and collapsed panels as part of `Accordion` styles
- apply them to panel's element slot, such that any element rendered there will have these styles

# Problem: ensure styles applied to any slot elements
The last point of fix strategy was problematic to achieve, due to Stardust haven't provided/supported any reliable mechanism to apply styles to element in the slot.

Common strategy so far was to define `styles` value in the `defaultProps` of slot's shorthand factory. As an example, lets consider this snipper from `Loader`'s code:

```jsx
{Box.create(indicator, {
    defaultProps: {
        ...
        //  these styles won't be applied if 'indicator' slot will have, say <span /> element as its content
        styles: styles.indicator,
    },
})}
```

This approach of providing `styles` works till the moment when any non-Stardust component is provided to the slot: this component won't know how to handle `styles` value passed to it as prop and, thus, will just ignore this value (i.e., CSS styles) - and for intrinsic DOM components it will also result into `console.warning` being issued about not supported `styles` prop.

**To summarise - any non-Stardust component passed to the slot have resulted in broken styles, which have provided client with unexpected effects for many slots of Stardust componets.**

`Loader` component is just one of the examples - simple search for this pattern (`styles` in `defaultProps` for slot) will reveal lots of other places where this problem appears itself.

## Solution strategy
All components that client may pass to slots may be divided in two groups:
- Stardust components - are able to handle `styles` and `className` props
- other components: intrinsic DOM (e.g. `div`) and third-party ones - are able to handle `className` props

For Stardust components it is important to maintain order of styles applied. This fact has made approach to apply `className` to components from both groups to be unreliable: as for Stardust components it would result in race conditions for styles that CSS-in-JS engine (Fela) will apply, as order of styles would be defined by order of Fela's classes rendering.

As a result, the following strategy was taken:
- check which element should be rendered in the slot
    - if Stardust element - just provide `styles` prop to it
    - if any other element - convert `styles` to `classNames` and merge them to classes already defined by element

## Implementation
These are the tools that were used to achieve the goals of solution strategy:
- `applyStyles` function, which is introduced as a tool to apply styles to _any_ element
    - `const ElementWithStyles = applyStyles(<Element />, { .. CSS in JS styles .. })`
    - clones element with proper styling prop applied (`styles` or `className`)
- each Stardust component has been provided with boolean static prop that could be used to detect Stardust component/element (`__isStardust`)

# Benefits of the approach taken
- `styles` in shorthand factory's `defaultProps` now have effect that is expected by client
- styles are reliably applied to any slot's element
- **no changes to slots' DOM tree rendering logic are introduced** - thus, no chance of breaking changes for accessibility and client. Build tests verifies that 
  - no regressions in screener tests (those also check for DOM structure)
  - no regressions in unit/e2e tests (which, amongst other, check accessibility behaviors)

# P.S. apply styles to `Flex.Item's` child
This solution has been also applied to `Flex.Item` component, which had similar problems with applying flex styles to child element. Previosuly it was solved with ad-hoc function defined in `FlexItem` module, but it wasn't able to handle all the necessary cases reliably - now this issue is addressed as well.


layershifter (changes requested):
tenor

----

1. This breaks children API because we can't use `active` prop anymore.
2. This hides problem with applying props to React Elements, but doesn't resolve it.
3. The approach with `IS_STARDUST_PROP_NAME` is brittle and dangerous, we **should** avoid this pattern. We have already `__isChatMessage` in `Chat` and it's definitely evil thing.

jurokapsiar:
@silviuavram can you please take a look at this? does accessibility still work?

kuzhelov:
@layershifter 
> This breaks children API because we can't use active prop anymore.

Children API wasn't supported for `Accordion` component even before - it wasn't provided to the `Accordion`'s examples. 

The reason it is not working is how `activeIndex` tracking was implemented for `Accordion` - this is entirely controlled by `Accordion`'s component state, and it is not something that accordion's panel is supposed to _control_. With this component's implementation, panel is supposed to only _react_ on the `active` value that is prescribed by `Accordion`'s state - this makes such modes like `exclusive` to work as expected for `Accordion`.

image

> This hides problem with applying props to React Elements, but doesn't resolve it.

This PR wasn't aimed to resolve problem of Stardust props being applied to slot's React elements **implicitly**. 

> The approach with IS_STARDUST_PROP_NAME is brittle and dangerous, we should avoid this pattern. We have already __isChatMessage in Chat and it's definitely evil thing.

Specifically, why? what makes it to be non-working one? What are the edge cases? Even React folks using exactly the same approach to differentiate function vs class components in their core logic: https://github.com/facebook/react/blob/769b1f270e1251d9dbdce0fcbd9e92e502d059b8/packages/react-reconciler/src/ReactFiber.js#L299

This PR aims to provide functionality that guarantees that Stardust _styles_ could be reliably applied to any element (via `applyStyles` function). Also, it _uses_ this provided functionality for component slots, where styles should propagate to any element passed to the slot.

silviuaavram:
KB navi and screen reader still work.

miroslavstastny (commented):
*I have not read any PR comments to be able to make my own opinion, so I don't know how this goes together with the other comments.*

### 1. DOM element
The PR solves problem with `content: <span>Inner</span>`. Classnames are generated and applied to the span. But there is another problem - `Accordion` passes  `accordionTitleId` prop to `AccordionContent`. `span` obviously does not handle it, that means that:
- It will be added as an attribute on the span (where it makes no sense).
- Accessibility is now broken without anybody noticing.

### 2. React component
There is still a problem with using own non-stardust component inside a content:
```jsx
import React from 'react'
import { Accordion } from '@stardust-ui/react'

const P = props => {
  console.log(props)
  return <span {...props}>Use the haptic SDD circuit, then you can index the redundant pixel!</span>
}

const panels = [
  {
    key: 'p',
    title: 'What is a point?',
    content: <P />,
  },
]

const AccordionExample = () => <Accordion panels={panels} />
export default AccordionExample
```

Now both `styles` and `className` are passed to the inner `P` component, this component spreads all unhandled props on its element resulting in `styles` being spread and propTypes warning logged.

### Conclusion
While I see the problem with **React component** as an implementation bug, the problem described in **DOM element** is a real design issue. We pretend that a consumer can pass anything as `content`, but that's not true. There is a contract which must be fulfilled by the `content` - this PR solves styling, but there are others - accessibility in Accordion content, key handlers in Popup trigger...

Going back to original issue, I think the problem is in the shorthand factory: If I specify **'string'** as content, it **goes inside** content. If I specify **`<element />`** as a content, it **replaces** the content. That's unexpected, surprising and error prone.

I think that the correct solution would be to always render the passed content **inside**. If consumer needs to **replace**, she should use render callback.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -277,8 +277,8 @@ class Accordion extends AutoControlledComponent<WithAsProp<AccordionProps>, Acco
       children.push(
         AccordionContent.create(content, {
           defaultProps: {
+            styles: active ? styles.expandedContent : styles.collapsedContent,
             className: Accordion.slotClassNames.content,
-            active,
```
bmdalex, line 281: we remove `active` prop for `Accordion.Content` but we keep it for `Accordion.Title` => this brings:
- inconsistency
- confusion for user because it breaks current functionality:

```jsx
import React from 'react'
import { Accordion } from '@stardust-ui/react'

const panels = [
  {
    key: 'p',
    title: {
      active: true,
      content: 'title',
    },
    content: {
      active: true,
      content: 'content',
    },
  }
]

const AccordionExample = () => <Accordion styles={{ border: '1px dotted red' }} panels={panels} />

export default AccordionExample
```

used to render:

Screenshot 2019-06-19 at 15 53 20

and now it renders: 💣 
Screenshot 2019-06-19 at 15 53 35
**content is missing**

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -277,8 +277,8 @@ class Accordion extends AutoControlledComponent<WithAsProp<AccordionProps>, Acco
       children.push(
         AccordionContent.create(content, {
           defaultProps: {
+            styles: active ? styles.expandedContent : styles.collapsedContent,
```
bmdalex: Is this an exception or a pattern we want to start? as developers, we are definitely used to:
- write functionality based on properties
- write styles based on properties
- write accessibility behaviors based on properties

but this approach seems to suggest and emerge into a different pattern that I may formulate as:

- identify all UI states for all prop combinations of a specific component
- choose names for these UI states (e.g. `open => expandedContent`, `!open => collapsedContent`)
- write styles for the component for each of these UI states (e.g. `styles.expandedContent`, `styles.collapsedContent`)

Do we want to start this one? Are we sure it's the best idea?

layershifter:
> Children API wasn't supported for `Accordion` component even before - it wasn't provided to the `Accordion`'s examples.

Only second part is correct, we haven't implemented examples. I was able to reach the same markup using Children API: https://codesandbox.io/s/stardust-ui-example-djphs If I will add event handlers properly I will be able to have the same behavior, too.

kuzhelov:
@layershifter 

Which event handlers you are talking about? how these will be applied to child items? As an example, you might consider this code for your sandbox - suddenly, with children being added to Accordion, its mode switches to controlled one - this is absolutely not the effect client might expect from children API.

```jsx
{ /* -----> 'exclusive' prop will have no effect for Children API */ }
 <Accordion exclusive>
      <Accordion.Title content="What is a point?" />
      <Accordion.Content
        content="Use the haptic SDD circuit, then you can index the redundant pixel!"
      />

{ /* -----> this will be always active, no matter where user will click/press */ }
      <Accordion.Title active content="What is a point?" /> 
      <Accordion.Content
        active
        content="Use the haptic SDD circuit, then you can index the redundant pixel!"
      />
    </Accordion>
```

I don't see how the event handlers necessary for uncontrolled `Accordion`'s mode can be added to any _arbitrary_ children tree. Again, on why this happens is because of the way how state is handled by Accordion, and with this Children API cannot be supported (as it is `Accordion` who now controls the state of the panels).

bmdalex (commented):
I think this solution is treating just the effect and not the cause of the problem we currently have.

The bigger problem is that **we are inconsistent in the way we handle the shorthand pattern, more precisely, we should handle React Elements as primitive**. There is a PR opened by @layershifter that treats the cause and I believe solves these issues, see #1513 

Also, this solution is treating the effects just partially. Let's consider the following example, where not only the `Accordion` panel's `content` is a React Element, but the `title` as well:

```jsx
import React from "react";
import { Accordion } from "@stardust-ui/react";

const AccordionExample = () => (
  <Accordion
    panels={[
      {
        title: (
          <button style={{ background: "red", width: "200px", height: "32px" }}>
            click to expand
          </button>
        ),
        content: (
          <span style={{ background: "yellow" }}>content as element</span>
        )
      },
    ]}
    styles={{ background: "lightgray" }}
  />
);

export default AccordionExample;
```

**It will render just:**

image

**and contain 3 major issues:** 
1. the open/closed indicator is missing from the title :arrow_right: 💣 **styles are broken**
2. clicking on the title does not expand/collapse the content :arrow_right: 💣 **functionality is broken**
3. the markup does not contain required aria labels :arrow_right: 💣 **accessibility is broken**
it also contains unhandled attributes such as `contentRef`, `accordionContentId`, `accordionTitleId`  (see the markup below for more details)


```html
<div class="ui-provider__box" dir="ltr">
  <dl role="presentation" class="ui-accordion pd pt pu pv pw">
    <button
      style="background:red;width:200px;height:32px"
      class="ui-accordion__title"
      index="0"
      contentRef="[object Object]" // ???
      id="accordion-title-1130"
      accordionContentId="accordion-content-1140" // ???
    >
      click to expand
    </button>
    <span
      style="background:yellow"
      class="ui-accordion__content"
      id="accordion-content-1140"
      accordionTitleId="accordion-title-1130" // ???
    >
      content as element
    </span>
  </dl>
</div>
```

### Recommendation

1. Wrap the content of `title` and `content` props with an object with `content` as key:

```jsx
import React from "react";
import { Accordion } from "@stardust-ui/react";

const AccordionExample = () => (
  <Accordion
    panels={[
      {
        title: {
          content: (
            <button
              style={{ background: "red", width: "200px", height: "32px" }}
            >
              click to expand
            </button>
          )
        },
        content: {
          content: (
            <span style={{ background: "yellow" }}>content as element</span>
          )
        }
      }
    ]}
    styles={{ background: "lightgray", padding: "10px" }}
  />
);

export default AccordionExample;
```

that renders correct styles, markup and applies correct accessibility:
image

However, this approach is already working in current version of the code. For that reason I believe **this PR does not solve #1272** 

2. Consider #1513 as:
- big improvement in API by removing the "`content-content` hell" we're all aware of
- solution for #1272 
- solution to our problem mentioned above (**we are inconsistent in the way we handle the shorthand pattern, more precisely, we should handle React Elements as primitive**)

kuzhelov:
@Bugaa92 

Thanks for the detailed feedback! Let me, please, respond on couple of moments raised, will start by just brief overview on what client's expectations are - in this section I won't touch solution area, will just focus on what client expects.

# Client's expectations

## General expectations from slot area

Would like to start from client's expectations. Lets consider `title` slot of the Accordion component as an example. With the following markup I can clearly see what is the customisable area of this slot:

```jsx
<Accordion panels = {[
  {
    key: 'p',
    title: { styles: { border: '1px solid red' }, content: 'What is a point?' },
    content: 'Use the haptic SDD circuit, then you can index the redundant pixel!',
  },
  ...
]} />
```

image

so, this is what promised to me by slot's border. 

However, if React element passed to slot will always be treated as `children` or `content`, the effect won't be the one that I, as a client, expect - as the only area I will be able to customize is this one:

image

## Slot with custom component

This is where this problem has apparent visual/DOM effects to the client. It shows scenario of how custom components can be integrated to Stardust: consider `Attachment`, where we, as a client, would want to provide custom action button to corresponding `action` slot (this example was even presented on the conference).

Suppose that is the Material UI Button we would like to use:
```jsx
import Button from '@material-ui/core/Button';

<Button variant="contained">
   Hello World
</Button>
```

that looks the following way:

image

This is the markup that I see reasonable for client to try:

```jsx
<Attachment ... 
  action={<MaterialButton variant='contained'>X</MaterialButton>} />
```

However, this will render not the result expected - by rendering this

**WRONG STYLES**
image
image

instead of this:

**CORRECT STYLES**
image
image

And for the rendered DOM tree, this will provide an unexpected effect of "button in button".

If we will take a look again on the markup by client's eyes, **the markup is seen as the one should produce expected effect of button being rendered in the slot**, as slot is not called `actionContent` or something - this is `action` that is supposed to render _entire_ button. However, as a result there is half of Stardust button rendered, and half of custom one.

The only option this leaves client with is to use `renderComponent` pattern in the slot, which generally reduces to the following:

```jsx
(Component, props) => <MaterialButton variant='contained' {...props}>X</MaterialButton>
```

I am omitting point on the fact that it is more work than necessary required from the client, as there is more serious problem with that: this option is quite dangerous, as **`props` that are provided to `MaterialButton` are designed for _entirely different component (Stardust Button)_ - and there is only subset of `htmlProps` from these props that are safe to apply**. As simple example, `styles` prop won't be properly interpreted by slot's element, and this will result in broken styles for _all_ slots where `styles` are provided to shorthand's `defaultProp`:

```
Box.create({
  defaultProps: { styles: ... }
```

## Conclusion
In this section concluding on what are the client's expectations on how React element provided to the slot should be treated, to support composition model:
- this should be a replacement for the entire slot area, as it is promised by slot's boundaries. It is not surprising that component libraries follow this approach. 
  - For instance, **if slot is named `icon`, it is should be an _entire_ icon (not just icon _content_) replaced by element provided to the slot**: examples are Material UI's `label`/`icon` props of [BottomNavigationAction](https://material-ui.com/components/bottom-navigation/), React Bootstrap's `nextIcon/prevIcon` of [Carousel](https://react-bootstrap.github.io/components/carousel/#api), etc.
- this should be provided with _the set of safe html attributes_ - the ones that could be safely applied to any React element. Just as a side note, the very same idea of safe `htmlProps` is used by Reakit, to support composition model: https://reakit.io/docs/composition/
- props from the `render` function are not safe to be just spread to the component, as their intent was to be provided to entirely different component type.

With this, I think that these expectations largely form list of _actual_ problems that we need to fix - and with treating React element as primitive we will be able only to avoid some of them - however, apart from serious breaking change in paradigm, also at the cost of providing additional points for confusion:
- what is slot's area and why, by providing the element to the slot I am not replacing content of this slot area entirely
- scenarios of broken styles/DOM for custom components provided to the slot
- confusion about the shorthand markup syntax - JSON or JSON + JSX will provide different results for similar intents

To see what the last point is about, just take a look on the following syntax _by client's eyes_:

```jsx
<Chat items={[
  {
    message: (
    ),
    contentPosition: 'end',
    attached: 'top',
    key: 'message-id-1',
  },
  ...]}
/>
```

**Note that JSX is mixed with JSON in the above example. However, descriptivist approach prescribes to use the following form here - and this is how equivalent in JSON is expected to look like:**

```jsx
<Chat items={[
  {
    message: {
       content: "Hello" 
       author: "John Doe"
       timestamp: "Yesterday, 10:15 PM" 
       mine: true
    },
    contentPosition: 'end',
    attached: 'top',
    key: 'message-id-1',
  },
  ...]}
/>
```

Unfortunately, this won't result in desired effect - and I am pretty sure that it will be extremely confusing to the client to understand why these two forms behave differently.

Please, don't get me wrong - the last JSON form _is_, actually, the one we should get to, and I am not denying the problem of `content -> content`. The only thing I am saying is that React element is primitive won't _resolve_ it, but rather _mask_ it under the implicit assumptions of 'rendering as a content' it provides.

levithomason:
The original issue is solved by properly using the existing shorthand.  See here https://github.com/stardust-ui/react/issues/1272#issuecomment-504602359. **TL;DR** `content: <span />` should be `content: { children: <span /> }`.

I can say we will not merge `__isStardust` magic checking.  It is brittle and we need to embrace more solid solutions.

Per @miroslavstastny:
>If I specify <element /> as a content, it replaces the content. That's unexpected, surprising and error prone.
>
>I think that the correct solution would be to always render the passed content inside. If consumer needs to replace, she should use render callback.

This is exactly the case.  The easiest, intuitive, and smallest change is place elements inside the slot component as children.  If the user wants to opt-in to replacing the slot entirely, let them do so with the render callback where they are given all the pieces (element, props) necessary to do so correctly.

kuzhelov:
@levithomason, @miroslavstastny 

## Intuitive API effects to the client

>  If I specify 'string' as content, it goes inside content. If I specify <element /> as a content, it replaces the content. That's unexpected, surprising and error prone.

Cannot agree on the fact that `element will replace slot's area` is unexpected. As a client, if I would provide element to, say, `button` (not `buttonContent`) slot, I would provide some `<button />` element to it, and not just button's content:

```jsx
// this is what client expects to do
<StardustComponent button={<MyButton text="Click me" />} />
```

The same applies to slots that are named as `icon` or `image`:

```jsx
// this is what client expects to do
<StardustComponent icon={<span className='restore' />} image={<img .. />} />
```

As a proof of this point you may consider library APIs of React Bootstrap or Material UI I've referred to [in this post](https://github.com/stardust-ui/react/pull/1511#issuecomment-503985198).

## Robustness of proposed solution in this PR

> I can say we will not merge __isStardust magic checking. It is brittle and we need to embrace more solid solutions.

`__isStardust` flag is meant to mark Stardust components, and it does its role reliably. The rest of the logic also does reliable choices based on the components categorisation `Stardust` vs `others`. Note that this approach is used even in React itself, to differentiate class components from function ones: https://github.com/facebook/react/blob/769b1f270e1251d9dbdce0fcbd9e92e502d059b8/packages/react-reconciler/src/ReactFiber.js#L299

Please, just try scenarios you are interested to check for `Accordion` component, where proposed changes were applied. Also, please, post exact scenarios that you see as broken/unsolvable ones (given the Accordion's example) with this approach being applied?

## Treat element as slot's children
> The easiest, intuitive, and smallest change is place elements inside the slot component as children.

This is an easy change to make on Stardust side, agree, but this will result in huge breaking changes to the client. In addition, things won't be made more intuitive - as a proof we might consider already mentioned `Chat` case. Consider the following forms of declaring Chat items - **could you, please, say if it is expected that these two will produce entirely different effects?**

```jsx
<Chat items={[
  {
    message: (
    ),
```
VS
```jsx
<Chat items={[
  {
    message: {
       content: "Hello" 
       author: "John Doe"
       timestamp: "Yesterday, 10:15 PM" 
       mine: true
    }
```
 
Also couple of words on integrating custom components. Please, just try to do it with current code of Stardust `master`, where **there is no props splitting on 'safe' vs 'unsafe' ones, as all of them provided in single `props` object**  - in general, it simply won't work as expected by client, and at worst will even conflict with props of client's component. 

Should note that changes provided in this PR were addressing this problem of props splitting.

PR #1513, opened 2019-06-18
fix(factories): handle ReactElement as primitive

layershifter:
# BREAKING CHANGES

This PR changes how `React.Element`s are handled in shorthands, this change affects all components. To replace elements consider to use `render()` callback.

## Before
`React.Element` replaced matching slot.

```tsx
<Attachment header={<span>Document.docx</span>} />
```
```html
<div class="ui-attachment" tabindex="0" data-is-focusable="true">
  <div><span>Document.docx</span></div>
</div>
```

```jsx
<Button icon={<Icon className="bar" name="book" />} primary />
```

## After
`React.Element` will be inside matching slot.

```tsx
<Attachment header={<span>Document.docx</span>} />
// now these lines are equal
<Attachment header={{ content: <span>Document.docx</span> }} />
```
```html
<div class="ui-attachment" tabindex="0" data-is-focusable="true">
  <div><span class="ui-text"><span>Document.docx</span></span></div>
</div>
```

To keep existing behavior with replacement consider to use [Render Callback](https://stardust-ui.github.io/react/shorthand-props#render-callback-argument) to completely replace contents of slot.  However, in the most case enough to use [Object Shorthand](https://stardust-ui.github.io/react/shorthand-props#shorthand-value-as-object).

```jsx
// use object shorthand
<Button icon={{ className: 'bar', name: 'book' }} primary />
// or render callback
<Button
  icon={render =>
    render(null, (ComponentType, props) => <Icon className="foo" name="book" {...props} />)
  }
  primary
/>
```

See [Shorthand props docs](https://stardust-ui.github.io/react/shorthand-props#function-as-value) for more details.

***

## Problem

```jsx
<Popup content='Foo' /> // it's cool, I can set message 😎 
<Popup content={<i>Foo</i>} /> // layout is broken, WTF? 💣
```

Our component factories handles primitives (`string`, `number`) and `React.Element`s in different ways. This issue comes from Semantic UI React where it was a single way to modify rendered element in matching slot. But, we have `render()` callback that covers that purpose.

The most common use case is over-complicated and disguised from consumer: 
```jsx
<Popup content={{ content: <i>Foo</i>}} />
// I want just pass an element and I have to use `{ content: <i />}` 😭 
```

We can't always apply props properly to passed element, it's a black box for us:
```jsx
<Popup content={<Foo />} />
// will `Foo` handle `styles`, `pointerRef` or any other props? 🤔 
```

## Solution

When passing JSX, it should be a child of the slot component, this is the more intuitive behavior:

```jsx
<Popup content='Foo' />
<Popup content={<span>Foo</span>} /> // will have a <span> in DOM tree, but look will be same
```

I.e. we will handle `React.Element` in the same way as `string`, so the approach will be consistent.

## Shorthand without JSX

Shorthands like `icon`s and `image`s can't render `children` and have mapping to other props than `content`. That's why we should handle these cases separately and render `null` in that case. Factories were updated to handle a new `allowJSX` param in config.


docs/src/examples/components/Avatar/Variations/AvatarExampleImageCustomization.shorthand.tsx
```diff
@@ -12,21 +12,6 @@ const AvatarExampleImageCustomizationShorthand = () => (
...
-        <Icon
-          name="lock"
-          circular
-          bordered
-          variables={{ color: 'blue' }}
-          styles={{ boxSizing: 'border-box', padding: '8px' }}
-        />
-      }
-      status={{ color: 'green', icon: 'check', title: 'Available' }}
-    />
```
layershifter: Our examples should not share tricks as use cases to avoid confusions.

If we want to support icons on `Avatar` it should be an `icon` slot? 🤔 
This change produces single regression in this PR.

docs/src/examples/components/Popup/Types/index.tsx
```diff
@@ -20,11 +20,6 @@ const Types = () => (
       description="Note that if Popup is controlled, then its 'open' prop value could be changed either by parent component, or by user actions (e.g. key press) - thus it is necessary to handle 'onOpenChange' event. Try to type some text into popup's input field and press ESC to see the effect."
       examplePath="components/Popup/Types/PopupControlledExample"
     />
-    <ComponentExample
-      title="Content Wrapper"
-      description="Use 'content' prop of the Popup to set whether Popup content should be rendered with the default wrapper."
-      examplePath="components/Popup/Types/PopupContentWrapperExample"
-    />
```
layershifter, line 27: This example again show the trick as working pattern, it's definitely not related to `Types` section.

packages/react/src/components/Dropdown/DropdownSelectedItem.tsx
```diff
@@ -125,28 +126,39 @@ class DropdownSelectedItem extends UIComponent<WithAsProp<DropdownSelectedItemPr
...
+          render(image, (ComponentType, props) =>
+            Image.create(image, {
+              defaultProps: {
+                avatar: true,
+                className: DropdownSelectedItem.slotClassNames.image,
+                styles: styles.image,
+              },
+              overrideProps: props,
+            }),
+          )
```
layershifter, line 161: I am still disappointed with this.

As I see we used our own components in the wrong way, I expect that we can reuse `Image` and `Icon` components, but should we reuse `Label` component in this case? Not sure 🌵

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -441,7 +441,7 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
     const popupContentAttributes =
       accessibility.focusTrap || accessibility.autoFocus ? {} : popupWrapperAttributes
 
-    const popupContent = Popup.Content.create(content, {
+    const popupContent = Popup.Content.create(content || {}, {
```
layershifter, line 452: If we will pass `false` or `null` to this slot we will render nothing and it will break `Popper` and `Ref` components that expects children. I want to consider refactoring of this place more deeply.

packages/react/src/lib/factories.ts
```diff
@@ -230,9 +227,16 @@ function createShorthandFromValue({
...
+  if (generateKey && _.isNil(props.key)) {
+    if (valIsPrimitive) {
+      // use string/number shorthand values as the key
+      props.key = value
+    }
+
+    if (valIsReactElement) {
+      // use the key from React Element
+      props.key = (value as React.ReactElement).key
+    }
```
layershifter, line 267: I am not sure about this place, but usage like below looks understandable to me:

```jsx
<List
  items={[
    'string', // key will be 'string'
    10,  // key will be '10'
    <span key='foo' />, // key will be 'foo'
    { key: 'foo', content: <span /> }, // otherwise only this option will be available 👎 
  ]}
/>
```

packages/react/src/components/Icon/Icon.tsx
```diff
@@ -93,7 +93,7 @@ class Icon extends UIComponent<WithAsProp<IconProps>, any> {
   }
 }
 
-Icon.create = createShorthandFactory({ Component: Icon, mappedProp: 'name' })
+Icon.create = createShorthandFactory({ Component: Icon, mappedProp: 'name', primitive: true })
```
levithomason: `primitive` is not very clear to me.  What about `doesSupportJSX` or `allowJSX`, something explicit? Default to `true`.

packages/react/src/lib/factories.ts
layershifter: We decided to keep behavior with `key` definition

packages/react/src/components/Label/Label.tsx
```diff
@@ -62,9 +62,9 @@ class Label extends UIComponent<WithAsProp<LabelProps>, any> {
   static propTypes = {
     ...commonPropTypes.createCommon({ color: true }),
     circular: PropTypes.bool,
-    icon: customPropTypes.itemShorthand,
+    icon: customPropTypes.itemShorthandWithoutJSX,
```
mnajdova, line 64: Maybe we should rethink how we create these custom prop types - the user here must know that the image and icon are this special type...

packages/react/src/components/Dropdown/DropdownSelectedItem.tsx
mnajdova: Is this the only place where we implicitly pass shorthand of a shorthand? This will be issue in every case we have like this. Meaning, we have shorthand for something that is shorthand in another element..

packages/react-proptypes/src/index.ts
```diff
@@ -411,6 +411,10 @@ export const itemShorthandWithKindProp = (kindPropValues: string[]) => {
     ]),
   ])
 }
+export const itemShorthandWithoutJSX = every([
+  disallow(['children']),
+  PropTypes.oneOfType([PropTypes.func, PropTypes.number, PropTypes.object, PropTypes.string]),
```
mnajdova: Can we extract this and reuse it in both `itemShorthand` and `itemShorthandWithoutJSX`?

packages/react/src/components/Dropdown/DropdownSelectedItem.tsx
layershifter: We have this only in `Dropdown` components. I feel that it will be easier to have some custom components instead of this composition

e2e/tests/popupEscHandling-example.tsx
```diff
@@ -12,21 +12,19 @@ export const selectors = {
 const PopupEscHandlingExample = () => (
   <Popup
     trigger={<Button id={selectors.popupTriggerId} content="Open popup" style={{ margin: 50 }} />}
-    content={{
-      content: (
-        <React.Fragment>
-          <div>Hello from inner popup!</div>
+    content={
+      <>
```
mnajdova, line 16: 👍

packages/react-proptypes/src/index.ts
layershifter: These types are different, in `itemShorthand` it's `PropTypes.node`

docs/src/pages/ShorthandProps.mdx
```diff
@@ -0,0 +1,186 @@
...
+- it should return React Element as a result
+
+Thus, in its simplest form, it could be used the following way:
+
+```jsx
+<Button icon={() => <Icon name="chess rook" />} />
+```
+
+However, if it will be just about that, this function form wouldn't introduce any additional value. So, lets take a look on the scenarios where it is really helpful.
+
```
mnajdova, line 74: Not sure why we have this section if we say that it doesn't introduce any additional value.

docs/src/pages/ShorthandProps.mdx
```diff
@@ -0,0 +1,186 @@
...
+<Button
+  icon={render =>
+    render(
+      /* what to render */
+      { name: 'chess rook' },
+
+      /* how to render */
+      (ComponentType, props) => (
+        <div class="my-icon-container">
+          <i class="my-chess-rook-icon" {...props.accessibili.root} />
```
mnajdova:
```suggestion
          <i class="my-chess-rook-icon" {...props.accessibility.root} />
```

docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx
```diff
@@ -8,121 +8,103 @@ const janeAvatar = {
...
-    },
+    gutter: <Avatar {...janeAvatar} />,
+    message: (
+      <Chat.Message content="Looks good!" author="Jane Doe" timestamp="Yesterday, 10:15 PM" />
+    ),
     attached: true,
     key: 'message-id-5',
   },
   {
     gutter: { content: <Avatar {...janeAvatar} /> },
```
alinais: missed line in the refactoring process

docs/src/examples/components/Avatar/Variations/AvatarExampleImageCustomization.shorthand.tsx
mnajdova: This is already used on the client's side, so would suggest proposing fix for this, rather then just remove it...

packages/react/src/components/Icon/Icon.tsx
layershifter: Replaced with `allowsJSX`

packages/react/src/components/Label/Label.tsx
layershifter: Can you please provide more info? We already have a custom propTypes for them...

mnajdova: My point was, the user must know that the `Icon` and `Image` have different type then the other shorthands, and this is very prone to mistake in the future... Maybe we can make the component type param to the item shorthand and have one centralized place where this logic will be implemented. Other then this don't have any better idea.. What do you think?

packages/react-proptypes/src/index.ts
mnajdova: Got it, then should we add these values as well: `boolean | null | undefined | {}` too?

docs/src/examples/components/Avatar/Variations/AvatarExampleImageCustomization.shorthand.tsx
layershifter: Not sure about this, even previously it had custom `styles` what is definitely not the best option

packages/react-proptypes/src/index.ts
layershifter: Good catch, we definitely should handle `false` as a valid value:
```diff
+PropTypes.oneOf([false])
```

docs/src/examples/components/Avatar/Variations/AvatarExampleImageCustomization.shorthand.tsx
layershifter: Restored this one as we decided during offline discussion.

packages/react/src/lib/factories.ts
```diff
@@ -153,22 +173,28 @@ function createShorthandFromValue({
...
+      )
+    }
+
+    if (!allowsJSX && valIsReactElement) {
+      /* eslint-disable-next-line no-console */
+      console.error(
+        [
+          'Shorthand value must be a string|number|object|array when it is a `allowsJSX={false}`.',
+          ' Use null|undefined|boolean for none.',
+          ` Received: ${value}`,
```
levithomason, line 196: Invoke this error in the browser and make sure it makes good sense to the user.  Probably would be a good idea to show the component name, prop name, and value in the error.  Something like:

```
The shorthand prop `{prop}` was passed a JSX element but this slot only supports string|number|object|array
```

This is a rough idea, just make sure user's will be well informed and happy 👍

layershifter: `The shorthand prop for "${Component.displayName}" component was passed a JSX element but this slot only supports string|number|object|array|ReactElements.`,

The best that I was able to do because factories don't know about parent components 😭

docs/src/pages/ShorthandProps.mdx
```diff
@@ -0,0 +1,186 @@
...
+  <Popup content={<em>Awesome Popup!</em>} />
+  {/* 💡 has identical effect to the previous one */}
+  <Popup content={{ content: <em>Awesome Popup!</em> }} />
+</>
+```
+
+### Disable slot's rendering
+
+It is also possible to pass falsy values (`false`, `null` or
+`undefined`) to shorthand prop - iin that case there will be nothing rendered for the component's slot.
```
miroslavstastny:
```suggestion
`undefined`) to shorthand prop - in that case there will be nothing rendered for the component's slot.
```

docs/src/pages/ShorthandProps.mdx
```diff
@@ -0,0 +1,186 @@
...
+```jsx
+<Dropdown toggleIndicator={null} />
+```
+
+## Function as value
+
+Providing function as a shorthand value is the most involving but, at the same time, the most
+powerful option for customizing component's slot. The only requirements for this function are:
+
+- it should finish syncronously
```
miroslavstastny:
```suggestion
- it should finish synchronously
```

miroslavstastny: + I would say **must** instead of **should** for both

docs/src/pages/ShorthandProps.mdx
```diff
@@ -0,0 +1,186 @@
...
+      getData={fetch('https://some/url')}
+      /* using 'render' arg here to render loading icon as well */
+      onLoading={render('loading')}
+      onLoaded={data => render({ name: data.iconName })}
+    />
+  )}
+/>
+```
+
+### Customizing rendered tree
```
miroslavstastny, line 153: We should also add new use case now - completely replace slot content.

layershifter: Actually, the example in this section does required thing. I removed a paragraph that doesn't have any sense now, https://github.com/stardust-ui/react/commit/1fbc89a1f148398e7aa2836603a8b540dd48ffb4

packages/react/src/components/Dropdown/DropdownSelectedItem.tsx
```diff
@@ -125,28 +126,39 @@ class DropdownSelectedItem extends UIComponent<WithAsProp<DropdownSelectedItemPr
...
       defaultProps: {
         as: 'span',
         className: DropdownSelectedItem.slotClassNames.header,
         styles: styles.header,
       },
     })
+    const renderIcon = _.isNil(icon)
+      ? icon
+      : render =>
+          render(image, (ComponentType, props) =>
```
miroslavstastny: `image`?

layershifter:
```suggestion
          render(icon, (ComponentType, props) =>
```

packages/react/src/components/Popup/Popup.tsx
miroslavstastny: Are you going to create a separate issue for that?

layershifter: #1630 👍

miroslavstastny (approved):

mnajdova (approved):

yarn.lock
```diff
@@ -9,7 +9,27 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
-"@babel/core@^7.1.0", "@babel/core@^7.2.2", "@babel/core@^7.3.4":
+"@babel/core@^7.1.0", "@babel/core@^7.3.4":
```
mnajdova: Can you check again why we have changes in the yarn.lock and if we have new dependencies that require OSS scan?

layershifter: I reverted these changes because I don't why they happened 🤔 There are no any changes in any `package.json`

PR #1514, opened 2019-06-18
test(attachment):  Adding unit tests for fix done in Attachment PR

kolaps33:
Related to the PR:
https://github.com/stardust-ui/react/pull/1452

adding unit tests. 

And adding information into changed log as I missed it in previous PR.


CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Fixes
 - Fix prop types of `Tooltip` component @kuzhelov ([#1499](https://github.com/stardust-ui/react/pull/1499))
 - Fix `Popup` styling with `pointer` when it is wrapped by `FocusZones` @layershifter ([#1492](https://github.com/stardust-ui/react/pull/1492))
+- Fix `Attachment` more option button invoke correctly action on space/enter @kolaps33 ([#1452] (https://github.com/stardust-ui/react/pull/1452))
```
silviuaavram: do we need to add changelog for a unit test?

packages/react/test/specs/components/Attachment/Attachment-test.tsx
```diff
@@ -18,5 +21,32 @@ describe('Attachment', () => {
...
+        <Attachment
+          actionable
+          action={{
+            icon: 'more',
+            onClick: () => onClickButton(),
+          }}
+          onClick={onClickAttachment}
+        />,
+      )
+      const moreOptionButton = attachement.find('button')
```
silviuaavram: try to use a class selector rather than type of element.

silviuaavram: probably you need to add slot class names to `Attachment`, and use the one you will add for `action` here as class selector.

packages/react/test/specs/components/Attachment/Attachment-test.tsx
```diff
@@ -18,5 +21,32 @@ describe('Attachment', () => {
...
+
+      attachement.simulate('keydown', { keyCode: keyboardKey.Enter })
```
silviuaavram: maybe get the actual DOM element and simulate a keydown on it? instead of `attachment`, use attachment.find(`.${Attachment.className}`).simulate(...)

packages/react/test/specs/components/Attachment/Attachment-test.tsx
```diff
@@ -18,5 +21,32 @@ describe('Attachment', () => {
...
+      expect(onClickAttachment).toHaveBeenCalledTimes(1)
+
+      onClickAttachment.mockClear()
+      moreOptionButton.simulate('keydown', { keyCode: keyboardKey.Enter })
+      expect(onClickAttachment).not.toBeCalled()
+
+      // simulate the similar behavior as browser does when Enter key is pressed on button element
+      moreOptionButton.simulate('click')
+      expect(onClickButton).toHaveBeenCalledTimes(1)
```
silviuaavram: I think this part should be tested separately, if not already tested in `isConformant`. Maybe you can check with someone from Stardust whether this handler call is not already tested.

CHANGELOG.md
kolaps33: this is related to the previous PR, where I forgot to update changelog:
https://github.com/stardust-ui/react/pull/1452

packages/react/test/specs/components/Attachment/Attachment-test.tsx
kolaps33: I cannot see difference if button is found by 'type of element' or by the class. What is advantage of that? and why I should adding slot class name into Attachment only because the test?

packages/react/test/specs/components/Attachment/Attachment-test.tsx
kolaps33: why it is different, if I use attachment object which is created, or if I will find the object in DOM?
I took this approach from "ListItem-test.tsx" where it was done some time ago. 
If there is some benefit getting object from DOM, then we should consider do the change across the tests.

packages/react/test/specs/components/Attachment/Attachment-test.tsx
```diff
@@ -18,5 +21,30 @@ describe('Attachment', () => {
...
+
+    test('handleClick is executed when Enter is pressed for attachment, not for more option button', () => {
+      const onClickAttachment = jest.fn()
+      const onClickButton = jest.fn()
+      const attachment = mountWithProvider(
+        <Attachment
+          actionable
+          action={{
+            icon: 'more',
+            onClick: () => onClickButton(),
```
silviuaavram: can be `onClick: onClickButton` as the similar one below

silviuaavram (approved):

packages/react/test/specs/components/Attachment/Attachment-test.tsx
kolaps33: changed :)

packages/react/test/specs/components/Attachment/Attachment-test.tsx
```diff
@@ -18,5 +21,30 @@ describe('Attachment', () => {
     handlesAccessibility(Attachment, {
       defaultRootRole: undefined,
     })
+
+    test('handleClick is executed when Enter is pressed for attachment, not for more option button', () => {
```
mnajdova: Can we separate these in two tests (one when it should be clicked and one where it should not be clicked)?

mnajdova (approved):
Please consider comments before merging

PR #1515, opened 2019-06-18
fix(RadioGroup): First item should be tabbable by default when none of the items selected

### LINKED ISSUE #1351: RadioGroupItems not tab-accessible if NO default RadioGroupItem checked
### Steps
Use the https://stardust-ui.github.io/react/components/radio-group#item-radio-group-item-example (Radio group when no default radio item is checked) and try to navigate there with tab.

### Expected Result
The group should be navigable with tab, in order for clients to be able to select an item.

### Actual Result
The RadioGroup is not accessible at all.

### Version
0.30.0

### Testcase
https://codesandbox.io/s/gxoz7

sophieH29:
Fixes https://github.com/stardust-ui/react/issues/1351


jurokapsiar (approved):

jurokapsiar:
The arrow navigation looks good.

Since now it is possible to focus unchecked item, we need to add support for Space bar as suggested by ARIA:
```
If the radio button with focus is not checked, changes the state to checked.
Otherwise, does nothing.
Note: The state where a radio is not checked only occurs on page load.
```

PR #1518, opened 2019-06-18
feat(Toolbar): add menu prop

### LINKED ISSUE #1411: Toolbar - menu
Folloup of #1408.

A toolbar item can open a menu:
image

miroslavstastny:
Closes #1411.
A Toolbar item can now have a menu:

image

### API

Following props were added to the `ToolbarItem`:
- `menu` - depending on `kind` either shorthand for `ToolbarMenuItem` or `ToolbarMenuDivider`, each Toolbar Menu Item:
  - can have `content` and `icon`
  - can be `active` and `disabled`
  - has `onClick`, `onFocus` and `onBlur` callback
- `menuOpen` - boolean, the menu is rendered if `true`.
- `onMenuOpenChange` - callback to request `menuOpen` value change (the whole `Toolbar` is controlled component, parent must handle its state).

### Known Limitations

- accessibility to be handled separately
- no styles for active (no redlines)
- no submenu support (out of scope)

- [ ] changelog


packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -46,6 +53,19 @@ export interface ToolbarItemProps
...
+
+  /** Indicates if the menu inside the item is open. */
+  menuOpen?: boolean
+
+  /**
+   * Event for request to change 'menuOpen' value.
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props and proposed value.
+   */
+  onMenuOpenChange?: ComponentEventHandler<ToolbarItemProps>
```
miroslavstastny, line 72: Should we also hoist `onOpenChange` from `Popup` to here to have both on the same level?

docs/src/examples/components/Toolbar/Types/ToolbarExampleEditor.shorthand.tsx
```diff
@@ -49,6 +49,8 @@ const ToolbarExampleShorthand = () => {
   const [highlightOpen, setHighlightOpen] = React.useState(false)
   const [fontColorActive, setFontColorActive] = React.useState(false)
 
+  const [moreMenuOpen, setMoreMenuOpen] = React.useState(false)
```
layershifter: Can we use knobs there? 🤔

docs/src/examples/components/Toolbar/Types/ToolbarExampleEditor.shorthand.tsx
```diff
@@ -129,7 +131,36 @@ const ToolbarExampleShorthand = () => {
         { key: 'bullets', icon: { name: 'bullets', outline: true } },
         { key: 'number-list', icon: { name: 'number-list', outline: true } },
         { key: 'divider3', kind: 'divider' },
-        { key: 'more', icon: { name: 'more', outline: true } },
+        {
+          key: 'more',
+          icon: { name: 'more', outline: true },
+          active: moreMenuOpen,
+          menu: {
```
layershifter: I believe that `items` are redundant there:

```jsx
menu: [{ item }, { item }]
```

As we have `mappedArrayProp` to handle this

docs/src/examples/components/Toolbar/Types/ToolbarExampleEditor.shorthand.tsx
```diff
@@ -129,7 +131,36 @@ const ToolbarExampleShorthand = () => {
...
+            items: [
+              {
+                content: 'Quote',
+                icon: 'quote',
+                onClick: () => alert('... -> Quote'),
```
layershifter: nit:

I am not a big fun of using `alert()` in example, can we use `Event Log` instead:
https://stardust-ui.github.io/react/components/portal#types-portal

To get prettier output we can use `CodeSnippet`:

```jsx
import { CodeSnippet } from '@stardust-ui/docs-components'

<CodeSnippet label="Output log"  mode="json" value={log} />
```

docs/src/examples/components/Toolbar/Types/ToolbarExampleEditor.shorthand.tsx
```diff
@@ -129,7 +131,36 @@ const ToolbarExampleShorthand = () => {
...
+              {
+                content: 'Code snippet',
+                icon: 'code-snippet',
+                onClick: () => alert('... -> Code snippet'),
+              },
+            ],
+          },
+          menuOpen: moreMenuOpen,
+          onMenuOpenChange: (e, { menuOpen }) => {
+            console.log(`setting menu to ${menuOpen ? 'open' : 'close'}`)
```
layershifter: If we will use knobs I believe that we can't remove this.

And I think that we should never use `console.log()` in our examples because it's an item that is hidden from user until dev console is opened.

docs/src/examples/components/Toolbar/Types/ToolbarExampleMenu.shorthand.tsx
```diff
@@ -0,0 +1,35 @@
...
+          active: menuOpen,
+          menu: {
+            items: [
+              { key: 'play', content: 'Play', icon: 'play' },
+              { key: 'pause', content: 'Pause', icon: 'pause' },
+              { key: 'divider', kind: 'divider' },
+              'Without icon',
+            ],
+            onItemClick: (e, { content }) => {
+              console.log(`Click - ${content}`)
```
layershifter: Same thing about `console.log()`

docs/src/examples/components/Toolbar/Types/index.tsx
```diff
@@ -15,6 +15,11 @@ const Types = () => (
       }
       examplePath="components/Toolbar/Types/ToolbarExamplePopup"
     />
+    <ComponentExample
+      title="Toolbar can contain a menu"
+      description="Toolbar item can open a menu."
+      examplePath="components/Toolbar/Types/ToolbarExampleMenu"
```
layershifter: I believe that it's a part of `Content` section

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -135,6 +161,31 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>, ToolbarItemS
       </ElementType>
     )
 
+    const submenu = menuOpen ? (
+      <>
+        <Ref innerRef={this.menuRef}>
+          <Popper align="start" position="above" targetRef={this.itemRef}>
+            {ToolbarMenu.create(menu, {
+              overrideProps: predefinedProps => ({
+                onItemClick: (e, itemProps) => {
```
layershifter:
```suggestion
                onItemClick: (e, itemProps: ToolbarItemProps) => {
```

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -135,6 +161,31 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>, ToolbarItemS
       </ElementType>
     )
 
+    const submenu = menuOpen ? (
+      <>
+        <Ref innerRef={this.menuRef}>
+          <Popper align="start" position="above" targetRef={this.itemRef}>
+            {ToolbarMenu.create(menu, {
+              overrideProps: predefinedProps => ({
```
layershifter:
```suggestion
              overrideProps: (predefinedProps: ToolbarItemProps) => ({
```

layershifter: Are we going to move `overrideProps` to a class method?

packages/react/src/components/Toolbar/ToolbarMenu.tsx
```diff
@@ -0,0 +1,100 @@
...
+    accessibility: submenuBehavior,
+    as: 'ul',
+  }
+
+  handleItemOverrides = variables => predefinedProps => ({
+    onClick: (e, itemProps) => {
+      _.invoke(predefinedProps, 'onClick', e, itemProps)
+      _.invoke(this.props, 'onItemClick', e, itemProps)
+    },
+    variables: mergeComponentVariables(variables, predefinedProps.variables),
```
layershifter, line 65: Should this be done in a factory?

And I don't remember exactly the case, but these `variables` silently override `ToolbarItemVariables` as I remember. Are we going to fix that in this PR?

packages/react/src/components/Toolbar/ToolbarMenuDivider.tsx
```diff
@@ -0,0 +1,62 @@
...
+  static create: Function
+
+  static className = 'ui-toolbar__menudivider'
+
+  static propTypes = {
+    ...commonPropTypes.createCommon(),
+  }
+
+  static defaultProps = {
+    accessibility: defaultBehavior as Accessibility,
```
layershifter:
```suggestion
    accessibility: defaultBehavior,
```

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
```diff
@@ -0,0 +1,181 @@
...
+    icon: customPropTypes.itemShorthand,
+    onClick: PropTypes.func,
+    onFocus: PropTypes.func,
+    onBlur: PropTypes.func,
+    wrapper: PropTypes.oneOfType([PropTypes.node, PropTypes.object]),
+  }
+
+  static defaultProps = {
+    as: 'button',
+    accessibility: defaultBehavior as Accessibility,
```
layershifter:
```suggestion
    accessibility: defaultBehavior,
```

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+  static propTypes = {
+    ...commonPropTypes.createCommon(),
+    active: PropTypes.bool,
+    disabled: PropTypes.bool,
+    icon: customPropTypes.itemShorthand,
+    onClick: PropTypes.func,
+    onFocus: PropTypes.func,
+    onBlur: PropTypes.func,
+    wrapper: PropTypes.oneOfType([PropTypes.node, PropTypes.object]),
```
layershifter:
```suggestion
    wrapper: customPropTypes.itemShorthand,
```

Should it be?

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
```diff
@@ -0,0 +1,181 @@
...
+        ) : (
+          <>
+            {Icon.create(icon, { defaultProps: { xSpacing: !!content ? 'after' : 'none' } })}
+            {content}
+          </>
+        )}
+      </ElementType>
+    )
+
+    if (!wrapper) {
```
layershifter, line 131:
```suggestion
    if (_.isNil(wrapper)) {
```

Otherwise it will fail on empty string or `0`

packages/react/src/themes/teams/components/Toolbar/toolbarMenuDividerVariables.ts
```diff
@@ -0,0 +1,3 @@
+import toolbarVariables from './toolbarVariables'
+
+export default toolbarVariables
```
layershifter, line 3: I am not sure that I like this pattern because all variables are defined in a single place will affect others 🤔

packages/react/src/themes/teams/components/Toolbar/toolbarVariables.ts
```diff
@@ -46,6 +53,26 @@ export interface ToolbarVariables {
...
+  menuBorderRadius: string
+  menuBoxShadow: string
+
+  menuItemForeground: string
+  menuItemForegroundHover: string
+  menuItemBackgroundHover: string
+  menuItemBackgroundFocus: string
+  menuItemForegroundFocus: string
+  menuItemForegroundDisabled: string
+  menuItemBackgroundDisabled: string
```
layershifter, line 71: As I remember we use following naming: `PROP + SLOT + VARNAME`

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -135,6 +161,31 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>, ToolbarItemS
       </ElementType>
     )
 
+    const submenu = menuOpen ? (
```
sophieH29: what if the menu's items is an empty array?

sophieH29: nit: I would keep consistency as we have across our lib, to have it as a function `renderSubmenu = () => { ... }`

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -147,7 +198,12 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>, ToolbarItemS
       })
     }
 
-    return renderedItem
+    return (
+      <>
+        <Ref innerRef={this.itemRef}>{renderedItem}</Ref>
+        {submenu}
+      </>
```
sophieH29: based on our discussion and [ARIA spec](https://www.w3.org/TR/wai-aria-practices/examples/toolbar/toolbar.html), toolbar button and menu should be wrapped in one div so then a screen reader will work correctly. @jurokapsiar correct me if I am wrong

Another aspect, when the focus is in menu and press ESC, the menu should get closed and focus should go on the button.
In theory, this type of keyboard event should be caught by a wrapper element, which wraps a toolbar button and menu and put a focus on the button. That seems for me the most logical way of doing it. Let me know what you think.

packages/react/src/components/Toolbar/ToolbarItem.tsx
miroslavstastny: I don't think so. This is the only place where it is called, just once.

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
miroslavstastny: This is intentional. `wrapper: false` is valid.

packages/react/src/components/Toolbar/ToolbarItem.tsx
miroslavstastny: Then in renders empty submenu, consistent with `Menu`. Do you think it is incorrect?

sophieH29: I would say if there are no menu items we don't need to render it. What the use case of rendering an empty menu?

docs/src/examples/components/Toolbar/Types/ToolbarExampleEditor.shorthand.tsx
miroslavstastny: Added event log, cannot use `CodeSnippet` now - missing `minHeight`, no light variant...
Let's think about creating reusable `Log` component

docs/src/examples/components/Toolbar/Types/ToolbarExampleEditor.shorthand.tsx
miroslavstastny: Changed to knobs, but, still, this needs to be here to close the menu when it wants to be closed.

packages/react/src/themes/teams/components/Toolbar/toolbarVariables.ts
miroslavstastny: In color scheme we have `foregroundDisabled`. Here I just prefixed it with `SLOT`. Let's discuss this with @mnajdova separately.

packages/react/src/themes/teams/components/Toolbar/toolbarMenuDividerVariables.ts
miroslavstastny: This is now consistent with `Menu`. Let's discuss it separately and eventually refactor both.

packages/react/src/components/Toolbar/ToolbarMenu.tsx
miroslavstastny: Related to variables inheritance.

packages/react/src/components/Toolbar/ToolbarItem.tsx
miroslavstastny: Agree, but I think it is consumer's responsibility. `menu` can be array or object or element and we cannot check all. So I think it is better to be consistently dump than to try to be smart but inconsistent.

packages/react/src/components/Toolbar/ToolbarItem.tsx
miroslavstastny: Makes sense. I added a wrapper slot, it is rendered only if there is a menu prop defined on the item, no matter whether the menu is open ( = rendered) or closed ( = not in dom).

packages/react/src/components/Toolbar/ToolbarItem.tsx
layershifter: Let's skip it for now, we still have an access to it via `popup` shorthand

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
layershifter: `_.isNil(wrapper) || wrapper === false`?

layershifter (approved):

PR #1520, opened 2019-06-20
docs(Accessibility): Improve layout of behaviors pages

jurokapsiar:
- improve layout of behaviors pages by fixing styles and introducing markdown
- descriptions and specifications form transformed into bullet list (only in render, not for unit tests)
- replace "component's part" with "slot"

### DIFF
--- a/docs/src/components/ComponentDoc/BehaviorDescription.tsx
+++ b/docs/src/components/ComponentDoc/BehaviorDescription.tsx
@@ -6,12 +6,14 @@ const AccessibilityDescription = React.lazy(() => import('./AccessibilityDescrip
 const item = '- '
 
 const BehaviorDescription: React.FunctionComponent<{ value: string }> = ({ value }) => {
+  // doctrine has a bug where it ignores list item indicator (-) if it is in the beginning of the comment
+  // because of that, add the list item indicators after parsing
   const markdown =
     item +
     value
       .split('\n')
       .join(`\n${item}`)
-      .replace(/'(?!s )/g, '\u0060')
+      .replace(/'(?!s )/g, '\u0060') // replace ' with backtick as regexp rules with backtick would not be understandablein unit tests
   return (
     <React.Suspense fallback={<Loader />}>
       <AccessibilityDescription value={markdown} />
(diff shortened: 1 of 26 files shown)


docs/src/components/DocsBehaviorRoot.tsx
```diff
@@ -18,72 +19,59 @@ class DocsBehaviorRoot extends React.Component<any, any> {
...
-                key={keyValue}
-                className="docs-example"
-                id={_.kebabCase(variation.name)}
-                styles={exampleStyle}
-              >
-                <ComponentExampleTitle
-                  title={this.baseName(variation.name)}
-                  description={`Name: ${variation.name.replace('.ts', '')}`}
-                />
+              <>
```
layershifter: Oops, `key` is broken

```suggestion
              <React.Fragment key={keyValue}>
```

docs/src/components/DocsBehaviorRoot.tsx
```diff
@@ -18,72 +19,59 @@ class DocsBehaviorRoot extends React.Component<any, any> {
...
+                <Segment
+                  key={keyValue}
```
layershifter:
```suggestion
```

docs/src/components/ComponentDoc/BehaviorDescription.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react'
+import { Loader } from '@stardust-ui/react/src'
+
+const AccessibilityDescription = React.lazy(() => import('./AccessibilityDescription'))
+
+const item = '- '
+
+const BehaviorDescription: React.FunctionComponent<{ value: string }> = ({ value }) => {
+  // doctrine has a bug where it ignores list item indicator (-) if it is in the beginning of the comment
+  // because of that, add the list item indicators after parsing
```
layershifter, line 10: Broken parser sucks 👎 

I tried to inspect and found that `doctrine` parses only `@example` correctly:
https://github.com/eslint/doctrine/blob/master/lib/doctrine.js#L682

I also tried [`comment-parser`](https://github.com/yavorskiy/comment-parser) that currently is used in [eslint-plugin-jsdoc](https://github.com/gajus/eslint-plugin-jsdoc), but there is the same issue even on `@example` 💣

layershifter (approved):

PR #1523, opened 2019-06-20
fix(Dropdown, DropdownItem, ListItem): Updating theme files

bcalvery:
Redline match for Dropdown and DropdownItem.

before:
image
image

after:
image
image
image

image
image
image

The redlines call for a "selected indictor" (checkmark) next to any item that is currently selected in the dropdown, and that is not implemented here because this is supposed to just be a style update.
image


packages/react/src/themes/teams/components/Dropdown/dropdownItemStyles.ts
```diff
@@ -17,13 +20,51 @@ const dropdownItemStyles: ComponentSlotStylesInput<DropdownItemProps, DropdownVa
...
+          },
+        }),
       }),
     }),
   }),
+  image: ({ props: p }): ICSSInJSStyle => ({
+    margin: `${pxToRem(3)} ${pxToRem(12)} ${pxToRem(3)} ${pxToRem(4)}`,
+    // using !important because:
+    // a ' > :not(:last-child)' selector affecting the margin-right style
+    marginRight: `${pxToRem(12)} !important`,
```
bcalvery: I'm using !important because the image is being affected by a selector:
`'> *:not(:last-child)'`
Which is applying an incorrect margin-right value. It is taking precedence over my style. It may be coming from flexStyles?

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix `Popup` styling with `pointer` when it is wrapped by `FocusZones` @layershifter ([#1492](https://github.com/stardust-ui/react/pull/1492))
 - Apply custom focus style on `TreeItem` and `TreeTitle` @silviuavram ([#1506](https://github.com/stardust-ui/react/pull/1506))
 - Fix `theme` types, remove duplication @kuzhelov ([#1508](https://github.com/stardust-ui/react/pull/1508))
+- Fix `Dropdown` and `DropdownItem` styles to match redlines ([#1523] (https://github.com/stardust-ui/react/pull/1523))
```
codepretty: For style related changes, the request is to have the comment include that it's for Teams theme and an indicator for consumers to know they might need to remove style overrides with a `[redlines]` identifier. So something like

```
Fix `Dropdown` and `DropdownItem` Teams theme styles to match redlines [redlines]
```

packages/react/src/themes/teams/siteVariables.ts
```diff
@@ -17,7 +17,7 @@ export {
 // BORDER STYLES
 //
 export const borderWidth = '1px'
-export const borderRadius = '2px'
+export const borderRadius = '3px'
```
codepretty: just wanted to check and see if you meant to change this for all the instances currently using this value? should we only ever be using a 3px border radius?

bcalvery: Good catch, I don't think these even affects my changes.

packages/react/src/themes/teams-high-contrast/siteVariables.ts
```diff
@@ -6,6 +6,7 @@ export * from './colors'
 //
 // BORDER STYLES
 //
+export const borderWidth = '2px'
```
levithomason: Cruft?

packages/react/src/themes/teams/components/Dropdown/dropdownItemStyles.ts
```diff
@@ -1,12 +1,15 @@
 import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
 import { DropdownVariables } from './dropdownVariables'
-import { DropdownItemProps } from '../../../../components/Dropdown/DropdownItem'
+import DropdownItem, { DropdownItemProps } from '../../../../components/Dropdown/DropdownItem'
 import getBorderFocusStyles from '../../getBorderFocusStyles'
+import { pxToRem } from '../../../../lib'
 
 const dropdownItemStyles: ComponentSlotStylesInput<DropdownItemProps, DropdownVariables> = {
   root: ({ props: p, variables: v, theme: { siteVariables } }): ICSSInJSStyle => ({
+    minHeight: pxToRem(0),
```
levithomason: Let's just use `0`.

packages/react/src/themes/teams/components/Dropdown/dropdownItemStyles.ts
```diff
@@ -17,13 +20,51 @@ const dropdownItemStyles: ComponentSlotStylesInput<DropdownItemProps, DropdownVa
...
+    margin: `${pxToRem(3)} ${pxToRem(12)} ${pxToRem(3)} ${pxToRem(4)}`,
+    // using !important because:
+    // a ' > :not(:last-child)' selector affecting the margin-right style
+    marginRight: `${pxToRem(12)} !important`,
+  }),
+  header: ({ props: p, variables: v }): ICSSInJSStyle => ({
+    fontSize: v.listItemHeaderFontSize,
+    // if the item doesn't have content - i.e. it is header only - then it should use the content color
+    // or Should the default behavior of handling a generic list be changed, so that the "content" property
+    // is populated by default and not the header?
```
levithomason: Let's change `DropdownItem.create()` to follow suit with `ListItem.create()`.  The string should be passed to the `content` prop, instead of the `header` prop.

levithomason (approved):
👍

levithomason (approved):
If the redline is correct here, let's merge 👍

PR #1524, opened 2019-06-21
feat(Popup): show on `contextmenu`

jurokapsiar:
Often, what appears to be just menu is not only a menu. Below are two examples that are called "Menu" by the component libraries, but are actually combining Menu with other components in a Popup - Semantic UI and Fabric:

image

There are two use cases for context menu (or ... menu):
- Simple menu using only menu items
- Augumented menu that also has a search box or other elements

It makes sense to allow both and the plan to do that is:
- allow the Popup to open on contextmenu event
- reuse the Popup and create "ContextMenu" component based on this (name is not final and will be discussed as apart of the actual PR that introduces it)

Reasons to reuse Popup in Context menu:
Both create a new surface - intline or portal
Both have trigger element
Both can have similar events opening and closing: click, hover, focus (contextmenu)
Positioning - above, below
Alignment
Click outside to close
Opening children surfaces
Focus trap, auto focus functionality

What is not currently in Popup:
1) Positioning around cursor
2) Context action in popup

It makes sense to add both of these two and allow to create augumented menus using Popup.

Right click popup cannot be easily repositioned as it is attached to a fixed point (mouse position). There are two ways how the libraries or apps are handling it:
- dismiss the menu/popup when user scrolls
- do not allow scrollwhile the popup is open

Currently it is easier just to dismiss, but we will add an option to prevent scrolling: https://github.com/stardust-ui/react/issues/1579


sophieH29:
I am afraid that we are overcomplicating Popup with too many things it can do. Hence, make it hard for the consumer to figure out all the API and abilities of the component.
We might need to discuss all the trade-offs here. But how about creating `<ContextMenu>` component, it can render content using `<PopupContent>` as well.  It's very straightforward from the component's name what are the capabilities - open menu on the right click. 
@jurokapsiar & folks, what are your thoughts?

jurokapsiar:
> too many things it can do

This PR only adds the possibility to open the popup on right click. It can already open on focus, hover, click and combination of them so I don't think it is a big addition to the API. I agree that `<ContextMenu>` should be separate, but it is not part of the PR. `<ContextMenu>` can be used if you only have menu items in it. If there is more complicated structure with other focusable elements, `<Popup>` would still need to be used.

bmdalex:
> we are overcomplicating Popup

> how about creating `<ContextMenu>` component

I agree with @sophieH29 that a separate component makes sense.

@jurokapsiar can you please update the description of the PR with the problems and user scenarios we're trying to solve based on what we discussed offline?

I did a bit of research and **here is what other libraries are doing:**

### 1. Use a context menu
- [Fabric's `ContextualMenu`](https://developer.microsoft.com/en-us/fabric#/controls/web/contextualmenu)
- [Blueprint's `Context menu`](https://blueprintjs.com/docs/#core/components/context-menu)
- [Ant Design's `Dropdown` with `Menu` overlay and `contextMenu` trigger](https://ant.design/components/dropdown/#components-dropdown-demo-context-menu)

### 2. Have a select / overflow menu (that we can customize to work for right-click)
- [Carbon Design System's `Overflow Menu`](http://react.carbondesignsystem.com/?path=/story/overflowmenu--custom-trigger)
- [Vuetify's `Menu`](https://vuetifyjs.com/en/components/menus#absolute-position-without-activator)
- [Evergreen's `Select Menu`](https://evergreen.segment.com/components/select-menu/)
- [Material's `MenuList` composition](https://material-ui.com/components/menus/#menulist-composition)

Let's take the best out of these and come up with best compromise based on the user scenarios we're trying to solve.

jurokapsiar:
@Bugaa92 we had a discussion offline and the agreement was indeed to create a separate component for the context menu, but to reuse the Popup for that. This PR only includes changes necessary for the Popup, there will be a followup PR with the rest. I updated the description with the reasons.

packages/react/src/lib/accessibility/Behaviors/Popup/popupBehavior.ts
```diff
@@ -99,4 +102,6 @@ export type PopupBehaviorProps = {
     /** Element type. */
     type?: string
   }
+  /** Trigger's ressponsibility to be focusable */
+  triggerHandlesFocus?: boolean
```
sophieH29: typo -> `responsibility`

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -112,6 +112,9 @@ export interface PopupProps
   /** Element to be rendered in-place where the popup is defined. */
   trigger?: JSX.Element
 
+  /** Trigger is responsible for being focusable */
+  triggerHandlesFocus?: boolean
```
sophieH29: Jush sharing my thoughts. This `prop` seems to be only needed to define if set `tabindex` to trigger or not. Frankly, for me a bit confusing from the prop name what it is for. 
If this prop is `false`, the trigger will be set attribute `tabindex=0`, what will make it tabbable, nit just focusable.
As one of the options:
1. we can have a prop `shouldTriggerBeTabbable` and make it `true` by default
2. Or, utilize attribute `data-is-focusable`. If it set to "false" to trigger, in `popupBehavior` check it and do not set  tabindex. Though it needs proper documentation, but kind of consistency of using it (we use it for FocusZone)

Maybe other folks will have thoughts around that :)

packages/react/src/lib/accessibility/Behaviors/Popup/popupBehavior.ts
```diff
@@ -16,7 +16,9 @@ const popupBehavior: Accessibility<PopupBehaviorProps> = props => {
   return {
     attributes: {
       trigger: {
-        tabIndex: getAriaAttributeFromProps('tabIndex', props, 0),
+        ...(props.triggerHandlesFocus
```
mnajdova: What is the use case of this prop? If this is connected with the `'context'` value on the `on` prop can we use that one instead?

packages/react/src/components/Popup/Popup.tsx
jurokapsiar: I like the first option, seems cleaner

mnajdova: In my opinion the second option is better, because we already have a mechanism for this, why introduce a new one. If we continue with this new pattern, we will start adding tabbable props on different components, and it's not something we want to do...

packages/react/src/lib/accessibility/Behaviors/Popup/popupBehavior.ts
jurokapsiar: it's not only related to context value. I have renamed the prop, hopefully it is more clear now.

sophieH29 (approved):

docs/src/examples/components/Popup/Usage/PopupExampleOnMultiple.shorthand.tsx
```diff
@@ -2,18 +2,62 @@ import * as React from 'react'
...
+        content="Hello from popup on hover!"
+        on={['context', 'hover']}
+      />
+      <Popup
+        trigger={
+          <Button
+            icon="expand"
+            content="Context + Hover + Focus"
+            aria-label="Right click or hover or focus button"
+            onClick={() => alert('Click!')}
```
layershifter: Please avoid `alert()` usages in examples

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -112,6 +112,9 @@ export interface PopupProps
   /** Element to be rendered in-place where the popup is defined. */
   trigger?: JSX.Element
 
+  /** should trigger be made tabbable */
```
```suggestion
  /** Should trigger be made tabbable. */
```

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -403,6 +436,7 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
         rtl={rtl}
         unstable_pinned={unstable_pinned}
         targetRef={target ? toRefObject(target) : this.triggerRef}
+        referenceObject={this.rightClickReferenceObject}
```
layershifter: From API point these things (`targetRef` or `referenceObject`) can't be specified together, there are mutually exclusive

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -534,6 +569,33 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
...
+        right,
+        bottom,
+      }
+    }
+
+    this.rightClickReferenceObject = {
+      getBoundingClientRect,
+      clientWidth: 1,
+      clientHeight: 1,
+    }
```
layershifter: I believe that this logic can be extracted to a function like `createReferenceFromContextClick(e)`

layershifter (changes requested):
context-bug

context-bug2


1. `Popup` flies in unpredictable way
2. A new `Popup` is not closed on right click

sophieH29:
Just noticed that we have a paragraph on right-click support in documentation, would be great to update it too https://stardust-ui.github.io/react/accessibility#right-click-support

packages/react/src/components/Popup/Popup.tsx
jurokapsiar: But that means that `data-is-focusable` prop would add `data-is-focusable` attribute as well as add tabindex. It would make that prop both handled and unhandled, do we really want that?

packages/react/src/components/Popup/Popup.tsx
jurokapsiar: do you mean to move it outside of this file?

layershifter: Outside component at least...

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -112,12 +112,16 @@ export interface PopupProps
   /** Element to be rendered in-place where the popup is defined. */
   trigger?: JSX.Element
 
+  /** Should trigger be made tabbable */
```
bmdalex: nit
```suggestion
  /** Whether the trigger should be tabbable */
```

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -486,21 +522,68 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
...
+                />
+              </>
+            )}
           </>
         )}
       </Unstable_NestingAuto>
     )
   }
 
+  preventScroll(e: Event) {
```
bmdalex: pls move to a helper function in `src/lib`; this should be probably tested but can be done in another PR

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -486,21 +522,68 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
...
     if (newValue) {
       this.updateTriggerFocusableDomElement()
+      this.updateContextPosition(isOpenedByRightClick && eventArgs.nativeEvent)
     }
-    this.trySetState({ open: newValue })
+    this.trySetState({ open: newValue, isOpenedByRightClick })
     _.invoke(this.props, 'onOpenChange', eventArgs, { ...this.props, ...{ open: newValue } })
   }
 
+  isRightClick(eventArgs: any) {
```
bmdalex: same as `preventScroll`, pls extract in a helper in `src/lib`; you can define one file for helpers related to events for instance

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -69,7 +69,10 @@ const Popper: React.FunctionComponent<PopperProps> = props => {
     () => {
       destroyInstance()
 
-      if (!enabled || !targetRef.current || !contentRef.current) {
+      const targetRefCurrent = targetRef && (targetRef as React.RefObject<Element>).current
+      const reference = targetRefCurrent || (targetRef as _PopperJS.ReferenceObject)
```
bmdalex: nit:
```typescript
const reference = (targetRef as React.RefObject<Element>).current || (targetRef as _PopperJS.ReferenceObject)

**Review by bmdalex — COMMENTED** [review]
looks good in general; please take a look at my comments; also, did you manage to address @layershifter 's comments? I can still see a couple of issues he pointed out as well:
- preventing scrolling mechanism does not work when opening popups on context
- right click does not dismiss context popups; what's the expected behavior?

File: packages/react/src/components/Popup/Popup.tsx (hunk 7 above)
Comment (bmdalex) [reply packages/react/src/components/Popup/Popup.tsx]: +1 for that, it can be in the same folder `packages/react/src/components/Popup/createReferenceFromContextClick.ts`

**Comment by layershifter** [comment]
https://github.com/stardust-ui/react/pull/1524#pullrequestreview-256407003

I merged the latest `master` and issues that were mentioned in that comment are still present 😭 

bug-popup

@jurokapsiar can you please check?

**Comment by jurokapsiar** [comment]
@Bugaa92, @layershifter thanks for the review. I have resolved your comments. For your points:
- preventing scrolling mechanism does not work when opening popups on context
- right click does not dismiss context popups; what's the expected behavior?

1 I was not able to repro but I will check with you tomorrow to see if it's not version specific behavior - but I checked chrome and FF
2 is fixed now, probably some forgottnen commit, I thought I already fixed that

**Comment by layershifter** [comment]
Yep, second issue is fixed 🎉 

However, scrolling behavior is different in IE/Edge/Firefox and Chrome.

### Chrome

bug-popup1


### Edge

bug-popup2

But, as I remember we agreed to hide it on scroll...

File: packages/react/src/components/Popup/Popup.tsx [hunk 12]
```diff
@@ -211,6 +220,9 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
       e.preventDefault()
       this.setPopupOpen(true, e)
     },
+    preventScroll: e => {
```
layershifter, line 215:
```suggestion
    closeOnScroll: e => {
```

Should have a different name now...

packages/react/src/lib/accessibility/Behaviors/Popup/popupBehavior.ts
```diff
@@ -30,6 +32,17 @@ const popupBehavior: Accessibility<PopupBehaviorProps> = props => {
         closeAndFocusTrigger: {
           keyCombinations: [{ keyCode: keyboardKey.Escape }],
         },
+        preventScroll: {
```
layershifter, line 35: https://github.com/stardust-ui/react/pull/1524/files#r304360250

layershifter:
To fix CI, please update `build/gulp/tasks/test-projects/rollup/rollup.config.js`:
```js
 namedExports: {
        'node_modules/keyboard-key/src/keyboardKey.js': [
        // items
          'PageDown',
          'PageUp',
      ]
}

layershifter (approved):

PR #1540, opened 2019-06-24
feat(Tree): virtualized Tree prototype

silviuaavram:
WIP


silviuaavram:
Blocked ATM due to the fact that our Tree does not render a flat collection of items.

PR #1541, opened 2019-06-25
docs(Toolbar): Prototype for custom toolbar

miroslavstastny:
Adds a prototype for custom-styled toolbar


docs/src/index.ejs
```diff
@@ -124,6 +124,20 @@
     .hidden {
       display: none;
     }
+
+    /*
+      TODO: this could solve the isFromKeyboard and onFocus complication.
+            Need to work out how we integrate this static style into style functions.
+
+    [data-whatinput="mouse"] *:focus {
```
jurokapsiar: that would be good. when we checked a year back, fela did not allow these kind of selectors. But with unified focus indicator, it might be easier to have just global style that has nothing to do with fela. We will need to try and see. It would be also good to have screener tests of focused state for all components before (if) we decide to try this

docs/src/prototypes/customToolbar/CustomToolbar.tsx
```diff
@@ -0,0 +1,341 @@
...
+        // },
+
+        // TODO: Winner. Meets visual parity with Teams and code is most sane.  Confirm acc with Juraj.
+        // Flexbox has focus
+        {
+          key: 'recording2',
+          kind: 'custom',
+          fitted: 'horizontally',
+          content: (
+            <FocusableFlexBox fill gap="gap.medium" vAlign="center">
```
jurokapsiar: important thing here is that if the focus lands on the FocusableFlexBox div, the onClick handler should also be there (and we will need to add keyboard handlers for enter/space as well as aria role). Based on this:
- can we prevent the user from putting onClick to the 'custom' kind using types or prevent spreading it to the root?
- should we rename it to ClickableFlexBox?

I understand that in this particular use case they do not use onClick, but if this will become a generic thing, it should handle it correctly.

Side note - if this would have popup/menu/tooltip attached, the FocusableFlexBox should become the trigger for it.
But maybe we will just say that custom items cannot have popup/menu/tooltip and we will let consumers handle it - in that case the type should also be limited appropriately and not expose those props.

docs/src/prototypes/customToolbar/CustomToolbar.tsx
```diff
@@ -0,0 +1,341 @@
...
+  standard: props => [...commonLayout(props), ...sidebarButtons(props), layoutItems.endCall(props)],
+
+  'desktop-share': props => [
+    ...commonLayout(props),
+    ...sidebarButtons(props),
+    { key: 'divider-sidebar', kind: 'divider' },
+    {
+      key: 'stop-sharing',
+      kind: 'custom',
+      content: <Button content="Stop Sharing" />,
```
jurokapsiar, line 164: we will need to consider the case when this is transformed to a menu item in the overflow menu. It should not be a button there - menu only allows menu items, menu item checkboxes and menu item radios.
Btw this also shows why the custom item type should not have onClick prop available - the user needs to add onClick to the button. Challenge is that for the overflow menu item, the onClick will need to be on the menu item, not on something what is inside of it.

docs/src/prototypes/customToolbar/CustomToolbar.tsx
```diff
@@ -0,0 +1,341 @@
...
+    },
+
+    onClick: () => {
+      _.invoke(props, 'onMoreClick')
+    },
+
+    variables: {
+      primary: true,
+    },
+  },
```
jurokapsiar: The exampe does not show a spin button, but we will need to figure out how that is going to be transformed into a menu item for the overflow. Should we generate two menu items for it? Should we introduce something similar to the chrome zoom menu - which is a group of menu items aligned horizontaly?

image

docs/src/index.ejs
layershifter: Moved to comment https://github.com/stardust-ui/react/issues/749#issuecomment-507267679

docs/src/prototypes/customToolbar/CustomToolbar.tsx
```diff
@@ -0,0 +1,234 @@
...
+  pptSlide?: string
+  onPptPrevClick?: () => void
+  onPptNextClick?: () => void
+
+  onEndCallClick?: () => void
+}
+
+type CustomToolbarItem =
+  | ((ToolbarItemProps | ToolbarCustomItemProps) & {
+      as?: any
```
jurokapsiar: is `as` needed in the prototype?

layershifter: Good catch, removed it 👍

docs/src/prototypes/customToolbar/CustomToolbar.tsx
layershifter: > we will need to consider the case when this is transformed to a menu item in the overflow menu. It should not be a button there - menu only allows menu items, menu item checkboxes and menu item radios.

We don't touch responsive layout in this case, but it's a good mention. Should be considered in future updates

> Btw this also shows why the custom item type should not have onClick prop available - the user needs to add onClick to the button.

It was addressed previously with PR for `custom` `kind`

CHANGELOG.md
```diff
@@ -16,6 +16,9 @@
 This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+### Fixes
```
mnajdova: Oops, let's remove this.

docs/src/prototypes/customToolbar/CustomToolbar.tsx
```diff
@@ -0,0 +1,230 @@
...
+  micActive?: boolean
+  onMicChange?: (state: boolean) => void
+
+  screenShareActive?: boolean
+  onScreenShareChange?: (state: boolean) => void
+
+  sidebarSelected: false | 'chat' | 'participant-add'
+  onSidebarChange?: (state: false | 'chat' | 'participant-add') => void
+
+  chatHasDot?: boolean
```
mnajdova: `chatHasNotification`

mnajdova:
```suggestion
  chatHasNotification?: boolean
```

docs/src/prototypes/customToolbar/CustomToolbar.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+const commonLayout: CustomToolbarLayout = props =>
+  [
+    props.isRecording && {
+      key: 'recording',
+      kind: 'custom' as ToolbarItemShorthandKinds,
+      focusable: true,
+      content: (
+        <Status
+          state="error"
```
mnajdova: Hmm we can strange name for error... Maybe danger would be better choice.. :\

docs/src/prototypes/customToolbar/CustomToolbar.tsx
```diff
@@ -0,0 +1,230 @@
...
+          title="Recording"
+          variables={siteVars => ({ borderColor: siteVars.colors.white })}
```
mnajdova: Let's move this to variables/styles files.

mnajdova: We don't want values for variables in tsx files.

docs/src/prototypes/customToolbar/CustomToolbar.tsx
```diff
@@ -0,0 +1,230 @@
...
+    layoutItems.endCall(props),
+  ],
+}
+
+const CustomToolbar: React.FunctionComponent<CustomToolbarProps> = props => {
+  const { layout = 'standard' } = props
+
+  return (
+    <Toolbar
+      variables={{ dividerMargin: 0, borderRadius: 0, itemHeight: '4rem', isCt: true }}
```
mnajdova: I don't like that we have these defined in the tsx file. Can we move them to variables?

docs/src/prototypes/customToolbar/darkThemeOverrides.ts
```diff
@@ -0,0 +1,227 @@
...
+      ctItemBackground: siteVars.colorScheme.default.background1,
+      ctItemBackgroundHover: siteVars.colorScheme.brand.backgroundHover1,
+      ctItemBorderColorFocus: siteVars.colorScheme.default.borderFocus,
+      ctItemColor: siteVars.colorScheme.default.foreground,
+      ctItemColorFocus: siteVars.colorScheme.default.foregroundFocus,
+      ctItemColorHover: siteVars.colorScheme.default.foregroundHover,
+
+      ctItemActiveBackground: siteVars.colorScheme.brand.backgroundHover1,
+      ctItemActiveBackgroundOverlay:
+        'linear-gradient(90deg,rgba(60,62,93,.6),rgba(60,62,93,0) 33%),linear-gradient(135deg,rgba(60,62,93,.6) 33%,rgba(60,62,93,0) 70%),linear-gradient(180deg,rgba(60,62,93,.6) 70%,rgba(60,62,93,0) 94%),linear-gradient(225deg,rgba(60,62,93,.6) 33%,rgba(60,62,93,0) 73%),linear-gradient(270deg,rgba(60,62,93,.6),rgba(60,62,93,0) 33%),linear-gradient(0deg,rgba(98,100,167,.75) 6%,rgba(98,100,167,0) 70%)',
```
mnajdova, line 83: Let's check before merge whether these colors exists in the color palette.

docs/src/prototypes/customToolbar/darkThemeOverrides.ts
```diff
@@ -0,0 +1,227 @@
...
+      ctItemActiveColor: siteVars.colorScheme.default.foreground3,
+
+      ctItemDangerBackground: siteVars.colorScheme.red.background2,
+      ctItemDangerBackgroundHover: siteVars.colorScheme.red.backgroundHover,
+      ctItemDangerColorHover: siteVars.colorScheme.red.foregroundHover,
+
+      ctItemIndicatorPadding: '8px',
```
mnajdova: We have lots of mix of px vs rem values. Let's stick with one of them (rem preferably)

docs/src/prototypes/customToolbar/darkThemeOverrides.ts
```diff
@@ -0,0 +1,227 @@
...
+
+      ctItemNotificationBackgroundColor: 'red',
```
mnajdova: nit: let's use some color schema value here if possible

docs/src/prototypes/customToolbar/darkThemeOverrides.ts
```diff
@@ -0,0 +1,227 @@
...
+      }: ComponentStyleFunctionParam<ToolbarCustomItemProps, CustomToolbarVariables>) => ({
+        ...(v.isCt && {
+          background: v.ctItemBackground,
+          borderStyle: v.ctBorderStyle,
+          borderWidth: v.ctBorderWidth,
+
+          ...(v.isCtItemPrimary && { background: v.ctItemPrimaryBackground }),
+          ...(v.isCtItemIndicator && { padding: v.ctItemIndicatorPadding }),
+
+          ...(p.isFromKeyboard && {
```
mnajdova, line 140: Let's check with design whether these focusing styles are correct. And at least add comment that those should be changed if they should appear as in the other components.

layershifter: /cc @codepretty

docs/src/prototypes/customToolbar/CustomToolbar.tsx
miroslavstastny: This PR should not change any library code

docs/src/prototypes/customToolbar/CustomToolbar.tsx
miroslavstastny: Hmm, correct. But `Status` does magic computations based on this variable.
Variable removed, handled in styles.

docs/src/prototypes/customToolbar/CustomToolbar.tsx
miroslavstastny: Variables removed.
That means built-in variables are useless for (conditional) style overrides :-(

docs/src/prototypes/customToolbar/darkThemeOverrides.ts
miroslavstastny: Our plan is not to match redlines (for now), comment added.

docs/src/prototypes/customToolbar/darkThemeOverrides.ts
miroslavstastny: Changed this to rems, now px is used only for borders

docs/src/prototypes/customToolbar/darkThemeOverrides.ts
miroslavstastny: This rather shows how to change styles for focus than matches redlines.

mnajdova: I agree, just if we can use the general `getBorderFocusStyles` styles, we would be able to remove some styles... Let's not address this now until we are sure what is the expected state here...

docs/src/prototypes/customToolbar/darkThemeOverrides.ts
```diff
@@ -0,0 +1,273 @@
...
+
+              ...(v.isCtItemPrimary && {
+                color: v.ctItemPrimaryColorHover,
+                background: v.ctItemPrimaryBackgroundHover,
+              }),
+            }),
+          }),
+
+          ...(v.isCtItemIconNoFill && {
+            '& .ui-icon__filled': {
```
mnajdova, line 245: Can we use here `getIconFilledOrOutlineStyles.ts`? Or at least can we use `[`& .${teamsIconClassNames.filled}`]` and `[`& .${teamsIconClassNames.outline}`]` as selectors?

mnajdova (approved):
Let's review comments before merging..

docs/src/prototypes/customToolbar/darkThemeOverrides.ts
miroslavstastny: Unfortunately there is currently no way how to access it from the outside of the library.
This is a bummer we should think about.

PR #1542, opened 2019-06-25
fix(plugins): use compiled dependency instead of `css-expand-shorthand`

### LINKED ISSUE #1534: [IE11] Dependency js file has arrow function
## Bug Report
Stardust uses `css-shorthand-expand` which contains a js file with an arrow function.
This breaks IE11 compatibility:

See line 28:
https://github.com/kapetan/css-shorthand-expand/blob/master/source/background.js

### Steps
Open https://stardust-ui.github.io/react/ in IE11.

### Expected Result
Page renders something.

### Actual Result
Page doesn't render anything.

### Version
0.33.0

layershifter:
Fixes #1534.

### Problem

IE11 does not have support of arrow functions while our dependency [`css-shorthand-expand`](https://github.com/kapetan/css-shorthand-expand) has its [usage](https://github.com/kapetan/css-shorthand-expand/blob/master/source/background.js#L28). And for client there is only one option to fix - compile these dependency with `babel-loader` or any other thing 👎 

### Solution

I used [`@pika/pack`](https://github.com/pikapkg/pack) to quicky get bundled library, so our `cssExpandShorthand.ts` contains `css-shorthand-expand` and all its dependencies in the single and tree-shaken file. As we compile all our files with Babel: arrow functions and any other unsupported things will be compiled properly 🚀

### DIFF
--- a/docs/src/index.ejs
+++ b/docs/src/index.ejs
@@ -6,7 +6,6 @@
   <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
   <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.5.0/css/all.css" integrity="sha384-B4dIYHKNBt8Bc12p+WXckhzcICo0wtJAoU8YZTY5qE0Id1GSseTk6S+L3BlXeVIU" crossorigin="anonymous">
   <link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.6.3/css/brands.css" integrity="sha384-1KLgFVb/gHrlDGLFPgMbeedi6tQBLcWvyNUN+YKXbD7ZFbjX6BLpMDf0PJ32XJfX" crossorigin="anonymous">
-  <link rel="stylesheet" href="https://unpkg.com/react-vis@<%= htmlWebpackPlugin.options.versions.reactVis %>/dist/style.css" />
   <link rel="shortcut icon" type="image/x-icon" href='<%= __BASENAME__ + 'logo.png' %>'/>
   <title>Stardust</title>
   <script>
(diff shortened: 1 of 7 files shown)


docs/src/index.ejs
```diff
@@ -32,24 +31,14 @@
   </style>
 </head>
 <body>
-  <script src="https://cdn.jsdelivr.net/npm/@babel/polyfill@7/dist/polyfill.min.js"></script>
+  <script src='https://cdn.jsdelivr.net/npm/core-js-bundle/minified.js'></script>
+  <script src='https://cdn.jsdelivr.net/npm/whatwg-fetch/dist/fetch.umd.min.js'></script>
```
layershifter, line 35: `fetch()` is not supported by IE11

docs/src/index.ejs
```diff
@@ -32,24 +31,14 @@
   </style>
 </head>
 <body>
-  <script src="https://cdn.jsdelivr.net/npm/@babel/polyfill@7/dist/polyfill.min.js"></script>
+  <script src='https://cdn.jsdelivr.net/npm/core-js-bundle/minified.js'></script>
```
layershifter, line 34: `core-js-bundle` is bundled version of `core-js` which is a successor for `@babel/polyfill`:
https://babeljs.io/docs/en/babel-polyfill

docs/src/index.ejs
```diff
@@ -32,24 +31,14 @@
...
-    src="https://unpkg.com/prettier@<%= htmlWebpackPlugin.options.versions.prettier %>/parser-babylon.js"
-  ></script>
-  <script
-    crossOrigin="anonymous"
-    src="https://unpkg.com/prettier@<%= htmlWebpackPlugin.options.versions.prettier %>/parser-html.js"
-  ></script>
-  <script
-    crossOrigin="anonymous"
-    src="https://unpkg.com/prettier@<%= htmlWebpackPlugin.options.versions.prettier %>/parser-typescript.js"
+    src="https://cdn.jsdelivr.net/combine/npm/prettier@<%= htmlWebpackPlugin.options.versions.prettier %>/standalone.min.js,npm/prettier@<%= htmlWebpackPlugin.options.versions.prettier %>/parser-babylon.min.js,npm/prettier@<%= htmlWebpackPlugin.options.versions.prettier %>/parser-html.min.js,npm/prettier@<%= htmlWebpackPlugin.options.versions.prettier %>/parser-typescript.min.js"
```
layershifter, line 39: Instead of 4 files unminified files from CDN we will require only minified one 🚤

docs/src/index.ejs
```diff
@@ -62,7 +51,8 @@
 
   <script src="https://cdn.jsdelivr.net/npm/@babel/standalone@<%= htmlWebpackPlugin.options.versions.babelStandalone %>/babel.min.js"></script>
 
-  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/prismjs@1.16.0/themes/prism-tomorrow.css">
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/prismjs@1.16.0/themes/prism-tomorrow.min.css">
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/react-vis@<%= htmlWebpackPlugin.options.versions.reactVis %>/dist/style.min.css" />
```
layershifter, line 55: Also minified versions

packages/docs-components/src/CodeSnippet/CodeSnippetLabel.tsx
```diff
@@ -28,14 +43,14 @@ const CodeSnippetLabel: React.FunctionComponent<CopySnippetLabelProps> = props =
...
-          top: '1rem',
+          right: '0.8rem',
+          top: '0.8rem',
           zIndex: 100,
         }}
         title={copyable ? 'Copy' : undefined}
       >
         <div>{label || mode}</div>
-        {copyable && <div style={{ marginLeft: '5px' }}>{active ? '✔ ' : '⎘'}</div>}
+        {copyable && <div style={{ marginLeft: '5px' }}>{active ? checkIcon : copyIcon}</div>}
```
layershifter, line 53: image

Current icon looks ugly and there are no good replacement, so I decided to use SVGs.
Also it's not enough contrast, so change `color` to `#ccc` from PrismJS theme.


### After

codesnippet-icons

miroslavstastny (approved):

PR #1550, opened 2019-06-26
docs: create a visible Best Practices section for each component

alinais:
Before:
image

After: 
image


packages/react/src/components/Menu/Menu.tsx
```diff
@@ -245,10 +245,6 @@ Menu.create = createShorthandFactory({ Component: Menu, mappedArrayProp: 'items'
  * A menu displays grouped navigation actions.
  * @accessibility
  * Implements ARIA [Menu](https://www.w3.org/TR/wai-aria-practices-1.1/#menu), [Toolbar](https://www.w3.org/TR/wai-aria-practices-1.1/#toolbar) or [Tabs](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel) design pattern, depending on the behavior used.
- * Do not use Children API (`<MenuItem>` component directly), use [Shorthand API](https://stardust-ui.github.io/react/shorthand-props) with `items` prop instead.
  * For render tree customization, use [render callback argument](https://stardust-ui.github.io/react/shorthand-props#render-callback-argument).
```
jurokapsiar, line 248: I think this can also be moved to the Do section

docs/src/examples/components/Menu/BestPractices/MenuBestPractices.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react'
+
+import ComponentBestPractices from 'docs/src/components/ComponentBestPractices'
+
+const doList = [
+  {
+    key: 'one',
+    content: 'Choose desired accessibility behavior depending on the use case.',
```
codepretty: Would it be helpful to add examples of what use cases would use which behaviors?

docs/src/examples/components/Popup/BestPractices/PopupBestPractices.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+
+import ComponentBestPractices from 'docs/src/components/ComponentBestPractices'
+
+const doList = [
+  {
+    key: 'one',
+    content: 'Use popupFocusTrapBehavior if the focus needs to be trapped inside of the Popup.',
```
codepretty: What does `trapped` mean here - just that the focus remains inside the popup until the popup is closed? What is the default behavior if you don't have this set?

docs/src/examples/components/Popup/BestPractices/PopupBestPractices.tsx
```diff
@@ -0,0 +1,21 @@
...
+  },
+  {
+    key: 'two',
+    content:
+      "If Popup's content is lazy loaded and focus needs to be trapped inside - make sure to use state change to trigger componentDidUpdate, so the focus can be set correctly to the first tabbable element inside Popup or manually set focus to the element inside once content is loaded.",
```
codepretty: Are both triggering `componentDidUpdate` and manually setting focus both equally good options or is one more preferable than another? Is there a reason to use one over the other?

docs/src/examples/components/Reaction/BestPractices/ReactionBestPractices.tsx
```diff
@@ -0,0 +1,22 @@
...
+const doList = [
+  {
+    key: 'one',
+    content:
+      'Use actionable components (for example Button) if the reactions need to be actionable.',
+  },
+  {
+    key: 'two',
+    content:
+      'Add textual representation to the icon slot if it only contains an icon (using title, aria-label or aria-labelledby props on the slot).',
```
codepretty: For accessibility best practices, when would I use title vs aria-label vs aria-labelledby? Or is there an accessibility FAQ page that would be helpful for answering questions like this and just linking to?

docs/src/examples/components/Menu/BestPractices/MenuBestPractices.tsx
alinais: @jurokapsiar, can you please provide some short examples?

docs/src/examples/components/Accordion/BestPractices/AccordionBestPractices.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+
+import ComponentBestPractices from 'docs/src/components/ComponentBestPractices'
+
+const doList = [
+  {
+    key: 'one',
```
kuzhelov: may suggest to omit keys from the client's API - we could safely use items' indexes. Actually, we could omit `content` prop from the API as well - this should be enough for client just to pass strings array:

```jsx
const doList = [
 'Use Accordion for grouping parts of the UI (multipart forms, articales...)',
  'Use Tree component to display a hierarchical structure that allows user to select one item.'
]
```

All the necessary mapping to `key` and `content` could be done by `ComponentBestPractices` component under the hood

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -137,8 +137,5 @@ class Alert extends UIComponent<WithAsProp<AlertProps>, AlertState> {
 /**
  * A Alert displays information that explains nearby content.
  * @accessibility
```
kuzhelov: should we remove @accessibility section altogether, as there is no content there now?

docs/src/examples/components/Dropdown/BestPractices/DropdownBestPractices.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react'
+
+import ComponentBestPractices from 'docs/src/components/ComponentBestPractices'
+
+const doList = [
+  {
+    key: 'one',
+    content:
+      'Provide getA11ySelectionMessage, getA11yStatusMessage, noResultsMessage and loadingMessage props to announce state changes correctly.',
```
kuzhelov: for the sake of maintaining consistency, lets enclose prop names (e.g. `getA11ySelectionMessage`), prop values (`warning`, `danger`)  as well as attribute names (e.g. `aria-label`) in back ticks (`). This will provide proper formatting for these tokens and improve readability for a client

kuzhelov: nit: 

> to announce state changes correctly

May suggest to rephrase a bit: 

> to visualize dropdown state correctly

packages/react/src/components/Alert/Alert.tsx
alinais: I think this would be a valid update.

docs/src/examples/components/Reaction/BestPractices/ReactionBestPractices.tsx
alinais: These are more basic details about accessibility.
- `title` you use when you want the description of the element to be shown in a tooltip when you hover the element  as well as to be read by the screen reader
- `aria-label` is used for screen reader (no tooltip is shown by default out of it)
- `aria-labelledby` is used when the element refers to some other element
I found this nice description here:
https://stackoverflow.com/questions/22039910/what-is-aria-label-and-how-should-i-use-it

docs/src/examples/components/Popup/BestPractices/PopupBestPractices.tsx
alinais: Trapped means that as long as the popup is opened, the focus stays inside of it. But we will add more details about it as part of the issue #1572

docs/src/examples/components/Popup/BestPractices/PopupBestPractices.tsx
alinais: I think that here you need to try and see which version works better for a specific case. We'll add more information to the details of Popup as part of #1572 work.

docs/src/examples/components/Alert/BestPractices/AlertBestPractices.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+
+import ComponentBestPractices from 'docs/src/components/ComponentBestPractices'
+
+const doList = [
+  'Use warning and danger variants to announce the alert by the screen reader.',
+  'Use other libraries (for example `react-aria-live`) if the content of default or success variant needs to be announced.',
```
kuzhelov, line 7: unfortunately, these markdown ticks are rendered 'as is' at the moment - however, think that this is a right direction taken, and it is just about providing simple markdown renderer to process these lines (like `react-markdown`). Otherwise, if we will use React markup in `do` and `don't` for these simple `code` items, we will end up with barely readable code. Lets just take this as a follow-up thing

kuzhelov (approved):
❤️

PR #1551, opened 2019-06-26
docs(Composition): add section about `as` prop and DOM attributes

layershifter:
This PR:
- adds support of MDX
- adds docs for `as` prop and unhandled props


docs/src/pages/Composition.mdx
```diff
@@ -0,0 +1,45 @@
...
+
+```jsx
+<>
+  {/* 🧱 Each Stardust UI component has a default value for `as` prop */}
+  {/* Will output: <h1 class="ui-header" /> */}
+  <Header />
+  {/* Uses another tag: <h2 class="ui-header" /> */}
+  <Header as="h2" />
+</>
+```
```
jurokapsiar, line 21: can you please add following:
```
Note: Using 'as' prop can change the behavior of the component, therefore Stardust does not handle accessibility and styling of the component if this prop is used. It is the responsibility of the consumer to verify these aspects and handle them correctly.
```

Not entirely sure about the styling part, you are in better position to say that, but I think even for styling, you cannot guarantee that the styles will be applied properly if you use as.

bmdalex (approved):
looks good 👍

docs/src/pages/Composition.mdx
layershifter: image

sophieH29 (approved):

PR #1553, opened 2019-06-26
feat(Toolbar): Add ARIA attributes and keyboard handling for a menu

sophieH29:
This PR adds:

- correct ARIA attributes to the Toolbar's rendered menu by using `menuItemBehavior`
- added `toolbarItemBehavior` - which uses `buttonBehavior` and addtionally sets `aria-haspopup` attribute based if Toolbar button opens a popup or a menu. Also closes and focuses toolbar button on `ESC` press


packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -194,8 +210,8 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>, ToolbarItemS
     const renderedItem = (
       <ElementType
         {...accessibility.attributes.root}
-        {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.root, unhandledProps)}
         {...unhandledProps}
+        {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.root, unhandledProps)}
```
bmdalex, line 214: should we document that `applyAccessibilityKeyHandlers` needs to be spread after `unhandledProps` on a component?

bmdalex (approved):
👍 one small comment

packages/react/src/components/Toolbar/ToolbarItem.tsx
sophieH29: @layershifter have you documented it when added this? frankly, I don't remember why 😄

layershifter: It's done for correctly overriding `onKeyDown` on slots. It should be applied after `unhandledProps` because they can contain `onKeyDown` from user and is handled by UTs in `isConformant()`

sophieH29: added as a comment

miroslavstastny (approved):

PR #1557, opened 2019-06-27
perf(docs): remove DLL, use splitChunks & runtimeChunk

layershifter:
### Teaser

image

### Remove DLL

Previously DLL was used to speed build, but there are no benefits currently:
- we used it to build extra small 2 packages
- it was improperly configured and actually we have used `vendor` chunk
- it doesn't make sense to use in our case at all: https://github.com/zeit/next.js/issues/4123#issuecomment-379457733

### Fix double build

Previously we had build that was running Wepback two times during `yarn start`:
- first time under `build:docs:webpack`:
image

Actually first time build doesn't have any sense because these files will not be served.

- second time under `webpack-dev-middleware` on `serve:docs`:
image

### Open browser tab
Behavior taken from CRA, window will be opened once build is ready. Done via Webpack plugin.

### Build optimizations
- use `filename: [name].js` in `DEV`: inspired by [Webpack docs](https://webpack.js.org/guides/build-performance/#avoid-production-specific-tooling)
- disabled `cacheCompression` on `babel-loader`: inspired by [CRA](https://github.com/facebook/create-react-app/blob/v3.0.1/packages/react-scripts/config/webpack.config.js#L397)
- use `splitChunks` and `runtimeChunk` optimizations: inspired by [CRA](https://github.com/facebook/create-react-app/blob/v3.0.1/packages/react-scripts/config/webpack.config.js#L256)


build/gulp/plugins/webpack-open-browser.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(options: OpenBrowserPluginOptions) {
+    this.options = options
+  }
+
+  apply(compiler) {
+    compiler.hooks.done.tap('Hello World Plugin', () => {
+      if (!this.opened) {
+        this.opened = true
+
+        open(`http://${this.options.host}:${this.options.port}`).catch(() => {})
```
levithomason: This is really annoying as it opens up a new doc tab even when it is already open.  I think the user should just open their own doc site when they are ready.

If the majority of the team wants this, let's duplicate what CRA does.  It only opens the tab if necessary otherwise it focuses the already open tab.

https://github.com/facebook/create-react-app/blob/2c6dd45cf4cb81f3cbd2e4073713647ebd033bbf/packages/react-dev-utils/openBrowser.js

layershifter: Switched this one to use `react-dev-utils/openBrowser` 👍

levithomason:
👍 🚀 😸 

**Before**: `yarn start` 2m 40s

**After**:
`yarn start` 1m
(restart) 24s

levithomason (approved):
:shipit:

PR #1558, opened 2019-06-27
feat(Toolbar): add `custom` kind of ToolbarItem

miroslavstastny:
`custom` Toolbar Item kind allows to insert any content into a toolbar item - either non-focusable (`Text`) or focusable (`Button`):

```jsx
<Toolbar
  items={[
    {
      key: 'custom-text',
      kind: 'custom',
      content: <Text content="text" />,
    },
    {
      key: 'custom-button',
      kind: 'custom',
      fitted: 'horizontally',
      content: <Button content="button" />,
    },
  ]}
/>
```

image

### `fitted` prop
This custom item adds padding by default. This can be customised by specifying `fitted` prop, which can be:
- `false` (default)
- `true` - removes all padding
- `'horizontally'` - removes left and right padding
- `'vertically'` - removes top and bottom padding

### `focused` prop

A custom item can be focused.

image


packages/react/src/components/Toolbar/ToolbarCustomItem.tsx
```diff
@@ -0,0 +1,68 @@
...
+  childrenExist,
+  commonPropTypes,
+} from '../../lib'
+
+import { WithAsProp, withSafeTypeForAs } from '../../types'
+import { Accessibility } from '../../lib/accessibility/types'
+import { defaultBehavior } from '../../lib/accessibility'
+
+export interface ToolbarCustomItemProps
+  extends UIComponentProps,
```
jurokapsiar, line 22: is it possible to restrict the prop types and remove onClick? onClick should be put on the content element instead as the custom item is not focusable.

docs/src/examples/components/Toolbar/Content/index.tsx
```diff
@@ -25,6 +25,11 @@ const Content = () => (
       description="Toolbar items can be grouped into radio group. Up/Down arrow keys can be used to cycle between radio items. Only one of the radio items can be selected at a time, should be implemented additionally."
       examplePath="components/Toolbar/Content/ToolbarExampleRadioGroup"
     />
+    <ComponentExample
+      title="Toolbar can contain custom content"
+      description="Toolbar item can contain custom content."
```
sophieH29, line 32: I would add here or in the accessibility section of the whole Toolbar page, that when custom item used, accessibility and styles should be covered manually

sophieH29 (approved):

packages/react/src/components/Toolbar/ToolbarCustomItem.tsx
```diff
@@ -0,0 +1,122 @@
...
+  isFromKeyboard: boolean
+}
+
+class ToolbarCustomItem extends UIComponent<
+  WithAsProp<ToolbarCustomItemProps>,
+  ToolbarCustomItemState
+> {
+  static displayName = 'ToolbarCustomItem'
+
+  static className = 'ui-toolbar__customitem'
```
kuzhelov, line 64: `ui-toolbar__custom-item`?

packages/react/src/themes/teams/components/Toolbar/toolbarCustomItemStyles.ts
```diff
@@ -0,0 +1,47 @@
...
+      backgroundColor: v.background,
+      borderColor: 'transparent',
+      borderWidth: v.borderWidth,
+      borderStyle: 'solid',
+      height: v.itemHeight,
+      color: v.foreground || colors.foreground1,
+      display: 'flex',
+      alignItems: 'center',
+      justifyContent: 'center',
+      ...(p.fitted !== true &&
```
kuzhelov, line 23: just to be sure - which case we are protecting against with explicit `p.fitted !== true && ..`, instead of `!p.fitted && ..`?

packages/react/src/components/Toolbar/ToolbarCustomItem.tsx
layershifter: We have a UT for this one:

```tsx
test(`is a static equal to "${info.componentClassName}"`, () => {
```

It expects to have: `ui-toolbar__customitem`

packages/react/src/components/Toolbar/ToolbarCustomItem.tsx
kuzhelov: probably, I am missing something - but currently not sure how aforementioned statement can coexist with `focusable` prop :)

```jsx
  /** A custom item can be focused. */
  focusable?: boolean
```

@miroslavstastny, @jurokapsiar, could you, please, suggest which part I am missing?

packages/react/src/themes/teams/components/Toolbar/toolbarCustomItemStyles.ts
layershifter:
```jsx
const hasPadding = (fitted) => fitted !== true && fitted !== 'horizontally'
const hasPaddingModified = (fitted) => !fitted && fitted !== 'horizontally'
```

```js
console.log(hasPadding(true), hasPadding('horizontally'), hasPadding('vertically'))
```

```
false false true
```

```jsx
console.log(hasPaddingModified(true), hasPaddingModified('horizontally'), hasPaddingModified('vertically'))
```
```
false false false
```

I believe it's the reason

docs/src/examples/components/Toolbar/Content/index.tsx
layershifter: Section was added 👍

packages/react/src/components/Toolbar/ToolbarCustomItem.tsx
layershifter: Per offline discussion: it should be `focusable` conditionally, but it can't be `actionable` (`onClick: never`). We can consider to change this later based on provided use cases.

docs/src/examples/components/Toolbar/Content/index.tsx
```diff
@@ -25,6 +27,19 @@ const Content = () => (
...
+      examplePath="components/Toolbar/Content/ToolbarExampleCustomContent"
+    >
+      <Alert warning>
+        <p>
+          When <code>custom</code> kind is used it is the responsibility of the consumer to verify
+          accessibility and styling aspects of the component and handle them correctly. This kind of
+          items can't be actionable.
```
jurokapsiar: You can add:
```
This kind of items can't be actionable, but actionable components might be added to the content slot of the custom kind.
```

PR #1559, opened 2019-06-28
feat(slider): v1

### LINKED ISSUE #1539: Slider component
## Slider component

Related to #1255 

### Description

- a slider is an input that allows the user to choose a value from within a specific range of values
- sliders typically have a slider thumb that can be moved along a bar or track to change the value of the slider.

 image
image

### API proposal v1

- `defaultValue?: number | string | string[]` - The value of the slider.

- `disabled?: boolean`- A slider can be read-only and unable to change states.

- `discrete?: boolean` (alternatives: `segmented`, `dotted`) - Whether the slider has discrete styling or not.

- `fluid?: boolean` - A slider can take the width of its container.

- `input?: ShorthandValue` - Shorthand for the input component.

- `max?: string | number` - The greatest value in the range of permitted values.

- `min?: string | number` - The lowest value in the range of permitted values.

- `vertical?: boolean` - The slider orientation. Horizontal by default.

- `step?: string | number` - A number that specifies the granularity that the value must adhere to or the special value any.

- `inputRef?: React.Ref<HTMLElement>` - Ref for input DOM node.

- `onChange?: ComponentEventHandler<SliderProps>` - Change event handler.

- `value?: number | string | string[]` - The value of the slider.

### v2:

- #### `items` / `marks` prop
These are predetermined values to which the user can move the slider. If true the marks will be spaced according the value of the step prop. If an array, it should contain objects with value and an optional label keys.
Screenshot 2019-06-24 at 17 47 38

- #### Do we need a callback function that is fired when the mouseup is triggered?
PROPOSAL: `onChanged `, `onChangeComitted` (other names: `onMouseUp`, `onChangeSubmitted`) 

- #### `label` prop vs `Form.Field`?
What should be the preferred way to label a slider? e.g.:
image

- #### `media` prop needed or should we use `Flex` + `Icon` or other components?
56600347-51584e00-65ad-11e9-9703-b0159fe95515 (1)

- #### `valueFormat` prop in v1?
Optional function to format the slider value.

- #### `showValue` prop in v1?
Whether to show the value on the right of the Slider.

Inspired from:
- [Material UI](https://material-ui.com/api/slider/)
- [Fabric](https://developer.microsoft.com/en-us/fabric#/controls/web/slider)
- [MDN docs for `<input type="range">`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range)
- [ARIA practices](https://www.w3.org/TR/wai-aria-practices-1.1/#slider)

bmdalex:
# feat(slider): v1

Closes #1539 

## Description

This PR consists of the first version of the `Slider` component. It incorporates:

### 1. Functionality (demo below):
Screen Recording 2019-06-28 at 07 59 34

### 2. Styling (according to design files and redlines for all themes):

| Theme Name | Slider flavor |
|:-------:|:------:|
| Base | Screenshot 2019-06-28 at 08 05 07 |
| Teams | Screenshot 2019-06-28 at 07 58 09 
| Teams Dark | Screenshot 2019-06-28 at 07 58 25 |
| Teams HC | Screenshot 2019-06-28 at 07 58 44 |

### 3. Accessibility:
  - leverages native keyboard navigation
  - adds correct accessibility attributes via the `sliderBehavior`

## API

```typescript
export interface SliderProps {
  /**
   * Accessibility behavior if overridden by the user.
   * @default sliderBehavior
   */
  accessibility?: Accessibility

  /** The default value of the slider. */
  defaultValue?: SupportedIntrinsicInputProps['defaultValue'] // TODO 'value' type is React.ReactText but it doesn't work for 'defaultValue'

  /**
   * A slider can be read-only and unable to change states.
   * @default false
   */
  disabled?: SupportedIntrinsicInputProps['disabled']

  /**
   * A slider can take the width of its container.
   * @default false
   */
  fluid?: boolean

  /**
   * Callback that creates custom accessibility message a screen reader narrates when the value changes.
   * @param {SliderProps} props - Slider props.
   */
  getA11yValueMessageOnChange?: (props: SliderProps) => string

  /** Shorthand for the input component. */
  input?: ShorthandValue

  /** Ref for input DOM node. */
  inputRef?: React.Ref<HTMLElement>

  /**
   * The minimum value of the slider
   * @default 0
   */
  min?: SupportedIntrinsicInputProps['min']

  /**
   * The maximum value of the slider.
   * @default 100
   */
  max?: SupportedIntrinsicInputProps['max']

  /**
   * Called after item checked state is changed.
   */
  onChange?: ComponentEventHandler<SliderProps>

  /**
   * A number that specifies the granularity that the value must adhere to, or the special value 'any'.
   * A string value of any means that no stepping is implied, and any value is allowed
   * (barring other constraints, such as min and max).
   * @default 1
   */
  step?: SupportedIntrinsicInputProps['step']

  /** The value of the slider. */
  value?: React.ReactText

  /**
   * A slider can be displayed vertically.
   * @default false
   */
  vertical?: boolean
}
```

## TODO

- [x] Focus borders: still didn't find a way to add double borders to the slider's thumb
- [x] Remove slider marks in IE
- [x] Improve perf for recalculating props (accessibility props, styling props, etc) with every change of the slider's value
- [x] Add more visual tests


layershifter (commented):
Oops, CI fails and also there is warning somewhere is examples:
image

layershifter:
Not sure that performance is an issue:
image

Do you have real problems there?

docs/src/examples/components/Slider/Types/SliderExample.shorthand.tsx
```diff
@@ -0,0 +1,6 @@
+import * as React from 'react'
+import { Slider } from '@stardust-ui/react'
+
+const SliderExample = () => <Slider />
+
+export default SliderExample
```
layershifter:
```suggestion
export default SliderExampleShorthand
```

docs/src/examples/components/Slider/Types/SliderExampleDisabled.shorthand.tsx
```diff
@@ -0,0 +1,6 @@
+import * as React from 'react'
+import { Slider } from '@stardust-ui/react'
+
+const SliderExampleDisabled = () => <Slider disabled />
+
+export default SliderExampleDisabled
```
layershifter:
```suggestion
export default SliderExampleDisabledShorthand
```

docs/src/examples/components/Slider/Types/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+const Types = () => (
+  <ExampleSection title="Types">
+    <ComponentExample
+      title="Default"
+      description="A default slider."
+      examplePath="components/Slider/Types/SliderExample"
+    />
+    <ComponentExample
+      title="Disabled"
```
layershifter: `disabled` prop describes state

docs/src/examples/components/Slider/Variations/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
+
+const Variations = () => (
+  <ExampleSection title="Variations">
+    <ComponentExample
+      title="Icon"
+      description="A slider can have an icon."
+      examplePath="components/Slider/Variations/SliderExampleIcon"
```
layershifter: `icon` and `iconPosition` can be combined into single example, what do you think?

And also placed under `Slots` section

docs/src/examples/components/Slider/Types/SliderExample.shorthand.tsx
bmdalex: done thx

docs/src/examples/components/Slider/Types/SliderExampleDisabled.shorthand.tsx
bmdalex: done thx

docs/src/examples/components/Slider/Types/index.tsx
bmdalex: done thx

docs/src/examples/components/Slider/Variations/index.tsx
bmdalex: done thx

bmdalex:
> Oops, CI fails and also there is warning somewhere is examples:
> image

@layershifter  fixed...but it wasn't clear at all how to do this so I created #1571 to address that; pls provide your feedback

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,236 @@
...
+                value,
+                onChange: this.handleChange,
+                onFocus: this.handleFocus,
+                styles: styles.input,
+                ...applyAccessibilityKeyHandlers(accessibility.keyHandlers.input, htmlInputProps),
+              },
+            })}
+          </Ref>
+        </div>
+        {iconPosition !== 'start' && this.renderIcon(variables, styles)}
```
codepretty: Should the icon optionally be a button instead of just icon since it needs the ability to perform some action when clicked? when clicked it should mute
image

codepretty: We will also need hover/focus styles for the icon button

bmdalex: good point, let me think about it; will update the thread soon
@codepretty is the 20 volume mark part of the slider? I don't see that in the design files I have

bmdalex: discussed with the team and we decided not to do that in v1 since it's not a requirement that's generic enough; we might reconsider; I will add an example on how to compose the `Slider` with `Button` component to achieve the behavior you're mentioning:
> it needs the ability to perform some action when clicked

codepretty: The screenshot above is from the Windows default speaker controls - I was just trying to show a common volume control. Another requirement would be the icon updating at different points on the slider to indicate volume level. I'm not sure when something becomes a new component vs a usage example vs a prototype, but something to capture these requirements would be nice! Thank you @Bugaa92

bmdalex: sure, np @codepretty 
I just added an example to show how to compose the Slider with Button component to achieve the behavior you're mentioning:

> it needs the ability to perform some action when clicked

so I'm resolving this comment

miroslavstastny (commented):
RTL is broken :-(
slider-rtl

Both mouse and keyboard.

mnajdova:
RTL doesn't seem to work as expected:
LTR: 
image
RTL: 
image

miroslavstastny:
Disabled slider with icon does not look correct (the icon):
image

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,276 @@
...
+  WithAsProp,
+  withSafeTypeForAs,
+  Omit,
+} from '../../types'
+import { Accessibility } from '../../lib/accessibility/types'
+import { sliderBehavior } from '../../lib/accessibility'
+import { SupportedIntrinsicInputProps } from '../../lib/htmlPropsUtils'
+import Box from '../Box/Box'
+
+const defaultPropValues = {
```
mnajdova: Why not inlining these to the defaultProps? It's hard to follow what is actually in the default props

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,276 @@
...
+  inputWrapper: string
+  rail: string
+  thumb: string
+  track: string
+}
+
+export interface SliderProps
+  extends UIComponentProps,
+    ChildrenComponentProps,
+    Omit<SupportedIntrinsicInputProps, 'defaultValue'> {
```
mnajdova, line 57: I understand that we want to reuse these handlers, but it seems a bit awkward why on some handlers, like `onChange` we provide the props, but on others, like `onMouseDown` we don't... Can we somehow unify this behavior?

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,276 @@
...
+                className: Slider.slotClassNames.input,
+                as: 'input',
+                min,
+                max,
+                step,
+                type,
+                value,
+                onChange: this.handleChange,
+                onFocus: this.handleFocus,
+                onMouseDown: this.handleMouseDown,
```
mnajdova: Shouldn't the handlers go in the `overrideProps`? This way those will be overriden if the user provides the same one.

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,276 @@
...
+  /**
+   * A number that specifies the granularity that the value must adhere to, or the special value 'any'.
+   * A string value of any means that no stepping is implied, and any value is allowed
+   * (barring other constraints, such as min and max).
+   * @default 1
+   */
+  step?: SupportedIntrinsicInputProps['step']
+
+  /** The value of the slider. */
+  value?: React.ReactText
```
layershifter: This type looks strange... Can we use `number | string` instead?

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,276 @@
...
+   * @default 0
+   */
+  min?: SupportedIntrinsicInputProps['min']
+
+  /**
+   * Called after item checked state is changed.
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props.
+   */
+  onChange?: ComponentEventHandler<SliderProps>
```
layershifter:
```suggestion
  onChange?: ComponentEventHandler<SliderProps & { value: string }>
```

I believe that this is more accurate typing

docs/src/examples/components/Slider/Playground.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <Slider
+      disabled={disabled}
+      fluid={fluid}
+      min={min}
+      max={max}
+      step={step}
+      value={value}
+      vertical={vertical}
+      onChange={(e, data) => setValue(String(data.value))}
```
layershifter: https://github.com/stardust-ui/react/pull/1559/files#r302911394
Do we need this type cast?

docs/src/examples/components/Slider/Usage/SliderExampleAction.shorthand.tsx
```diff
@@ -0,0 +1,72 @@
+import * as React from 'react'
+import { Button, Input, Slider, Flex, SliderProps } from '@stardust-ui/react'
+import { useBooleanKnob } from '@stardust-ui/docs-components'
+
+interface SliderAction {
+  type: 'toggle_mute' | 'change_value'
+  value?: React.ReactText
```
layershifter: https://github.com/stardust-ui/react/pull/1559/files#r302911394

docs/src/examples/components/Slider/Usage/SliderExampleAction.shorthand.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+  const [state, dispatch] = React.useReducer(stateReducer, {
+    mute: false,
+    value: defaultValue,
+    currentValue: min,
+    minValue: min,
+  })
+
+  const handeChange = React.useCallback(
+    (e, data) => dispatch({ type: 'change_value', value: Number(data.value) }),
```
layershifter: A bit strange for me, in `Playground` we used `String()`...

Will https://github.com/stardust-ui/react/pull/1559/files#r302911394 solve this?

docs/src/examples/components/Slider/Usage/SliderExampleAction.shorthand.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+    case 'change_value':
+      return { ...state, mute: action.value <= state.minValue, value: action.value }
+
+    default:
+      throw new Error(`Action ${action.type} is not supported`)
+  }
+}
+
+const SliderWithAction: React.FunctionComponent<
```
layershifter: Do we really need this a separate component? It looks that this creates an additional complication with types and `commonProps`...

packages/react/src/components/Slider/Slider.tsx
mnajdova: Basically for all components, the right type is `Props & State`. Would suggest addressing this for all components in a separate PR

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,276 @@
...
+import Box from '../Box/Box'
+
+const defaultPropValues = {
+  max: 100,
+  min: 0,
+  step: 1,
+  value: 50,
+}
+
+const processInputValues = (
```
layershifter, line 31: Can we do this inside `getDerivedStateFromProps()` (`getAutoControlledStateFromProps()`). In that case we will have more accurate values for accessibility and styles...

packages/react/src/themes/base/components/Slider/sliderStyles.ts
```diff
@@ -0,0 +1,92 @@
+import { SliderVariables } from './sliderVariables'
+import { SliderProps, SliderState } from '../../../../components/Slider/Slider'
+import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
+import { selectors } from '../../../../styles/selectors'
```
layershifter: Having this in a separate file is confusing for me...

packages/react/src/themes/base/components/Slider/sliderStyles.ts
```diff
@@ -0,0 +1,92 @@
...
+
+const getSliderSlotStyles = (p: SliderProps, v: SliderVariables, slot: 'rail' | 'track') => ({
```
layershifter: _nit_: We can simplify this if we will have:

```jsx
const getCommonSlotStyles = (p: SliderProps, v: SliderVariables) => ({
  cursor: 'pointer',
  pointerEvents: 'none',
  position: 'absolute',
  border: 0,
  height: v.railHeight,
  marginTop: `calc(${v.height} / 2 - ${v.railHeight} / 2)`,
})
```

And conditions below will go back to matching slots:
```jsx
  ...(slot === 'rail' && { width: '100%', background: v.railColor }),
  ...(slot === 'track' && { background: v.trackColor }),
  ...(p.disabled && slot === 'rail' && { background: v.disabledRailColor }),
  ...(p.disabled && slot === 'track' && { background: v.disabledTrackColor }),
```

packages/react/src/themes/base/components/Slider/sliderVariables.ts
```diff
@@ -0,0 +1,48 @@
+import { pxToRem } from '../../../../lib'
+import { SiteVariablesPrepared } from '../../../types'
+
+export interface SliderVariables {
+  height: React.CSSProperties['height']
```
layershifter: Styles should be framework agnostic and variables, too. So we should not use `React.*` stuff

packages/react/test/specs/components/Slider/Slider-test.tsx
```diff
@@ -0,0 +1,50 @@
...
+        onKeyPress: 'input',
+        onKeyUp: 'input',
+      },
+    })
+  })
+
+  describe('auto-controlled', () => {
+    it('sets slider value from user when the value prop is not set (non-controlled mode)', () => {
+      const sliderComp = mount(<Slider />)
+      const domNode = getInputDomNode(sliderComp)
```
layershifter, line 28: Why we need to capture this from DOM? I believe that `expect(wrapper.find('input').prop('value')).toEqual('30')` should be enough

packages/react/test/specs/components/Slider/Slider-test.tsx
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react'
+import { ReactWrapper } from 'enzyme'
+
+import { mountWithProvider as mount } from 'test/utils'
+import { isConformant, handlesAccessibility } from 'test/specs/commonTests'
+import Slider from 'src/components/Slider/Slider'
+
+const testValue = '30'
```
layershifter: This value is used two times, it will be more obvious if we will just inline it

packages/react/test/specs/components/Slider/Slider-test.tsx
```diff
@@ -0,0 +1,50 @@
...
+const getInputDomNode = (sliderComp: ReactWrapper): HTMLInputElement =>
+  sliderComp.find('input').getDOMNode() as HTMLInputElement
+
+const setUserInputValue = (sliderComp: ReactWrapper, value: string) => {
+  sliderComp.find('input').simulate('change', { target: { value } })
+}
+
+describe('Slider', () => {
+  describe('conformance', () => {
```
layershifter:
```suggestion
```

We don't need a separate `describe()` there

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,276 @@
...
+  thumb: `${Slider.className}__thumb`,
+  track: `${Slider.className}__track`,
+}
+
+/**
+ * A slider is an input that allows the user to choose a value from within a specific range of values.
+ * @accessibility
+ * Implements [ARIA Slider](https://www.w3.org/TR/wai-aria-practices-1.1/#slider) design pattern.
+ */
+export default withSafeTypeForAs<typeof Slider, SliderProps, 'input'>(Slider)
```
bmdalex, line 267: when doing this, I cannot access `Slider.defaultProps` outside of this file, @kuzhelov any idea why?

Screenshot 2019-07-12 at 13 38 58

docs/src/examples/components/Slider/Playground.tsx
bmdalex: yes; reasons:
- I don't think https://github.com/stardust-ui/react/pull/1559/files#r302911394 is the correct fix; the value can be number as well
- `useStringKnob` requires `string` and `value` coming from `Slider` has type `string | number`

bmdalex:
> RTL is broken :-(
> slider-rtl
> 
> Both mouse and keyboard.


> RTL doesn't seem to work as expected:
> LTR:
> image
> RTL:
> image

@mnajdova 
@miroslavstastny 

great catch, thanks! RTL was fixed and I also added an example for it in the docs. Please check again.

packages/react/src/components/Slider/Slider.tsx
bmdalex: done, I was using them outside the component and thought it's more obvious, but managed to find a way, thanks!

docs/src/examples/components/Slider/Playground.tsx
layershifter:
```
      const value = _.get(e, 'target.value')
      _.invoke(this.props, 'onChange', e, { ...this.props, value })
```

According to these lines it can't be `string | number`:

```js
const e:  React.ChangeEvent<HTMLInputElement> = {}
e.target.value // is `string`
```

See https://github.com/stardust-ui/react/pull/1559#discussion_r302911394

bmdalex:
> Disabled slider with icon does not look correct (the icon):
> image

@miroslavstastny I think you were testing an earlier version 😕 removed `icon` and `iconPosition` altogether from `Slider`

packages/react/src/components/Slider/Slider.tsx
bmdalex: ooops... great catch, done

packages/react/src/components/Slider/Slider.tsx
bmdalex: no idea how, but I don't think it's a blocker for this PR
I added the `onChange` here just because it's very important for the `Slider` (same is done in `Input`)...

packages/react/src/components/Slider/Slider.tsx
bmdalex: done, thanks

packages/react/src/components/Slider/Slider.tsx
bmdalex: I don't agree with the `onChange?: ComponentEventHandler<SliderProps & { value: string }>` suggestion. `value` can be number as well. Also `value` is part of `SliderProps` so I don't see what the issue is exactly. Can I resolve?
@mnajdova @layershifter

docs/src/examples/components/Slider/Usage/SliderExampleAction.shorthand.tsx
bmdalex: done, thanks

docs/src/examples/components/Slider/Usage/SliderExampleAction.shorthand.tsx
bmdalex: removed, not needed

docs/src/examples/components/Slider/Usage/SliderExampleAction.shorthand.tsx
bmdalex: refactored, thanks

packages/react/src/components/Slider/Slider.tsx
bmdalex: does that mean `min`, `max` and `valueAsPercentage` will become part of state? why do we need that?
From accessibility POV it works anyway with both strings and numbers.

Are there any other concerns? It seems strange to add these to the state. @layershifter

packages/react/src/themes/base/components/Slider/sliderStyles.ts
bmdalex: previously I needed `selectors` in more files. moved them back to  `sliderStyles.ts`, thanks

packages/react/src/themes/base/components/Slider/sliderStyles.ts
bmdalex: I like this, done, thanks

packages/react/test/specs/components/Slider/Slider-test.tsx
bmdalex: agreed, done, thanks

packages/react/test/specs/components/Slider/Slider-test.tsx
bmdalex: done, thanks

packages/react/test/specs/components/Slider/Slider-test.tsx
bmdalex: I copied this from `Input-test.tsx`. Any major concerns with it?

packages/react/src/themes/base/components/Slider/sliderVariables.ts
bmdalex: done, thanks

packages/react/src/components/Slider/Slider.tsx
mnajdova: For this component this is definitely not a blocker, but it is something that we will need to tackle for other components.

layershifter: Value in callback can't be `number`, please check it again.
https://github.com/stardust-ui/react/pull/1559#discussion_r302952170

docs/src/examples/components/Slider/Playground.tsx
bmdalex: well that was a good argument 😄 applied suggestion, we should find a way to generalize this. thanks, resolving

packages/react/src/components/Slider/Slider.tsx
bmdalex: indeed, @layershifter is right; applied change but as @mnajdova says we need to align these interfaces in a separate PR; thanks guys, resolving 😃

docs/src/examples/components/Slider/Types/SliderExample.shorthand.steps.ts
```diff
@@ -0,0 +1,32 @@
+import { Slider } from '@stardust-ui/react/src'
+
+const selectors = {
+  input: `.${Slider.slotClassNames.input}`,
+}
+
+const focusSliderStep: ScreenerStep = (builder, keys) => builder.keys('body', keys.tab)
+
+const config: ScreenerTestsConfig = {
```
mnajdova, line 3: Let's make sure that we capture this for the rtl example as well

packages/react/src/components/Slider/Slider.tsx
layershifter: > does that mean `min`, `max` and `valueAsPercentage` will become part of state?

Yep

> From accessibility POV it works anyway with both strings and numbers.
> Are there any other concerns? It seems strange to add these to the state.

In this case: no.

docs/src/examples/components/Slider/Types/SliderExample.shorthand.steps.ts
bmdalex: good idea, done

packages/react/test/specs/components/Slider/Slider-test.tsx
layershifter: It's a strange thing to test DOM instead of React in this case. But you can keep it, probably `Input-tests` should be also refactored...

layershifter (approved):
👍

PR #1561, opened 2019-06-28
fix(Popup): update styles for popup colors in Teams themes

codepretty:
Updated the box shadow and border colors for popup in Teams theme

**Teams theme**
image

**Teams-dark theme**
image

**Teams-high-contrast theme**
image


kuzhelov (approved):

packages/react/src/themes/teams/components/Popup/popupContentVariables.ts
```diff
@@ -13,9 +16,12 @@ export interface PopupContentVariables {
 
 export default (siteVars: any): PopupContentVariables => {
   return {
-    borderColor: siteVars.colors.grey[250],
+    borderColor: siteVars.colorScheme.default.border2,
     borderRadius: pxToRem(3),
```
notandrew, line 20: Should this be an explicit value instead of rems because of the border being 1px always?

notandrew (approved):

Issue #1359, opened 2019-05-20
Missing icons: chevron-left and chevron-right


kamilakorzec:
We're missing left and right chevron for the navigation:
![image](https://user-images.githubusercontent.com/11797698/58016240-63ba9e80-7afd-11e9-832c-ffb45aa52277.png)

codepretty:
What size are those icons? I believe this request is dependent on issue #1356 

Once we have all 3 sizes of icons, we just need to rotate them to get the desired look. eg)
```
<Icon name="chevron" size="small" rotate={270} />
```

jurokapsiar:
Icons can be added to the theme in overrides

PR #1562, opened 2019-07-01
feat(Icon): add props arg to SVG spec

### LINKED ISSUE #1356: Icon sizes that pull in different paths
## Feature Request
We have a lot of icons that are made for particular sizes. Due to how SVGs scale, it is not always optimal to just use css for scaling sometimes a different icons is needed.

### Problem description
We have the ability to set a size on an icon and adjust the size with css
`<Icon name="chevron" size="small" />`

But there are a number of icons that are designed to use a different SVG or paths for different sizes.

### Proposed solution
The Icon API should remain the same, but if an icon has a smaller version (either new SVG or paths or ??) then it should use that. If not, then it should scale via css.

### Use case
We need to add chevrons of different sizes. And instead of adding them as completely different assets SVGs with names like `chevron-medium` and `chevron-small`, it would be nice to let the person using the SVG just used the size prop to pull in the correct asset.

image

kuzhelov:
Fixes #1356 .

This PR provides `props` argument to SVG icon spec. This makes it possible to use `props` values, such as `size`, to define SVG paths accordingly.

## Example (added `props` arg)
```jsx
// call.tsx
export default {
  icon: ({ classes, props }) => (
     {props.size === 'small' && (
           <svg>
            ...
          </svg>
     )}
     ...
} as TeamsSvgIconSpec
```

## Opened question on icon's box sizes (width and height)

It still remains to be an opened question whether changes to logic of applying box sizes to SVG element (such as `width` and `height`) should be provided. 

As it is suggested by #1356, in cases if `size` prop (with, say, `small` value) is handled by icon's render spec, `width` and `height` of the icon should remain to be default ones (and shouldn't be changed to `small` value). Specifically:

> but if an icon has a smaller version (either new SVG or paths or ??) then it should use that. **If not, then it should scale via css**.

However, this behavior may provide unexpected effects to the client for, say, the following code:

```jsx
// this icon renders different SVG path for 'smaller', 
// but its box's width and height doesn't change with 'size' prop
<Icon name='call' size='smaller' bordered> 

<Icon name='video' size='smaller' bordered>
```

Lets suppose that **only `call` icon** provides different SVG path for `smaller` size - and, thus, box size of this `call` icon won't be changed to `smaller` width and height values. In that case this will be the rendered result:

image

Even given that both icons were rendered as `smaller`, they have different box sizes which, in its turn, breaks alignment.

-----

My proposal here would be to guarantee that _box sizing_ algorithms work the same way for icons of any type, the same invariant we had before for `Icon` styles - as this guarantees proper alignment for any icons defined. Should note that this strategy will be sufficient enough to address the case of #1359.

However, it may be the case that I am missing some use cases missing with that - so, lets discuss and get to decision before merging the PR. FYI @codepretty


mnajdova:
> My proposal here would be to guarantee that box sizing algorithms work the same way for icons of any type, the same invariant we had before for Icon styles - as this guarantees proper alignment for any icons defined. Should note that this strategy will be sufficient enough to address the case of #1359.

I agree with this, all types of icons should behave the same, and depending on the theme's requirement, we can make the box (border around the icon) to grow or always be the same for all icons. For the icons that do not support different sizes, the box will still always be the same, as well as the icon size.

packages/react/src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -23,10 +23,8 @@ const getBorderedStyles = (boxShadowColor: string): ICSSInJSStyle => {
   }
 }
 
-const getIconSize = (size: SizeValue, sizeModifier: IconSizeModifier, v: IconVariables): string => {
-  if (!sizeModifier) {
-    return v[`${size}Size`]
-  }
+const getIconSize = (size: SizeValue, v: IconVariables): string => {
+  const sizeModifier: IconSizeModifier = v.sizeModifier
```
mnajdova: Can we remove this const now?

packages/react/src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -35,7 +33,9 @@ const getIconSize = (size: SizeValue, sizeModifier: IconSizeModifier, v: IconVar
     },
   }
 
-  return modifiedSizes[size] && pxToRem(modifiedSizes[size][sizeModifier])
+  return sizeModifier && modifiedSizes[size] && modifiedSizes[size][sizeModifier]
```
mnajdova: Actually, in my opinion the prev conditional is more readable. Is there a solid reason why it was changed?

packages/react/src/themes/teams/components/Icon/iconStyles.ts
```diff
@@ -80,7 +80,7 @@ const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
 
   svg: ({ props: { size, color, disabled, rotate }, variables: v }): ICSSInJSStyle => {
     const colors = v.colorScheme[color]
-    const iconSizeInRems = getIconSize(size, v.sizeModifier, v)
+    const iconSizeInRems = getIconSize(size, v)
```
mnajdova, line 81: 👍

packages/react/src/themes/teams/components/Icon/iconStyles.ts
kuzhelov: this change was necessary to remove branching at method's start:

```jsx
 if (!sizeModifier) {
    return v[`${size}Size`]
 }
```

Essentially, in this method we have two branches to handle, thus it is an expected thing that we need just one `if` for that. Previously there were two `if`s (at start of the method and at `return`) for, essentially, differentiate between the same two branches (`sizeModifier` is provided or not), and that has complicated the overall's function logic - and now there is only one.

mnajdova (approved):

PR #1564, opened 2019-07-01
fix(Dropdown): fix Dropdown ArrowLeft keydown behavior for multiple variants

### LINKED ISSUE #1129: Accessibility > Dropdown > Multiple Search/Selection - Not able to navigate to selected user, when tab/shifTab
### Steps
1. Insert couple of user to multiple search dropdown 
2. Navigate by left arrow key into selected user 
3. press Shift+TAB (focus is outside of dropdown component)
4. press TAB (focus is back in dropdown) on the last selected user
5. press TAB (focus is in the input field)
6. press left arrow key to navigate to selected user

### Expected Result
user is able navigate to the selected user

### Actual Result
user is NOT able to navigate to selected user(s) at all...

### Comment
Multiple Selection dropdown is affected in similar way
TAB and then Shift+TAB will cause the same issue.

not_possible_to_navigate_to_users

silviuaavram:
fixes https://github.com/stardust-ui/react/issues/1129

It was broken when the user was (shift-)tabbing from the last selected item. If the last item was already marked as active, then ArrowLeft from SearchInput would not re-focus it, due to the logic in `DropdownSelectedItem`:

`if (!prevProps.active && this.props.active) {`

I don't like the solution chosen but I cannot think of anything else less intrusive than this.


jurokapsiar (approved):

jurokapsiar:
please add changelog

PR #1565, opened 2019-07-01
fix(Popup): Move `trapFocus` and `autoFocus` to props from accessibility behaviors

### LINKED ISSUE #1493: Inline Popup not Accessible
## Bug Report

It appears that a popup, which is rendered as part of a Menu is not accessible when the `inline` prop is set on it as well as the `popupFocusTrapBehavior`.

### Steps
1. [Open test codesandbox](https://codesandbox.io/s/stardust-ui-example-bd3zj)
1. Click on the Popup that uses the `inline` prop.
1. Focus is set on the first focusable element in the popup
1. Try to navigate to the next focusable element via TAB or SHIFT+TAB

### Expected Result
Focus moves to the next focusable element.

### Actual Result
Focus doesn't move.

### Version
0.32

sophieH29:
closes https://github.com/stardust-ui/react/issues/1493

# BREAKING CHANGES

- deleted `popupFocusTrapBehavior`. `trapFocus` prop added to `Popup`.
- deleted `popupAutoFocusBehavior`. `autoFocus` prop added to `Popup`.

### Improvements
- `FocusTrapZone` and `AutoFocusZone` are now integrated into `PopupContent` without rendering an additional `div`.
- `trapFocus` can be set to popup using `true` value or an object of `FocusTrapZoneProps`
- `autoFocus` can be set to popup using `true` value or an object of `AutoFocusZoneProps`
- warn user of using combination of `inline` and `trapFocus` props. Or `autoFocus` for non-`inline` Popup.


docs/src/views/AutoFocusZone.tsx
```diff
@@ -16,25 +16,30 @@ export default () => (
...
+      In {code('Popup')}, set {code('autoFocus')} to {code('true')} with default settings or set an
+      object with desired values for auto focus zone props.
     </p>
+    <p>By default, focus will be set to the first tabbable element in the Popup:</p>
+    <CodeSnippet
+      label="PopupExample.jsx"
+      value={`
+        const Popup = () => (
+          <Popup autoFocus />
+        )`}
```
layershifter: nit: can be one liner:

```jsx
value='const Example = () => <Popup autoFocus />'
```

**!**: At least name should be changed to avoid collisions with component

docs/src/views/AutoFocusZone.tsx
```diff
@@ -44,31 +49,15 @@ export default () => (
...
-      
-        return behavior
-      }
-      `}
-    />
-    And then use this new behavior by Popup component:
     <CodeSnippet
       label="PopupExample.jsx"
       value={`
         const Popup = () => (
```
layershifter, line 53:
```suggestion
        const Example = () => (
```

docs/src/views/FocusTrapZone.tsx
```diff
@@ -32,16 +33,24 @@ export default () => (
...
+      To apply {code('FocusTrapZone')} to {code('Popup')} set prop {code('trapFocus')} to
       {code('true')} with default settings or set an object with desired values for focus trap zone
-      props. <Link to="accessibility-behaviors">Read more about Accessibility Behaviors.</Link>{' '}
-      Currently, it is used for Popup via {code('popupFocusTrapBehavior')} and Dialog via{' '}
-      {code('dialogBehavior')}.
+      props.
+    </p>
+    <CodeSnippet
+      label="PopupExample.jsx"
+      value={`
```
layershifter:
```suggestion
      value='const Example = () => <Popup trapFocus />'`
```

docs/src/views/FocusTrapZone.tsx
```diff
@@ -50,31 +59,23 @@ export default () => (
...
       value={`
-      const overridenFocusTrapBehavior: Accessibility = (props: any) => {
-        const behavior = popupFocusTrapBehavior(props)
-      
-        behavior.trapFocus.disableFirstFocus = true;
-      
-        return behavior
-      }
-      `}
+        const Popup = () => (
```
layershifter, line 63:
```suggestion
        const Example = () => (
```

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -212,10 +218,23 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
...
+        console.warn(
+          'Using "trapFocus" in inline popup leads to broken behavior for screen reader users.',
+        )
+      }
+      if (!inline && autoFocus) {
+        console.warn(
+          'Beware, "autoFocus" prop will just grab focus at the moment of mount and will not trap it. As user is able to TAB out from popup, better use "inline" prop to keep correct tab order.',
+        )
+      }
+    }
```
layershifter, line 230: Not sure about logging in `render()`, it will produce a warning on each render cycle.

What about `componentDidMount()`?

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -414,42 +433,34 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
...
-
     const popupContent = Popup.Content.create(content, {
       defaultProps: {
         ...popupContentAttributes,
         placement,
         pointing,
         pointerRef: this.pointerTargetRef,
-        unstable_wrapped: accessibility.focusTrap || accessibility.autoFocus,
+        trapFocus,
+        autoFocus,
```
layershifter, line 463: ❤️

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -414,42 +433,34 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
...
+      mountDocument,
+      pointing,
+      trapFocus,
+      autoFocus,
+    } = this.props
     const content = renderContent ? renderContent(scheduleUpdate) : propsContent
     const documentRef = toRefObject(mountDocument)
 
-    const popupWrapperAttributes = {
+    const popupContentAttributes = {
```
layershifter: It seems that `popupContentAttributes` is useless now and can be merged with `popupContent`:

```jsx
    const popupContent = Popup.Content.create(content, {
      defaultProps: {
        ...(rtl && { dir: 'rtl' }),
        ...accessibility.attributes.popup,
        ...accessibility.keyHandlers.popup,
        className: popupPositionClasses,
        ...this.getContentProps(),
        placement,
        pointing,
        pointerRef: this.pointerTargetRef,
        trapFocus,
        autoFocus,
      },
      overrideProps: this.getContentProps,
    })
```

packages/react/src/components/Popup/PopupContent.tsx
```diff
@@ -123,7 +142,15 @@ class PopupContent extends UIComponent<WithAsProp<PopupContentProps>> {
...
-      </ElementType>
+      </>
+    )
+
+    return trapFocus ? (
+      <FocusTrapZone {...focusTrapZoneProps}>{popupContent}</FocusTrapZone>
+    ) : autoFocus ? (
+      <AutoFocusZone {...autoFocusZoneProps}>{popupContent}</AutoFocusZone>
+    ) : (
+      <ElementType {...popupContentProps}>{popupContent}</ElementType>
```
layershifter: I suggest to refactor it in a bit more obvious way:

```jsx
   const popupContentProps: PopupContentProps = {
      className: classes.root,
      ...rtlTextContainer.getAttributes({ forElements: [children, content] }),
      ...accessibility.attributes.root,
      ...unhandledProps,
      onMouseEnter: this.handleMouseEnter,
      onMouseLeave: this.handleMouseLeave,
    }

    const popupContent = (
      <>
        {pointing && (
          <Ref innerRef={pointerRef}>
            {Box.create({}, { defaultProps: { styles: styles.pointer } })}
          </Ref>
        )}
        {Box.create(
          {},
          {
            defaultProps: {
              children: childrenExist(children) ? children : content,
              styles: styles.content,
            },
          },
        )}
      </>
    )
    
    if (trapFocus) {
      const focusTrapZoneProps = {
        ...popupContentProps,
        ...((_.keys(trapFocus).length && trapFocus) as FocusTrapZoneProps),
        as: ElementType,
      }
      
      return <FocusTrapZone {...focusTrapZoneProps}>{popupContent}</FocusTrapZone> 
    }

    if (autoFocus) {
      const autoFocusZoneProps = {
        ...popupContentProps,
        ...((_.keys(autoFocus).length && autoFocus) as AutoFocusZoneProps),
        as: ElementType,
      }
      
      return <AutoFocusZone {...autoFocusZoneProps}>{popupContent}</AutoFocusZone>
    }
    
    return <ElementType {...popupContentProps}>{popupContent}</ElementType>
```

layershifter: 1. We will compute `focusTrapZoneProps` and `autoFocusZoneProps` only when they are really needed
2. We will avoid nested ternaries

sophieH29: ❤️

packages/react/src/components/Popup/Popup.tsx
sophieH29: but then if user updates some of the props, component re-renders, the console warn will disappear even if the problem is not fixed. Isn't it? So maybe componentDidUpdate?

layershifter: I am not that it's a real scenario. If it's really breaks accessibility probably it should throw an error instead. Let's discuss it offline

layershifter (approved):

PR #1576, opened 2019-07-02
docs(CodeSnippet): fixing issue with copy icon taking up entire page

notandrew:
Removing height property from copy icon SVG and replacing with style. Firefox ignored the height property and displayed the icon at full window size.

Before:
image

after:
image


layershifter:
Can we fix this also for IE11?

image

packages/docs-components/src/CodeSnippet/CodeSnippetLabel.tsx
```diff
@@ -36,8 +36,7 @@ const CodeSnippetLabel: React.FunctionComponent<CopySnippetLabelProps> = props =
           border: '1px solid #ccc',
           color: '#ccc',
           cursor: copyable ? 'pointer' : 'default',
-          display: 'grid',
-          gridTemplateColumns: 'auto auto',
+          display: 'flex',
```
layershifter, line 39: `grid` is not working in IE11

packages/docs-components/src/CodeSnippet/CodeSnippetLabel.tsx
```diff
@@ -17,7 +17,7 @@ const checkIcon = (
 )
 
 const copyIcon = (
-  <svg height="0.7rem" viewBox="0 0 16 16">
+  <svg style={{ height: '.7rem', width: '0.7rem' }} viewBox="0 0 16 16">
```
layershifter, line 20: `width` should be also fixed for IE11

layershifter:
### Chrome

image

### IE11

image

### Firefox

image

layershifter (approved):

PR #1578, opened 2019-07-03
feat(Checkbox): add `labelPosition` prop

### LINKED ISSUE #1535: Label for Checkbox component positioning on the left
## Feature Request
The Checkbox label should have a positioning option.

### Problem description
A clear and concise description of what the problem is. Ex. I have an issue when [...]

For PreJoin screen, we need the label of the Checkbox on the left instead of the right side (as it is now). 

### Proposed solution
A clear and concise description of what you want to happen. Add any considered drawbacks.
Add a positioning item to the label that could be { "start", "end" }. 

### MVP

layershifter:
Fixes #1535.

### `labelPosition={start|end}`

This PR adds a new `labelPosition` prop. Styles for a gap between `label` and `icon` was inspired by `Flex` component.

image

### Docs refactor
- `Disabled` example was moved to `States` section
- Added an example for `Checked` state


packages/react/src/themes/base/components/Checkbox/checkboxStyles.ts
```diff
@@ -17,6 +17,10 @@ const checkboxStyles: ComponentSlotStylesInput<CheckboxProps & CheckboxState, Ch
       color: v.disabledColor,
       cursor: 'default',
     }),
+
+    '> *:not(:last-child)': {
```
mnajdova: Not sure about this selector. Do we really need this complication?

mnajdova (approved):

PR #1580, opened 2019-07-03
feat(Tooltip): add svg beak in Teams theme

mnajdova:
This PR adds svg pointing beak to the Tooltip component in Teams theme. No API changes were made.


packages/react/src/themes/teams/getPointerWithSvgStyles.ts
```diff
@@ -0,0 +1,55 @@
+import { PopperChildrenProps } from '../../lib/positioner'
+
+const rtlMapping = {
+  left: 'right',
+  right: 'left',
+}
+
+const getPointerWithSvgStyles = (
```
mnajdova: Should be refactored, there are lots of duplications with the `getPointerStyles.ts`

packages/react/src/themes/teams/getPointerWithSvgStyles.ts
```diff
@@ -0,0 +1,55 @@
...
+  pointerOffset: string,
+  pointerMargin: string,
+  rtl: boolean,
+  popperPlacement?: PopperChildrenProps['placement'],
+) => {
+  const placementValue = (popperPlacement || '').split('-', 1).pop()
+  const placement = (rtl && rtlMapping[placementValue]) || placementValue
```
mnajdova: Fix rtl issue

packages/react/src/themes/teams/components/Tooltip/tooltipContentStyles.ts
```diff
@@ -11,19 +26,22 @@ const tooltipContentStyles: ComponentSlotStylesInput<
     borderRadius: v.borderRadius,
     display: 'block',
     maxWidth: v.maxWidth,
-
-    ...(p.pointing && getPointerStyles(v.pointerOffset, v.pointerMargin, rtl, p.placement).root),
+    color: v.color,
+    background: v.backgroundColor,
+    ...(p.pointing &&
+      getPointerStyles(getPointerOffset(p.placement, v), v.pointerMargin, rtl, p.placement, true)
```
kuzhelov: I would suggest to calculate pointer styles once and store them, given how long list of `getPointerStyles` arguments is. Something like

```jsx
// here we will be able to capture semantics of these styles in variable name
const svgPointerStyles = getPoiterStyles(...)

return {
  root: () => ({
    ...,
    ...svgPointerStyles.root
  }),
  pointer: () => ({
    ...,
    ...svgPointerStyles.pointer
  })
}
```

packages/react/src/themes/teams/components/Tooltip/tooltipContentStyles.ts
```diff
@@ -2,6 +2,21 @@ import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
 import { TooltipContentProps } from '../../../../components/Tooltip/TooltipContent'
 import { TooltipContentVariables } from './tooltipContentVariables'
 import getPointerStyles from '../../getPointerStyles'
+import svgContent from './tooltipPointerSvgUrl'
```
kuzhelov: nit: maybe `tooltipPointerSvg` would be more descriptive? I mean, surely, it is more verbose, but it makes the following line much easier to read, as `svgContent` is very abstract

```jsx
backgroundImage: svgContent(v.backgroundColor),
```

kuzhelov (approved):

packages/react/src/themes/teams/components/Tooltip/tooltipPointerSvgUrl.ts
```diff
@@ -0,0 +1,4 @@
+export default (backgroundColor: string) =>
+  `url(\"data:image/svg+xml,%3C%3Fxml version='1.0' encoding='utf-8'%3F%3E%3C!-- Generator: Adobe Illustrator 22.0.1, SVG Export Plug-In . SVG Version: 6.00 Build 0) --%3E%3Csvg version='1.1' id='Layer_1' xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink' x='0px' y='0px' fill='${encodeURIComponent(
```
layershifter: Can we remove cruft from there, like `Generator` comment, `id` and `xmlns;*`?

packages/react/src/themes/teams/getPointerStyles.ts
```diff
@@ -31,19 +32,19 @@ const getPointerStyles = (
...
     },
     bottom: {
       top: `-${pointerOffset}`,
-      transform: 'rotate(-135deg)',
+      transform: `rotate(${isSvg ? (rtl ? -90 : 90) : -135}deg)`,
     },
     left: {
       right: `-${pointerOffset}`,
-      transform: 'rotate(-45deg)',
+      transform: `rotate(${isSvg ? (rtl ? 0 : 180) : -45}deg)`,
```
layershifter: It's hard to read... What about this?

```suggestion
      transform: isSvg ? `rotate(${rtl ? 0 : 180}deg)` : 'rotate(-45deg)',
```

packages/react/src/themes/teams/components/Tooltip/tooltipPointerSvgUrl.ts
codepretty: If you run it through [SVGOMG](https://jakearchibald.github.io/svgomg/), it will remove a lot of the unnecessary stuff.

codepretty: i would suggest not naming this specific to tooltip, since we will want to use this same beak for any popup or coachmark  that is pointing.

packages/react/src/themes/teams/components/Tooltip/tooltipContentStyles.ts
mnajdova: I cannot extract this in the beginning, because the ComponentSlotStylesInput is an object, not a function, but both in the root and pointer, I can extract this as a variable in the beginning... Would that be enough?

PR #1581, opened 2019-07-03
feat(color): add new colors and schema tokens in Teams theme

mnajdova:
Added new `brand` and `onyx` colors and `background4` token for `default` and `brand` color schemes.


packages/react/src/themes/teams/colors.ts
```diff
@@ -158,7 +159,7 @@ export const transparentColors: TeamsTransparentColors = {
     600: undefined,
     700: undefined,
     800: 'rgba(27,26,26,0.9)',
-    900: undefined,
+    900: 'rgba(0,0,0,0.8)',
```
notandrew: Do we want this RGBA value instead of a HEX value? If so, can you add a comment to what that color's hex value is so we don't have duplicate colors built?

alinais (approved):

PR #1588, opened 2019-07-05
[RFC] feat(Accessibility): Make action handlers conditional based on Event object

### LINKED ISSUE #1502: Accessibility: Conditional key handlers
### Problem description
There is not a clear separation between behaviors and components if the action executed on the component should be different based on props or the event. See https://github.com/stardust-ui/react/pull/1488/files#r293815521

### Proposed solution
Allow conditional keyboard handlers that would only execute an action if a condition based on the props or event is met

sophieH29:
As agreed during the discussion, will keep this PR for further discussion on achieving better solution for having conditional action handler based on an event object.
Meanwhile, a split of action handlers was introduced in another PR https://github.com/stardust-ui/react/pull/1622
--------------------------------------

Fixes https://github.com/stardust-ui/react/issues/1502

**PROBLEM**
==================

There a lot of confusion in key actions, as initially key action was aimed to do always one thing, now it can do multiple things, plus on different conditions of props values and event object.

Example of such actions:
Dialog - `closeAndFocusTrigger`
```jsx
closeAndFocusTrigger: e => {
      this.handleDialogCancel(e)
      e.stopPropagation()

      _.invoke(this.triggerRef, 'current.focus')
},
```

Attachment:
```jsx
performClick = e => {
    if (e.currentTarget === e.target) {
      e.stopPropagation()
      this.handleClick(e)
    }
  }
```

TreeItem:
```jsx
collapseOrReceiveFocus: e => {
      const { items, open } = this.props

      e.preventDefault()

      // Focuses the title if the event comes from a child item.
      if (e.currentTarget !== e.target && items && items.length) {
        e.stopPropagation()
        this.itemRef.current.focus()
      } else if (open) {
        e.stopPropagation()
        _.invoke(this.props, 'onTitleClick', e, this.props)
      }
    },
```

Also already discussed with @kuzhelov, here is a comment which is totally valid
image

**SOLUTION**
==================
I've made some investigation, things that I found out:
- actions are called in that sequence that they appear in behaviors.
For example, we have the next defined actions:

```jsx
 keyActions: {
    root: {
      Spacebar: {
        keyCombinations: [{ keyCode: keyboardKey.Spacebar }], 
      },
      enterSpace: {
      },
      performClick: {
      },
      enter: {
        keyCombinations: [{ keyCode: keyboardKey.Enter }],
      },
    },
  },
```

When a user press `SPACE`, the sequence of called actions will be
1. Spacebar
2. enterSpace
3. performClick

When a user press `ENTER`, the sequence of called actions will be
1. enterSpace
2. performClick
3. enter

That means, in cases where we have actions like `closeAndFocusTrigger` in `Dialog`

`keyActions` can be transformed **FROM**:
```jsx
keyActions: {
      popup: {
        closeAndFocusTrigger: {
        },
  },
```
**TO**:
```jsx
keyActions: {
      popup: {
        close: {
        },
        focusTrigger: {
        },
  },
```
And then implementation in `Dialog`:
**FROM**:
```jsx
actionHandlers = {
    closeAndFocusTrigger: e => {
      this.handleDialogCancel(e)
      e.stopPropagation()

      _.invoke(this.triggerRef, 'current.focus')
    },
  }
```

**TO**:
```jsx
actionHandlers = {
    close: e => this.handleDialogCancel(e),
    focusTrigger: e => {
      e.stopPropagation()
      _.invoke(this.triggerRef, 'current.focus')
    },
  }
```

Also, for cases when we depend on conditions like checking values of props and state, event object, we can extend API of `keyActions` and allow user to specify condition when the specific action should be called.

Extended API:
```jsx
export interface KeyAction {
  keyCombinations: KeyCombinations[]
  condition?: (event, props) => boolean
}
```

I introduced changes in `TreeItem` and `treeItemBehavior` to showcase how it will work.

Example of usage of `condition` prop in `treeItemBehavior`
```jsx
    ...
    receiveFocus: {
        keyCombinations: [{ keyCode: keyboardKey.ArrowLeft }],
        condition: doesEventComesFromChildItem,
      },
      collapse: {
        keyCombinations: [{ keyCode: keyboardKey.ArrowLeft }],
        condition: (event, props: TreeItemBehaviorProps) =>
          !doesEventComesFromChildItem(event, props) && props.open,
      },
   ...
```

And then in `TreeItem`:
```jsx
receiveFocus: e => {
      e.preventDefault()
      e.stopPropagation()

      this.itemRef.current.focus()
    },
collapse: e => {
      e.preventDefault()
      e.stopPropagation()

      _.invoke(this.props, 'onTitleClick', e, this.props)
},
```

**CONCERNS**
==================

I hesitate if should we move all conditions to behaviors as they're a part of the inner logic of components. Clients have no idea how they can impact if to change them.
I want us to discuss and agree on these points:
- separate combined actions into small actions, so then we'll have few actions assigned to the same key(s).
- move conditions logic which defines if the action should be called to behaviors OR as we'll have broken down small actions, leave conditions inside them.


packages/react/src/lib/accessibility/Behaviors/Tree/treeItemBehavior.ts
```diff
@@ -30,21 +32,46 @@ const treeItemBehavior: Accessibility<TreeItemBehaviorProps> = props => ({
...
+        receiveFocus: {
+          keyCombinations: [{ keyCode: keyboardKey.ArrowLeft }],
+          shouldHandle: doesEventComesFromChildItem,
+        },
+        collapse: {
+          keyCombinations: [{ keyCode: keyboardKey.ArrowLeft }],
+          shouldHandle: event => !doesEventComesFromChildItem(event),
+        },
+      }),
+      ...(!isSubtreeOpen(props) && {
```
miroslavstastny, line 45: I am not sure I like the combination of conditionally **adding** key handlers based on `props` and conditionally **invoking** them based on `event` (and eventually `props`).
Wouldn't it be easier to switch everything to `shouldHandle` condition? (I am aware of the fact that this can result in added keyhandlers which never call any action.)
Other opinions?

packages/react/src/lib/accessibility/Behaviors/Tree/treeItemBehavior.ts
```diff
@@ -30,21 +32,46 @@ const treeItemBehavior: Accessibility<TreeItemBehaviorProps> = props => ({
...
-        keyCombinations: [{ keyCode: keyboardKey.ArrowRight }],
-      },
+      ...(isSubtreeOpen(props) && {
+        receiveFocus: {
+          keyCombinations: [{ keyCode: keyboardKey.ArrowLeft }],
+          shouldHandle: doesEventComesFromChildItem,
+        },
+        collapse: {
+          keyCombinations: [{ keyCode: keyboardKey.ArrowLeft }],
+          shouldHandle: event => !doesEventComesFromChildItem(event),
```
miroslavstastny, line 42: nit
```suggestion
          shouldHandle: event => !eventComesFromChildItem(event),
```

packages/react/src/lib/accessibility/Behaviors/Tree/treeItemBehavior.ts
sophieH29: @miroslavstastny I totally share your thought, have same concerns. 
@kuzhelov @mnajdova @layershifter need your input here

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -113,42 +113,44 @@ class TreeItem extends UIComponent<WithAsProp<TreeItemProps>> {
...
     },
-    expandOrPassFocus: e => {
-      const { open } = this.props
+    expand: e => {
+      e.preventDefault()
+      e.stopPropagation()
 
+      this.handleTitleClick(e)
+    },
+    passFocus: e => {
```
levithomason, line 136: Suggestion: `focusSubtree`?  It would be nice to say _where_ we are passing the focus to.  Currently, it says to pass it, but as a dev implementing this I'm not sure where to pass it to.

packages/react/src/lib/accessibility/Behaviors/Tree/treeItemBehavior.ts
```diff
@@ -30,21 +32,46 @@ const treeItemBehavior: Accessibility<TreeItemBehaviorProps> = props => ({
...
-export default treeItemBehavior
+/** Checks if the event comes from a child item */
+const doesEventComesFromChildItem = (event): boolean => {
+  return event.currentTarget !== event.target
+}
+
+/** Checks if current tree item has a subtree and it is opened */
+const isSubtreeOpen = (props: TreeItemBehaviorProps): boolean => {
+  const { items, open } = props
+  return !!(items && items.length && open)
```
levithomason, line 76: Consider if it is enough to check the `open` prop.  If so, consider if we can collapse this function to inline logic.

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -113,42 +113,44 @@ class TreeItem extends UIComponent<WithAsProp<TreeItemProps>> {
...
     },
-    collapseOrReceiveFocus: e => {
-      const { items, open } = this.props
+    receiveFocus: e => {
+      e.preventDefault()
+      e.stopPropagation()
 
+      this.itemRef.current.focus()
+    },
+    collapse: e => {
```
levithomason, line 124: We've all agreed splitting the handlers into single purpose functions is correct.  let's get this merged in a separate PR.  we'll continue to discuss the `shouldHandle` and other concerns in this PR.

jdhuntington:

PR #1590, opened 2019-07-08
fix(ChatMessage): action menu doesn't disappear on click

mnajdova:
The action menu on the ChatMessages in Teams theme, should be open or closed based on the user interaction with the options within. For example, clicking on the emoticons, should close it, but clicking on the '...' should leave it open for the client to be able to choose one of the options.

For this to be possible, we need to allow the user to control this in some way. Currently the showing/hiding of the action menu is done only by changing the opacity of it, based on the state's focused prop. The simplest way to allow the user to control this is adding a boolean variable that can alter these styles - `showActionMenu`. In the future, if we decide that adding/hiding the action menu is a common thing, we may make this component AutoControlled and change it via the props.

In my opinion this is the simplest way of fixing the issue, but we still need to make insure that we are not breaking any of the prev accessibility. @sophieH29 @jurokapsiar can we test this and ensure that the focusing behavior is still working as expected?


packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -85,7 +85,14 @@ const chatMessageStyles: ComponentSlotStylesInput<
     overflow: p.focused ? 'visible' : 'hidden',
 
     // hide and squash actions menu to prevent accidental hovers over its invisible area
-    opacity: p.focused ? 1 : 0,
+    opacity:
```
mnajdova: Is there a better way to do this?

docs/src/prototypes/chatMessages/ChatMessageWithPopover/ChatWithPopover.tsx
```diff
@@ -7,118 +7,130 @@ const janeAvatar = {
...
+        items={[
+          {
+            key: 'a',
+            message: {
+              content: (
+                <Chat.Message
+                  actionMenu={
+                    <Popover shouldCloseMenuHandler={val => setShouldCloseActionMenu(val)} />
+                  }
+                  variables={{ shouldCloseActionMenu }}
```
mnajdova: Variable added for opting-out the opacity of the action menu

docs/src/prototypes/chatMessages/ChatMessageWithPopover/ChatWithPopover.tsx
```diff
@@ -7,118 +7,130 @@ const janeAvatar = {
...
+    >
+      <Chat
+        items={[
+          {
+            key: 'a',
+            message: {
+              content: (
+                <Chat.Message
+                  actionMenu={
+                    <Popover shouldCloseMenuHandler={val => setShouldCloseActionMenu(val)} />
```
mnajdova: Callback for changing the state, regarding the newly added variable

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -83,12 +76,13 @@ const chatMessageStyles: ComponentSlotStylesInput<
     right: v.actionMenuPositionRight,
     top: v.actionMenuPositionTop,
     overflow: p.focused ? 'visible' : 'hidden',
-
-    // hide and squash actions menu to prevent accidental hovers over its invisible area
-    opacity: p.focused ? 1 : 0,
-    width: p.focused ? 'auto' : 0,
+    opacity: 0,
+    width: 0,
+    ...(v.open && {
```
mnajdova: This is the only source of truth about whether the action menu should be opened.

docs/src/prototypes/chatMessages/ChatMessageWithPopover/ReactionPopup.tsx
```diff
@@ -0,0 +1,58 @@
...
+    }
+  }
+
+  handleOpenChange = (e, { open }) => {
+    this.setState({ open })
+  }
+
+  render() {
+    return (
+      <Popup
```
mnajdova: This is the same ReactionPopup that we use in the other Prototype, with the only difference that it does not uses the autoFocusPopupBehavior. The reason for this is, if the submenu from the action menu is opened and we hover over some reaction, the focus is moved, thus the submenu is closed. Not sure if this is the behavior we want here...

docs/src/prototypes/chatMessages/ChatMessageWithPopover/ChatWithPopover.tsx
```diff
@@ -1,124 +1,177 @@
...
+          {
+            key: 'a',
+            message: {
+              content: (
+                <Chat.Message
+                  actionMenu={
+                    <Popover setFixedMode={setFixedModeCM1} setOpen={val => setOpenCM1(val)} />
+                  }
+                  onMouseEnter={() => setOpenCM1(true)}
+                  onMouseLeave={() => !fixedModeCM1 && setOpenCM1(false)}
```
mnajdova: On mouse leave we will close the action menu, unless it is in fixed mode (the submenu is opened).

docs/src/prototypes/chatMessages/ChatMessageWithPopover/ChatWithPopover.tsx
```diff
@@ -1,124 +1,147 @@
...
+const TeamsChatMessage: React.FC<ChatMessageProps> = (props: ChatMessageProps) => {
+  const [open, setOpen] = React.useState(false)
+  const [fixedMode, setFixedMode] = React.useState(false)
+
+  return (
+    <Chat.Message
+      {...props}
+      actionMenu={<Popover setFixedMode={setFixedMode} setOpen={setOpen} />}
+      onMouseEnter={() => setOpen(true)}
+      onMouseLeave={() => !fixedMode && setOpen(false)}
```
mnajdova: On mouse leave we will close the action menu, unless it is in fixed mode (the submenu is opened).

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -83,12 +74,13 @@ const chatMessageStyles: ComponentSlotStylesInput<
     right: v.actionMenuPositionRight,
     top: v.actionMenuPositionTop,
     overflow: p.focused ? 'visible' : 'hidden',
-
-    // hide and squash actions menu to prevent accidental hovers over its invisible area
-    opacity: p.focused ? 1 : 0,
-    width: p.focused ? 'auto' : 0,
+    opacity: 0,
+    width: 0,
+    ...(v.actionMenuOpened && {
```
mnajdova: This is the only source of truth about whether the action menu should be opened.

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -82,13 +71,12 @@ const chatMessageStyles: ComponentSlotStylesInput<
     position: 'absolute',
     right: v.actionMenuPositionRight,
     top: v.actionMenuPositionTop,
-    overflow: p.focused ? 'visible' : 'hidden',
 
-    // hide and squash actions menu to prevent accidental hovers over its invisible area
-    opacity: p.focused ? 1 : 0,
-    width: p.focused ? 'auto' : 0,
+    visibility: 'hidden',
```
mnajdova: This is the only source of truth about whether the action menu should be opened.

docs/src/prototypes/chatMessages/ChatMessageWithPopover/ChatWithPopover.tsx
```diff
@@ -1,124 +1,147 @@
...
+    variables: { meReacting: true },
+  },
+  {
+    icon: 'thumbs down',
+    content: 2,
+    key: 'dislikes',
+  },
+]
+
+const reactionsWithPopup = _.map(reactions, reaction => render =>
```
kuzhelov: nit: would suggest to use native `map` method here, as we are not consuming any functionality that is exclusively provided by lodash. As this is the only place where lodash is consumed, we the could just avoid this import.

docs/src/prototypes/chatMessages/ChatMessageWithPopover/ChatWithPopover.tsx
```diff
@@ -1,124 +1,147 @@
...
-        },
-      ]}
+        ]}
+      />
+    </Provider>
+  )
+}
+
+const TeamsChatMessage: React.FC<ChatMessageProps> = (props: ChatMessageProps) => {
+  const [open, setOpen] = React.useState(false)
```
kuzhelov: may suggest to rename this state variable to `showActionMenu` - it will better suggest its semantics

docs/src/prototypes/chatMessages/ChatMessageWithPopover/ChatWithPopover.tsx
```diff
@@ -1,124 +1,147 @@
...
+  const [fixedMode, setFixedMode] = React.useState(false)
```
kuzhelov: `forceShowActionMenu`?

docs/src/prototypes/chatMessages/ChatMessageWithPopover/Popover.tsx
```diff
@@ -4,6 +4,8 @@ import cx from 'classnames'
 
 export interface PopoverProps {
   className?: string
+  setFixedMode?: React.Dispatch<React.SetStateAction<boolean>>
```
kuzhelov: mb `forceOpen`? also, to align with the Controlled pattern, we may name these props as `forceOpen` and `open`, pass boolean as values, so that those will be reflected by `Popover`'s visual appearance

docs/src/prototypes/chatMessages/ChatMessageWithPopover/ReactionPopup.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return `${numberOfPersons} people reacted this message with a ${emojiType} emoji. Open menu to see people who reacted.`
+}
+
+class ReactionPopup extends React.Component<ReactionProps, { open: boolean }> {
+  state = {
+    open: false,
+  }
+
+  handleKeyDownOnMenu = e => {
+    if ((e.shiftKey && e.keyCode === keyboardKey.Tab) || e.keyCode === keyboardKey.Tab) {
```
kuzhelov: `e.keyCode === keyboardKey.Tab` is an equivalent of both conditions

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix `Dropdown` ArrowLeft keydown behavior for multiple variants @silviuavram ([#1564](https://github.com/stardust-ui/react/pull/1564))
 - Fix styles for shadows in `Popup` in Teams theme [redlines] @codepretty ([#1561](https://github.com/stardust-ui/react/pull/1561))
 - Fix typings for `Provider` props @miroslavstastny ([#1601](https://github.com/stardust-ui/react/pull/1601))
+- Fix condition for showing the `actionMenu` inside the `ChatMessage` in Teams theme @mnajdova ([#1590](https://github.com/stardust-ui/react/pull/1590))
```
kuzhelov: provided changes, unfortunately, are breaking ones

docs/src/prototypes/chatMessages/ChatMessageWithPopover/Popover.tsx
mnajdova: Not sure I understand the requested changes. I can rename these to `setOpen` and `setForceOpen` or `forceOpen`, but those are callback, not sure how to change them to booleans...

docs/src/prototypes/chatMessages/ChatMessageWithPopover/ReactionPopup.tsx
```diff
@@ -0,0 +1,55 @@
...
+  return `${numberOfPersons} people reacted this message with a ${emojiType} emoji. Open menu to see people who reacted.`
+}
+
+class ReactionPopup extends React.Component<ReactionProps, { open: boolean }> {
+  state = {
+    open: false,
+  }
+
+  handleKeyDownOnMenu = e => {
+    if (e.keyCode === keyboardKey.Tab) {
```
sophieH29, line 18: why not close the popup menu on blur?

docs/src/prototypes/chatMessages/ChatMessageWithPopover/Popover.tsx
kuzhelov: oh, yes, I see - in that case, still, by following Controlled pattern, lets name props accordingly: they are designed to request state changes, so lets use `on...` naming convention, something like the following. This will be just about rename refactoring.

```jsx
...
 return (
    <Chat.Message
       ...
       actionMenu={<Popover onForceOpenChange={setForceOpen} onOpenChange={setOpen} />}
/>)
```

docs/src/prototypes/chatMessages/ChatMessageWithPopover/ReactionPopup.tsx
mnajdova: blur is invoked even if I press arrow keys (up/down), so I was not able to use this event.

docs/src/prototypes/chatMessages/ChatMessageWithPopover/Popover.tsx
```diff
@@ -44,27 +48,47 @@ class Popover extends React.Component<PopoverProps, PopoverState> {
...
+              setOpen(false)
+              chatMessageRef.current.focus()
+            },
           },
           {
             key: 'smile2',
             icon: 'smile',
             className: 'smile-emoji',
             'aria-label': 'smile two',
+            onClick: () => {
```
sophieH29: nit: you have the same repeatable click handler, can be moved to 1 func
```jsx
handleItemClick = () => {
              setOpen(false)
              chatMessageRef.current.focus()
  },
...
onClick: this.handleItemClick
```

sophieH29 (approved):

docs/src/prototypes/chatMessages/ChatMessageWithPopover/Popover.tsx
mnajdova: I renamed that using the state names, `showActionMenu` => `onShowActionMenuChange` and `forceShowActionMenu` => `onForceShowActionMenuChange`

docs/src/prototypes/chatMessages/index.tsx
```diff
@@ -18,11 +17,5 @@ export default () => (
     >
       <ImportantAndMentionMessages />
     </ComponentPrototype>
-    <ComponentPrototype
```
mnajdova, line 21: Removing this prototype as the reactions with popup is already included in the first one

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -53,12 +53,15 @@ const chatMessageStyles: ComponentSlotStylesInput<
 
     ...getBorderFocusStyles({ siteVariables, isFromKeyboard: p.isFromKeyboard }),
 
-    ':hover': {
-      [`& .${ChatMessage.slotClassNames.actionMenu}`]: {
-        opacity: 1,
-        width: 'auto',
+    ...(v.showActionMenu === undefined && {
```
kuzhelov: to make logic more robust, would suggest to capture additional cases where `v.showActionMenu` may be nullish (like, when variable's value is `null`), by either:
```js
v.showActionMenu == null
```
or rely on lodash:
```js
_.isNil(v.showActionMenu)
```

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
jurokapsiar: I'm affraid this will break the accessibiity for JAWS, I will test in short

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -82,13 +85,21 @@ const chatMessageStyles: ComponentSlotStylesInput<
...
+      overflow: p.focused ? 'visible' : 'hidden',
+      // hide and squash actions menu to prevent accidental hovers over its invisible area
+      opacity: p.focused ? 1 : 0,
+      width: p.focused ? 'auto' : 0,
+    }),
 
+    ...(v.showActionMenu !== undefined && {
+      visibility: 'hidden',
+      ...(v.showActionMenu && {
```
miroslavstastny: nit: ternary operator?

docs/src/prototypes/chatMessages/ChatMessageWithPopover/Popover.tsx
```diff
@@ -31,10 +34,20 @@ class Popover extends React.Component<PopoverProps, PopoverState> {
...
+    const { onShowActionMenuChange, chatMessageRef } = this.props
+    onShowActionMenuChange(false)
+    // Currently when the action menu is closed because of some actionable item is clicked, we focus the ChatMessage
+    // this was not in the spec, so it may be changed if the requirement is different
+    e.type === 'keydown' && chatMessageRef && chatMessageRef.focus()
+  }
+
   render() {
+    const { onShowActionMenuChange, onForceShowActionMenuChange, ...rest } = this.props
+    delete rest.chatMessageRef
```
kuzhelov: nit: maybe just mention this prop while destructuring the object? I think it will be cleaner as all the intents will be reflected by single expression. 

```js
// it is clear from this line what 'rest' is, 
// and we are not mutating this 'rest' object anywhere in the code afterwards
const { onShowActionMenuChange, onForceShowActionMenuChange, chatMessageRef, ...rest } = this.props
```

However, please, feel free to decide what is better :)

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
mnajdova: Added

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -82,13 +86,21 @@ const chatMessageStyles: ComponentSlotStylesInput<
...
-  }),
+    ...(_.isNil(v.showActionMenu) && {
+      overflow: p.focused ? 'visible' : 'hidden',
+      // hide and squash actions menu to prevent accidental hovers over its invisible area
+      opacity: p.focused ? 1 : 0,
+      width: p.focused ? 'auto' : 0,
+    }),
 
+    ...(v.showActionMenu !== undefined && {
+      visibility: 'hidden',
```
jurokapsiar: I'm affraid this will break accessibility for JAWS

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -53,12 +53,15 @@ const chatMessageStyles: ComponentSlotStylesInput<
 
     ...getBorderFocusStyles({ siteVariables, isFromKeyboard: p.isFromKeyboard }),
 
-    ':hover': {
-      [`& .${ChatMessage.slotClassNames.actionMenu}`]: {
-        opacity: 1,
-        width: 'auto',
+    ...(v.showActionMenu === undefined && {
+      ':hover': {
```
kuzhelov, line 61: I think that it would be helpful to provide short comment on why this variable's value is checked, to clearly suggest its intent. Something like: 

> actions menu's appearance can be controlled by value of `showActionMenu` variable - in this case this variable will serve the single source of truth on whether actions menu should be shown. Otherwise, if variable is not provided, the default appearance logic will be used for actions menu.

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
mnajdova: Will the combination of these props works better:
```
      overflow: v.showActionMenu ? 'visible' : 'hidden',
      // hide and squash actions menu to prevent accidental hovers over its invisible area
      opacity: v.showActionMenu ? 1 : 0,
      width: v.showActionMenu ? 'auto' : 0,
```

mnajdova: I changed it to work with this three props, from visual point of view, it behaves the same. Let me know if this is ok

docs/src/prototypes/chatMessages/ChatMessageWithPopover/Popover.tsx
mnajdova: Why I did this is, the chatMessageRef is not used anywhere in the code, so I am getting an error...

docs/src/prototypes/chatMessages/ChatMessageWithPopover/Popover.tsx
```diff
@@ -4,6 +4,9 @@ import cx from 'classnames'
 
 export interface PopoverProps {
   className?: string
+  onForceShowActionMenuChange?: (val: boolean) => void
+  onShowActionMenuChange?: (val: boolean) => void
+  chatMessageRef?: HTMLElement
```
kuzhelov: may suggest to rename it to `chatMessageElement` - because, if it is a ref as announced by prop, as a client I would expect Ref interface to be applicable: `chatMessageRef.current`. Here, in contrast, this prop provides element directly, not a ref on element

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -82,13 +89,20 @@ const chatMessageStyles: ComponentSlotStylesInput<
...
-    overflow: p.focused ? 'visible' : 'hidden',
 
-    // hide and squash actions menu to prevent accidental hovers over its invisible area
-    opacity: p.focused ? 1 : 0,
-    width: p.focused ? 'auto' : 0,
-  }),
+    ...(_.isNil(v.showActionMenu) && {
+      overflow: p.focused ? 'visible' : 'hidden',
+      // hide and squash actions menu to prevent accidental hovers over its invisible area
+      opacity: p.focused ? 1 : 0,
```
kuzhelov, line 96: could we augment this comment with a note about `opacity` vs `visibility` strategy - specifically, mention that this affects JAWS behavior on Windows? Just to prevent my future attempts to improve styles by using `visibility` CSS prop instead :)

mnajdova: :D sure

mnajdova: added

kuzhelov (approved):

PR #1600, opened 2019-07-09
chore(accessibility): remove `defaultBehavior`

layershifter:
# BREAKING CHANGES

This PR removes `defaultBehavior` because the single reason to have it is how `getAccessibility()` function is working.


packages/react/src/lib/renderComponent.tsx
```diff
@@ -53,18 +57,25 @@ export interface RenderConfig<P> {
 }
 
 const getAccessibility = (
-  props: State & PropsWithVarsAndStyles,
+  props: State & PropsWithVarsAndStyles & { accessibility?: Accessibility },
   actionHandlers: AccessibilityActionHandlers,
   isRtlEnabled: boolean,
-) => {
-  const { accessibility: customAccessibility, defaultAccessibility } = props
```
layershifter, line 60: We don't have `defaultAccessibility` prop at all 💥

jurokapsiar (approved):

docs/src/prototypes/MenuButton/MenuButton.tsx
```diff
@@ -150,10 +150,11 @@ class MenuButton extends React.Component<MenuButtonProps, MenuButtonState> {
     const { button, disabled, menu, placement } = this.props
     const { menuOpen } = this.state
     const [position, align] = _.split(placement, '-') as [Position, Alignment]
-    const accessibilityBehavior: ReactAccessibilityBehavior = menuButtonBehavior({
-      ...this.props,
-      ...this.state,
-    })
+    const accessibilityBehavior: ReactAccessibilityBehavior = {
+      attributes: {},
```
sophieH29: why do you need an empty `attributes` object here? I see attributes already exist in `menuButtonBehavior`? and also `keyHandlers` which aren't used. I mean I understand why you're adding these empty mocks to replicate same behavior as in `renderComponent`, but I think it's not very understandable for someone who is looking at this prototype. And I believe if you create a behavior which returns an object people are wise on which object's props they can reference to

sophieH29 (approved):
small comment! Other than that, great job 👍 ❤️

bmdalex (approved):

docs/src/prototypes/MenuButton/MenuButton.tsx
layershifter: Extracted this to empty behavior

Issue #1597, opened 2019-07-08
Code in Theming Examples does not have a corresponding output


lucivpav:
### Steps
1. Navigate to the [docs section](https://stardust-ui.github.io/react/theming-examples#nesting-themes).
2. Observe the code and its output.

### Expected Result
1. The code for defining the Default theming buttons is provided.
2. The _darkred_ brand color is applied.

### Actual Result
1. There is no code corresponding to creation of the Default theming buttons. Instead, a code for a button with content "Branding" is provided, however this is not rendered in the output.
2. There is no sign of _darkred_ brand color being applied.

### Version
0.34.0

PR #1616, opened 2019-07-11
fix(docs): Code in Theming Examples does not have a corresponding output

lucivpav:
This resolves issue #1597. The [code in docs](https://stardust-ui.github.io/react/theming-examples#nesting-themes) did not correspond to the actual output. Futhermore the code was outdated, resulting in not doing what it was intended to do.

Before:
orig

After:
image


docs/src/views/ThemingExamples.tsx
```diff
@@ -277,15 +277,27 @@ export default () => (
...
+          <Icon name="chess rook" size="larger" />
+          <Label content="Label with icon" icon="close" />
 
           /* First nested theming */
           <Provider
             theme={{
-              siteVariables: { brand: 'darkred' },
+              componentStyles: {
+                Button: {
+                  root: ({variables, props}) => ({
```
mnajdova: Would rather introduce variable for the color value, so that we won't hardcode the `darkred` value. Also, as a pattern and maybe better readability, we can transform the usage into the following:

```
            componentStyles: {
                Button: {
                  root: ({ variables: v, props: p }) => ({
                    ...(p.primary && {
                      background: v.backgroundColor,
                    }),
                  }),
                },
```

docs/src/views/ThemingExamples.tsx
```diff
@@ -330,7 +359,23 @@ export default () => (
...
+                Button: {
+                  primaryBackgroundColor: 'darkred',
+                },
+              },
+              componentStyles: {
+                Button: {
+                  root: ({
+                    variables: v,
+                    props: p,
+                  }: ComponentStyleFunctionParam<any, ThemingExamplesVariables>) => ({
```
lucivpav: This type is missing from the example snippet.

lucivpav: Which is a good thing, as we do not want to force users to use TypeScript

docs/src/views/ThemingExamples.tsx
```diff
@@ -330,7 +359,23 @@ export default () => (
...
+                    ...(p.primary && {
```
mnajdova: You do not need the componentStyles here again, as those are not overriding anything from before. Just the componentVariables part should be enough

docs/src/views/ThemingExamples.tsx
```diff
@@ -277,15 +277,24 @@ export default () => (
       value={`
         <div>
           /* Default theming */
```
layershifter:
```suggestion
          {/* Default theming */}
```

Should be JSX comment...

docs/src/views/ThemingExamples.tsx
```diff
@@ -277,15 +277,24 @@ export default () => (
...
-          <Header as="h3" content="Default" />
-          <Button primary>Branding</Button>
+          <Header as="h3" content="Default theming" />
+          <Button content="Button" />
+          <Button icon="plus" iconOnly primary />
+          <Button icon="at" content="Send email" secondary />
+          <Icon name="chess rook" size="larger" />
+          <Label content="Label with icon" icon="close" />
 
           /* First nested theming */
```
layershifter:
```suggestion
          {/* First nested theming */}
```

layershifter (approved):

PR #1617, opened 2019-07-11
feat(Dialog): add headerAction slot

mnajdova:
This PR adds a `headerAction` to the `Dialog` component. This button is `text` and `iconOnly` and is rendered after the `header` slot. I noticed that it is the first thing focused when the `Dialog` is opened (not sure if this is the behavior we want) @jurokapsiar @sophieH29 please share your thoughts on this. Also the focus indicator is not implemented for the `textOnly` variant for the button. I may add that as part of this PR.


docs/src/examples/components/Dialog/Content/DialogExampleContent.shorthand.tsx
```diff
@@ -7,6 +7,7 @@ const DialogExampleContent: React.FC = () => (
     confirmButton="Confirm"
     content="Are you sure you want to confirm this action?"
     header="Action confirmation"
+    headerAction={{ icon: 'more' }}
```
jurokapsiar: please change to button

jurokapsiar: plase add `title` prop to this and the next example - let's educate the devs that they need to provide the text for icon only buttons

jurokapsiar:
@sophieH29 let's discuss - we might do something like looking for the focusable in content first and then in the header actions if we did not find anything

@mnajdova yes, the focus indicator should be there - I will leave it up to you if you want to add it in this PR or if you want to create an issue - it probably also depends on if you have the red lines or not.

docs/src/examples/components/Dialog/Content/DialogExampleContent.shorthand.tsx
mnajdova: added

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -54,6 +55,9 @@ export interface DialogProps
   /** A dialog can contain a header. */
   header?: ShorthandValue
 
+  /** A dialog can contain an button next to the header. */
```
lucivpav:
```suggestion
  /** A dialog can contain a button next to the header. */
```

packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -148,6 +148,12 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps & ButtonState, ButtonVa
             color: textPrimaryColorHover,
           },
         }),
+
+        ':focus': {
```
mnajdova, line 152: Added focus styles for the text button

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -54,6 +55,9 @@ export interface DialogProps
   /** A dialog can contain a header. */
   header?: ShorthandValue
 
+  /** A dialog can contain a button next to the header. */
+  headerAction?: ShorthandValue
```
layershifter:
```suggestion
  headerAction?: ShorthandValue<ButtonProps>
```

🙏

docs/src/examples/components/Dialog/Content/DialogExampleContent.shorthand.tsx
```diff
@@ -7,6 +7,7 @@ const DialogExampleContent: React.FC = () => (
     confirmButton="Confirm"
     content="Are you sure you want to confirm this action?"
     header="Action confirmation"
+    headerAction={{ icon: 'more', title: 'See more options' }}
```
layershifter: Can we add a separate example instead? 🤔 

And probably `close` icon will have more sense at that place. We can use a knob to handle that:

```tsx
const [open, setOpen] = useBooleanKnob({ name: 'open })

<Dialog
  onOpen={() => setOpen(true)}
  onCancel=...
  onConfirm=...
  headerAction={{ onClick: () => setOpen(true) }}
/>
```

mnajdova: Created new example under the Content section.

packages/react/src/components/Dialog/Dialog.tsx
mnajdova: Added for all ShorthandValue props in the Dialog. :)

layershifter (approved):

packages/react/src/themes/teams/components/Dialog/dialogVariables.ts
```diff
@@ -26,4 +28,6 @@ export default (siteVariables): DialogVariables => ({
 
   overlayBackground: 'rgba(37, 36, 36, .75)', // todo: update to a palette value when daisy has mapped one
   overlayZIndex: 1000,
+
+  headerActionMargin: `${pxToRem(-3)} ${pxToRem(-8)} 0 0`,
```
bmdalex, line 32: I understand why this is done and I had the same issue a few times. `<Button iconOnly text />` has some default padding that breaks most of the layouts where it's used.

**Should we add a `fitted` prop on the `Button` to remove that?**

On the other side, because of the negative margin, it looks a bit strange when we have the focus border around it:

Screenshot 2019-07-12 at 12 07 52

So my question is, what is expected here?
- align the icon within the button with the rest of the layout in regular mode? (what @mnajdova did here)
- align the button itself so that it looks good in focus mode?

fyi @codepretty

bmdalex (approved):
one comment

packages/react/src/themes/teams/components/Dialog/dialogVariables.ts
mnajdova: @Bugaa92 I am not sure whether we should change it, because this is exactly how the focus styles should look by design:

image

I agree that it looks strange, but would rather not make any changes at this moment. If this would have been unexpected behavior, then yes, I agree that we should have reconsidered this, but at this moment it is consistent with the designs...

docs/src/examples/components/Dialog/Content/DialogExampleHeaderAction.shorthand.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+import { Button, Dialog } from '@stardust-ui/react'
+import { useBooleanKnob } from '@stardust-ui/docs-components'
+
+const DialogExampleHeaderAction: React.FC = () => {
+  const [open, setOpen] = useBooleanKnob({ name: 'open' })
+  return (
+    <Dialog
+      open={open}
```
miroslavstastny, line 9: Not sure that this example is 100% descriptive.

If `headerAction` is used for a close button (I bet most common case) like in this example, the component must be obviously controlled = consumer must also handle `onCancel` and `onConfirm` in order to close the dialog after clicking any action button or outside the dialog.

This example does not do that (to show just the header) but it can be confusing for a consumer.

miroslavstastny (approved):

packages/react/src/themes/teams/components/Dialog/dialogVariables.ts
bmdalex: got it, thanks for the explanation @mnajdova makes sense

miroslavstastny:
We should add screener tests for open dialogs, popups, tooltips... (probably in a separate task)
image

docs/src/examples/components/Dialog/Content/DialogExampleHeaderAction.shorthand.tsx
mnajdova: Maybe we should add more examples for the Dialog for controlling state/custom handlers. Do you suggest any change in this particular example? I mean we can remove the open knobs and do nothing when clicking the close icon, but I think that would be even more confusing..

mnajdova:
> We should add screener tests for open dialogs, popups, tooltips... (probably in a separate task)

@miroslavstastny I thought about the same thing, so I added steps for the Dialog - the new example and the example with the content... In a separate pr we can add more examples for the popups and tooltips.

docs/src/examples/components/Dialog/Content/DialogExampleHeaderAction.shorthand.tsx
miroslavstastny: Agree. Let's leave it as it is for now.

PR #1622, opened 2019-07-11
feat(Accessibility): Split action handlers with "OR" condition

sophieH29:
Based on this initial PR https://github.com/stardust-ui/react/pull/1588, there was made a decision to split actions in separate PR. And re-think conditional handlers in further discussions.

### DIFF
--- a/packages/react/src/lib/accessibility/Behaviors/Tree/treeItemBehavior.ts
+++ b/packages/react/src/lib/accessibility/Behaviors/Tree/treeItemBehavior.ts
@@ -9,8 +9,11 @@ import { IS_FOCUSABLE_ATTRIBUTE } from '../../FocusZone/focusUtilities'
  * Adds 'tabIndex' as '-1' if the item is not a leaf.
  *
  * @specification
- * Triggers 'collapseOrReceiveFocus' action with 'ArrowLeft' on 'root'.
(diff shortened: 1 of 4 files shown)


packages/react/src/lib/accessibility/Behaviors/Tree/treeItemBehavior.ts
```diff
@@ -47,4 +60,10 @@ export type TreeItemBehaviorProps = {
   open?: boolean
 }
 
+/** Checks if current tree item has a subtree and it is opened */
+const isSubtreeOpen = (props: TreeItemBehaviorProps): boolean => {
+  const { items, open } = props
+  return !!(items && items.length && open)
```
levithomason, line 66: Was it determined that checking `open` alone was not sufficient?

levithomason (approved):
Just one comment, address as you see fit and merge 👍

packages/react/src/lib/accessibility/Behaviors/Tree/treeItemBehavior.ts
sophieH29: yes, in `TreeItem` subtree is only rendered if `open` === `true` and `items` is not an empty array

Issue #1591, opened 2019-07-08
An icon in docs not rendered properly


lucivpav:
## Bug Report

An icon in docs with color variable set to _cornflowerblue_ is not rendered properly.
![image](https://user-images.githubusercontent.com/10930160/60808157-e370fb00-a187-11e9-9d33-5663c15f4e9f.png)

### Steps
1. Navigate to the section of [docs](https://stardust-ui.github.io/react/theming#component-variables)
2. Take a look at the first block of code and its output

### Expected Result
The second icon has a color of _cornflowerblue_.

### Actual Result
The second icon has a black (default) color.

### Version
0.34.0

PR #1625, opened 2019-07-12
fix(docs): icons in docs not rendered properly

lucivpav:
This PR updates the outdated code snippets in [theming section](https://stardust-ui.github.io/react/theming) of docs. The icon colors were not set properly and the output did not match user's expectation. Relevant issue with error description: #1591.

Before:
image

After:
image


CHANGELOG.md
```diff
@@ -44,6 +44,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix broken fragments on CodeSandbox @lucivpav ([#1655](https://github.com/stardust-ui/react/pull/1655))
 - Improve a table with component props in docs @layershifter ([#1634](https://github.com/stardust-ui/react/pull/1634))
 - Improve rendering performance for docs examples @Bugaa92 ([#1654](https://github.com/stardust-ui/react/pull/1654))
+- Fix code in component variables section of theming page @lucivpav ([#1625](https://github.com/stardust-ui/react/pull/1625))
```
kuzhelov: nit: lets provide an exact name of the docs section in the changelog - otherwise it is not easy to comprehend which page the entry is about:

> Fix `Theming` page code examples @lucivpav

docs/src/views/Theming.tsx
```diff
@@ -87,7 +99,7 @@ export default () => (
       render={() => (
         <>
           <Icon name="user" circular />
-          <Icon name="user" circular variables={{ color: 'cornflowerblue' }} />
+          <Icon name="user" circular styles={{ color: 'cornflowerblue' }} />
```
kuzhelov: while it makes the code example correct, unfortunately, it makes the intent of the paragraph to be invalid, as here we suggesting to use variables:

> You can define component variables on a single instance of a component.

But then we are providing a code example that is about `styles` customization.

Lets decide here on the following:
- would we like to restructure the page, so that `variables` customization will be replaced by `styles` custiomization (probably, no - as in that case it will also invalidate the statement we are making about sparing usage of styles customization)
- rework example to use _working_ variables based on color scheme (and provide the necessary references)
- omit the example entirely (or, maybe, replace with the one that is about using other variables, like `gap`)

@mnajdova, @layershifter

docs/src/views/Theming.tsx
```diff
@@ -97,12 +109,57 @@ export default () => (
...
+          <Icon name="user" circular />
+
+          <Provider
+            theme={{
+              componentVariables: {
+                Icon: {
+                  iconColor: 'cornflowerblue'
+                  }
+              },
+              componentStyles: {
```
kuzhelov: again, while this fixes an example, this makes it being non-aligned with the docs section: here, essentially, we are demoing _styles_ customization (even if based on _variables_ - but this complicates example even more). Lets agree on the resolution for this comment https://github.com/stardust-ui/react/pull/1625/files#r305273759, and then we would be able to address this issue in the same way.

docs/src/views/Theming.tsx
alinais: Let's try to use colors from the color palette, to keep the consistency across the API examples.

lucivpav:
Please note that only a small fraction of icons support the color variable. User thus may be confused, why his icon does not have a proper color. Relevant issue #1039.

lucivpav:
This PR is deprecated as it will be fixed as part of #1764

docs/src/views/Theming.tsx
```diff
@@ -97,15 +97,44 @@ export default () => (
       the <NavLink to="components/provider">Provider</NavLink>.
     </p>
     <ExampleSnippet
+      value={`
```
miroslavstastny: What is the `value` good for?

miroslavstastny (approved):

lucivpav:
This was fixed by PR #1764. Closing this as it is no longer relevant.

Issue #1592, opened 2019-07-08
Button in docs not rendered properly


lucivpav:
## Bug Report
There are two issues with a button being rendered in the docs. Firstly, the icon has a wrong color. This has been reported [here](https://github.com/stardust-ui/react/issues/1591). Secondly, the text of the button and the button's background is rendered using default values.

![image](https://user-images.githubusercontent.com/10930160/60808945-c3423b80-a189-11e9-865c-cad57dd5331f.png)

### Proposal
1. Change variable `secondaryColor` to `color`.
2. Change variable `secondaryBackgroundColor` to `backgroundColor`.

### Steps
1. Navigate to the section of [docs](https://stardust-ui.github.io/react/theming-examples#changing-component-variables).
2. Take a look at the first block of code and its output.

### Expected Result
1. The icon has a _brown_ color.
2. The button's background has a _charcoal_ color.
3. The button's text has a _colar_ color.

### Actual Result
1. The icon has a _black_ color.
2. The button's background has a default (_transparent_?) color.
3. The button's text has a _black_ color.

### Version
0.34.0

PR #1626, opened 2019-07-12
fix(docs): changing component variables example snippet

lucivpav:
This PR updates the outdated code snippets in [changing component variables](https://stardust-ui.github.io/react/theming-examples#changing-component-variables) section of docs. The buttons were not rendered as a user would expect. Relevant issue with error description: #1592.

Before:
image

After:
image


docs/src/views/ThemingExamples.tsx
```diff
@@ -121,32 +121,46 @@ export default () => (
...
-              variables: { color: "brown" }
-            }}
-            content="Send email"
-            variables={{
-              secondaryColor: "coral",
-              secondaryBackgroundColor: "charcoal",
-              paddingLeftRightValue: 20,
-            }}
-          />
+        <Button
```
kuzhelov, line 124: unfortunately, this leads to change in the example's semantics. Originally the purpose was to demo `variables` usage - specifically, how it is possible to customize variables and obtain different look of the component. Now what we have is the `styles` demo, with additional aspect of using `variables` as style parameters.

alinais: Talked to Roman about it - we should advertise the colors from the color palette rather than the usage of the styles.

lucivpav:
The example looks like this now:
image

lucivpav:
Please note that only a small fraction of icons support the color variable. User thus may be confused, why his icon does not have a proper color. Relevant issue #1039.

docs/src/views/ThemingExamples.tsx
lucivpav: @alinais do you mean, for example, `siteVars.colors.pink[50]` instead of, say, `pink`? I have updated the PR, please let me know whether you like it now :)

kuzhelov (approved):

PR #1611, opened 2019-07-10
feat(docs): make categories collapsible

lucivpav:
This PR changes the docs in a way that the sidebar categories can be collapsed. Previously this was not possible, as the sidebar component did not have a tree structure.

collaps1

The sidebar items can also be accessed using keyboard only.

collaps2

PR #1627, opened 2019-07-12
feat(docs): make prototypes conditionally public

lucivpav:
This PR contains several changes to the docs site:
1. Prototypes section moved below Behaviors.
2. Each prototype has a `public` flag, which determines whether the prototype should be visible in *Prod* build.
3. Custom Styled Toolbar is public.

In non-*Prod* build:
image

In *Prod* build:
image

This PR should be merged after #1611.


docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -337,62 +364,64 @@ class Sidebar extends React.Component<any, any> {
       {
         key: 'chatMssages',
         title: { content: 'Chat Messages', as: NavLink, to: '/prototype-chat-messages' },
+        public: false,
```
bmdalex, line 366: nit:
add `public: true` as required type for `prototypesTreeItems`. it will become:
```typescript
const prototypesTreeItems: (ShorthandValue & { key: string; public: boolean })[] = []
```

bmdalex (approved):

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -262,6 +262,33 @@ class Sidebar extends React.Component<any, any> {
...
   }
+  getPrototypesGivenEnvironment(allPrototypes) {
+    if (process.env.NODE_ENV !== 'production') {
+      return allPrototypes
+    }
+    return allPrototypes.filter(p => p.public)
+  }
+
+  getSectionsWithPrototypeSectionIfApplicable(currentSections, allPrototypes) {
+    let prototypes = this.getPrototypesGivenEnvironment(allPrototypes)
```
```jsx
const prototypes = process.env.NODE_ENV === 'production' ? _.filter(allPrototypes, { public: true }) : allPrototypes
```

PR #1629, opened 2019-07-12
fix(docs): component descriptions

kuzhelov:
## Goal
This PR aims to ensure that for each component there are the following pieces provided
- concise description that reflects component's purpose
- brief accessibility description

Visual aspects of how this data is presented by docs are left outside of this PR, as its main focus is to refine, organise and make consistent the actual _content_ (component description) that should be rendered.

## Unify component names' formatting in descriptions
Previously there were no consistency, and both these approaches were used for Stardust components (based on previous component descriptions):

> A **Alert** displays information that explains nearby content.

> A **button** indicates a possible user action.

This PR unifies the approach for all components, by favouring the uppercased way for component references. This is done for the following reasons:
- to ensure close relationship to other Stardust component is conveyed by the description, and it is not that just some general noun/term used.
- at times it even makes an important semantic distinction - just compare:

>  A ChatItem is container for Chat item's content.

> A chat item is container for chat item's content.

This is not the only example - similar semantic differences could be seen for the following components:
- `Input`
- `TollbarMenuDivider`
- `Embed`
- `ToolbarDivider`

## References used
- [WAI-ARIA 1.1](https://www.w3.org/TR/wai-aria/)
- [Reakit](https://reakit.io/docs/get-started/)
- [Reach UI](https://ui.reach.tech/)
- [Material UI](https://material-ui.com/components/buttons/)
- [React Bootstrap](https://react-bootstrap.github.io/components/alerts/)
- [Blueprint](https://blueprintjs.com/docs/)
- [Ant Design](https://ant.design/docs/react/introduce/)
- [Semantic UI React](https://react.semantic-ui.com/elements/button/)
- [Office UI Fabric](https://developer.microsoft.com/en-us/fabric#/controls/web/)


packages/react/src/components/Button/ButtonGroup.tsx
```diff
@@ -107,6 +107,6 @@ ButtonGroup.create = createShorthandFactory({
 })
 
 /**
- * A button group presents multiple related actions.
+ * A ButtonGroup represents multiple related actions in a horizontal or vertical group.
```
mnajdova: would avoid `horizontal or vertical` as the `ButtonGroup` do not even have the `vertical` prop at this moment

packages/react/src/components/Chat/ChatItem.tsx
```diff
@@ -147,6 +147,7 @@ ChatItem.slotClassNames = {
 }
 
 /**
- * A chat item represents a single event in a chat.
+ * A ChatItem is container for Chat item's content.
```
mnajdova: Maybe replacing the `Chat item's content` with something like `one Chat row` would be more descriptive

packages/react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -175,7 +175,9 @@ Checkbox.create = createShorthandFactory({
 })
 
 /**
- * A single checkbox within a checkbox group.
+ * A Checkbox allows a user to make choice from a small set of options.
```
mnajdova: `small set of options` sounds confusing in my opinion. Can we be more specific with this, or at least state dual-state for now, until we actually support more options

packages/react/src/components/Dropdown/DropdownItem.tsx
```diff
@@ -120,7 +120,7 @@ DropdownItem.slotClassNames = {
 DropdownItem.create = createShorthandFactory({ Component: DropdownItem, mappedProp: 'header' })
 
 /**
- * A sub-component of the Dropdown.
- * Used to display items of the dropdown list.
+ * A DropdownItem represents item of Dropdown.
```
mnajdova: If the DropdownItem is used to display one option in the list, let's state it like this. As we have lots of Component around the Dropdown - DropdownItem, DropdownLabel etc, so it would be good to make the separation of these components in the description.

packages/react/src/components/Menu/MenuDivider.tsx
```diff
@@ -70,6 +70,6 @@ class MenuDivider extends UIComponent<WithAsProp<MenuDividerProps>> {
 MenuDivider.create = createShorthandFactory({ Component: MenuDivider, mappedProp: 'content' })
 
 /**
- * A menu divider visually segments menu items inside menu.
+ * A MenuDivider is non-focusable element that visually segments items of Menu.
```
mnajdova: Can we replace the `non-focusable` with `non-actionable` element, as in that case it will be easy to reason what is the mean difference between `MenuItem` and `MenuDivider`

packages/react/src/components/Text/Text.tsx
```diff
@@ -103,8 +103,6 @@ class Text extends UIComponent<WithAsProp<TextProps>, any> {
 Text.create = createShorthandFactory({ Component: Text, mappedProp: 'content' })
 
 /**
- * A Text component formats occurrences of text consistently.
- * @accessibility
- * Text is how people read the content on your website.
+ * A Text component is used to consistently style and format occurrences of text.
```
mnajdova: We didn't have any description with `is used` so far in the Components. Can we reformulate this description in a similar format as the others?

packages/react/src/components/Menu/MenuDivider.tsx
kuzhelov: have replace all the occurrences of `non-focusable` to `non-actionable`

packages/react/src/components/Text/Text.tsx
kuzhelov: have replaced `used to` with active-voice verbs

packages/react/src/components/Dropdown/DropdownItem.tsx
kuzhelov: reformulated :)

packages/react/src/components/Checkbox/Checkbox.tsx
kuzhelov: replaced with direct `two options`, as this is the only mode we currently support

packages/react/src/components/Button/ButtonGroup.tsx
kuzhelov: removed

packages/react/src/components/Chat/ChatItem.tsx
kuzhelov: have simplified description to 
> A ChatItem is container for single event in Chat (e.g. message, notification, etc).

I think that it is better to avoid visual commitments (like, `row`) in the component's description, to ensure that only component's semantics is provided.

mnajdova (approved):
No other comments from my side. Let's be sure to check the final description from accessibility point of view and it is good to go 👍

packages/react/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -148,6 +148,6 @@ AccordionTitle.slotClassNames = {
 }
 
 /**
- * A standard AccordionTitle that is used to expand or collapse content.
+ * An AccordionTitle represents the title of Accordion's item that client could interact with to expand or collapse the item's content.
```
sophieH29: nit: Somehow I don't like using `client` here. And seems it's the only place where you use it.
How about `An AccordionTitle represents the title of Accordion's item that can be interacted with to expand or collapse the item's content.`

packages/react/src/components/Avatar/Avatar.tsx
```diff
@@ -124,6 +124,6 @@ class Avatar extends UIComponent<WithAsProp<AvatarProps>, any> {
 Avatar.create = createShorthandFactory({ Component: Avatar, mappedProp: 'name' })
 
 /**
- * An avatar is a graphic representation of user.
+ * An Avatar is a graphic representation of user.
```
sophieH29: nit: `of a user`

packages/react/src/components/Chat/ChatItem.tsx
```diff
@@ -147,6 +147,6 @@ ChatItem.slotClassNames = {
 }
 
 /**
- * A chat item represents a single event in a chat.
+ * A ChatItem is container for single event in Chat (e.g. message, notification, etc).
```
sophieH29: first impression of `event` as event object 😄 
How about `A ChatItem is a container for a single entity in Chat (...)`

packages/react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -176,7 +176,8 @@ Checkbox.create = createShorthandFactory({
 })
 
 /**
- * A single checkbox within a checkbox group.
+ * A Checkbox allows a user to make choice from the set of two options.
```
sophieH29: `set of two options` sounds a bit confusing to me, but it may only to me :)
I like this definition which I found `... make a choice between one of two possible mutually exclusive options`. Choose what you like more

packages/react/src/components/Dropdown/DropdownSelectedItem.tsx
```diff
@@ -179,8 +179,8 @@ DropdownSelectedItem.create = createShorthandFactory({
 })
 
 /**
- * A a sub-component of multiple-selection Dropdown.
- * Used to display selected item.
+ * A DropdownSelectedItem represents item of 'multiple-selection' Dropdown.
+ * Used to display the search input field.
```
sophieH29: Maybe "to display **IN**  the search input...", as this a selected item which is displayed in the input

packages/react/src/components/Dropdown/DropdownSelectedItem.tsx
```diff
@@ -179,8 +179,8 @@ DropdownSelectedItem.create = createShorthandFactory({
 })
 
 /**
- * A a sub-component of multiple-selection Dropdown.
- * Used to display selected item.
+ * A DropdownSelectedItem represents item of 'multiple-selection' Dropdown.
```
sophieH29: `...represents a selected item of ...`

packages/react/src/components/Grid/Grid.tsx
```diff
@@ -77,6 +77,6 @@ class Grid extends UIComponent<WithAsProp<GridProps>, any> {
 }
 
 /**
- * A grid is used to harmonize negative space in a layout.
+ * A Grid is a layout component that harmonizes negative space, by controling both the row and column alignment.
```
sophieH29: `controlling`

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -428,6 +428,6 @@ class MenuItem extends AutoControlledComponent<WithAsProp<MenuItemProps>, MenuIt
 MenuItem.create = createShorthandFactory({ Component: MenuItem, mappedProp: 'content' })
 
 /**
- * A menu item is an actionable navigation item within a menu.
+ * A MenuItem is an actionable navigation item within a Menu.
```
sophieH29: it's not always `navigation`. `actionable` should be enough

packages/react/src/components/Checkbox/Checkbox.tsx
kuzhelov: Fully agree that `mutually exclusive` is important to add to the description- would it be fine if we'll end up with this variant: `make a choice from two mutually exclusive options`?

packages/react/src/components/List/List.tsx
```diff
@@ -218,6 +218,12 @@ class List extends AutoControlledComponent<WithAsProp<ListProps>, ListState> {
 }
 
 /**
- * A list displays a group of related content.
+ * A List displays a group of related sequential items.
```
sophieH29, line 226: This section is a bit confusing for me in a sense what we offer. Currently, we have just a static list (1st option) and navigable selectable list (2nd option). 
Navigability can be only achieved by specifying `selectable` prop and it will set `listbox` role out of the box, which indicated a selectable list semantic.
To achieve 3rd option, external tweaks are required, for e.g. specifying `selectable` prop but overriding `listbox` to `list` to get rid of 'selectable' semantic, what we don't offer

packages/react/src/components/Checkbox/Checkbox.tsx
sophieH29: yes, perfect!

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -306,6 +306,8 @@ Dialog.slotClassNames = {
 
 /**
  * A Dialog displays important information on top of a page which usually requires user's attention, confirmation or interaction.
+ * Dialogs are purposefully interruptive, so they should be used sparingly.
+ *
```
hughreeling, line 307: Would be great to have some text about how to create modal / non-modal dialogs, or at least hint about using focusTrap...

packages/react/src/components/List/List.tsx
kuzhelov: so, would you suggest to remove `Selectable List` from the set of options, so that we will only target scenarios that are easy (and, what is more important, _safe_) to achieve out-of-the-box?

packages/react/src/components/Dropdown/DropdownSelectedItem.tsx
kuzhelov: this is an awesome catch - completely overlooked this moment. Actually, there is nothing in this component to deal with `search` input field - this component just displays selected options, and there is no need to dig deeper in the presentational aspects. Will make the necessary changes here, thank you!

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -177,7 +177,7 @@ class Embed extends AutoControlledComponent<WithAsProp<EmbedProps>, EmbedState>
 Embed.create = createShorthandFactory({ Component: Embed })
 
 /**
- * An embed component displays a placeholder and matching content based on user's interaction.
+ * An Embed displays content from external websites, like Instagram post or Youtube video.
```
hughreeling: I think you need ' component' here - an Embed component displays content.

packages/react/src/components/Form/Form.tsx
```diff
@@ -100,6 +100,6 @@ class Form extends UIComponent<WithAsProp<FormProps>, any> {
 }
 
 /**
- * A Form displays a set of related user input fields in a structured way.
+ * A Form collects, validates, and submits the user input, in a structured way.
```
hughreeling: We don't do any validation at present.... A Form is used to collect, optionally validate and submit the user input in a structured way?

packages/react/src/components/Label/Label.tsx
```diff
@@ -150,6 +150,6 @@ class Label extends UIComponent<WithAsProp<LabelProps>, any> {
 Label.create = createShorthandFactory({ Component: Label, mappedProp: 'content' })
 
 /**
- * A Label is used to classify content.
+ * A Label allows user to classify a content.
```
hughreeling: without a.  Allows user to classify content.

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -122,7 +122,9 @@ export interface PopupState {
 }
 
 /**
- * A Popup displays additional information on top of a page.
+ * A Popup is a non-modal element that floats around its target on top of a page.
```
hughreeling: Don't like 'floats around' - sounds like swimming.  A Popup is a non-modal element that is displayed above the target and is often used to display additional rich content relating to it's target.

Would be great to also describe concept.  The target is the element upon which an action caused the popup to be displayed....

alinais (approved):

packages/react/src/components/Embed/Embed.tsx
kuzhelov: actually, `component` is consistently omitted for all the descriptions by intent - and it seems to be a common pattern for UI libraries - just several examples (everywhere there it is possible to add `component` in the description):
- https://react.semantic-ui.com/elements/button/
- https://ant.design/components/tag/
- https://material-ui.com/components/slider/

Could you, please, suggest what sort of confusion the lack of this word adds to the description?

packages/react/src/components/Reaction/Reaction.tsx
```diff
@@ -126,6 +126,6 @@ Reaction.slotClassNames = {
 }
 
 /**
- * A reaction is used to indicate user's reaction.
+ * A Reaction indicates user's emotion or perception.
```
hughreeling, line 126: Should identify more clearly why this is different from button?

packages/react/src/components/List/List.tsx
sophieH29: I meant to leave what we support now
- static non-navigable list (when using `<List ...>`)
- selectable navigable list (when using `<List selectable ...>`)

remove 3rd option - "navigable actionable list", for now. But this is something to discuss - having a navigable list (not as selectable) with actionable items

packages/react/src/components/Text/Text.tsx
```diff
@@ -103,8 +103,6 @@ class Text extends UIComponent<WithAsProp<TextProps>, any> {
 Text.create = createShorthandFactory({ Component: Text, mappedProp: 'content' })
 
 /**
- * A Text component formats occurrences of text consistently.
- * @accessibility
- * Text is how people read the content on your website.
+ * A Text consistently styles and formats occurrences of text.
  */
```
hughreeling, line 107: Why would I used Text component, not just React  \<span\> for example?

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -79,8 +79,11 @@ export interface TooltipProps
 }
 
 /**
- * A tooltip that displays information related to an element when the element receives keyboard focus
- * or the mouse hovers over it.
+ * A Tooltip is non-modal element that floats around its target and displays its label.
```
hughreeling: ditto, floating around and swimming

packages/react/src/components/Text/Text.tsx
kuzhelov: because it would inconsistently style and format text for the client? :) could you, please, suggest what you think should be added here?

packages/react/src/components/Video/Video.tsx
```diff
@@ -98,6 +98,6 @@ class Video extends UIComponent<WithAsProp<VideoProps>> {
 Video.create = createShorthandFactory({ Component: Video, mappedProp: 'src' })
 
 /**
- * An video is a graphicical and audio representation of something.
+ * A Video provides ability to embed video content.
```
hughreeling, line 101: And how is it different from Embed?

packages/react/src/components/Popup/Popup.tsx
kuzhelov: > Would be great to also describe concept. The target is the element upon which an action caused the popup to be displayed....

we have a test that checks that description is no longer than 25 words :) will try my best to do this

packages/react/src/components/Video/Video.tsx
kuzhelov: `Embed`: An Embed displays content from external websites, like Instagram post or Youtube video.

Should we somehow emphasize more on the `external` part there? Or, maybe, replace `Youtube video` with `Google Map`?

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -306,7 +303,10 @@ Dialog.slotClassNames = {
...
  * @accessibility
  * Implements [ARIA Dialog (Modal)](https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal) design pattern.
+ * Do use `trapFocus` prop to control focus trapping behavior.
```
jurokapsiar: just a minor comment - since now we have the Best Pracitces section, I think kind of information is more visible in the Do section

packages/react/src/components/List/List.tsx
jurokapsiar: navigable list is coming in the next sprint - maybe sooner if you @sophieH29 will be as fast as usual :)

packages/react/src/components/Reaction/Reaction.tsx
kuzhelov: actually, there is no much difference between these two components. From those things that we may suggest in the description those are
- `Button`, by means of its API, provides set of props that better describe button's semantics
- `Reaction` is more suitable for the chat message cases, and it has far less scenarios for customization supported

However, main reason behind `Reaction` component was to introduce button-like element that would have separate styles in theme, because of its different semantic meaning. If there would be a better mechanism to add this variant of `Button`, it would be chosen then, without need for a dedicated component.

Will introduce sentence that suggests close relation of this component with `Chat.Message`

silviuaavram (approved):

kolaps33 (approved):

packages/react/src/components/Dialog/Dialog.tsx
kuzhelov: moved

sophieH29 (approved):

PR #1605, opened 2019-07-09
chore(types): make Props param in ShorthandValue and ShorthandCollection required

layershifter:
# BREAKING CHANGES

Only if you have used `ShorthandValue` or `ShorthandCollection` types.

### Before
```tsx
const item: ShorthandValue = {}
const items: ShorthandCollection = []
```

### After
```tsx
const item: ShorthandValue<ListItemProps> = {}
const items: ShorthandCollection<ListItemProps> = []
```

***

This PR makers `P` in `ShorthandValue<P>` required, this will allow to get better autocomplete, improve types and generate better prop definitions.

PR #1634, opened 2019-07-15
docs(Props): improve table with props

layershifter:
# BREAKING CHANGES

1. `LoaderPosition` is no longer exported.

```tsx
// BEFORE
const position: LoaderPosition
// AFTER 
const position: LoaderProps['position']
```

# Problem

To generate proper playgrounds for components we need to have a set of values props. It's impossible to get them from strings.

### Allow to use Markdown in props description

This will allow us provide more meaningful descriptions for props.

### Missing default values
Only values that were specified by `@default` tag in JSX were present, for [`Loader` ](https://stardust-ui.github.io/react/components/loader) we have missed `as`, `delay`, `labelPosition` and `size` 💥 

✅ Fixed by update of `parseDefaultValue`:

image

### Unable to manipulate values

Previously we had just string as values (`"gap.smaller" | "gap.small" | "gap.medium" | "gap.large"`) from `react-typescript-docgen`, so we can't to use them for playground generation.

✅ An updated `parseType()` function powered by Babel parses these strings and create proper definition structure:
```json
    {
      "types": [
        {"value": "gap.smaller"},
        {"value": "gap.small"},
        {"value": "gap.medium"},
        {"value": "gap.large"}
      ],
      "name": "gap",
    },
```

#### Why use Babel in `parseType()`?

Only AST parser can provide proper output. Any implementation based on regexps will be fragile on things like generic types.

#### Why not use only Babel?

It will be hard to resolve types that are coming interfaces `CommonProps`, i.e. Babel good at parse, but it can't resolve TS types.

#### Handle edge cases

`react-typescript-docgen` uses `typeToString()` to generate strings with types mentioned above. Sometimes it generates things like this:
image

ℹ️ I tried to configure output of `typeToString()` manually via 3rd param (`TypeFormatFlags`), but it still resolved types.
✅ Fixed by direct lookup in Babel AST, see `getTypeFromBabelTree()`

### Wrong `as` prop definition

It was marked as required and missed description.

image

🆚 

image

### Hard to understand what props can be passed to slot

image

_What can be passed to `clearIndicator`?_

image

Based on #1605 we know which props are used. With parsed types we can obviously understand relationship between components. In this example, `BoxProps` is a link to `Box` component 🚀


build/gulp/tasks/docs.ts
```diff
@@ -93,11 +93,7 @@ const markdownSrc = [
 
 task('build:docs:component-info', () =>
   src(componentsSrc, { since: lastRun('build:docs:component-info') })
-    .pipe(
-      cache(gulpReactDocgen(['DOMAttributes', 'HTMLAttributes']), {
-        name: 'componentInfo',
-      }),
-    )
+    .pipe(cache(gulpReactDocgen(['DOMAttributes', 'HTMLAttributes']), { name: 'componentInfo-1' }))
```
layershifter, line 96: Should be increased otherwise everyone should clear cache manually

packages/react/src/components/Slider/Slider.tsx
```diff
@@ -55,25 +55,16 @@ export interface SliderProps
...
   accessibility?: Accessibility
 
   /** The default value of the slider. */
   defaultValue?: string | number
 
-  /**
-   * A slider can be read-only and unable to change states.
-   * @default false
-   */
+  /** A slider can be read-only and unable to change states. */
```
miroslavstastny, line 64: I understand `@default` is no longer used by docsite, but it is still used by intellisense.
Do we want to remove it or rather test that it is present and correct?

packages/react/src/components/Loader/Loader.tsx
```diff
@@ -15,19 +15,14 @@ import { Accessibility } from '../../lib/accessibility/types'
 import { WithAsProp, ShorthandValue, withSafeTypeForAs } from '../../types'
 import Box, { BoxProps } from '../Box/Box'
 
-export type LoaderPosition = 'above' | 'below' | 'start' | 'end'
```
miroslavstastny, line 18: Removal of this makes it a BREAKING CHANGE. Is it necessary?

docs/src/components/ComponentDoc/ComponentPropsTable/ComponentPropsRow.tsx
```diff
@@ -1,68 +1,76 @@
...
-    }
-    return _.some(tags, ['title', 'slot'])
-  }
+  return <span>{name}</span>
+}
+
+const ComponentPropsRow: React.FunctionComponent<ComponentPropsRowProps> = props => {
+  const { defaultValue, description, name, required, types } = props
+
+  const shorthand = types.find(
```
miroslavstastny:
```suggestion
  const shorthand = types.some(
```

and `!!shorthand` -> `shorthand`

miroslavstastny (commented):
### Shorthand docs and typings not correct for collection shorthand with `kind`

Let's take `ToolbarProps.items` as an example:
- it accepts `kind` attribute with 5 different values, one of them being default
- based on the `kind` it calls shorthand factory for 4 different components.

But documentation just states `ShorthandCollection<ToolbarItemProps>` - it is missing information about `kind` and respective props.
Also the typing is not correct (though `ObjectOf<any>` "saves" that).

Sorry that I have not realised it when reviewing #1605 :-/

### Link from `ShorthandCollection<SubComponent>` goes to SubComponent page

If you follow the link from `ShorthandCollection<ToolbarItemProps>` you will get to ToolbarItemProps page - which correctly describes the props but other than that looks like a dead page (no examples...).
I think I can confuse consumers, it should rather link to parent component page with subcomponent props opened.

build/gulp/plugins/util/getComponentInfo.ts
```diff
@@ -37,111 +55,127 @@ const getComponentInfo = (filepath: string, ignoredParentInterfaces: string[]) =
...
-      }
-    } else {
-      delete info.props[propName]
+        required: propDef.required,
+      })
     }
   })
 
+  // manually insert `as` prop
+  if (info.props.as) {
```
bmdalex, line 143: can we generate something similar for most common events? `onClick`, `onChange`, `onFocus`, etc

build/gulp/plugins/util/getShorthandInfo.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  const object = left.get('object')
+  const property = left.get('property') as NodePath<t.Identifier>
+  const callee = right.get('callee')
+
+  return (
+    object.isIdentifier({ name: componentName }) &&
+    property.isIdentifier({ name: 'create' }) &&
+    callee.isIdentifier({ name: 'createShorthandFactory' })
```
bmdalex, line 33: hope we never rename this function 😄

build/gulp/tasks/docs.ts
bmdalex: don't we need to change `.gitignore` as well?

docs/src/components/ComponentDoc/ComponentPropsTable/ComponentPropsRow.tsx
```diff
@@ -1,68 +1,76 @@
...
-  static propTypes = {
-    defaultValue: PropTypes.string,
-    description: PropTypes.arrayOf(PropTypes.string),
-    name: PropTypes.string,
-    required: PropTypes.bool,
-    tags: PropTypes.array,
-    type: PropTypes.string,
-    value: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)]),
-  }
+const InlineMarkdown = React.lazy(() => import('../InlineMarkdown'))
```
bmdalex, line 9: 👍

docs/src/components/ComponentDoc/ComponentPropsTable/ComponentPropsRow.tsx
```diff
@@ -1,68 +1,76 @@
...
+
+type ComponentPropsRowProps = ComponentProp
 
-  state: any = {}
+const ComponentPropValue: React.FunctionComponent<ComponentPropType> = props => {
+  const { name, parameters } = props
 
-  toggleEnums = () => this.setState({ showEnums: !this.state.showEnums })
+  if (!name) return <span>enum</span>
```
bmdalex: what's happening here? :)

docs/src/components/ComponentDoc/ComponentPropsTable/ComponentPropsRow.tsx
```diff
@@ -1,68 +1,76 @@
...
-    return _.some(tags, ['title', 'slot'])
-  }
+  return <span>{name}</span>
+}
+
+const ComponentPropsRow: React.FunctionComponent<ComponentPropsRowProps> = props => {
+  const { defaultValue, description, name, required, types } = props
+
+  const shorthand = types.find(
+    type => type.name === 'ShorthandValue' || type.name === 'ShorthandCollection',
```
bmdalex, line 44: how can we protect against renaming these types?

alinais (approved):

packages/react/src/components/Loader/Loader.tsx
bmdalex: Isn't @layershifter just inlining this?

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -12,7 +12,7 @@ import getElementType from '../../getElementType'
 import callable from '../../callable'
 
 /** AutoFocusZone is used to focus inner element on mount. */
-export class AutoFocusZone extends React.Component<AutoFocusZoneProps> {
+export default class AutoFocusZone extends React.Component<AutoFocusZoneProps> {
```
bmdalex, line 15: is this necessary or just an improvement?

bmdalex (approved):
looks good to me, few comments; awesome job 👍

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
layershifter: It's necessary to make working `defaultProps` parser because all our components has a default export. It's even covered by `isConformant()` test.

packages/react/src/components/Loader/Loader.tsx
layershifter: Technically it's breaking change, will mention this in `CHANGELOG` and in a PR header.
It's necessary to avoid manual type definition for autogenerated playground.

packages/react/src/components/Slider/Slider.tsx
layershifter: It's hard to maintain things like these because they don't aligned with code. I think that the best way is to remove them and rely on `defaultProps`.

build/gulp/plugins/util/getComponentInfo.ts
layershifter: image

We have already these, but they don't look good 👎

I am going to improve this thing in a separate PR:
- `ComponentEventHandler` will contain event type, i.e. `ComponentEventHandler<React.MouseEvent, ButtonProps>`
- provide more details and describe it as a function: `(e: React.MouseEvent, data: ButtonProps) => void`

build/gulp/tasks/docs.ts
layershifter: Nope, this cache is stored in your TMP directory

docs/src/components/ComponentDoc/ComponentPropsTable/ComponentPropsRow.tsx
layershifter: Improved this part in https://github.com/stardust-ui/react/pull/1634/commits/0bd34495c8acfb26c47b43865fa2dd4f97e96134 👍

docs/src/components/ComponentDoc/ComponentPropsTable/ComponentPropsRow.tsx
layershifter: I don't see any way because these TS types don't exist in runtime 😭

layershifter:
@miroslavstastny great catches, fixed them in https://github.com/stardust-ui/react/pull/1634/commits/99561150ab9f0c91bf89bfcae71ffd4e6e38b65e and https://github.com/stardust-ui/react/pull/1634/commits/83c157717c7cee3fa685bc414fcd6fd305c46052 👍

packages/react/src/components/Loader/Loader.tsx
bmdalex: oh, we were exporting this type 🤦‍♂  great catch @miroslavstastny

docs/src/components/ComponentDoc/ComponentPropsTable/ComponentPropsRow.tsx
bmdalex: true 😢

miroslavstastny (approved):

PR #1635, opened 2019-07-15
feat(Tooltip): Add label behavior for Tooltip

sophieH29:
This PR creates `tooltipAsLabelBehavior` accessibility behavior, so when used, it will apply `aria-labelledby` attribute to trigger button in case it is needed to label visual-only elements. `aria-describedby` doesn't work well for JAWS


CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Split action handlers with "OR" condition in accessibility behaviors @sophieH29 ([#1622](https://github.com/stardust-ui/react/pull/1622))
 - Add `headerAction` slot to the `Dialog` component @mnajdova ([#1617](https://github.com/stardust-ui/react/pull/1617))
 - Add `Slider` component @Bugaa92 ([#1559](https://github.com/stardust-ui/react/pull/1559))
+- Add label accessibility behavior for `Tooltip` ([#1635](https://github.com/stardust-ui/react/pull/1635))
```
mnajdova:
```suggestion
- Add `tooltipAsLabelBehavior` accessibility behavior for `Tooltip` ([#1635](https://github.com/stardust-ui/react/pull/1635))
```

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -81,6 +83,8 @@ export interface TooltipProps
 /**
  * A tooltip that displays information related to an element when the element receives keyboard focus
  * or the mouse hovers over it.
+ * @accessibility
```
mnajdova: Should we add example for this one on the docs?

sophieH29: I was thinking about it. From one side this behavior changes one attribute and it is listed in the `accessibility` section on cases when to use it. And at other side, it is a variance of a tooltip.
If you think better to add it, I'll do that

bmdalex (approved):

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -81,6 +83,8 @@ export interface TooltipProps
...
+ * If adding tooltip to icon-only button or to another visual-only element without any text, title or label, apply `tooltipAsLabelBehavior` to `accessibility` prop to correctly label the element for all screen readers.
```
jurokapsiar: please move this information to the Best Practices section of the tooltip site - it is more visible there.

packages/react/src/components/Tooltip/Tooltip.tsx
sophieH29: @mnajdova I added a tooltip best practices section and mentioned it there. I think should be enough for now. Let's observe, if discoverability is bad, will add a separate example in future

docs/src/examples/components/Tooltip/BestPractices/TooltipBestPractices.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react'
+
+import ComponentBestPractices from 'docs/src/components/ComponentBestPractices'
+
+const doList = [
+  'Use `tooltipAsLabelBehavior` if adding tooltip to icon-only button or to another visual-only element without any text, title or label.',
```
mnajdova, line 6: 👍

mnajdova (approved):

packages/react/src/lib/accessibility/Behaviors/Tooltip/tooltipAsLabelBehavior.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Adds attribute 'aria-labelledby' based on the property 'aria-labelledby' to 'trigger' slot.
+ * Triggers 'close' action with 'Escape' on 'trigger'.
+ */
+const tooltipAsLabelBehavior: Accessibility<TooltipBehaviorProps> = props => {
+  const behaviorData = tooltipBehavior(props)
+  const defaultAriaLabeledBy = getDefaultAriaLabelledBy(props)
+
+  behaviorData.attributes = {
+    trigger: {
+      'aria-labelledby': defaultAriaLabeledBy || props['aria-labelledby'],
```
kolaps33, line 17: do we want to add props['aria-labelledby'] ?, if we do already this check(verification) in "getDefaultAriaLabelledBy" function? :)

PR #1636, opened 2019-07-15
feat(docs): add usage example with Tooltip for the actionable elements

mnajdova:
This PR is adding usage example for the `Button`, `MenuItem` and `ToolbarItem` components for using `Tooltip`

Open questions:
Should we add tooltip for the disabled elements? - there is an example of this in the Toolbar example
Should we add tooltip examples of any other components, like Slider, or Popup and Dialog - as they have usually Button as a trigger


docs/src/examples/components/Button/Usage/index.tsx
```diff
@@ -9,6 +10,16 @@ const Usage = () => (
       description='A button used in cards is a "tinted" version of a default button.'
       examplePath="components/Button/Usage/ButtonUsageExample"
     />
+    <ComponentExample
+      title="With tooltip"
+      description={
+        <>
+          {'The button as actionable element should be rendered with '}
+          <Link to="tooltip">tooltip</Link>
```
layershifter:
```suggestion
          <Link to="/components/tooltip">tooltip</Link>
```

docs/src/examples/components/Toolbar/Usage/ToolbarExampleWithTooltip.shorthand.tsx
```diff
@@ -0,0 +1,282 @@
+import * as React from 'react'
```
sophieH29, line 1: This example is super huge. How about to make it with few items (2-3) just to showcase the usage of a tooltip. Because as a client I can get lost to find the logic where the tooltip is rendered

sophieH29:
> Should we add tooltip examples of any other components, like Slider, or Popup and Dialog - as they have usually Button as a trigger

I think `Popup` is good to have. I know there is one for the toolbar with Popup. But would be a plus to see a separate example of combining that nested logic with Popup's trigger and Tooltip's trigger

docs/src/examples/components/Toolbar/Usage/ToolbarExampleWithTooltip.shorthand.tsx
mnajdova: I simplified this a bit, but still I want to have example for each tooltip item: toggle, divider, item with popup, with menu, radio group etc. These are the options we have now:
image

Let me know if you want some of there to be removed, and will update it

bmdalex: I agree with @sophieH29 maybe this should be a prototype... 
@mnajdova , maybe we can shorten the example and make it point to the prototype (with a link) for more details since we plan to have a public prototypes section anyway.

docs/src/examples/components/Button/Usage/index.tsx
```diff
@@ -9,6 +10,16 @@ const Usage = () => (
       description='A button used in cards is a "tinted" version of a default button.'
       examplePath="components/Button/Usage/ButtonUsageExample"
     />
+    <ComponentExample
+      title="With tooltip"
+      description={
+        <>
+          {'The button as actionable element should be rendered with '}
```
kuzhelov: sorry, cannot understand what is the main idea this description is going to convey to the client?

docs/src/examples/components/Menu/Usages/index.tsx
```diff
@@ -24,6 +25,16 @@ const Usages = () => (
       description="When Submenu in MenuItem is controlled, then its 'menuOpen' prop value could be changed either by parent component, or by user actions (e.g. key press) - thus it is necessary to handle 'onMenuOpenChange' event."
       examplePath="components/Menu/Usages/MenuExampleWithSubmenuControlled"
     />
+    <ComponentExample
+      title="Menu with items containing tooltip"
+      description={
+        <>
+          {'The items inside the Menu as actionable element should be rendered with '}
```
kuzhelov: same here (as it was for https://github.com/stardust-ui/react/pull/1636/files#r303463181) - are we advising client to always use `Tooltip` with any actionable element, or it is some different idea intended to be provided?

docs/src/examples/components/Menu/Usages/index.tsx
```diff
@@ -24,6 +25,16 @@ const Usages = () => (
       description="When Submenu in MenuItem is controlled, then its 'menuOpen' prop value could be changed either by parent component, or by user actions (e.g. key press) - thus it is necessary to handle 'onMenuOpenChange' event."
       examplePath="components/Menu/Usages/MenuExampleWithSubmenuControlled"
     />
+    <ComponentExample
+      title="Menu with items containing tooltip"
```
kuzhelov: nit: for the sake of consistency, lets choose either shorter `With tooltip` form (which is used for `Button` - i.e. format where component's name is omitted), or use the long form with component name. Personally, think that `With tooltips` should be enough to describe the example

docs/src/examples/components/Menu/Usages/index.tsx
mnajdova: It is exactly that:

> The items inside the Menu as actionable element should be rendered with tooltip.


Can you suggest a better description?

mnajdova:
@sophieH29 I tried adding example for the Popup with tooltip, but it seems that accessibility is broken in this case:
image

With escape, the tooltip is only closed, we cannot close the Popup with escape... Not sure if we should add this example until we have working solution. Here is the code I tried:
https://codesandbox.io/s/stardust-ui-example-7ho97

docs/src/examples/components/Toolbar/Usage/ToolbarExampleWithTooltip.shorthand.tsx
kuzhelov: agree - in its current form this example is rather a prototype. 

Would suggest to avoid `group` complications here, just place three or four items in the `Toolbar` and add tooltip to them - this should perfectly make a `Toolbar example with tooltip` :)

> but still I want to have example for each tooltip item: toggle, divider, item with popup, with menu, radio group etc

As it is the same logic used for any kind of tooltip item, I think it would be enough to provide example for one kind of item only, and just say in the comments that all other kind of items could be treated the same way.

mnajdova:
As oppose to the prev comment for the issues of the  `Popup` with `tooltip`, the toolbar item containing popup and tooltip on escape is closing only the popup, the tooltip can never be closed with this action. Should I remove this example for now and we can address these issues in a separate PR and add the examples together with the fix? @miroslavstastny I know you were investigating this, maybe you have some different proposal. @sophieH29 FYI

docs/src/examples/components/Toolbar/Usage/ToolbarExampleWithTooltip.shorthand.tsx
mnajdova: Thanks for the suggestions guys, I refactored the example, it's much shorter now, please take a look!

docs/src/examples/components/Menu/Usages/index.tsx
kuzhelov: may suggest to embrace `as` in commas in both places, to proper indicate `because` semantics. This is similar to the following example, where `as` also makes semantic difference:

> I drove the car home as it was snowing. (i.e. `I was the car home as a very careful driver.`)
> I drove the car home, as it was snowing. (i.e. `I drove the car home because it was snowing.`)

docs/src/examples/components/Toolbar/Usage/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
+
+const Usage = () => (
+  <ExampleSection title="Usage">
+    <ComponentExample
+      title="With tooltips"
+      description={
+        <>
+          {'The items inside the Toolbar as actionable element should be rendered with '}
```
kuzhelov: nit: `as actionable elements`?

docs/src/examples/components/Menu/Usages/index.tsx
```diff
@@ -24,6 +25,16 @@ const Usages = () => (
       description="When Submenu in MenuItem is controlled, then its 'menuOpen' prop value could be changed either by parent component, or by user actions (e.g. key press) - thus it is necessary to handle 'onMenuOpenChange' event."
       examplePath="components/Menu/Usages/MenuExampleWithSubmenuControlled"
     />
+    <ComponentExample
+      title="With tooltips"
+      description={
+        <>
+          {'The items inside the Menu as actionable element should be rendered with '}
```
kuzhelov: nit: as actionable elements (plural)?

kuzhelov (approved):

bmdalex (approved):

miroslavstastny (approved):
Let's solve `Popup` and `Tooltip` attached to the same trigger in a separate issue, discussing desired behavior first. As you described, accessibility is broken and there are some other problems I have found so far:
- `Popup` and `Tooltip` are not spreading props to trigger
- both have `z-index: 1000`, my gut feeling is Tooltip should have higher z-index than Popup

mnajdova:
Thanks everyone, will add steps for the new screens and will merge it

sophieH29 (approved):

PR #1641, opened 2019-07-15
docs(slider): best practices and usage in form

bmdalex:
### Description

This PR contains documentation improvements for the `Slider` component:
- best practices section;
- new example on how to use the `Slider` in a `Form`

### DIFF
--- a/docs/src/examples/components/Slider/Usage/SliderExampleForm.shorthand.tsx
+++ b/docs/src/examples/components/Slider/Usage/SliderExampleForm.shorthand.tsx
@@ -1,0 +1,43 @@
+import * as React from 'react'
+import { Slider, SliderProps, Flex, Form, Label } from '@stardust-ui/react'
+
+const CustomSlider: React.FunctionComponent<SliderProps> = props => {
+  const [value, setValue] = React.useState<string>('35')
+  return (
+    <Flex inline hAlign="center" vAlign="center" gap="gap.smaller">
+      <Slider min="18" value={value} onChange={(e, data) => setValue(data.value)} {...props} />
+      <Label content={value} />
+    </Flex>
+  )
+}
+
+const SliderExampleFormShorthand = () => (
+  <Form
+    onSubmit={() => alert('Form submitted')}
+    fields={[
+      {
+        label: 'First name',
+        name: 'firstName',
+        id: 'first-name-shorthand',
+        key: 'first-name',
+        required: true,
+      },
+      {
+        label: 'Last name',
+        name: 'lastName',
+        id: 'last-name-shorthand',
+        key: 'last-name',
+        required: true,
+      },
+      {
+        label: 'Age',
+        control: { as: CustomSlider },
+        id: 'age-shorthand',
+        key: 'age',
+        required: true,
+      },
+    ]}
+  />
+)
+
+export default SliderExampleFormShorthand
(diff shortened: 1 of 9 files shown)


kolaps33 (approved):
During verification of accessibility found issues with JAWS. 

Bugs to JAWS reported:

Slider - JAWS narrates slider value twice when using PageUp / PageDown
https://github.com/FreedomScientific/VFO-standards-support/issues/220

Slider - JAWS narrates current and new value in vertical slider
https://github.com/FreedomScientific/VFO-standards-support/issues/219

docs/src/examples/components/Slider/Usage/SliderExampleForm.shorthand.tsx
```diff
@@ -0,0 +1,43 @@
...
+      {
+        label: 'Last name',
+        name: 'lastName',
+        id: 'last-name-shorthand',
+        key: 'last-name',
+        required: true,
+      },
+      {
+        label: 'Age',
+        control: { as: CustomSlider },
```
jurokapsiar, line 34: I mentioned this offline, but why can't we just use:
```
control: <Slider />
```
It's true that doing that will add the `id` to the root of the Slider element. But doing the same with `Input` will add the `id` to the `input` slot. Why are these two different?

silviuaavram (approved):

docs/src/examples/components/Slider/Usage/SliderExampleForm.shorthand.tsx
bmdalex: > but why can't we just use:
> `control: <Slider />`

That's because `control: <Slider  />,` places the slider inside the control slot rather than replacing it (see https://github.com/stardust-ui/react/pull/1513)

> It's true that doing that will add the id to the root of the Slider element. But doing the same with Input will add the id to the input slot. Why are these two different?

It's working for `Input` just because we use Input as default control in the form: https://github.com/stardust-ui/react/blob/424ba3c8b20835921b79cf1591115b98623c7fef/packages/react/src/components/Form/FormField.tsx#L74

So if we want to replace the default `Input` control with a new one (`Slider` in our case), the only way to achieve this currently is:

`control: { as: Slider }`

@jurokapsiar I agree this might not be the most intuitive API. I talked to @layershifter and there might be ways to improve the API and use an element instead of shorthand for `control` slot. That means your suggestion:

`control: <Slider />`

would work. However, this solution diverges from the current Stardust patterns where we treat slots as shorthands. Adding @mnajdova who originally implemented the `Form` component.

Do you guys think we should discuss this?

mnajdova: In my opinion the API: `control: <Slider />` is the most intuitive. Usually when we want to support different components in the same slot we are introducing the kind prop - like in the Menu or Toolbar, but here, as we can have lots of different controls - custom  or Stardust, I would say that even that option may not be quite helpful. A bit afraid with the changes of the factories, as now every custom component should use the render callback, but I guess the prev has it caveats as well.

layershifter: https://react.semantic-ui.com/collections/form/#shorthand-subcomponent-control

Please check existing API of SUIR. `control` is not shorthand prop there and it allows to avoid complications

bmdalex: then I guess we should discuss it, created #1648

Issue #1531, opened 2019-06-21
`FAQ` entry for usages of `Box`, `Segment`, `Flex`, etc


bmdalex:
### Problem description
Users have been complaining about not being able to understand when to use different components that seem more abstract at first glance, such as `Box`, `Segment`, `Flex`, `Grid`, `Layout`

### Proposed solution
- better docs
- FAQ (https://stardust-ui.github.io/react/faq) entry with a compare and contrast section for these components.

Issue #1530, opened 2019-06-21
`Segment` doc examples


bmdalex:
### Problem description
Users have been complaining that it's not clear:
- what a `Segment` does
- what are some real life scenarios of how it can be used

### Proposed solution
Add more doc examples for `Segment` (inspiration: https://react.semantic-ui.com/elements/segment/)

jurokapsiar:
docs are sufficient now

PR #1646, opened 2019-07-16
feat(docs): add FAQ entry for Flex, Grid, Layout, Box, Segment

lucivpav:
This PR aims to clarify the differences between usages of the following components: `Flex`, `Grid`, `Layout`, `Box`, `Segment`.

This PR resolves Issue #1531.


lucivpav:
Example(s) should be given on Segment component.

lucivpav:
> Example(s) should be given on Segment component.

Relevant issue #1530

docs/src/views/FAQ.tsx
```diff
@@ -111,6 +111,26 @@ export default () => (
...
+        <p>
+          Flex, Grid and Layout components handle layout aspects. The Flex component is for laying
+          out items in one direction, while the Grid component is made for two dimensional layouts.
+          Visit the {link('Layout guide', '/layout')} page for a detailed comparison between Flex
+          and Grid. The Layout component is now deprecated. Its purpose was arrangement of the
+          content of a component. Flex or Grid component should be used instead of the Layout
+          component.
+          <br />
+          Box is an abstract component that by default renders a `div` without any styles. Box
+          should be used carefully, in most cases {link("you don't need it ", '/components/box')}.
```
bmdalex: I think we need to improve on this one:

`Box should be used carefully, in most cases you don't need it.`

We also need to tell users when they might need it. Not sure how to formulate this. Can you guys suggest something?

@layershifter @kuzhelov @mnajdova @miroslavstastny

layershifter: > `Box` is an abstract component that by default renders a `div` by default and can process defined styles. To keep your styles inside theme consider to use `createComponent()` instead of `Box`.

What about this?

kuzhelov: maybe we could even simplify it more, by saying that

> Box is a utility component that is often used by Stardust to implement higher-level components. By default, it renders styled `div` element.  

> The cases when client might want to use it are very exceptional, and all are about applying custom styles to rendered `div` element. However, more robust approach that won't break theming consistency is to instead create a custom component and define related styles as part of theme (link to Integrate Custom Components section).

docs/src/views/FAQ.tsx
```diff
@@ -111,6 +111,26 @@ export default () => (
       }
     />
 
+    <Question content="What's the difference between Flex, Grid, Layout, Box and Segment components?" />
+    <Answer
+      content={
+        <p>
+          Flex, Grid and Layout components handle layout aspects. The Flex component is for laying
```
kuzhelov: nit: as we are doing throughout the docs, lets wrap component names in `code('Flex')`, so that it would lead to proper formatting (just as an example, here is the first paragraph from the `Layout` section): 

image

docs/src/views/FAQ.tsx
```diff
@@ -111,6 +111,42 @@ export default () => (
...
+    </Question>
+    <Answer
+      content={
+        <div>
+          <p>
+            {code('Flex')}, {code('Grid')} and {code('Layout')} components handle layout aspects.
+            The {code('Flex')} component is for laying out items in one direction, while the{' '}
+            {code('Grid')} component is made for two dimensional layouts. Visit the{' '}
+            {link('Layout guide', '/layout')} page for a detailed comparison between {code('Flex')}
+            and {code('Grid')}. The {code('Layout')} component is now deprecated. Its purpose was
```
kuzhelov: nit: may suggest to put comments about `Layout` in a separate paragraph, it will make the content easier to read

CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix broken fragments on CodeSandbox @lucivpav ([#1655](https://github.com/stardust-ui/react/pull/1655))
 - Improve a table with component props in docs @layershifter ([#1634](https://github.com/stardust-ui/react/pull/1634))
 - Improve rendering performance for docs examples @Bugaa92 ([#1654](https://github.com/stardust-ui/react/pull/1654))
+- Add FAQ entry for Flex, Grid, Layout, Box, Segment @lucivpav ([#1646](https://github.com/stardust-ui/react/pull/1646))
```
kuzhelov: nit: please, embrace component names in back-ticks, for the sake of formatting consistency in the changelog

kuzhelov (approved):
looks good to me 👍 
 @mnajdova, @layershifter, please, share your thoughts on this. Thank you!

layershifter:
Please add the question in the title

image

docs/src/views/FAQ.tsx
```diff
@@ -111,6 +111,45 @@ export default () => (
...
+          <p>
+            {code('Box')} is a utility component that is often used by Stardust to implement
+            higher-level components. By default, it renders styled {code('div')} element.
+          </p>
+          <p>
+            The cases when client might want to use it are very exceptional, and all are about
+            applying custom styles to rendered {code('div')} element. However, more robust approach
+            that won't break theming consistency is to instead create a custom component and define
+            related styles as part of theme{' '}
+            {link('Integrate Custom Components', '/integrate-custom-components')}.
```
mnajdova: The ending of the paragraph is not good. Let's end it with something like: 

> Visit the `Integrate Custom Components` page to see how this can be done

Or something similar, at least to give transition of what the link is about.

lucivpav:
> Please add the question in the title
> 
> image

Fixed.

PR #1647, opened 2019-07-16
feat(FocusZone): add bidirectional navigation following DOM order

sophieH29:
## tl;dr
Usually, when screen reader users navigate to menus or lists, they want to use Up/Down arrow keys as the semantic of these components stands for. In real-world, we have a lot of examples, where we have a horizontal list/menu or a grid with items. Keyboard users (not blind people) intentionally will want to use Left/Right arrow keys to navigate within these components.

To achieve consistency, Right/Down arrow keys should navigate to the next focusable item in the component. Left/Up arrow keys - to the previous focusable item. Note: in RTL mode, only left and arrow keys are switched.

Also, it is useful for Grid-like components (for e.g. emoji pickers). Screen readers are not aware of the grid so they would use Up/Down arrow keys to navigate through emoji as in the list. Keyboard users most probably will use Left/Right arrow keys to navigate in the grid, the focus will move horizontally row by row.

## changes
- Add one more direction to `FocusZone`'s enum - `bidirectionalDomOrder`
- Substitute `bidirectional` with `bidirectionalDomOrder` in `menuAsToolbarBehavior` and `tablistBheavior`, as `bidirectional` is more calculation heavy by calculating position of the elments on a page
- add `gridHorizontalBehavior` with using FocusZone with `bidirectionalDomOrder`, to allow navigation described above.


jurokapsiar (approved):

bmdalex (approved):

docs/src/examples/components/Grid/BestPractices/GridBestPractices.tsx
```diff
@@ -3,7 +3,9 @@ import * as React from 'react'
 import ComponentBestPractices from 'docs/src/components/ComponentBestPractices'
 
 const doList = [
-  'Use Grid behavior for bidirectional keyboard navigation. Use appropriate ARIA role for the grid and actionable components inside of it.',
+  'Use `gridBehavior` for bidirectional keyboard navigation with 4 arrow keys.',
```
hughreeling: I find wording 'with 4 arrow keys' somehow confusing.  Can there be more?  Why specify 4?  I think it would read better as just 'with arrow keys'...

PR #1652, opened 2019-07-16
feat(Text): add largest variation of font size

alinais:
image


alinais:
@codepretty please let me know if 32px or 36px is more suitable as `largest` size for Text (or other value).

alinais:
The requested change is no longer needed. Closing this PR.

PR #1654, opened 2019-07-16
docs(perf): improve rendering perf for examples

bmdalex:
### Description:

This PR is focused on improving docs usage experience.
Performance is gained by using placeholders for invisible components.
Each page is loading faster now 🚀

Inspired by @layershifter 's [PR in SUIR](https://github.com/Semantic-Org/Semantic-UI-React/pull/3549) 🙇 

### Before
https://imgur.com/a/t6iKNNW

### After
https://imgur.com/a/cjN5oNM


docs/src/components/DelayedRenderer.tsx
```diff
@@ -0,0 +1,37 @@
...
+  smaller: 5,
+  small: 15,
+  medium: 25,
+  large: 40,
+  larger: 65,
+  largest: 100,
+}
+
+const createPlaceholder = (placeholder: DelayedRendererProps['placeholder']): JSX.Element => {
+  const size = typeof placeholder === 'string' ? placeholder : 'medium'
```
miroslavstastny: nit: I would move this bellow line 23 to do it only when necessary (no performance gain, just more understandable code)

miroslavstastny (approved):
🎉

docs/src/components/DelayedRenderer.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react'
+import { Loader, Segment, LoaderProps } from '@stardust-ui/react'
+
+type DelayedRendererProps = Partial<{
+  children: JSX.Element
+  placeholder: JSX.Element | LoaderProps['size']
+  visible: boolean
+}>
```
layershifter: Do we need really need this complication? I don't see any usage of `placeholder` as element.

I tried to rewrite this file and it looks more simple:

```tsx
import { Loader, Segment, LoaderProps } from '@stardust-ui/react'

type DelayedRendererProps = {
  placeholder: LoaderProps['size']
  visible: boolean
}

const paddings: Record<LoaderProps['size'], string> = {
  smallest: '0px',
  smaller: '5px',
  small: '15px',
  medium: '25px',
  large: '40px',
  larger: '65px',
  largest: '100px',
}

const DelayedRenderer: React.FunctionComponent<DelayedRendererProps> = props => {
  const { children, placeholder, visible } = props

  return visible ? <>{children}</> : (
    <Segment variables={{ padding: paddings[placeholder] }}>
      <Loader size={placeholder}/>
    </Segment>
  )
}

export default DelayedRenderer
```

bmdalex: We don't need it for sure, agreed that if users of this component need custom placeholder we can implement something on demand and that probably won't happen.

Then I'll probably change the name of `placeholder` prop to `size` or `placeholderSize`. Which one are you voting for @layershifter ?

Also,  is `DelayedRenderer` a good enough name?

layershifter: > Then I'll probably change the name of placeholder prop to size 

👍 

> Also, is DelayedRenderer a good enough name?

Not, sure because there are no any `delay` in that component

layershifter: What about `ExamplePlaceholder`?

docs/src/components/DelayedRenderer.tsx
bmdalex: simplified interface:
- `placeholder` cannot be ReactElement anymore;
- `placeholder` was renamed to `size`

docs/src/components/DelayedRenderer.tsx
bmdalex: applied suggestions

Issue #1519, opened 2019-06-19
Docs: examples that contain React.Fragments are broken on CodeSandbox


layershifter:
## Bug Report

`<>` is working in newly created CodeSandboxes: https://codesandbox.io/s/infallible-leftpad-gczx1
However, `<>`  is broken in ours 😭

It seems that we're using outdated template (`sandbox.config.json`):

```json
{"template":"create-react-app"}
```

**It works properly for TS examples.**

### Steps
1. Go to example: https://stardust-ui.github.io/react/components/checkbox#types-disabled
2. Click to "CodeSandbox" button

### Expected Result

Working example.

### Actual Result

![image](https://user-images.githubusercontent.com/14183168/59775202-0eff7480-92b1-11e9-8c80-8febd87abc29.png)

### Version
latest

lucivpav:
Note: "react-scripts": "latest" should be added to `package.json`, this resolves the error

PR #1655, opened 2019-07-16
fix(docs): broken fragments on CodeSandbox

lucivpav:
This PR resolves issue #1519. The code examples with `<>...</>` fragments resulted in an error on CodeSandbox if JavaScript language was chosen.

### DIFF
--- a/docs/src/components/Playground/renderConfig.ts
+++ b/docs/src/components/Playground/renderConfig.ts
@@ -24,7 +24,6 @@ export const imports = {
   react: React,
   'react-dom': ReactDOM,
   'react-fela': ReactFela,
-  'react-scripts': 'latest',
 }
 
 export const importResolver = importName => imports[importName]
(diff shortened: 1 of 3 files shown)


docs/src/components/Playground/renderConfig.ts
```diff
@@ -24,6 +24,7 @@ export const imports = {
   react: React,
   'react-dom': ReactDOM,
   'react-fela': ReactFela,
+  'react-scripts': 'latest',
```
layershifter: I suggest to place this thing in `createPackageJson.ts` instead:

```jsx
const dependencies = {
  ..._.mapValues(imports, () => 'latest'),
  // required to enable all features due old templates in https://github.com/codesandbox/codesandbox-importers
  // https://github.com/stardust-ui/react/issues/1519
  'react-scripts': 'latest',
}
```

If we will place this here it can be resolved to string in our examples 💥

layershifter (approved):

PR #1656, opened 2019-07-16
fix(Checkbox): refined styles to match redline and add dark and contrast theme

bcalvery:
Before:
image
image
image

After:
image
image
image


packages/react/src/themes/teams/components/Checkbox/checkboxStyles.ts
```diff
@@ -1,13 +1,104 @@
...
 
 const checkboxStyles: ComponentSlotStylesInput<CheckboxProps & CheckboxState, CheckboxVariables> = {
   root: ({ props: p, variables: v, theme: t }): ICSSInJSStyle => ({
+    color: v.checkboxTextColor,
+    padding: v.rootPadding,
+
     ...getBorderFocusStyles({
       siteVariables: t.siteVariables,
       isFromKeyboard: p.isFromKeyboard,
+      borderRadius: '3px', // shouldn't default be correct?
```
bcalvery: I had to modify the "universal" getBorderFocusStyles to make the radius of the corner match the redline. It seems like that should not be something that is necessary, is global getBorderFocusStyles correct?

CHANGELOG.md
```diff
@@ -39,6 +39,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Refine component descriptions @kuzhelov ([#1629](https://github.com/stardust-ui/react/pull/1629))
 - Add best practices and form usage example for `Slider` component @Bugaa92 ([#1641](https://github.com/stardust-ui/react/pull/1641))
 - Add examples with `Tooltip` for the actionable components @mnajdova ([#1636](https://github.com/stardust-ui/react/pull/1636))
+- Update `Checkbox` theme files to match [redlines] @bcalvery ([#1656](https://github.com/stardust-ui/react/pull/1656))
```
codepretty: Teams theme

packages/react/src/themes/teams-dark/components/Checkbox/checkboxVariables.ts
```diff
@@ -0,0 +1,5 @@
+import { CheckboxVariables } from '../../../teams/components/Checkbox/checkboxVariables'
+
+export default (siteVars: any): Partial<CheckboxVariables> => {
+  return {}
```
codepretty: Is this file needed since it's empty?

bcalvery: If a tree falls in the forest, and no one is there to hear it, does it make a sound?

packages/react/src/themes/teams/components/Checkbox/checkboxVariables.ts
```diff
@@ -1,22 +1,75 @@
...
+  checkboxBorderColorHover: siteVars.colorScheme.default.foreground,
+  checkboxCheckedBackgroundHover: siteVars.colors.brand[600],
+
+  toggleBorderRadius: pxToRem(999),
+  toggleIndicatorColor: siteVars.colors.brand[600],
+  toggleMargin: `${pxToRem(20)}, 0`,
+  togglePadding: `${checkboxPadding} ${toggleMovementDistance} ${checkboxPadding} ${checkboxPadding}`,
+
+  checkboxCheckedTextColor: siteVars.colorScheme.default.foreground,
+  checkboxCheckedBackground: siteVars.colors.brand[600],
```
codepretty: I *think* `checkboxCheckedBackground` should be `siteVars.colorScheme.brand.backgroundActive1` 
image

This will also get rid of the need for defining `checkboxCheckedBackground` in the `teams-high-contrast` theme file.

packages/react/src/themes/teams/components/Checkbox/checkboxVariables.ts
```diff
@@ -1,22 +1,75 @@
...
+  checkboxBorderWidth: pxToRem(1),
+  checkboxIndicatorColor: 'transparent',
+  checkboxGap: pxToRem(12),
+  checkboxMargin: `${pxToRem(20)}, 0`,
+  checkboxPadding,
+  rootPadding: '3px 5px',
+
+  checkboxTextColorHover: siteVars.colorScheme.default.foreground,
+  checkboxBorderColorHover: siteVars.colorScheme.default.foreground,
+  checkboxCheckedBackgroundHover: siteVars.colors.brand[600],
```
codepretty: I think `checkboxCheckedBackgroundHover` should be using background hover design token from the brand color scheme.

image

packages/react/src/themes/teams/components/Checkbox/checkboxVariables.ts
```diff
@@ -1,22 +1,75 @@
...
+
+  toggleBorderRadius: pxToRem(999),
+  toggleIndicatorColor: siteVars.colors.brand[600],
+  toggleMargin: `${pxToRem(20)}, 0`,
+  togglePadding: `${checkboxPadding} ${toggleMovementDistance} ${checkboxPadding} ${checkboxPadding}`,
+
+  checkboxCheckedTextColor: siteVars.colorScheme.default.foreground,
+  checkboxCheckedBackground: siteVars.colors.brand[600],
+  checkboxCheckedBorderColor: 'transparent',
+  checkboxCheckedIndicatorColor: siteVars.colors.white,
```
codepretty: `checkboxCheckedIndicatorColor` should use the Foreground 4 design token from the brand color scheme

image

and you can remove the variable from the high contrast variables

packages/react/src/themes/teams/components/Checkbox/checkboxVariables.ts
```diff
@@ -1,22 +1,75 @@
...
+  toggleCheckedPadding: `${checkboxPadding} ${checkboxPadding} ${checkboxPadding} ${toggleMovementDistance}`,
+
+  disabledColor: siteVars.colorScheme.default.foregroundDisabled1,
+  disabledCheckboxBackground: siteVars.colorScheme.default.background,
+  disabledCheckboxBackgroundChecked: siteVars.colorScheme.default.backgroundDisabled,
+  disabledCheckboxBorderColor: siteVars.colorScheme.default.foregroundDisabled1,
+  disabledCheckboxCheckedIndicatorColor: siteVars.colorScheme.default.foregroundDisabled1,
```
codepretty: `disabledCheckboxCheckedIndicatorColor` should be `foregroundDisabled` instead of `foregroundDisabled1`. then you can remove from the high contrast variables file.

packages/react/src/themes/teams-high-contrast/components/Checkbox/checkboxVariables.ts
```diff
@@ -0,0 +1,15 @@
...
+  return {
+    checkboxBorderColorHover: siteVars.accessibleYellow,
+    checkboxCheckedBackground: siteVars.accessibleCyan,
+    checkboxCheckedBackgroundHover: siteVars.accessibleYellow,
+    checkboxCheckedIndicatorColor: siteVars.colors.black,
+
+    toggleIndicatorColor: siteVars.colors.white,
+
+    disabledCheckboxCheckedIndicatorColor: siteVars.colors.black,
+    disabledToggleIndicatorColor: siteVars.colors.accessibleGreen,
```
codepretty: i don't think you need the `disabledToggleIndicatorColor` variable defined here since `foregroundDisabled1` already sets this to green in high contrast theme

packages/react/src/themes/teams/components/Checkbox/checkboxVariables.ts
bcalvery: I've removed most, however the checkboxBorderColorHover in this file is still necessary.

packages/react/src/themes/teams/components/Checkbox/checkboxStyles.ts
```diff
@@ -1,13 +1,107 @@
...
+          }),
+          ...(!p.checked && {
+            borderColor: v.checkboxBorderColorHover,
+            ...(p.toggle && {
+              color: v.checkboxCheckedBackgroundHover,
+            }),
+          }),
+        },
+      },
+    }),
```
layershifter, line 43: These styles looks strange to me...

```js
...(p.disabled && {
  // disabled styles
}),
...(!p.disabled && {
  // regular styles
})
```

Should it be just like below?

```js
// regular styles
...(p.disabled && {
  // disabled styles
}),
```

packages/react/src/themes/teams/components/Checkbox/checkboxStyles.ts
```diff
@@ -1,13 +1,107 @@
...
+
+    ...(p.disabled && {
+      background: v.disabledCheckboxBackground,
+      borderColor: v.disabledCheckboxBorderColor,
+
+      ...(p.checked && {
+        color: v.disabledCheckboxCheckedIndicatorColor,
+        background: v.disabledCheckboxBackgroundChecked,
+      }),
+    }),
```
layershifter, line 65: nit: I suggest to simplify a bit:

```js
...(p.disabled && {}),
...(p.disabled && p.checked && {}),
```

We will have less nesting levels in that case

packages/react/src/themes/teams/components/Checkbox/checkboxVariables.ts
```diff
@@ -1,22 +1,75 @@
...
+  togglePadding: string
+  toggleCheckedPadding: string
+
+  disabledColor: string
+  disabledCheckboxBackground: string
+  disabledCheckboxBackgroundChecked: string
+  disabledCheckboxBorderColor: string
+  disabledCheckboxCheckedIndicatorColor: string
+  disabledToggleIndicatorColor: string
+}
```
layershifter, line 22: Is there any reason why we can't use variables from `base` theme? If we need something custom we can define them:

```tsx
export type CheckboxVariables = Partial<BaseCheckboxVariables> & {
  additionalVariableForTeamsTheme1: string
  additionalVariableForTeamsTheme2: string
}
```

packages/react/src/themes/teams/components/Checkbox/checkboxVariables.ts
```diff
@@ -1,22 +1,75 @@
...
+
+const [toggleMovementDistance, checkboxPadding] = [20, 2].map(v => pxToRem(v))
```
layershifter: Can you please just inline this because it's hard to read things like this...

```
togglePadding: `${pxToRem(2)} ${pxToRem(20)} ${pxToRem(2)} ${pxToRem(2)}`,
```

packages/react/src/themes/teams/components/Checkbox/checkboxStyles.ts
```diff
@@ -1,13 +1,107 @@
...
+  }),
+
+  label: ({ props: p, variables: v }): ICSSInJSStyle => ({
+    [p.labelPosition === 'start' ? ':after' : ':before']: {
+      content: '" "',
+      display: 'inline-block',
+      width: v.checkboxGap,
+    },
+  }),
```
layershifter, line 73: We don't need this section with `label` at all because it duplicates all styles from `base` theme

packages/react/src/themes/base/components/Checkbox/checkboxStyles.ts
```diff
@@ -30,9 +30,9 @@ const checkboxStyles: ComponentSlotStylesInput<CheckboxProps & CheckboxState, Ch
     padding: v.checkboxPadding,
 
     ...(p.checked && {
-      background: v.checkedCheckboxBackground,
-      borderColor: v.checkedCheckboxBorderColor,
-      color: v.checkedCheckboxColor,
+      background: v.checkboxCheckedBackground,
```
mnajdova: Not sure why the name of these were changed. More over, not sure why we have the word `checkbox` in any variable of the `Checkbox` component. We may simplify this to `checkedBackground`

packages/react/src/themes/base/components/Checkbox/checkboxStyles.ts
```diff
@@ -54,29 +54,29 @@ const checkboxStyles: ComponentSlotStylesInput<CheckboxProps & CheckboxState, Ch
   }),
 
   toggle: ({ props: p, variables: v }): ICSSInJSStyle => ({
-    background: v.toggleBackground,
+    background: v.checkboxToggleBackground,
```
mnajdova: No need for adding `checkbox` in the variables names

packages/react/src/themes/teams/components/Checkbox/checkboxStyles.ts
mnajdova: By default, in `siteVariables` the `borderRadius` is defined to be `2px`. If we need 3px only for the checkbox, that we may parametrize the `getBorderFocusStyles`, to receive this param, but the default can be the value of the siteVariables. Would that solve your issue?

packages/react/src/themes/teams/components/Checkbox/checkboxVariables.ts
bcalvery: I thought teams theme was supposed to stand alone from base theme - and also changes to base theme could impact teams unexpectedly? @levithomason wasn't this something I heard from you?

packages/react/src/themes/teams/components/Checkbox/checkboxVariables.ts
bcalvery: I disagree in this case, those variables are important for maintaining proper behavior of the toggle (see the togglePadding and toggleCheckedPadding variables' values, they both use the movement distance. An aesthetic problem could occur if the inline values are changed in one and not the other.

packages/react/src/themes/teams/components/Checkbox/checkboxStyles.ts
bcalvery: I looked into it more. getBorderFocusStyles produces the correct result in the case of button and dropdown, but the focus indicator in the redline for the checkbox does have more radius in the corners. I think just changing it as a one-off within checkboxStyles is fine for now.

packages/react/src/themes/teams/components/Checkbox/checkboxVariables.ts
bcalvery:
```
const [toggleMovementDistance, padding] = [20, 2].map(v => pxToRem(v))
```
```
togglePadding: `${padding} ${toggleMovementDistance} ${padding} ${padding}`,
toggleCheckedPadding: `${padding} ${padding} ${padding} ${toggleMovementDistance}`,
```

packages/react/src/themes/teams/components/Checkbox/checkboxVariables.ts
levithomason: Base theme is intended to be the base theme for all themes.  It should only include values and variables that are applicable to all themes.  Things like "buttons have horizontal and vertical padding".  Base theme variables should be exposed for these things, the lowest common denominators for all themes.

alinais (approved):

packages/react/src/themes/teams/components/Checkbox/checkboxVariables.ts
levithomason: The `map` syntax looks strange to my eyes, but let's not bikeshed too long on it.  Progress trumps this one in priority.

bcalvery: I changed it to not use map, just two separate variables.

packages/react/src/themes/base/components/Checkbox/checkboxStyles.ts
```diff
@@ -30,27 +29,27 @@ const checkboxStyles: ComponentSlotStylesInput<CheckboxProps & CheckboxState, Ch
...
+    margin: v.margin,
+    padding: v.padding,
 
     ...(p.checked && {
-      background: v.checkedCheckboxBackground,
-      borderColor: v.checkedCheckboxBorderColor,
-      color: v.checkedCheckboxColor,
+      background: v.checkedBackground,
+      borderColor: v.checkedBorderColor,
+      color: v.checkboxCheckedColor,
```
alinais: nit: checkedColor

packages/react/src/themes/base/components/Checkbox/checkboxVariables.ts
```diff
@@ -1,79 +1,75 @@
...
-  checkedCheckboxBorderColor: string
-  checkedCheckboxColor: string
-  checkedToggleColor: string
-  checkedToggleBackground: string
-  checkedToggleBorderColor: string
-  checkedTogglePadding: string
+  checkedBackground: string
+  checkedBorderColor: string
+  checkboxCheckedColor: string
+  checkboxToggleCheckedColor: string
```
alinais, line 26: I think that `checkedToggleColor`, `checkedToggleBackground` and `checkedToggleBorderColor` namings were fine.

PR #1657, opened 2019-07-16
feat(Toolbar): responsive

miroslavstastny:
This PR adds possibility for a `Toolbar` to move items which do not fit to an overflow menu:
image

Based on requested use cases it is not enough to just remove items from end one by one and move them to the overflow menu. There is a requirement for the consumer to be able to specify custom "resize steps".

### API
`onReduceItems(renderedItems, measures): newItems` function has been added to `ToolbarProps`. If this prop is specified, `Toolbar` is watching its size and if it does not fit, it calls this function so that the consumer can restructure the items.

### Basic idea
Technically this work is inspired by [Office UI Fabric's ResizeGroup](https://developer.microsoft.com/en-us/fabric/#/controls/web/resizegroup).

image

1. the whole toolbar is wrapped by an additional `wrapper` div.
2. inside the div, there is a hidden `measurement` div.
3. inside the `measurement` div, a `hidden toolbar` is rendered
4. once the `hidden toolbar` is rendered, all its DOM children are measured and their rects are compared with `wrapper` rect to find out what overflows.
5. if there is an overflow, `onReduceItems` callback is called for the consumer to restructure the items, the `hidden toolbar` is re-rendered with the new items and the whole process repeats
6. once the `hidden toolbar` fits the `wrapper`, current items are rendered to `visible toolbar` and `hidden toolbar` is removed
7. there is a `resize detector` inside the `wrapper` to detect width changes - if width changes, `visible toolbar` is not touched, but the whole process inside `hidden toolbar` starts again with the initial (full) set of items

### Part of `Toolbar` vs `ResponsiveToolbar` vs universal wrapper

There are three options how to implement this:
1. as part of Toolbar implementation
👍user starts with `Toolbar`, wants to just "enable responsiveness"
👍implementation can be optimized for Toolbar
🛑complicates Toolbar implementation
🛑only for Toolbar
2. ResponsiveToolbar which composes Toolbar
👍separation of concerns
👍implementation can be optimized for Toolbar
🛑user needs to replace components to enable responsiveness
🛑only for Toolbar
3. universal wrapper
👍usable for any component
👍separation of concerns
🛑less intuitive for users
🛑cannot be optimized for Toolbar

After offline discussion with @levithomason we agreed to start with option 1. 

### Known issues, todos
- no performance test
- no performance optimizations
- `onReduceItems` implementation is in hands of a consumer and is not a straightforward/simple thing
- given current `ToolbarMenu` can contain only items and divider, there is no way how to move radiogroup and custom items to the overflow menu
- `requestAnimationFrame` registrations must be canceled on unmount
- the example should show how to handle dividers
- if overflow menu is opened and screen is resized so that the whole toolbar fits (=overflow menu disappears), and then resized back, the menu is automatically opened again


packages/react/src/themes/base/components/Toolbar/toolbarRadioGroupStyles.ts
```diff
@@ -0,0 +1,9 @@
+import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
+
+const toolbarRadioGroupStyles: ComponentSlotStylesInput = {
+  root: (): ICSSInJSStyle => ({
+    whiteSpace: 'nowrap',
```
miroslavstastny, line 5: without this items in radio group are wrapped if toolbar does not fit

layershifter:
### Dependency

- ❌ `react-resize-detector^4.2.0` 🚀 
  - ✔️ `lodash^4.17.11` - `4.17.12` 🚀 
  - ❌ `lodash-es^4.17.11` - `4.17.15` 🚀 
  - ✔️ `prop-types^15.7.2` - `15.7.2`
  - ❌ `raf-schd^4.0.0` - `4.0.2` 🚀 
  - ❌ `resize-observer-polyfill^1.5.1` - `1.5.1` 🚀

kuzhelov:
Detected by approved dependency check:

failed constrains | approved candidates
--- | --- 
lodash-es@^4.17.11 | _there are no any approved packages_
raf-schd@^4.0.0 | _there are no any approved packages_
react-resize-detector@^4.2.0 | _there are no any approved packages_
resize-observer-polyfill@^1.5.1 | _there are no any approved packages_

layershifter (approved):

miroslavstastny:
Screener test is not working for overflow Toolbar, created a separate issue #1788.

PR #1662, opened 2019-07-17
feat(Button): add loading prop

mnajdova:
We need to allow user to define loading state of the Buttons. This is a common thing already defined in many UI libraries, like: 
- Semantic UI React (loading prop) - https://react.semantic-ui.com/elements/button/#states-loading
- Ant design (loading prop) -	https://ant.design/components/button/#components-button-demo-loading
- Vuetify (loading prop) - https://vuetifyjs.com/en/components/buttons#button
- Evergreen (isLoading prop) - https://evergreen.segment.com/components/button/
- Atlaskit (loading prop) - https://atlaskit.atlassian.com/packages/core/button

Usually, the way it is implemented is, either with showing the loader indicator before the content of the button, or replacing the content of the button with loader:
image
image

In Teams theme, we need the first behavior - showing the loader indicator before the content (instead of the icon). 

# Proposed API
For showing the loading indicator, we may add `loading` prop in the Button component, that will show the indicator before the content. 

Furthermore, if we want to allow the users to customize this, we may add loaderPosition: `'start' | 'end' | 'center'`, which will position the loader before, after the content, or replace it if center is provided. Not sure if we want to add this right-away, but it's just an idea of how we may extend the API in the future.

# Alternative API
As we already have the icon slot, we may reuse that slot for rendering the `Loader` component if the loading prop is introduced. Some open questions here are, whether we want to change the component rendered there, in means of instead of `Icon.create(...)`, should we have `Loader.create(...)` and this applies then, do we want to reuse the icon slot for the `Loader`'s props as well, or introduce additional slot here...

In the styles, conditionally we may apply different margins if the loading prop is provided, but at least the anatomy of the component will be clear.

# Blockers
Currently this is blocked by the Loader component, as we don't have two variations for the loader that we may use for default and primary Button. The issues is tracked here: https://github.com/stardust-ui/react/issues/1663


packages/react/src/components/Button/Button.tsx
```diff
@@ -147,21 +152,31 @@ class Button extends UIComponent<WithAsProp<ButtonProps>, ButtonState> {
...
-      defaultProps: {
-        styles: styles.icon,
-        xSpacing: !content ? 'none' : iconPosition === 'after' ? 'before' : 'after',
-        variables: variables.icon,
-      },
-    })
+    const { icon, iconPosition, content, loading } = this.props
+
+    return loading
+      ? Loader.create(
```
mnajdova: Should we add slot of this? Should we make the loading shorthand for the Loader component?

packages/react/src/components/Button/Button.tsx
```diff
@@ -147,21 +152,31 @@ class Button extends UIComponent<WithAsProp<ButtonProps>, ButtonState> {
         {Box.create(!hasChildren && content, {
           defaultProps: { as: 'span', styles: styles.content },
         })}
-        {!hasChildren && iconPosition === 'after' && this.renderIcon(variables, styles)}
+        {!hasChildren && iconPosition === 'after' && !loading && this.renderIcon(variables, styles)}
```
mnajdova: In all cases that I found the Loader was never after the content, that's why I added this condition. We may reconsider this...

docs/src/examples/components/Button/States/ButtonExampleLoading.shorthand.tsx
```diff
@@ -0,0 +1,22 @@
...
+        <Button loading={loading} content="Secondary" secondary />
+        <Button loading={loading} icon="book" content="Click me" iconPosition="before" primary />
+        <Button loading={loading} circular icon="coffee" />
+      </Flex>
+      <Button loading={loading} fluid content="Fluid" />
+    </Flex>
+  )
+}
+
+export default ButtonExampleLoading
```
layershifter, line 11: Do we need to have all of these in this example? Can we have just `<Button content="Processing..." loading />` and move all others to hidden visual tests section? What do you think?

packages/react/src/components/Button/Button.tsx
```diff
@@ -144,21 +154,28 @@ class Button extends UIComponent<WithAsProp<ButtonProps>, ButtonState> {
...
+            styles: styles.loader,
+          },
+        })
+      : Icon.create(icon, {
+          defaultProps: {
+            styles: styles.icon,
+            xSpacing: !content ? 'none' : iconPosition === 'after' ? 'before' : 'after',
+            variables: variables.icon,
+          },
+        })
```
layershifter:
```jsx
const loaderElement = loading && Loader.create()
const iconElement = !loading && Icon.create()
```

If we will specify this inline in `renderComponent()`, will it be more clear?

mnajdova: Actually we are rendering both elements now, and the theme can hide one of them if necessary.

docs/src/examples/components/Button/States/ButtonExampleLoading.shorthand.tsx
mnajdova: Make sense!

packages/react/src/lib/accessibility/Behaviors/Loader/loaderBehavior.ts
```diff
@@ -8,7 +8,7 @@ import { Accessibility } from '../../types'
  * Adds role 'progressbar' to 'root' slot.
  */
 
-const loaderBehavior: Accessibility = () => ({
+const loaderBehavior: Accessibility = props => ({
```
miroslavstastny: why are you adding unused param?

mnajdova: Will remove, Milan was changing something for accessibility testing, but I refactored it later.. It's a craft.

layershifter (approved):

PR #1665, opened 2019-07-17
chore: check for approval of dependencies

kuzhelov:
This PR adds approvals checks for runtime dependencies of `@stardust-ui/react` packages, and compares related constraints against approved dependencies listed in `approvedPackages.ts` file. **If this check is failed, then PR that fails this check is blocked from being merged**. 

The check itself takes ~9s on my machine.

## How the check is working
Essentially, the logic of this check is captured by the following comment: https://github.com/stardust-ui/react/pull/1665/files#diff-9daef8cb0bbb429fbb252c54d14f1854R12 - however, for the sake of completeness will put it here:

 - request runtime dependencies of @stardust-ui/react package (by crawling the code, starting from index file),
 - for each of the runtime dependencies:
    - get corresponding set of version restrictions (by analyzing related package.json files),
    - get list of approved dependency's versions (from `approvedPackages.ts`),
    - check if list of approved versions covers the set of version restrictions.

## Output example on unapproved dependency detection
The following package version constraints missing approved candidate:

failed constrains | approved candidates
--- | --- 
@babel/runtime@^7.1.2 | @babel/runtime@2.1.2, @babel/runtime@2.4.5

## Notes on the approach

There are couple of moments that worth special attention. 

First is that only runtime dependencies are considered to be checked - this is done by crawling the actual code of `@stadust-ui/react` package and detecting the packages referenced (here logic relies on `license-webpack-plugin`). This ensures that only the code that is actually running is checked.

Second is that version restrictions are collected by analyzing all the necessary `package.json` files in the dependencies tree. This guarantees that the most strict set of package version constraints will be applied during the check.


build/dangerjs/detectNonApprovedDependencies/utils/getRuntimeDependencies.ts
```diff
@@ -0,0 +1,30 @@
+import { spawnSync } from 'child_process'
+
+import config from '../../../../config'
+
+const { paths } = config
+
+const getRuntimeDependencies = () => {
+  const dependencyRegex = /^dependency:\s+(.*)$/
+  const result = spawnSync("yarn gulp test:dependencies:list --prefix='dependency: '", {
```
kuzhelov: unfortunately, it is necessary to introduce this complication of running weback in a separate process - otherwise, if run in DangerJS's process, `license-webpack-plugin` won't detect dependencies (presumably, because of the way how DangerJS handles imports, but for the sake of saving time haven't made a deep dive into this)

build/dangerjs/detectNonApprovedDependencies/utils/getVersionConstrains.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+const parsePackageJson = (packageJsonPath: string): Promise<PackageJson> => {
+  return new Promise((resolve, reject) => {
+    readPackageJson(packageJsonPath, (error, data) => {
+      if (error) {
+        reject(new Error(`There was an error reading the ${packageJsonPath} file.`))
+      }
+
+      const dependencies = data.dependencies || {}
```
kuzhelov: constraints of `peerDependencies` should also be included

dangerfile.ts
```diff
@@ -1,136 +1,12 @@
...
-        shouldLogWarning = true
-      }
-      if (!_.isEmpty(changedPeerDependencies)) {
-        markdownChangedDependencies(filepath, changedPeerDependencies, 'peerDependencies')
-        shouldLogWarning = true
-      }
-      return shouldLogWarning
-    }, false)
-}
+const dangerJS = { danger, fail, warn, markdown }
```
kuzhelov, line 10: unfortunately, this trick (of explicitly passing `dangerJS` utils as function arg, instead of importing them at places where needed) is necessary due to the magic DangerJS is doing with imports: https://spectrum.chat/danger/javascript/danger-js-actually-runs-your-imports-as-globals~0a005b56-31ec-4919-9a28-ced623949d4d

build/dangerjs/detectChangedDependencies.ts
```diff
@@ -0,0 +1,98 @@
+import * as _ from 'lodash'
+import { DangerDSLType } from 'danger'
+
+type DangerJS = {
```
layershifter: Can we extract this type and use it everywhere, i.e. `checkChangelog()`, `detectChangedDependencies()`, etc?

dangerfile.ts
layershifter: Let's add a comment to code instead of PR

build/dangerjs/checkChangelog.ts
```diff
@@ -0,0 +1,68 @@
...
+      return acc.concat(filteredLines)
+    }, [])
+  })
+}
+
+export default async ({
+  danger,
+  fail,
+  warn,
+}: {
```
layershifter: A use case for `DangerJS` type

build/dangerjs/detectNonApprovedDependencies/approvedPackages.ts
```diff
@@ -0,0 +1,69 @@
+export default [
+  '@babel/runtime@7.1.2',
+  '@babel/runtime@7.4.5',
+  'classnames@2.2.6',
+  'compute-scroll-into-view@1.0.11',
+  'css-in-js-utils@2.0.1',
+  'css-in-js-utils@3.0.0',
```
layershifter, line 7: Can we cleanup this file to have only actual dependencies there? In a follow up PR of course

build/dangerjs/detectNonApprovedDependencies/utils/getRuntimeDependencies.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  if (error) {
+    throw new Error(error)
+  }
+
+  return output
+    .split('\n')
+    .map(line => line.match(dependencyRegex))
+    .filter(match => !!match)
+    .map(match => match[1])
```
layershifter, line 35: This thing looks like a bit overcomplicated to me. To get dependencies we need to run `webpack` with `license-webpack-plugin`, instead of this:
1) Gun Gulp task
2) Run Webpack
3) Parse task output

So, may be we can inline this check into this function?

build/dangerjs/detectNonApprovedDependencies/utils/getVersionConstrains.ts
```diff
@@ -0,0 +1,123 @@
...
+    })
+  })
+}
+
+export const getDependenciesVersionConstraints = async (
+  packageJsonPath: string,
+  dependencyChain: string[],
+): Promise<Constraints> => {
+  let detectedConstraints: Constraints = {}
+  const dependenciesWithConstraints = (await parsePackageJson(packageJsonPath)).dependencies // and peer dependencies?
```
layershifter: Is this comment valid? I see that in `parsePackageJson()` we return both:

```
const dependencies = { ...data.peerDependencies, ...data.dependencies }
```

build/dangerjs/detectNonApprovedDependencies/utils/getRuntimeDependencies.ts
layershifter: It also fails on Windows...

```
[11:12:01] Task never defined: '
[11:12:01] To list available tasks, try running: gulp --tasks

Error: [11:12:01] Task never defined: '
[11:12:01] To list available tasks, try running: gulp --tasks

    at Object.getRuntimeDependencies (C:\Users\olfedias\WebstormProjects\react\build
\dangerjs\detectNonApprovedDependencies\utils\getRuntimeDependencies.ts:17:15)
    at Object.<anonymous> (C:\Users\olfedias\WebstormProjects\react\build\dangerjs\d
etectNonApprovedDependencies\index.ts:57:48)
    at step (C:\Users\olfedias\WebstormProjects\react\build\dangerjs\detectNonApprov
edDependencies\index.ts:32:23)
    at Object.next (C:\Users\olfedias\WebstormProjects\react\build\dangerjs\detectNo
nApprovedDependencies\index.ts:13:53)
    at C:\Users\olfedias\WebstormProjects\react\build\dangerjs\detectNonApprovedDepe
ndencies\index.ts:7:71
    at new Promise (<anonymous>)
    at __awaiter (C:\Users\olfedias\WebstormProjects\react\build\dangerjs\detectNonA
pprovedDependencies\index.ts:3:12)
    at Object.detectNonApprovedDependencies [as default] (C:\Users\olfedias\Webstorm
Projects\react\build\dangerjs\detectNonApprovedDependencies\index.ts:50:66)
    at Object.<anonymous> (dangerfile.ts:45:81)
    at step (dangerfile.ts:32:23)
```

build/dangerjs/detectNonApprovedDependencies/index.ts
```diff
@@ -0,0 +1,64 @@
...
+ * - for each of the runtime dependencies:
+ *    - get corresponding set of version restrictions (by analyzing related package.json files),
+ *    - get list of approved dependency's versions,
+ *    - check if list of approved versions covers the set of version restrictions.
+ */
+const detectNonApprovedDependencies = async dangerJS => {
+  const { fail, markdown } = dangerJS
+  const failedVersionConstraints: FailedConstraintsExplanation[] = []
+
+  const dependencyPackageIds = getRuntimeDependencies('react')
```
layershifter, line 25: We use `dependencyPackageIds` only in this file and it is an array of strings:

```jsx
[
    'rtl-css-js@1.10.0',
    'scheduler@0.13.6',
    'webpack@4.35.0',
]
```

But in `checkPackageVersionConstraints` we do:
```jsx
const packageName = getPackageName(packageId)
```

`getPackageName()` is used only once, so I think that we can change the structure to `object`:
```
{
  webpack: '4.35.0',
}
```

build/dangerjs/detectNonApprovedDependencies/utils/checkPackageVersionConstraints.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const approvedPackageIds = approvedPackages.filter(item => item.startsWith(`${packageName}@`))
+
+  approvedPackageIds.map(getPackageVersion).forEach(approvedPackageVersion => {
+    Object.keys(versionConstraintsCheckResult).forEach(versionConstraint => {
+      if (semver.satisfies(approvedPackageVersion, versionConstraint)) {
+        versionConstraintsCheckResult[versionConstraint] = true
+      }
+    })
+  })
```
layershifter: It seems that there is a bug with `webpack` at least:

```
versionConstraintsCheckResult webpack {}
approvedPackageIds [ 'webpack@4.25.1' ]
approvedPackageIds.map(getPackageVersion) [ '4.25.1' ]
versionConstraintsCheckResult webpack {}
```

It's safe to ignore for Webpack, but can you please check other packages?

build/dangerjs/detectNonApprovedDependencies/utils/checkPackageVersionConstraints.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const failedConstraints = Object.keys(versionConstraintsCheckResult).reduce(
+    (acc, versionConstraint) =>
+      !versionConstraintsCheckResult[versionConstraint] ? [...acc, versionConstraint] : acc,
+    [],
+  )
```
layershifter: Can use `lodash` in this case? It's hard to understand what happens here...

build/dangerjs/detectNonApprovedDependencies/utils/checkPackageVersionConstraints.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const versionConstraintsCheckResult = versionConstraints.reduce(
+    (acc, versionConstraint) => ({ ...acc, [versionConstraint]: false }),
+    {},
+  )
+
+  const approvedPackageIds = approvedPackages.filter(item => item.startsWith(`${packageName}@`))
+
+  approvedPackageIds.map(getPackageVersion).forEach(approvedPackageVersion => {
+    Object.keys(versionConstraintsCheckResult).forEach(versionConstraint => {
```
layershifter: Can we split this to multiple steps to make it more obvious?

build/dangerjs/detectNonApprovedDependencies/utils/checkPackageVersionConstraints.ts
```diff
@@ -0,0 +1,62 @@
...
+    ? { isApproved }
+    : {
+        isApproved: false,
+        explain: {
+          failedConstraints: failedConstraints.map(
+            versionConstraint => `${packageName}@${versionConstraint}`,
+          ),
+          approvedPackages: approvedPackageIds,
+        },
+      }
```
layershifter: We can return something only on fail, it will also simplify structure that will be returned:

```tsx
  if (failedConstraints.length > 0) {
    return {
      failedConstraints: failedConstraints.map(
        versionConstraint => `${packageName}@${versionConstraint}`,
      ),
      approvedPackages: approvedPackageIds,
    },
  }
```

build/dangerjs/detectNonApprovedDependencies/approvedPackages.ts
kuzhelov: yes, we could - however, this file is intended to answer the following question: what are the packages that are approved? It is not this file's intent to reflect the dependencies that are 'actually' used - to save on maintainability costs. We may display actually used dependencies from the approved ones on each build, as part of DangerJS output instead

build/dangerjs/detectNonApprovedDependencies/utils/checkPackageVersionConstraints.ts
kuzhelov: have discussed offline, this `reduce` expression will be removed, overall code of the module will be simplified

build/dangerjs/detectNonApprovedDependencies/utils/getRuntimeDependencies.ts
kuzhelov: for the first part of the question - unfortunately, this is a necessary complication, as otherwise there will be a problem for `license-webpack-plugin` to crawl through the imported modules. Have added comment that will describe this.

Have debugged Windows-related issue with @layershifter, it was caused by double quotes in the string command. This issue was fixed

build/dangerjs/checkChangelog.ts
kuzhelov: yep, introduced 👍

dangerfile.ts
kuzhelov: added comment

build/dangerjs/detectNonApprovedDependencies/utils/checkPackageVersionConstraints.ts
kuzhelov: discussed offline, agreed that logic of the approvals check handles the case correctly

layershifter (approved):

alinais (approved):

build/dangerjs/detectNonApprovedDependencies/utils/getRuntimeDependencies.ts
```diff
@@ -0,0 +1,39 @@
+import { spawnSync } from 'child_process'
+
+import config from '../../../../config'
+
+const { paths } = config
+
+/**
+ * It is necessary to cotrresppondign Gulp task in a separate process.
```
kuzhelov: typo

build/dangerjs/detectNonApprovedDependencies/index.ts
kuzhelov: I would like to leave it as it is for now, because this format resembles output of most related NPM utils, so it is quite easy to copy and paste their output to this file, without any additional parsing/composing/processing actions. This, essentially, was a key factor that made it very easy to me to prepare original version of this list.

However, if there will be strong need for providing strict data structure for this list, we could always rethink the approach.

PR #1667, opened 2019-07-17
feat(Popup): altered controlled example to contain close icon in the content

mnajdova:
This PR adds closing icon on the Popup's content in the controlled example. Not sure if this should be added in the API at this moment, as I could not see it as a common thing in the other UI libraries. For the Dialog we added `headerAction` button that is appearing in the right corner, maybe we should add something like that for the popup content, but it still seems awkward as we don't have the a header prop there... Even if we add it as a header action, it would still be user's responsibility to close the popup when that action is triggered, so not sure if it would bring any value.

Thoughts: @kuzhelov @layershifter @miroslavstastny


kuzhelov:
I would vote for the safest option we have now - add this as prototype/example, do not change the API and, thus, **_ensure that we are providing client with all the necessary tools to handle this scenario_**. This also will allow us to rethink this approach in future, if necessary.

docs/src/examples/components/Popup/Types/PopupControlledExample.shorthand.tsx
```diff
@@ -1,24 +1,27 @@
...
+  const [open, setOpen] = useBooleanKnob({ name: 'open' })
 
-  handleOpenChange = (e, { open }) => {
-    alert(`Popup requested to change its open state to "${open}".`)
-    this.setState({ open })
-  }
+  const popupContent = (
+    <Flex column>
+      <Flex.Item align="end">
+        <Button text iconOnly icon="close" onClick={() => setOpen(false)} />
```
jurokapsiar: please add `title='Close'`

jurokapsiar: also, not directly related to the PR but since the popup contains actionable components, it should have `trapFocus`.

mnajdova: Will add the `trapFocus` prop. Also added title on the trigger button as it only contains an icon

miroslavstastny (approved):

PR #1669, opened 2019-07-17
chore(Dropdown): screener test to select dropdown item outside popup

### LINKED ISSUE #1673: Selecting a participant in "add to chat" dialog does nothing
Create screener test for the issue below:

In the add participant dropdown, if I select a row beyond the scope of the parent UI element, that participant is not added. 

If you select on Karvel's name which is outside of the add participant dialog boundary then the selection is ignored. 

image

silviuaavram:
Adds test for a previous issue with a `Dropdown` in a `Popup`: if the item was visible outside the popup, it was not possible to select it.

Added the popup with dropdown as an `Example` although it should've been added as a `Prototype`. The reason is the fact that we don't have infrastructure yet for Prototypes to receive screener tests.

Closes https://github.com/stardust-ui/react/issues/1673


jurokapsiar (approved):

docs/src/examples/components/Popup/Usage/PopupExampleCloseButton.shorthand.tsx
```diff
@@ -0,0 +1,115 @@
...
+        trigger={
+          <Button
+            icon="user"
+            content="People Picker"
+            onClick={this.openPopup}
+            aria-label="Choose a person."
+          />
+        }
+        content={{
+          content: (
```
layershifter: We don't need double `content` anymore 🐱 

```diff
-        content={{
-          content: (
+       content={(
```

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -132,6 +133,7 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
 
   static slotClassNames: PopupSlotClassNames = {
     content: `${Popup.className}__content`,
+    trigger: `${Popup.className}__trigger`,
```
layershifter: I am not sure that we really need this because we don't apply any styles for `trigger`. `trigger` is not also a slot.

docs/src/examples/components/Popup/Usage/PopupExampleCloseButton.shorthand.steps.ts
```diff
@@ -0,0 +1,23 @@
...
+  item: (itemIndex: number) => `.${Dropdown.slotClassNames.itemsList} li:nth-child(${itemIndex})`,
+  popupTrigger: `.${Popup.slotClassNames.trigger}`,
+}
+
+const config: ScreenerTestsConfig = {
+  themes: ['teams', 'teamsDark', 'teamsHighContrast'],
+  steps: [
+    (builder, keys) =>
+      builder
+        .click(selectors.popupTrigger)
```
layershifter, line 14: https://github.com/stardust-ui/react/pull/1669/files#r305348228

You can use `Button.className` instead

docs/src/examples/components/Popup/Usage/PopupExampleCloseButton.shorthand.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  closePopup = e => {
+    this.setState({
+      open: false,
+    })
+  }
+
+  openPopup = () => {
+    this.setState({
+      open: true,
```
layershifter: Can we use knobs instead?

layershifter (changes requested):
docs-popup

It seems that click outside and on a trigger is not properly handled

docs/src/examples/components/Popup/Usage/index.tsx
```diff
@@ -35,6 +35,11 @@ const Usage = () => (
       description="Popups can be nested."
       examplePath="components/Popup/Usage/PopupExampleNested"
     />
+    <ComponentExample
+      title="With close button"
+      description="Popups can have close button."
```
layershifter, line 40: Does this `title` and `description` represent actual example? 🤔

layershifter (approved):

docs/src/examples/components/Popup/Usage/PopupExampleCloseButton.shorthand.tsx
```diff
@@ -0,0 +1,96 @@
+import { useBooleanKnob } from '@stardust-ui/docs-components'
+import * as React from 'react'
+import { Button, Dropdown, Popup, Flex } from '@stardust-ui/react'
+
+const inputItems = [
```
bmdalex, line 5: we want to make examples as short as possible; can we make this just a minimal:

```typescript
const inputItems = ['Bruce Wayne', 'Natasha Romanoff' ]
```

I guess it's enough for this example. Would be great to open a PR and go through `Dropdown` examples and clean them up a little bit so that users can focus on what's relevant to that particular example. 

fyi @layershifter

docs/src/examples/components/Popup/Usage/PopupExampleCloseButton.shorthand.tsx
```diff
@@ -0,0 +1,96 @@
...
+    content: 'Partner Software Engineer',
+  },
+  {
+    header: 'Selina Kyle',
+    image: 'public/images/avatar/small/ade.jpg',
+    content: 'Graphic Designer',
+  },
+]
+
+const getA11ySelectionMessage = {
```
bmdalex, line 53: is this really needed for this example?

docs/src/examples/components/Popup/Usage/PopupExampleCloseButton.shorthand.tsx
```diff
@@ -0,0 +1,96 @@
...
+    setOpen(false)
+  }
+
+  const popupContent = (
+    <>
+      <Dropdown
+        search
+        multiple
+        items={inputItems}
+        placeholder="Start typing a name"
```
bmdalex, line 71: is this really needed for this example?

docs/src/examples/components/Popup/Usage/PopupExampleCloseButton.shorthand.tsx
```diff
@@ -0,0 +1,96 @@
...
+        getA11ySelectionMessage={getA11ySelectionMessage}
```
bmdalex, line 72: is this really needed for this example?

docs/src/examples/components/Popup/Usage/PopupExampleCloseButton.shorthand.tsx
```diff
@@ -0,0 +1,96 @@
...
+        noResultsMessage="We couldn't find any matches."
```
bmdalex, line 73: is this really needed for this example?

docs/src/examples/components/Popup/Usage/PopupExampleCloseButton.shorthand.tsx
```diff
@@ -0,0 +1,96 @@
...
+      />
+      <br />
```
bmdalex, line 75: nit: just wondering if `Flex` `gap` is better for achieve this, if it gets complicated leave it as is

docs/src/examples/components/Popup/Usage/PopupExampleCloseButton.tsx
```diff
@@ -0,0 +1,97 @@
+import { useBooleanKnob } from '@stardust-ui/docs-components'
+import * as React from 'react'
+import { Button, Dropdown, Popup, Flex } from '@stardust-ui/react'
+
+const inputItems = [
```
bmdalex, line 5: same as previous example

bmdalex (approved):
no comments are critical, feel free to decide what's relevant

silviuaavram:
@Bugaa92 I think those are needed yes, as we want that Dropdown to behave well from every perspective when a user checks it out in the docs, not just to use it in our test.

I don't want to strip the Dropdowns of everything they have (placeholder, a11y status message etc) so I can illustrate the fact that you can add a `noResultsMessage`. If we decide to do that, fine, we are in the process of improving the docs anyway.

@layershifter how would you rename the example? I am pretty bad at this :D

Issue #1585, opened 2019-07-03
Muted attribute doesn't work on Embed videos


stuartlong:
## Bug Report

Settings `muted` or `muted={true}` on `Embed` components doesn't actually mute them.

### Steps
1. Create an embed component
2. Set it's source video as a mp4 video with sound
3. Add `muted={true}` to the embed control

### Expected Result
There should be a way to mute the component

### Actual Result
Can't mute

### Version
0.34.0

### Testcase
https://codesandbox.io/s/stardust-ui-example-owdmg

layershifter:
`muted` is a prop of `Video` components, but this issue exists there, too: https://codesandbox.io/s/62dpz?module=/example.js

stuartlong:
Right, yes, I don't expect Embed to take a specific muted attribute. But even setting the intrinsic muted attribute doesn't work.

layershifter:
Was fixed in #1847.

PR #1670, opened 2019-07-17
docs(Playground): add auto generated playgrounds

layershifter:
# Autogenerated playgrounds 🚀 

This PR adds autogenerated playgrounds for components (see exceptions below). It allows:
- to disable playgrounds (via blacklist)
- allows to use manually defined (`Menu` playground was untouched)

Adds also `useNumberHook`().

## Generation strategy

We use existing hooks like `useSelectKnob`, `useStringKnob`, etc. to handle state and reuse existing knobs. Each generator is described by `KnobGenerator` which contains `hook` & options for it.

- `componentGenerators.ts` contains generators for props specific to components, has the highest priority.
- `propGenerators.tsx` contains generators for props, they will be preferred over other options. Is used to handle corner cases like `src` where type is `string` but it should a url to an image.
- `typeGenerators.ts` contains generators for types
- `createHookGenerator.tsx` contains blacklist for unsupported props 

#### No shorthands for `Box`

It's very hard to understand what represents `Box` component.

### Add `animation` pro

As @mnajdova [suggested](https://github.com/stardust-ui/react/pull/1670#issuecomment-512339937), we can easily have it.

### Remove/disable incorrect prop combinations

As @mnajdova [mentioned](https://github.com/stardust-ui/react/pull/1670#issuecomment-512339937), some prop combinations are conflicting. Examples:
```tsx
<Button content="Foo" iconOnly />
<Button content="Foo" primary secondary />
```

### No type for `array` in props table

Present in `Accordion`:
image

We should specify `array` of what types. Expected `number | number[]`.

### No support for action handlers

We should add support for `onClick`, `onFocus`, etc. And logs actions from them.

### Missing autocontrolled props

We should handle them in controlled way by knobs, it's supported but needs additional work.

### Enable `Playground` for components

`Provider` and `Box` components probably should not have `Playground` at all.

It's hard to generate collections for components below. Playground for them is disabled, but will be enabled later.

- `Chat`
- `Accordion`
- `Dropdown`
- `Flex`
- `Form`
- `Grid`
- `Portal`
- `RadioGroup`
- `Toolbar`
- `Tree`

### TODO
- [x] allow to set `useSelectionKnob` to none
- [x] resolve an issue with icons in `Input`, `Checkbox`
- [x] fix `Divider` preview in `Playground`
- [x] hide autocontrolled props
- [x] remove playgrounds for `Layout` and `ItemLayout`
- [x] fix missing `label` on `Loader`
- [x] add missing `ColorComponentProps` interface for `Segment`


docs/src/examples/components/Chat/Playground.tsx
```diff
@@ -0,0 +1,2 @@
+// Playground is not supported by this component
+export default null
```
mnajdova: Can't we just check if this file is not present? This seems really awkward.. Or at least let's create Empty playground component, or something like that..

layershifter: We have:
- autogenerated playground (no file)
- we have a generated playground (file is present)
- no playground at all

Can you suggest a better option for the last item?...

I believe that once we will create missing playgrounds or improve their generation third option will not be required anymore. But I don't have more simple idea for now.

mnajdova:
Generally, looks awesome. I have few comments/suggestions from user's perspective, let's just note them here and we can create a follow up PRs later:

- select knob cannot be set to none, we need to have some option to remove that prop from the component
- animation prop - can be a select with the animations defined in the theme, something similar to how you are picking the icons I suppose
- one potential problem that I see is that, for the components that we have some `stardust-*` icon provided, like checkbox, input, dropdown, it seems really awkward that we have some other random icon there... Also, maybe we should have the stardust-* icons in the select always?
- It looks funny that all buttons `iconOnly`, `text` that we have across the components, now have text :D. if there is a way to see that this prop is provided, it would be great to add icon there as an option, but it may be hard as we are setting this when the component is created. Anyway this is just something to think about
- If I check the open boolean knob on `Dialog`, I cannot close it
- The divider doesn't have the line for some reason...
- There is an issue with the colors, for example on the Header playground, by default the color is brand, but it is not applied - if I change it to something else and then back to brand it works
- Let's remove item layout's and layouts's playground
- Maybe it would be nice to have rtl on the playground as well :) 
- Loader - there is labelPosition prop, but there is not input for the label text
- Segment doesn't display color prop - strange that I cannot see it in the props, but we have even example for it...
- Slider doesn't have input for value - not sure if we want to add it at this moment, see next comment
- On the Tooltip playground, check open and it is opened, on unchecked nothing happens... Not sure if we should expose this auto controlled props, until we make them work...

layershifter:
@mnajdova thanks for the detailed feedback 👍 

> * select knob cannot be set to none, we need to have some option to remove that prop from the component

Should be fixed before merge, moved to TODOs.

> * animation prop - can be a select with the animations defined in the theme, something similar to how you are picking the icons I suppose

Can be easily implemented, added to improvements list.

> * one potential problem that I see is that, for the components that we have some `stardust-*` icon provided, like checkbox, input, dropdown, it seems really awkward that we have some other random icon there... Also, maybe we should have the stardust-* icons in the select always?

Moved to TODOs to resolve before merge.

> * It looks funny that all buttons `iconOnly`, `text` that we have across the components, now have text :D. if there is a way to see that this prop is provided, it would be great to add icon there as an option, but it may be hard as we are setting this when the component is created. Anyway this is just something to think about

There should be probably a logic that disallows to use some prop combinations. Move to feature list because this thing is quite complicated.

> * If I check the open boolean knob on `Dialog`, I cannot close it

I will hide all autocontrolled props, we will reenable them later - added to feature list.

> * The divider doesn't have the line for some reason...

Added to TODOs, will fix.

> * There is an issue with the colors, for example on the Header playground, by default the color is brand, but it is not applied - if I change it to something else and then back to brand it works

Will be fixed by `none` in selects.

> Maybe it would be nice to have rtl on the playground as well :)

We have it in designs so we will have it in future 👍 

> Loader - there is labelPosition prop, but there is not input for the label text

It's related to `Box` components, however we should use `Text` in this case.

> Segment doesn't display color prop - strange that I cannot see it in the props, but we have even example for it...

Will fix.

> * Slider doesn't have input for value - not sure if we want to add it at this moment, see next comment
> * On the Tooltip playground, check open and it is opened, on unchecked nothing happens... Not sure if we should expose this auto controlled props, until we make them work...

Will be resolved as other cases with autocontrolled props.

docs/src/examples/components/Chat/Playground.tsx
lucivpav: I agree that we should not specify this using `null`. How about `NoPlayground` or `DisablePlayground` constant? Then we can also get rid of all the associated comments.

docs/src/components/ComponentPlayground/ComponentPlayground.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <div
+          style={{
+            fontSize: 10,
+            border: '2px dotted pink',
+            margin: 5,
+            padding: 5,
+            maxWidth: '200px',
+          }}
+        >
+          <b>Props not supported by generators:</b> {unsupportedProps.join(' | ')}
```
lucivpav: I think it may not be clear to the user, what "generators" actually means. How about rephrasing this to something like: "Props not supported in playground: ".

lucivpav:
Thank you for the PR, it's really useful. Here are my notes, what I have found. It is possible that some of these issues are actually present in the component themselves, rather than in this PR.

`Alert` component: `attached` prop does nothing. Actually, it is supposed to do nothing in this case, but it can be confusing to newbies, who try to figure out why is this toggle not having any effect. How about:
- blacklisting this prop, for the `Alert` component
- Adding another content, so that `Alert` can attach to it (but it may be also confusing; user may ask, why is there the other content)

`Attachment` component: `disabled` prop has no visual effect. should it be this way?

`Button` component: `secondary` prop has no visual effect either

`Dialog` component: `color` has no effect, which is confusing. `trapFocus` contains only one element. can we remove such single-element dropdown elements?

`Divider` component: since `fitted` makes no visual change, should it be available to play with?

`Circular` component: `circular`, `color`, `circular`: no effect

`Input` component: `inline` makes (not surprisingly) no effect, should it be removed?

`List` component: seems like it will always stay as an empty list

`Loader`: only the `size` prop seems to have an effect

`Popup`: what does `inline` do?

`Text`: `disabled`, `error`, `success` - no effect, because of the color set. And there is no option not to specify a color. `size` - no effect.

`Video`: `muted` - no effect

layershifter:
@mnajdova 

> one potential problem that I see is that, for the components that we have some stardust-* icon provided, like checkbox, input, dropdown, it seems really awkward that we have some other random icon there... Also, maybe we should have the stardust-* icons in the select always?

Partially solved with `none`, by default we have a proper `icon`. We should improve this with better defaults, I think.

layershifter:
@lucivpav thanks for the feedback 👍 Please check comments below.

> `Alert` component: `attached` prop does nothing. Actually, it is supposed to do nothing in this case, but it can be confusing to newbies, who try to figure out why is this toggle not having any effect. How about:

image

image

It's not true, but the effect is not obvious.

> `Attachment` component: `disabled` prop has no visual effect. should it be this way?

It affects only `onClick` handler, but it's not present in playgrounds yet. Please create a new issue about it, I believe that there should be an visual representation.

> Button component: secondary prop has no visual effect either

Yep, it's a know issue because Teams theme doesn't have it.

> Dialog component: color has no effect, which is confusing. trapFocus contains only one element. can we remove such single-element dropdown elements?

Great catch, now it's a `boolean`:
image

> Divider component: since fitted makes no visual change, should it be available to play with?
> Input component: inline makes (not surprisingly) no effect, should it be removed?

Not sure about these one. @mnajdova do you have any ideas?

> List component: seems like it will always stay as an empty list

Good catch, I removed playground for it.

> Loader: only the size prop seems to have an effect

https://stardust-ui.github.io/react//components/loader#variations-size
It has, but not all sizes are present in Teams theme.

> Circular component: circular, color, circular: no effect

We don't have such component, can you please clarify?

> Popup: what does inline do?

It renders a component as a sibling in DOM instead of `document.body`. Looks like the same thing with `Divider` and `Input`.

> Text: disabled, error, success - no effect, because of the color set. And there is no option not to specify a color. size - no effect.

`size` has an effect. Other things are not related to this PR.

> Video: muted - no effect

It's a bug, #1585

docs/src/components/ComponentPlayground/ComponentPlayground.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    return <ComponentPlaygroundTemplate component={component} />
+  }
+
+  const [element, unsupportedProps] = usePlaygroundComponent(props.componentName)
+
+  return (
+    <ComponentPlaygroundTemplate element={element}>
+      {process.env.NODE_ENV === 'production' ? null : (
+        <div
```
bmdalex, line 54: what about using `Alert` instead?

docs/src/components/ComponentPlayground/ComponentPlaygroundSnippet.tsx
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * This component uses `react-element-to-jsx-string` to get a generated markup with props.
+ */
+const ComponentPlaygroundSnippet: React.FunctionComponent<
+  ComponentPlaygroundSnippetProps
+> = props => {
+  const { element, component, ...rest } = props
+
+  if (process.env.NODE_ENV !== 'production') {
+    if (typeof component === 'function' && !!component.prototype.isReactComponent) {
```
bmdalex, line 20: this is pretty ugly... React does not offer any helper methods for this? similar to `React.isValidElement`

docs/src/components/ComponentPlayground/propGenerators.tsx
```diff
@@ -0,0 +1,128 @@
...
+    return {
+      hook: useSelectKnob,
+      name: propName,
+      allowsNone: _.isNil(propDef.defaultValue),
+      initialValue: propDef.defaultValue,
+      values,
+    }
+  }
+
+  if (componentInfo.displayName === 'Avatar') {
```
bmdalex: 💣 this code creates tight coupling between `propGenerators.tsx` and `Avatar.tsx`. Can we extract this in some `enhancers` file and have the logic there and consume `enhancers` here?

docs/src/components/ComponentPlayground/propGenerators.tsx
```diff
@@ -0,0 +1,128 @@
...
+  allowsNone: true,
+  initialValue: propDef.defaultValue,
+  values: Object.keys({
+    ...theme.siteVariables.contextualColors,
+    ...theme.siteVariables.naturalColors,
+  }),
+})
+
+export const name: KnobGenerator<string> = ({ componentInfo, propDef, propName, theme }) => {
+  if (componentInfo.displayName === 'Icon') {
```
bmdalex: same as https://github.com/stardust-ui/react/pull/1670#discussion_r305893437

docs/src/components/ComponentPlayground/propGenerators.tsx
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+
+  throw new Error(
+    `A generated value for "name" prop is supported only on "Avatar" and "Icon" components`,
+  )
+}
+
+export const poster: KnobGenerator<string> = ({ componentInfo, propName }) => {
+  if (componentInfo.displayName !== 'Video') {
```
bmdalex: same as https://github.com/stardust-ui/react/pull/1670#discussion_r305893437

docs/src/components/ComponentPlayground/propGenerators.tsx
```diff
@@ -0,0 +1,128 @@
...
+  if (propDef.types.length > 1) {
+    throw new Error(
+      `A "${
+        componentInfo.displayName
+      }" for "size" prop defines multiple types, it is not supported`,
+    )
+  }
+
+  // Workaround for `Divider` component that supports size in different way
+  if (componentInfo.displayName === 'Divider' && propDef.types[0].name === 'number') {
```
bmdalex: same as https://github.com/stardust-ui/react/pull/1670#discussion_r305893437

docs/src/components/ComponentPlayground/propGenerators.tsx
```diff
@@ -0,0 +1,128 @@
...
+})
+
+export const trapFocus: KnobGenerator<boolean> = ({ componentInfo, propDef, propName }) => ({
+  hook: useBooleanKnob,
+  name: propName,
+  initialValue: propDef.defaultValue,
+})
+
+export const src: KnobGenerator<string> = ({ componentInfo, propName }) => {
+  if (componentInfo.displayName === 'Image') {
```
bmdalex: same as https://github.com/stardust-ui/react/pull/1670#discussion_r305893437

docs/src/components/ComponentPlayground/propGenerators.tsx
```diff
@@ -0,0 +1,128 @@
...
+    return {
+      hook: useStringKnob,
+      name: propName,
+      initialValue: faker.image.avatar(),
+    }
+  }
+
+  if (componentInfo.displayName === 'Video') {
```
bmdalex: same as https://github.com/stardust-ui/react/pull/1670#discussion_r305893437

docs/src/components/ComponentPlayground/typeGenerators.ts
```diff
@@ -0,0 +1,80 @@
...
+  const shorthandComponentName: string = shorthandPropType.replace(/Props$/, '')
+
+  const shorthandComponentInfo: ComponentInfo =
+    componentInfoContext.byDisplayName[shorthandComponentName]
+
+  const mappedShorthandProp = shorthandComponentInfo.mappedShorthandProp
+  const shorthandPropDef = _.find(shorthandComponentInfo.props, { name: mappedShorthandProp })
+
+  // TODO: fix support for boxes
+  if (shorthandComponentName === 'Box') {
```
bmdalex: same as https://github.com/stardust-ui/react/pull/1670#discussion_r305893437

docs/src/components/ComponentPlayground/typeGenerators.ts
```diff
@@ -0,0 +1,80 @@
...
+    return null
+  }
+
+  if (shorthandComponentName === 'Icon') {
```
bmdalex: same as https://github.com/stardust-ui/react/pull/1670#discussion_r305893437

docs/src/examples/components/Chat/Playground.tsx
bmdalex: It can be option
- we have a generated playground (file is present)

and it's just going to render an empty playground, you can even create this as @mnajdova suggests

packages/docs-components/src/knobs/defaultComponents.tsx
```diff
@@ -34,6 +34,16 @@ const KnobBoolean: React.FunctionComponent<KnobComponentProps> = props => (
   />
 )
 
+const KnobNumber: React.FunctionComponent<KnobComponentProps> = props => (
```
bmdalex, line 37: do we really need this? can't we have string, number and boolean together in one `KnobPrimitive`?

packages/docs-components/src/knobs/useSelectKnob.ts
```diff
@@ -1,11 +1,19 @@
 import { UseKnobOptions } from './types'
 import useKnob from './useKnob'
 
-const useSelectKnob = <T extends string>(options: UseKnobOptions<T>) =>
-  useKnob<T>({
+type UseSelectKnobOptions<T extends string> = UseKnobOptions<T> & {
+  allowsNone?: boolean
```
bmdalex, line 5: this is strange. can't we just rely on `undefined` as value?

docs/src/components/ComponentPlayground/ComponentPlayground.tsx
layershifter: This part renders debug info for dev env:
image

Not sure that it's relevant there to use `Alert`...

docs/src/components/ComponentPlayground/ComponentPlaygroundSnippet.tsx
layershifter: Nope, as I know because for React there is no difference: https://stackoverflow.com/a/52211951/6488546

Actually, this component is quite hacky and tricky 😒

packages/docs-components/src/knobs/defaultComponents.tsx
layershifter: Yep, because it renders different components:
- `KnobBoolean` - `input` with `type="checkbox"`, will be replaced with `Checkbox` later
- `KnobNumber` - `input` with `type="number"`, value is parsed with `parseInt()`
- `KnobString` - usual `input`

packages/docs-components/src/knobs/useSelectKnob.ts
layershifter:
```tsx
const [values] = useSelectKnob({ allowsNone: true, values: ['a', 'b'] })
const [values] = useSelectKnob({ values: [undefined, 'b'] })
```

Are you sure that second one is more obvious? 🤔

docs/src/components/ComponentPlayground/propGenerators.tsx
layershifter: For now I want to keep it dumb, even with this relationship because it's easy to understand why exact value was generated. Not all follow ups are implemented yet and don't see all cases, premature refactor there can lead into wrong direction.

docs/src/components/ComponentPlayground/ComponentPlaygroundTemplate.tsx
```diff
@@ -0,0 +1,67 @@
...
+  // @ts-ignore
+  const shouldStretch = props.element && props.element.type.displayName === 'Divider'
+
+  return (
+    <Grid columns="75% 25%" rows="1fr auto" styles={{ gridColumnGap: '1rem' }}>
+      <Segment
+        styles={{
+          display: 'flex',
+          flexDirection: 'column',
+          alignItems: shouldStretch ? 'stretch' : 'center',
```
mnajdova: Wut? Can we avoid this somehow? Maybe for the Divider example we can just add this around it?

docs/src/components/ComponentPlayground/ComponentPlaygroundTemplate.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const ComponentPlaygroundTemplate: React.FunctionComponent<
+  ComponentPlaygroundTemplateProps
+> = props => {
+  // @ts-ignore
+  const shouldStretch = props.element && props.element.type.displayName === 'Divider'
+
+  return (
+    <Grid columns="75% 25%" rows="1fr auto" styles={{ gridColumnGap: '1rem' }}>
+      <Segment
```
mnajdova: Why do we use Segment here?

docs/src/components/ComponentPlayground/propGenerators.tsx
mnajdova: Yeah I would agree that this looks strange... Let's create a separate file to store all these mappings.

docs/src/components/ComponentPlayground/usePlaygroundComponent.tsx
```diff
@@ -0,0 +1,62 @@
...
+  if (process.env.NODE_ENV !== 'production') {
+    if (!componentInfo) {
+      throw new Error(
+        `Cannot find a definition for ${componentName}, please check that "docs/src/componentInfo/${componentName}.info.json" file exists`,
+      )
+    }
+
+    if (!Stardust[componentName]) {
+      throw new Error(
+        `Cannot find an for ${componentName}, please check that it is exported from "@stardust-ui/react"`,
```
mnajdova: Cannot find what?

docs/src/examples/components/Animation/Playground.tsx
```diff
@@ -0,0 +1,2 @@
+// Playground is not supported by this component
+export default null
```
mnajdova: So about this one, can we export some component like `<AutogeneratedPlayground />` that will trigger that we want auto generated playground for the component, and skip it for the components that we don't want to have a playground?

packages/docs-components/src/knobs/useSelectKnob.ts
mnajdova: I would vote for `const [values] = useSelectKnob({ allowsNone: true, values: ['a', 'b'] })` It is more obvious what will be rendered, but that's just me :)

mnajdova (commented):
Functionality looks awesome! Let's review final comments.

docs/src/components/ComponentPlayground/ComponentPlaygroundTemplate.tsx
layershifter: Why not? 🤔 It has a proper look to be used there...

docs/src/components/ComponentPlayground/usePlaygroundComponent.tsx
layershifter: `Cannot find an import`, fixed 👍

docs/src/examples/components/Animation/Playground.tsx
layershifter: We agreed to have a blacklist separately

docs/src/components/ComponentPlayground/ComponentPlaygroundTemplate.tsx
layershifter: _Fixed_ with `fluid` prop, will be removed after will refactor `Divider`'s styles

docs/src/examples/components/Chat/Playground.tsx
layershifter: We agreed with @mnajdova on blacklist, is sounds like the best option

docs/src/components/ComponentPlayground/propGenerators.tsx
layershifter: Done, please check `componentGenerators.ts`

layershifter:
I also had some comments from @mnajdova in offline:

> - `Embed` & `Video` components are oversized

Fixed in https://github.com/stardust-ui/react/pull/1670/commits/f295ca33dc6fad4c489837153149565c1ada5c78 via `variables` usage. The better solution is to have `size` prop on these components

> `Icon` by default doesn't render any icon 

Opened an issue for it #1715, solved via hack now: https://github.com/stardust-ui/react/pull/1670/commits/7ab9a3ea1f37d15a0e0df4b7320d1b308207596f

> Remove `color` prop from `Loader`

Done 👍 

> try to set `mouseLeaveDelay` to `1000` in `Tooltip` - it shrinks the display area of the component itself 

tooltip-bug

This looks more like an issue with `Tooltip` component itself.

alinais (approved):

levithomason:
I'd really love to add a list of examples or presets next!  Here is the original branch that prototyped the playground.  Note the examples list on the left.

image

PR #1677, opened 2019-07-18
chore: accessible flat tree prototype

silviuaavram:
WIP

Purpose is to test the flat-equivalent tree with screen readers.

Not virtualized.


docs/src/examples/components/Tree/Types/TreeExampleExperiment.shorthand.tsx
```diff
@@ -0,0 +1,166 @@
+import * as React from 'react'
```
bmdalex, line 1: this example is way to complex to have it here; I'd suggest making it an actual prototype under `Prototypes/` section

docs/src/examples/components/Tree/Types/TreeExampleExperiment.shorthand.tsx
```diff
@@ -0,0 +1,166 @@
...
+  level?: number
+  size?: number
+  id?: string
+}
+
+type TreeState = {
+  open?: boolean
+}
+
+class Tree extends React.Component<TreeProps, TreeState> {
```
bmdalex, line 17: nit: we started to use more and more `React.FunctionComponent`s and hooks, you can consider implementing this with hooks

docs/src/examples/components/Tree/Types/TreeExampleExperiment.shorthand.tsx
```diff
@@ -0,0 +1,166 @@
...
+                id={`item-${level}-${index}`}
+              />
+            ))}
+        </div>
+      )
+    }
+
+    return (
+      <>
+        <div
```
bmdalex, line 57: consider converting this to `TreeItem` custom component to improve readability of code

Issue #1678, opened 2019-07-18
Padding breaks Icon in Input 


lucivpav:
## Bug Report
Consider an Input component with an icon. When you apply a style with padding to it, the icon is not padded appropriately.

### Steps
```
<Input
    fluid
    icon="search"
    iconPosition="start"
    placeholder="Search"
    style={{ padding: "6px 60px 6px 60px" }}
  />
```
### Expected Result
![image](https://user-images.githubusercontent.com/10930160/61447082-92fc5900-a950-11e9-9cbd-c84a452f59cb.png)

### Actual Result
![image](https://user-images.githubusercontent.com/10930160/61446898-3c8f1a80-a950-11e9-8cb7-b248c6fbe2ca.png)

### Version
0.34.1

### Testcase
https://codesandbox.io/s/stardust-ui-example-g6mbe

mnajdova:
The `icon` in the `Input` component has position `absolute`, because we want it to appear on top of the input. So instead of `padding`, try using `margin` and it should work as expected. Also, always use the `styles` prop instead of `style` when you use Stardust components.

lucivpav:
The same issue happens with `clearable` prop.
![image](https://user-images.githubusercontent.com/10930160/62042033-59073e80-b1fd-11e9-9db0-4e5f43ea49e8.png)

PR #1682, opened 2019-07-18
feat(docs): add search capability

lucivpav:
This PR adds a search input to the top part of sidebar. The search goes through all the below tree's sections, effectively filtering those that match the query. The items are matched in case-insensitive fashion.

search

As you can see, the icon is not rendered at an expected position. This looks like a bug in Input component, reported [here](https://github.com/stardust-ui/react/issues/1678).

The search input is not fully accessible - it does not let the user know, how many items were found (matching the query). This would be solved in a future PR.

You can search by typing even when the element is not focused (provided no other input element is in focus).


lucivpav:
Problem: category of currently selected section cannot be collapsed anymore.

lucivpav:
> Problem: category of currently selected section cannot be collapsed anymore.

Fixed.

lucivpav:
Icon's location has been fixed by manually setting the icon's padding.

image

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -479,10 +434,41 @@ class Sidebar extends React.Component<any, any> {
...
+            </Box>
+          </a>
+          <a href={changeLogUrl} target="_blank" rel="noopener noreferrer" style={topItemTheme}>
+            <Box>
+              CHANGELOG
+              <Icon name="file alternate outline" styles={{ float: 'right' }} />
+            </Box>
+          </a>
+          <Input
+            style={topItemTheme}
```
mnajdova: Please always use `styles` instead of `style` with Stardust components. Also, instead of padding, try applying margin and everything should work. The icon inside the menu is positioned absolutelly, that's why padding is not working: https://codesandbox.io/s/stardust-ui-example-i4e6k

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -38,6 +39,17 @@ class Sidebar extends React.Component<any, any> {
...
   selectedRoute: any
   filteredMenu = componentMenu
+  treeRef: Ref
+  searchInputRef: Ref
+
+  constructor(props) {
+    super(props)
+    this.handleQueryChange = this.handleQueryChange.bind(this)
+    this.findActiveCategoryIndex = this.findActiveCategoryIndex.bind(this)
+    this.setSearchInput = this.setSearchInput.bind(this)
```
mnajdova: Please use arrow functions instead of binding this methods in the constructor.

mnajdova:
I would expect that Concepts, Guides and Prototypes would not be visible in this case
image

docs/src/components/Sidebar/Sidebar.tsx
lucivpav: I think that in this case, the code will look cleaner if I stick to padding.

lucivpav:
> I would expect that Concepts, Guides and Prototypes would not be visible in this case
> image

Good point, fixed.

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -488,10 +445,41 @@ class Sidebar extends React.Component<any, any> {
...
+              <Icon name="github" styles={{ float: 'right' }} />
+            </Box>
+          </a>
+          <a href={changeLogUrl} target="_blank" rel="noopener noreferrer" style={topItemTheme}>
+            <Box>
+              CHANGELOG
+              <Icon name="file alternate outline" styles={{ float: 'right' }} />
+            </Box>
+          </a>
+          <Input
```
mnajdova, line 457: I think it would look better if we use the clearable input with the search icon positioned in the end. What do you think @lucivpav

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -38,6 +39,14 @@ class Sidebar extends React.Component<any, any> {
   _searchInput: any
   selectedRoute: any
   filteredMenu = componentMenu
+  treeRef: Ref
+  searchInputRef: Ref
```
layershifter: Oops, it should be

```suggestion
  searchInputRef: React.RefObject<HTMLElement>
```

However, you don't need `constructor()` 🐱 
```suggestion
  searchInputRef: React.createRef()
```

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -488,10 +445,41 @@ class Sidebar extends React.Component<any, any> {
...
+          <Input
+            styles={topItemTheme}
+            fluid
+            icon={{ name: 'search', style: { padding: topItemTheme.padding } }}
+            placeholder="Search"
+            iconPosition="start"
+            role="search"
+            onChange={this.handleQueryChange}
+            value={this.state.query}
+            ref={this.searchInputRef}
```
```suggestion
            inputRef={this.searchInputRef}
```

We have `inputRef`, so we can simplify it

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -52,9 +61,14 @@ class Sidebar extends React.Component<any, any> {
     document.removeEventListener('keydown', this.handleDocumentKeyDown)
   }
 
-  setSearchInput() {
-    // TODO: Replace findDOMNode with Ref component when it will be merged
-    this._searchInput = (findDOMNode(this) as any).querySelector('.ui.input input')
+  setSearchInput = () => {
+    this._searchInput = this.searchInputRef.current.inputRef.current
+  }
```
layershifter: You don't need this, check this one: https://github.com/stardust-ui/react/pull/1682/files#r307795621

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -66,67 +80,17 @@ class Sidebar extends React.Component<any, any> {
     if (!hasModifier && isAZ && bodyHasFocus) this._searchInput.focus()
```
```suggestion
    if (!hasModifier && isAZ && bodyHasFocus) this._searchInput.current.focus()
```

https://github.com/stardust-ui/react/pull/1682/files#r307795621

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -63,69 +64,23 @@ class Sidebar extends React.Component<any, any> {
...
-          as={NavLink}
-          to={getComponentPathname(info)}
-        >
-          {info.displayName}
-          {isSelected && selectedItemLabel}
-        </Menu.Item>
-      )
-    })
-*/
+  treeActiveIndexChanged = data => {
```
mnajdova: Usually these kind of callbacks sends two params here:
- event
- all props

So this one should be refactored to something like:
```
treeActiveIndexChange = (e, props) => {
 this.setState({activeCategoryIndex: props.activeIndex})
}
```

mnajdova: Take a look on the `onMenuOpenChange` in the `MenuItem` component.

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -488,10 +433,42 @@ class Sidebar extends React.Component<any, any> {
...
+            value={this.state.query}
+            inputRef={this.searchInputRef}
+          />
+        </Flex>
         <Tree
-          defaultActiveIndex={activeCategoryIndex}
           items={allSections}
           renderItemTitle={titleRenderer}
+          activeIndex={this.state.activeCategoryIndex}
+          onActiveIndexChange={this.treeActiveIndexChanged}
```
mnajdova, line 474: Let's add the method in the Tree in this PR, so we can test it with the real implementation. Then we can extract it to a separate PR

lucivpav: Ok. Please take a look: https://github.com/stardust-ui/react/pull/1728

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -52,6 +53,13 @@ export interface TreeProps extends UIComponentProps, ChildrenComponentProps {
    * @param {ReactNode|ReactNodeArray} children - The computed children for this slot.
    */
   renderItemTitle?: ShorthandRenderFunction
+
+  /** Called when activeIndex changes.
+   *
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props and proposed value.
+   */
+  onActiveIndexChange?: ComponentEventHandler<TreeProps & { activeIndex: number[] | number }>
```
mnajdova:
```suggestion
  onActiveIndexChange?: ComponentEventHandler<TreeProps>
```
The `TreeProps` interface already contains the `activeIndex`

docs/src/components/Sidebar/Sidebar.tsx
lucivpav: I moved the icon to the right side, because if we have the input `clearable`, it looks bad on the left side.

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -288,79 +244,24 @@ class Sidebar extends React.Component<any, any> {
...
-      fontWeight: 700,
-    }
-
-    const flexDisplayStyle: any = { width: `${parseInt(this.props.width, 10) - 50}px` }
-
-    const changeLogUrl: string = `${constants.repoURL}/blob/master/CHANGELOG.md`
+  handleQueryChange = e => {
+    const query = e.target.value ? e.target.value : ''
+    this.setState({ query })
+  }
```
layershifter, line 247:
```jsx
handleQueryChange = (e, data) => {
  this.setState({ query: data.value })
}
```

We have `value` in the second param on `Input`

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -34,27 +34,28 @@ class Sidebar extends React.Component<any, any> {
     history: PropTypes.object.isRequired,
     style: PropTypes.object,
   }
-  state: any = { query: '' }
+  state: any = { query: '', activeCategoryIndex: 0 }
   _searchInput: any
   selectedRoute: any
   filteredMenu = componentMenu
+  searchInputRef: React.RefObject<HTMLInputElement> = React.createRef()
```
```suggestion
  searchInputRef = React.createRef<HTMLInputElement>()
```

nit:

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -34,27 +34,28 @@ class Sidebar extends React.Component<any, any> {
...
 
   componentDidMount() {
     document.addEventListener('keydown', this.handleDocumentKeyDown)
-    this.setSearchInput()
-  }
 
-  componentDidUpdate() {
-    this.setSearchInput()
+    const at = window.location.href.replace(window.location.origin, '')
```
layershifter: Can we use `this.props.location.pathname`?

`Sidebar` is wrapped with `withRouter()`:
https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/withRouter.md

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -63,69 +64,23 @@ class Sidebar extends React.Component<any, any> {
...
-  handleItemClick = () => {
+  handleItemClick = e => {
     const { query } = this.state
 
-    if (query) this.setState({ query: '' })
-    // TODO: as part of search input re-enabling
-    // if (document.activeElement === this._searchInput) this._searchInput.blur()
+    if (query) {
+      this.setState({ query: '' })
+      const at = e.target.href.replace(e.target.baseURI, '/')
```
layershifter: We can get value from `data` somehow?

```tsx
handleItemClick = (e: React.SyntheticEvent, data: TreeItemProps) => {
  const at = data.title.to // ???
}
```

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -63,69 +64,23 @@ class Sidebar extends React.Component<any, any> {
...
-          as={NavLink}
-          to={getComponentPathname(info)}
-        >
-          {info.displayName}
-          {isSelected && selectedItemLabel}
-        </Menu.Item>
-      )
-    })
-*/
+  treeActiveIndexChanged = (e, props) => {
```
```suggestion
  treeActiveIndexChanged = (e: React.SyntheticEvent, props: TreeItemProps) => {
```

Can we use types?

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -449,31 +350,78 @@ class Sidebar extends React.Component<any, any> {
...
+    const allSections = _.filter(
+      allSectionsWithPossibleEmptySections,
+      (section: ShorthandValue<TreeItemProps>) => (section as any).items.length > 0,
+    )
+
+    if (this.state.query !== '') {
+      // open all sections
+      _.forEach(allSections, (section: ShorthandValue<TreeItemProps>) => {
+        ;(section as any).open = true
+      })
```
layershifter, line 400: Can we avoid mutation?

```tsx
allSections = _.map(allSections, (section: ShorthandValue<TreeItemProps>) => {
  return { ...section, open: true }
})
```

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -449,31 +350,78 @@ class Sidebar extends React.Component<any, any> {
...
+    const changeLogUrl: string = `${constants.repoURL}/blob/master/CHANGELOG.md`
+    const allSectionsWithoutSearchFilter = this.getSectionsWithoutSearchFilter()
+
+    const allSectionsWithPossibleEmptySections = _.map(
+      allSectionsWithoutSearchFilter,
       (section: ShorthandValue<TreeItemProps>) => {
-        return _.find((section as any).items, item => item.title.to === at)
+        if ((section as any).items) {
+          ;(section as any).items = _.filter((section as any).items, item =>
+            item.title.content.toLowerCase().includes(this.state.query.toLowerCase()),
```
layershifter: Use regexp instead?

```tsx
const escapedQuery = _.escapeRegExp(this.state.query)
const regexQuery = new RegExp(`^${escapedQuery}`, 'i')
```

```tsx
(section: ShorthandValue<TreeItemProps>) => {
  return {
    ...section,
    items: _.filter(section.items, item => regexQuery.test(item.title)) // _.filter works with undefined
  }
}
```

layershifter (approved):

PR #1683, opened 2019-07-18
fix(Checkbox): fix gap styling in RTL

### LINKED ISSUE #1676: No space between label and checkbox in RTL
## Bug Report
The checkbox component (toggle=true) has incorrect layout in RTL.

### Steps
1. Go to https://stardust-ui.github.io/react/components/checkbox
2. Scroll to the toggle example
3. Click RTL.

Expected: There should be some space between the label and the control.
Actual: There is no space.

### Version
0.34.1

layershifter:
Fixes #1676.

----

## Problem

We need to have a gap between `label` and `icon` slots. There are three options to solve this issue.

#### Flex-like approach

```tsx
'> *:not(:last-child)': {
  marginRight: v.checkboxGap,
},
```

**CONS:** `marginRight` will be applied in unpredictable way and will override `margin` provided by user on slots.

#### Element approach

Create a new element and specify a margin/width on it.

**CONS:** We will have a useless element in markup.

#### Grid approach

CSS Grid has a native support for a gap. There are issues with support in IE11, but all of them are resolved via prefixed properties. Approach with a gap support was taken from `autoprefixer`.

✅ Winner

### IE11

image

### Firefox

image

### Chrome

image

## Layout fix

Using Grid has more one benefit, it stratches columsn and it is actually fixes the common use case.

```tsx
const CheckboxExample = () => (
  <div
    style={{
      display: 'flex',
      flexDirection: 'column',
      margin: 10,
      width: '300px',
      border: '2px dotted orange',
      padding: 10,
    }}
  >
    <Checkbox label="Make my profile visible" labelPosition="start" />
    <Checkbox label="I agree with terms" labelPosition="start" />
  </div>
)
```

### `0.34.1`

image

### Current PR

image


packages/react/src/themes/base/components/Checkbox/checkboxStyles.ts
```diff
@@ -7,7 +7,14 @@ const checkboxStyles: ComponentSlotStylesInput<CheckboxProps & CheckboxState, Ch
     alignItems: 'center',
     position: 'relative',
 
-    display: 'inline-flex',
+    // CSS Grid is polifilled only with latest inline-style-prefixer
```
kuzhelov, line 10: could you, please, provide exact version you are implying here?

packages/react/src/themes/base/components/Checkbox/checkboxStyles.ts
```diff
@@ -45,15 +55,15 @@ const checkboxStyles: ComponentSlotStylesInput<CheckboxProps & CheckboxState, Ch
...
 
-  label: ({ props: p, variables: v }): ICSSInJSStyle => ({
-    [p.labelPosition === 'start' ? ':after' : ':before']: {
-      content: '" "',
-      display: 'inline-block',
-      width: v.checkboxGap,
-    },
+  label: ({ props: p }): ICSSInJSStyle => ({
+    '-ms-grid-column': p.labelPosition === 'start' ? 1 : 3,
+    display: 'block', // IE11: should be forced `block`, `inline-block` will be broken
```
kuzhelov: nit:
> IE11: should be forced to be `block`, as `inline-block` is not supported

kuzhelov (approved):

packages/react/src/themes/base/components/Checkbox/checkboxStyles.ts
layershifter: I created an issue #1686, so it will be tracked. We use `5.0.3`, but it works from `5.1.0`: https://github.com/weserio/inline-style-prefixer/blob/master/CHANGELOG.md#510

alinais (approved):

PR #1684, opened 2019-07-19
feat(Input): add `inverted` prop to change background color of an input

codepretty:
Need to be able to have 2 different background options for inputs.

New prop `inverted` for changing the background color of the input
`<input content="i am an input" inverted />`

image

image

image

---

**Question:** This works for Teams theme scenarios, but not sure if the `inverted` prop is the best way to do this... thoughts?

**Conclusion:** After speaking with Design, we decided that `inverted` is the best solution we could come up with for now.


bcalvery:
I wish there was a way for input to just know it's on a background that matches/clashes with the default background and automagically use an alternate based on the color scheme.
I think "inverted' is a bit of a misnomer because just reading it I would think "normal" is black-on-white - so "inverted" must mean white-on-black (specifically the text/background relationship). It's more like "enhancedBackgroundContrast" or "increasedBackgroundContrast"

What is making the parent have that same background color? Is there a property somewhere higher up that says this is a "callout" or something that can be used to infer that the background matches and then relevant form elements can respond - I can imagine a checkbox on that alternative background is going to blend in and might also need to have certain colors react appropriately.

Sorry I don't really have a helpful thought of how to actually solve this though.

mnajdova:
The important things is how Design is describing this. Do they have some design token for this input? Inverted seems reasonable to me, if developers see some design and there is stated that inverted Input should be used it will be much easier to reason about, rather then seeing the background and deciding what should the background of the input be. So I am okay with this proposal..

docs/src/examples/components/Toolbar/Content/ToolbarExamplePopup.shorthand.tsx
```diff
@@ -37,13 +12,7 @@ const ToolbarExamplePopupShorthand = () => {
...
           active: highlightOpen,
           popup: {
-            content: (
-              <HighlightPopup
-                onConfirm={() => {
-                  setHighlightOpen(false)
-                }}
-              />
-            ),
+            content: <p>This is a popup</p>,
```
mnajdova: Why did we change these examples? I am pretty sure they were meaningful exactly because of the inputs inside them.

packages/react/src/themes/teams-high-contrast/components/Input/inputVariables.ts
```diff
@@ -1,12 +1,10 @@
 import { InputVariables } from '../../../base/components/Input/inputVariables'
+import { pxToRem } from '../../../../lib'
 
 export default (siteVars: any): Partial<InputVariables> => {
   return {
-    backgroundColor: siteVars.bodyBackground,
```
mnajdova, line 5: 👍 for removing the overrides

miroslavstastny (approved):
Technically ok, make sure all the screener regressions are expected.

PR #1687, opened 2019-07-19
feat(Chat): add Control messages prototype

mnajdova:
This PR adds prototype for the Control messages that can be used in the Chat. They are based on the Tree* components, with some custom overrides on the behaviors. The styling is not done currently, we want to focus on the accessibility first and test all scenarios.


docs/src/prototypes/chatMessages/ControlMessages/index.tsx
```diff
@@ -0,0 +1,178 @@
...
+        },
+        content: (
+          <Tree
+            data-is-focusable="true"
+            accessibility={overridenTreeBehavior}
+            items={[
+              {
+                onKeyDown: e => {
+                  // Without this override, the focus zone for the tree title is not working..
+                  console.log('Tree item 1 key down')
```
mnajdova: The inner Focus zone doesn't work if I remove this event handler.. I tried just adding data-is-focusable, tabIndex: 0, but these two didn't help..

docs/src/prototypes/chatMessages/ControlMessages/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+          const eventCode = keyboardKey.getCode(e)
+          if (eventCode === keyboardKey.Enter || eventCode === keyboardKey.Spacebar) {
+            setOpen(true)
+          }
+          if (eventCode === keyboardKey.Escape) {
+            setOpen(false)
+          }
+        },
+        content: (
+          <Tree
```
lucivpav, line 113: Looking at the chat example, it was not clear to me that this is actually an expansible tree.

docs/src/prototypes/chatMessages/ControlMessages/index.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+const janeAvatar = {
+  image: 'public/images/avatar/small/ade.jpg',
+  status: {
+    color: 'green',
+    icon: 'check',
+  },
+}
+
+const treeItems = [
```
lucivpav, line 24: These two items are almost identical. How about extracting the code into a function?

docs/src/prototypes/chatMessages/ControlMessages/index.tsx
mnajdova: The styling are not implemented for this prototype, we are focusing only on the behavior and whether everything works as expected

docs/src/prototypes/chatMessages/ControlMessages/index.tsx
mnajdova: will refactor it once we decide on the final implementation

jdhuntington:

PR #1688, opened 2019-07-19
fix(Toolbar): close menu when it looses focus

### LINKED ISSUE #1624: Toolbar menu does not close when it loses focus
### Steps
Pressing Tab or Shift+Tab in the toolbar menu does not close it

### Expected Result
When focus is moved outside of the menu within toolbar, it should close

### Actual Result
It does not close 💣

### Version
0.34.1

sophieH29:
Fixes https://github.com/stardust-ui/react/issues/1624


jurokapsiar:
I have tested this - tab correctly closes the menu, but shift+tab focuses the triggering button and the menu is not closed

packages/react/src/lib/accessibility/Behaviors/Toolbar/toolbarItemBehavior.ts
```diff
@@ -22,7 +22,10 @@ const toolbarItemBehavior: Accessibility<ToolbarItemBehaviorProps> = props => {
   behaviorData.keyActions.wrapper = {
     ...behaviorData.keyActions.wrapper,
     closeMenuAndFocusTrigger: {
-      keyCombinations: props.menu && props.menuOpen ? [{ keyCode: keyboardKey.Escape }] : null,
+      keyCombinations:
+        props.menu && props.menuOpen
+          ? [{ keyCode: keyboardKey.Escape }, { keyCode: keyboardKey.Tab, shiftKey: true }]
```
sophieH29: handling `shift+tab` key, so when user press it, focus will go to the prev tabbable element before toolbar, not the trigger button of the menu inside toolbar

mnajdova: Just to be sure, `tab` should go to the next focusable element after the toolbar, and `shift+tab` should go to the prev focusable element before the toolbar? `Escape` should focus the `trigger`?

packages/react/src/lib/accessibility/Behaviors/Toolbar/toolbarItemBehavior.ts
```diff
@@ -22,7 +22,10 @@ const toolbarItemBehavior: Accessibility<ToolbarItemBehaviorProps> = props => {
   behaviorData.keyActions.wrapper = {
     ...behaviorData.keyActions.wrapper,
     closeMenuAndFocusTrigger: {
```
mnajdova, line 24: A bit worried that the name here implies close the menu and **focus the trigger**, but actually the combination `{ keyCode: keyboardKey.Tab, shiftKey: true }` is not focusing the trigger :\

packages/react/src/lib/accessibility/Behaviors/Toolbar/toolbarItemBehavior.ts
sophieH29: yes, it all works. Why only `shift+tab`? - cause trigger is rendered before the menu and trigger has `tabindex=0` when menu is opened. So by default when pressing `shift+tab` - focus goes to to trigger from menu. But need to be at the previous tabbable element

packages/react/src/lib/accessibility/Behaviors/Toolbar/toolbarItemBehavior.ts
sophieH29: makes sense, @mnajdova. Created a separate action handler to handle that

jurokapsiar (approved):

mnajdova (approved):

mnajdova:
Actually, with the latest changes, shift tab is focusing the trigger element :\ Can you check this again

PR #1689, opened 2019-07-19
fix(Popup): fix broken background with pointing and FocusTrap

### LINKED ISSUE #1679: Teams theme: Popup pointing styles still broken with focus trap
## Bug Report

After fix in #1433 Popup pointing with focus trap is still broken:

image

### Version
0.34.1

### Testcase
https://codesandbox.io/s/popup-pointing-with-focustrap-broken-styles-8kenr

layershifter:
Fixes #1679.

## Problem

In `master` we have already resolved this issue by #1565, however we need to provide a fix without breaking changes for now.

### Before

image

### After

image

#### Test snippet

```tsx
import { Button, Flex, Popup, popupFocusTrapBehavior } from '@stardust-ui/react'

const plainContentStyle = {
  zIndex: 1000,
  padding: 5,
}

const PopupExample = () => (
  <Flex vAlign='center' space='between' styles={{ padding: '50px', height: '500px', backgroundColor: 'orange' }}>
    <Popup content="Hello from popup!" pointing open accessibility={popupFocusTrapBehavior}>
      <Button icon="expand"/>
    </Popup>

    <Popup open content={<p style={plainContentStyle}>Plain popup content rendered 'as is'.</p>}>
      <Button icon="expand" content="Popup with plain content"/>
    </Popup>
    <Popup open accessibility={popupFocusTrapBehavior}
           content={<p style={plainContentStyle}>Plain popup content rendered 'as is'.</p>}>
      <Button icon="expand" content="Popup with plain content"/>
    </Popup>

    <Popup
      open
      content={{ content: <p style={plainContentStyle}>Popup content rendered in wrapper.</p> }}
    >
      <Button icon="expand" content="Popup with wrapped content"/>
    </Popup>
  </Flex>
)

export default PopupExample
```

### DIFF
--- a/packages/react/src/themes/teams/components/Popup/popupStyles.ts
+++ b/packages/react/src/themes/teams/components/Popup/popupStyles.ts
@@ -1,3 +1,4 @@
+import * as _ from 'lodash'
 import * as React from 'react'
 import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
 import { PopupProps } from '../../../../components/Popup/Popup'
@@ -14,7 +15,7 @@ const popupStyles: ComponentSlotStylesInput<PopupProps, PopupVariables> = {
     /*
      * This fix handles all cases with wrapped focus trap.
      */
-    ...(React.isValidElement(p.content) && {
+    ...((React.isValidElement(p.content) || _.isFunction(p.content)) && {
       // when FocusTrap exists and `context` is JSX element
       color: v.contentColor,
       background: v.contentBackgroundColor,
--- a/packages/react/src/themes/teams/components/Popup/popupStyles.ts
+++ b/packages/react/src/themes/teams/components/Popup/popupStyles.ts
@@ -1,3 +1,4 @@
+import * as React from 'react'
 import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
 import { PopupProps } from '../../../../components/Popup/Popup'
 import { PopupVariables } from './popupVariables'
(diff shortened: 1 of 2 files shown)


packages/react/src/themes/teams/components/Popup/popupStyles.ts
```diff
@@ -1,16 +1,34 @@
...
     zIndex: v.zIndex,
     position: 'absolute',
     textAlign: 'left',
-    color: v.contentColor,
-    background: v.contentBackgroundColor,
+
+    /*
+     * This fix handles all cases with wrapped focus trap.
+     */
+    ...(React.isValidElement(p.content) && {
```
kuzhelov: actually, this should also apply in case if render function is provided as an argument :( as in that case slot component will be also replaced. So, I think that we may just check that this is not an object or primitive

kuzhelov (approved):

layershifter:
Adjusted to handle `render()` callback cases:

```tsx
    <Popup open trigger={<Button icon="expand"/>} content={render => render(
      {},
      (ComponentType, props) => (
        <p style={{ zIndex: 1000, padding: 5 }}>Plain popup content rendered 'as is'.</p>
      ),
    )}/>
    <Popup open trigger={<Button icon="expand"/>} pointing content={render => render(
      {},
      (ComponentType, props) => (
        <div {...props} style={{ zIndex: 1000, padding: 5}}>
          Plain popup content rendered 'as is'.
        </div>
      ),
    )}/>
```

image

PR #1690, opened 2019-07-19
feat(renderComponent): remove `focusZoneRef`

### LINKED ISSUE #1569: Remove `focusZoneRef` from `renderComponent`
`focusZoneRef` is exposed in UI component and is passed all the way through `renderComponent` and then `createComponent` functionality, which is only used in one place - `Chat` component, to focus a message item when pressing ESC inside chat message.

TODO:
- remove `focusZoneRef` from all the places
- move key action `focus` from `chatBehavior` to `chatMessageBehavior`
- implement an action handler in `chatMessageBehavior` to focus chat message on ESC

sophieH29:
Closes https://github.com/stardust-ui/react/issues/1569


packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -142,12 +143,24 @@ class ChatMessage extends UIComponent<WithAsProp<ChatMessageProps>, ChatMessageS
...
     preventDefault: event => {
       event.preventDefault()
     },
+
+    focus: event => {
+      if (this.messageRef) {
+        this.messageRef.current.focus()
+
+        if (document.activeElement === this.messageRef.current) {
+          event.stopPropagation()
```
jurokapsiar: why is this needed? can you maybe add a comment in the code as well as it is not so obvious.

layershifter: Yep, this combination looks a bit strange... Is `.focus()` async?

sophieH29: it's more about stop propagation only if focus was set to chat message successfully, for e.g. if chat message won't have tabindex and it is a `div`, focus will go to body. If we allow propagation, most probably it can be handled by parent components.

Ok, discussed with Shift, this is very rare case. I'll just remove it to avoid concerns

layershifter (approved):

PR #1691, opened 2019-07-19
fix(factories): fix key handling on React.Element

layershifter:
Fixes a regression from #1600.

### Problem

Previously we picked `key` always from React Element, but it's wrong because each passed value to `key` prop will be stringified.

Current behavior results into errors:
```
Warning: Encountered two children with the same key, `null`.
Keys should be unique so that components maintain their identity across updates.
Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
```

image


alinais (approved):

packages/react/src/lib/factories.ts
```diff
@@ -263,7 +263,14 @@ function createShorthandFromValue({
 
     if (valIsReactElement) {
       // use the key from React Element
-      props.key = (value as React.ReactElement).key
+      const elementKey = (value as React.ReactElement).key
+      // <div /> - key is not passed as will be `null`
+      // <div key={null} /> - key is passed as `null` and will be stringified
+      const isNullKey = elementKey === 'null' || elementKey === null
```
kuzhelov: afraid that there may be legit case when user would want to provide `null` as a string key for an element

kuzhelov (approved):

packages/react/src/lib/factories.ts
layershifter: Valid point 👍 
Fixed

PR #1696, opened 2019-07-22
feat(MenuButton): add component

### LINKED ISSUE #345: [RFC] ContextMenu Component

jurokapsiar:
Allows opening a "dropdown" menu on left/right click, hover or focus.
Reuses Popup, adds a Menu to its content.

Fixes #345


packages/react/src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,234 @@
...
+import { contextMenuBehavior } from '../../lib/accessibility'
+import { focusMenuItem, focusNearest } from './focusUtils'
+import { ALIGNMENTS, POSITIONS } from '../../lib/positioner'
+
+export interface ContextMenuSlotClassNames {
+  menu: string
+}
+
+export interface ContextMenuProps
+  extends Omit<PopupProps, 'content' | 'renderContent' | 'children' | 'trapFocus'> {
```
jurokapsiar: Do we want to extend or do we want to put all props here explicitly?

packages/react/src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,234 @@
...
+    menu: `${ContextMenu.className}__menu`,
+  }
+
+  static propTypes = {
+    ...commonPropTypes.createCommon({
+      animated: false,
+      as: true,
+      content: false,
+    }),
+    align: PropTypes.oneOf(ALIGNMENTS),
```
jurokapsiar: Do we want to put all Popup props here explicitly? If yes, should we have a unit test that makes sure these two are in sync?

packages/react/src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+  static defaultProps: ContextMenuProps = {
+    accessibility: contextMenuBehavior,
+    align: 'start', // top
+    position: 'below', // after
+  }
+
+  static autoControlledProps = ['open']
+
+  // TODO: this does not persist generated ids across re-renders, see ContextMenu-test.tsx
```
jurokapsiar: @sophieH29 when I use `getOrGenerateIdFromShorthand` here, each re-render will generate new ID. Is that expected or is it a bug?

packages/react/src/components/ContextMenu/ContextMenu.tsx
```diff
@@ -0,0 +1,234 @@
...
+        <Ref innerRef={this.triggerRef}>
+          <Popup
+            {...popupProps}
+            accessibility={() => accessibility}
+            open={this.state.open}
+            onOpenChange={this.handleOpenChange}
+            inline
+            autoFocus
+            content={{
+              variables: { padding: '', borderSize: '0px' }, // TODO: this should probably be in the context menu variables?
```
jurokapsiar: I don't like this, hoping to find a better solution - for example make these two a variable of the ContextMenu itself?

packages/react/src/lib/accessibility/Behaviors/ContextMenu/contextMenuBehavior.ts
```diff
@@ -0,0 +1,67 @@
+import * as _ from 'lodash'
+import * as keyboardKey from 'keyboard-key'
+import { Accessibility } from '../../types'
+import popupBehavior, { PopupBehaviorProps } from '../Popup/popupBehavior'
+
+const contextMenuBehavior: Accessibility<ContextMenuBehaviorProps> = props => {
```
jurokapsiar: spec section needs to be added

packages/react/src/components/ContextMenu/ContextMenu.tsx
jurokapsiar: override styles, create them for menu slot, pass them down to popup and popup content

packages/react/src/components/ContextMenu/ContextMenu.tsx
jurokapsiar: after discussion, all props are here explicitly

docs/src/examples/components/MenuButton/Rtl/MenuButtonExample.rtl.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react'
+import { Button, MenuButton } from '@stardust-ui/react'
+
+const ContextMenuExampleRtl = () => (
+  <MenuButton
+    trigger={<Button content="ا يجلبه إلينا الأس" />}
+    menu={{ items: ['English text!', 'غالباً ونرفض الشعور'] }}
```
layershifter:
```suggestion
    menu={['English text!', 'غالباً ونرفض الشعور']}
```

docs/src/examples/components/MenuButton/Rtl/MenuButtonExample.rtl.tsx
```diff
@@ -0,0 +1,11 @@
...
+  />
+)
+
+export default ContextMenuExampleRtl
```
layershifter:
```suggestion
export default MenuButtonExampleRtl
```

docs/src/examples/components/MenuButton/Usage/MenuButtonExampleControlled.shorthand.tsx
```diff
@@ -0,0 +1,20 @@
...
+    <MenuButton
+      open={open}
+      onOpenChange={(e, { open }) => setOpen(open)}
+      trigger={<Button icon="expand" title="Open MenuButton" />}
+      menu={{ items }}
+    />
+  )
+}
+
+export default MenuButtonControlledExample
```
layershifter:
```suggestion
export default MenuButtonOpenExample
```

layershifter: Please move this example under `State` section because it represents `Open` state.

docs/src/examples/components/MenuButton/Usage/MenuButtonExampleControlled.shorthand.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+const MenuButtonControlledExample = () => {
+  const [open, setOpen] = useBooleanKnob({ name: 'open', initialValue: true })
+
+  return (
+    <MenuButton
+      open={open}
+      onOpenChange={(e, { open }) => setOpen(open)}
+      trigger={<Button icon="expand" title="Open MenuButton" />}
+      menu={{ items }}
```
layershifter:
```suggestion
      menu={['1', '2', '3', { content: 'submenu', menu: { items: ['4', '5'] } }]}
```

layershifter: Just inline items, they are pretty simple

docs/src/examples/components/MenuButton/Usage/MenuButtonExampleOnElement.shorthand.tsx
```diff
@@ -0,0 +1,33 @@
...
+    return (
+      <>
+        <MenuButton
+          trigger={
+            <div style={{ padding: '4rem', border: 'red dashed' }}>
+              <Button content="Random button" onClick={this.showAlert} />
+            </div>
+          }
+          shouldTriggerBeTabbable={false}
+          menu={{ items }}
```
layershifter: Inline items, please

docs/src/examples/components/MenuButton/Usage/index.tsx
```diff
@@ -0,0 +1,26 @@
...
+    />
+    <ComponentExample
+      title="Context menu"
+      description="A menu button can be attached to any element to create a context menu."
+      examplePath="components/MenuButton/Usage/MenuButtonExampleOnElement"
+    />
+    <ComponentExample
+      title="Controlled"
+      description="Note that if MenuButton is controlled, then its 'open' prop value could be changed either by parent component, or by user actions (e.g. key press) - thus it is necessary to handle 'onOpenChange' event."
+      examplePath="components/MenuButton/Usage/MenuButtonExampleControlled"
```
layershifter: https://github.com/stardust-ui/react/pull/1696/files#r310158939

docs/src/examples/components/MenuButton/Usage/index.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react'
+
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
+
+const Usage = () => (
+  <ExampleSection title="Usage">
+    <ComponentExample
+      title="Triggering menu on different actions"
+      description="A context menu can be triggered on click, hover or focus."
```
layershifter: Example also contains `context` usage, BTW.

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+  getOrGenerateIdFromShorthand,
+  commonPropTypes,
+  StyledComponentProps,
+} from '../../lib'
+import { ShorthandValue, ComponentEventHandler } from '../../types'
+
+import { Accessibility } from '../../lib/accessibility/types'
+import { createShorthandFactory } from '../../lib/factories'
+import Popup, { PopupProps, PopupEvents, PopupEventsArray } from '../Popup/Popup'
+import { Menu, MenuItemProps, MenuProps, Ref } from '../..'
```
layershifter: Please use direct imports

```suggestion
import { Menu, MenuItemProps, MenuProps } from '../Menu/Menu'
import { Ref } from '@stardust-ui/react-component-ref'
```

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+  target?: HTMLElement
+
+  /** Element to be rendered in-place where the popup is defined. */
+  trigger?: JSX.Element
+
+  /** Whether the trigger should be tabbable */
+  shouldTriggerBeTabbable?: boolean
+
+  /** Ref for Popup content DOM node. */
+  contentRef?: React.Ref<HTMLElement>
```
layershifter: ❗️ Why this is exposed?

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+
+  /** Shorthand for menu configuration */
+  menu?: ShorthandValue<MenuProps>
```
layershifter:
```suggestion
  menu?: ShorthandValue<MenuProps> | ShorthandCollection<MenuItemProps>
```

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+
+  /** Determines if the MenuButton behaves as context menu */
+  contextMenu?: boolean
```
layershifter, line 83: Why it's required? Why we can't use `const isContextMenu = this.props.on === 'content'`?

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+  triggerId: string
+  autoFocus: boolean
+}
+
+/**
+ * A MenuButton displays a menu connected to trigger element.
+ * @accessibility
+ */
+export default class MenuButton extends AutoControlledComponent<
+  ContextMenuProps,
```
layershifter:
```suggestion
  MenuButtonProps,
```

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+  ContextMenuState
```
layershifter:
```suggestion
  MenuButtonState
```

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+> {
+  static displayName = 'MenuButton'
+
+  static className = 'ui-menubutton'
+
+  static create: Function
+
+  static slotClassNames: ContextMenuSlotClassNames = {
```
layershifter:
```suggestion
  static slotClassNames: MenuButtonSlotClassNames = {
```

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+    unstable_pinned: PropTypes.bool,
+    contentRef: customPropTypes.ref,
+    menu: customPropTypes.itemShorthandWithoutJSX,
+    contextMenu: PropTypes.bool,
+  }
+
+  static defaultProps: ContextMenuProps = {
+    accessibility: contextMenuBehavior,
+    align: 'start', // top
+    position: 'below', // after
```
layershifter:
```suggestion
    position: 'below',
```

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+    shouldTriggerBeTabbable: PropTypes.bool,
+    unstable_pinned: PropTypes.bool,
+    contentRef: customPropTypes.ref,
+    menu: customPropTypes.itemShorthandWithoutJSX,
+    contextMenu: PropTypes.bool,
+  }
+
+  static defaultProps: ContextMenuProps = {
+    accessibility: contextMenuBehavior,
+    align: 'start', // top
```
layershifter:
```suggestion
    align: 'start',
```

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+    position: 'below', // after
+  }
+
+  static autoControlledProps = ['open']
+
+  // TODO: this does not persist generated ids across re-renders, see menubutton-test.tsx
```
layershifter: Was TODO completed?

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+      {
+        open: true,
+        autoFocus: false, // focused by renderCallback
+      },
+      renderCallback,
+    )
+    e.preventDefault()
+  }
+
+  handleOpenChange = (e, { open }) => {
```
layershifter:
```suggestion
  handleOpenChange = (e, { open }: PopupProps) => {
```

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+    contextMenu: PropTypes.bool,
+  }
+
+  static defaultProps: ContextMenuProps = {
+    accessibility: contextMenuBehavior,
+    align: 'start', // top
+    position: 'below', // after
+  }
+
+  static autoControlledProps = ['open']
```
layershifter, line 144: Is this component autocontrolled? I don't see any `trySetState()`...

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+      this.setState({ open: false, autoFocus: false })
+    }
+  }
+
+  handleMenuItemOverrides = (menuItemAccessibilityAttributes: any) =>
+    _.map(_.get(this.props.menu, 'items'), (item: ShorthandValue<MenuItemProps>) =>
+      typeof item === 'object'
+        ? {
+            ...item,
+            onClick: this.handleMenuItemClick(item as MenuItemProps),
```
layershifter: Let's add `onItemClick` prop to `Menu` instead, this looks weird...

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+  renderComponent({
+    ElementType,
+    classes,
+    unhandledProps,
+    accessibility,
+    styles,
+  }: RenderResultConfig<ContextMenuProps>): React.ReactNode {
+    const { contextMenu, menu, ...popupProps } = this.props
+    const content =
+      menu &&
```
layershifter:
```suggestion
```

We don't need this check, it will be done by factories

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+      >
+        <Ref innerRef={this.triggerRef}>
+          {Popup.create(popupProps, {
+            defaultProps: {
+              inline: true,
+              autoFocus: true,
+              unstable_pinned: true,
+            },
+            overrideProps: {
+              accessibility: () => accessibility,
```
layershifter:
```suggestion
              accessibility,
```

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,305 @@
...
+    )
+
+  renderComponent({
+    ElementType,
+    classes,
+    unhandledProps,
+    accessibility,
+    styles,
+  }: RenderResultConfig<ContextMenuProps>): React.ReactNode {
+    const { contextMenu, menu, ...popupProps } = this.props
```
layershifter: This spread is not safe, we need to pass down popup props manually

packages/react/src/lib/accessibility/Behaviors/MenuButton/menuButtonBehavior.ts
```diff
@@ -0,0 +1,63 @@
+import * as _ from 'lodash'
+import * as keyboardKey from 'keyboard-key'
+import { Accessibility } from '../../types'
+import popupBehavior, { PopupBehaviorProps } from '../Popup/popupBehavior'
+
+const contextMenuBehavior: Accessibility<ContextMenuBehaviorProps> = props => {
```
layershifter:
```suggestion
const menuButtonBehavior: Accessibility<MenuButtonBehaviorProps> = props => {
```

packages/react/src/lib/accessibility/Behaviors/MenuButton/menuButtonBehavior.ts
```diff
@@ -0,0 +1,63 @@
...
+    },
+  })
+}
+
+export interface ContextMenuBehaviorProps extends PopupBehaviorProps {
+  /** menu id */
+  menuId?: string
+  /** button id */
+  triggerId?: string
+  /** open */
```
layershifter: It is worse to have such comments...

```suggestion
```

packages/react/src/lib/accessibility/index.ts
```diff
@@ -50,3 +50,4 @@ export { default as checkboxBehavior } from './Behaviors/Checkbox/checkboxBehavi
 export { default as tooltipBehavior } from './Behaviors/Tooltip/tooltipBehavior'
 export { default as tooltipAsLabelBehavior } from './Behaviors/Tooltip/tooltipAsLabelBehavior'
 export { default as sliderBehavior } from './Behaviors/Slider/sliderBehavior'
+export { default as contextMenuBehavior } from './Behaviors/MenuButton/menuButtonBehavior'
```
layershifter:
```suggestion
export { default as menuButtonBehavior } from './Behaviors/MenuButton/menuButtonBehavior'
```

packages/react/src/lib/getOrGenerateIdFromShorthand.ts
```diff
@@ -6,20 +6,25 @@ const getOrGenerateIdFromShorthand = <P extends Record<string, any>>(
...
   }
 
-  if (_.isPlainObject(value)) {
-    return (value as Record<string, any>).id
+  if (fallbackToUnique && !result) {
+    result = _.uniqueId(prefix)
   }
 
-  return currentValue || _.uniqueId(prefix)
+  return result
```
layershifter: Why this is needed? Please clarify.

As I see, this condition (`if (fallbackToUnique && !result) {)` can work only if `value` is `nil`...

packages/react/test/specs/components/MenuButton/MenuButton-test.tsx
```diff
@@ -0,0 +1,95 @@
...
+  isConformant(MenuButton)
+
+  describe('accessibility', () => {
+    handlesAccessibility(MenuButton, {
+      defaultRootRole: 'none',
+    })
+
+    describe('onOpenChange', () => {
+      test('is called on click', () => {
+        const spy = jest.fn()
```
layershifter:
```suggestion
        const onOpenChange = jest.fn()
```

packages/react/src/components/MenuButton/MenuButton.tsx
jurokapsiar: no, that is a different use case. contextMenu is for any element where you can click on. on=context is just a regular MenuButton which opens on context. The use case: For reactions, we want the popup appear on hover and on contextmenu so that both screen reader users can activate it (screen readers do not have a concept of hover). On click, an action of selecting/unselecting the reaction should be executed. So it will be something like:
```
<MenuButton
  trigger={reaction}
  menu={...}
  on=['context', 'hover']
  />
```

Whereas contextMenu will be something like 
```
<MenuButton
  trigger={<TreeItem ... />}
  menu={...}
  on=['context']
  />
```

It is currently not documented at all, but I will add the explanation to the examples

packages/react/src/components/MenuButton/MenuButton.tsx
jurokapsiar: yes

packages/react/src/lib/getOrGenerateIdFromShorthand.ts
jurokapsiar: these changes are basically to make sure that an id is generated even if value is an element or shorthand without an ID.

packages/react/src/components/MenuButton/MenuButton.tsx
jurokapsiar: it's because of `TypeError: accessibility is not a function`. I'm not really happy about this construct, but I did not find any other way how to pass the pre-populated behavior to the Popup

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,297 @@
...
+  /**
+   * DOM element that should be used as popup's target - instead of 'trigger' element that is used by default.
+   */
+  target?: HTMLElement
+
+  /** Element to be rendered in-place where the popup is defined. */
+  trigger?: JSX.Element
+
+  /** Whether the trigger should be tabbable */
+  shouldTriggerBeTabbable?: boolean
```
levithomason: We don't follow the boolean prefixing props, like `should*`.  Example, we use `pointing` instead of `isPointing`, etc.

Propose something like `tabbableTrigger` for this instead.  The props are meant to be read together as a natural language string:

"An open tabbable trigger menu button" ==
```jsx
<MenuButton open tabbableTrigger />
```

For contrast, we do not do this:
```jsx
// Counterexample!
<MenuButton isOpen hasTabbleTrigger />
```

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,337 @@
...
+        pointing,
+        position,
+        tabbableTrigger,
+        styles: stylesProp,
+        target,
+        trigger,
+        unstable_pinned,
+        variables,
+      },
+      _.isUndefined,
```
jurokapsiar: @layershifter is this correct? I had to remove props with `undefined` value, otherwise the default props were overriden to `undefined` when not present.

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,337 @@
...
+          trapFocus: true,
+          unstable_pinned: true,
+          tabbableTrigger: false,
+        },
+        overrideProps,
+      })
+    }
+
+    return (
+      <ElementType
```
mnajdova, line 286: Why exactly do we need this additional element around the trigger? We are not adding any  role to it, and it is not used much for styling..

jurokapsiar: We are adding keyboard handlers to it - they apply for both the trigger as well as the inline popup/menu that is inside.

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### BREAKING CHANGES
 - Fix `firstFocusableSelector` in `FocusTrapZone` and `AutoFocusZone` @sophieH29 ([#1732](https://github.com/stardust-ui/react/pull/1732))
+- Rename Popup `shouldTriggerBeTabbable` prop to `tabbleTrigger` @jurokapsiar ([#1696](https://github.com/stardust-ui/react/pull/1696))
```
```suggestion
- Rename Popup `shouldTriggerBeTabbable` prop to `tabbableTrigger` @jurokapsiar ([#1696](https://github.com/stardust-ui/react/pull/1696))
```

docs/src/examples/components/MenuButton/State/MenuButtonExampleControlled.shorthand.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+const MenuButtonControlledExample = () => {
+  const [open, setOpen] = useBooleanKnob({ name: 'open', initialValue: true })
+
+  return (
+    <MenuButton
+      open={open}
+      onOpenChange={(e, { open }) => setOpen(open)}
+      trigger={<Button icon="expand" title="Open MenuButton" />}
+      menu={['1', '2', '3', { content: 'submenu', menu: { items: ['4', '5'] } }]}
```
layershifter:
```suggestion
      menu={['1', '2', '3', { content: 'submenu', menu: ['4', '5'] }]}
```

_nit_

docs/src/examples/components/MenuButton/State/MenuButtonExampleControlled.shorthand.tsx
```diff
@@ -0,0 +1,18 @@
...
+    />
+  )
+}
+
+export default MenuButtonControlledExample
```
layershifter:
```suggestion
export default MenuButtonOpenExample
```

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,331 @@
...
+  accessibility?: Accessibility
+
+  /** Additional CSS class name(s) to apply.  */
+  className?: string
+
+  /** Initial value for 'open'. */
+  defaultOpen?: boolean
+
+  /** Whether the Popup should be rendered inline with the trigger or in the body. */
+  inline?: boolean
```
layershifter: Remove `inline` in favor of `contextMenu`

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  closeMenu() {
+    this.trySetState({
+      open: false,
+      autoFocus: false,
+    })
+  }
+
+  openAndFocus(e: Event, which: 'first' | 'last') {
```
layershifter:
```suggestion
  openAndFocus(e: React.KeyboardEvent, which: 'first' | 'last') {
```

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,331 @@
...
+        open: true,
+        autoFocus: false, // focused by renderCallback
+      },
+      renderCallback,
+    )
+    e.preventDefault()
+  }
+
+  handleOpenChange = (e, { open }: PopupProps) => {
+    _.invoke(this.props, 'onOpenChange', e, { ...this.props, ...{ open } })
```
layershifter:
```suggestion
    _.invoke(this.props, 'onOpenChange', e, { ...this.props, open })
```

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,331 @@
...
+        className={classes.root}
+        {...accessibility.attributes.root}
+        {...unhandledProps}
+        {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.root, unhandledProps)}
+      >
+        <Ref innerRef={this.triggerRef}>
+          {Popup.create(popupProps, {
+            defaultProps: {
+              inline: true,
+              autoFocus: true,
```
layershifter: Please test this, should be `this.state.autoFocus` or `this.state.autoFocus` should be removed

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,331 @@
...
+        styles: styles.popupContent,
+        content: content && <Ref innerRef={this.menuRef}>{content}</Ref>,
+      },
+      children: undefined, // force-reset `children` defined for `Popup` as it collides with the `trigger
+    }
+
+    if (contextMenu) {
+      return Popup.create(popupProps, {
+        defaultProps: {
+          on: 'context',
```
layershifter: Go to override props: `overrideProps`

packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -0,0 +1,331 @@
...
+    )
+
+    const content = Menu.create(menu, {
+      defaultProps: {
+        ...accessibility.attributes.menu,
+        vertical: true,
+      },
+      overrideProps: {
+        onItemClick: (e: React.SyntheticEvent, itemProps: MenuItemProps) =>
+          this.handleMenuItemClick(content)(e, itemProps),
```
layershifter:
```
  handleMenuOverrides = (predefinedProps?: MenuProps) => ({
    onItemClick: (e: React.SyntheticEvent, itemProps: MenuItemProps) => {
      _.invoke(predefinedProps, 'onItemClick', e, itemProps)
      if (!itemProps || !itemProps.menu) {
        // do not close if clicked on item with submenu
        this.trySetState({ open: false, autoFocus: false })
      }
    },
  })
```

packages/react/src/components/MenuButton/MenuButton.tsx
layershifter: Covered offline, no other way now

packages/react/src/lib/getOrGenerateIdFromShorthand.ts
layershifter:
```tsx
import * as _ from 'lodash'
import { ShorthandValue } from '../types'

const getOrGenerateIdFromShorthand = <P extends Record<string, any>>(
  prefix: string,
  value: ShorthandValue<P>,
  currentValue?: string,
): string | undefined => {
  if (_.isNil(value)) {
    return undefined
  }

  let result: string

  if (React.isValidElement(value)) {
    result = (value as React.ReactElement<{ id?: string }>).props.id
  } else if (_.isPlainObject(value)) {
    result = (value as Record<string, any>).id
  }

  return result || currentValue || _.uniqueId(prefix)
}

export default getOrGenerateIdFromShorthand
```

layershifter (approved):

PR #1700, opened 2019-07-22
chore(package): bump lodash

layershifter:
## Security upgrade

`lodash@4.17.11` contains a security vulnerability:
- lodash/lodash#4348
- https://snyk.io/vuln/npm:lodash@4.17.11

I also have to upgrade `snyk` and `syncpack` because they had direct dependency on vulnerable version.


levithomason (approved):
👍

PR #1706, opened 2019-07-23
fix(Dialog): support nested Dialogs and Popups

### LINKED ISSUE #1674: Click inside a child Dialog of a Popup closes the whole React tree
Steps
- Create a Popup with Dialog as a child item. 
- Create a button inside the Dialog

Expected Result
- Click should not be considered as a click 'outside' pop-up since, dialog is a child of the pop-up.

Actual Result
- Dialog and pop-up close with a click inside the dialog because Pop-up doesn't register the dialog as a child hence, during the calculations to find whether the click was inside of the children returns false and pop-up closes. This results in closing of the whole React tree since, for React it is infact the child.
- Issue doesn't repro with Popup as a child of a Popup.

Example:
- CodeSandbox link: https://codesandbox.io/embed/stardust-ui-example-y3bqd

layershifter:
Fixes #1674.

This PR:
- fixes matching bug via usage of `Unstable_NestingAuto` component, same thing as in `Popup`
- adds a new prototype with different behaviors of nested `Popup`s and `Dialog`s
- adds E2E for fixed behavior 🚀


packages/react-component-nesting-registry/src/types.ts
```diff
@@ -2,7 +2,7 @@ import * as React from 'react'
 
 export type GetContextRefs = (needleRef: NodeRef) => NodeRef[]
 export type GetRefs = () => NodeRef[]
-export type NodeRef<T extends Node = Node> = React.RefObject<T>
+export type NodeRef<T extends Node = Node> = React.MutableRefObject<T>
```
layershifter, line 5: There refs are mutable, it allows to avoid usage of `handleRef` or `@tsignore`

e2e/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "../build/tsconfig.common.json",
+  "compilerOptions": {
+    "module": "esnext"
+  },
+  "include": ["../packages/react/src", "../e2e", "../types"]
+}
```
layershifter, line 7: Add `tsconfig.json` to get proper autocomplete

packages/react/src/lib/handleRef.ts
```diff
@@ -1,29 +0,0 @@
-import * as React from 'react'
```
layershifter, line 1: Cruft, this function is a part of `@stardust-ui/react-component-ref`

packages/react/src/components/Portal/Portal.tsx
```diff
@@ -17,11 +17,11 @@ import {
 import PortalInner from './PortalInner'
 import { FocusTrapZone, FocusTrapZoneProps } from '../../lib/accessibility/FocusZone'
 import { AccessibilityAttributes } from '../../lib/accessibility/types'
-import { AccessibilityKeyHandlers } from '../../lib/accessibility/reactTypes'
+import { AccessibilityHandlerProps } from '../../lib/accessibility/reactTypes'
 
 export type TriggerAccessibility = {
   attributes?: AccessibilityAttributes
-  keyHandlers?: AccessibilityKeyHandlers
+  keyHandlers?: AccessibilityHandlerProps
```
layershifter, line 24: Type was wrong previously

e2e/e2eApi.ts
```diff
@@ -39,7 +39,22 @@ export class E2EApi {
...
+      return
+    }
+
+    const dimensions: { x: number; y: number } = await this.page.evaluate(s => {
+      const rect = document.querySelector(s).getBoundingClientRect()
+
+      return { x: Math.round(rect.left), y: Math.round(rect.top) }
+    }, selector)
+
+    await this.page.mouse.click(dimensions.x + x, dimensions.y + y)
```
layershifter: https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pageclickselector-options

> This method fetches an element with selector, scrolls it into view if needed, and then uses page.mouse to click in the center of the element. If there's no element matching selector, the method throws an error.

To make this test working I need a different behavior because the center of `overlay` is `content` and click will be ignored

e2e/tests/dialogInPopup-test.ts
```diff
@@ -0,0 +1,51 @@
...
+  it('"Popup" should be open after "Dialog" will be opened', async () => {
+    await e2e.clickOn(popupTrigger)
+    expect(await e2e.exists(popupContent)).toBe(true)
+
+    await e2e.clickOn(dialogTrigger)
+    expect(await e2e.exists(popupContent)).toBe(true)
+    expect(await e2e.exists(dialogHeader)).toBe(true)
+  })
+
+  it('"Popup" should be open after "Dialog" will closed', async () => {
```
mnajdova:
```suggestion
  it('"Popup" should be open after "Dialog" will be closed', async () => {
```

e2e/tests/dialogInDialog-test.ts
```diff
@@ -0,0 +1,62 @@
...
+  it('An outer "Dialog" should be open after inner "Dialog" will be opened', async () => {
+    await e2e.clickOn(outerTrigger)
+    expect(await e2e.exists(outerHeader)).toBe(true)
+
+    await e2e.clickOn(innerTrigger)
+    expect(await e2e.exists(outerHeader)).toBe(true)
+    expect(await e2e.exists(innerHeader)).toBe(true)
+  })
+
+  it('A click inside inner "Dialog" should do nothing', async () => {
```
mnajdova:
```suggestion
  it('A click inside inner "Dialog" should not close dialogs', async () => {
```

mnajdova: do nothing seems awkward :)

e2e/tests/dialogInDialog-test.ts
```diff
@@ -0,0 +1,62 @@
...
+    await e2e.clickOn(outerTrigger)
+    await e2e.clickOn(innerTrigger)
+    await e2e.clickOn(innerHeader)
+
+    expect(await e2e.exists(outerHeader)).toBe(true)
+    expect(await e2e.exists(innerHeader)).toBe(true)
+  })
+
+  it('A click on overlay should close only matching "Dialog"', async () => {
```
mnajdova:
```suggestion
  it('A click on overlay should close only the last opened "Dialog"', async () => {
```

mnajdova: Is this what matching means?

e2e/e2eApi.ts
```diff
@@ -39,7 +39,22 @@ export class E2EApi {
...
-  public clickOn = async (selector: string) => await (await this.getElement(selector)).click()
+  public clickOn = async (selector: string, x?: number, y?: number) => {
+    if (x === undefined && y === undefined) {
+      const elementHandle = await this.getElement(selector)
+      await elementHandle.click()
+
+      return
+    }
+
+    const dimensions: { x: number; y: number } = await this.page.evaluate(s => {
```
kuzhelov: `boundingbox` method of `ElementHandle` can be used here instead: https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#elementhandleboundingbox

e2e/e2eApi.ts
```diff
@@ -39,7 +39,22 @@ export class E2EApi {
...
   }
 
-  public clickOn = async (selector: string) => await (await this.getElement(selector)).click()
+  public clickOn = async (selector: string, x?: number, y?: number) => {
+    if (x === undefined && y === undefined) {
+      const elementHandle = await this.getElement(selector)
+      await elementHandle.click()
+
+      return
+    }
```
kuzhelov: as at this point it might be the case that either `x` or `y` is undefined, lets ensure that any `undefined` value is coerced to default of `0`

layershifter: Introduced a separate `clickOnPosition()` method instead

mnajdova (approved):
Changed look good to me, please check one last time with accessibility team whether everything works as expected

layershifter:
Confirmed this with @jurokapsiar, however we found another issue #1709 that should be fixed separately.

PR #1711, opened 2019-07-23
feat(colors): add new set for unified ramp of category colors in Teams theme

natashamayurshah:
Add all variants for the new set of [category colors](https://share.goabstract.com/ff44bfb8-840e-4396-91e4-750adb1e5efc?mode=design&sha=35ed81e812ac420a053175b09551bc973dd6e5e6).


packages/react/src/themes/teams-dark/colors.ts
```diff
@@ -291,3 +291,40 @@ export const colorScheme: ColorSchemeMapping = {
...
+const createCategoryColorScheme = (color: string, customValues = {}) => {
+  return {
+    foreground1: categoryColors[color][250],
+    foreground2: categoryColors[color][550],
+    background: categoryColors[color][800],
+    ...customValues,
+  }
+}
+
+export const categoryColorSchemes: CategoryColorSchemeMapping = {
```
mnajdova: Let's preserve the previous naming convention and rename this to `categoryColorScheme`

packages/react/src/themes/teams-dark/colors.ts
```diff
@@ -291,3 +291,40 @@ export const colorScheme: ColorSchemeMapping = {
     background: colors.yellow[100],
   }),
 }
+
+const createCategoryColorScheme = (color: string, customValues = {}) => {
+  return {
+    foreground1: categoryColors[color][250],
```
mnajdova: Let's replaced `foreground1` with `foreground` and `foreground2` with `foreground1`

mnajdova: Also, is there a plan for the future, more of these colors to be added in the color scheme? Not sure why we have so many values in the color palette, but we clearly only use a subset of them..

packages/react/src/themes/types.ts
```diff
@@ -129,6 +137,39 @@ type EmphasisColorsStrict = Partial<{
 
 export type EmphasisColors = Extendable<EmphasisColorsStrict, ColorVariants>
 
+/**
+ * A type for a predefined category colors.
+ */
+type CategoryColorsStrict = Partial<{
```
mnajdova: If this is supposed to be `Strict`, let's remove the `Partial` from the definition. If it is not suppose to be `Strict`, let's change the name to CategoryColors.

mnajdova: Also these typings are specific to team theme, so we may move them to `packages/react/src/themes/teams/types.ts`

packages/react/src/themes/types.ts
```diff
@@ -142,6 +183,14 @@ export type ColorValues<T, Colors extends string | number | symbol = ColorNames>
   T
 >
 
+/**
+ * A type for an extendable set of CategoryColorNames properties of type T
+ */
+type CategoryColorValues<
```
mnajdova: `ColorValues` is already a generic, can't we just reuse it with different params

packages/react/src/themes/types.ts
```diff
@@ -213,6 +262,17 @@ export type StrictColorSchemeMapping<
   default?: Scheme
 }
 
+export type CategoryColorScheme<T extends string | number | symbol = ComponentAreaName> = Partial<
+  Record<T, string>
+>
+
+export type CategoryColorSchemeMapping<
```
mnajdova: We may also reuse the generics from the original `ColorSchemeMapping` type

packages/react/src/themes/teams-dark/colors.ts
natashamayurshah: Moving the second comment discussion into the conversation with Sue Nguyen.

mnajdova:
Final thing :) Let's add entry in the CHANGELOG.md file under `## [Unreleased]`'s `### Features` section, something like this:
- Add `categoryColors` and `categoryColorScheme` in Teams theme @natashamayurshah  ([#1711](https://github.com/stardust-ui/react/pull/1711))

mnajdova (approved):

PR #1716, opened 2019-07-24
feat(Button): add `size` prop

### LINKED ISSUE #1376: Button text is cut off in the bottom
### Steps
Create a Button, for example
< Button content= "See how this very long text shows up on the button" / >
 
### Expected Result
Button text is fully visible and not cut off in the bottom

### Actual Result
Button text is cut off in the bottom on letter such as 'y' and 'g'

### Version
^0.29.1

### Testcase

### LINKED ISSUE #1671: Button `size` prop or `small`/`large` variations
### Problem description

We need a proper design term to suggest users they can use the `Button` component at a smaller size or with a more compact style.

This prop will only impact styling, functionality and accessibility stay the same

#### 1. string prop `size`

- API:
```jsx
<Button size="small" />
```
- implemented in:
  - [SUIR](https://react.semantic-ui.com/elements/button/#variations-size)
  - [Material UI](https://material-ui.com/api/button/#props)
  - [Reactstrap](https://reactstrap.github.io/components/buttons/)
  - [Bootstrap](https://getbootstrap.com/docs/4.2/components/buttons/#sizes)
  - [Evergreen](https://evergreen.segment.com/components/button/)
  - [Carbon Components React](http://react.carbondesignsystem.com/?path=/story/buttons--default)

#### 2. boolean prop `compact`

- API:
```jsx
<Button compact />
```
- alternative names:  `small`, `fitted` ?
- implemented in:
  - [SUIR](https://react.semantic-ui.com/elements/button/#variations-compact)
  - [Vuetify](https://vuetifyjs.com/en/components/buttons#api) (search for `small` or `large`)
  - [Blueprint](https://blueprintjs.com/docs/#core/components/button.props) (search for `small`)

layershifter:
Fixes #1376.
Fixes #1671.

# BREAKING CHANGES

- `paddingLeftRightValue` was renamed to `padding`.

---

### Changes

This PR adds a new `size` prop and styles for `size=small`.

I also made a small refactor to use `p` & `v` pattern.


layershifter:
@codepretty can you please check these changes? 🙏

alinais (approved):

packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -1,215 +1,177 @@
...
+  content: ({ props: p, variables: v }): ICSSInJSStyle => ({
     overflow: 'hidden',
     textOverflow: 'ellipsis',
     whiteSpace: 'nowrap',
-    fontWeight: variables.contentFontWeight,
+    fontSize: v.fontSize,
+    fontWeight: v.contentFontWeight,
+
+    ...(p.size === 'small' && {
+      fontSize: v.sizeSmallFontSize,
```
codepretty: I think this need to have a line-height set for the small size and one for the default (medium) size to ensure that no descender are cut off...
image

codepretty: Adding the medium line-height as a default will also fix this - https://github.com/stardust-ui/react/issues/1376

layershifter: Cool, applied 👍

docs/src/examples/components/Button/Variations/index.tsx
```diff
@@ -24,6 +24,11 @@ const Variations = () => (
       description="A button can be shown in form of a text to indicate some less-pronounced actions."
       examplePath="components/Button/Variations/ButtonExampleText"
     />
+    <ComponentExample
+      title="Size"
+      description="A button can have assorted sizes."
```
mnajdova, line 29: It would be nice to consolidate the description of the size in the components' examples. We have different for Avatar, Text, Button.. :\

packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -1,215 +1,179 @@
...
         },
       }),
 
       // circular button defaults
-      ...(circular &&
-        !text && {
-          minWidth: height,
+      ...(p.circular &&
+        !p.text && {
+          minWidth: v.height,
```
mnajdova, line 69: Hmm, shouldn't this be updated depending on the size? It seems like a special case..

layershifter: image

```tsx
      <Button size="small" circular />
      <Button circular />
```

Yep, it should

layershifter: image

Updated 👍

PR #1718, opened 2019-07-24
feat(dropdown): updating label styles and dropdown selection

notandrew:


notandrew:
This requires a bunch of dropdown updates too...

PR #1721, opened 2019-07-26
feat(List): add horizontal prop

mnajdova:
Added horizontal prop to the `List` component. This prop is considered in the keyboard navigation of the selectable List.

### DIFF
--- a/docs/src/examples/components/List/Variations/ListExampleHorizontal.shorthand.tsx
+++ b/docs/src/examples/components/List/Variations/ListExampleHorizontal.shorthand.tsx
@@ -1,0 +1,32 @@
+import * as React from 'react'
+import { List, Image } from '@stardust-ui/react'
+
+const items = [
+  {
+    key: 'irving',
+    media: <Image src="public/images/avatar/small/matt.jpg" avatar />,
+    header: 'Irving Kuhic',
+    headerMedia: '7:26:56 AM',
+    content: 'Program the sensor to the SAS alarm through the haptic SQL card!',
+  },
+  {
+    key: 'skyler',
+    media: <Image src="public/images/avatar/small/steve.jpg" avatar />,
+    header: 'Skyler Parks',
+    headerMedia: '11:30:17 PM',
+    content: 'Use the online FTP application to input the multi-byte application!',
+  },
+  {
+    key: 'dante',
+    media: <Image src="public/images/avatar/small/nom.jpg" avatar />,
+    header: 'Dante Schneider',
+    headerMedia: '5:22:40 PM',
+    content: 'The GB pixel is down, navigate the virtual interface!',
+  },
+]
+
+const ListExampleSelectable = () => (
+  <List selectable defaultSelectedIndex={0} items={items} vertical={false} />
+)
+
+export default ListExampleSelectable
--- a/docs/src/examples/components/List/Variations/ListExampleHorizontal.shorthand.tsx
+++ b/docs/src/examples/components/List/Variations/ListExampleHorizontal.shorthand.tsx
@@ -20,7 +20,7 @@ const items = [
 ]
 
 const ListExampleSelectable = () => (
-  <List selectable defaultSelectedIndex={0} items={items} vertical={false} />
+  <List selectable defaultSelectedIndex={0} items={items} horizontal />
 )
 
 export default ListExampleSelectable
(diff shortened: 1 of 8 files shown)


lucivpav:
Since the default value of `vertical` is `true`, wouldn't it be more user-friendly, if we would specify `horizontal` instead? E.g. `<List horizontal>` or `<List horizontal={true}>`?

mnajdova:
@lucivpav that's a good point, but I chose the `vertical` only because we have that as a term in other components too. I think it will be easier for users to just learn that one prop, rather than remember where they should provide the `vertical`, and where the `horizontal` prop. Is this strong enough reason for you? :)

lucivpav:
> @lucivpav that's a good point, but I chose the `vertical` only because we have that as a term in other components too. I think it will be easier for users to just learn that one prop, rather than remember where they should provide the `vertical`, and where the `horizontal` prop. Is this strong enough reason for you? :)

Yes, it makes very good sense now :)

packages/react/src/lib/accessibility/Behaviors/List/selectableListBehavior.ts
```diff
@@ -8,8 +8,10 @@ import { ListBehaviorProps } from './listBehavior'
...
  * Adds role='listbox'.
- * Triggers 'moveNext' action with 'ArrowDown' on 'root'.
- * Triggers 'movePrevious' action with 'ArrowUp' on 'root'.
+ * Triggers 'moveNext' action with 'ArrowDown' on 'root', when orientation is vertical.
+ * Triggers 'moveNext' action with 'ArrowRight' on 'root', when orientation is horizontal.
+ * Triggers 'movePrevious' action with 'ArrowUp' on 'root', when orientation is vertical.
+ * Triggers 'movePrevious' action with 'ArrowLeft' on 'root', when orientation is horizontal.
  * Triggers 'moveFirst' action with 'Home' on 'root'.
  * Triggers 'moveLast' action with 'End' on 'root'.
  */
```
kolaps33, line 18: Please add "aria-orientation" attribute(on the root element, under line 19), based on the aria:
https://www.w3.org/TR/wai-aria-practices/#Listbox
If options are arranged horizontally, the element with role listbox has aria-orientation set to horizontal. The default value of aria-orientation for listbox is vertical.

Thank you :)

docs/src/examples/components/List/Variations/ListExampleHorizontal.shorthand.tsx
```diff
@@ -0,0 +1,32 @@
...
+    key: 'dante',
+    media: <Image src="public/images/avatar/small/nom.jpg" avatar />,
+    header: 'Dante Schneider',
+    headerMedia: '5:22:40 PM',
+    content: 'The GB pixel is down, navigate the virtual interface!',
+  },
+]
+
+const ListExampleSelectable = () => (
+  <List selectable defaultSelectedIndex={0} items={items} vertical={false} />
```
levithomason: Propose we use `horizontal` instead.  This is because the default list is expected to be vertical and the descriptive alternative is one that is horizontal.

This is similar to why there is an HTML attribute for `disabled` and not `enabled="false"`.

packages/react/src/themes/teams/components/List/listStyles.ts
```diff
@@ -3,18 +3,15 @@ import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
...
-        listStyle: 'none',
-        display: 'block',
-        padding: 0,
-        margin: 0,
-      }),
-    }
-  },
+  root: ({ props: p }): ICSSInJSStyle => ({
+    ...(p.debug && debugRoot()),
+    display: p.vertical ? 'block' : 'flex',
```
levithomason: What's the reasoning on horizontal lists being block level elements vs inline elements (`inline-flex`)? What's use cases are we targeting?

packages/react/src/lib/accessibility/Behaviors/List/selectableListBehavior.ts
mnajdova: Added, please take a look and propose how we can test it :\

docs/src/examples/components/List/Variations/ListExampleHorizontal.shorthand.tsx
mnajdova: Okay, my reasoning was that we already have the `vertical` prop, so users will have to think about before applying `vertical`/`horizontal` on different components, based on how they appear by default...

packages/react/src/themes/teams/components/List/listStyles.ts
mnajdova: No strong reasoning for this, I was just following the paradigm we had before. `inline-flex` works as expected as well, and may be even better to the usecases we have, something like this:
image

Will change it

layershifter (approved):

Issue #1715, opened 2019-07-24
Icon: "name" prop is not required


layershifter:
## Bug Report

The usage `Icon` component without `name` prop is fully legit now, but it's wrong. We should make `name` required or have a proper default value.

### Version
0.34.3

PR #1723, opened 2019-07-26
fix(Icon): make "name" prop required

lucivpav:
An instantiated `Icon` without a `name` makes no sense. This PR makes the `name` prop mandatory. An alternative would be to set a default name value. This resolves issue #1715.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - `Loader` uses `Text` component for `label` slot instead of `Box` @layershifter ([#1701](https://github.com/stardust-ui/react/pull/1701))
 - Fix test cut off in `Button` component @layershifter ([#1716](https://github.com/stardust-ui/react/pull/1716))
 - Close `Toolbar`'s menu when it looses focus @sophieH29 ([#1688](https://github.com/stardust-ui/react/pull/1688))
-- Fix `Icon` `name` prop not required @lucivpav ([#1723](https://github.com/stardust-ui/react/pull/1723))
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -48,6 +48,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - `Loader` uses `Text` component for `label` slot instead of `Box` @layershifter ([#1701](https://github.com/stardust-ui/react/pull/1701))
 - Fix test cut off in `Button` component @layershifter ([#1716](https://github.com/stardust-ui/react/pull/1716))
 - Close `Toolbar`'s menu when it looses focus @sophieH29 ([#1688](https://github.com/stardust-ui/react/pull/1688))
+- Fix `Icon` `name` prop not required @lucivpav ([#1723](https://github.com/stardust-ui/react/pull/1723))
```
```suggestion
- Require `name` prop in `Icon` component @lucivpav ([#1723](https://github.com/stardust-ui/react/pull/1723))
```

This like a bit cleaner to me

layershifter: Please also update `propTypes`:

```diff
-name: PropTypes.string,
+name: PropTypes.string.isRequired,
```

layershifter (approved):

packages/react/src/components/Icon/Icon.tsx
```diff
@@ -30,7 +30,7 @@ export interface IconProps extends UIComponentProps, ColorComponentProps {
   disabled?: boolean
 
   /** Name of the icon. */
-  name?: string
+  name: string
```
mnajdova, line 33: propTypes too?

lucivpav: Fixed just now

lucivpav:
@layershifter can you please take a look at the new changes? I have fixed the test errors.

PR #1726, opened 2019-07-29
feat: add support for styles debugging

kuzhelov:
This PR introduces first version of the toolset to debug Stardust styles (for most commonly reported debug scenarios). 

# Usage
Toolset is working in `development` builds only, and this is the set of steps to obtain debug information about Stardust component's styles:
- open React Dev Tools tab in browser
- in the elements tree select _**Stardust component**_ which styles should be inspected
- enter a query in console, e.g. `$stardust.whosProp('...')`, `$stardust.whosValue('..')` (or other supported API method)
  - result output will be provided with all the requested insights for the selected component

# API

### List of commands
- `$stardust.whosProp(propName | (propName) => boolean)`
  - filters all the `siteVariables`, `variables` and `styles` values to remain only those that have `propName` as a prop name.
  - supports predicate function being passed as argument.
- `$stardust.whosPropContains(propNameSubstring)`
  - filters `siteVariables`, `variables` and `styles` entries to only those that contain `propNameSubstring` in their prop names
- `$stardust.whosValue(value | (value) => boolean)`
  - filters all the `siteVariables`, `variables` and `styles` values to remain only those that have `value` as value
   - supports predicate function being passed as argument.
- `$stardust.whosValueContains(valueSubstring)`
  - filters `siteVariables`, `variables` and `styles` entries to only those that contain string values that contain `valueSubstring`.
- `$stardust()`
  - this command provides extensive summary of all the available debug information about component's styles. This output is used by the rest commands from the list as an input for filtering.

### Output 
Each of these methods filters the `debugData` object (which could be accessed by `$stardust()` with the component being selected in the React Dev Tools - this object contains extensive information about all the style ingredients of the component). 

The result of this filtering generally contains the following props: `componentName`, `siteVariables`, `variables`, `styles`. Each of these categories contains resolved values grouped in the following categories:
- `themes` - an array of theme values, in the same order as those are defined by `Provider` components, starting from the root-level component down to the component which is debugged
- `instanceOverrides` - values that are defined directly for component's instance by, for example, `styles` or `variables` prop
- `result` - contains result of all the values from all the sources being merged. This reflects the final value that is passed to component.

It is worth to notice that empty categories are omitted from the output of filtering. For example, if there are only component `styles` that satisfy filtering conditions, then the following object will be provided as an output:

```js
// EXAMPLE: only `styles` contain data that satisfy filter condition
> $stardust.whosProp(' ... ')

// outputs
{
  componentName: ...
  styles: {
     themes: [ 
        0: { ... },
        1: { }   // empty object - if there are no matching values defined in this theme
     ],
     instanceOverrides: { ... }
     result: { ... }
  }
}
```

And, if there is no filtered data in both `siteVariables`, `variables` and `styles`categories, then the empty object will be returned as a result:

```js
// EXAMPLE: nor `styles`, `siteVariables` or `variables` contain data that satisfy filter condition
> $stardust.whosProp(' ... ')

// outputs
{ }
```

# Implementation Notes
Below there is a list of criterias the implementation approach was aiming on:

- minimal changes to existing API to instrument debug scenarios
- modular implementation, easy to extend
- minimal affect on the existing component's scenarios and functionality
- minimal possible runtime costs
   - zero costs for `production` builds
   - for `development` builds
      - 0% CPU cost, as all the debug data is evaluated on-demand
      - 2% memory increase for `development` builds (tested on Stardust docs site's pages)


packages/react/src/lib/debug/debugDataProvider.ts
```diff
@@ -0,0 +1,151 @@
...
+  }: DebugDataProviderArgs) {
+    this.componentName = componentName
+    this.themes = themes
+    this.instanceStylesOverrides = instanceStylesOverrides
+    this.instanceVariablesOverrides = instanceVariablesOverrides
+    this.resolveStyles = resolveStyles
+    this.resolveVariables = resolveVariables
+  }
+
+  public resolve() {
```
lucivpav, line 51: Maybe break this down into smaller private functions? E.g. getSiteVariablesDebugData(...)

packages/react/src/lib/debug/utils.ts
```diff
@@ -0,0 +1,51 @@
...
+    return {}
+  }
+
+  return Object.keys(object).reduce((acc, currentPropName) => {
+    if (predicate(currentPropName, object[currentPropName])) {
+      return { ...acc, [currentPropName]: object[currentPropName] }
+    }
+
+    if (typeof object[currentPropName] === 'object') {
+      const deepPickResult = deepPickBy(object[currentPropName], predicate)
```
lucivpav: Is it possible that this would result in stack overflow if the object contains a circular reference?
E.g. obj.key1.key2.[...].keyX === obj
If so, should the user be responsible for ensuring no such circular reference exists?

kuzhelov: if this circular reference exists in the objects structure, than it means that the same circular reference will affect even the rendering process (so that even on the rendering stage the styles/variables won't be resolved). Thus, we may safely claim the following: if Stardust component was rendered successfully, it is safe to debug this component's styles

packages/react/src/lib/debug/debugDataProvider.ts
kuzhelov: sorry, not sure that I understand the comment: currently `resolve` method consists of several helper calls that, essentially, have the same semantics:

```js
resolve() {
   const siteVariablesDebugData = this.createDebugCategory(...)
   const variablesDebugData = this.createDebugCategory(...)
   ...
}
```

Could you, please, suggest what you'd like to split?

packages/react/src/lib/debug/utils.ts
kuzhelov: will add comment to the code. Thanks!

alinais (approved):

PR #1727, opened 2019-07-29
fix(DropdownSelectedItem): apply 'button' role

### LINKED ISSUE #1653: Dropdown -  VoiceOver doesn't narrate and navigate to user pill in multiple selection
### Steps
1. turn on VoiceOver
2. navigate to stardust page into "Search Multiple Selection" dropdown
3. Select 2 users into selection
4. try to navigate to each user pill with left/right arrow key

### Expected Result
VoiceOver navigate on the user pill and narrate corresponding message

### Actual Result
VoiceOver get focus around the whole input and narrate some unrelated strings... 
Screenshot 2019-07-16 at 6 22 05 PM

silviuaavram:
Fixes https://github.com/stardust-ui/react/issues/1653.

Now it should focus without any issues.
MaterialUI also has them as buttons. https://material-ui.com/components/autocomplete/

Since we also have onClick on it then maybe it should be a button.

@kolaps33 @jurokapsiar @mshoho @sophieH29

### DIFF
--- a/packages/react/src/components/Dropdown/DropdownSelectedItem.tsx
+++ b/packages/react/src/components/Dropdown/DropdownSelectedItem.tsx
@@ -165,6 +165,7 @@ class DropdownSelectedItem extends UIComponent<WithAsProp<DropdownSelectedItemPr
         <Label
           className={classes.root}
           tabIndex={active ? 0 : -1}
+          role="button"
           styles={styles.root}
           circular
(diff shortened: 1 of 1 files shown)


sophieH29 (approved):

silviuaavram:
will close this since it seems to be a bug in Chrome and not on our side. Will follow up with an issue on their side.

PR #1728, opened 2019-07-29
feat(Tree): add onActiveIndexChange prop

lucivpav:
This PR adds an ability to respond when the active index of `Tree` component changes.

### DIFF
--- a/packages/react/src/components/Tree/Tree.tsx
+++ b/packages/react/src/components/Tree/Tree.tsx
@@ -120,11 +120,11 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
             return acc
           }, [])
         : []
-      this.setActiveIndexAndTriggerEvent(e, activeIndex)
+      this.trySetActiveIndexAndTriggerEvent(e, activeIndex)
     },
   }
 
-  setActiveIndexAndTriggerEvent = (e, activeIndex) => {
+  trySetActiveIndexAndTriggerEvent = (e, activeIndex) => {
     this.trySetState({ activeIndex })
     _.invoke(this.props, 'onActiveIndexChange', e, { ...this.props, activeIndex })
   }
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Tree/Tree.tsx
```diff
@@ -52,6 +53,13 @@ export interface TreeProps extends UIComponentProps, ChildrenComponentProps {
    * @param {ReactNode|ReactNodeArray} children - The computed children for this slot.
    */
   renderItemTitle?: ShorthandRenderFunction
+
+  /** Called when activeIndex changes.
+   *
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props and proposed value.
+   */
+  onActiveIndexChange?: ComponentEventHandler<TreeProps & { activeIndex: number[] | number }>
```
mnajdova:
```suggestion
  onActiveIndexChange?: ComponentEventHandler<TreeProps>
```

lucivpav:
@mnajdova can you please take a look at this PR again?

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -111,10 +120,15 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
             return acc
           }, [])
         : []
-      this.trySetState({ activeIndex })
+      this.setActiveIndexAndTriggerEvent(e, activeIndex)
     },
   }
 
+  setActiveIndexAndTriggerEvent = (e, activeIndex) => {
```
kuzhelov: nit: just for the sake of preserving semantics in function name, would suggest to rename the function to `try SetActiveIndexAndTriggerEvent` - in that case function's name will clearly advertise that index, in fact, may not be changed (if controlled)

kuzhelov (approved):

PR #1732, opened 2019-07-29
fix(FocusTrapZone): firstFocusableSelector is fragile

### LINKED ISSUE #1566: FocusTrapZone firstFocusableSelector prop is fragile
## Bug Report

FocusTrapZone firstFocusableSelector prop is fragile.

1) it is called selector, but it expects class, as it prepends the selector with .
-> only prepend . if the string does not start with . or #
2) it allows seting focus to elements outside of the focus trap zone
-> if selector evaluates to something outside of the zone or to nothing, find the first focusable in the zone

### Version
0.33

sophieH29:
Fixes https://github.com/stardust-ui/react/issues/1566

### Breaking change
Before:
`firstFocusableSelector` was used only to specify the class name string selector. `.` was appended to the passed value.

After:
`firstFocusableSelector` can be any string selector as is used in API for `document.querySelector`

--------------------------------
This PR introduces changes to `FocusTrapZone` and `AutoFocusZone` to not limit a user to using class selectors only.
Now it's a user responsibility to pass the whole string selector whether it's a `.blah` or `#blah` or tag `blah` etc. If this element won't be found in the container, the first tabbable element will be selected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Features
 - Expose `isFromKeyboard` in `Grid` component @sophieH29 ([#1729](https://github.com/stardust-ui/react/pull/1729))
 
+### Fixes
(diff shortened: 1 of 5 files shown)


silviuaavram (approved):

layershifter (approved):

layershifter:
Please describe breaking changes in PR description 😼

PR #1734, opened 2019-07-29
updating button styles

notandrew:


packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -43,25 +42,65 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps & ButtonState, ButtonVa
...
       ...(!p.text && {
-        outline: 0,
-        borderWidth,
         borderStyle: 'solid',
         borderColor: v.borderColor,
         boxShadow: v.boxShadow,
+        borderRadius: '.2rem',
+        border: 0,
+
+        '::after': {
```
codepretty: what is the `after` and `before` here for?

notandrew: This is for applying the correct focus border styles to the buttons, one inner border and one outer border. When we use the normal border property, the content can shift, this just applied pseudo borders so the shape doesn't change.

packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -41,30 +40,73 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps, ButtonVariables> = {
...
+            content: '""',
+            borderRadius: '.2rem',
+            border: '.1rem solid transparent',
+            top: '.1rem',
+            right: '.1rem',
+            bottom: '.1rem',
+            left: '.1rem',
+          },
+
+          ...(p.isFromKeyboard && {
```
mnajdova, line 77: `isFromKeyboard` does not exists any more, use the":focus-visible" instead. Also, never use multiple pseudo selectors at the same time. You can change the following lines to:
```
":focus": {
  ":active": {
    borderColor: '...',
  },
}
```

packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -41,30 +40,73 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps, ButtonVariables> = {
...
+            ':focus::after': {
+              borderColor: '#252423',
```
mnajdova, line 79: No hex values please

packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -41,30 +40,73 @@ const buttonStyles: ComponentSlotStylesInput<ButtonProps, ButtonVariables> = {
...
+            },
+
+            ':focus::before': {
+              borderColor: '#fff',
```
mnajdova, line 83: No hex values please

Issue #1708, opened 2019-07-23
Using space to activate Menu Item scrolls the page


mnajdova:
### Steps
Focus some menu item element and press space

### Expected Result
The element should be activated, but there should be no scrolling on the page.

### Actual Result
The element is activated, but the whole page scrolls.💣

### Short gif with the behavior
https://i.imgur.com/aIYJlOC.gif

v0.34.2

sophieH29:
fixed https://github.com/stardust-ui/react/pull/1735

PR #1735, opened 2019-07-30
fix(MenuItem): make spacebar not scroll the page, fix submenu not closing

lucivpav:
This PR makes sure that pressing *space* on focused `MenuItem` does not lead the page being scrolled. The issue is described here: #1708.

It also fixes a problem when clicking on a submenu item with keyboard does not lead to the menu to be closed.


packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -335,6 +335,8 @@ class MenuItem extends AutoControlledComponent<WithAsProp<MenuItemProps>, MenuIt
         e.stopPropagation()
         e.preventDefault()
       }
+    } else {
+      e.preventDefault()
```
layershifter: > * Triggers 'performClick' action with 'Enter' or 'Spacebar' on 'root'.

So this action can be dispatched by `Enter` key, too. @sophieH29 is it correct to prevent default on it?

mnajdova: We cannot do this, this broke the submenu behavior, try opening with keyboard some submenu and press enter on some action and you will see that the menus are not closed. Awkward that this was not caught by any test :(

mnajdova: Actually, not sure whether it is broken, but we have to ensure that we will not break the behavior...

sophieH29: @layershifter yes should be ok, no implications. We've done it for the most action handlers - on Enter/Space keys

miroslavstastny: I don't know what the exact reason is but as @mnajdova wrote submenu now does not close on Enter/Space with `menuBehavior` or `menuAsToolbarBehavior`. In master it does not work for `menuBehavior` but it closes as expected when `menuAsToolbarBehavior` is used.

@lucivpav could you please investigate it before merging the PR?

packages/react/test/specs/components/Menu/Menu-test.tsx
```diff
@@ -66,6 +77,19 @@ describe('Menu', () => {
...
+      const menu = mountWithProvider(<Menu items={getNestedItems()} />)
+      const menuItems = menu.find('MenuItem')
+
+      menuItems.at(1).simulate('keydown', { keyCode: keyboardKey.Spacebar })
+      menuItems
+        .at(1)
+        .at(0)
+        .simulate('keydown', { keyCode: keyboardKey.Spacebar })
+
+      expect(menuItems.at(1).state('menuOpen')).toBe(false)
```
lucivpav, line 96: Without the fix, the `Menu` component behaves in a weird way. The `menuOpen` state is `undefined`, instead of `true`. Upon investigation, `this.props` in `handleClick` are not what one would expect. They seem like the props of the root component. This probably causes the `menuOpen` not to be defined. I spent too much time on this, trying different ways to make the value `true` instead of `undefined`, but without luck.

packages/react/src/lib/accessibility/Behaviors/Menu/menuItemBehavior.ts
```diff
@@ -43,12 +43,10 @@ const menuItemBehavior: Accessibility<MenuItemBehaviorProps> = props => ({
   },
 
   keyActions: {
-    root: {
+    wrapper: {
```
jurokapsiar: we have one corner case where wrapper is removed by setting { wrapper: null } in the menuitem shorthand. I think enter/space would not work in that case. why does click on root close the menu but enter on root does not?

lucivpav: > why does click on root close the menu but enter on root does not?

When Enter is pressed, `handleClick` is called on `MenuItem`. If root is used, the `this.props` in `handleClick` is not `MenuItemProps`, but instead something that looks like the root props. As a result of this, when `performClick` is called (from `handleClick`), the following branch of code is not executed (`this.props` does not contains `menu`):
```
    const { active, menu } = this.props

    if (menu) {
         ...
    }
    ...
```

Is it a correct behavior, that `this.props` is not set to `MenuItemProps` when we are using `root` and Enter?

sophieH29: @lucivpav 
> Is it a correct behavior, that this.props is not set to MenuItemProps when we are using root and Enter?

I think it's not correct behavior. I would expect same menu item props to be passed

> but instead something that looks like the root props

What do you mean by that? What is the difference between those props objects?

lucivpav: As discussed with Sofiya, if we are using `root`, the event does not propagate to other `root` elements, because they are siblings. Lets just use `wrapper` instead of `root`.

jurokapsiar: as I mentioned - we have a case where wrapper is not used, this PR would break that case. Let's discuss in person tomorrow

lucivpav: I have updated the PR. Note that if we are not using wrapper, then a menu cannot contain a submenu.

layershifter (approved):

packages/react/src/lib/accessibility/Behaviors/Menu/menuItemBehavior.ts
```diff
@@ -44,11 +44,18 @@ const menuItemBehavior: Accessibility<MenuItemBehaviorProps> = props => ({
...
+      ...(!props.hasWrapper && {
+        performClick: {
+          keyCombinations: [{ keyCode: keyboardKey.Enter }, { keyCode: keyboardKey.Spacebar }],
+        },
+      }),
     },
     wrapper: {
+      ...(props.hasWrapper && {
+        performClick: {
+          keyCombinations: [{ keyCode: keyboardKey.Enter }, { keyCode: keyboardKey.Spacebar }],
```
jurokapsiar: did you try leaving performClick on root and adding closeAllNenus on wrapper when enter/space is pressed?

lucivpav: Yes, but closeAllMenus would be called on the item. Since it contains no submenu, nothing gets closed.

jurokapsiar: let's say we have parent and child menu. You press enter on an item of the child menu. The key handler on the wrapper should get an Enter keydown. That item will not do anything as it does not have a submenu. So the event propagates to the parent menu item, which has a submenu - key handler on the parent warpper will handle it and the closeAllMenus method will close the menu - but maybe my thinking is wrong, can you please debug?

lucivpav: I managed to make it work using `closeAllMenus()`. This simplifies the implementation 😇

jurokapsiar (approved):

PR #1738, opened 2019-07-31
feat(Dropdown): add checkable and checkableIndicator prop

mnajdova:
We have a requirement for Teams theme for adding checkmark icon in the selected dropdown item. We have several options:

1. add `selectedIndicator` icon shorthand prop in the `Dropdown`, for allowing setting checkmark on the selected drop down item:

image

2. Add `checkable` & `checkableIndicator` props on the Dropdown & DropdownItem - similar to the clearable and clearableIndicator that we already have (thanks @layershifter for the proposal)

3. Adding endMedia prop in the Dropdown Item as it is already a list item, and users can use it if they want to add some icon when the item is selected.

We need to decide on a common pattern for these things. In the Input, we have only the clearable boolean prop, so we have no way of customizing that slot. In the Dropdown together with the clearable we have the clearableIndicator. As we tend to say that every slot in the Stardust components can be customized, I believe we need to have slots for these things as well, although that they have some defaults.


docs/src/examples/components/Dropdown/Types/DropdownExample.shorthand.tsx
```diff
@@ -17,6 +17,7 @@ const DropdownExample = () => (
   <Dropdown
     items={inputItems}
     placeholder="Select your hero"
+    selectedIndicator="stardust-checkmark"
```
mnajdova: Not sure if we should add separate example for this one..

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -285,7 +288,8 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
     search: PropTypes.oneOfType([PropTypes.bool, PropTypes.func]),
     searchQuery: PropTypes.string,
     searchInput: customPropTypes.itemShorthand,
-    toggleIndicator: customPropTypes.itemShorthand,
+    selectedIndicator: customPropTypes.itemShorthandWithoutJSX,
+    toggleIndicator: customPropTypes.itemShorthandWithoutJSX,
```
mnajdova: fixed prop type

packages/react/src/components/Dropdown/DropdownItem.tsx
```diff
@@ -43,8 +45,11 @@ export interface DropdownItemProps extends UIComponentProps<DropdownItemProps> {
    */
   onClick?: ComponentEventHandler<DropdownItemProps>
 
-  /** A dropdown item can be selected if single selection Dropdown is used. */
+  /** A dropdown item can be selected. */
   selected?: boolean
```
mnajdova: Added missing prop in the interface

packages/react/src/themes/teams/components/Dropdown/dropdownItemStyles.ts
```diff
@@ -59,6 +59,10 @@ const dropdownItemStyles: ComponentSlotStylesInput<DropdownItemProps, DropdownVa
     fontSize: v.listItemContentFontSize,
     color: v.listItemContentColor,
   }),
+  selectedIndicator: ({ variables: v }) => ({
+    position: 'relative',
+    left: pxToRem(3),
```
mnajdova: The padding on the `DropdownItem` is `11px`, but the icon should be `8px` from the right border, that's why we need styles

packages/react/src/components/Dropdown/DropdownItem.tsx
```diff
@@ -102,6 +115,17 @@ class DropdownItem extends UIComponent<WithAsProp<DropdownItemProps>> {
             styles: styles.content,
           },
         })}
+        endMedia={
```
mnajdova, line 123: This is different then how the rest of the elements are created and where the styles are applied, but there are issues with the height if we don't provide these styles.. I am open for other ideas...

mnajdova: Ping @layershifter @miroslavstastny

layershifter:
### Atlaskit

https://atlaskit.atlassian.com/packages/core/select
Please check `Named Exports`:

image
image

### BlueprintJS

`MultiSelect` there has similar idea:
https://blueprintjs.com/docs/#select/multi-select

image

###  AntDesign

https://ant.design/components/select/
Check multiple selection.

image

silviuaavram (approved):

kolaps33:
We will need to set aria-selected on selected option:
Based on example 2 on this page:
https://www.w3.org/TR/wai-aria-practices-1.1/examples/listbox/listbox-rearrangeable.html

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -285,7 +293,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
     search: PropTypes.oneOfType([PropTypes.bool, PropTypes.func]),
     searchQuery: PropTypes.string,
     searchInput: customPropTypes.itemShorthand,
-    toggleIndicator: customPropTypes.itemShorthand,
+    toggleIndicator: customPropTypes.itemShorthandWithoutJSX,
```
mnajdova, line 296: fixed prop type

packages/react/src/themes/teams/components/Dropdown/dropdownItemStyles.ts
```diff
@@ -59,6 +59,13 @@ const dropdownItemStyles: ComponentSlotStylesInput<DropdownItemProps, DropdownVa
     fontSize: v.listItemContentFontSize,
     color: v.listItemContentColor,
   }),
+  checkableIndicator: ({ variables: v }) => ({
+    position: 'relative',
+    left: pxToRem(3),
```
mnajdova, line 64: The padding on the `DropdownItem` is `11px`, but the icon should be `8px` from the right border, that's why we need styles

kolaps33 (approved):

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Add `onActiveIndexChange` prop to `Tree` component @lucivpav ([#1728](https://github.com/stardust-ui/react/pull/1728))
 - Add `horizontal` prop for `List` component @mnajdova ([#1721](https://github.com/stardust-ui/react/pull/1721))
 - Export `call-blocked` icon to Teams theme @francescopalmiotto ([#1736](https://github.com/stardust-ui/react/pull/1736))
+- add `checkable` and `checkableIndicator` to the `Dropdown` and `DropdownItem` components @mnajdova ([#1738](https://github.com/stardust-ui/react/pull/1738))
```
```suggestion
- Add `checkable` and `checkableIndicator` to the `Dropdown` and `DropdownItem` components @mnajdova ([#1738](https://github.com/stardust-ui/react/pull/1738))
```

layershifter (approved):

PR #1739, opened 2019-07-31
fix(ChatMessage): actions menu's appearance

kuzhelov:
This PR changes the way positioning is applied to `actionsMenu` of `ChatMessage` - now `Popper` component is used to control positioning of this floating section. 

This change is necessary to handle situations where `actionsMenu`'s content may be cut by the edges of the viewport - this may be the case for messages with short content and long list of `actionMenu`'s items.

## Before
Actions menu content is cut off.

image

## Now
Actions menu's position is adjusted.

image

## Notes on the approach
To handle `actionsMenu` cut-off case, the strategy of using `Popper`'s positioning was taken. Essentially, previously the set of static CSS props (in a sense that they were prescribed once and for entire component's time) were used for absolutely position `actionsMenu`, and these CSS values were part of the theme.

Unfortunately, these values should be set dynamically, in order to handle the cases when viewport's width is not enough to show the entire `actionMenu`'s content - that's why `Popper`'s positioning utilities were considered to be part of the solution

There were an alternative strategy to use `Popper` wrapper for the slot's content, instead of providing `Popper` as part of `ChatMessage`'s implementation. However, this introduces couple of problems:
- there will be 'two sources of truth' for the `actionMenu` positioning: one is the slot's wrapper, and another one - CSS values defined in theme
- this results in quite complicated and brittle logic for the client

Thus, it was decided to pay the costs of extensibility and introduce `Popper` positioning for `actionsMenu` as part of `ChatMessage`'s implementation, at least for now.


packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -153,14 +157,16 @@ class ChatMessage extends UIComponent<WithAsProp<ChatMessageProps>, ChatMessageS
     },
 
     focus: event => {
-      if (this.messageRef) {
```
kuzhelov, line 156: this check was guaranteed to pass

packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -138,13 +141,14 @@ class ChatMessage extends UIComponent<WithAsProp<ChatMessageProps>, ChatMessageS
     reactionGroupPosition: 'start',
   }
 
+  updateActionsMenuPosition = () => {}
+
   state = {
     focused: false,
     isFromKeyboard: false,
+    messageDomNode: null,
```
kuzhelov, line 149: it is necessary to use `state` to store this DOM node, as otherwise `Popper` component that is used for `actionsMenu` rendering won't update its target once `messageDomNode` is set

packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -178,6 +184,33 @@ class ChatMessage extends UIComponent<WithAsProp<ChatMessageProps>, ChatMessageS
...
+    })
+
+    if (!actionMenuElement) {
+      return actionMenuElement
+    }
+
+    return (
+      <Popper targetRef={this.state.messageDomNode} position="above" align="end">
+        {({ scheduleUpdate }) => {
+          this.updateActionsMenuPosition = scheduleUpdate
```
kuzhelov, line 207: it is necessary to trigger position updates for the cases when actions menu appear, as it has width of `0` when is hidden (https://github.com/stardust-ui/react/blob/master/packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts#L95)

packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -246,10 +272,15 @@ class ChatMessage extends UIComponent<WithAsProp<ChatMessageProps>, ChatMessageS
...
-      <Ref innerRef={this.messageRef}>
+      <Ref
+        innerRef={domNode => {
+          !this.state.messageDomNode && this.setState({ messageDomNode: domNode })
+        }}
+      >
         <ElementType
           onBlur={this.handleBlur}
           onFocus={this.handleFocus}
+          onMouseEnter={() => this.updateActionsMenuPosition()}
```
jurokapsiar, line 283: is it ok not to update the position when you are only using keyboard?

kuzhelov: yes, as logic of `this.handleFocus` is responsible for making position updates in this case: https://github.com/stardust-ui/react/pull/1739/files#diff-36ca6d6c5aa2d1fa2b0e2d9f04902716R168

alinais (approved):

PR #1741, opened 2019-07-31
fix(fela): plugin order, invokeKeyframes

lucivpav:
### plugin order
This PR resolves and tests a problem with `felaRenderer`. The order of plugins being executed actually matters. The problem is that `felaPluginFallbackValue()` comes before `felaInvokeKeyframesPlugin()`.

`felaPluginFallbackValue()` does the following:
**Input**
{
  color: [ 'rgba(0, 0, 0, 0.5)', '#ccc']
}
**Output**
{
  color: 'rgba(0, 0, 0, 0.5);color:#ccc'
}

But if we have, say, an Animation with `keyframeParams=[ '0%', '50%', '100%' ]`, the array would be processed `felaPluginFallbackValue()`. If the `keyframe` function expects an array, we get an error, because the `keyFrameParams` is no longer an array.

### invokeKeyframes
The `checkboxStyles` definition contains `root: display: ['inline-grid', '-ms-inline-grid']`. This gets processed by `felaInvokeKeyframesPlugin` incorrectly, rendering it as `root: display: { 0: 'inline-grid', 1: '-ms-inline-grid' }`.


lucivpav:
This PR is to be merged after #1744

packages/internal-tooling/package.json
```diff
@@ -24,8 +24,11 @@
     "eslint-plugin-jsx-a11y": "^6.2.1",
     "eslint-plugin-react": "^7.12.4",
     "eslint-plugin-react-hooks": "^1.6.0",
+    "fela": "^10.6.1",
```
miroslavstastny, line 27: We should wait for #1743 to be merged first and remove this change here.
@layershifter Do you agree?

packages/internal-tooling/package.json
```diff
@@ -24,8 +24,11 @@
...
     "jest": "^24.5.0",
-    "jest-axe": "^3.1.1"
+    "jest-axe": "^3.1.1",
+    "jest-react-fela": "^10.6.1",
+    "react-fela": "^10.6.1"
```
miroslavstastny, line 31: The same here

packages/react/test/specs/lib/felaRenderer-test.tsx
```diff
@@ -0,0 +1,90 @@
...
+import Icon from 'src/components/Icon/Icon'
+import Provider from 'src/components/Provider/Provider'
+import Text from 'src/components/Text/Text'
+import { felaRenderer } from 'src/lib'
+import { teams } from 'src/themes'
+
+test('css fallback value', () => {
+  const snapshot = createSnapshot(
+    <EmptyThemeProvider>
+      <Box styles={{ color: 'red' }} />
```
miroslavstastny, line 12: This is not testing css fallback value.
- please rename this to something like 'basic styles'.
- add css fallback value test with `styles={{ color: ['red', 'blue'] }}`

packages/react/test/specs/lib/felaRenderer-test.tsx
```diff
@@ -0,0 +1,90 @@
...
+    }),
+    keyframeParams: {
+      fromColor: 'red',
+      toColor: 'blue',
+    },
+    duration: '5s',
+  }
+
+  const snapshot = createSnapshot(
+    <Provider theme={teams}>
```
miroslavstastny: This adds insane amount of styles to the snapshot (fonts, static styles).
It makes the snapshot difficult to review and the test fragile.
Can we do it with just inline minimalistic theme?

packages/react/test/specs/lib/felaRenderer-test.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <Provider theme={{ animations: { spinner } }}>
+        <Animation name="spinner">
+          <Icon name="umbrella" circular bordered />
```
miroslavstastny: The same story - Icon is unnecessarily complex component to be used here. Go with just a box.

packages/react/test/specs/lib/felaRenderer-test.tsx
```diff
@@ -0,0 +1,90 @@
...
+        </Animation>
+      </Provider>
+    </Provider>,
+    {},
+    felaRenderer,
+  )
+  expect(snapshot).toMatchSnapshot()
+})
+
+test('keyframe returning css fallback value', () => {
```
miroslavstastny: Rename - this is not css fallback value. This is 'keyframe is array as an argument'.

packages/react/test/specs/lib/felaRenderer-test.tsx
```diff
@@ -0,0 +1,90 @@
...
+  const steps = ['0%', '100%']
+
+  const spinner = {
+    keyframe: ({ steps }) => {
+      const obj = {}
+      steps.forEach((step: string, index) => {
+        ;(obj as any)[step] = { opacity: 0.28 }
```
miroslavstastny: This (the `opacity`) would have to be an array to test 'array returned by keyframe results in CSS fallback values'

packages/react/test/specs/lib/felaRenderer-test.tsx
```diff
@@ -0,0 +1,90 @@
...
+      })
+      return obj
+    },
+  }
+
+  const snapshot = createSnapshot(
+    <Provider theme={teams}>
```
miroslavstastny: As mentioned in one of the previous tests, remove this theme to reduce snapshot size.

packages/react/test/specs/lib/felaRenderer-test.tsx
```diff
@@ -0,0 +1,90 @@
...
+      <Provider theme={{ animations: { spinner } }}>
+        <Animation name="spinner" keyframeParams={{ steps }}>
+          <Icon name="umbrella" circular bordered />
```
miroslavstastny: -> Box

packages/react/test/specs/lib/felaRenderer-test.tsx
```diff
@@ -0,0 +1,90 @@
...
+        </Animation>
+      </Provider>
+    </Provider>,
+    {},
+    felaRenderer,
+  )
+  expect(snapshot).toMatchSnapshot()
+})
+
+test('rtl', () => {
```
miroslavstastny: This does not test anything from RTL plugin.
Add `styles={{marginLeft: '10px'}}` to verify it works.

miroslavstastny (commented):
For each test:
- use a descriptive name which clearly indicates what the test verifies (or in other words what is broken if the test is red)
- check generated snapshot to verify it is as small as possible and it tests the right thing
- try to make the test red (depending on what the test it testing either disable a plugin in fela config or change the plugin order) to verify the tests are working.

packages/react/test/specs/lib/felaRenderer-test.tsx
lucivpav: If I do not wrap it in Teams theme, then the resulting snapshot doesn't contain the `@keyframes` section. How can I create such a minimal theme, that will allow me to use the Animation's `keyframe`s?

packages/react/test/specs/lib/felaInvokeKeyframesPlugin-test.ts
```diff
@@ -56,4 +56,12 @@ describe('felaRenderKeyframesPlugin', () => {
       animationDuration: '2s',
     })
   })
+
+  test('does not transform a list of strings', () => {
```
lucivpav, line 60: Maybe the `felaInvokeKeyframesPlugin` fix and test should be moved into a separate commit? The reason for this fix is that screener tests are failing, if the order of plugins is changed.

packages/react/src/lib/felaInvokeKeyframesPlugin.ts
```diff
@@ -25,6 +29,13 @@ export default () => {
         }
       }
 
+      if (Array.isArray(cssPropertyValue)) {
+        return {
+          ...acc,
+          [cssPropertyName]: cssPropertyValue.map(arrayElement => invokeKeyframes(arrayElement)),
+        }
+      }
```
layershifter: Please add matching for plugin. I don't get why it's required...

packages/react/test/specs/lib/felaInvokeKeyframesPlugin-test.ts
lucivpav: The `checkboxStyles` definition contains `root: display: ['inline-grid', '-ms-inline-grid']`. This gets processed by `felaInvokeKeyframesPlugin` incorrectly, rendering it as `root: display: { 0: 'inline-grid', 1: '-ms-inline-grid' }`.

packages/react/src/lib/felaInvokeKeyframesPlugin.ts
layershifter: Why it fails and still is not safe?

```tsx
typeof [] // 'object'
typeof null // 'object'
```

I suggest to use `_.isPlainObject` instead:
```tsx
      if (_.isPlainObject(cssPropertyValue)) {
        if (cssPropertyName === 'animationName') {
          if (cssPropertyValue.keyframe) {
            styles[cssPropertyName] = callable(cssPropertyValue.keyframe)(
              cssPropertyValue.params || {},
            )
          }

          return {
            ...acc,
            [cssPropertyName]: styles[cssPropertyName],
          }
        }

        return {
          ...acc,
          [cssPropertyName]: invokeKeyframes(cssPropertyValue),
        }
      }
```

miroslavstastny (approved):

PR #1749, opened 2019-08-01
fix(FocusZone): Elements with `contenteditable` attribute should be respected on Home/End key press

### LINKED ISSUE #1608: Home/End don't work in contenteditable divs underneath FocusZone
## Bug Report
Home/End keys don't work correctly when the user is inside of a contenteditable div inside of a FocusZone.

### Steps
1. Place a `contenteditable="true"` `div` as the content for a `Chat`
2. Place your cursor in the div
3. Hit `end` key

### Expected Result
The cursor goes to the end of the content editable text

### Actual Result
The user is navigated to the last chat in the list

### Version
0.34.0

### Testcase
Unable to generate a test case due to https://github.com/stardust-ui/react/issues/1607

sophieH29:
Fixes https://github.com/stardust-ui/react/issues/1608

When pressing `Home` \ `End` keys, focus should not move out of element with `contenteditable` attribute.


jurokapsiar (approved):

lucivpav:
Is it possible to add test(s) for this?

PR #1757, opened 2019-08-05
chore(List): use FocusZone in selectable list

jurokapsiar:
Use FocusZone for selectable list instead of the FocusHandler. This will allow us to use focus restoration on item removed once the appropriate change from fabric gets merged. It is also much more performant as the List does not have to update state on each keypress.

### DIFF
--- a/packages/react/src/lib/accessibility/Behaviors/List/selectableListBehavior.ts
+++ b/packages/react/src/lib/accessibility/Behaviors/List/selectableListBehavior.ts
@@ -1,6 +1,6 @@
-import * as keyboardKey from 'keyboard-key'
-import { Accessibility } from '../../types'
+import { Accessibility, FocusZoneMode } from '../../types'
 import { ListBehaviorProps } from './listBehavior'
+import { FocusZoneDirection } from '../../FocusZone'
 
 /**
  * @description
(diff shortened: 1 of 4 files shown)


jurokapsiar:
TODO: changelog

packages/react/src/lib/accessibility/Behaviors/List/selectableListBehavior.ts
```diff
@@ -25,24 +21,11 @@ const selectableListBehavior: Accessibility<ListBehaviorProps> = props => ({
...
-      moveFirst: {
-        keyCombinations: [{ keyCode: keyboardKey.Home }],
-      },
-      moveLast: {
-        keyCombinations: [{ keyCode: keyboardKey.End }],
-      },
+  focusZone: {
+    mode: FocusZoneMode.Embed,
+    props: {
+      shouldFocusInnerElementWhenReceivedFocus: true,
```
sophieH29, line 27: shouldn't we then put 'tabindex=0' to list container? So once a user TAB to list, focus will be moved inside using `shouldFocusInnerElementWhenReceivedFocus `.

Another thought which came up, previously the last selected item has set "tabindex=0", so when user come back to list, focus moves to the previously selected item (not previously focused). Does it still work like that?

jurokapsiar: I tried tabindex=0, but for some reason I was not then able to tab away from the list. Probably we would need to listen on focus/blur events and set tabIndex 0 only if the focus is outside. I decided not to include it in this PR.

For focused/selected - no, tabIndex=0 is on the last focused. I think in majority of cases that is sufficient, as the user typically focuses an item and then selects it. We might be able to use `defaultTabbableElement` but again, I would keep it as a separate task/PR

sophieH29 (approved):

Issue #1742, opened 2019-07-31
Embed component has no indication it's focused


stuartlong:
## Bug Report
Focusing the embed component via the keyboard yields no visual indication that the component is focused.

### Steps
1. Navigate to the play/pause button of an embed component via the keyboard

### Expected Result
There should be a border or some other indication that the play button is focused

### Actual Result
no indication

### Version
0.35.0

### Testcase
https://codesandbox.io/s/stardust-ui-example-bdi2c

Issue #1586, opened 2019-07-03
Embed video has a white flash when toggle active state


stuartlong:
## Bug Report
When playing or pausing an embedded video there is a white flash when switching between the placeholder and the video. This is particualry annoying when the placeholder is the first frame of the video.

### Steps
Create an Embed component and press play

### Expected Result
Playing a video should smoothly show the video.

### Actual Result
There is a whiteflash before the video starts

### Version
0.34.0

### Testcase
https://codesandbox.io/s/stardust-ui-example-fm1ow

lucivpav:
<del>Please see if PR #1768 resolves this. There is now a background color set. It behaves the same as Semantic UI's `Embed`.</del>

lucivpav:
@jurokapsiar PR #1758 did not resolve this issue

PR #1758, opened 2019-08-05
fix(embed): make component focusable, only show control/placeholder on video

lucivpav:
This PR resolves issue #1767, which states that:
> the pause control button is visible, even if we are using an iframe in Embed component.

This PR also resolves issue #1742 by making the `Embed` component focusable using keyboard.
image

**Other changes**
* <del>control button, placeholder is not shown when `video` prop not set</del>
* <del>background set to `colorScheme.brand.background`</del>
* added YouTube example

**TODO**
* accessibility is broken even on *master*, this will be fixed in another PR


sophieH29 (approved):

alinais (approved):

lucivpav:
Let's first fix the control button of `Embed` component #1768

lucivpav:
@alinais please review this PR again, it has changed quite a bit since last time

alinais (approved):

packages/react/src/themes/teams/components/Embed/embedVariables.ts
```diff
@@ -13,4 +14,5 @@ export default (siteVariables): EmbedVariables => ({
   width: undefined,
   height: undefined,
   focusBorderColor: siteVariables.colors.brand[500],
+  backgroundColor: siteVariables.colorScheme.brand.background,
```
miroslavstastny: Why this?
Now when you play a video, because of the delay, there is a purple rectangle displayed.

Plus embed's `span` is 5px higher than the iframe, causing purple line to be displayed bellow the iframe:
image

To address this I would create a separate issue.

miroslavstastny (changes requested):

packages/react/src/themes/teams/components/Embed/embedVariables.ts
miroslavstastny: https://stackoverflow.com/questions/21025319/iframe-creates-extra-space-below

lucivpav: This was a poor attempt to resolve issue #1586. I have reverted it, let's address it by another PR in the future.

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -165,7 +165,7 @@ class Embed extends AutoControlledComponent<WithAsProp<EmbedProps>, EmbedState>
                 },
               },
             })}
-            {Box.create(iframe, { defaultProps: { as: 'iframe' } })}
+            {Box.create(iframe, { defaultProps: { as: 'iframe', style: { display: 'block' } } })}
```
miroslavstastny: This is the correct fix but I don't like the implementation.
1. not `style` but `styles`,
2. do not write the styles directly here but rather add a new `iframe` slot to `embedStyles` and here just pass `styles.iframe` (For inspiration see the usage of `styles.control` below)

miroslavstastny (approved):
👍

CHANGELOG.md
```diff
@@ -31,6 +31,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Restore docs for `Ref` component @layershifter ([#1777](https://github.com/stardust-ui/react/pull/1777))
 - Add prototype for expandable control messages in `Chat` @sophieH29 ([#1765](https://github.com/stardust-ui/react/pull/1765))
 
+### Fixes
+- Make `Embed` focusable via keyboard @lucivpav ([#1758](https://github.com/stardust-ui/react/pull/1758))
```
miroslavstastny: Move this up to the other Fixes

PR #1763, opened 2019-08-06
docs(variables): use color picker in theme it

levithomason:
image


docs/src/components/ComponentDoc/ComponentExample/ComponentExampleVariables.tsx
```diff
@@ -21,13 +21,23 @@ type ComponentExampleVariablesProps = {
   usedVariables: Record<string, string[]>
 }
 
+const getGroupName = (variableName: string): string => {
+  if (/^(font|letter|line)/i.test(variableName)) return 'Typography'
+  if (/color/i.test(variableName)) return 'Colors'
```
mnajdova, line 26: Let's not show the undefined values from the color palette, because if one of them is picked, the variable disappears from the form..

docs/src/components/ComponentDoc/ComponentExample/ComponentExampleColorPicker.tsx
```diff
@@ -0,0 +1,76 @@
...
+  ComponentExampleColorPickerProps
+> = props => {
+  const { onChange, variableValue } = props
+  const { theme } = React.useContext<ProviderContextPrepared>(ThemeContext)
+
+  const handleClick = (colorValue: string) => () => {
+    onChange(colorValue)
+  }
+  // Some colors are strings only, i.e. black and white
+  const filteredColors = _.pickBy(theme.siteVariables.colors, _.isPlainObject)
```
mnajdova, line 25: It would be great in the future if we have here color scheme values :)

docs/src/components/ComponentDoc/ComponentExample/ComponentExampleColorPicker.tsx
```diff
@@ -0,0 +1,76 @@
...
+          <strong style={{ display: 'block', width: '12ch', background: '#fff' }}>
+            {_.startCase(colorName)}
+          </strong>
+
+          {_.map(colorShades, (shadeValue, shadeName) => {
+            const isActive = variableValue === shadeValue
+            const contrastColor = Color(shadeValue).isDark() ? '#fff' : '#000'
+
+            return (
+              <div
```
mnajdova, line 51: Can we have a title here with the actual color value?

layershifter: image

Added 👍

docs/src/components/ComponentDoc/ComponentExample/ComponentExampleVariables.tsx
layershifter: Fixed, thank you 👍

layershifter (approved):

Issue #1762, opened 2019-08-06
Remove Font Awesome icons from docs


lucivpav:
### Problem description
There are examples in docs that use Font Awesome icons. If such an example snippet is run on e.g. Codesandbox, the icons are not rendered at all, due to missing Font Awesome. The Font Awesome should not be used in docs at all.

### Proposed solution
1. Do not merge with Font Awesome theme in docs
2. All icon usages are from Teams theme
3. Icons that are not in Teams theme, but are used in docs, should be added to Teams theme

layershifter:
Was fixed in #1764.

PR #1764, opened 2019-08-06
feat(docs): remove Font Awesome icons from examples

lucivpav:
**Overview**
There are examples in docs that use Font Awesome icons. If such an example snippet is run on e.g. Codesandbox, the icons are not rendered at all, due to missing Font Awesome. The Font Awesome should not be used in docs at all. Relevant issue #1762

**Changes**
1. <del>Font Awesome theme is no longer merged with `Teams` theme in `app.tsx` of docs</del>
2. All icon usages in code examples are from Teams theme
3. <del>GitHub icon has been added to `Teams` theme</del>
4. arrow-left, arrow-right were added to `Teams` theme (they are taken from `ProcessedIcons` folder)

<del>**TODO**
We need the following icons, that are currently missing from the `Teams` theme (they are also not in processedIcons):</del>
* <del>Copy, Copy outline</del>
* <del>Spinner</del>
* <del>Align right</del>
* <del>Minus</del>
* <del>Clock</del>
* <del>Thumbs down (aka Dislike)</del>
* <del>Folder</del>
* <del>Folder open</del>
* <del>User</del>


levithomason:
Quick clarification, the Teams theme shouldn't use FontAwesome icons.  However, other themes may use FontAwesome icons.  It is also OK for the doc site itself to use FontAwesome icons.

I think only the actual examples files in `docs/src/examples` should avoid FontAwesome icons, and only for now while we work out icon name inconsistency issues.  However, this is also a temporary solution since switching the theme to anything other than Teams will also break once all the icons are switched to Teams.

Suggestions and reasoning:
- **Move forward with removing FontAwesome icon names from _examples_ only**
Just a temporary solution since the main theme being worked on is Teams right now.  This will have to updated again in the near future as soon as we start adding other themes.  Theme switching from Teams to something else should still work.

- **Leave FontAwesome support in Stardust**
Other themes likely will use it since it is the most popular and has a free license.  Most likely, will provide a base theme which also uses FontAwesome.

- **Leave the doc site on FontAwesome icons**
Stardust is not a Microsoft product but a general open source library.  We don't want to associate our docs with Microsoft's product glyphs.

levithomason (dismissed):
See above

lucivpav:
@levithomason I have updated this PR. Only the example code snippets are using Teams icons instead of Font Awesome icons. The usages of missing icons from Teams theme were replaced by other meaningful icons from Teams, so that we don't need to ask for new icons from the design team.

docs/src/examples/components/Popup/Variations/PopupExampleOffset.shorthand.tsx
```diff
@@ -4,7 +4,8 @@ import { Button, Grid, Popup } from '@stardust-ui/react'
 const renderButton = rotateArrowUp => (
   <Button
     icon={{
-      name: 'arrow circle up',
+      name: 'arrow-up',
+      circular: true,
```
miroslavstastny, line 8: `circular` without `bordered` does not do anything visible. I would rather remove `circular`.
Also think about unifying Popup examples to use the same visual representation among examples.

docs/src/examples/components/Popup/Variations/PopupExamplePosition.shorthand.tsx
```diff
@@ -53,7 +58,7 @@ const positionAndAlignValues = [
   'after-bottom',
 ]
 
-const icons: Record<Position, string> = {
+const iconNames: Record<Position, string> = {
   above: 'arrow circle up',
```
miroslavstastny: These are still FA

docs/src/examples/components/Popup/Variations/PopupExampleOffset.shorthand.tsx
```diff
@@ -4,7 +4,8 @@ import { Button, Grid, Popup } from '@stardust-ui/react'
...
       styles: { transform: `rotate(${rotateArrowUp})` },
```
miroslavstastny: No need for styles, `Icon` has `rotate` prop

packages/react/src/themes/teams/components/Icon/svg/icons/index.ts
```diff
@@ -153,6 +155,8 @@ export default {
   add,
   'arrow-up': arrowUp,
   'arrow-down': arrowDown,
+  'arrow-left': arrowLeft,
```
miroslavstastny, line 158: Let only design add icons. Use `arrow-up` and rotate it as needed.

docs/src/views/Accessibility.tsx
```diff
@@ -152,8 +152,8 @@ export default () => (
     <CodeSnippet
       value={`
         <>
-          <Button icon='envelope' />
-          <Button icon='envelope' aria-label='Send message' />
+          <Button icon='email' />
```
miroslavstastny, line 155: As @levithomason mentioned - this is general doc, I would stick with FA here.

docs/src/examples/components/Popup/Variations/PopupExampleOffset.shorthand.tsx
lucivpav: In order to unify the Popup examples, I have decided to keep this icon `circular` and `bordered`.

packages/react/src/themes/teams/components/Icon/svg/icons/index.ts
lucivpav: The two icons `arrow-left`, `arrow-right` are simply copied from `ProcessedIcons` folder. I did not design them.

docs/src/views/Accessibility.tsx
lucivpav: Are you sure? If user copies these code examples into the `codesandbox` template, the icons won't render.

miroslavstastny: I am not sure, but let's go with what you have now.

miroslavstastny (approved):
Make sure you accept screen changes before merging.

PR #1765, opened 2019-08-06
feat(Chat): Add expandable control messages prototype

sophieH29:
Based on @mnajdova 's work https://github.com/stardust-ui/react/pull/1687

**Collapsed:**
image

**Expanded:**
image


docs/src/prototypes/chatMessages/ControlMessages/ControlMessage.tsx
```diff
@@ -0,0 +1,38 @@
+import * as React from 'react'
+import { Ref, ChatMessageProps, Chat, Icon } from '@stardust-ui/react'
+
+interface ControlMessageProps {
+  focused?: boolean
+  icon?: boolean
+  message: ChatMessageProps
+}
+class ControlMessage extends React.Component<ControlMessageProps> {
```
jurokapsiar, line 9: I'm wondering if it would make sense to use createComponent here. It might be a nice example.

docs/src/prototypes/chatMessages/ControlMessages/GroupControlMessages.tsx
```diff
@@ -0,0 +1,107 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+import * as keyboardKey from 'keyboard-key'
+
+import { List, chatBehavior, Accessibility, Flex, Icon } from '@stardust-ui/react'
+import ControlMessage from './ControlMessage'
+
+const overridenChatBehavior: Accessibility<any> = props => {
```
jurokapsiar: `controlMessagesGroupBehavior`?

docs/src/prototypes/chatMessages/ControlMessages/GroupControlMessages.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const behaviorData = chatBehavior(props)
+
+  behaviorData.attributes.root = {
```
jurokapsiar: I think using `_.merge()` might be safer than spreading

docs/src/prototypes/chatMessages/ControlMessages/GroupControlMessages.tsx
```diff
@@ -0,0 +1,107 @@
...
+      ...behaviorData.focusZone.props,
+      shouldFocusOnMount: true,
+      shouldFocusInnerElementWhenReceivedFocus: true,
+      defaultTabbableElement: undefined,
+    },
+  }
+  return behaviorData
+}
+
+const GroupControlMessages = () => {
```
jurokapsiar: please consider using `createComponent` in this case as well - it might help to standardize styling. Also, it migh allow you to move keyboard handling to the behavior

sophieH29: we won't be able to move keyboard handling, as `actionHandlers` are not yet supported in `createComponent` thus we can't move it to behaviors. So `createComponent` might be useful only for styles 🤷‍♀ @mnajdova what do you think?

docs/src/prototypes/chatMessages/ControlMessages/ControlMessage.tsx
sophieH29: discussed w/ @mnajdova , doesn't make much sense here

docs/src/prototypes/chatMessages/ControlMessages/GroupControlMessages.tsx
sophieH29: Can't agree that it is safer. Because the 2nd param of `_.merge()` can be any object, there is no type, basically, you can merge the behavior to anything. If I would want to use for that object type as `AccessibilityDefinition` it will require me to fill in required props, for e.g. `FocusZone`'s mode, which would be redundant.

docs/src/prototypes/chatMessages/ControlMessages/ControlMessage.tsx
```diff
@@ -0,0 +1,38 @@
...
+    return (
+      <>
+        {this.props.icon ? <Icon name="participant-add" /> : null}
+        <Ref innerRef={this.messageRef}>
+          <Chat.Message
+            {...this.props.message}
+            styles={{
+              padding: 0,
+              marginLeft: '10px',
+              backgroundColor: '#f3f2f1',
```
mnajdova: Can we have Provider around this prototype, and defined overrides there? That's way it would be more obvious what is happening if we replace this color with `colorScheme.default.background2` for example

docs/src/prototypes/chatMessages/ControlMessages/GroupControlMessages.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+const controlMessagesGroupBehavior: Accessibility<any> = props => {
+  const behaviorData = chatBehavior(props)
+
+  behaviorData.attributes.root = {
+    ...behaviorData.attributes.root,
+    'data-is-focusable': true,
+  }
+
+  behaviorData.focusZone = {
```
mnajdova: maybe it would be more readable if we replace only the props here, not the whole `focusZone` object.

docs/src/prototypes/chatMessages/ControlMessages/GroupControlMessages.tsx
```diff
@@ -0,0 +1,107 @@
+import * as React from 'react'
+import * as _ from 'lodash'
+import * as keyboardKey from 'keyboard-key'
+
+import { List, chatBehavior, Accessibility, Flex, Icon } from '@stardust-ui/react'
+import ControlMessage from './ControlMessage'
+
+const controlMessagesGroupBehavior: Accessibility<any> = props => {
```
mnajdova: Let's separate this in a separate file.

docs/src/prototypes/chatMessages/ControlMessages/GroupControlMessages.tsx
```diff
@@ -0,0 +1,107 @@
...
+      ...behaviorData.focusZone.props,
+      shouldFocusOnMount: true,
+      shouldFocusInnerElementWhenReceivedFocus: true,
+      defaultTabbableElement: undefined,
+    },
+  }
+  return behaviorData
+}
+
+const GroupControlMessages = () => {
```
mnajdova: Can we have here props with items param, and we can use this component in a more realistic way. We can extract the items in a separate file, called mockData or something similar.

docs/src/prototypes/chatMessages/ControlMessages/GroupControlMessages.tsx
```diff
@@ -0,0 +1,107 @@
...
+        styles: { padding: 0, display: 'block', minHeight: '25px' },
+      }
+    })
+  }
+
+  return (
+    <Flex
+      onKeyDown={e => {
+        const eventCode = keyboardKey.getCode(e)
+        if (eventCode === keyboardKey.Enter) {
```
mnajdova, line 32: Should we handle Space here as well?

docs/src/prototypes/chatMessages/ControlMessages/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+const ChatExample = () => {
+  const items: ShorthandCollection<ChatItemProps> = [
+    {
+      message: {
+        content: (
+          // Adding control message
+          <ControlMessage
+            icon={true}
+            message={{
+              content: (
```
mnajdova: We can avoid the content here after the @layershifter's refactoring of the factories. You can immediately use the <div> element inside the message :)

docs/src/prototypes/chatMessages/ControlMessages/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+          <GroupControlMessages />
+        ),
+        styles: {
+          marginLeft: 0,
+        },
+      },
+    },
+    {
+      gutter: {
+        content: <Avatar {...janeAvatar} />,
```
mnajdova: See previous comment

packages/react/src/lib/accessibility/Behaviors/Chat/chatMessageBehavior.ts
```diff
@@ -20,6 +20,7 @@ const chatMessageBehavior: Accessibility = () => ({
   attributes: {
     root: {
       [IS_FOCUSABLE_ATTRIBUTE]: true,
+      tabIndex: -1,
```
mnajdova, line 23: Is this change for the prototyping or a real fix?

docs/src/prototypes/chatMessages/ControlMessages/GroupControlMessages.tsx
sophieH29: Only `Enter`, based on current behavior in WebClient

docs/src/prototypes/chatMessages/ControlMessages/index.tsx
sophieH29: hmm, will try

packages/react/src/lib/accessibility/Behaviors/Chat/chatMessageBehavior.ts
sophieH29: it's additional safety thing. Helps both for prototype and overall behavior. Chat Message should be always focusable, meaning - to have set `tabIndex=-1`. When component re-renders based on state change, `FocusZone` won't pick this change up unless you move arrow keys, so the chat message will be rendered without tabIndex - thus not focusable

docs/src/prototypes/chatMessages/ControlMessages/ControlMessage.tsx
sophieH29: I was thinking about this. Will do

docs/src/prototypes/chatMessages/ControlMessages/ControlMessage.tsx
```diff
@@ -0,0 +1,30 @@
...
+      this.messageRef.current.focus()
+    }
+  }
+
+  render() {
+    return (
+      <>
+        {this.props.icon ? <Icon name="participant-add" /> : null}
+        <Ref innerRef={this.messageRef}>
+          <Chat.Message {...this.props.message} className="control-message" />
```
mnajdova: Let's follow the convention for the other stardust component's classNames. It should be something like `ui-chat__message_control`

mnajdova (approved):
See one final comment, looks good otherwise :)

Issue #1767, opened 2019-08-07
Embed iframe should not show control icon when active


lucivpav:
### Steps
Currently, the pause control button is visible, even if we are using an `iframe` in `Embed` component.

### Expected Result
![image](https://user-images.githubusercontent.com/10930160/62621506-463aeb00-b91c-11e9-83c3-3bb17013b8a8.png)

### Actual Result
![image](https://user-images.githubusercontent.com/10930160/62616798-ad52a280-b910-11e9-91f7-c02afbb06c2f.png)

### Version
0.35.0

### Testcase
https://codesandbox.io/s/l7v94?module=/example.js

PR #1768, opened 2019-08-07
fix(Embed): do not show control when not appropriate

lucivpav:
This PR resolves issue #1767, which states that:
> the pause control button is visible, even if we are using an iframe in Embed component.

**Changes**
* control button is not shown when not appropriate
* background set to <del>`grey`</del> `colorScheme.brand.background`
* added YouTube example
* changed icon from Play to Search, on the example with map


docs/src/examples/components/Embed/Types/index.tsx
```diff
@@ -10,6 +10,16 @@ const Types = () => (
...
       examplePath="components/Embed/Types/EmbedExample"
     />
+    <ComponentExample
+      title="IFrame"
+      description="An embed component can embed any iframe."
+      examplePath="components/Embed/Types/EmbedExampleIframe"
+    />
+    <ComponentExample
+      title="YouTube"
+      description="An embed component can embed a YouTube video by iframe."
```
layershifter, line 10: Let's move this under `Usage` section.
And keep `Iframe` under `Slots`

docs/src/examples/components/Embed/Types/EmbedExampleIframe.shorthand.tsx
```diff
@@ -10,6 +10,7 @@ const EmbedExampleIframe = () => (
       height: 450,
       width: 600,
     }}
+    control={{ name: 'search' }}
```
layershifter:
```suggestion
    control='search'
```
nit: can be just icon's name

packages/react/src/themes/teams/components/Embed/embedStyles.ts
```diff
@@ -13,6 +13,7 @@ export default {
       width: v.width,
       height: v.height || 'auto',
       outline: 0,
+      background: 'grey',
```
layershifter: Let's create a variable instead and use some color from color schema

docs/src/examples/components/Embed/Types/index.tsx
lucivpav: Both IFrame example and video example belong under Slots section right?

layershifter: Yep, no `Types`. It doesn't have sense there

lucivpav:
These changes will be integrated in #1758 instead.

PR #1779, opened 2019-08-09
feat: flat Tree component

silviuaavram:
Since now this tree may be virtualized, subtree open/closed state should be kept only on the main `Tree`, as other Items/Titles/Trees can be removed during the virtualisation process. Also removed sub-trees altogether and kept only the first Tree.

The Tree renders TreeItems which in turn render TreeTitles. open/close logic is kept on the main `Tree` state in a Map. Each item has an id (map key) and we keep its ref and open/close state as map object value.

At Tree constructor time, all items will receive props that should not be changed by the user, as they are used in aria attribute generation, keyboard navigation and collapsing/expansion. These are: level, siblings, index, parent. Each item will receive these props.

Each item will pass level, treeSize, index to the TreeTitle (for aria attribute generation in the behavior).

`id` is now required from the user for reach tree item.


docs/src/examples/components/Tree/Types/TreeTitleCustomizationExample.shorthand.tsx
```diff
@@ -0,0 +1,42 @@
...
+        key: '5',
+        title: 'two one',
+      },
+    ],
+  },
+]
+
+const titleRenderer = (Component, { content, open, hasSubtree, ...restProps }) => (
+  <Component open={open} hasSubtree={hasSubtree} {...restProps}>
+    {hasSubtree && <Icon name={open ? 'arrow down' : 'arrow right'} />}
```
lucivpav: Same here.

docs/src/examples/components/Tree/Types/TreeExclusiveExample.shorthand.tsx
```diff
@@ -0,0 +1,52 @@
...
+        key: '5',
+        title: 'two one',
+      },
+    ],
+  },
+]
+
+const titleRenderer = (Component, { content, open, hasSubtree, ...restProps }) => (
+  <Component open={open} hasSubtree={hasSubtree} {...restProps}>
+    {hasSubtree && <Icon name={open ? 'arrow down' : 'arrow right'} />}
```
lucivpav: Let's only use icons from the Teams theme in code examples. Replace by 'triangle-down' and 'triangle-right'.

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,250 @@
...
+      items,
+      title,
+      renderItemTitle,
+      open,
+      level,
+      siblings,
+      indexInTree,
+      indexInSubtree,
+    } = this.props
+    const hasSubtree = !_.isNil(items)
```
lucivpav: What if the collection is empty?

packages/react/src/lib/accessibility/Behaviors/Tree/treeBehavior.ts
```diff
@@ -0,0 +1,39 @@
...
+    attributes: {
+      root: {
+        role: 'tree',
+        'aria-labelledby': props['aria-labelledby'],
+      },
+    },
+    keyActions: {
+      root: {
+        expandSiblings: {
+          keyCombinations: [{ keyCode: keyboardKey['*'] }],
```
lucivpav, line 24: Does this mean that the siblings are expanded no matter which key is pressed?

packages/react/src/themes/teams-dark/components/Tree/treeTitleVariables.ts
```diff
@@ -0,0 +1,7 @@
+import { HierarchicalTreeTitleVariables } from '../../../teams/components/HierarchicalTree/hierarchicalTreeTitleVariables'
+
+export default (siteVars: any): HierarchicalTreeTitleVariables => {
```
lucivpav: Why not just `TreeTitleVariables`?

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+  constructor(props, context) {
+    super(props, context)
+
+    const activeItems = new Map()
+
+    if (props.items) {
+      const setItemsLevelAndSize = (items = props.items, level = 1, parent?) => {
+        items.forEach((item: ShorthandValue<TreeItemProps>, index: number) => {
+          item['level'] = level
```
sophieH29: how about creating a type for such structure for type safety?

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,271 @@
...
+    onTitleClick: (
+      e: React.SyntheticEvent,
+      treeItemProps: TreeItemProps,
+      predefinedProps: TreeItemProps,
+    ) => {
+      const { activeItems } = this.state
+      const itemId = treeItemProps.id
+      const activeItemValue = activeItems.get(itemId)
+
+      if (treeItemProps.items) {
```
sophieH29: what if `items` is an empty array? Are we ok of expanding an empty subtree?

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,271 @@
...
+    onFocusParent: (e: React.SyntheticEvent, treeItemProps: TreeItemProps) => {
+      const { parent } = treeItemProps
+
+      if (!parent) {
+        return
+      }
+
+      const { activeItems } = this.state
+
+      activeItems.get(parent['id']).element.focus()
```
sophieH29: might be good to check if `element` and `focus` exist

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,271 @@
...
+    },
+    onFocusFirstChild: (e: React.SyntheticEvent, treeItemProps: TreeItemProps) => {
+      const { items } = treeItemProps
+
+      if (!items) {
+        return
+      }
+
+      const { activeItems } = this.state
+      const firstChildElement = activeItems.get(items[0]['id']).element
```
sophieH29: might be good to check `activeItems.get(items[0]['id'])` !== NULL

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+      _.invoke(predefinedProps, 'onSiblingsExpand', e, treeItemProps)
+    },
+  })
+
+  renderItems(items: ShorthandCollection<TreeItemProps>) {
+    const { activeItems } = this.state
+
+    return items.reduce((renderedItems, item, index) => {
+      const isSubtree = !!item['items']
```
sophieH29: isn't it more `hasSubtree`? As I understood, you're checking if an item has items, meaning has a subtree

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,271 @@
...
+      const isSubtreeOpen = isSubtree && activeItems.get(item['id']).open
+      const renderedItem = TreeItem.create(item, {
+        defaultProps: {
+          className: Tree.slotClassNames.item,
+          open: isSubtreeOpen,
+        },
+        overrideProps: this.handleTreeItemOverrides,
+      })
+      const renderedFinalItem =
+        isSubtree || isFirstChild ? (
```
sophieH29: what does it mean final item? Would be great to have some comments on explaining what and why is going to be rendered

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,271 @@
...
+      return [
+        ...(renderedItems as any[]),
+        renderedFinalItem,
+        ...[isSubtreeOpen ? this.renderItems(item['items']) : []],
+      ]
+    }, [])
+  }
+
+  renderContent() {
+    const { items } = this.props
```
sophieH29: do you need this method wrapper to check if `items` is not null? Why not check it in `renderItems` method

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,271 @@
...
+    super(props, context)
+
+    const activeItems = new Map()
+
+    if (props.items) {
+      const setItemsLevelAndSize = (items = props.items, level = 1, parent?) => {
+        items.forEach((item: ShorthandValue<TreeItemProps>, index: number) => {
+          item['level'] = level
+          item['siblings'] = items
+          item['index'] = index
```
sophieH29: `level`, `siblings`, `id` and so on exist as props on `TreeItemProps`. Why not use `item.level` to have type validation?

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,240 @@
...
+}
+
+export interface TreeItemProps extends UIComponentProps, ChildrenComponentProps {
+  /** Accessibility behavior if overridden by the user. */
+  accessibility?: Accessibility
+
+  /** Only allow one subtree to be open at a time. */
+  exclusive?: boolean
+
+  id?: string
```
sophieH29: comment missing as for the props below

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,240 @@
...
+  renderContent() {
+    const { items, title, renderItemTitle, open, level, siblings, index } = this.props
+    const hasSubtree = !_.isNil(items)
+
+    return TreeTitle.create(title, {
+      defaultProps: {
+        className: TreeItem.slotClassNames.title,
+        open,
+        hasSubtree,
+        as: hasSubtree ? 'span' : 'a',
```
sophieH29, line 197: we can consider to have it as part of `treeTitleBehavior` to define if render `span` or `a`

packages/react/src/lib/accessibility/Behaviors/Tree/treeBehavior.ts
sophieH29: That means that all siblings are expanded when `*` is pressed. But ARIA design pattern

layershifter:
### `exclusive` is broken

bug-tree

docs/src/examples/components/Tree/index.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
+import Types from './Types'
+
+const TreeExamples = () => (
+  <div>
```
layershifter:
```suggestion
  <>
```
_nit_

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,271 @@
...
+    item: `${Tree.className}__item`,
+  }
+
+  static propTypes = {
+    ...commonPropTypes.createCommon({
+      content: false,
+    }),
+    activeItemsList: customPropTypes.every([
+      customPropTypes.disallow(['children']),
+      PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.number), PropTypes.number]),
```
layershifter: PropTypes do not match the actual implementation

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,271 @@
...
+    ]),
+    defaultActiveItemsList: customPropTypes.every([
+      customPropTypes.disallow(['children']),
+      PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.number), PropTypes.number]),
+    ]),
+    exclusive: PropTypes.bool,
+    items: customPropTypes.collectionShorthand,
+    renderItemTitle: PropTypes.func,
+    rtlAttributes: PropTypes.func,
+    onactiveItemsListChange: PropTypes.func,
```
layershifter: I don't this in props interface, cruft?

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,271 @@
...
+          if (item['items']) {
+            setItemsLevelAndSize(item['items'], level + 1, item)
+            activeItems.set(item['id'], { open: false })
+          }
+        })
+      }
+      setItemsLevelAndSize()
+    }
+
+    this.state = { activeItems }
```
layershifter: I don't see `getDerivedStateFromProps()` or `componentDidUpdate()`, it means that on any props update it will be broken

packages/react/src/components/Tree/Tree.tsx
layershifter: This mutates the original `props.items` and it's definitely unexpected because `props` should be immutable:

image

packages/react/src/components/Tree/Tree.tsx
layershifter:
```tsx
import React from 'react'
import { Button, Divider, Tree } from '@stardust-ui/react'

const TreeExampleShorthand = () => {
  const [items, setItems] = React.useState([
    {
      key: '1',
      title: '1',
    },
  ])

  return (
    <>
      <Button
        content="Add item"
        onClick={() => {
          const newItem = {
            key: items.length + 1,
            title: items.length + 1,
          }

          setItems([...items, newItem])
        }}
      />

      <Divider />
      <Tree items={items} />
    </>
  )
}

export default TreeExampleShorthand
```

kolaps33:
Look on it with JAWS and NVDA, tree works as expected. :)

Just minor issue with reader in  browse mode:
- NVDA - browse mode - not navigate into tree element at all. Fix is if we can add on tree element tabIndex=-1

Generaly for both NVDA and JAWS is good to label tree, otherwise user will only navigate to tree component and reader narrate "Tree view" and then skip whole tree in virtual cursor/browse mode. 
So if the tree is not label user has no clue what tree he reach.

In my opinion, I would add accessibility description something like: 
"For good screen reader experience set aria-label or aria-labelledby attribute on the tree element."

And when I was trying to use "*" to expand all items I could not get it working... Does it work for you?

packages/react/src/components/Tree/TreeItem.tsx
silviuaavram: it was like this in HierarchicalTree but sure I will add an extra check thanks!

packages/react/src/components/Tree/Tree.tsx
silviuaavram: it was the same in the old tree, will add another check. thanks!

lucivpav:
The playground does not seem very useful.
image
How about putting this component into the [blacklist](https://github.com/stardust-ui/react/blob/7a1daca414f32e6202a5da8640dbbc4ed5dd0f65/docs/src/components/ComponentPlayground/ComponentPlayground.tsx#L13)?

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,395 @@
...
+
+  static slotClassNames: TreeSlotClassNames = {
+    item: `${Tree.className}__item`,
+  }
+
+  static propTypes = {
+    ...commonPropTypes.createCommon({
+      content: false,
+    }),
+    activeItems: PropTypes.any,
```
lucivpav:
```suggestion
    activeItems: PropTypes.instanceOf(Map),
```

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,395 @@
...
+    defaultActiveItems: PropTypes.any,
```
lucivpav:
```suggestion
    defaultActiveItems: PropTypes.instanceOf(Map),
```

packages/react/src/components/Tree/TreeTitle.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { ComponentEventHandler, WithAsProp, withSafeTypeForAs } from '../../types'
+
+export interface TreeTitleProps
+  extends UIComponentProps,
+    ChildrenComponentProps,
+    ContentComponentProps {
+  /** Accessibility behavior if overridden by the user. */
+  accessibility?: Accessibility
+
+  /** Whether or not the item has a subtree. */
```
lucivpav:
```suggestion
  /** Whether or not the tree title has a subtree. */
```

packages/react/src/components/Tree/TreeTitle.tsx
```diff
@@ -0,0 +1,110 @@
...
+  hasSubtree?: boolean
+
+  /** The index of the item among its siblings. */
```
lucivpav:
```suggestion
  /** The index of the tree title among its siblings. */
```

packages/react/src/components/Tree/TreeTitle.tsx
```diff
@@ -0,0 +1,110 @@
...
+  index?: number
+
+  /** Level of the tree/subtree that contains this item. */
```
lucivpav:
```suggestion
  /** Level of the tree/subtree that contains this title. */
```

packages/react/src/components/Tree/TreeTitle.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  /**
+   * Called on click.
+   *
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props.
+   */
+  onClick?: ComponentEventHandler<TreeTitleProps>
+
+  /** Whether or not the subtree of the item is in the open state. */
```
lucivpav:
```suggestion
  /** Whether or not the subtree of the tree title is in the open state. */
```

packages/react/src/components/Tree/TreeTitle.tsx
```diff
@@ -0,0 +1,110 @@
...
+  open?: boolean
+
+  /** Array with the ids of the tree item's siblings, if any. */
```
lucivpav:
```suggestion
  /** Array with the ids of the tree title's siblings, if any. */
```

lucivpav (commented):
Can we have some tests please?

docs/src/examples/components/Tree/Types/TreeInitiallyOpenExample.shorthand.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from 'react'
+import { Tree } from '@stardust-ui/react'
+
+const items = [
+  {
+    key: '1',
+    title: 'House Lannister',
+    initialOpen: true,
```
layershifter: What is a purpose of this prop? 😕 

- we use `defaultOpen`/`open` for such things
- at the same time I see `activeItems` in props, but I don't see a controlled usage example

I suggest to remove `initialOpen` and use `activeItems` to avoid multiple patterns.

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,334 @@
...
+  id: string
+  index: number
+  level: number
+  parent: ShorthandValue<TreeItemProps>
+  siblings: ShorthandCollection<TreeItemProps>
+}
+
+export interface TreeState {
+  activeItemIds: string[]
+  itemsForRender: { [s: string]: TreeItemForRenderProps }
```
```suggestion
  itemsForRender: Record<string, TreeItemForRenderProps>
```

_nit_

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,334 @@
...
+  static propTypes = {
+    ...commonPropTypes.createCommon({
+      content: false,
+    }),
+    activeItemIds: customPropTypes.collectionShorthand,
+    defaultActiveItemIds: customPropTypes.collectionShorthand,
+    exclusive: PropTypes.bool,
+    items: customPropTypes.collectionShorthand,
+    renderItemTitle: PropTypes.func,
+    rtlAttributes: PropTypes.func,
```
layershifter: Cruft?

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,334 @@
...
+    }
+  }
+
+  getInitialAutoControlledState() {
+    return { activeItemIds: [] }
+  }
+
+  treeRef = React.createRef<HTMLElement>()
+
+  handleTreeItemOverrides = (predefinedProps: TreeItemProps) => ({
```
layershifter, line 150:
```suggestion
  handleTreeItemOverrides = (predefinedProps: TreeItemProps): TreeItemProps => ({
```

It will also highlight an real issue:

image

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,334 @@
...
+      const { exclusive } = this.props
+
+      if (!Tree.isSubtree(treeItemProps)) {
+        return
+      }
+
+      const indexOfActiveItem = activeItemIds.indexOf(id)
+
+      if (indexOfActiveItem > -1) {
+        activeItemIds.splice(indexOfActiveItem, 1)
```
layershifter: This mutates original array 😥

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,334 @@
...
+      } else {
+        if (exclusive) {
+          siblings.some(sibling => {
+            const activeSiblingIndex = activeItemIds.indexOf(sibling['id'])
+            if (activeSiblingIndex > -1) {
+              activeItemIds.splice(activeSiblingIndex, 1)
```
layershifter: Same mutation...

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,334 @@
...
+              return true
+            }
+            return false
+          })
+        }
+
+        activeItemIds.push(id)
```
layershifter: Same mutation...

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,334 @@
...
+    },
+  })
+
+  renderContent() {
+    const { activeItemIds, itemsForRender } = this.state
+    const { items, renderItemTitle } = this.props
+
+    if (!items) return null
+
+    const renderItems = (items: ShorthandCollection<TreeItemProps>): any[] => {
```
```tsx
    const renderItems = (items: ShorthandCollection<TreeItemProps>): any[] /* !!! */ => {
```

Why this is any? Let's type it properly

packages/react/src/components/Tree/Tree.tsx
silviuaavram: it's our array in our state. why should we care? is it the same reference as the potentially controlled one from props?

packages/react/src/components/Tree/Tree.tsx
layershifter: However, I am not fully understand why we need to have there a callback.

Why we can do simply return?

```tsx
  renderContent() {
    const { activeItemIds, itemsForRender } = this.state
    const { items, renderItemTitle } = this.props

    if (!items) return null
    return items.reduce((renderedItems: any[], item: ShorthandValue<TreeItemProps>) => {
    //
```

silviuaavram: in List and Menu our `renderItems` are typed as `any[]`

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,334 @@
...
+          },
+          overrideProps: this.handleTreeItemOverrides,
+        })
+
+        // Only need refs of the items that spawn subtrees, when they need to be focused
+        // by any of their children, using Arrow Left.
+        const finalRenderedItem = isSubtree ? (
+          <Ref
+            key={item['key'] || item['id']}
+            innerRef={(itemElement: HTMLElement) => {
```
```diff
-            innerRef={(itemElement: HTMLElement) => {
-              handleRef(elementRef, itemElement)
-            }}
+            innerRef={elementRef}
```

Can be simplified

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,334 @@
...
+            ...restItemForRender,
+          },
+          overrideProps: this.handleTreeItemOverrides,
+        })
+
+        // Only need refs of the items that spawn subtrees, when they need to be focused
+        // by any of their children, using Arrow Left.
+        const finalRenderedItem = isSubtree ? (
+          <Ref
+            key={item['key'] || item['id']}
```
```suggestion
            key={item['id']}
```

I am not fully understand the fate of `key` there. Do we need `key`s at all, or only `id`s?

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,234 @@
...
+      e.preventDefault()
+      e.stopPropagation()
+
+      this.handleSiblingsExpand(e)
+    },
+  }
+
+  eventComesFromChildItem = e => {
+    return e.currentTarget !== e.target
+  }
```
layershifter: Cruft?

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+  handleTitleClick = e => {
+    _.invoke(this.props, 'onTitleClick', e, this.props)
+  }
```
layershifter, line 179: _nit:_ All of them are used once, do we need these methods?

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,234 @@
...
+    const hasSubtree = !_.isNil(items) && items.length > 0
+
+    return TreeTitle.create(title, {
+      defaultProps: {
+        className: TreeItem.slotClassNames.title,
+        open,
+        hasSubtree,
+        as: hasSubtree ? 'span' : 'a',
+        level,
+        siblingsLength: siblings.length,
```
layershifter: No solution, but I don't like to have accessibility only props.

packages/react/src/lib/accessibility/Behaviors/Tree/treeItemBehavior.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Triggers 'collapse' action with 'ArrowLeft' on 'root', when has an opened subtree.
+ * Triggers 'expand' action with 'ArrowRight' on 'root', when has a closed subtree.
+ * Triggers 'focusFirstChild' action with 'ArrowRight' on 'root', when has an opened subtree.
+ */
+const treeItemBehavior: Accessibility<TreeItemBehaviorProps> = props => ({
+  attributes: {
+    root: {
+      role: 'none',
+      ...(props.items &&
+        props.items.length && {
```
layershifter: Leaking React abstraction... Let's compute `hasSubtree` on state instead

packages/react/src/components/Tree/Tree.tsx
layershifter: It can cause the same issues as are described there:
https://daveceddia.com/why-not-modify-react-state-directly/

Because this code actually does:
```tsx
this.state.activeItemIds.splice(indexOfActiveItem, 1)
```

packages/react/src/components/Tree/Tree.tsx
layershifter:
```tsx
    const renderItems = (items: ShorthandCollection<TreeItemProps>): React.ReactElement[] => {
      return items.reduce((renderedItems: React.ReactElement[], item: ShorthandValue<TreeItemProps>) => {
```

We checked and this works properly, it hightlights another issue typings under factories will create an issue for it.

layershifter: https://github.com/stardust-ui/react/issues/1865

packages/react/src/components/Tree/Tree.tsx
silviuaavram: not sure what this means

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,331 @@
...
+
+export interface TreeProps extends UIComponentProps, ChildrenComponentProps {
+  /** Accessibility behavior if overridden by the user. */
+  accessibility?: Accessibility
+
+  /** Map with the subtrees and information related to their open/closed state. */
+  activeItemIds?: string[]
+
+  /** Initial activeIndex value. */
+  defaultActiveItemIds?: string[]
```
sophieH29, line 42: typo in comment `activeIndex` -> `activeItemIds`

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,331 @@
...
+        ]
+      } else {
+        if (exclusive) {
+          siblings.some(sibling => {
+            const activeSiblingIdIndex = activeItemIds.indexOf(sibling['id'])
+            if (activeSiblingIdIndex > -1) {
+              activeItemIds = [
+                ...activeItemIds.slice(0, activeSiblingIdIndex),
+                ...activeItemIds.slice(activeSiblingIdIndex + 1),
+              ]
```
sophieH29: lines 163-169 and 173-178 are duplicated, you can export it to function thus will be easier to read

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,331 @@
...
+
+  treeRef = React.createRef<HTMLElement>()
+
+  handleTreeItemOverrides = (predefinedProps: TreeItemProps) => ({
+    onTitleClick: (e: React.SyntheticEvent, treeItemProps: TreeItemProps) => {
+      let { activeItemIds } = this.state
+      const { id, siblings } = treeItemProps
+      const { exclusive } = this.props
+
+      if (!Tree.hasSubtree(treeItemProps)) {
```
sophieH29: nit: I would move it to the very beginning, as anything else above doesn't matter if this fails

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,331 @@
...
+    },
+    onFocusParent: (e: React.SyntheticEvent, treeItemProps: TreeItemProps) => {
+      const { parent } = treeItemProps
+
+      if (!parent) {
+        return
+      }
+
+      const { itemsForRender } = this.state
+      const elementToBeFocused = itemsForRender[parent['id']].elementRef
```
sophieH29: would `itemsForRender[parent['id']]` never return NULL ? just to make sure it is safe

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,331 @@
...
+      const { id, siblings } = treeItemProps
+      const { activeItemIds } = this.state
+
+      siblings.forEach(sibling => {
+        if (Tree.hasSubtree(sibling) && activeItemIds.indexOf(sibling['id']) < 0) {
+          activeItemIds.push(sibling['id'])
+        }
+      })
+
+      if (Tree.hasSubtree(treeItemProps) && activeItemIds.indexOf(id) < 0) {
```
sophieH29: nit: `activeItemIds.indexOf` have been used 5 times to either check if item is active or not. How about to export this to func?
```
isActiveItem = (id): boolean => {
   const { activeItemIds } = this.state
    return activeItemIds.indexOf(id) > -1
}
```

docs/src/examples/components/Tree/Types/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+    <ComponentExample
+      title="Default"
+      description="A default Tree."
+      examplePath="components/Tree/Types/TreeExample"
+    />
+    <ComponentExample
+      title="Custom Title"
+      description="A Tree with customized title rendering."
+      examplePath="components/Tree/Types/TreeTitleCustomizationExample"
+    />
```
layershifter: Let's move this to `Usage` section

docs/src/examples/components/Tree/Types/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+    <ComponentExample
+      title="Exclusive"
+      description="A Tree with only one subtree open at a time."
+      examplePath="components/Tree/Types/TreeExclusiveExample"
+    />
+    <ComponentExample
+      title="Initially Open"
+      description="A Tree with some items initially open."
+      examplePath="components/Tree/Types/TreeInitiallyOpenExample"
+    />
```
layershifter: Let's move this to `Usage` section

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,331 @@
...
+
+export interface TreeSlotClassNames {
+  item: string
+}
+
+export interface TreeProps extends UIComponentProps, ChildrenComponentProps {
+  /** Accessibility behavior if overridden by the user. */
+  accessibility?: Accessibility
+
+  /** Map with the subtrees and information related to their open/closed state. */
```
```suggestion
  /** Ids of opened items. */
```

Not sure that my suggestion is the best, but description should be updated.

packages/react/src/components/Tree/Tree.tsx
layershifter: I mean that it's used nowhere and is missing in types definition. Can we remove it safely?

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,236 @@
...
+    onTitleClick: PropTypes.func,
+    onFocusFirstChild: PropTypes.func,
+    onFocusParent: PropTypes.func,
+    onSiblingsExpand: PropTypes.func,
+    open: PropTypes.bool,
+    parent: customPropTypes.itemShorthand,
+    renderItemTitle: PropTypes.func,
+    siblings: customPropTypes.collectionShorthand,
+    title: customPropTypes.itemShorthand,
+    treeItemRtlAttributes: PropTypes.func,
```
layershifter: Same thing: https://github.com/stardust-ui/react/pull/1779/files#r319033540

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,331 @@
...
+    accessibility: treeBehavior,
+  }
+
+  static autoControlledProps = ['activeItemIds']
+
+  static hasSubtree(item: TreeItemProps | ShorthandValue<TreeItemProps>): boolean {
+    return !_.isNil(item['items']) && item['items'].length > 0
+  }
+
+  static getItemsForRender = _.memoize((itemsFromProps: ShorthandCollection<TreeItemProps>) => {
```
```suggestion
  static getItemsForRender = (itemsFromProps: ShorthandCollection<TreeItemProps>) => {
```

I am really not sure that `_.memoize()` will work properly with nested objects. I suggest to remove it until we will meet a performance bottleneck

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,331 @@
...
+  })
+
+  renderContent(): React.ReactElement[] {
+    const { activeItemIds, itemsForRender } = this.state
+    const { items, renderItemTitle } = this.props
+
+    if (!items) return null
+
+    const renderItems = (items: ShorthandCollection<TreeItemProps>): React.ReactElement[] => {
+      return items.reduce((renderedItems: any[], item: ShorthandValue<TreeItemProps>) => {
```
```suggestion
      return items.reduce((renderedItems: React.ReactElement[], item: ShorthandValue<TreeItemProps>) => {
```

Oops, still any

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -0,0 +1,331 @@
...
+        const finalRenderedItem = isSubtree ? (
+          <Ref key={item['id']} innerRef={elementRef}>
+            {renderedItem}
+          </Ref>
+        ) : (
+          renderedItem
+        )
+
+        return [
+          ...(renderedItems as any[]),
```
```suggestion
          ...renderedItems,
```

Is not required after https://github.com/stardust-ui/react/pull/1779/files#r319034709

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  state = {
+    hasSubtree: false,
+    treeSize: 0,
+  }
+
+  static getDerivedStateFromProps(props: TreeItemProps) {
+    return {
+      hasSubtree: Tree.hasSubtree(props),
```
```
ERROR in packages/react/dist/es/components/Tree/Tree.js -> packages/react/dist/es/components/Tree/TreeItem.js -> packages/react/dist/es/components/Tree/Tree.js
```

This create a circular reference, I suggest to create a simple function as there are no context requirements and move it to `Tree/lib/hasSubtree.ts`:

 ```suggestion
      hasSubtree: hasSubtree(props),
```

File: packages/react/src/themes/teams/components/Tree/treeTitleVariables.ts [hunk 54]
```diff
@@ -0,0 +1,9 @@
+export interface TreeTitleVariables {
+  defaultColor: string
+}
+
+export default (siteVars: any): TreeTitleVariables => {
+  return {
+    defaultColor: siteVars.colors.grey[750],
```
layershifter: https://stardust-ui.github.io/react/color-schemes
This color perfectly maps `foreground` from `default`, can we use it? And we will be to remove variable overrides in Dark and HC themes.

I also suggest to rename it to `color` to match our variable naming convention:

```suggestion
    color: siteVars.colorScheme.default.foreground,
```

packages/react/src/themes/teams-high-contrast/components/Tree/treeTitleVariables.ts
```diff
@@ -0,0 +1,7 @@
+import { TreeTitleVariables } from '../../../teams/components/Tree/treeTitleVariables'
```
layershifter: Can be removed:
https://github.com/stardust-ui/react/pull/1779/files#r319038491

packages/react/src/themes/teams-dark/components/Tree/treeTitleVariables.ts
```diff
@@ -0,0 +1,7 @@
+import { TreeTitleVariables } from '../../../teams/components/Tree/treeTitleVariables'
+
+export default (siteVars: any): TreeTitleVariables => {
+  return {
+    defaultColor: siteVars.colors.white,
+  }
+}
```
layershifter: Can be removed:
https://github.com/stardust-ui/react/pull/1779/files#r319038491

packages/react/src/themes/teams/components/Tree/treeTitleStyles.ts
```diff
@@ -0,0 +1,18 @@
+import { ICSSInJSStyle } from '../../../types'
+import { pxToRem } from '../../../../lib'
+import getBorderFocusStyles from '../../getBorderFocusStyles'
+
+const treeTitleStyles = {
+  root: ({ variables, theme: { siteVariables } }): ICSSInJSStyle => ({
+    padding: `${pxToRem(1)} 0`,
```
layershifter: Can we move this to variables?

```suggestion
    padding: v.padding,
```

packages/react/src/themes/teams/components/Tree/treeTitleStyles.ts
```diff
@@ -0,0 +1,18 @@
...
+    cursor: 'pointer',
+    color: variables.defaultColor,
```
layershifter:
```suggestion
    color: v.color,
```

https://github.com/stardust-ui/react/pull/1779/files#r319038491

packages/react/src/themes/teams/components/Tree/treeTitleStyles.ts
```diff
@@ -0,0 +1,18 @@
+import { ICSSInJSStyle } from '../../../types'
+import { pxToRem } from '../../../../lib'
+import getBorderFocusStyles from '../../getBorderFocusStyles'
+
+const treeTitleStyles = {
+  root: ({ variables, theme: { siteVariables } }): ICSSInJSStyle => ({
```
layershifter:
```suggestion
  root: ({ variables: v, theme: { siteVariables } }): ICSSInJSStyle => ({
```

Let's use `v.*` naming

packages/react/src/themes/teams/components/Tree/treeTitleVariables.ts
```diff
@@ -0,0 +1,9 @@
+export interface TreeTitleVariables {
+  defaultColor: string
```
layershifter:
```suggestion
  color: string
```
https://github.com/stardust-ui/react/pull/1779/files#r319038491

packages/react/test/specs/behaviors/testDefinitions.ts
```diff
@@ -558,7 +558,7 @@ definitions.push({
   regexp: /Triggers '(\w+)' action with '(\w+)' on '([\w-]+)', when has an opened subtree\./g,
   testMethod: (parameters: TestMethod) => {
     const [action, key, elementToPerformAction] = [...parameters.props]
-    const propertyOpenedSubtree = { open: true, items: [{ a: 1 }] }
+    const propertyOpenedSubtree = { open: true, items: [{ a: 1 }], siblings: [], hasSubtree: true }
```
layershifter, line 561:
```suggestion
    const propertyOpenedSubtree = { open: true, hasSubtree: true }
```

I think that is not required anymore

packages/react/src/themes/teams/components/Tree/treeItemStyles.ts
```diff
@@ -0,0 +1,24 @@
+import { ComponentSlotStylesInput, ICSSInJSStyle } from '../../../types'
+import { pxToRem } from '../../../../lib'
+import getBorderFocusStyles from '../../getBorderFocusStyles'
+import { TreeItemProps } from '../../../../components/Tree/TreeItem'
+import TreeTitle from '../../../../components/Tree/TreeTitle'
+
+const treeItemStyles: ComponentSlotStylesInput<TreeItemProps> = {
+  root: ({ theme: { siteVariables }, props: { level } }): ICSSInJSStyle => ({
```
layershifter:
```suggestion
  root: ({ theme: { siteVariables }, props: p } }): ICSSInJSStyle => ({
```

Please use `p.*` naming

packages/react/test/specs/behaviors/testDefinitions.ts
silviuaavram: still required as it will be used for the Hierarchical tree tests.

docs/src/examples/components/Tree/Usage/index.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
+
+const Usage = () => (
+  <ExampleSection title="">
```
layershifter:
```suggestion
  <ExampleSection title="Usage">
```

Oops, missing title

layershifter (approved):
Approved on my side, great job!

***

@miroslavstastny @mnajdova please also take a look

silviuaavram:
Merging this. We can create issues to follow up, if any.

PR #1780, opened 2019-08-09
feat(Provider): add overwrite prop

layershifter:
# ♻️ `overwrite` prop

This PR adds `overwrite` prop for `Provider` component. It allows to reset current context and avoid merging with existing themes. Existing behavior is preserved.

### `overwrite=false`

```jsx
<Provider theme={themes.teams}> // Teams theme
  <Provider theme={themes.base} /> // Teams + Base theme
```

### `overwrite=true`

```jsx
<Provider theme={themes.teams}> // Teams theme
  <Provider overwrite theme={themes.base} /> // Base theme
```

### DIFF
--- a/packages/react/src/components/Provider/Provider.tsx
+++ b/packages/react/src/components/Provider/Provider.tsx
@@ -158,11 +158,11 @@ class Provider extends React.Component<WithAsProp<ProviderProps>> {
       disableAnimations,
       renderer,
     }
+
+    const incomingContext: ProviderContextPrepared = overwrite ? {} : this.context
     // rehydration disabled to avoid leaking styles between renderers
     // https://github.com/rofrischmann/fela/blob/master/docs/api/fela-dom/rehydrate.md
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Provider/Provider.tsx
```diff
@@ -177,7 +181,7 @@ class Provider extends React.Component<WithAsProp<ProviderProps>> {
         target={target}
         {...{ rehydrate: false }}
       >
-        <ThemeProvider theme={outgoingContext}>
+        <ThemeProvider theme={outgoingContext} overwrite>
```
layershifter, line 184: `ThemeProvider` comes from Fela, we should always use **overwrite** as Fela has own shallow merging.

https://github.com/robinweser/fela/blob/v10.6.1/packages/fela-bindings/src/ThemeProviderFactory.js

lucivpav:
Let's test this change.

miroslavstastny (approved):

PR #1793, opened 2019-08-14
fix(Prototypes): fix invoking suggestion when user type "@" in "At mention"

kolaps33:
Hot to test fix:
 invoke "@" by key combination like:
- alt+64 
- ctrl + alt + v (when czech or slovak keyboard layout)


sophieH29 (approved):

docs/src/prototypes/mentions/MentionsEditor.tsx
```diff
@@ -82,8 +82,9 @@ const MentionsEditor: React.FunctionComponent<
     dispatch({ type: 'RESET_UPDATE_FLAG' })
   }, [state.shouldUpdate])
 
-  const handleEditorKeyUp = (e: React.KeyboardEvent) => {
-    if (!state.open && e.shiftKey && keyboardKey.getCode(e) === keyboardKey.AtSign) {
+  const handleEditorKeyChange = (e: any) => {
+    const editorContent = e.target.textContent
+    if (!state.open && editorContent[editorContent.length - 1] === '@') {
```
jurokapsiar: it would be better to check which character is at the cursor position. see for example https://www.techcycle.me/post/reactjs-track-cursor-in-textarea on how to find the selection start/end

docs/src/prototypes/mentions/MentionsEditor.tsx
```diff
@@ -133,7 +134,7 @@ const MentionsEditor: React.FunctionComponent<
       <div
         contentEditable
         ref={contendEditableRef}
-        onKeyUp={handleEditorKeyUp}
+        onInput={event => handleEditorKeyChange(event)}
```
layershifter:
```suggestion
        onInput={handleEditorKeyChange}
```

docs/src/prototypes/mentions/MentionsEditor.tsx
```diff
@@ -82,8 +82,9 @@ const MentionsEditor: React.FunctionComponent<
     dispatch({ type: 'RESET_UPDATE_FLAG' })
   }, [state.shouldUpdate])
 
-  const handleEditorKeyUp = (e: React.KeyboardEvent) => {
-    if (!state.open && e.shiftKey && keyboardKey.getCode(e) === keyboardKey.AtSign) {
+  const handleEditorKeyChange = (e: any) => {
```
layershifter: Please don't use any there, please use valid typings.

docs/src/prototypes/mentions/MentionsEditor.tsx
```diff
@@ -82,8 +82,13 @@ const MentionsEditor: React.FunctionComponent<
...
 
-  const handleEditorKeyUp = (e: React.KeyboardEvent) => {
-    if (!state.open && e.shiftKey && keyboardKey.getCode(e) === keyboardKey.AtSign) {
+  const handleEditorKeyChange = () => {
+    const selectionContent = window.getSelection().anchorNode.nodeValue
+    if (
+      !state.open &&
+      selectionContent &&
+      selectionContent[window.getSelection().focusOffset - 1] === '@'
+    ) {
```
layershifter: Nit, we can rewrite it to make a bit more readable:

```tsx
    const { anchorNode, focusOffset } = window.getSelection()
    const lastCharacter = anchorNode.nodeValue && anchorNode.nodeValue[focusOffset - 1]

    if (!state.open && lastCharacter === '@') {
```

***

But even current changes should be adjusted, I can't close `Dropdown` by `Escape` key or `Backspace` anymore.

bug-input

layershifter: It works now 👍

layershifter:
And let's replace `defaultOpen` with `open` in `docs/src/prototypes/mentions/MentionsDropdown.tsx`.

layershifter (approved):

PR #1794, opened 2019-08-14
fix(FocusZone): Bidirectional - U/D Arrow key navigation bug when hitting grid edge

### LINKED ISSUE #1659: U/D Arrow key navigation bug when hitting grid edge
## Bug Report
When navigating UP and I reach the end of the grid, the focus moves LEFT instead (expected - same behavior with DOWN moving RIGHT). However, after reaching the edge of the grid and UP causes me to navigate further LEFT, behavior of the DOWN key is wrong. I do not end up one space below my current position. Instead, I end up one space down from the last valid UP movement. 

### Steps
1. Refer to below keypad grid. Start at 8, UP UP UP (LEFT) to end up at 1. 
2. Press DOWN. Expect to land on 4. Land on 5 instead.

1 2 3
4 5 6
7 8 9
\* 0 #

Can be repro'd using the grid here: https://stardust-ui.github.io/react/components/grid

### Expected Result
DOWN key after series of UP keys should land focus immediately below current focus.

### Actual Result
DOWN key after series of UP keys lands focus immediately below the result of the last valid UP movement.

### Version
0.34.1

sophieH29:
Fix https://github.com/stardust-ui/react/issues/1659

Let's say we have the next grid
1 2 3
4 5 6
7 8 9

1. Case: the focus is on "5". User presses UP - focus goes to "2". UP again - "1"
- **Prev Behavior**: Press DOWN from "1" - focus goes to "5"
- **Fixed behavior**: Press DOWN from "1" - focus goes to "4"

2. Case: "5" cell is `disabled` or has `data-is-visible` === `true`. Focus is on "2"
- **Prev Behavior**:
      - Press DOWN from "2" - focus goes to "4". Press DOWN from "4" - focus goes to "8"
      - Press UP from "8" - focus goes to "6". Press UP from "6" - focus goes to "2"

- **Fixed Behavior**:
      - Press DOWN from "2" - focus goes to "4". Press DOWN from "4" - focus goes to "7"
      - Press UP from "8" - focus goes to "6". Press UP from "6" - focus goes to "3"

The created issue in Fabric UI https://github.com/OfficeDev/office-ui-fabric-react/issues/10145

### Side note
`Fabric UI` had a PR https://github.com/OfficeDev/office-ui-fabric-react/pull/7277 for not allowing to move to the left when UP pressed (same moving right when DOWN pressed). They eventually closed the PR as such behavior is by design. But this is something we can reconsider and incorporate to our `FocusZone`. That will solve the current bug and simplify the overall behavior. Let me know your thoughts around that.


silviuaavram (approved):

layershifter (approved):

dzearing:
@sophieH29 This probably wasn't the right fix.

In asymetrical tile scenarios, resetting the horizontal centering on up/down would cause the focus to pull to the left or right. This is the scenario I mean:

image

By not resetting the horizontal alignment on vertical movement, it keeps the alignment centered, so that you can down arrow "in the middle of the set" and it stays "in the middle".

dzearing:
I *think* the right fix would be to either prevent the up/down arrow to do things when we hit the edge, or to wrap if the wrap setting is on. Moving horizontally when you press a vertical arrow seems weird anyway.

sophieH29:
> I _think_ the right fix would be to either prevent the up/down arrow to do things when we hit the edge, or to wrap if the wrap setting is on. Moving horizontally when you press a vertical arrow seems weird anyway.

@dzearing I agree on your point that it is rather better to stop navigating horizontally with vertical keys when we hit the edge than the current fix. And I for fixing it in this way.
@jurokapsiar what do you think?

jurokapsiar:
@sophieH29 agree

PR #1807, opened 2019-08-16
fix(Popup|Dialog): ESC should close opened Popup/Dialog if focus is lost and moved to body

### LINKED ISSUE #1709: Dialog: can't close if not focused
### Steps
1. Go to `Dialog` examples: https://stardust-ui.github.io/react/components/dialog
2. Open any `Dialog`
3. Click inside of it, on a header for example
4. Press `Escape` key

### Expected Result
`Dialog` is closed

### Actual Result
`Dialog` is open because focus was moved to `body` while we have an event handler on `Dialog` `root`

### Version
0.34.2

sophieH29:
Fixes https://github.com/stardust-ui/react/issues/1709


lucivpav:
Can we add a test?

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -203,6 +204,20 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
     }
   }
 
+  handleDocumentKeydown = (getRefs: Function) => e => {
```
layershifter:
```suggestion
  handleDocumentKeydown = (getRefs: Function) => (e: KeyboardEvent) => {
```

More type safety 🏋

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -203,6 +204,20 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
...
+    // if focus was lost from Dialog, for e.g. when click on Dialog's content
+    // and ESC is pressed, the opened Dialog should get closed and the trigger should get focus
+    const lastOverlayRef = getRefs().pop()
+    const isLastOpenedDialog: boolean =
+      lastOverlayRef && lastOverlayRef.current === this.overlayRef.current
+
+    const keyCode = keyboardKey.getCode(e)
+    if (keyCode === keyboardKey.Escape && isLastOpenedDialog && this.state.open) {
```
layershifter:
```suggestion
    if (keyboardKey.getCode(e) === keyboardKey.Escape && isLastOpenedDialog) {
```

You don't need `this.state.open` as this event listener will be mounted only when `Dialog` is `open`:

```tsx
<Portal open={open} /* here we are */>
  <EventListener type="keydown" />
</Portal>
```

layershifter (approved):

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -312,6 +326,12 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
                 type="click"
                 capture
               />
+              <EventListener
+                listener={this.handleDocumentKeydown(getRefs)}
+                targetRef={documentRef}
```
layershifter:
```suggestion
                targetRef={targetRef}
```

Issue #1805, opened 2019-08-16
Sidebar not expanded on behavior link


lucivpav:
## Bug Report
When clicking an accessibility link, such as the one for Default Button behavior, the left sidebar is not expanded appropriately.

### Steps
1. Navigate to [https://stardust-ui.github.io/react/behaviors/Button](https://stardust-ui.github.io/react/behaviors/Button)
2. Observe sidebar not expanded

### Expected Result
The result that you expected.
![image](https://user-images.githubusercontent.com/10930160/63154145-4ae45b00-c010-11e9-85a6-41f58fed25c5.png)

### Actual Result
The actual result that happened 💣
![image](https://user-images.githubusercontent.com/10930160/63154126-4455e380-c010-11e9-95db-03e9389d998a.png)

### Version
0.36.1

PR #1811, opened 2019-08-16
docs(sidebar): fix sidebar not expanding on behavior link

lucivpav:
This fixes issue #1805


docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -60,7 +60,9 @@ class Sidebar extends React.Component<any, any> {
     }
 
     return _.findIndex(sections, (section: ShorthandValue<HierarchicalTreeItemProps>) => {
-      return _.find((section as any).items, item => item.title.to.startsWith(newAt))
+      return _.find((section as any).items, item =>
+        new RegExp(`^${newAt}`, 'i').test(item.title.to),
+      )
```
layershifter, line 65: I think that we should fix links instead

lucivpav: I agree. Let's not create a fix, as @kolaps33 is working on a PR that changes the Accessibility tab. With his changes, we should no longer have a way to click on the broken a11y links.

PR #1816, opened 2019-08-19
feat(Focus): focus indicator for additional documents

jurokapsiar:
Add event handlers that detect input type to additional documents as well


packages/react/src/lib/whatInput.ts
```diff
@@ -234,6 +234,23 @@ if (isBrowser() && 'addEventListener' in window && Array.prototype.indexOf) {
...
+
+export const setUpWhatInput = (target: Document) => {
+  if (isBrowser() && 'addEventListener' in window && Array.prototype.indexOf) {
+    const whatInputInitialized = 'whatInputInitialized'
+    if (target[whatInputInitialized] === true) {
+      return
+    }
+    target[whatInputInitialized] = true
+
+    addListeners(target)
```
layershifter: It should use `window`, we can use `defaultView`: https://developer.mozilla.org/en-US/docs/Web/API/Document/defaultView

```suggestion
    addListeners(target.defaultView)
```

layershifter (commented):
There is also `docElem` with calls like:

```tsx
docElem.setAttribute(`data-whatinput`, currentInput)
docElem.setAttribute('data-whatelement', currentElement)
```

Should we also update it?

jurokapsiar:
> Should we also update it?
Agreed offline to handle it separately.

packages/react/src/lib/whatInput.ts
```diff
@@ -82,45 +82,45 @@ const setUp = () => {
   // add correct mouse wheel event mapping to `inputMap`
   inputMap[detectWheel()] = 'mouse'
 
-  addListeners()
+  addListeners(window)
```
sophieH29, line 85: Thinking from a side of simplifying things, curious if it would make sense to add listeners initially to `document` as for these types of events there is no difference `document` or `window`. And then in your method attach additional documents as you do. So we'll refer only to document object and won't diversify between window and document

layershifter: I think that reason to use `window` is conditions like this:
```
if (eventTarget.PointerEvent) {
} else if (window.MSPointerEvent) {
```

packages/react/src/lib/whatInput.ts
```diff
@@ -234,6 +234,23 @@ if (isBrowser() && 'addEventListener' in window && Array.prototype.indexOf) {
   setUp()
 }
 
+/*
+ * set up for document
+ */
+
+export const setUpWhatInput = (target: Document) => {
+  if (isBrowser() && 'addEventListener' in window && Array.prototype.indexOf) {
```
layershifter:
```suggestion
  if (isBrowser() && 'addEventListener' in target && Array.prototype.indexOf) {
```

packages/react/src/lib/whatInput.ts
jurokapsiar: we decided to do the opposite - always attach them to window (or to document.defaultView which is also a Window)

layershifter (approved):

PR #1819, opened 2019-08-19
chore: release 0.36.2

layershifter:
# 0.36.2

Release that contains only required features and fixes.

### Changelog

- Update vulnerable version of lodash dependency @layershifter (#1700)
- Use target from Provider in ReactDOM.createPortal() calls @layershifter (#1810)
- Focus indicator for additional documents @jurokapsiar (#1816)
- Add possibility for a Toolbar to rearrange its items according to space available @miroslavstastny (#1657)


lucivpav:
@layershifter should we merge this into master? Master currently shows v0.36.1, which is confusing because we have already released v0.36.2

layershifter:
I don't have a clear answer there as we definitely can't merge all commits. We can try to merge only related to release and changelog, but I don't see a big benefit there. Release process should be reviewed and improved once we will release `1.0.0`.

PR #1820, opened 2019-08-19
fix(FocusZone|FocusTrapZone): correctly define current document object of the components

### LINKED ISSUE #1817: FocusZone and FocusTrapZone do not respect target in Provider context
## Bug Report

FocusZone and FocusTrapZone do not respect target in Provider context

### Steps
Open https://codesandbox.io/s/stardust-ui-example-1phxh
Try to navigate in focus zone

### Expected Result
Focus moves when arrow keys are used

### Actual Result
Focus does not move

### Version
0.36.1

sophieH29:
Fixes https://github.com/stardust-ui/react/issues/1817

By replacing `document` to `getDocument` from `focusUtilities` we're making sure that the right document object will be used for correct `FocusZone` and `FocusTrapZone` function.

`getDocument` is called with node elements `this._root._current` and implementation looks like this:
```js
export function getDocument(rootElement?: Element | null): Document | undefined {
  return (rootElement && rootElement.ownerDocument) || document
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix `hand` icon in Teams theme @lucivpav ([#1782](https://github.com/stardust-ui/react/pull/1782))
 - Use `target` from `Provider` in `ReactDOM.createPortal()` calls @layershifter ([#1810](https://github.com/stardust-ui/react/pull/1810))
 - ESC key should close the last opened `Popup` or `Dialog` if body has focus @sophieH29 ([#1807](https://github.com/stardust-ui/react/pull/1807))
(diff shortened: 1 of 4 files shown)


jurokapsiar (approved):

lucivpav:
Looks good to me

dzearing (approved):
Yep change makes sense. If this is a bug in the other copy, we need to merge the two asap to stop the branching 😀 or fix the bug in both places.

sophieH29:
> Yep change makes sense. If this is a bug in the other copy, we need to merge the two asap to stop the branching 😀 or fix the bug in both places.

@davezuko totally agree on keeping consistency between copies. We encountered the bug when had a case of opening another window with components through `window.open`, so when `document.active` was used, it would use the wrong (initial) `document` object from the parent window. I'll create a PR on Fabric side to replicate the change.

**Update:** created a PR https://github.com/OfficeDev/office-ui-fabric-react/pull/10187

PR #1821, opened 2019-08-19
feat(alert): add Header and Icon slots

lucivpav:
This PR adds an ability to specify `header` and `icon` props when using an `Alert` component.
Example:
image


docs/src/examples/components/Alert/Variations/index.tsx
```diff
@@ -44,6 +44,16 @@ const Variations = () => (
...
+    <ComponentExample
+      title="Icon"
+      description="An Alert can contain an icon."
+      examplePath="components/Alert/Variations/AlertExampleIcon"
+    />
+    <ComponentExample
+      title="Header"
+      description="An Alert can contain a header."
+      examplePath="components/Alert/Variations/AlertExampleHeader"
+    />
```
layershifter: Please move this to `Slots` part

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -94,10 +107,28 @@ class Alert extends UIComponent<WithAsProp<AlertProps>, AlertState> {
...
+          },
+        })}
+        {Box.create(header, {
+          defaultProps: {
+            className: Alert.slotClassNames.header,
+            ...accessibility.attributes.header,
+            styles: {
+              ...styles.header,
+              fontWeight: 'bold',
+              marginRight: '10px',
```
layershifter: Please move these styles to `alertStyles`

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -36,6 +39,12 @@ export interface AlertProps
   /** Button shorthand for the action slot. */
   action?: ShorthandValue<ButtonProps>
 
+  /** An alert may contain an icon. */
+  icon?: ShorthandValue<IconProps>
+
+  /** An alert may contain a header. */
+  header?: string
```
layershifter: Should be shorthand

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -94,10 +107,28 @@ class Alert extends UIComponent<WithAsProp<AlertProps>, AlertState> {
   }
 
   renderContent = ({ styles, accessibility }: RenderResultConfig<AlertProps>) => {
-    const { action, content } = this.props
+    const { action, icon, header, content } = this.props
 
     return (
       <>
```
jurokapsiar, line 183: alertWarningBehavior adds aria-live to the content slot of the Alert which makes the screen reader read the content. Since now we have header with the textual information as well, it would be good to read that as well. However, that would require us to have nested slots - let's discuss how to deal with it. fyi @layershifter

packages/react/src/components/Alert/Alert.tsx
lucivpav: ShorthandValue<string>?

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -179,20 +185,22 @@ class Alert extends AutoControlledComponent<WithAsProp<AlertProps>, AlertState>
...
-          },
-        })}
-        {Box.create(content, {
-          defaultProps: {
-            className: Alert.slotClassNames.content,
-            styles: styles.content,
-            ...accessibility.attributes.content,
-          },
-        })}
+        <Flex as="div" className="wrapper" {...accessibility.attributes.wrapper} id={wrapperId}>
```
layershifter: We should not use utility components in our internals, consider to create a slot for it.

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -136,11 +140,13 @@ class Alert extends AutoControlledComponent<WithAsProp<AlertProps>, AlertState>
     success: PropTypes.bool,
     visible: PropTypes.bool,
     warning: PropTypes.bool,
+    wrapperId: PropTypes.string,
   }
 
   static defaultProps = {
     accessibility: alertBehavior,
     dismissAction: { icon: 'close' },
+    wrapperId: _.uniqueId('alert-wrapper-'),
```
layershifter: This will be the same for all components, ids should be stored under component's state.

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -87,6 +99,9 @@ export interface AlertProps
 
   /** An alert may be formatted to display a warning message. */
   warning?: boolean
+
+  /** Id of a wrapper containing header and content */
```
jurokapsiar: consider renaming wrapper to body, we use wrapper in other components for wrapping the whole component (root)

packages/react/src/lib/accessibility/Behaviors/Alert/alertWarningBehavior.ts
```diff
@@ -1,17 +1,21 @@
...
-const alertWarningBehavior: Accessibility = () => ({
+const alertWarningBehavior: Accessibility<AlertBehaviorProps> = props => ({
   attributes: {
-    content: {
+    wrapper: {
       role: 'alert',
       'aria-live': 'polite',
     },
+    dismissAction: {
+      'aria-labelledby': props.wrapperId,
```
jurokapsiar: Use aria-describedby instead of aria-labelledby - as label should be something like 'Dismiss' - that needs to come from outside as translations are out of scope for stardust

packages/react/src/lib/accessibility/Behaviors/Alert/alertWarningBehavior.ts
```diff
@@ -1,17 +1,21 @@
 import { Accessibility } from '../../types'
+import AlertBehaviorProps from './alertBehaviorProps'
 
 /**
  * @specification
- * Adds role 'alert' to 'content' slot.
- * Adds attribute 'aria-live=polite' to 'content' slot.
+ * Adds role 'alert' to 'wrapper' element.
+ * Adds attribute 'aria-live=polite' to 'wrpper' element.
```
kolaps33: typo
wrapper

packages/react/src/lib/accessibility/Behaviors/Alert/alertWarningBehavior.ts
kolaps33: @jurokapsiar 
I think it could be pointing to wrapperId.
But it need be there aria-describedby prop not "aria-labelledby"

Then the patter is, depend on close button:
if icon
`<button aria-describedby="wrapperId" aria-label="dismiss"> </button>`
if button has text
`<button aria-describedby="wrapperId"> dismiss </button>`

And if will add this additional line, then we need add specification line as well and verify that is tested. I will help Pavel with it.

packages/react/src/lib/accessibility/Behaviors/Alert/alertWarningBehavior.ts
```diff
@@ -1,17 +1,21 @@
 import { Accessibility } from '../../types'
+import AlertBehaviorProps from './alertBehaviorProps'
 
 /**
  * @specification
- * Adds role 'alert' to 'content' slot.
- * Adds attribute 'aria-live=polite' to 'content' slot.
+ * Adds role 'alert' to 'wrapper' element.
```
jurokapsiar: please keep 'slot' here as it is a slot from the behavior point of view (I understand it is not a slot in the component). This will also make the unit tests work again.

packages/react/src/components/Alert/Alert.tsx
lucivpav: If its a state and not a prop, then how can I retrieve the value in [behavior](https://github.com/stardust-ui/react/pull/1821/files#diff-4757d822a5e7f14212c430689bfb9d1dR17)?

layershifter: Accessibility behavior receive `state` and `props`, so you will have it ✨

lucivpav:
TODO: fix: dismissible is now not at the end of component, but right after text
image
The same issue applies to `actions`

packages/react/src/lib/accessibility/Behaviors/Alert/alertProps.ts
```diff
@@ -0,0 +1,10 @@
+type AlertProps = {
+  /** An alert may be formatted to display a danger message. */
+  danger?: boolean
+  /** An alert may be formatted to display a warning message. */
+  warning?: boolean
+  /** id of the alert body element. */
+  bodyId?: string
+}
+
+export default AlertProps
```
layershifter: Can you please clarify why we need to move out types?..

packages/react/src/lib/accessibility/Behaviors/Alert/alertWarningBehavior.ts
```diff
@@ -1,17 +1,22 @@
...
+const alertWarningBehavior: Accessibility<AlertProps> = props => ({
   attributes: {
-    content: {
+    body: {
       role: 'alert',
       'aria-live': 'polite',
     },
+    dismissAction: {
+      'aria-describedby': props.bodyId,
+      'aria-label': 'dismiss',
```
layershifter: Is it valid to have unlocalized defaults? @jurokapsiar

packages/react/src/themes/teams/components/Alert/alertStyles.ts
```diff
@@ -100,6 +100,16 @@ const alertStyles: ComponentSlotStylesInput<AlertProps, AlertVariables> = {
     margin: v.actionsMargin,
   }),
 
+  header: (): ICSSInJSStyle => ({
+    fontWeight: 'bold',
+    marginRight: '10px',
```
layershifter: Let's create a variable `headerMargin` and `headerFontWeight`. We can also use there `pxToRem()`:

```tsx
headerFontWeight: siteVars.fontWeightBold,
headerMargin: `0 ${pxToRem(10)} 0 0`,
```

packages/react/src/lib/accessibility/Behaviors/Alert/alertProps.ts
```diff
@@ -0,0 +1,10 @@
+type AlertProps = {
+  /** An alert may be formatted to display a danger message. */
+  danger?: boolean
+  /** An alert may be formatted to display a warning message. */
+  warning?: boolean
+  /** id of the alert body element. */
```
layershifter:
```suggestion
  /** Id of the alert body element. */
```

_nit_

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -153,17 +177,44 @@ class Alert extends AutoControlledComponent<WithAsProp<AlertProps>, AlertState>
...
       <>
-        {Box.create(content, {
+        {Icon.create(icon, {
           defaultProps: {
-            className: Alert.slotClassNames.content,
-            styles: styles.content,
-            ...accessibility.attributes.content,
+            className: Alert.slotClassNames.icon,
+            styles: styles.icon,
+            xSpacing: 'after',
```
layershifter:
```suggestion
```

We can move this to styles instead?

packages/react/src/lib/accessibility/Behaviors/Alert/alertProps.ts
lucivpav: Because both `alertBehavior.ts` and `alertWarningBehavior.ts` depends on it.

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -153,17 +177,44 @@ class Alert extends AutoControlledComponent<WithAsProp<AlertProps>, AlertState>
...
+                  },
+                })}
+                {Box.create(content, {
+                  defaultProps: {
+                    className: Alert.slotClassNames.content,
+                    styles: styles.content,
+                    ...accessibility.attributes.content,
+                  },
+                })}
+              </>
```
layershifter: Can we move out this fragment?

```tsx
const bodyContent = (
  <>
  </>
)
//
// ...
//

{Box.create(body, {
   defaultProps: { /* ... */ },
   overrideProps: { children: bodyContent }, // <<<< children should be defined in overrideProps in this case
```

packages/react/src/components/Alert/Alert.tsx
lucivpav: But `xSpacing` is a prop of `Icon`. If I set it in styles prop, the value is not taken into account.

packages/react/src/lib/accessibility/Behaviors/Alert/alertProps.ts
layershifter: Can we keep them `alertBehavior.ts`?

packages/react/src/components/Alert/Alert.tsx
layershifter: ### Problem for styles

It's just a `margin` value. We should avoid styling via props as it created hidden dependencies. We have already discussed this problem and it's related to children API.

```tsx
<Alert icon="book" />
// This things look equal 🤔 
// But rendered result will be different as there are no matching styles for it
<Alert>
  <Icon name="book" />  
</Alert>
// With usage component props we are making it even more complicated
```

### To coupled

If `horizontalSpace` variable in `IconVariables` will be changed, it will accidentally affect this component. Will be it valid?

packages/react/src/lib/accessibility/Behaviors/Alert/alertWarningBehavior.ts
jurokapsiar: no, please remove. User will need to provide this

packages/react/src/lib/accessibility/Behaviors/Alert/alertProps.ts
lucivpav: I added it to `alertWarningBehavior` to avoid circular dependency (`alertBehavior` imports `alertWarningBehavior`)

packages/react/src/components/Alert/Alert.tsx
lucivpav: Fixed

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -153,17 +177,47 @@ class Alert extends AutoControlledComponent<WithAsProp<AlertProps>, AlertState>
   }
 
   renderContent = ({ styles, accessibility }: RenderResultConfig<AlertProps>) => {
-    const { actions, dismissible, dismissAction, content } = this.props
+    const { actions, dismissible, dismissAction, content, icon, header, body } = this.props
 
-    return (
+    const bodyContent = (
       <>
+        {Box.create(header, {
```
layershifter: Is it `Box` or `Text`?

https://github.com/stardust-ui/react/pull/1821/files#diff-55e5002251082bee3582b3d7cad2ceedR57

layershifter (approved):

packages/react/src/components/Alert/Alert.tsx
lucivpav: Changed it to `Text`

PR #1822, opened 2019-08-19
test(sliderBehavior): add description and unit tests

kolaps33:


silviuaavram (approved):

jurokapsiar (approved):

packages/react/src/lib/accessibility/Behaviors/Slider/sliderBehavior.ts
```diff
@@ -1,5 +1,15 @@
...
+/**
+ * @description
+ * Adds attribute 'aria-disabled=true' based on the property 'disabled'.
+ * Adds attribute 'aria-orientation=vertical' or 'horizontal' based on the property 'vertical'.
+ * Adds attribute 'aria-valuemin' with value based on the property 'min'.
+ * Adds attribute 'aria-valuemax' with value based on the property 'max'.
+ * Adds attribute 'aria-valuenow' with value based on the property 'value'.
+ * Adds attribute 'aria-valuetext' with value based on the callback function 'getA11yValueMessageOnChange'.
+ */
+
```
layershifter:
```suggestion

```

nit

packages/react/test/specs/behaviors/sliderBehavior-test.ts
```diff
@@ -0,0 +1,42 @@
+import { sliderBehavior } from 'src/lib/accessibility'
+
+function generatePropertyTest(
+  propName: string,
+  propValue: string | boolean | number,
+  slot: string,
+  expectedAttr: string,
+  customExpResult?: string,
+) {
```
layershifter: Can we use object instead? It's hard to follow all function params

```tsx
generatePropertyTest('disabled', true, 'root', 'aria-disabled')
```
**vs**
```tsx
generatePropertyTest({ 
  propName: 'disabled',
  propValue: true, 
  slot: 'root',
  attr: 'aria-disabled'
})
```

layershifter (approved):

PR #1823, opened 2019-08-19
feat(Alert): add actions

layershifter:
# BREAKING CHANGES

- `action` slot was renamed to `dismissAction`
- to show `dismissAction`, you need to use `dismissible` prop

### Before
```tsx
<Alert action={{ icon: 'close', onClick: this.handleClick('Remove') }} />
```

### After
```tsx
<Alert dismissible onDismiss={this.handleClick} />
```

----

## Add `visible` prop

It provides a better experience when `Alert` is used with `dismissible` prop as because it allows to use `Alert` in autocontrolled mode.

```tsx
const [visible] = React.useState(true)

<Alert
  dismissible
  onDismiss={() => setVisible(false)}
  visible={visible}
/>
```

## `action` slot was renamed to `dismissAction`

The most common thing that all `Alert`-like components have it's a dismiss action that hides the component. It's the main action that I found with these components, so it's usage should be simplified.

#### Bootstrap | `dismissible`

image

### SUIR/SUI | `onDismiss`, `Dismissable`

image

### Fabric UI | `dismissButton`

image

### AntDesign | `closable`

image

### Lightning DS | `close button`

image

### Backpack | `Dismissable`

image

### AtlasKit | `Dismiss`

image

## Add `actions` slot

To have multiple action buttons is also quite common and this is a different that have `dismissable`.

image

### AtlasKit | `actions`

image

### Fabric UI | `actions`

image


levithomason:
What about the cases where the dismiss is not an icon X, but it is in the actions?

image

Seems most implementations use a separate concept for `close` and `actions` as you've proposed, so let's move forward.  However, we will need guidance and perhaps a Usage example for the case where the dismiss button is in the `actions` as above.

sophieH29 (approved):

PR #1824, opened 2019-08-19
fix(button): make circular background color of button white instead of transparent

### LINKED ISSUE #1813: Circular button is transparent in teams theme
`<Button />` and `<Button circular />` have different background colors. The first button has a white background in the Teams light theme, but that background becomes transparent when it's made circular.

### Expected Result

Button should have the same background as the default button (white), just with rounded borders.

### Actual Result

Button loses its background color.

### Version
0.36.1

### Testcase

```jsx
import React from 'react'
import { Provider, Button, Flex, themes } from '@stardust-ui/react'

const ButtonExampleCircular = () => (
  <Provider theme={themes.teams}>
    <Flex gap="gap.smaller" styles={{ background: "rgb(170, 170, 170)" }}>
       <Button circular content="Circular" />
       <Button content="Default" />
     </Flex>
   </Provider>
)
```

lucivpav:
Fixes #1813


miroslavstastny:
Discussed with @bcalvery, #1813 is not a bug.

PR #1825, opened 2019-08-19
fix(Tooltip): Fixing incorrect tooltip shadow

bcalvery:
Changed TooltipContentVariables to use shadow value defined in siteVariables rather than separate values defined within tooltip.

Before:
image
After:
image


CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix `hand` icon in Teams theme @lucivpav ([#1782](https://github.com/stardust-ui/react/pull/1782))
 - ESC key should close the last opened `Popup` or `Dialog` if body has focus @sophieH29 ([#1807](https://github.com/stardust-ui/react/pull/1807))
 - Correctly define current document object of the `FocusZone` and `FocusTrapZone` @sophieH29 ([#1820](https://github.com/stardust-ui/react/pull/1820))
+- Fix `Tooltip` shadow incorrect [redlines] @bcalvery ([#1825](https://github.com/stardust-ui/react/pull/1825))
```
codepretty: @miroslavstastny, does a variable rename need to be in the BREAKING CHANGES section?

codepretty (approved):
Just make sure that it is in the right spot in the changelog. Rest of it looks good to me!

CHANGELOG.md
layershifter: @codepretty yep, as it's a part of public API

PR #1826, opened 2019-08-20
feat(colors): added more color schemes in Teams theme

mnajdova:
Added yellow, green, orange, pink, onyx, silver and amethyst color schemas


docs/src/views/ColorSchemes.tsx
```diff
@@ -31,7 +31,18 @@ export default () => {
...
+              'brand',
+              'red',
+              'green',
+              'yellow',
+              'orange',
+              'pink',
+              'silver',
+              'onyx',
+              'amethyst',
+            ]}
```
layershifter, line 45: Can we get this from `Provider.Consumer`?

packages/react/src/themes/teams-high-contrast/colors.ts
```diff
@@ -231,7 +312,119 @@ export const colorScheme: ColorSchemeMapping = {
...
+  yellow: {
+    foreground: colors.white,
+    foreground1: colors.black,
+    foreground2: colors.black,
+    background: colors.white,
+    background1: undefined,
+    background2: colors.white,
+    border: undefined,
+    shadow: undefined,
+    foregroundHover: undefined,
```
layershifter: Can we introduce an `emptyScheme` and merge with it?

```tsx
const emptyScheme: SchemeType = {
    border: undefined,
    shadow: undefined,
    foregroundHover: undefined,
    // ...
}

//
{
  yellow: {
    ...emptyScheme,
    foreground: colors.white,
    foreground1: colors.black,
    foreground2: colors.black,
  }
}
```

docs/src/components/ColorBox.tsx
```diff
@@ -87,23 +94,25 @@ const ColorBox = createComponent<ColorBoxProps>({
...
-      {!copyToClipboardIcon && showColorValue && (
-        <span className={classes.value}>{value || 'Not defined'}</span>
-      )}
+        {copyToClipboardIcon && (
+          <CopyToClipboard value={value}>
+            {(active, onClick) => (
+              <div className={classes.value}>
+                <span onClick={onClick}>
+                  {value && <Icon name={active ? 'checkmark' : 'copy outline'} size="small" />}
+                  {value || 'Not defined'}
```
layershifter, line 110: Is not readable :(

image

---

What about special background for such cases? Something like this: https://codepen.io/tennowman/pen/ynrih

docs/src/components/ColorBox.tsx
```diff
@@ -53,10 +53,17 @@ export const colorBoxVariables = (siteVariables): ColorBoxVariables => ({
 
 export const colorBoxStyles: ComponentSlotStylesInput<ColorBoxProps, ColorBoxVariables> = {
   root: ({ props: p, variables: v }): ICSSInJSStyle => ({
-    backgroundColor: p.value,
+    ...(p.showColorValue && {
+      backgroundImage:
+        'url("data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs+9AAAAKUlEQVQoU2NkYGAwZkAD////RxdiYBwKCv///4/hGUZGkNNRAeMQUAgAtxof+nLDzyUAAAAASUVORK5CYII=")',
+      backgroundRepeat: 'repeat',
+    }),
```
layershifter: image

### vs.

image

---

On tables second looks better, what do you think?

docs/src/views/ColorSchemes.tsx
mnajdova: Unfortunately, we don't have red-lined color schemes for all colors currently, so until that's true, I would rather enum this manually...

docs/src/components/ColorBox.tsx
mnajdova: Fixed, that border looked awkward.. :) Please take a look now

docs/src/components/ColorBox.tsx
mnajdova: Agreed, will change it with something different

mnajdova: Actually will make it transparent, as, any other pattern I choose is somehow distracting the user to look into the empty blocks :\

layershifter (approved):

PR #1831, opened 2019-08-20
fix(Menu): Menu width and Submenu indicator fix

bcalvery:
**Objective:** Improve width behavior and update indicator icon to chevron type.

**Before:**
image
image

**After:**
image
image


lucivpav:
I think that when using the RTL mode, the submenu arrow should point to the opposite direction. Also, the expanded submenu should go to the opposite side in RTL mode.
image

codepretty:
> Should icons in the menu cause items without icons to shift right in order to align?

In teams theme, I don't think it should shift menu items without icons. I don't think it is a common case but the few examples I found do not shift.

image

image

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -239,11 +238,13 @@ class MenuItem extends AutoControlledComponent<WithAsProp<MenuItemProps>, MenuIt
...
             })}
-          {rtlTextContainer.createFor({ element: content })}
+          {Box.create(content, {
+            defaultProps: { as: 'span', styles: styles.content },
+          })}
           {menu &&
             Icon.create(indicatorWithDefaults, {
               defaultProps: {
-                name: vertical ? 'stardust-arrow-end' : 'stardust-arrow-down',
+                name: vertical ? 'chevron-right' : 'chevron-down',
```
codepretty: I _think_ the `stardust` prefix for icons is used to indicate when a different icons is used per theme. So I think you might want to update or add a reference.

https://github.com/stardust-ui/react/blob/master/packages/react/src/themes/teams/index.tsx

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -255,8 +256,8 @@ class MenuItem extends AutoControlledComponent<WithAsProp<MenuItemProps>, MenuIt
         <>
           <Ref innerRef={this.menuRef}>
             <Popper
-              align={vertical ? 'top' : 'start'}
-              position={vertical ? 'after' : 'below'}
+              align={vertical ? 'top' : rtl ? 'end' : 'start'}
```
dvdzkwsk, line 261: Stardust noobie  question here for @codepretty, @levithomason , etc.: would this be better done within `Popper`? Making Popper itself respect RTL settings would eliminate this work on the consumer's end, which may otherwise be forgotten.

packages/react/src/components/Menu/MenuItem.tsx
bcalvery: I'm not sure how I would I go about changing modifying 'stardust-arrow-down' and 'stardust-arrow-end' just in the context of Menu.
image
Would change it for every case of 'stardust-arrow-down and -right' which isn't what I want - I don't think it is at least.
Doing this would mean new assets are needed for the default Stardust theme right?
image

packages/react/src/components/Menu/MenuItem.tsx
bcalvery: I'm not sure. Popper seems to be a bit more generic - it depends on the element using it to feed it position information, and the positioning of Menu is affected by RTL possibly differently than other items using Popper. @levithomason ?
https://github.com/stardust-ui/react/blob/master/packages/react/src/lib/positioner/Popper.tsx

packages/react/src/themes/base/components/Icon/iconNames.ts
```diff
@@ -11,6 +11,8 @@ export const icons: ThemeIcons = {
   'stardust-arrow-down': fontIcon('25BE'),
   'stardust-arrow-end': fontIcon('25B8'),
   'stardust-arrow-up': fontIcon('25B4'),
+  'stardust-menu-arrow-down': fontIcon('25BE'),
+  'stardust-menu-arrow-end': fontIcon('25B8'),
```
mnajdova, line 15: I think it's nice to scope these icons per component, so people would ideally configure via the theme which icons should appear on which components. 👍

packages/react/src/components/Menu/MenuItem.tsx
mnajdova: But Popper should respect rtl, this works perfectly if you use the Popup component for example: 
image

Let me dive into this tomorrow and see what's happening

bcalvery: @mnajdova Menu is using "Popper" not "Popup" - maybe that is part of the difference?

bcalvery: I am trying to swap in Popup for Popper in MenuItem, and it just is not working:
image

Popper has a `targetRef` property, while Popup seems to have a `target` property which is passed to Popper in Popup), but `target` of Popup doesn't want to accept `{this.itemRef}`.

codepretty: @bcalvery, i think you have a good solution. i think it probably makes sense to keep this rtl logic where you have it unless we know how popper will be used in all cases, it might not make sense to prematurely move this logic to popper.

CHANGELOG.md
```diff
@@ -56,7 +57,6 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixes
 - Update text `Button` styles for Teams Theme to match [redlines] @codepretty ([#1719](https://github.com/stardust-ui/react/pull/1719))
-- Fix `Menu` and `MenuItem` styles to match [redlines] @bcalvery ([#1712](https://github.com/stardust-ui/react/pull/1712))
```
codepretty: Shouldn't this stay here?

packages/react/src/themes/teams/components/Menu/menuItemStyles.ts
```diff
@@ -402,21 +403,47 @@ const menuItemStyles: ComponentSlotStylesInput<MenuItemPropsAndState, MenuVariab
     }
   },
 
+  content: ({ props: p }): ICSSInJSStyle => {
+    const widthAdjust = (p.icon ? 26 : 0) + (p.menu ? 16 : 0)
+    return {
+      whiteSpace: 'normal',
```
codepretty, line 409: I think the default value is already `normal`.. is this `white-space` property overriding something?

bcalvery: The root element has `whiteSpace: 'nowrap'` and that seems to affect all children, but the text within the "content" should wrap if necessary, so this returns it to `'normal'`.

codepretty (approved):

PR #1832, opened 2019-08-21
docs(Alert): add usage example with actions

layershifter:
This PR is a follow up for https://github.com/stardust-ui/react/pull/1823#issuecomment-522638836. Adds an example that show how to use `actions` in `Alert` component to dismiss it.

### DIFF
--- a/docs/src/examples/components/Alert/Usage/index.tsx
+++ b/docs/src/examples/components/Alert/Usage/index.tsx
@@ -1,0 +1,16 @@
+import * as React from 'react'
+
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
+
+const Usage = () => (
+  <ExampleSection title="Usage">
+    <ComponentExample
+      title="Dismiss on actions"
+      description="You can dismiss alerts also on actions."
+      examplePath="components/Alert/Usage/AlertExampleDismissActions"
+    />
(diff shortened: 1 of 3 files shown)


docs/src/examples/components/Alert/Usage/index.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react'
+
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
+
+const Usage = () => (
+  <ExampleSection title="Usage">
+    <ComponentExample
+      title="Dismiss on actions"
+      description="You can dismiss alerts also on actions."
```
lucivpav:
```suggestion
      description="You can also dismiss alerts on actions."
```

docs/src/examples/components/Alert/index.tsx
```diff
@@ -13,6 +14,7 @@ const AlertExamples = () => (
     <State />
     <Slots />
     <Rtl />
+    <Usage />
```
lucivpav: `<Usage />` section should not go to `index.tsx` of any component example. The Usage is displayed under Usage tab.

sophieH29 (approved):

docs/src/examples/components/Alert/index.tsx
layershifter: Oops, good catch 👍

PR #1837, opened 2019-08-21
fix(renderComponent): do not throw if context is missing

### LINKED ISSUE #1834: renderComponent: usage without Provider throws an unclear error
### Steps
1. Open https://codesandbox.io/s/stardust-ui-example-3pxmg. It contains a usage of `Button` without `Provider` component

### Expected Result
See rendered `Button` and an error in console

### Actual Result
App throws 💣

image

Without any meaningful description.

### Version
0.36.2

layershifter:
Fixes #1834.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix `hand` icon in Teams theme @lucivpav ([#1782](https://github.com/stardust-ui/react/pull/1782))
 - ESC key should close the last opened `Popup` or `Dialog` if body has focus @sophieH29 ([#1807](https://github.com/stardust-ui/react/pull/1807))
 - Correctly define current document object of the `FocusZone` and `FocusTrapZone` @sophieH29 ([#1820](https://github.com/stardust-ui/react/pull/1820))
+- Fix a runtime error if context if missing @layershifter ([#1837](https://github.com/stardust-ui/react/pull/1837))
 
 ### Features
 - Add `overwrite` prop to `Provider` @layershifter ([#1780](https://github.com/stardust-ui/react/pull/1780))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix `hand` icon in Teams theme @lucivpav ([#1782](https://github.com/stardust-ui/react/pull/1782))
 - ESC key should close the last opened `Popup` or `Dialog` if body has focus @sophieH29 ([#1807](https://github.com/stardust-ui/react/pull/1807))
 - Correctly define current document object of the `FocusZone` and `FocusTrapZone` @sophieH29 ([#1820](https://github.com/stardust-ui/react/pull/1820))
-- Fix a runtime error if context if missing @layershifter ([#1837](https://github.com/stardust-ui/react/pull/1837))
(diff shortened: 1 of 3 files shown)


packages/react/src/lib/mergeThemes.ts
```diff
@@ -177,36 +189,29 @@ export const mergeStyles = (...sources: ComponentSlotStyle[]) => {
...
+      acc.staticStyles = mergeStaticStyles(...acc.staticStyles, ...(next.staticStyles || []))
 
-    acc.animations = mergeAnimations(acc.animations, next.animations)
+      acc.animations = mergeAnimations(acc.animations, next.animations)
 
-    return acc
-  }, emptyTheme)
+      return acc
+    },
+    { ...emptyTheme },
```
layershifter, line 214: Clone is required as it will modify the source object. I have two options:
- leave a comment
- use `Object.freeze` on `emptyTheme`

packages/react/src/lib/renderComponent.tsx
```diff
@@ -203,7 +195,7 @@ const renderComponent = <P extends {}>(
     displayName,
     props: stateAndProps,
     variables: resolvedVariables,
-    theme: context.theme,
```
layershifter, line 206: `context` can be `undefined` if there are no wrapping `Provider` and it will throw an error

packages/react/src/lib/renderComponent.tsx
```diff
@@ -142,7 +142,7 @@ const resolveStyles = (
 
 const renderComponent = <P extends {}>(
   config: RenderConfig<P>,
-  context: ProviderContextPrepared,
+  context?: ProviderContextPrepared,
```
layershifter, line 145: I marked `context` as optional to emphasize that it can be `undefined`

CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix `hand` icon in Teams theme @lucivpav ([#1782](https://github.com/stardust-ui/react/pull/1782))
 - ESC key should close the last opened `Popup` or `Dialog` if body has focus @sophieH29 ([#1807](https://github.com/stardust-ui/react/pull/1807))
 - Correctly define current document object of the `FocusZone` and `FocusTrapZone` @sophieH29 ([#1820](https://github.com/stardust-ui/react/pull/1820))
+- Fix a runtime error if context if missing @layershifter ([#1837](https://github.com/stardust-ui/react/pull/1837))
```
lucivpav:
```suggestion
- Fix a runtime error if context is missing @layershifter ([#1837](https://github.com/stardust-ui/react/pull/1837))
```

packages/react/src/lib/renderComponent.tsx
mnajdova: I understand that this would fix the problem of not throwing when Provider is not defined, but we are still not telling the user, why the Button is not rendered as expected, if there is no Provider :\ It seems like we are silencing the error, instead of explaining it...

layershifter: Actually, we are telling: 
image

And this error will still exist.

---

But then happens:
image

And I want to fix it, because it causes the app crash in the wrong place.

packages/react/src/lib/mergeThemes.ts
mnajdova: leave a comment :)

mnajdova (approved):

packages/react/src/lib/renderComponent.tsx
mnajdova: Got it, I missed that, thanks for the clarification

PR #1842, opened 2019-08-23
fix(Provider): should not use felaRenderer explicitly

### LINKED ISSUE #1814: Provider doesn't always use props.renderer
`Provider` takes in an optional `renderer` prop to customize the Fela renderer used in the React tree. This prop is correctly passed down via context, but there are still places where Stardust's singleton are used. This means that two renderers can be active at once.

If the renderer is customized, Stardust should always prefer it over its own in order to better support server-side rendering.

Here are a few examples:

* https://github.com/stardust-ui/react/blob/6607278b148f9a9404c4ea48c93be9a75e3d6856/packages/react/src/components/Provider/Provider.tsx#L99
* https://github.com/stardust-ui/react/blob/6607278b148f9a9404c4ea48c93be9a75e3d6856/packages/react/src/components/Provider/Provider.tsx#L105
* https://github.com/stardust-ui/react/blob/6607278b148f9a9404c4ea48c93be9a75e3d6856/packages/react/src/components/Provider/Provider.tsx#L129

### Version
0.36.1

### Testcase
N/A

lucivpav:
Fixes #1814


layershifter (commented):
Let's add some tests for it

packages/internal-tooling/eslint/index.js
```diff
@@ -115,6 +115,12 @@ module.exports = {
         '@stardust-ui/no-visibility-modifiers': 'error',
       },
     },
+    {
+      files: '**/*.{ts,tsx}',
+      rules: {
+        'no-dupe-class-members': 'off',
```
lucivpav, line 121: Otherwise we get false positive in `RendererMock.tsx`, see https://github.com/typescript-eslint/typescript-eslint/issues/291

lucivpav:
> Let's add some tests for it

@layershifter I have added tests. But I feel its a bit ugly to add `RendererMock.tsx` file. Can you point me to the right direction?

packages/react/test/specs/components/Provider/Provider-test.tsx
```diff
@@ -184,4 +187,46 @@ describe('Provider', () => {
...
   })
+
+  describe('calls provided renderer', () => {
+    test('calls renderFont', () => {
+      const theme: ThemeInput = {
+        fontFaces: teams.fontFaces,
+      }
+
+      const renderFontCalled = jest.fn()
+      const rendererMock = new RendererMock(renderFontCalled, jest.fn())
```
layershifter: https://jestjs.io/docs/en/jest-object.html#jestspyonobject-methodname

Have you tried `spyOn` instead?

packages/react/test/specs/components/Provider/Provider-test.tsx
```diff
@@ -184,4 +187,44 @@ describe('Provider', () => {
       })
     })
   })
+
+  describe('calls provided renderer', () => {
+    test('calls renderFont', () => {
+      const theme: ThemeInput = {
+        fontFaces: teams.fontFaces,
+      }
+      const renderer = createRenderer()
```
lucivpav, line 201: I cannot use `felaRenderer`, because the test would pass even without this fix.

packages/internal-tooling/eslint/index.js
lucivpav: No longer needed at the moment, but I think it will be useful to keep this change, as described in the link above

layershifter: Yep, let's keep it

packages/react/test/specs/components/Provider/Provider-test.tsx
```diff
@@ -184,4 +187,44 @@ describe('Provider', () => {
       })
     })
   })
+
+  describe('calls provided renderer', () => {
+    test('calls renderFont', () => {
+      const theme: ThemeInput = {
+        fontFaces: teams.fontFaces,
```
layershifter: Can we use some dumb definition instead?

layershifter (approved):

Issue #1808, opened 2019-08-16
Checkbox: add documentation how to compose with Form


layershifter:
## Feature Request

We need to add documentation how to compose `Checkbox` with `Form` component and usage section.

PR #1845, opened 2019-08-23
docs(Form): add usage example with Checkbox

lucivpav:
Resolves PR: #1808
There are similar examples under _Types_ section. Should I move them under _Usage_ instead?

New example:
image


docs/src/examples/components/Form/Usage/FormExampleCheckbox.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+const theme = {
+  componentStyles: {
+    Checkbox: {
+      root: {
+        padding: 0,
+      },
+    },
+  },
+}
```
layershifter: Let's remove this one, we definitely should not show this in examples

docs/src/examples/components/Form/Usage/FormExampleCheckbox.tsx
```diff
@@ -0,0 +1,52 @@
...
+    name: 'email',
+    id: 'email-inline-shorthand',
+    key: 'email',
+    required: true,
+    inline: true,
+  },
+  {
+    control: {
+      as: Checkbox,
+      labelPosition: 'start',
```
layershifter:
```suggestion
```

It's not required, so let's omit it

layershifter:
> There are similar examples under Types section. Should I move them under Usage instead?

Yes, please 🐱

docs/src/examples/components/Form/Usage/FormExampleCheckbox.tsx
lucivpav: Without this, however, a padding breaks the visual structure of the form.
image

docs/src/examples/components/Form/Usage/FormExampleCheckbox.tsx
lucivpav: I think it looks better, when the form's structure is unified, i.e. description on the left, user's input on the right.

docs/src/examples/components/Form/Usage/FormExampleCheckbox.tsx
layershifter: Let design team to care about it, we shouldn't tune theme specific things in theme agnostic examples

docs/src/examples/components/Form/Usage/FormExampleCheckbox.tsx
layershifter: We can go whatever you will decide, but should mention:
- that default position is on left across Web:
  - https://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_type_checkbox
  - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox
- we should keep examples minimal and avoid questions like "`labelPosition` is required to use this component"?

docs/src/examples/components/Form/Usage/FormExampleCheckbox.tsx
lucivpav: Did you mean `shouldn't`? 😄

lucivpav: fixed

layershifter (approved):

PR #1846, opened 2019-08-26
fix(emptyTheme): avoid theme object mutation

layershifter:
Fixes a small regression from #1837 that was mentioned by @levithomason.

## ???


lucivpav:
How about testing this fix?

jdhuntington:

PR #1847, opened 2019-08-26
fix(Video): fix `muted` prop

layershifter:
This PR fixes the `muted` prop on `Video` component, now it works properly.

Tested in Chrome, FF, Edge and IE11.

### DIFF
--- a/packages/react/src/components/Video/Video.tsx
+++ b/packages/react/src/components/Video/Video.tsx
@@ -67,11 +67,7 @@ class Video extends UIComponent<WithAsProp<VideoProps>> {
     // React doesn't guaranty that props will be set:
     // https://github.com/facebook/react/issues/10389
     if (this.videoRef.current) {
-      if (this.props.muted) {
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Video/Video.tsx
```diff
@@ -67,11 +67,7 @@ class Video extends UIComponent<WithAsProp<VideoProps>> {
     // React doesn't guaranty that props will be set:
     // https://github.com/facebook/react/issues/10389
     if (this.videoRef.current) {
-      if (this.props.muted) {
-        this.videoRef.current.setAttribute('muted', 'true')
-      } else {
-        this.videoRef.current.removeAttribute('muted')
-      }
+      this.videoRef.current.muted = !!this.props.muted
```
layershifter, line 70: https://github.com/facebook/react/issues/10389#issuecomment-512565626

We should set `muted` directly, `setAttribute` only shows this attribute on DOM.

lucivpav (approved):

PR #1850, opened 2019-08-26
chore: remove isFromKeyboard prop from components

layershifter:
# BREAKING CHANGES

1. `Alert`, `Attachment`, `Button`, `ChatMessage`, `Checkbox`, `Embed`, `Grid`, `ListItem`, `RadioGroupItem`, `Reaction`, `Slider`, `ToolbarCustomItem`, `ToolbarItem`, `ToolbarMenuItem` do not have `isFromKeyboard` in their style functions anymore. Use `:focus-visible` selector instead.
2. `AttachmentState`, `ButtonState`, `GridState`, `ListItemState`, `ReactionState`, `ToolbarCustomItemState`, `ToolbarItemState`, `ToolbarMenuItemState` interfaces are no longer exported.

### Before
```tsx
const buttonStyles = {
  root: ({ props: p }) => ({
    ...p.isFromKeyboard && ({ border: '1px solid red' })
  })
}
```
### After
```tsx
const buttonStyles = {
  root: ({ props: p }) => ({
    ':focus-visible': { border: '1px solid red' },
  })
}
```

## 💣 Issue

To understand that event comes from a keyboard currently we have `isFromKeyboard` understate component's state. This solution creates at least two issues:
- it's not scalable: to apply styles we need to update a component
- it's not performant: to apply styles we have `onFocus` handlers that perform state changes and useless component render

##  💡 Idea

The original proposal (#749) was to use polyfill for `:focus-visible`, but we meet at least two issues:
- the polyfill has some bugs
- it's not supported really at any browser

To solve this I propose to use `:focus-visible` in our styles that will simplify the migration later and create a custom [Fela enhancer](http://fela.js.org/docs/advanced/Enhancers.html).

#### Why not plugin?

Fela plugins allow to modify styles, but to make this working I should change `selector` generation.

## 🏎  What it does?

This enhancer performs a transform of generated selectors to replace `:focus-visible` with `html[data-whatinput="keyboard"] + :focus`.

#### `buttonStyles.ts`

```tsx
{
  ':focus': { color: 'red' },
  ':focus-visible': { color: 'green' },
}
```

#### Generated styles

```css
a:focus { color: red }
html[data-whatinput="keyboard"] a:focus { color: green }
```

The original prototype is there: https://codesandbox.io/s/young-dust-fy3b0

### Untouched components

`Menu` & `Dropdown` components still will have `isFromKeyboard` under theirs states because they style container elements and this should be solved by `:focus-within` instead or other tricks separately.


levithomason:
👍

sophieH29 (approved):

PR #1852, opened 2019-08-26
feat(Dialog): add `backdrop` and `closeOnOutsideClick` props

layershifter:
This PR adds multiple props to `Dialog` component.

### `backdrop`
A dialog can have a backdrop on its overlay.

#### `backdrop=true` (default)

image

#### `backdrop=false`

image

### `closeOnOutsideClick`

Controls whether or not a dialog should close when a click outside is happened.


packages/react/src/themes/teams/components/Dialog/dialogVariables.ts
```diff
@@ -11,23 +11,17 @@ export interface DialogVariables extends BaseDialogVariables {
...
+export default (siteVariables): Partial<DialogVariables> => ({
   boxShadow: siteVariables.shadowLevel4,
   foregroundColor: siteVariables.colors.grey[900],
 
-  rootBackground: siteVariables.colors.white,
-  rootBorderRadius: pxToRem(3),
   rootPadding: `${pxToRem(27)} ${pxToRem(32)} ${pxToRem(20)} ${pxToRem(32)}`,
-  rootWidth: '50vw',
-
-  contentMargin: `0 0 ${pxToRem(20)} 0`,
```
layershifter, line 23: Is duplicated in base theme

docs/src/examples/components/Dialog/Variations/DialogExampleInert.shorthand.tsx
```diff
@@ -0,0 +1,14 @@
+import { Button, Dialog } from '@stardust-ui/react'
+import * as React from 'react'
+
+const DialogExampleInert = () => (
+  <Dialog
+    cancelButton="Cancel"
+    confirmButton="Confirm"
+    header="Action confirmation"
+    inert
+    trigger={<Button content="Open a dialog" />}
```
mnajdova: Why this is not a knob?

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -43,9 +43,15 @@ export interface DialogProps
...
   actions?: ShorthandValue<BoxProps>
 
+  /** A dialog can have a backdrop on its overlay. */
+  backdrop?: boolean
+
   /** A dialog can contain a cancel button. */
   cancelButton?: ShorthandValue<ButtonProps>
 
+  /** Controls whether or not a dialog should close when the document is clicked. */
+  closeOnDocumentClick?: boolean
```
mnajdova: Why we don't have example for this?

mnajdova: Also why not `closeOnOutsideClick`? Document click is kind a true even when you click inside the dialog :)

packages/react/src/themes/base/components/Dialog/dialogVariables.ts
```diff
@@ -19,6 +20,7 @@ export default (siteVariables): DialogVariables => ({
   rootBorderRadius: pxToRem(3),
   rootPadding: pxToRem(24),
   rootWidth: '50vw',
+  zIndex: 1001,
```
mnajdova: Do we have some convention of what should the values of the `zIndex` be?

docs/src/examples/components/Dialog/Variations/DialogExampleBackdrop.shorthand.tsx
```diff
@@ -0,0 +1,19 @@
+import { useBooleanKnob } from '@stardust-ui/docs-components'
+import { Button, Dialog } from '@stardust-ui/react'
+import * as React from 'react'
+
+const DialogExampleBackdrop = () => {
+  const [backdrop] = useBooleanKnob({ initialValue: false, name: 'backdrop' })
```
mnajdova: Why this is not true by default? The example intend is to show the backdrop.

docs/src/examples/components/Dialog/Variations/DialogExampleInert.shorthand.tsx
```diff
@@ -0,0 +1,14 @@
+import { Button, Dialog } from '@stardust-ui/react'
+import * as React from 'react'
+
+const DialogExampleInert = () => (
+  <Dialog
```
mnajdova: opening this dialog is scrolling the page to the top. Is that expected?

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -58,6 +64,12 @@ export interface DialogProps
   /** A dialog can contain a button next to the header. */
   headerAction?: ShorthandValue<ButtonProps>
 
+  /**
+   * Whether the dialog should be inert, e.g. not dismiss when focusing/clicking outside of the dialog. Hides an
+   * overlay.
+   */
+  inert?: boolean
```
mnajdova: Is it expected not to be able to actually write text in an input outside of the dialog? Click and focusing works as expected, but writing something inside form controls outside of the Dialog doesn't work...

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -58,6 +64,12 @@ export interface DialogProps
   /** A dialog can contain a button next to the header. */
   headerAction?: ShorthandValue<ButtonProps>
 
+  /**
+   * Whether the dialog should be inert, e.g. not dismiss when focusing/clicking outside of the dialog. Hides an
+   * overlay.
```
mnajdova: Maybe backdrop would be better word here, as it is pointing to the actual prop we have.

docs/src/examples/components/Dialog/Variations/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+const DialogVariationsExamples = () => (
+  <ExampleSection title="Variations">
+    <ComponentExample
+      title="Backdrop"
+      description="A dialog can have a backdrop on its overlay."
+      examplePath="components/Dialog/Variations/DialogExampleBackdrop"
+    />
+    <ComponentExample
+      title="Inert"
+      description="Whether the dialog should be inert, e.g. not dismiss when focusing/clicking outside of the dialog. Hides an overlay."
```
levithomason: `inert` should prevent the user from being able to interact with the content below the dialog.

https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

image


image

Perhaps this prop is then better called `modal`?

packages/react/src/components/Dialog/Dialog.tsx
levithomason: +1 to `closeOnOutsideClick`, it also common in the community.  More importantly, the click may not be registered on the document at all but just registered that it happened "outside" the dialog.

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -127,6 +142,9 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
   static defaultProps = {
     accessibility: dialogBehavior,
     actions: {},
+    backdrop: true,
```
mnajdova, line 138: In my opinion it feels awkward that the defaults are true for some of these props.. I had the same doubts for the List and the vertical prop => it is true by default, so we agreed that better prop is probably horizontal, which is false by default... Just sharing opinion, not sure if we should rename these.

packages/react/src/components/Dialog/Dialog.tsx
layershifter: Done 👍

docs/src/examples/components/Dialog/Variations/DialogExampleInert.shorthand.tsx
layershifter: IDK, added 😄

packages/react/src/themes/base/components/Dialog/dialogVariables.ts
layershifter: Nope, that's why this value is `+1`

docs/src/examples/components/Dialog/Variations/DialogExampleBackdrop.shorthand.tsx
layershifter: Updated to `true`

layershifter:
@mnajdova @levithomason I reworked this PR. I removed `inert` & `backdrop` in favor of `modal`, it looks better for me now. Can you please take one more?

packages/react/src/lib/accessibility/Behaviors/Dialog/dialogBehavior.ts
```diff
@@ -61,6 +61,7 @@ const getDefaultAriaDescribedBy = (props: DialogBehaviorProps) => {
 export default dialogBehavior
 
 type DialogBehaviorProps = {
+  closeOnDocumentClick?: boolean
```
mnajdova: Oops, not renamed

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -58,6 +61,12 @@ export interface DialogProps
   /** A dialog can contain a button next to the header. */
   headerAction?: ShorthandValue<ButtonProps>
 
+  /**
+   * Whether content outside should be inert, e.g. when focusing/clicking outside is not possible and the dialog will\
```
mnajdova:
```suggestion
   * Whether content outside should be inert, e.g. when focusing/clicking outside is not possible and the dialog will
```

mnajdova:
I know it's documented in the PR already, but I feel awkward that with this config, close on outside click doesn't work :\ 
```
  <Dialog
    cancelButton="Cancel"
    modal={false}
    closeOnOutsideClick={true}
    confirmButton="Confirm"
    header="Action confirmation"
    trigger={<Button content="Open a dialog" />}
  />
```

docs/src/examples/components/Dialog/Variations/DialogExampleInert.shorthand.tsx
layershifter: Nope, was fixed 👍

packages/react/src/components/Dialog/Dialog.tsx
layershifter: Description was changed

packages/react/src/components/Dialog/Dialog.tsx
layershifter: Good catch 👍

layershifter:
> I know it's documented in the PR already, but I feel awkward that with this config, close on outside click doesn't work :\

Same thing with `trapFocus`, actually.

packages/react/src/themes/base/components/Dialog/dialogVariables.ts
levithomason: https://psuter.net/2019/07/07/z-index 😂

levithomason: It isn't enough:

image

mnajdova:
Per offline discussion, `closeOnOutsideClick` and `trapFocus` should always work, because the `modal` prop is more generic. If we have modal, the defaults are `closeOnOutsideClick` true, `trapFocus` true and `overlay` true, if ti is false the opposite, but if some of these props are defined, they should win. Good to merge after these changes :) Nice work 👍

mnajdova (approved):

levithomason:
I don't think `modal={false}` makes sense for our Dialog since it is strictly a modal Dialog component only.  Its slots and usages dictate that it only works for the modal Dialog use cases.  It is always intended for "interrupting the users workflow".  It cannot reasonably serve any other "dialog" use cases such as a Popup, Date picker, etc.

A general purpose "dialog" for all use cases would just be a "window" component, basically a portal with no slots and little to no styling.  I think we should have specific components for the dialog use cases, hardcode the correct modal behaviors, and provide props for the optional parts.

Per ARIA:

**[Dialog](https://www.w3.org/TR/wai-aria-1.1/#dialog)** _(Doesn't describe our `Dialog`)_
- A descendant window of the primary window of a web application (Alert, Message, Popup, etc)
- Focus may move outside

**[Dialog (Modal)](https://www.w3.org/TR/wai-aria-practices/#dialog_modal)** _(Describes our `Dialog`)_
- Content outside is inert (cannot interact with it)
- Optionally has a backdrop/overlay
- Optionally closes on outside click
- Usually for interrupting the users workflow
- Focus cannot move outside

**✓** always, **×** never, **−** sometimes

| Feature                        | Dialog | +Modal Behavior |
|--------------------------------|--------|--------|
| Focuses an element inside      | ✓      | ✓      |
| Circular tab nav within        | ✓      | ✓      |
| Focus can leave dialog         | −      | ×      |
| Inert content outside          | −      | ✓      |
| Visible backdrop               | −      | −      |
| Close on outside click         | −      | −      |
| Close on ESC                   | −      | ✓      |

The modal behavior is a specific implementation of a Dialog where most features are hard coded.  The only features that remain optional are whether the backdrop is visible and whether it closes on outside click.

Rendering a Stardust Dialog without modal capabilities doesn't make sense and is an invalid interaction model for the component.  So, we should consider a props API that follows the **+Modal Behavior** column above.

### Modal vs Dialog

Have a look through all Dialog and Modal links in the [Open UI component matrix](https://open-ui.org/analysis/component-matrix).  There is consistency in all implementations that use the name "Modal".  They indicate a dialog with this particular anatomy and modal behavior.

The things called "Dialog" include modal dialogs, toast messages, popups, and invalid modal dialog cases.  The term dialog is not consistent nor clear in the wild.

Propose we also rename this component to `Modal`.  Although it is intended to describe a behavior in the aria specs, in the real world it is used to describe a component with the same anatomy, appearance, and behavior as what we've implemented.

## Proposed API

Note that in all cases of this component the content is to remain inert.  This is done by the backdrop intercepting clicks, whether opaque (on) or transparent (off).

**`<Modal />`** 
Inert outside, show backdrop, close on outside click.  Default behavior acts just as you'd expect a modal dialog to behave.

**`<Modal backdrop={false} />`**
Inert outside, transparent backdrop, close on outside click.  Propose the backdrop stay in the DOM so that it can intercept clicks but just made transparent.  This is because even when no backdrop is shown the content outside is still supposed to be inert.

**`<Modal closeOnOutsideClick={false} />`**
Inert outside, show backdrop, do not close on outside click.  Here we just ignore clicks on the backdrop, but it is still visible and in the DOM.

**`<Modal backdrop={false} closeOnOutsideClick={false} />`**
Inert outside, transparent backdrop, no not close on outside click.

## Conclusion

I think calling our component a "Modal" would make it much more clear to the community.  It also makes it more clear why it has the props it has and why it is lacking other features.

This also leaves room for us to implement other specific components for the other "dialog" components, like a date picker, popup, alert, tool window, etc.

We may also consider / prototype a lower level Dialog component which would allow configuration for all these use cases.  Sounds much like the [`Portal`](http://react.semantic-ui.com/addons/portal) component in SUIR which powers the Modal and Popup there.

layershifter:
I discussed this with @jurokapsiar and we agreed to implement `backdrop` and `closeOnOutsideClick` props.

I moved out the discussion about component to #1889 to unblock current PR.

PR #1854, opened 2019-08-26
feat(design): add prop and component for adjusting layouts

levithomason:
# Short Story

Layout Stardust components using the `design` prop.

```jsx
<Button design={{ marginRight: '1rem' }} primary>Primary</Button>
<Button design={{ marginRight: '1rem' }} secondary>Secondary</Button>
```

Layout non Stardust components using the `<Design />` component.

```jsx
<Design config={{ padding: '10px', marginLeft: '0.5rem' }}>
  {({ className }) => (
    <p className={className}>
      All the things together
    </p>
  )}
</Design>
```

***

# Details 

This PR is experimenting with providing a way to adjust/override the layout of components while not breaking theming.

The current approach is to provide a new prop and component that allows passing only layout styles through (display, position, padding, margin, etc).

We have plans to further develop this prop to also restrict the values passed to only theme acceptable values.  Example, currently in this PR you can pass `design={{ padding: '32px' }}` but in the future we may restrict it to something like `design={{ padding: 4 }}` where `4` would be the 4th stop on a predefined theme scale ramp.

## Target UI

This is the card UI we tried to layout.

image

## Initial Result

Here's the result using Stardust's Segment, Header, Image, and Text components.  Note the size and layout of the components isn't even close to the target UI.

image

## Design adjustments
Here we adjust layout properties on the components only.

image

***

# Conclusions

**👍 Pros**
- Restricting to layout properties means theme switching still works (light/dark/contrast)
- Users can more easily create UIs they need by having direct control over layouting
- We can iterate toward even more safe values for this property (theme enum)

**👎 Cons**
Some "design" oriented things could not be achieved:
- Segment radius and shadow is wrong
- Border between top and bottom Segments
- Font size and line heights are off

The cons listed above may not be cons if the goal is purely layout, but it just strikes me odd when "designing" a component to be able to change it's position, display, and size to match a redline but not be able to adjust the line height or add/remove a border/box shadow.


packages/react/src/lib/commonPropTypes.ts
```diff
@@ -33,6 +33,50 @@ export const createCommon = (config: CreateCommonConfig = {}) => {
     ...(as && {
       as: PropTypes.elementType,
     }),
+    design: PropTypes.oneOfType([
```
miroslavstastny: Add this conditionally - either new boolean or, preferably, move this under `styled`

miroslavstastny (approved):

PR #1856, opened 2019-08-27
docs(Dropdown): update docs regarding `renderSelectedItem`

### LINKED ISSUE #1584: Dropdown's renderSelectedItem prop doesn't get called
### Steps
We are using Dropdown component and we pass it renderSelectedItem as prop to provide custom rendering but that method never gets called (unless we allow multiple selection)

From https://github.com/stardust-ui/react/blob/master/packages/react/src/components/Dropdown/Dropdown.tsx:

image

### Expected Result
When item is selected from the Dropdown component and proper function is passed as renderSelectedItem prop, that function gets called

### Actual Result
When item is selected from the Dropdown component and proper function is passed as renderSelectedItem prop, that function doesn't get called 💣

### Version
0.34.0

### Testcase

lucivpav:
Fixes #1584


lucivpav:
@layershifter is this actually a bug? If the behavior on master is expected, then we should update the prop documentation (which doesn't make it clear the prop only works when `multiple` is used)
image

layershifter:
> @layershifter is this actually a bug?

Not, it's not a bug. `renderSelectItem` should be applied to `DropdownSelectedItem`, so proper fix is to update prop description.

image

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -185,7 +185,7 @@ export interface DropdownProps
   renderItem?: ShorthandRenderFunction
 
   /**
-   * A custom render function for the selected item.
+   * A custom render function for the selected item. Only applicable with the multiple prop.
```
```suggestion
   * A custom render function for the selected item. Only applicable with the `multiple` prop.
```

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -1441,4 +1441,25 @@ describe('Dropdown', () => {
...
+        <SelectedItem {...props} header={`${props.header} (selected)`} />
+      ))
+      const wrapper = mountWithProvider(
+        <Dropdown multiple items={items} renderSelectedItem={renderSelectedItem} />,
+      )
+      const triggerButton = getTriggerButtonWrapper(wrapper)
+
+      triggerButton.simulate('click')
+      const firstItem = getItemAtIndexWrapper(wrapper)
+      firstItem.simulate('click', { nativeEvent: { stopImmediatePropagation: jest.fn() } })
```
layershifter:
```suggestion
      firstItem.simulate('click', { nativeEvent: { stopImmediatePropagation: _.noop } })
```

_nit_, you can use `_.noop` from Lodash.

packages/react/test/specs/components/Dropdown/Dropdown-test.tsx
```diff
@@ -1441,4 +1441,25 @@ describe('Dropdown', () => {
       expect(stopPropagation).toBeCalledTimes(1)
     })
   })
+
+  describe('renderSelectedItem', () => {
+    it('calls renderSelectedItem in multiple selection', () => {
+      const renderSelectedItem = jest.fn((SelectedItem, props) => (
+        <SelectedItem {...props} header={`${props.header} (selected)`} />
+      ))
```
layershifter: There are two options:
- simplify `jest.fn`:  `jest.fn(() => null)` as we don't test rendered content
- expand this UT to catch that these elements were rendered:
```tsx
      const className = faker.random.word()
      const renderSelectedItem = jest.fn((SelectedItem, props) => (
        <SelectedItem {...props} className={className} header={`${props.header} (selected)`} />
      ))
//

expect(wrapper.find(`.${className}`)).toHaveLength(5)
```

layershifter (approved):

PR #1857, opened 2019-08-27
fix(Checkbox): Checkbox should be vertically aligned with content

bcalvery:
**Before:**
image

**After:**
image


mnajdova:
Let's add one example to potentially catch regressions for this in future.

codepretty (approved):

packages/react/src/themes/teams/components/Checkbox/checkboxStyles.ts
```diff
@@ -53,6 +55,10 @@ const checkboxStyles: ComponentSlotStylesInput<CheckboxProps & CheckboxState, Ch
     margin: v.margin,
     padding: v.padding,
 
+    [`&.${Checkbox.slotClassNames.indicator}`]: {
+      marginTop: pxToRem(2),
```
layershifter: image

image

This override is really strange. I expect that `margin` variable will be overridden instead:

#### `checkboxStyles.ts`
```diff
-    [`&.${Checkbox.slotClassNames.indicator}`]: {
-      marginTop: pxToRem(2),
-    },
```
#### `checkboxVariables.ts`
```diff
-  margin: `${pxToRem(20)}, 0`, // THERE IS A MISTAKE, COMMA!
+  margin: `pxToRem(2) 0 0 0`,
```

layershifter (changes requested):
Should be the same thing applied to toggle?

packages/react/src/themes/teams/components/Checkbox/checkboxStyles.ts
bcalvery: Thank you for catching that. I wonder why that mistake was not causing any errors.

PR #1855, opened 2019-08-27
chore: add screener tests for keyboard focus

layershifter:
Extracted from #1850.

This PR adds more screener tests to catch possible bugs with focus styles.

PR #1859, opened 2019-08-28
chore: fix broken screener tests

layershifter:
Extracted from #1855.

---

Fixes an issus with wrongly persisted mouse state between steps and themes. We found it when verified new portion of screenshots in #1855. The easiest repro can be found with high contrast theme examples. For example I compare below `PopupExample.shorthand.tsx: Theme: teamsHighContrast`:

### Wrong

image

_It's wrong because the button is hovered on default screenshot 💥_ 

### Correct

image

***

I also tried to use a trick with _hovering `body`_, but it produces wrong screenshots: elements are not hovered on click anymore. So there is only one correct solution, to reload page between tests.

### DIFF
--- a/build/screener/screener.states.ts
+++ b/build/screener/screener.states.ts
@@ -31,12 +31,14 @@ const getStateForPath = (examplePath: string) => {
   const rtl = exampleNameWithExtension.endsWith('.rtl.tsx')
   const exampleUrl = _.kebabCase(exampleNameWithoutExtension)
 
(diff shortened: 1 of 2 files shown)


lucivpav:
I think this one should be renamed. Its confusing to see both of the elements appear hovered. https://screener.io/v2/states/rkGqtccHX.stardust-ui-react/chore%2Fmake-screenshots-correct/1024x768/Chrome/dropdownexamplesearchmultipleshorthandtsx-hovers-first-selected-element-2

This is a problem on master though, not with your PR.

lucivpav (approved):

PR #1860, opened 2019-08-28
feat(CopyToClipboard): add prototype

lucivpav:
### Props
* value - the value to be copied
* copyPrompt - text describing what is going to be copied
* timeout - how long to show _copied_ prompt, in milliseconds
* attached - displays _copied_ prompt as a popup vs renders the message in the middle of parent element
* pointing - whether the popup is pointing

### Examples
image

image

image

image

### Issues
When clicking the not attached version, use has no way to interact with elements other than the message.


docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+export type CopyToClipboardProps = {
+  value: string
+  copyPrompt: string
+  timeout?: number
+  attached?: boolean
+  pointing?: boolean
+}
+
+class CopyToClipboard extends React.Component<CopyToClipboardProps> {
```
layershifter: Let's be trendy and use hooks 🐤

docs/src/prototypes/CopyToClipboard/index.tsx
```diff
@@ -0,0 +1,55 @@
+import * as React from 'react'
+import { Flex, Text, Provider, themes } from '@stardust-ui/react/src'
+import CopyToClipboard from './CopyToClipboard'
+import { PrototypeSection, ComponentPrototype } from '../Prototypes'
+
+class CopyToClipboardPrototypeAttached extends React.Component {
+  render() {
+    const commitID = '3422f7d'
+    return (
+      <Provider theme={themes.teams} style={{ padding: 10 }}>
```
layershifter: I've seen this multiple times, can you please clarify why are you using `Provider`?
You don't need it there as doc site uses `teams` theme by default.

To specify `padding`:
- you can use cool old `div`s: `<div style={{ padding: 10 }} />`
- you can specify it on `Flex`: `<Flex styles={{ padding: '10px' }} />`
- you can use `padding` prop on `Flex`: `<Flex padding="padding.medium" />`, this is the best probably

docs/src/prototypes/CopyToClipboard/index.tsx
```diff
@@ -0,0 +1,55 @@
+import * as React from 'react'
+import { Flex, Text, Provider, themes } from '@stardust-ui/react/src'
+import CopyToClipboard from './CopyToClipboard'
+import { PrototypeSection, ComponentPrototype } from '../Prototypes'
+
+class CopyToClipboardPrototypeAttached extends React.Component {
```
layershifter: Let's use functional components as there is no state.

layershifter: You can also have single component as markup is the same:

```tsx
const CopyToClipboardPrototype: React.FC = (props) => (
  <Flex gap="gap.medium" vAlign="center">
    <Text content="Commit: " />
    <Text content={props.commitId} color="brand" />
    
    <CopyToClipboard attached={props.attached} pointing value={props.commitId} copyPrompt="Copy commit ID" />
  </Flex>
)
```

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,135 @@
...
+      position: 'fixed' as 'fixed',
+      backgroundColor: 'black',
+      color: 'white',
+      zIndex: 1000,
+      top: `calc(50% - ${this.state.copiedHeight / 2}px)`,
+      left: `calc(50% - ${this.state.copiedWidth / 2}px)`,
+    }
+    return (
+      <>
+        <Popup
```
layershifter: Let's use `Tooltip` component instead. API is almost same

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,135 @@
...
+              ? siteVariables.colorScheme.brand.foreground4
+              : siteVariables.colors.black,
+            contentBackgroundColor: this.state.copied
+              ? siteVariables.colorScheme.brand.background4
+              : siteVariables.colors.white,
+            ...(hideContent && hiddenPopupVariables),
+          })}
+        />
+        {hideContent && (
+          <Portal
```
layershifter: Let's extract this to a separate component like `CopyButtonNotification`. We can reuse our [`createComponent()`](https://stardust-ui.github.io/react/integrate-custom-components) to host styles.

```tsx
const CopyButtonNotification = createComponent({
  displayName: 'CopyButtonNotification',
  render: () => {
    return (
      <Portal>
        <div className={classes.root}>
          <div className={classes.content}>{children}</div>
        </div>
       </Portal>
    )
  },
})
```

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,135 @@
...
+      display: 'flex',
+      alignItems: 'center',
+      justifyContent: 'center',
+      position: 'fixed' as 'fixed',
+      backgroundColor: 'black',
+      color: 'white',
+      zIndex: 1000,
+      top: `calc(50% - ${this.state.copiedHeight / 2}px)`,
+      left: `calc(50% - ${this.state.copiedWidth / 2}px)`,
+    }
```
layershifter: I believe that we can start it in more simple way and we will not need a ref 😎 

I prepared CodeSandbox: https://codesandbox.io/s/busy-pond-9v49o

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,135 @@
...
+      copiedHeight: this.copiedTextRef.current.clientHeight,
+      copiedWidth: this.copiedTextRef.current.clientWidth,
+    })
+  }
+
+  render() {
+    const { attached, pointing, copyPrompt } = this.props
+    const hideContent = this.state.copied && !attached
+    const copiedStr = 'Copied to clipboard'
+    const copiedText = hideContent ? (
```
layershifter: I am not fully understand this logic, can you please clarify?

I expect something like this:

```tsx
{attached && copied && <CopyButtonNotification>{notice}</CopyButtonNotification>}
{!attached && <Tooltip>{copied ? notice : usualMessage}</CopyButtonNotification>}
```

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,135 @@
...
+    }
+    const button = (
+      <Button
+        icon="clipboard-copied-to"
+        iconOnly
+        onClick={this.handleClick}
+        variables={siteVariables => ({
+          colorHover: !this.state.copied && siteVariables.colors.brand[400],
+        })}
+      />
```
layershifter: Can we use shorthand factory instead?

```tsx
CopyToClipboard.defaultProps = {
  button: { icon: 'clipboard-copied-to' },
}

// ...

{Button.create(button, {
  defaultProps: { iconOnly: true },
  overrideProps: () => ({ onClick }) // some work
})
```

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,135 @@
...
+          position="below"
+          align="center"
+          trigger={button}
+          content={{
+            content: popupContent,
+            variables: {
+              ...(hideContent && hiddenPopupContentVariables),
+            },
+          }}
+          on={['hover', 'context']}
```
jurokapsiar: I tried the prototype - it shows the message correctly after click, but then when I move my cursor outside of the button and return back to it, it shows the message again - I believe it's because of 'hover'. I think it should be on={[]}. Not sure if Toolbar would behave better or worse

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
jurokapsiar: would we advise developers to use the factory? given that this is not internal Stardust code, but a prototype?

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
jurokapsiar: if the button would be inside of a popup or dialog, would it still be centered relative to the window or relative to the popup/dialog? The requirement is to center it relative to the window (if not attached).

Next level: if the button is in a popup and popup would be closed, would the notification popup be closed as well?

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
jurokapsiar: Agree with createComponent, also consolidating styles in the provider or theme would be nice.

Would it be possible to create a "Notification" component instead so that you can use without the button? (which I believe Shift's example might be showing as well)

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
layershifter: Factories are public API (_but never advertised_) and exist on almost each component, we will also fix typings in #1865.

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
layershifter: > if the button would be inside of a popup or dialog, would it still be centered relative to the window or relative to the popup/dialog?

It will be rendered inside `Portal`, so it will be centered in window.

***



> Next level: if the button is in a popup and popup would be closed, would the notification popup be closed as well?

Yep, like a nested `Popup` because React's Tree will be unmounted.

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
layershifter: > Would it be possible to create a "Notification" component instead so that you can use without the button?

I think so 👍

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
jurokapsiar: agree - with Tooltip you might not even need to use style overrides

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
jurokapsiar: so this might end up being a Notificaton provider/consumer, something like (pseudocode)

```jsx
const notifications = useContext(Notifications)
const ref = useRef();

const handleClick = () => {
  randomAction()
    .then(() => notifications.show( { content: 'Copied', attachTo: ref} ))
    .fail(() => notifications.show( { content: 'Failed to copy', attachTo: ref} ))
}

return (
  <Ref innerRef={ref}><Button content='copy to clipboard' onClick={handleClick}/></Ref>
)
```

then the provider would just render a tooltip (using target prop).

docs/src/prototypes/CopyToClipboard/index.tsx
lucivpav: When I thought of an element that would wrap the content, I thought of Provider. It's wrong though, thank you for letting me know.

layershifter: `Provider` does more more than just content wrap, it merges top-level and its context that include themes and other things 🔬 

_Usually_, you don't more than one `Provider` per app.

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
lucivpav: We don't want to show the window-centered message when we are in _attached_ mode. In _attached_ mode, we just change the content of the Tooltip.

And to make the centered message more visible, I'm setting `Text`'s size to _larger_.

lucivpav: `hideContent` implies that we should hide the `Tooltip`'s content, as instead we are displaying the centered message.

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
lucivpav: This is intentional. Upon clicking the button, the components gets into a _copied_ state. During the duration of this state, I'm showing 'Copied to clipboard' message. Would you prefer to show the original "Copy commit ID" message instead?

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
lucivpav: @layershifter the consequence of setting `right` and `bottom` props is that user can no longer click on any other elements than the centered message

docs/src/prototypes/CopyToClipboard/index.tsx
lucivpav: I also thought that `<>` is a short for `Provider`. So I decided to choose `Provider` to make it more clear for me

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
jurokapsiar: the message should be visible also when you do not hover/focus on the button.

lucivpav: It is now visible also when the button is not hovered/focused

layershifter (approved):

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+import Notification from './Notification'
+
+export type CopyToClipboardProps = {
+  attached?: boolean
+  pointing?: boolean
+  timeout?: number
+  value: string
+
+  button?: ShorthandValue<ButtonProps>
```
lucivpav: should be `trigger` instead (trigger?: ReactElemenet), like in `Portal`. Need to clone existing element, using React.cloneElement...

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -0,0 +1,93 @@
...
+  noticeText?: ShorthandValue<TextProps>
+  promptText?: ShorthandValue<TextProps>
+}
+
+const CopyToClipboard = createComponent<CopyToClipboardProps>({
```
lucivpav: we don't need createComponent anymore, let it be React.FC

docs/src/prototypes/CopyToClipboard/Notification.tsx
```diff
@@ -0,0 +1,24 @@
+import { Portal, createComponent } from '@stardust-ui/react'
+import * as React from 'react'
+
+type NotificationProps = {
+  children?: React.ReactNode
+  open?: boolean
+}
+
+const Notification = createComponent<NotificationProps>({
```
lucivpav: Consider use case when CopyToClipboard is in a Menu as a MenuItem. Create example prototype to test it. Rename to `NotificationProvider`.

lucivpav: Discuss or pair with Shift...

docs/src/prototypes/CopyToClipboard/NotificationProvider.tsx
```diff
@@ -0,0 +1,50 @@
+import { Portal, createComponent } from '@stardust-ui/react'
+import * as React from 'react'
+
+type NotificationProps = {
+  children?: React.ReactNode
+}
+
+type NotificationContextValue = (value: React.ReactNode, timeout: number) => void
+export const NotificationContext = React.createContext<NotificationContextValue>(() => {
+  throw new Error('Foo')
```
layershifter: Oops, let's add a proper warning there

docs/src/prototypes/CopyToClipboard/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+      ],
+    },
+  ]
+  return <Menu items={items} />
+}
+
+const CopyToClipboardPrototypes: React.FC = () => {
+  const commitID = '3422f7d'
+  return (
+    <PrototypeSection title="Copy to Clipboard">
```
jurokapsiar, line 58: can you please add a note to the beginning of the prototype page, saying:

Note: For screen reader users, make sure to use react-aria-live or similar library to announce the notification.

docs/src/prototypes/CopyToClipboard/NotificationProvider.tsx
lucivpav: How about `  throw new Error('No matching NotificationContext.Provider')`?

layershifter: Will work 👍

PR #1864, opened 2019-08-28
fix(Menu Animation): Adding Animation to Menu

bcalvery:
I started to add animation to Menu using animations from the Teams Theme. Unfortunately this has the same problem as the icons... To add the animation to Menu across the board I would have to add it to MenuItem.tsx, therefore tying the component's code to the theme.

What I'm attempting to do here is create an intermediary, similar to what has been done with the icons which use the `stardust-` prefix. My understanding is stretched pretty thin here but I think I have enough here to at least get my idea across.

@levithomason @mnajdova @codepretty @notandrew


packages/react/src/themes/types.ts
```diff
@@ -548,3 +548,13 @@ export type RequiredIconNames =
...
+
+export type RequiredAnimationNames =
+  | 'stardust-menu-animation'
+  | 'stardust-popup-animation'
+  | 'stardust-other-animation'
+
+// I'm pretty sure I need to do something with the ThemeAnimation interface here
+export type ThemeAnimations = Partial<Record<RequiredAnimationNames, string>> & {
+  [animationName: string]: string
+}
```
bcalvery, line 560: this should probalby use some sort of ThemeAnimation spec or interface?

packages/react/src/themes/base/animations/animationNames.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export const animations: ThemeAnimations = {
+  'stardust-menu-animation': 'fadeEnterFast',
+  'stardust-popup-animation': 'fadeEnterFast',
+  'stardust-other-animation': 'fadeEnterFast',
+}
+
+export const emptyAnimation: string = ''
+
+export default (name: string): string => animations[name] || emptyAnimation
```
bcalvery, line 11: I copied this from the iconNames file, which has some more complex typing which I'm not sure how to translate to the animations.

PR #1871, opened 2019-09-02
fix(Input): background color in dark theme

### LINKED ISSUE #1870: Input with dark theme blends with background
## Bug Report
Using dark theme on an input causes it to use the same color as the body background, and it blends in becoming invisible.  

### Steps
1) go to: https://stardust-ui.github.io/react/components/input/definition
2) Change theme to teams dark

### Expected Result
Input should have a more gray background, similar to dropdown component

### Actual Result
Becomes invisible

image

image

image

image

### Version
0.37.0

lucivpav:
Fixes #1870

Before:
image

After:
image


packages/react/src/themes/teams-dark/components/Input/inputVariables.ts
```diff
@@ -0,0 +1,5 @@
+import { InputVariables } from '../../../base/components/Input/inputVariables'
+
+export default (siteVars): Partial<InputVariables> => ({
+  backgroundColor: siteVars.colors.grey[650],
```
layershifter, line 4: @codepretty is this valid change? Should we use another value from `colorScheme`?

codepretty: No, we don't want to do this. There are valid cases when we want the input to have that background color. I added `inverted` prop to Input in https://github.com/stardust-ui/react/pull/1684. This is what should be used when the input should swap background colors. I updated the [issue](https://github.com/stardust-ui/react/issues/1870) as well.

codepretty (changes requested):
We don't need this change. Please see my comments above.

PR #1872, opened 2019-09-02
feat(Alert): add `fitted` prop

layershifter:
### `fitted` prop

This PR adds a new prop to `Alert` component, it allows to only take up the width of its content.

***

Also adds an example into `Usage` with fixed container width that wraps `Alert`.

### DIFF
--- a/docs/src/examples/components/Alert/Usage/AlertExampleWidth.shorthand.tsx
+++ b/docs/src/examples/components/Alert/Usage/AlertExampleWidth.shorthand.tsx
@@ -1,0 +1,20 @@
+import { useRangeKnob } from '@stardust-ui/docs-components'
+import { Alert } from '@stardust-ui/react'
+import * as React from 'react'
+
+const AlertExampleWidth = () => {
+  const [width] = useRangeKnob({ name: 'width', initialValue: '500px' })
+
+  return (
+    <div style={{ minWidth: 350, maxWidth: 800, width }}>
+      <Alert
+        actions={[{ content: 'Join and add the room', primary: true }]}
+        header="There is a conference room close to you."
+        dismissible
+        icon="screencast"
+      />
+    </div>
+  )
+}
+
+export default AlertExampleWidth
--- a/packages/react/src/components/Alert/Alert.tsx
+++ b/packages/react/src/components/Alert/Alert.tsx
@@ -60,7 +60,7 @@ export interface AlertProps
   attached?: boolean | 'top' | 'bottom'
 
   /** An alert can only take up the width of its content. */
-  compact?: boolean
+  fitted?: boolean
 
   /** An alert may be formatted to display a danger message. */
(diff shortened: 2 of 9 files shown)


docs/src/examples/components/Alert/Usage/AlertExampleWidth.shorthand.tsx
```diff
@@ -0,0 +1,20 @@
+import { useRangeKnob } from '@stardust-ui/docs-components'
+import { Alert } from '@stardust-ui/react'
+import * as React from 'react'
+
+const AlertExampleWidth = () => {
+  const [width] = useRangeKnob({ name: 'width', initialValue: '500px' })
```
lucivpav, line 6: Wouldn't it be better to specify the bounds on knob instead? Let's create a ticket for it?

layershifter: I am working on PR that will introduce it

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -59,6 +59,9 @@ export interface AlertProps
   /** Controls Alert's relation to neighboring items. */
   attached?: boolean | 'top' | 'bottom'
 
+  /** An alert can only take up the width of its content. */
+  compact?: boolean
```
mnajdova: Why not `fitted`? It feels awkward to have two different terms across the components for the same thing? If we decide that one is better then the other, we can rename all of them at once. What do you think?

docs/src/examples/components/Alert/Usage/AlertExampleWidth.shorthand.tsx
layershifter: Here it is #1876.

packages/react/src/components/Alert/Alert.tsx
layershifter: Good catch, `fitted` will work better there and will be consistent 👍

mnajdova (approved):

packages/react/src/themes/teams/components/Alert/alertVariables.ts
```diff
@@ -55,12 +56,13 @@ export default (siteVars: SiteVariablesPrepared): AlertVariables => {
     color: siteVars.colors.grey[500],
     fontWeight: siteVars.fontWeightRegular,
     minHeight,
-    padding: `0 0 0 ${pxToRem(16)}`,
+    padding: `0 ${pxToRem(16)}`,
```
layershifter, line 59: @codepretty can you please check this one? Is it correct?

codepretty: Looks good to me

codepretty (approved):

PR #1875, opened 2019-09-02
fix(factories): add types to createShorthandFactory() methods

### LINKED ISSUE #1865: Typings: Factories are not properly typed
### Steps
1. Use any `.create()` method:
```tsx
{Avatar.create(Symbol("foo"), "FOO", "BAR")}
```
https://codesandbox.io/s/stardust-ui-example-0tkkx

Issue exists because all our `.create()` functions are typed as `Function` 💥 

image

### Expected Result
Types error.

### Actual Result
No type error 💣 

### Version
0.37.0

lucivpav:
Fixes #1865 

The new type is
```tsx
export type ShorthandFactory = (
  value: ShorthandValue<UIComponentProps>,
  options?: CreateShorthandOptions,
) => React.ReactElement | null | undefined
```

## Won't compile
```tsx
{Avatar.create(Symbol("foo"), "FOO", "BAR")}
{Avatar.create(Symbol("foo"))}
```

## Will compile
```tsx
avatar?: ShorthandValue<AvatarProps> = ...
{Avatar.create(avatar}, {defaultProps: {...}, overrideProps: {...}}
{Avatar.create(avatar}}
```

And this will unfortunately also compile
```tsx
button?: ShorthandValue<ButtonProps> = ...
{Avatar.create(button}}
```

### DIFF
--- a/docs/src/components/ColorSchemes.tsx
+++ b/docs/src/components/ColorSchemes.tsx
@@ -57,7 +57,7 @@ const ColorSchemes = createComponent<ColorVariantsProps>({
     return (
       <div className={classes.root}>
         <Grid columns={columns}>
-          {headers && headers.map(header => Header.create(header))}
+          {headers && headers.map(header => Header.create(header, {}))}
           {elements}
         </Grid>
       </div>
--- a/docs/src/components/ColorSchemes.tsx
+++ b/docs/src/components/ColorSchemes.tsx
@@ -57,7 +57,7 @@ const ColorSchemes = createComponent<ColorVariantsProps>({
     return (
       <div className={classes.root}>
         <Grid columns={columns}>
-          {headers && headers.map(header => Header.create(header, {}))}
+          {headers && headers.map(header => Header.create(header))}
           {elements}
         </Grid>
       </div>
--- a/packages/react/src/components/Accordion/AccordionTitle.tsx
+++ b/packages/react/src/components/Accordion/AccordionTitle.tsx
@@ -66,7 +66,7 @@ export interface AccordionTitleProps
 class AccordionTitle extends UIComponent<WithAsProp<AccordionTitleProps>, any> {
   static displayName = 'AccordionTitle'
 
-  static create: ShorthandFactory
+  static create: ShorthandFactory<AccordionTitleProps>
 
   static className = 'ui-accordion__title'
 
@@ -107,7 +107,8 @@ class AccordionTitle extends UIComponent<WithAsProp<AccordionTitleProps>, any> {
 
   renderComponent({ ElementType, classes, unhandledProps, styles, accessibility }) {
     const { contentRef, children, content, indicator, active } = this.props
-    const indicatorWithDefaults = indicator === undefined ? {} : indicator
+    const defaultIndicator = { name: active ? 'stardust-arrow-down' : 'stardust-arrow-end' }
+    const indicatorWithDefaults = indicator === undefined ? defaultIndicator : indicator
 
     const contentElement = (
       <Ref innerRef={contentRef}>
@@ -119,7 +120,6 @@ class AccordionTitle extends UIComponent<WithAsProp<AccordionTitleProps>, any> {
           {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.content, unhandledProps)}
           start={Icon.create(indicatorWithDefaults, {
             defaultProps: {
-              name: active ? 'stardust-arrow-down' : 'stardust-arrow-end',
               styles: styles.indicator,
             },
           })}
--- a/packages/react/src/components/Accordion/AccordionTitle.tsx
+++ b/packages/react/src/components/Accordion/AccordionTitle.tsx
@@ -14,7 +14,7 @@ import {
   commonPropTypes,
   rtlTextContainer,
   applyAccessibilityKeyHandlers,
-  CreateShorthandFactoryResult,
+  ShorthandFactory,
 } from '../../lib'
 import { WithAsProp, ComponentEventHandler, ShorthandValue, withSafeTypeForAs } from '../../types'
 import Icon, { IconProps } from '../Icon/Icon'
@@ -66,7 +66,7 @@ export interface AccordionTitleProps
 class AccordionTitle extends UIComponent<WithAsProp<AccordionTitleProps>, any> {
   static displayName = 'AccordionTitle'
 
-  static create: CreateShorthandFactoryResult
+  static create: ShorthandFactory
 
   static className = 'ui-accordion__title'
 
--- a/packages/react/src/components/Accordion/AccordionTitle.tsx
+++ b/packages/react/src/components/Accordion/AccordionTitle.tsx
@@ -14,6 +14,7 @@ import {
   commonPropTypes,
   rtlTextContainer,
   applyAccessibilityKeyHandlers,
+  CreateShorthandFactoryResult,
 } from '../../lib'
 import { WithAsProp, ComponentEventHandler, ShorthandValue, withSafeTypeForAs } from '../../types'
 import Icon, { IconProps } from '../Icon/Icon'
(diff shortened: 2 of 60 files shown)


packages/react/src/components/Tree/Tree.tsx
```diff
@@ -263,7 +263,7 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
           const isSubtree = hasSubtree(item)
           const isSubtreeOpen = isSubtree && this.isActiveItem(item['id'])
 
-          const renderedItem = TreeItem.create(item, {
+          const renderedItem: any = TreeItem.create(item, {
```
lucivpav: Without `any`, a strange TS error shows up: 
```
Type 'ShorthandValue<TreeItemProps & { kind?: never; }>' is not assignable to type 'ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)>) | (new (props: any) => Component<any, any, any>)>[]'.
  Type 'string' is not assignable to type 'ReactElement<any, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)>) | (new (props: any) => Component<any, any, any>)>[]'.ts(2322)
```
I didn't find a way to resolve it

packages/react/src/lib/factories.ts
```diff
@@ -10,6 +10,7 @@ import {
   ShorthandRenderer,
 } from '../types'
 import { mergeStyles } from './mergeThemes'
+import { UIComponentProps } from '.'
```
layershifter: Ouch, circular reference

```suggestion
import { UIComponentProps } from './commonPropInterfaces'
```

packages/react/src/lib/factories.ts
```diff
@@ -91,6 +92,10 @@ type CreateShorthandFactoryConfigInner<TPropName = string> = {
   mappedArrayProp?: TPropName
 }
 export type CreateShorthandFactoryConfig = CreateShorthandFactoryConfigInner
+export type CreateShorthandFactoryResult = (
```
layershifter: Actually, the result of `createShorthandFactory` is `ShorthandFactory`

```suggestion
export type ShorthandFactory = (
```

packages/react/src/lib/factories.ts
```diff
@@ -91,6 +92,10 @@ type CreateShorthandFactoryConfigInner<TPropName = string> = {
...
+  a: ShorthandValue<UIComponentProps>,
```
layershifter:
```suggestion
  value: ShorthandValue<UIComponentProps>,
```

packages/react/src/lib/factories.ts
```diff
@@ -91,6 +92,10 @@ type CreateShorthandFactoryConfigInner<TPropName = string> = {
...
+  b: CreateShorthandOptions,
```
layershifter:
```suggestion
  options?: CreateShorthandOptions,
```

This param is also optional

packages/react/src/lib/factories.ts
layershifter:
```suggestion
  a: ShorthandValue<P>,
```

Actually, let's make it generic and accept component props there. Hope that it will work

docs/src/components/ColorSchemes.tsx
```diff
@@ -57,7 +57,7 @@ const ColorSchemes = createComponent<ColorVariantsProps>({
     return (
       <div className={classes.root}>
         <Grid columns={columns}>
-          {headers && headers.map(header => Header.create(header))}
+          {headers && headers.map(header => Header.create(header, {}))}
```
layershifter:
```suggestion
          {headers && headers.map(header => Header.create(header))}
```

Second param should be optional

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -743,8 +743,10 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
       !loading &&
         items.length === 0 &&
         ListItem.create(noResultsMessage, {
-          key: 'no-results-message',
-          styles: styles.noResultsMessage,
+          defaultProps: {
+            key: 'no-results-message',
+            styles: styles.noResultsMessage,
+          },
```
layershifter, line 749: ❤️

packages/react/src/components/Tree/Tree.tsx
layershifter: Looks like TS bug, I suggest to move any to other place instead

```tsx
const renderItems = (items: ShorthandCollection<TreeItemProps>) => {
      return items.reduce<React.ReactElement[]>(
        (renderedItems, item): React.ReactElement[] => {
          const itemForRender = itemsForRender[item['id']]
          const { elementRef, ...restItemForRender } = itemForRender
          const isSubtree = hasSubtree(item)
          const isSubtreeOpen = isSubtree && this.isActiveItem(item['id'])

          const renderedItem = TreeItem.create(item, {
            defaultProps: {
              className: Tree.slotClassNames.item,
              open: isSubtreeOpen,
              renderItemTitle,
              key: item['id'],
              ...restItemForRender,
            },
            overrideProps: this.handleTreeItemOverrides,
          })

          // Only need refs of the items that spawn subtrees, when they need to be focused
          // by any of their children, using Arrow Left.
          const finalRenderedItem = isSubtree ? (
            <Ref key={item['id']} innerRef={elementRef}>
              {renderedItem}
            </Ref>
          ) : (
            renderedItem
          )

          return [
            ...renderedItems,
            finalRenderedItem,
            ...[isSubtreeOpen ? renderItems(item['items']) : []] as any,
          ]
        },
        [],
      )
    }
```

packages/react/src/lib/factories.ts
```diff
@@ -91,6 +92,10 @@ type CreateShorthandFactoryConfigInner<TPropName = string> = {
...
+) => React.ReactElement<Props> | null | undefined
```
layershifter:
```suggestion
) => React.ReactElement | null | undefined
```

It defaults to `any`

packages/react/src/lib/factories.ts
lucivpav: Yes it works

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -214,7 +215,8 @@ class MenuItem extends AutoControlledComponent<WithAsProp<MenuItemProps>, MenuIt
     } = this.props
     const { menuOpen } = this.state
 
-    const indicatorWithDefaults = indicator === undefined ? {} : indicator
+    const defaultIndicator = { name: vertical ? 'stardust-arrow-end' : 'stardust-arrow-down' }
+    const indicatorWithDefaults = indicator === undefined ? defaultIndicator : indicator
```
layershifter, line 219: Are these changes required? Can we revert them?

packages/react/src/components/Accordion/AccordionTitle.tsx
```diff
@@ -106,7 +107,8 @@ class AccordionTitle extends UIComponent<WithAsProp<AccordionTitleProps>, any> {
 
   renderComponent({ ElementType, classes, unhandledProps, styles, accessibility }) {
     const { contentRef, children, content, indicator, active } = this.props
-    const indicatorWithDefaults = indicator === undefined ? {} : indicator
+    const defaultIndicator = { name: active ? 'stardust-arrow-down' : 'stardust-arrow-end' }
+    const indicatorWithDefaults = indicator === undefined ? defaultIndicator : indicator
```
layershifter, line 111: https://github.com/stardust-ui/react/pull/1875/files#diff-d5390340926344f30b1c6712c4e548c7R219

packages/react/src/components/Menu/MenuItem.tsx
lucivpav: It is required, because Type `{}` is not assignable to type `Props<IconProps>`.

layershifter (approved):

lucivpav:
Another PR: `CreateShorthandOptions` should be typed as `CreateShorthandOptions<P>`

PR #1876, opened 2019-09-02
feat(Knobs): improve `useRangeKnob()` and add `LogInspector`/`useLogKnob()`

layershifter:
## `useLogKnob()` and `LogInspector`

This PR adds and knob and component, this drastically simplifies examples and allows to log callbacks. There are two examples of `useLogKnob()`: `PortalExampleOpen.tsx` and `DialogExampleCallbacks.shorthand.tsx`.

Second is more advanced as contains custom output formatter:

callback-example

## `useRangeKnob()`

Was improved to support `min`, `max` and `step` props. Logic with parsing values was moved out from `KnobRange` so it will be easier to integrate custom components.


packages/docs-components/package.json
```diff
@@ -20,6 +20,7 @@
   "main": "dist/commonjs/index.js",
   "module": "dist/es/index.js",
   "peerDependencies": {
+    "prettier": "^1.18.2",
```
layershifter, line 23: We are using it under `packages/docs-components/src/CodeSnippet/formatCode.ts`.

packages/docs-components/src/knobs/useLogKnob.ts
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react'
+import { LogContext } from './KnobContexts'
+
+const defaultFormatter = (name: string) => `${new Date().toLocaleTimeString()}: ${name}`
+
+const useLogKnob = <T = (...args: any[]) => any>(
+  name: string,
+  callback?: T,
+  formatter: Function = defaultFormatter,
```
lucivpav: Can we specify the `formatter` type, instead of using a general `Function`? For instance, it would be nice to specify that the function should return `string`.

layershifter: Done 👍

PR #1882, opened 2019-09-03
perf(Tooltip): fix Popper positioning called when unnecessary

### LINKED ISSUE #1880: Popper in Tooltip always executes positioning code
## Bug Report

The `Tooltip`'s `Popper`'s positioning code is always executed, even if the tooltip content is not visible to the user. This is a performance issue if we have many `Tooltip` components.

### Expected Result
The content positioning code is only executed when the content is visible to the user.

### Actual Result
The content positioning code is executed even if the content is not visible 💣

### Version
0.37.0

lucivpav:
Fixes #1880

### DIFF
--- a/packages/react/src/components/Tooltip/Tooltip.tsx
+++ b/packages/react/src/components/Tooltip/Tooltip.tsx
@@ -236,13 +236,14 @@ export default class Tooltip extends AutoControlledComponent<TooltipProps, Toolt
     accessibility: ReactAccessibilityBehavior,
   ): JSX.Element {
     const { align, position, target } = this.props
+    const { open } = this.state
 
     return (
       <Popper
         pointerTargetRef={this.pointerTargetRef}
         align={align}
         position={position}
-        enabled={this.state.open}
+        enabled={open}
         rtl={rtl}
         targetRef={target ? toRefObject(target) : this.triggerRef}
         children={this.renderPopperChildren.bind(this, tooltipPositionClasses, rtl, accessibility)}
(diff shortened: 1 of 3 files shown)


packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -238,6 +242,7 @@ export default class Tooltip extends AutoControlledComponent<TooltipProps, Toolt
         pointerTargetRef={this.pointerTargetRef}
         align={align}
         position={position}
+        enabled={this.state.open}
```
layershifter:
```diff
    const { align, position, target } = this.props
+    const { open } = this.state
```

```suggestion
        enabled={open}
```

_nit:_ to aligned let's use destruction

packages/react/test/specs/components/Tooltip/Tooltip-test.tsx
```diff
@@ -60,4 +60,13 @@ describe('Tooltip', () => {
       )
     })
   })
+
+  describe('open', () => {
+    it('is passed to "Popper" as "enabled"', () => {
+      const wrapper = mountWithProvider(<Tooltip trigger={<button />} content="Foo" />)
+      expect(wrapper.find('Popper').prop('enabled')).toBe(false)
+      wrapper.setProps({ open: true })
```
layershifter, line 69:
```suggestion

      wrapper.setProps({ open: true })
```

nit: empty line

layershifter (approved):

PR #1884, opened 2019-09-03
fix(Dropdown): pass custom window to Downshift

silviuaavram:
https://github.com/downshift-js/downshift/pull/755

Can be merged without the Downshift fix.

### DIFF
--- a/packages/react/src/components/Dropdown/Dropdown.tsx
+++ b/packages/react/src/components/Dropdown/Dropdown.tsx
@@ -430,6 +430,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
           highlightedIndex={highlightedIndex}
           onStateChange={this.handleStateChange}
           labelId={this.props['aria-labelledby']}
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -430,6 +430,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
           highlightedIndex={highlightedIndex}
           onStateChange={this.handleStateChange}
           labelId={this.props['aria-labelledby']}
+          environment={this.context.target.defaultView}
```
layershifter, line 433: https://developer.mozilla.org/en-US/docs/Web/API/Document/defaultView
Works in IE11, I am happy 👍

layershifter (approved):
Changelog?

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: the exact place I got it from!

PR #1887, opened 2019-09-03
feat(shadows): updating the shadow variables

notandrew:
Removing extra 0s, fixing shadowLevel4, and adding the 'dark' style.


packages/react/src/themes/teams/siteVariables.ts
```diff
@@ -26,11 +26,11 @@ export const focusOuterBorderColor = colors.black
...
-export const shadowLevel2 = '0 .4rem 0.7rem -0.1rem rgba(0, 0, 0, .1)'
-export const shadowLevel3 = '0 .8rem 1rem -0.2rem rgba(0, 0, 0, .1)'
-export const shadowLevel4 = '0 .6rem 1.8rem -0.4rem rgba(0, 0, 0, .1)'
-
-export const shadowLevel1Darker = '0 .2rem .4rem -.075rem rgba(0, 0, 0, 0.5)'
+export const shadowLevel2 = '0 .4rem .7rem -.1rem rgba(0, 0, 0, .1)'
+export const shadowLevel3 = '0 .8rem 1rem -.2rem rgba(0, 0, 0, .1)'
+export const shadowLevel4 = '0 1.6rem 1.8rem -.4rem rgba(0, 0, 0, .1)'
+export const shadowLevel1Dark = '0 .2rem .4rem -.075rem rgba(0, 0, 0, .25)'
+export const shadowLevel1Darker = '0 .2rem .4rem -.075rem rgba(0, 0, 0, .5)'
```
codepretty, line 33: Can we remove `shadowLevel1Darker`? Is there a reason to keep it?

notandrew: We should keep it because we have it in the current app.

codepretty: We could remove it and add this as a Breaking Change if we don't want both values to actually be used in the app. If we don't want to cause churn for the main app, we could just update the `shadowLevel1Darker` variable instead of adding a new variable.

notandrew: We currently have this in the angular app:
.shadow-level-1 { box-shadow: 0 .2rem .4rem -.075rem rgba(0,0,0,.1)}
shadow-level-1-dark { box-shadow: 0 .2rem .4rem -.075rem rgba(0,0,0,.25) }
.shadow-level-1-darker { box-shadow: 0 .2rem .4rem -.075rem rgba(0,0,0,.5) }
.shadow-level-2 { box-shadow: 0 .4rem .7rem -.1rem rgba(0,0,0,.1) }
.shadow-level-3 { box-shadow: 0 .8rem 1rem -.2rem rgba(0,0,0,.1) }
.shadow-level-4 { box-shadow: 0 1.6rem 1.8rem -.4rem rgba(0,0,0,.1) }

codepretty: Ah, got it. Looks good to me.

codepretty (approved):

PR #1890, opened 2019-09-04
feat(Tree): add virtualized tree prototype

silviuaavram:
Attempt to provide virtualization capability to a `Tree` via a render prop.

`react-virtualized` adds a `style` param to the component it renders, and it needs to be there at render.

Tried `React.cloneElement` to continue to pass rendered items to the Virtualized render prop but it's not working as expected.

Added another way of doing it by Tree passing render callbacks to the Virtualized render prop, that will render it along with the `style` prop, and that works.

Work in progress.


packages/react/src/components/Tree/Tree.tsx
```diff
@@ -263,32 +266,42 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
-            },
-            overrideProps: this.handleTreeItemOverrides,
-          })
-
-          // Only need refs of the items that spawn subtrees, when they need to be focused
-          // by any of their children, using Arrow Left.
-          const finalRenderedItem = isSubtree ? (
+          const renderedItem = (
+            <Ref key={item['id']} innerRef={elementRef}>
+              {TreeItem.create(item, {
```
levithomason: This line does 3 things: 
- resolve user's `item` value into ElementType and props
- Mix in defaultProps and overrideProps
- Create an element from the final ElementType and mixed props

It needs to only do items 1 and 2, get final ElementType and props.  The creation of the JSX should be done at a later phase by the virtualized step where it can mix in additional props, like style.

There needs to be some prop or pattern available to the user to get the computed ElementType and props before the JSX is created.  It should also allow them to create the JSX themselves.

Consider updating the `virtualized` prop to take this data instead and I'd also rename it to something like `renderItems` since we don't know what someone will use it for.

docs/src/prototypes/VirtualizedTree/VirtualizedTree.tsx
```diff
@@ -0,0 +1,56 @@
...
+  cache = null
+
+  constructor(props) {
+    super(props)
+
+    this.cache = new CellMeasurerCache({
+      defaultHeight: 20,
+      fixedWidth: true,
+    })
+  }
```
layershifter: nit, `constructor` is redundant

```
class TreeVirtualizer extends React.Component<TreeVirtualizerProps> {
  cache = new CellMeasurerCache({
    defaultHeight: 20,
    fixedWidth: true,
  })
```

docs/src/prototypes/VirtualizedTree/VirtualizedTree.tsx
```diff
@@ -0,0 +1,56 @@
...
+        rowHeight={this.cache.rowHeight}
+        rowRenderer={this.rowRenderer}
+        estimatedRowSize={20}
+        height={300}
+        rowCount={this.props.renderedItems.length}
+        width={600}
+      />
+    )
+  }
+}
```
layershifter, line 37: #### ☢️ NIT
Do we really need a class component there?

docs/src/prototypes/VirtualizedTree/itemsGenerator.ts
```diff
@@ -0,0 +1,24 @@
+import * as _ from 'lodash'
+
+function getItems(minItems = 20, maxItems = 40, maxLevel = 1) {
+  function getItemsNumber(minItems, maxItems) {
+    return _.random(minItems, maxItems)
+  }
+
+  function generateLevel(level, parent = '') {
+    const result = []
+    for (let index = 0; index < getItemsNumber(minItems, maxItems); index++) {
```
layershifter: `_.times` from Lodash?

packages/react/package.json
```diff
@@ -25,7 +25,8 @@
     "prop-types": "^15.7.2",
     "react-fela": "^10.6.1",
     "react-is": "^16.6.3",
-    "react-resize-detector": "^4.2.0"
+    "react-resize-detector": "^4.2.0",
+    "react-virtualized": "^9.21.1"
```
layershifter: Oops 💣 Please move this into root `package.json`

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -104,6 +108,7 @@ class TreeItem extends UIComponent<WithAsProp<TreeItemProps>, TreeItemState> {
     ...commonPropTypes.createCommon({
       content: false,
     }),
+    contentRef: PropTypes.object,
```
```suggestion
    contentRef: customPropTypes.ref,
```

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -343,6 +343,11 @@ class Sidebar extends React.Component<any, any> {
         title: { content: 'MenuButton', as: NavLink, to: '/menu-button' },
         public: false,
       },
+      {
+        key: 'virtualized-tree',
+        title: { content: 'VirtualizedTree', as: NavLink, to: '/virtualized-tree' },
+        public: false,
```
```suggestion
        public: true,
```

Should be `true`?

docs/src/prototypes/VirtualizedTree/VirtualizedTree.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+interface TreeVirtualizerProps {
+  renderedItems: React.ReactElement[]
+}
+
+function TreeVirtualizer(props: TreeVirtualizerProps) {
+  const cache = new CellMeasurerCache({
+    defaultHeight: 20,
+    fixedWidth: true,
+  })
```
layershifter, line 14: Should it be persistent between renders?

```tsx
const [cache] = React.useState(() => 
  new CellMeasurerCache({
    defaultHeight: 20,
    fixedWidth: true,
  })
)
```

layershifter:
tree-v

I am not that it's a PR blocker, but there are render delays on few opened items.

silviuaavram:
@layershifter if you go to http://bvaughn.github.io/react-virtualized/#/components/List and make the row size ~20 px those take a while to render as well. probably it will perform better if we had fewer, taller items in our example.

docs/src/prototypes/VirtualizedTree/VirtualizedTree.tsx
silviuaavram: not sure, I just looked at his example. 

https://github.com/bvaughn/react-virtualized/blob/a390b2aa3652bdf488a1eb37348d98558617dcfd/source/CellMeasurer/CellMeasurer.DynamicHeightList.example.js#L19

probably not needed.

layershifter (approved):

jurokapsiar (approved):

PR #1892, opened 2019-09-05
docs(props): put props on a single page, fix props links

### LINKED ISSUE #1878: Props links do not point to the exact location
### Steps
1. Navigate to https://stardust-ui.github.io/react/components/toolbar/props#ToolbarMenu
2. Observe the link to `ToolbarMenuItemProps`
3. Clicking the link redirects you to https://stardust-ui.github.io/react/components/toolbar 💣

### Expected Result
The link points to https://stardust-ui.github.io/react/components/toolbar/props#ToolbarMenuItem

### Actual Result
The link points to https://stardust-ui.github.io/react/components/toolbar 💣

### Version
0.37.0

lucivpav:
Fixes #1878. Furthermore, all props under one tab were collapsed on a single page, instead of having a dropdown prop selection.


levithomason:
It seems like we could fix this easier by just iterating all the props on a single page, opposed to having a dropdown to select the component props you want to see.

Thoughts?

lucivpav:
> It seems like we could fix this easier by just iterating all the props on a single page, opposed to having a dropdown to select the component props you want to see.
> 
> Thoughts?

I agree, that way the logic gets easier. I suggest we also have an outline of the components for which we have props. That is because the Props page can be long and it would be hard to navigate. 
image
So the outline would be similar compared to what we already have in Accessibility tab.
image

@layershifter do you think we should get rid of the props dropdown and list all the props on a single page instead?

lucivpav:
Shift suggests that we could use a submenu of Props to list the various options. Something like:
image
Then we can use react-router instead of programmatically maintaining correct dropdown value etc.
Codesandbox example: https://codesandbox.io/s/stardust-ui-example-vgy5i

lucivpav:
> It seems like we could fix this easier by just iterating all the props on a single page, opposed to having a dropdown to select the component props you want to see.
> 
> Thoughts?

I can't make the links working if all is on single page. I suggest that would be a separate PR and we merge this change as is.

docs/src/components/ComponentDoc/ComponentProps/ComponentPropsOutline.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+  const items: string[] = _.map(displayNames, displayName =>
+    displayName === parentDisplayName
+      ? displayName
+      : displayName.replace(parentDisplayName, `${parentDisplayName}`),
+  )
+  return (
+    <ul>
+      {_.map(items, item => (
+        <li>{link(item, `#${_.kebabCase(item)}`)}</li>
```
layershifter: Missing `key`

```suggestion
        <li key={item}>{link(item, `#${_.kebabCase(item)}`)}</li>
```

docs/src/components/ComponentDoc/ComponentPropsTable/ComponentPropsRow.tsx
```diff
@@ -23,11 +24,14 @@ const ComponentPropValue: React.FunctionComponent<ComponentPropType> = props =>
     const kindParam = parameters[1] && parameters[1].name !== 'never'
     const kindIsVisible = name === 'ShorthandCollection' && kindParam
 
+    const showHash = parentInfo.parentDisplayName || _.size(getComponentGroup(componentName)) > 1
+    const propsSection = showHash ? `#${_.kebabCase(componentName)}` : ''
```
layershifter, line 28: Why we can't always add hash? Is it bad?

docs/src/components/DocsLayout.tsx
```diff
@@ -59,7 +67,8 @@ class DocsLayout extends React.Component<any, any> {
     // Anchor links has issues with <base>
     // https://stackoverflow.com/questions/8108836/make-anchor-links-refer-to-the-current-page-when-using-base
     document.querySelectorAll('a.anchor-link').forEach(link => {
-      link.setAttribute('href', `${document.location.href}${link.getAttribute('href')}`)
+      const value = `${this.getUrlWithoutHash(document.location.href)}${link.getAttribute('href')}`
```
layershifter: Can you please explain why `getUrlWithoutHash()` is needed there?

docs/src/components/ComponentDoc/ComponentProps/ComponentPropsOutline.tsx
```diff
@@ -0,0 +1,28 @@
...
+      ))}
+    </ul>
+  )
+}
+
+ComponentPropsOutline.propTypes = {
```
layershifter: Let's use TS types instead

docs/src/components/ComponentDoc/ComponentProps/ComponentPropsOutline.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { link } from 'docs/src/utils/helpers'
+
+const ComponentPropsOutline: any = ({ displayNames, parentDisplayName }) => {
+  if (displayNames.length < 2) return null
+
+  const items: string[] = _.map(displayNames, displayName =>
+    displayName === parentDisplayName
+      ? displayName
+      : displayName.replace(parentDisplayName, `${parentDisplayName}`),
+  )
```
layershifter: What it should do? I don't see any difference in links:

image

docs/src/components/ComponentDoc/ComponentProps/ComponentProps.tsx
```diff
@@ -2,74 +2,62 @@ import * as _ from 'lodash'
...
-              <ComponentTableProps componentName={activeDisplayName} />
-            </>
-          </Flex.Item>
-        )}
+        {_.map(displayNames, displayName => {
+          const { docblock } = (componentGroup[displayName] || {}) as any
+          const description = _.get(docblock, 'description', [])
+          const showHeader = displayNames.length > 1
+          return (
+            <Flex.Item>
```
layershifter: You don't need `Flex.Item` there as I see

docs/src/components/ComponentDoc/ComponentProps/ComponentProps.tsx
```diff
@@ -2,74 +2,62 @@ import * as _ from 'lodash'
...
-          <Flex.Item style={propsContainerStyle}>
-            <>
-              <ComponentPropsDescription description={description} />
-              <ComponentTableProps componentName={activeDisplayName} />
-            </>
-          </Flex.Item>
-        )}
+        {_.map(displayNames, displayName => {
+          const { docblock } = (componentGroup[displayName] || {}) as any
+          const description = _.get(docblock, 'description', [])
```
layershifter:
```suggestion
          const description = _.get(componentGroup, [displayName, 'docblock', 'description'], '')
```

One liner 📦

docs/src/components/ComponentDoc/ComponentProps/ComponentPropCard.tsx
```diff
@@ -0,0 +1,25 @@
...
+    const { name, description } = this.props
+    return (
+      <Segment styles={cardStyle}>
+        <div>{description}</div>
+        <Divider />
+        <ComponentTableProps componentName={name} />
+      </Segment>
+    )
+  }
+}
```
layershifter: Function component?

docs/src/components/ComponentDoc/ComponentPropsTable/ComponentPropsRow.tsx
lucivpav: We don't add it in case when its the main prop of a component (e.g. ButtonProp, but not ToolbarMenuItemProp) and there are no other related props (in that case, the prop page does not have the header to navigate to, it only contains one prop table)

docs/src/components/DocsLayout.tsx
lucivpav: Because `document.location.href` can already end with a hash, so we would have two hashes in a link, which is wrong

docs/src/components/ComponentDoc/ComponentProps/ComponentPropsOutline.tsx
lucivpav: It's a cruft. Removed.

docs/src/components/DocsLayout.tsx
layershifter: So we can use `location.pathname` instead? Will it work?

lucivpav: Use `origin`, `pathname` instead

layershifter (approved):

PR #1895, opened 2019-09-06
perf: use `Object.assign` in `@babel/plugin-proposal-object-rest-spread`

layershifter:
_Thanks to Mario Baricevic for pointing to this._

## 🐢 Issue

We use `@babel-preset-env` to transform our code.

According to `.browserslistrc` it will compile our code to be used in IE11, there are no support for [`Object.assign()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) or [object spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) and it means that Babel helper will be used. However, we have already required `Object.assign()` polyfill and it means that we can use it instead of Babel helper.

#### Example input

```ts
const a = { foo: 'bar'}
const b = { baz: 'qux'}
const c = { ...a, ...b }
```

#### Example output before

```js
import _objectSpread from "@babel/runtime/helpers/esm/objectSpread";
var a = { foo: 'bar' };
var b = { baz: 'qux' };
var c = _objectSpread({}, a, b);
```

[@babel/runtime/helpers/esm/objectSpread.js](https://cdn.jsdelivr.net/npm/@babel/runtime@7.5.5/helpers/esm/objectSpread.js) contains own implementation which is definitely slower.

#### Example output with `{ loose: true, useBuiltIns: true }`

```js
var a = { foo: 'bar' };
var b = { baz: 'qux' };
var c = Object.assign({}, a, b);
```

## 🏁 Measures

```sh
cross-env PERF=true gulp perf --filter *Chat.perf.tsx --times=100
```

I run the command above 4 times to test changes. During each sample run perf example was run 100 times.

Performance win is around **13%**, see results below.

#### Before

| Example | Avg | Median | 
| --- |  ---  |  ---  | 
| Chat.perf.tsx (Sample 1) | 576.69 | 575.25 | 
| Chat.perf.tsx (Sample 2) | 583.62 | 577.66 | 
| Chat.perf.tsx (Sample 3) | 588.64 | 580.58 | 
| Chat.perf.tsx (Sample 4) | 578.11 | 575.83 | 
| Total | **581.765** | |

#### After

| Example | Avg | Median | 
| --- |  ---  |  ---  | 
| Example | Avg |  Median | 
| Chat.perf.tsx (Sample 1) | 504.76 | 503.9 | 
| Chat.perf.tsx (Sample 2) | 507.85 | 505.52 | 
| Chat.perf.tsx (Sample 3) | 505.25 | 504.13 | 
| Chat.perf.tsx (Sample 4) | 513.71 | 507.39 | 
| Total | **507.8925** | |


docs/src/examples/components/Chat/Performance/Chat.perf.tsx
```diff
@@ -28,7 +28,7 @@ const ChatExample = () => {
       }}
     >
       <Chat
-        items={_.times(30, i => ({
+        items={_.times(100, i => ({
```
layershifter, line 31: Adjusted to have cleaner results.

lucivpav (approved):

Issue #755, opened 2019-01-22
Add a TextArea component


levithomason:
We need a `<TextArea />` as in SUIR: https://react.semantic-ui.com/addons/text-area.

Users are trying to handle this use case with an `Input` and it is not going well: #754.

PR #1897, opened 2019-09-06
feat(TextArea): add component

lucivpav:
Implements #755

|Prop|Meaning|
|---|---|
|accessibility?|Accessibility behavior if overridden by the user.|
|defaultValue?|The default value of the text area.|
|onChange?|Called on change.|
|value?|The value of the text area.|
|disabled?|The text area becomes read-only.|

[Semantic UI's TextArea for reference](https://react.semantic-ui.com/addons/text-area)
[Mozilla's docs on `<textarea>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea)


docs/src/components/ComponentPlayground/ComponentPlayground.tsx
```diff
@@ -27,6 +27,7 @@ const unsupportedComponents = [
   'Portal',
   'Provider',
   'RadioGroup',
+  'TextArea',
```
lucivpav: There is a problem (along the lines: "A component is changing an uncontrolled input of type text to be controlled error in ReactJS") when one plays with `placeholder` in the playground. So I decided to disable it.

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,123 @@
...
+  /** The value of the text area. */
+  value?: string
+
+  /** The maximum number of characters allowed in the text area. */
+  maxLength?: number
+
+  /**
+   * Text describing the expected value of the text area.
+   */
+  placeholder?: string
```
layershifter: There is no need in it as it will be passed down as HTML prop

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,123 @@
...
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props.
+   */
+  onChange?: ComponentEventHandler<TextAreaProps>
+
+  /** The value of the text area. */
+  value?: string
+
+  /** The maximum number of characters allowed in the text area. */
+  maxLength?: number
```
layershifter: https://github.com/stardust-ui/react/pull/1897/files#r322261384

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,123 @@
...
+    )
+  }
+
+  handleChange = (e: React.ChangeEvent | React.FormEvent) => {
+    const value = _.get(e, 'target.value')
+
+    _.invoke(this.props, 'onChange', e, { ...this.props, value })
+
+    this.setState({ value })
+  }
```
layershifter, line 93: nit, let's move it upper

docs/src/examples/components/TextArea/index.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react'
+
+import Types from './Types'
+
+const TextAreaExamples = () => (
+  <div>
```
layershifter:
```suggestion
  <>
```

packages/react/src/lib/accessibility/Behaviors/TextArea/textAreaBehavior.ts
```diff
@@ -0,0 +1,15 @@
+import { Accessibility } from '../../types'
+
+const textAreaBehavior: Accessibility<TextAreaBehaviorProps> = props => ({
```
layershifter, line 6: Let's add a description and tests for it

layershifter (commented):
Let's add a common test

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,123 @@
...
+}
+
+TextArea.slotClassNames = {
+  textArea: `${TextArea.className}__textarea`,
+}
+
+export default withSafeTypeForAs<typeof TextArea, TextAreaProps, 'div'>(TextArea)
```
layershifter:
```suggestion
export default withSafeTypeForAs<typeof TextArea, TextAreaProps, 'textarea'>(TextArea)
```

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,112 @@
...
+    styles,
+    unhandledProps,
+  }: RenderResultConfig<TextAreaProps>) {
+    const { disabled } = this.props
+    const { value = '' } = this.state
+
+    return (
+      <ElementType
+        value={value}
+        className={TextArea.slotClassNames.textArea}
```
miroslavstastny:
```suggestion
        className={classes.root}
```

For each slot you need to set `className` to value from `classes`.
For all slots other than `root`, you must concat that with slot class name using `cx`.

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,112 @@
...
+        styles={styles.textArea}
```
miroslavstastny: Never pass `styles` to `ElementType`. `className` handles the styling.

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,112 @@
...
+        onChange={this.handleChange}
+        disabled={disabled}
+        {...unhandledProps}
+        {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.textArea, unhandledProps)}
```
miroslavstastny:
```suggestion
        {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.root, unhandledProps)}
```

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,112 @@
...
+export interface TextAreaState {
+  value?: TextAreaProps['value']
+}
+
+class TextArea extends AutoControlledComponent<WithAsProp<TextAreaProps>, TextAreaState> {
+  static className = 'ui-textarea'
+
+  static displayName = 'TextArea'
+
+  static slotClassNames: TextAreaSlotClassNames
```
miroslavstastny: Remove this. You do not have any slots, just the `root` which is `TextArea.className`

miroslavstastny:
> Are there any props you would change/add? Such as `minHeight?`?

No `minHeight` we have `design` prop for that. Once you set correct `className` on `ElementType` this will work: `<TextArea design={{minHeight: '400px'}} />`

lucivpav:
> Let's add a common test

I added unit test as well as a behavior test

lucivpav:
Bug: now the maxLength doesn't work. Need to use `maxlength` instead of `maxLength` prop.

lucivpav:
There should be a text text to the textarea that has limited number of characters. (And used aria-describedby/labeledby)

lucivpav:
Do not merge this without adding accessibility features with Milan. (Let Milan know once this PR is approved)

docs/src/examples/components/TextArea/Types/TextAreaDisabledExample.shorthand.tsx
```diff
@@ -0,0 +1,6 @@
+import * as React from 'react'
+import { TextArea } from '@stardust-ui/react'
+
+const TextAreaDisabledExample = () => <TextArea placeholder="You can't type here" disabled />
```
layershifter:
```suggestion
const TextAreaDisabledExample = () => <TextArea disabled placeholder="You can't type here" />
```

_nit_

docs/src/examples/components/TextArea/Types/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <ComponentExample
+      title="Max length"
+      description="A text area with a limited space."
+      examplePath="components/TextArea/Types/TextAreaMaxLengthExample"
+    />
+    <ComponentExample
+      title="Disabled"
+      description="A text area that is read-only."
+      examplePath="components/TextArea/Types/TextAreaDisabledExample"
+    />
```
layershifter: `Disabled` should be under `State`

docs/src/examples/components/TextArea/Types/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <ComponentExample
+      title="Placeholder"
+      description="A text area with a placeholder."
+      examplePath="components/TextArea/Types/TextAreaExample"
+    />
+    <ComponentExample
+      title="Max length"
+      description="A text area with a limited space."
+      examplePath="components/TextArea/Types/TextAreaMaxLengthExample"
+    />
```
layershifter: `Usage` section?

docs/src/examples/components/TextArea/Types/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <ComponentExample
+      title="Disabled"
+      description="A text area that is read-only."
+      examplePath="components/TextArea/Types/TextAreaDisabledExample"
+    />
+  </ExampleSection>
```
layershifter, line 12: Let's add an example for `value` with `knobs` under `State` section

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,106 @@
...
+   *
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props.
+   */
+  onChange?: ComponentEventHandler<TextAreaProps>
+
+  /** The value of the text area. */
+  value?: string
+
+  /**
```
layershifter:
```suggestion
```

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,106 @@
...
+   * The text area becomes read-only.
+   */
```
layershifter:
```suggestion
```

packages/react/src/components/TextArea/TextArea.tsx
```diff
@@ -0,0 +1,106 @@
...
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props.
+   */
+  onChange?: ComponentEventHandler<TextAreaProps>
+
+  /** The value of the text area. */
+  value?: string
+
+  /**
+   * The text area becomes read-only.
```
layershifter:
```suggestion
   /** The text area becomes read-only. */
```

packages/react/src/lib/accessibility/Behaviors/TextArea/textAreaBehavior.ts
```diff
@@ -0,0 +1,18 @@
+import { Accessibility } from '../../types'
+/**
+ * @specification
+ * Adds attribute 'aria-disabled=true' based on the property 'disabled'. This can be overriden by providing 'aria-disabled' property directly to the component.
```
layershifter:
```suggestion
* Adds attribute 'aria-disabled=true' based on the property 'disabled'.
```

packages/react/test/specs/behaviors/textAreaBehavior-test.tsx
```diff
@@ -0,0 +1,18 @@
...
+  })
+
+  test('is enabled', () => {
+    const behavior = textAreaBehavior({ disabled: false })
+    expect(behavior.attributes['root']['aria-disabled']).toEqual(false)
+  })
+
+  test('is disabled', () => {
+    const behavior = textAreaBehavior({ disabled: true })
+    expect(behavior.attributes['root']['aria-disabled']).toEqual(true)
```
layershifter: I think that this file is redundant as should be tested by comments:
https://github.com/stardust-ui/react/blob/master/.github/test-a-feature.md#behavior-tets

https://github.com/stardust-ui/react/pull/1897/files#r322706395

packages/react/test/specs/components/TextArea/TextArea-test.tsx
```diff
@@ -0,0 +1,65 @@
...
+const testValue = 'test value'
+
+describe('TextArea', () => {
+  describe('conformance', () => {
+    isConformant(TextArea, {
+      eventTargets: {
+        onChange: 'textarea',
+        onKeyDown: 'textarea',
+        onKeyPress: 'textarea',
+        onKeyUp: 'textarea',
```
layershifter: Not sure, but `textarea` is `root` so `isConformant(TextArea)` should work without any config

packages/react/test/specs/components/TextArea/TextArea-test.tsx
```diff
@@ -0,0 +1,65 @@
...
+  textAreaComp.find('textarea').getDOMNode() as HTMLTextAreaElement
+
+const setUserTextAreaValue = (textAreaComp: ReactWrapper, value: string) => {
+  textAreaComp.find('textarea').simulate('change', { target: { value } })
+}
+
+const testValue = 'test value'
+
+describe('TextArea', () => {
+  describe('conformance', () => {
```
layershifter:
```suggestion
  describe('conformance', () => {
```

`isConformant` creates own `describe`

packages/react/test/specs/components/TextArea/TextArea-test.tsx
```diff
@@ -0,0 +1,65 @@
...
+      setUserTextAreaValue(textAreaComp, testValue)
+
+      expect(domNode.value).toEqual(testValue)
+    })
+
+    it('cannot set input value from user when the value prop is already set (controlled mode)', () => {
+      const controlledTextAreaValue = 'controlled input value'
+      const textAreaComp = mount(<TextArea value={controlledTextAreaValue} />)
+      const domNode = getTextAreaDomNode(textAreaComp)
+      setUserTextAreaValue(textAreaComp, testValue)
```
layershifter:
```suggestion
      wrapper.simulate('change', { target: { value } })
```

Let's avoid useless helper function

packages/react/test/specs/components/TextArea/TextArea-test.tsx
```diff
@@ -0,0 +1,65 @@
...
+      const textAreaComp = mount(<TextArea />)
+      const domNode = getTextAreaDomNode(textAreaComp)
+      setUserTextAreaValue(textAreaComp, testValue)
+
+      expect(domNode.value).toEqual(testValue)
+    })
+
+    it('cannot set input value from user when the value prop is already set (controlled mode)', () => {
+      const controlledTextAreaValue = 'controlled input value'
+      const textAreaComp = mount(<TextArea value={controlledTextAreaValue} />)
```
layershifter:
```suggestion
      const wrapper= mount(<TextArea value={controlledTextAreaValue} />)
```

It's common to call it `wrapper`
https://airbnb.io/enzyme/docs/api/ReactWrapper/mount.html

packages/react/test/specs/components/TextArea/TextArea-test.tsx
```diff
@@ -0,0 +1,65 @@
...
+    })
+
+    it('cannot set TextArea value from user when the value prop is already set (controlled mode)', () => {
+      const controlledTextAreaValue = 'controlled TextArea value'
+      const textAreaComp = mount(<TextArea value={controlledTextAreaValue} />)
+      const domNode = getTextAreaDomNode(textAreaComp)
+      setUserTextAreaValue(textAreaComp, testValue)
+
+      expect(domNode.value).toEqual(controlledTextAreaValue)
+    })
```
layershifter, line 31: Duplicate of previous `describe`?

packages/react/test/specs/components/TextArea/TextArea-test.tsx
```diff
@@ -0,0 +1,65 @@
...
+      eventTargets: {
+        onChange: 'textarea',
+        onKeyDown: 'textarea',
+        onKeyPress: 'textarea',
+        onKeyUp: 'textarea',
+      },
+    })
+  })
+
+  describe('auto-controlled', () => {
```
layershifter: General thing, in SUIR it was common to structure tests in this way and the same thing is covered in our docs:
https://github.com/stardust-ui/react/blob/master/.github/test-a-feature.md#writing-tests

```tsx
describe('TextArea', () => {
  isConformant(TextArea)

  describe('defaultValue', () => {
     test('sets "defaultValue" as initial "value"', () => {
       const wrapper = mount(<TextArea defaultValue="foo" />)
       
       expect(wrapper.prop('value')).toBe('foo')
     })
  })

  describe('onChange', () => {
    test('is called with (e, data)', () => {
      const onChange = jest.fn()
      const wrapper = mount(<TextArea onChange={onChange} />)

      wrapper.simulate('change', { target: { value: 'foo' })
      expect(onChange).toBeCalledTimes(1)
      expect(onChange).toBeCalledWith(expect.objectContaining({ type: 'change' }), expect.objectContaining({ value: 'foo' })
    })
  })
})
```

docs/src/components/ComponentPlayground/ComponentPlayground.tsx
layershifter: Can be reenabled now?

kolaps33 (approved):

kolaps33 (approved):

docs/src/examples/components/TextArea/Types/index.tsx
lucivpav: You mean that I would use `useStringKnob()` for the value? Or other knobs for other props too?

packages/react/test/specs/components/TextArea/TextArea-test.tsx
lucivpav: I don't know why, but if I inline it, the test will fail

docs/src/components/ComponentPlayground/ComponentPlayground.tsx
lucivpav: Since there is no longer `placeholder` prop on (nor `maxLength` prop), it can be enabled. But I am not sure about its value

lucivpav:
@layershifter one more review please :)

docs/src/examples/components/TextArea/Usage/TextAreaMaxLengthExample.shorthand.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+import { TextArea } from '@stardust-ui/react'
+
+const maxLength = 5
+
+const TextAreaMaxLengthExample = () => (
+  <>
+    <TextArea
+      placeholder={`No more than ${maxLength} letters can be here...`}
```
layershifter:
```suggestion
      placeholder='No more than 5 letters can be here...'
```

Let's inline it

docs/src/examples/components/TextArea/Usage/TextAreaMaxLengthExample.shorthand.tsx
```diff
@@ -0,0 +1,15 @@
...
+      maxLength={maxLength}
```
layershifter:
```suggestion
      maxLength={5}
```

docs/src/examples/components/TextArea/Usage/index.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react'
+import ComponentExample from 'docs/src/components/ComponentDoc/ComponentExample'
+import ExampleSection from 'docs/src/components/ComponentDoc/ExampleSection'
+
+const Usage = () => (
+  <ExampleSection title="">
```
layershifter, line 6:
```suggestion
  <ExampleSection title="Usage">
```

packages/react/test/specs/components/TextArea/TextArea-test.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { isConformant } from 'test/specs/commonTests'
+
+const testValue = 'test value'
+
+describe('TextArea', () => {
+  isConformant(TextArea)
+
+  describe('defaultValue', () => {
+    test('sets "defaultValue" as initial "value"', () => {
+      const wrapper = mount(<TextArea defaultValue={testValue} />)
```
layershifter:
```suggestion
      const wrapper = mount(<TextArea defaultValue='foo' />)
```

Can we inline it? It's more obvious in the most cases

packages/react/test/specs/components/TextArea/TextArea-test.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+      expect(wrapper.find('textarea').prop('value')).toBe(testValue)
+    })
+  })
+
+  describe('auto-controlled', () => {
```
layershifter: Note sure that it's useful at all now as it tests the same thing as `AutoControlledComponent` does...

docs/src/examples/components/TextArea/Usage/index.tsx
lucivpav: Actually, we don't specify `Usage`. As all the examples under Usage naturally belong to `Usage`.

layershifter (commented):
Left minor comments, LGTM 👍 

Let's add some dumb styles and variables (for `margin: 0`, `resize: vertical`) to create boilerplate under Teams theme. And add an entry to `CHANGELOG.md` before merging 😼

packages/react/test/specs/components/TextArea/TextArea-test.tsx
lucivpav: But it's also more prone to errors, if you have to specify the same text at several places, you can make a typo. In the worst case, a test would pass while component is broken.

packages/react/test/specs/components/TextArea/TextArea-test.tsx
lucivpav: Does it mean the test code of `Input` is also irrelevant? https://github.com/stardust-ui/react/blob/master/packages/react/test/specs/components/Input/Input-test.tsx#L72

layershifter: Yep, it's strange

packages/react/test/specs/components/TextArea/TextArea-test.tsx
lucivpav: Ok, I replaced all the occurences of `testValue` by `foo`.

layershifter: At the same time it makes code less readable.

> But it's also more prone to errors, if you have to specify the same text at several places, you can make a typo.

It _should_ be the same only in single test, it's good to use [faker](https://www.npmjs.com/package/faker), actually.

docs/src/examples/components/TextArea/Usage/index.tsx
layershifter:
```suggestion
  <ExampleSection>
```

So it should be or even

```suggestion
  <UsageSection>
```

Otherwise it extremely confusing

docs/src/examples/components/TextArea/Usage/TextAreaMaxLengthExample.shorthand.tsx
```diff
@@ -0,0 +1,10 @@
+import * as React from 'react'
+import { TextArea } from '@stardust-ui/react'
+
+const TextAreaMaxLengthExample = () => (
+  <>
```
layershifter: Oops, `Fragment` is useless

```suggestion
```

lucivpav: Nice catch

packages/react/test/specs/components/TextArea/TextArea-test.tsx
lucivpav: I am using `faker` now. Hopefully, the code is still readable 🤔

packages/react/test/specs/components/TextArea/TextArea-test.tsx
lucivpav: Ok, I removed it from the other components, #1917

docs/src/examples/components/TextArea/Usage/index.tsx
lucivpav: I agree its confusing 👍

lucivpav: Let me create and use `UsageSection` in a seperate PR

docs/src/examples/components/TextArea/Usage/TextAreaMaxLengthExample.shorthand.tsx
```diff
@@ -0,0 +1,8 @@
+import * as React from 'react'
+import { TextArea } from '@stardust-ui/react'
+
+const TextAreaMaxLengthExample = () => (
+  <TextArea placeholder={`No more than 5 letters can be here...`} maxLength={5} />
```
layershifter:
```suggestion
  <TextArea placeholder='No more than 5 letters can be here...' maxLength={5} />
```

_nit_, should be handled in future by ESLint

layershifter (approved):

PR #1898, opened 2019-09-06
[WIP] fix(PortalInner): wrap children with ProviderBox

mnajdova:
None of the styles of the `ProviderBox` are applied when the `PortalInner` component is rendered. Simple fix introduced for wrapping the children of the `PortalInner` component with `ProviderBox` (ensured that it works with `Popup`, `Tooltip`  and `Dialog` components).

TODO:
- [ ] add screenshot tests for `Popup`, `Dialog` and `Tooltip`
- [x] fix Popup tests

### DIFF
--- a/docs/src/examples/components/Popup/Types/PopupExample.shorthand.tsx
+++ b/docs/src/examples/components/Popup/Types/PopupExample.shorthand.tsx
@@ -1,6 +1,25 @@
-import * as React from 'react'
-import { Button, Popup } from '@stardust-ui/react'
+import React from 'react'
+import { Button, Popup, Dialog } from '@stardust-ui/react'
 
-const PopupExample = () => <Popup trigger={<Button icon="more" />} content="Hello from popup!" />
+const PopupExample = () => (
+  <>
+    <Popup
+      trigger={<Button icon="more" />}
+      content={{
+        content:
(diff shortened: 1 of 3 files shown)


docs/src/examples/components/Popup/Types/PopupExample.shorthand.tsx
```diff
@@ -1,6 +1,25 @@
-import * as React from 'react'
-import { Button, Popup } from '@stardust-ui/react'
+import React from 'react'
+import { Button, Popup, Dialog } from '@stardust-ui/react'
 
-const PopupExample = () => <Popup trigger={<Button icon="more" />} content="Hello from popup!" />
+const PopupExample = () => (
+  <>
```
mnajdova: Disregard this example for now, will add separate for all components which are using the PortalInner

layershifter:
Can be keep `ProviderBox` in `Popup`, `Dialog`, etc? I think that `Portal` is too general for this and should be potentially moved to `react-bindings`.

mnajdova:
Closing this for now, until we decide what really should go in the ProviderBox, and whether all portal components created from Stardust should by default have the ProviderBox inside.. It's not a blocker as currently can be achieved by wrapping this components with ProviderBox on client's side.

PR #1900, opened 2019-09-09
feat(prototypes): Allow attaching notification to a target ref in CopyToClipboard prototype

jurokapsiar:
Notification button can be inside of a submenu - after clicking on it, the submenu closes and developer might want to attach the notification to the parent menu item.


docs/src/prototypes/CopyToClipboard/NotificationProvider.tsx
```diff
@@ -28,9 +35,21 @@ export const NotificationProvider: React.FC = props => {
 
   return (
     <>
-      <Portal open={!!notification}>
-        <Notification>{notification}</Notification>
-      </Portal>
+      {!!targetRef ? (
+        !!notification && (
+          <Tooltip
+            content={notification}
```
lucivpav:
```suggestion
            content={{content: notification, variables: {primary: true}}}
```

docs/src/prototypes/CopyToClipboard/NotificationProvider.tsx
```diff
@@ -28,9 +35,21 @@ export const NotificationProvider: React.FC = props => {
...
+            pointing={true}
+            open={!!notification}
+            target={targetRef.current}
+            variables={{ primary: true }}
```
lucivpav:
```suggestion
```

lucivpav: The `primary` prop should be set on `content`, otherwise the appropriate style is not applied:
image

docs/src/prototypes/CopyToClipboard/NotificationProvider.tsx
```diff
@@ -28,9 +35,21 @@ export const NotificationProvider: React.FC = props => {
 
   return (
     <>
-      <Portal open={!!notification}>
-        <Notification>{notification}</Notification>
-      </Portal>
+      {!!targetRef ? (
+        !!notification && (
+          <Tooltip
```
lucivpav: The following props are missing, yet they are used in `CopyToClipboard.tsx`'s `Tooltip`: align, pointing, position.

docs/src/prototypes/CopyToClipboard/index.tsx
```diff
@@ -27,27 +27,37 @@ const CopyToClipboardPrototype: React.FC<CopyToClipboardPrototypeProps> = props
 }
 
 const CopyToClipboardInMenu: React.FC = props => {
```
lucivpav, line 30: Let's keep the original example, and add a new example demonstrating the `targetRef` prop.

packages/react/src/components/Tooltip/Tooltip.tsx
```diff
@@ -161,9 +161,9 @@ export default class Tooltip extends AutoControlledComponent<TooltipProps, Toolt
     const { mountNode, children, trigger } = this.props
     const tooltipContent = this.renderTooltipContent(classes.content, rtl, accessibility)
 
-    const triggerElement = React.Children.only(
-      childrenExist(children) ? children : trigger,
-    ) as React.ReactElement<any>
+    const triggerElement =
+      (children || trigger) &&
+      (React.Children.only(childrenExist(children) ? children : trigger) as React.ReactElement<any>)
```
layershifter: It's hard to follow what happens there, can we simplify?

```tsx
    const triggerNode = childrenExist(children) ? children : trigger
    const triggerElement = triggerNode && React.Children.only(triggerNode) as React.ReactElement
    const triggerProps = this.getTriggerProps(triggerElement)
```

docs/src/prototypes/CopyToClipboard/index.tsx
```diff
@@ -27,27 +27,37 @@ const CopyToClipboardPrototype: React.FC<CopyToClipboardPrototypeProps> = props
...
-      ],
-    },
+    render =>
+      render(item, (Component, props) => (
+        <Ref innerRef={targetRef}>
+          <Component {...props} />
+        </Ref>
+      )),
   ]
   return <Menu items={items} />
```
layershifter, line 77: Can keep existing example and add one more instead?

docs/src/prototypes/CopyToClipboard/NotificationProvider.tsx
```diff
@@ -28,9 +35,21 @@ export const NotificationProvider: React.FC = props => {
...
+            pointing={true}
+            open={!!notification}
+            target={targetRef.current}
+            variables={{ primary: true }}
+          />
+        )
+      ) : (
+        <Portal open={!!notification}>
+          (<Notification>{notification}</Notification>)
+        </Portal>
```
layershifter: Condition there is very unclear.

Why we show `Notification` or `Tooltip` based on `targetRef`? Should we introduce an additional `position` or prop?

docs/src/prototypes/CopyToClipboard/NotificationProvider.tsx
```diff
@@ -28,9 +35,21 @@ export const NotificationProvider: React.FC = props => {
...
   return (
     <>
-      <Portal open={!!notification}>
-        <Notification>{notification}</Notification>
-      </Portal>
+      {!!targetRef ? (
+        !!notification && (
+          <Tooltip
+            content={notification}
+            pointing={true}
```
layershifter: It doesn't respect `pointing` value from the parent component

docs/src/prototypes/CopyToClipboard/NotificationProvider.tsx
```diff
@@ -28,9 +35,21 @@ export const NotificationProvider: React.FC = props => {
...
+            open={!!notification}
+            target={targetRef.current}
```
layershifter: This is dangerous, let's use `target`:

```suggestion
            target={targetRef}
```

docs/src/prototypes/CopyToClipboard/index.tsx
```diff
@@ -27,27 +27,37 @@ const CopyToClipboardPrototype: React.FC<CopyToClipboardPrototypeProps> = props
 }
 
 const CopyToClipboardInMenu: React.FC = props => {
+  const targetRef = React.useRef<HTMLElement>(null)
```
layershifter: https://github.com/stardust-ui/react/pull/1900/files#r322141672

```suggestion
  const [target, setTarget] = React.useState<HTMLElement>(null)
```

docs/src/prototypes/CopyToClipboard/index.tsx
```diff
@@ -27,27 +27,37 @@ const CopyToClipboardPrototype: React.FC<CopyToClipboardPrototypeProps> = props
...
-                value={'Julius Caesar'}
-                trigger={<Component {...props} />}
-              />
-            )
-          }),
-      ],
-    },
+    render =>
+      render(item, (Component, props) => (
+        <Ref innerRef={targetRef}>
```
layershifter: https://github.com/stardust-ui/react/pull/1900/files#r322141672

```suggestion
        <Ref innerRef={setTarget}>
```

docs/src/prototypes/CopyToClipboard/index.tsx
jurokapsiar: why? we already have the example for not attached, no need to duplicate

docs/src/prototypes/CopyToClipboard/NotificationProvider.tsx
jurokapsiar: according to design, this should never be pointing, so I will always set it to false

docs/src/prototypes/CopyToClipboard/NotificationProvider.tsx
jurokapsiar: wouldn't that be inconsistent with Popup component where we have
```
target?: HTMLElement
```

layershifter: Oops, wrong suggestion. It should be the same as in `Popup` and be:

```tsx
target={target}
```

So we will pass `target` as `HTMLElement` instead of refs.

***

> Keep in mind that useRef doesn’t notify you when its content changes. Mutating the .current property doesn’t cause a re-render. If you want to run some code when React attaches or detaches a ref to a DOM node, you may want to use a callback ref instead.

https://reactjs.org/docs/hooks-reference.html#useref

docs/src/prototypes/CopyToClipboard/NotificationProvider.tsx
jurokapsiar: the notification should be black according to the spec

docs/src/prototypes/CopyToClipboard/NotificationProvider.tsx
jurokapsiar: changed the API - now there is a separate prop for tooltip (shorthand) and a separate for ataching the notification (TooltipProps or boolean, but we can change it to just subset of tooltip props)

docs/src/prototypes/CopyToClipboard/index.tsx
```diff
@@ -27,27 +26,51 @@ const CopyToClipboardPrototype: React.FC<CopyToClipboardPrototypeProps> = props
...
+    content: 'Edit',
+    menu: [
+      'Open File...',
+      'Save File...',
+      render =>
+        render('Copy text', (Component, props) => {
+          return (
+            <CopyToClipboard
+              attach={{ target: targetRef.current }}
+              value={'Julius Caesar'}
```
layershifter:
```suggestion
              value='Julius Caesar'
```

_nit_

docs/src/prototypes/CopyToClipboard/index.tsx
layershifter: Oops, was not really applied :(

docs/src/prototypes/CopyToClipboard/index.tsx
```diff
@@ -27,27 +26,51 @@ const CopyToClipboardPrototype: React.FC<CopyToClipboardPrototypeProps> = props
...
+    key: 'edit',
+    content: 'Edit',
+    menu: [
+      'Open File...',
+      'Save File...',
+      render =>
+        render('Copy text', (Component, props) => {
+          return (
+            <CopyToClipboard
+              attach={{ target: targetRef.current }}
```
layershifter:
```suggestion
              attach={{ target }}
```

https://github.com/stardust-ui/react/pull/1900/files#r322141672

docs/src/prototypes/CopyToClipboard/NotificationProvider.tsx
```diff
@@ -1,24 +1,34 @@
...
-  children?: React.ReactNode
+  content: React.ReactNode
+  target?: HTMLElement
+  trigger?: JSX.Element
 }
 
-type NotificationContextValue = (value: React.ReactNode, timeout: number) => void
+type NotificationContextValue = (
+  content: React.ReactNode,
+  targe: HTMLElement | null,
```
pompomon: Shouldn't it be "target" instead of "targe"?

docs/src/prototypes/CopyToClipboard/CopyToClipboard.tsx
```diff
@@ -1,41 +1,25 @@
...
 import * as React from 'react'
 
-import { NotificationContext } from './NotificationProvider'
+import { Notification, NotificationContext } from './NotificationProvider'
 
 export type CopyToClipboardProps = {
-  attached?: boolean
-  pointing?: boolean
+  tooltip?: ShorthandValue<TooltipProps>
+  attach?: boolean
```
layershifter:
```suggestion
  attached?: boolean
```

Should it be `attached` again?

docs/src/prototypes/CopyToClipboard/NotificationProvider.tsx
```diff
@@ -28,23 +38,38 @@ export const NotificationProvider: React.FC = props => {
...
+
+    if (target || trigger) {
+      return Tooltip.create({ ...tooltipProps, offset: '0 10' })
+    }
+
     return (
-      <div className={classes.root}>
-        <div className={classes.overlay}>
-          <div className={classes.content}>{children}</div>
+      <Portal open={true}>
```
layershifter:
```suggestion
      <Portal open>
```

_nit_

docs/src/prototypes/CopyToClipboard/themeOverrides.ts
```diff
@@ -78,44 +52,16 @@ const themeOverrides: ThemeOverrides = {
...
-
-        ...(v.primary && {
-          backgroundImage: pointerSvgUrl(v.primaryContentBackgroundColor),
-        }),
-      }),
-    },
   },
   componentVariables: {
     Notification: (siteVariables): NotificationVariables => ({
       contentBackgroundColor: siteVariables.colorScheme.default.foreground,
```
layershifter: All these variables can be removed as they are not used.

layershifter (approved):

PR #1901, opened 2019-09-09
fix(Accessibility): fix order of applying unhandled props and key handlers

### LINKED ISSUE #1899: Embed does not apply key handlers correctly
### Steps
Use popup with a trigger that has key handlers coming from a behavior - for example `<Embed />`.

### Expected Result
User should be able to use keyboard keys generated from behaviors

### Actual Result
onKeyDown generated from behavior is ignored because the unhandled props are applied after key hadlers.

I can notice that in Embed, Button, HierarchicalTreeItem
 
### Version
0.38.0

### Testcase
 https://codesandbox.io/s/stardust-ui-example-tdgx9

jurokapsiar:
Fixes #1899 

Key handlers need to be applied after unhandled props so that onKeyDown is not overwritten but correctly respects both the key handlers from behaviors as well as onKeyDown passed by the consumer or by a wrapping component (popup)

### DIFF
--- a/packages/react/src/components/Embed/Embed.tsx
+++ b/packages/react/src/components/Embed/Embed.tsx
@@ -131,8 +131,8 @@ class Embed extends AutoControlledComponent<WithAsProp<EmbedProps>, EmbedState> 
         className={classes.root}
         onClick={this.handleClick}
         {...accessibility.attributes.root}
-        {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.root, unhandledProps)}
         {...unhandledProps}
+        {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.root, unhandledProps)}
       >
         {active ? (
           <>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixes
 - Fix `Menu` and `Submenu` to use correct indicator icon and have correct width behavior [redlines] @bcalvery ([#1831](https://github.com/stardust-ui/react/pull/1831))
+- Fix order of applying unhandled props and key handlers @jurokapsiar ([#1901](https://github.com/stardust-ui/react/pull/1901)) 
 
 <!--------------------------------[ v0.38.0 ]------------------------------- -->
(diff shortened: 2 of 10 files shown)


layershifter (commented):
1. Let's add an entry to CHANGELOG
2. Let's add a proper UT, may be it will catch additional cases (under `isConformant`):

```tsx
    test(`handles "onKeyDown" overrides`, () => {
      const actionHandler = jest.fn()
      const eventHandler = jest.fn()

      const actionBehavior: Accessibility = () => ({
        keyActions: {
          root: {
            mockAction: {
              keyCombinations: [{ keyCode: keyboardKey.Enter }],
            },
          },
        },
      })
      const wrapperProps = {
        ...requiredProps,
        accessibility: actionBehavior,
        onKeyDown: eventHandler,
      }
      const wrapper = mount(<Component {...wrapperProps} />)

      ;(wrapper.instance() as UIComponent<any, any>).actionHandlers.mockAction = actionHandler
      // Force render component to apply updated key handlers
      wrapper.setProps({})

      getEventTargetComponent(wrapper, 'onKeyDown').simulate(
        'keydown',
        { keyCode: keyboardKey.Enter },
      )

      expect(actionHandler).toBeCalledTimes(1)
      expect(eventHandler).toBeCalledTimes(1)
    })
```

layershifter (approved):
Please fix CI before merge 🚀

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -131,8 +131,8 @@ class Embed extends AutoControlledComponent<WithAsProp<EmbedProps>, EmbedState>
         className={classes.root}
         onClick={this.handleClick}
         {...accessibility.attributes.root}
-        {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.root, unhandledProps)}
         {...unhandledProps}
+        {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.root, unhandledProps)}
```
mnajdova, line 135: This is correct `applyAccessibilityKeyHandlers` should always be spread after the `unhandledProps`, as it will invoke the key events provided by the behavior.

jurokapsiar: yes, there's now even a test verifying that for all compoennts (that have the actions defined)

PR #1903, opened 2019-09-09
fix(Chat): fix handling of `onMouseEnter`

layershifter:
### 💥 Issue

Previously `onMouseEnter` was not handled properly, if you will pass custom `onMouseEnter` the position update will be never called.

### DIFF
--- a/packages/react/src/components/Chat/ChatMessage.tsx
+++ b/packages/react/src/components/Chat/ChatMessage.tsx
@@ -90,6 +90,13 @@ export interface ChatMessageProps
    */
   onFocus?: ComponentEventHandler<ChatMessageProps>
 
+  /**
+   * Called after user will enter by mouse.
+   * @param {SyntheticEvent} event - React's original SyntheticEvent.
+   * @param {object} data - All props.
+   */
+  onMouseEnter?: ComponentEventHandler<ChatMessageProps>
+
   /** Reaction group applied to the message. */
   reactionGroup?: ShorthandValue<ReactionGroupProps> | ShorthandCollection<ReactionProps>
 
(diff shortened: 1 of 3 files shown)


lucivpav:
But how could you previously pass `onMouseEnter`, when `ChatMessage` didn't have such prop? Shouldn't this be a feature PR, that adds such a prop?

packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -90,6 +90,13 @@ export interface ChatMessageProps
    */
   onFocus?: ComponentEventHandler<ChatMessageProps>
 
+  /**
+   * Called after user will enter by mouse.
```
lucivpav:
```suggestion
   * Called after user enters by mouse.
```
nit

packages/react/test/specs/components/Chat/ChatMessage-test.tsx
```diff
@@ -20,4 +23,14 @@ describe('ChatMessage', () => {
       focusZoneDefinition: (chatMessageBehavior as AccessibilityDefinition).focusZone,
     })
   })
+
```
lucivpav, line 26: How about another test, that will check that a custom `onMouseEnter` function gets called?

layershifter: It's already tested by `isConformant` test, see `handles events transparently`.

layershifter:
> But how could you previously pass `onMouseEnter`, when `ChatMessage` didn't have such prop?

It was overridden as an unhandled prop:
image

PR #1904, opened 2019-09-09
feat(List): behavior, example and prototype for navigable list

jurokapsiar:
After discussions with screen reader vendors, the behavior should be:
- right/down arrow move focus to the next item
- left/up arrow move focus to the previous item
- use menu and menuitem roles

Prototype shows how to attach context menu to the list items. This PR includes necessary fixes to make it work correctly.


packages/react/src/components/MenuButton/MenuButton.tsx
```diff
@@ -255,6 +255,7 @@ export default class MenuButton extends AutoControlledComponent<MenuButtonProps,
     })
 
     const overrideProps = {
+      accessibility: () => accessibility,
```
jurokapsiar, line 268: this was necessary so that the trigger would be focuses after closing the menu also for the contextMenu scenario

docs/src/examples/components/List/Types/ListExampleNavigbale.shorthand.tsx
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react'
```
miroslavstastny: This file has misspelled name -> does not show up in docsite docs/src/examples/components/List/Types/ListExampleNavig**ba**le.shorthand.tsx

packages/react/src/components/List/List.tsx
```diff
@@ -83,6 +85,7 @@ class List extends AutoControlledComponent<WithAsProp<ListProps>, ListState> {
     debug: PropTypes.bool,
     items: customPropTypes.collectionShorthand,
     selectable: PropTypes.bool,
+    navigable: PropTypes.bool,
```
miroslavstastny: I don't think it is valid to create `<List selectable navigable ... />`
I would go with someting like this:
```jsx
selectable: customPropTypes.every([customPropTypes.disallow(['navigable']), PropTypes.bool]),
navigable: customPropTypes.every([customPropTypes.disallow(['selectable']), PropTypes.bool]),
```

docs/src/examples/components/List/Types/ListExampleNavigbale.shorthand.tsx
jurokapsiar: good catch, thanks!

miroslavstastny (approved):

PR #1907, opened 2019-09-10
fix(Theme): Deep merge variables

miroslavstastny:
Do deep merge of site variables and component variables.

### Motivation
Let's have two themes (parent, child) merged.
To resolve styles, follow steps occur
1. `child.siteVariables` are merged on top of `parent.siteVariables` to compute `siteVariables`
2. computed `siteVariables` are used to compute `parent.componentVariables` and `child.componentVariables`, `child.componentVariables` are merged on top of `parent.componentVariables`
3. computed `componentVariables` are used to compute `parent.styles` and `child.styles`

That means that if child variables remove a variable from parent and parent styles use that variable, an exception will be thrown.


layershifter:
Please measure it before merging 🤔

packages/react/src/themes/teams-high-contrast/components/Attachment/attachmentVariables.ts
```diff
@@ -6,7 +6,7 @@ export default (siteVariables: any): Partial<AttachmentVariables> => ({
   backgroundColorHover: siteVariables.accessibleYellow,
   textColor: siteVariables.colors.white,
   textColorHover: siteVariables.colors.black,
-  boxShadow: undefined,
+  boxShadow: null,
```
levithomason: Try using `unset` CSS value here instead of the `null` concept.  Hopefully, we can stick with built-in CSS functionality here.

levithomason: Looking at the rest of the code, specifically variables, if the `null` concept works everywhere (styles/variables) for "removing" values or "unsetting" them, then so be it 👍 

Let's wait to hear some evidence from users.

packages/react/test/specs/lib/deepmerge-test.ts
```diff
@@ -0,0 +1,118 @@
...
+    const merged = deepmerge({ color: 'black' }, { color: undefined })
+    expect(merged).toStrictEqual({ color: undefined })
+  })
+
+  test('null overwrites previous value', () => {
+    const merged = deepmerge({ color: 'black' }, { color: null })
+    expect(merged).toStrictEqual({ color: null })
+  })
+
+  test('undefined gets overwritten  by next value', () => {
```
layershifter:
```suggestion
  test('undefined gets overwritten by next value', () => {
```

nit

packages/react/test/specs/lib/deepmerge-test.ts
```diff
@@ -0,0 +1,118 @@
...
+    const merged = deepmerge({ color: undefined }, { color: 'black' })
+    expect(merged).toStrictEqual({ color: 'black' })
+  })
+
```
layershifter, line 33:
```suggestion
  test('DOES SOMETHING WITH ARRAYS', () => {
    const merged = deepmerge({ color: 'black' }, { color: ['green', 'red'] })
    expect(merged).toStrictEqual({ color: ??? })
  })
```

Can you please a test for this? As we have fallback values that are arrays

layershifter: https://github.com/stardust-ui/react/pull/1907/files#diff-72b4c3a8ba5c9f962bf97db03ae032abR110

Second should always win as I understand. Just want to keep it clear in tests

miroslavstastny:
> Please measure it before merging 🤔

master:

| --- |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |
| ChatWithPopover.perf.tsx | 561.7 | +3.84% | 955.3 | **955.3** | 614.07 | +27.37% | 571.12 | **571.12** |
| Chat.perf.tsx | 540.91 | **540.91** | 956.14 | +0.09% | 482.11 | **482.11** | 1164 | +103.81% |

deep merge branch:

| --- |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |
| Chat.perf.tsx | 511.91 | **511.91** | 969.01 | **969.01** | 482.27 | **482.27** | 1122.21 | +74.7% |
| ChatWithPopover.perf.tsx | 558.12 | +9.03% | 996.02 | +2.79% | 609.04 | +26.29% | 642.37 | **642.37** |

My interpretation of the results: there is no measurable perf impact.

packages/react/src/themes/teams-high-contrast/components/Attachment/attachmentVariables.ts
miroslavstastny: switched back to `undefined`

packages/react/test/specs/lib/deepmerge-test.ts
miroslavstastny: If one of the values is not an object -> last one wins. Tests added, deep merge fixed.

levithomason (approved):
👍

PR #1908, opened 2019-09-10
refactor(Theme): move base theme styles to teams theme

mnajdova:
We decided to remove the base theme from Stardust, as it was created in an early stage, where we didn't have more use-cases to test it against. All theme definitions (styles, variables, siteVariables etc( that were currently in the base theme, are now moved to the teams theme.

The plan is to start to work on the base theme, with adding some more designs/applications where we will test whether the definitions added in the base theme are sufficient for achieving all different cases.

For Teams theme, this is not breaking change, as all styles from base theme are now included directly in Teams theme.

Introduced PR https://github.com/stardust-ui/react/pull/1941 for explicitly adding typings for teams theme as `ThemePrepared`, as we no longer use mergeThemes when exporting it.


docs/src/components/ColorSchemes.tsx
```diff
@@ -3,7 +3,7 @@ import * as _ from 'lodash'
 import {
   createComponent,
   ComponentSlotStylesInput,
-  ThemePrepared,
+  ThemeInput,
```
mnajdova: TODO: review this once again...

levithomason (commented):
image
❤️

packages/react/src/themes/font-awesome/components/Icon/iconStyles.ts
```diff
@@ -1,19 +0,0 @@
-import { ComponentSlotStylesInput, ICSSInJSStyle, ThemeIconSpec } from '../../../types'
-import { IconProps } from '../../../../components/Icon/Icon'
-import { IconVariables } from '../../../teams/components/Icon/iconVariables'
-import { emptyIcon } from '../../../base/components/Icon/iconNames'
-
-const iconStyles: ComponentSlotStylesInput<IconProps, IconVariables> = {
```
levithomason: Let's not delete font-awesome theme just yet.  Also, let's not port FA theme info to Teams theme.  Keep them separate.

levithomason (approved):
👍

PR #1909, opened 2019-09-10
fix(Embed): get rid of white flash

### LINKED ISSUE #1586: Embed video has a white flash when toggle active state

lucivpav:
Now, when you activate a `video` or `iframe` `Embed`, no white flash appears. Instead, the placeholder image is used.

Fixes #1586


packages/react/src/components/Embed/Embed.tsx
```diff
@@ -142,25 +151,27 @@ class Embed extends AutoControlledComponent<WithAsProp<EmbedProps>, EmbedState>
...
               },
             })}
-            {Box.create(iframe, { defaultProps: { as: 'iframe', styles: styles.iframe } })}
+            {Box.create(iframe, {
+              defaultProps: {
+                as: 'iframe',
+                styles: styles.iframe,
+                style: { visibility: iframeLoaded ? 'visible' : 'hidden' },
+                onLoad: () => {
+                  this.setState({ iframeLoaded: true })
```
layershifter: Should be handled in `overrideProps` otherwise `onLoad` passed by user will override our handler.

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -142,25 +151,27 @@ class Embed extends AutoControlledComponent<WithAsProp<EmbedProps>, EmbedState>
...
                   height: variables.height,
                 },
               },
             })}
-            {Box.create(iframe, { defaultProps: { as: 'iframe', styles: styles.iframe } })}
+            {Box.create(iframe, {
+              defaultProps: {
+                as: 'iframe',
+                styles: styles.iframe,
+                style: { visibility: iframeLoaded ? 'visible' : 'hidden' },
```
layershifter: Move these styles into theme please.

lucivpav: How do I access state in `EmbedStyles`?

layershifter: It will be available under props, use there a mix of types (actually it's already there):

image

packages/react/src/components/Embed/Embed.tsx
layershifter: 1. Original handler is still not called.
2. Please move it to class method:

```tsx
handleFrameOverrides = (predefinedProps) => ({
  onLoad: (e: EVENT_TYPE) => {
     _.invoke(predefinedProps, 'onLoaded', e)
     this.setState({ iframeLoaded: true })
  }
}}
```

packages/react/src/components/Embed/Embed.tsx
```diff
@@ -134,6 +142,7 @@ class Embed extends AutoControlledComponent<WithAsProp<EmbedProps>, EmbedState>
         {...unhandledProps}
         {...applyAccessibilityKeyHandlers(accessibility.keyHandlers.root, unhandledProps)}
       >
+        {iframe && active && !iframeLoaded && placeholderImage}
```
layershifter: What does this condition? Is there way to simplify rendering there?
There is too many conditions to be honest.

lucivpav: I have extracted it into a function, please check again

layershifter (approved):

PR #1910, opened 2019-09-10
chore(downshift): update to 3.2.14

### LINKED ISSUE #1818: Dropdown does not respect target in Provider context
### Steps
open https://codesandbox.io/s/stardust-ui-example-1phxh
Try to interact with the dropdown

### Expected Result
Aria live messages are rendered in the correct document
User can open the single selection non-search dropdown when clicking on it

### Actual Result
One of the aria live regions are rendered in the main document

### Version
0.36.1

silviuaavram:
Do not merge until approval.

Closes https://github.com/stardust-ui/react/issues/1818

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -11,7 +11,7 @@
     "@stardust-ui/react-component-ref": "^0.38.0",
     "@stardust-ui/react-proptypes": "^0.38.0",
     "classnames": "^2.2.5",
-    "downshift": "^3.2.10",
+    "downshift": "^3.2.14",
     "fast-memoize": "^2.5.1",
     "fela": "^10.6.1",
     "fela-plugin-embedded": "^10.6.1",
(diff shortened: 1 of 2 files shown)


packages/react/package.json
```diff
@@ -11,7 +11,7 @@
     "@stardust-ui/react-component-ref": "^0.38.0",
     "@stardust-ui/react-proptypes": "^0.38.0",
     "classnames": "^2.2.5",
-    "downshift": "^3.2.10",
+    "downshift": "^3.2.14",
```
levithomason: Heads up, because we use the caret `^`, consumers' `yarn install @stardust-ui/react` will automatically resolve to the latest `3.x.x` version.

If we want consumers to get `3.2.x` then we need to specify the tilde `~3.2.10`, which will only allow the latest patch version.

Lastly, for users using yarn, the `yarn.lock` file will remember whatever actual version was resolved for a given `package.json` specified version.  So, there is that caveat.  Example, if a user installed Stardust when downshift `3.2.14` was available as the latest `3.x.x` version, then the resolution of `^3.2.10` would point to `3.2.14` and they would get that version every time until they updated their `yarn.lock`.

silviuaavram: sorry for the rollup issue, it wasn't obvious for me.

I will update to 3.2.14 fixed version in this PR. 3.3.0 is released but I did not ask for approval on that one.

Is this ok?

PR #1911, opened 2019-09-10
feat(Accessibility): Initial accessibility attributes schema.

mshoho:
This is a work-in-progress. The Stardust part shouldn't change, the JSON schema is about to be updated, but we can already start talking about it.

The key idea is that there is a schema which declaratively defines allowed accessibility attributes and how the accessible components can be composed. And there is a validator (which is being added to the non-production builds only) which validates the integrity at runtime and shows errors in case something is not satisfying the schema.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,6 +1,7 @@
 coverage:
   ignore:
     - docs/*
+    - packges/ability-attributes/src/*
     - packges/react/src/lib/*
     - packges/react/src/themes/*
     - packges/react/src/styles/*
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,7 +1,7 @@
 coverage:
   ignore:
     - docs/*
-    - packges/ability-attributes/src/*
+    - packages/ability-attributes/src/*
     - packges/react/src/lib/*
     - packges/react/src/themes/*
     - packges/react/src/styles/*
--- a/packages/ability-attributes/package.json
+++ b/packages/ability-attributes/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@stardust-ui/ability-attributes",
   "description": "Accessibility attributes schema for Stardust UI",
-  "version": "0.37.0",
+  "version": "0.38.0",
   "author": "Marat Abdullin <marata@microsoft.com>",
   "bugs": "https://github.com/stardust-ui/react/issues",
   "dependencies": {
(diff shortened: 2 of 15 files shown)


packages/ability-attributes/package.json
```diff
@@ -0,0 +1,37 @@
...
+  ],
+  "homepage": "https://github.com/stardust-ui/react/tree/master/packages/ability-attributes",
+  "jsnext:main": "dist/es/index.js",
+  "license": "MIT",
+  "main": "dist/commonjs/index.js",
+  "module": "dist/es/index.js",
+  "peerDependencies": {
+    "react": "^16.8.0",
+    "react-dom": "^16.8.0"
+  },
```
layershifter: Are these package are real peer dependencies?

packages/ability-attributes/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": "stardust-ui/react.git",
+  "scripts": {
+    "build": "npm run schema && gulp bundle:package:no-umd --package ability-attributes",
+    "schema": "allyschema -c \"process.env.NODE_ENV !== 'production'\" schema.json > ./src/schema.ts",
+    "preinstall": "npm run build",
+    "prepare": "npm run build"
```
layershifter: If we will add watch mode `preinstall` and `prepare` are not required anymore I think

packages/ability-attributes/src/index.ts
```diff
@@ -0,0 +1,9 @@
+import { DevEnv } from 'ability-attributes'
+
+export * from './schema'
+
+export function setup(settings?: DevEnv.DevEnvSettings) {
+  if (process.env.NODE_ENV !== 'production') {
+    DevEnv.setup({ ...{ enforceClasses: false, ignoreUnknownClasses: true }, ...settings })
```
layershifter:
```suggestion
    DevEnv.setup({ enforceClasses: false, ignoreUnknownClasses: true, ...settings })
```

nit

packages/react/src/index.ts
```diff
@@ -245,5 +245,11 @@ export * from './lib/accessibility/FocusZone/FocusTrapZone.types'
 export * from './lib/accessibility/FocusZone/AutoFocusZone.types'
 export * from './lib/accessibility/types'
 export * from './lib/accessibility/reactTypes'
+// Dev-time accessibility attributes integrity validation.
+import { setup } from '@stardust-ui/ability-attributes'
+
+if (process.env.NODE_ENV !== 'production') {
+  setup()
+}
```
layershifter: Can we do this only for doc site, now?
Do we really want to ship this stuff to consumers as default now?

codecov.yml
```diff
@@ -1,6 +1,7 @@
 coverage:
   ignore:
     - docs/*
+    - packges/ability-attributes/src/*
```
layershifter:
```suggestion
    - packages/ability-attributes/src/*
```

Let's fix this typo, I will fix others in a separate PR. Good catch 👍

layershifter (approved):

PR #1912, opened 2019-09-10
fix(Tree): fix focus styles

### LINKED ISSUE #1620: Tree item has focus ring when navigating using mouse (Teams Theme)
### Steps
Click on a tree item.

### Expected Result
No focus ring visible, Teams theme displays focus ring only when navigating using keyboard.

### Actual Result
image

### Version
0.34.1

silviuaavram:
Fixes https://github.com/stardust-ui/react/issues/1620

Adds screener tests for 4 cases:
1. leaf focus with mouse
2. leaf focus with kb
3. non-leaf(subtree item) focus with mouse
4. non-leaf(subtree item) focus with keyboard


packages/react/src/themes/teams/components/HierarchicalTree/hierarchicalTreeTitleStyles.ts
```diff
@@ -8,7 +8,12 @@ const hierarchicalTreeTitleStyles = {
     cursor: 'pointer',
     color: variables.defaultColor,
     position: 'relative',
-    ...getBorderFocusStyles({ siteVariables }),
+    ':focus': {
+      outline: 0,
+    },
+    ':focus-visible': {
+      ...getBorderFocusStyles({ siteVariables })[':focus-visible'],
+    },
```
layershifter: Previous code does the same as I see

silviuaavram: before these changes it was not working, we are missing something.

before the removal of isFromKeyboard we had the issue that borderfocus style was applied every time. now, if you check on the stardust docsite, it won't be applied in any case (mouse or keyboard).

these changes fixes it.

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -101,6 +101,8 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
 
   static autoControlledProps = ['activeItemIds']
 
+  static Item = TreeItem
```
lucivpav: Why is this needed?

silviuaavram: I think I used in the screener test as selector then reverted that and left this. will remove, thanks!

layershifter:
image

_Please do not accept screener tests before merging. Now, it's unclear what was really changed._

packages/react/src/themes/teams/components/HierarchicalTree/hierarchicalTreeTitleStyles.ts
layershifter: Can you please check this thing again?

image

It looks that this change is invalid.

packages/react/src/themes/teams/components/Tree/treeTitleStyles.ts
```diff
@@ -7,7 +7,12 @@ const treeTitleStyles = {
     cursor: 'pointer',
     color: v.color,
     position: 'relative',
-    ...getBorderFocusStyles({ siteVariables }),
+    ':focus': {
+      outline: 0,
+    },
+    ':focus-visible': {
+      ...getBorderFocusStyles({ siteVariables })[':focus-visible'],
+    },
```
layershifter: https://github.com/stardust-ui/react/pull/1912/files#r323679125

layershifter (approved):

layershifter:
Please add CHANGELOG entry.

PR #1913, opened 2019-09-10
docs(Knobs): improve LogInspector

layershifter:
This is next iteration of usage `LogInspector` and `useLogKnob`.

- it makes formatter for Stardust callbacks shareable
- improves `LogInspector` styles

### Before

image

### After

image


docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/createPackageJson.ts
```diff
@@ -7,7 +7,6 @@ const name = 'stardust-ui-example'
 const description =
   'An exported example from Stardust UI React, https://stardust-ui.github.io/react/'
 const dependencies = {
-  '@stardust-ui/code-sandbox': 'latest',
```
layershifter, line 10: Moved to `renderConfig.ts`

packages/code-sandbox/src/callbackLogFormatter.ts
```diff
@@ -0,0 +1,20 @@
+import { formatCode } from '@stardust-ui/docs-components'
+import * as React from 'react'
+
+const callbackLogFormatter = (props: string[]) => (
```
lucivpav: It's not easy to read this code. What do these double arrows mean? Can we extract some of it to make this simpler?
```
(props: string[]) => (
  name: string,
  e: React.SyntheticEvent,
  data: Object,
) => {
 ...
}
```

lucivpav: Can we have some documentation on how to use `callbackLogFormatter`?  It's not clear at first glance, what the parameter is about

packages/code-sandbox/src/callbackLogFormatter.ts
```diff
@@ -0,0 +1,20 @@
...
+  data: Object,
+) => {
+  const pickedProps = props.reduce((acc, propName) => {
+    acc[propName] = data[propName]
+    return acc
+  }, {})
+
+  return [
+    `/*[${new Date().toLocaleTimeString()}]*/`,
+    `${name} (e: { "type": "${e.type}" }, data: ${formatCode(JSON.stringify(pickedProps), 'json')}`,
```
mnajdova:
```suggestion
    `${name} (e: { "type": "${e.type}" }, data: ${formatCode(JSON.stringify(pickedProps), 'json')})`,
```

packages/code-sandbox/src/callbackLogFormatter.ts
```diff
@@ -0,0 +1,20 @@
...
+  e: React.SyntheticEvent,
+  data: Object,
+) => {
+  const pickedProps = props.reduce((acc, propName) => {
+    acc[propName] = data[propName]
+    return acc
+  }, {})
+
+  return [
+    `/*[${new Date().toLocaleTimeString()}]*/`,
```
mnajdova: Not sure why this is in comment... It's not like the rest of the information is a running code...

mnajdova:
Probably not related to these changes, but the boolean knob is not working as expected: https://i.imgur.com/yivHrLl.gif

docs/src/examples/components/Toolbar/Content/ToolbarExampleMenu.shorthand.tsx
```diff
@@ -1,48 +1,39 @@
-import * as React from 'react'
+import { callbackLogFormatter } from '@stardust-ui/code-sandbox'
+import { useBooleanKnob, useLogKnob } from '@stardust-ui/docs-components'
 import { Toolbar } from '@stardust-ui/react'
+import * as React from 'react'
 
 const ToolbarExampleMenuShorthand = () => {
-  const [menuOpen, setMenuOpen] = React.useState(false)
+  const [menuOpen, setMenuOpen] = useBooleanKnob({ initialValue: false, name: 'menuOpen' })
```
mnajdova: See my prev comment, this boolean knob is not working as expected.

docs/src/examples/components/Toolbar/Content/ToolbarExampleMenu.shorthand.tsx
```diff
@@ -1,48 +1,39 @@
...
 
-  const [log, setLog] = React.useState<string[]>([])
-  const writeLog = message => {
-    setLog(prevLog => [`${new Date().toLocaleTimeString()}: ${message}`, ...prevLog])
-  }
+  const onItemClick = useLogKnob('onItemClick', null, callbackLogFormatter(['content']))
+  const onMenuOpenChange = useLogKnob(
+    'onItemClick',
```
mnajdova:
```suggestion
    'onMenuOpenChange',
```
?

packages/code-sandbox/src/callbackLogFormatter.ts
mnajdova: Plus, if we log something different then callback, it would be different then the default formatter styles...

packages/code-sandbox/src/callbackLogFormatter.ts
layershifter: Actually, it's a factory that creates a new function that's why there double arrows. It should be named `createCallbackLogFormatter()`.

layershifter: I added a comment there.

packages/code-sandbox/src/callbackLogFormatter.ts
layershifter: In the it will be highlighted by `Prism.JS` as Javascript code, without comment it looks ugly 😭 

image

docs/src/examples/components/Toolbar/Content/ToolbarExampleMenu.shorthand.tsx
layershifter: Reverted it for now, I don't that it's related to the knob itself. Should be tested separately.

docs/src/examples/components/Toolbar/Content/ToolbarExampleMenu.shorthand.tsx
layershifter: Yep, good catch 👍

packages/code-sandbox/src/callbackLogFormatter.ts
mnajdova: Ok, let's then change the default one as well... Maybe that would be better

mnajdova (approved):
👍

PR #1915, opened 2019-09-10
fix(Build): Use downshift 3.2.x (instead of 3.x.x)

miroslavstastny:
Downshift 3.3.0 breaks rollup build: 
```
[!] Error: 'getKey' is not exported by node_modules/keyboard-key/src/keyboardKey.js
https://rollupjs.org/guide/en/#error-name-is-not-exported-by-module
node_modules/downshift/dist/downshift.esm.js (9:9)
 7: import { isForwardRef } from 'react-is';
 8: import computeScrollIntoView from 'compute-scroll-into-view';
 9: import { getKey } from 'keyboard-key';
             ^
```

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -11,7 +11,7 @@
     "@stardust-ui/react-component-ref": "^0.38.0",
     "@stardust-ui/react-proptypes": "^0.38.0",
     "classnames": "^2.2.5",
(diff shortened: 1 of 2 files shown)


mnajdova (approved):

levithomason (approved):
👍

Issue #1918, opened 2019-09-11
Create `UsageSection` component


lucivpav:
### Problem description
Currently, when you want to add examples to the Usage section, you need to use the following confusing code: `<ExampleSection title="">`. The title should be empty, because we won't need to label a section that is already under Usage as Usage.

### Proposed solution
`<UsageSection>` would be used instead.

PR #1919, opened 2019-09-11
docs(Usage): create `UsageSection` component

lucivpav:
Implements #1918

### DIFF
--- a/docs/src/components/ComponentDoc/UsageSection.tsx
+++ b/docs/src/components/ComponentDoc/UsageSection.tsx
@@ -7,11 +7,9 @@ const gridStyle: React.CSSProperties = {
 
 // minmax = prevent example overflow - https://stackoverflow.com/a/43312314
 const ExampleSection: React.FC = ({ children }) => (
-  <>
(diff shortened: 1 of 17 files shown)


docs/src/components/ComponentDoc/UsageSection.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react'
+import { Grid } from '@stardust-ui/react'
+
+const gridStyle: React.CSSProperties = {
+  marginTop: '1.5em',
+}
+
+// minmax = prevent example overflow - https://stackoverflow.com/a/43312314
+const ExampleSection: React.FC = ({ children }) => (
+  <>
```
layershifter:
```suggestion
```

`Fragment` is useless there

levithomason:
Heads up, we're getting a lot of complaints that it's hard to find examples that are hidden in the Usage section. We might want to consider keeping all examples in the Definition tab and maybe renaming that Examples?

Thoughts?

lucivpav:
Closing in favor of #1948

PR #1920, opened 2019-09-11
feat(Toolbar): add support for radio & checkbox groups

layershifter:
This PR:
- adds a new `toggle` `kind` of items in `ToolbarItem`'s `menu`
- adds a new `group` `kind` of items in `ToolbarItem`'s `menu`, which is implemented by `ToolbarMenuRadioGroup`
- added new accessibility behaviors to implements there features

Based on: https://www.w3.org/TR/wai-aria-practices/examples/menubar/menubar-2/menubar-2.html


packages/react/src/components/Toolbar/ToolbarMenu.tsx
```diff
@@ -70,16 +71,38 @@ class ToolbarMenu extends UIComponent<ToolbarMenuProps> {
...
     const dividerOverridesFn = this.handleDividerOverrides(variables)
+    const radioGroupOverrides = this.handleRadioGroupOverrides(variables)
 
     return _.map(items, item => {
       const kind = _.get(item, 'kind', 'item')
-      if (kind === 'divider') {
-        return ToolbarMenuDivider.create(item, { overrideProps: dividerOverridesFn })
+
+      switch (kind) {
+        case 'checkbox':
```
jurokapsiar: just noticed that we call it `toggle` in the `Toolbar` component. We should probably have the same naming across these two

packages/react/test/specs/components/Toolbar/ToolbarMenuRadioGroup-test.ts
```diff
@@ -0,0 +1,10 @@
+import { isConformant } from 'test/specs/commonTests'
+
+import Box from 'src/components/Box/Box'
+import ToolbarMenuRadioGroup from 'src/components/Toolbar/ToolbarMenuRadioGroup'
+
+describe('ToolbarMenuRadioGroup', () => {
+  isConformant(ToolbarMenuRadioGroup, {
```
jurokapsiar, line 7: pls add handlesAccessibility test

docs/src/components/ColorSchemes.tsx
```diff
@@ -3,18 +3,18 @@ import * as _ from 'lodash'
...
   ShorthandCollection,
+  ThemeInput,
 } from '@stardust-ui/react'
 
 import ColorBox from './ColorBox'
 
 type ColorVariantsProps = {
   name?: string
-  themes?: ThemePrepared[]
+  themes?: ThemeInput[]
```
jurokapsiar: should this be in this PR?

packages/react/src/lib/accessibility/Behaviors/Toolbar/toolbarMenuItemRadioBehavior.ts
```diff
@@ -0,0 +1,37 @@
...
+ *  Adds attribute 'tabIndex=0' to 'root' slot.
+ */
+const toolbarMenuItemRadioBehavior: Accessibility<ToolbarMenuItemRadioBehaviorProps> = props => ({
+  attributes: {
+    root: {
+      [IS_FOCUSABLE_ATTRIBUTE]: true,
+      'aria-checked': props.checked,
+      'aria-disabled': props.disabled,
+      role: 'menuitemradio',
+      tabIndex: 0,
```
jurokapsiar: we should not need tabIndex if data-is-focusable is used and the item is in focus zone

packages/react/src/components/Toolbar/ToolbarMenuRadioGroup.tsx
```diff
@@ -0,0 +1,147 @@
...
+  extends UIComponentProps,
+    ChildrenComponentProps,
+    ContentComponentProps {
+  /**
+   * Accessibility behavior if overridden by the user.
+   */
+  accessibility?: Accessibility
+
+  /** Index of the currently active item. */
+  activeIndex?: number
```
jurokapsiar, line 42: to meke it predictable, I think we should add activeIndex to toolbarRadioGroup as well

jurokapsiar (dismissed):

kolaps33 (approved):

docs/src/components/ColorSchemes.tsx
layershifter: Nope, #1941 fixed this issue.

packages/react/test/specs/components/Toolbar/ToolbarMenuRadioGroup-test.ts
layershifter: Added 👍

packages/react/src/lib/accessibility/Behaviors/Toolbar/toolbarMenuItemRadioBehavior.ts
layershifter: Good catch, removed 👍

packages/react/src/components/Toolbar/ToolbarMenuRadioGroup.tsx
layershifter: Added 👍

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
```diff
@@ -35,12 +35,21 @@ export interface ToolbarMenuItemProps
   /** A toolbar item can be active. */
   active?: boolean
 
+  /** Item can show check indicator if selected. */
+  checked?: boolean
```
jurokapsiar: sorry that I missed it before - for toggle in the toolbar we are using toggleButtonBehavior, there the prop is called `active`. That way it would even be consistent with the toolbar menu radio.

PR #1925, opened 2019-09-12
perf: use `inline-style-expand-shorthand` for expanding

### LINKED ISSUE #1995: Plugins: expand is not working for `rgba()` values with space
## Bug Report

We use `css-expand-shorthand` to expand longhand CSS rules to short chunks. But there is a bug in this library:

```js
// ✅ {border-top-color: "rgba(255,255,1,1)", ... }
expand("border-color", "rgba(255,255,1,1)")

// 🔴  {border-top-color: "rgba(255,", border-right-color: "255,1,1)", ... }
expand("border-color", "rgba(255, 255,1,1)") 
```

### Steps
1. Open https://codesandbox.io/s/stardust-ui-example-kg48v

### Expected Result
`borderColor` should be applied

### Actual Result
Notice that `borderColor` is not applied

### Version
0.39.0

layershifter:
# BREAKING CHANGES

As `inline-style-expand-shorthand` doesn't support expanding of [`font`](https://www.w3schools.com/cssref/pr_font_font.asp) CSS property this probably should be done manually in styles/overrides:

### Before

```tsx
{
  font: "15px arial, sans-serif",
}
```

### After

```tsx
{
  fontSize: "15px",
  fontFamily: "arial, sans-serif",
}
```

---

This PR replaces `css-shorthand-expand` (we used compiled version of it) with `inline-style-expand-shorthand` which is used in `fela-plugin-expand-shorthand`.

We cannot use `fela-plugin-expand-shorthand` as it introduced regressions due merging order so I just replaced `css-shorthand-expand` which allows us to gain 10% performance and drop complied code.

# 🏁 Measures

```sh
yarn perf --filter *ChatWithPop*
```

Performance diff is **9.71%** for the example below.

### Before

| Example                  | min    | avg    | median | max    |
| ------------------------ | ------ | ------ | ------ | ------ |
| ChatWithPopover.perf.tsx | 363.42 | 378.07 | 372.95 | 453.05 |
| ChatWithPopover.perf.tsx | 363.07 | 379.59 | 375.67 | 425.95 |
|                          |        | 378.83 |        |

### After

| Example                  | min    | avg     | median | max    |
| ------------------------ | ------ | ------- | ------ | ------ |
| ChatWithPopover.perf.tsx | 328.7  | 341.16  | 338.68 | 383.49 |
| ChatWithPopover.perf.tsx | 332.13 | 345.61  | 341.96 | 424.88 |
|                          |        | 343.385 |        |        |


layershifter:
✅ https://github.com/robinweser/inline-style-expand-shorthand/pull/11 solved
✅ robinweser/fela#735 solved

jdhuntington:

packages/react/src/utils/felaRenderer.tsx
```diff
@@ -67,12 +67,12 @@ const rendererConfig = {
     felaPluginEmbedded(),
     felaPluginPrefixer(),
 
+    felaExpandCssShorthandsPlugin(),
```
layershifter, line 70: This move is valid as `inline-style-expand-shorthand` handles arrays on its side.

mnajdova: How are we handling arrays exactly? As far as I can see, we are just spreading the array:
```
if (Array.isArray(cssPropertyValue)) {
    return { ...acc, [cssPropertyName]: cssPropertyValue }
}
```

Is the `expandProperty` function handling arrays?

packages/react/test/specs/utils/felaExpandCssShorthandsPlugin-test.ts
```diff
@@ -18,6 +18,21 @@ describe('felaExpandCssShorthandsPlugin', () => {
     })
   })
 
+  test('should expand handle "undefined" and "null"', () => {
```
mnajdova: Can we add test for the `array` value scenario?

packages/react/src/utils/felaExpandCssShorthandsPlugin.ts
```diff
@@ -1,52 +1,56 @@
...
-  borderRadius: 'border-radius',
-  background: 'background',
-  outline: 'outline',
+import { ICSSInJSStyle } from '@fluentui/styles'
+import * as CSS from 'csstype'
+import { expandProperty } from 'inline-style-expand-shorthand'
+
+// https://jsperf.com/array-indexof-vs-object-key-lookup2/12
+const handledCssProps: Partial<Record<keyof CSS.Properties, true>> = {
+  // 'font', Oops, is not supported by inline-style-expand-shorthand
```
mnajdova, line 7: Please add **BREAKING CHANGE** changelog entry

packages/react/test/specs/utils/felaExpandCssShorthandsPlugin-test.ts
layershifter: Sure 🎸

layershifter: Added 👍

packages/react/src/utils/felaExpandCssShorthandsPlugin.ts
layershifter: Added and updated PR's description

packages/react/src/utils/felaRenderer.tsx
layershifter: `expandProperty` exactly, I added this functionality in https://github.com/robinweser/inline-style-expand-shorthand/pull/13

mnajdova (approved):

Issue #1894, opened 2019-09-05
Support Popup in Toolbar menu


joheredi:
### Problem description
Currently Toolbar component's (overflow) menu doesn't allow triggering popups from the (overflow) menu, popup support would have the following requirements:

### Proposed solution
Support triggering popups from Toolbar (overflow) menu keeping in mind that:
* (Overflow) menu should remain open when popup is opened

### MVP
[Fork, update, and replace this sandbox if you can show the proposed solution]:
https://codesandbox.io/s/stardust-ui-example-8ismt

jurokapsiar:
Fixed by #1927, closing the issue. Target release 0.39

PR #1927, opened 2019-09-12
feat(Toolbar): support for a popup in menu

lucivpav:
Implements #1894

### Scenarios in E2E tests
* Popup can be opened using mouse
* Popup can be opened using keyboard
* Tab when the Popup is focused should not result in hiding the Popup
* Click inside Popup should not hide Popup 
* Popup can be closed when clicking outside of the elements
* Click outside of Popup but inside of Menu closes Popup but leaves Menu open

### TODO
* Try whether it still works in overflow


docs/src/examples/components/Toolbar/Usage/ToolbarExampleWithTooltip.shorthand.tsx
```diff
@@ -67,6 +71,7 @@ const ToolbarExampleShorthand = () => {
                   // Adding tooltipAsLabelBehavior as the ToolbarItems contains only icon
                   return (
                     <Tooltip
+                      key={`${rest.key}-tooltip`} // to avoid errors with keys
```
lucivpav: to avoid errors with keys

docs/src/examples/components/Toolbar/Usage/ToolbarExampleWithTooltip.shorthand.tsx
```diff
@@ -44,10 +44,14 @@ const ToolbarExampleShorthand = () => {
...
           menu: [
             {
+              wrapper: undefined, // to avoid issue with keys
+              key: 'quote',
               content: 'Quote',
               icon: 'quote',
             },
             {
+              wrapper: undefined,
+              key: 'code-snippet', // to avoid issue with keys
```
lucivpav: Move the comment one line up

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -264,7 +267,11 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>> {
   }
 
   handleWrapperBlur = e => {
-    if (this.props.menu && !e.currentTarget.contains(e.relatedTarget)) {
+    if (
+      this.props.menu &&
+      !e.currentTarget.contains(e.relatedTarget) &&
+      e instanceof KeyboardEvent
```
lucivpav: Looks like this was never correct as it always evaluates as `false`

packages/react/src/lib/accessibility/FocusZone/FocusTrapZone.tsx
```diff
@@ -209,6 +209,8 @@ export default class FocusTrapZone extends React.Component<FocusTrapZoneProps, {
     if (!this._root.current.contains(relatedTarget as HTMLElement)) {
       this._hasFocus = false
     }
+
+    ev.stopPropagation()
```
lucivpav: Without this, tabbing inside a `Popup` that was created as part of a `ToolbarMenu`, will cause the `Popup` to be closed. The `onBlur` event in `ToolbarItem` would be called, which I believe is not correct behavior (since the `Popup` is in a focus trap).

packages/react/src/types.ts
```diff
@@ -36,6 +36,8 @@ export type ComponentKeyboardEventHandler<TProps> = (
   data?: TProps,
 ) => void
 
+export type ComponentMouseEventHandler<TProps> = (event: MouseEvent, data?: TProps) => void
```
layershifter: Let's do not do this, it's to early for this change and it should be done separately for all components

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
```diff
@@ -115,6 +154,24 @@ class ToolbarMenuItem extends UIComponent<WithAsProp<ToolbarMenuItemProps>> {
       </ElementType>
     )
 
+    const hasChildren = childrenExist(children)
+
+    if (popup && !hasChildren) {
+      return Popup.create(popup, {
+        defaultProps: {
+          trapFocus: true,
```
jurokapsiar, line 171: I agree that this is more convenient, but since trapFocus is not the default for normal <Popup /> we probably should not put it here and be consistent.

docs/src/examples/components/Toolbar/Usage/ToolbarExampleWithTooltip.shorthand.tsx
```diff
@@ -44,10 +44,14 @@ const ToolbarExampleShorthand = () => {
           tooltip: 'More options',
           menu: [
             {
+              wrapper: undefined, // to avoid issue with keys
```
jurokapsiar: let's try to find a better solution - how would a developer know that he should use wrapper: undefined?

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
lucivpav: `ToolbarItem` also sets `trapFocus: true`

https://github.com/stardust-ui/react/blob/master/packages/react/src/components/Toolbar/ToolbarItem.tsx#L225

docs/src/examples/components/Toolbar/Usage/ToolbarExampleWithTooltip.shorthand.tsx
lucivpav: I can no longer reproduce any errors without setting `wrapper` to `undefined`. Removed.

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
jurokapsiar: didn't know about it. ok.

docs/src/examples/components/Toolbar/Usage/ToolbarExamplePopupInMenu.shorthand.tsx
```diff
@@ -0,0 +1,65 @@
+import * as React from 'react'
+import { Toolbar, ToolbarItemShorthandKinds, Input } from '@stardust-ui/react'
+
+const PopupContent = <Input icon="search" placeholder="Search..." />
+
+const ToolbarExamplePopupInMenu = () => {
+  const [menu1Open, setMenu1Open] = React.useState(false)
```
jurokapsiar, line 5: can you also add an example (not sure if this or a new file) that shows how they can close the popup after a successful action? for example, if the user selected the item that he wanted to select - in that case both the popup as well as the submenu should close. I want to understand how much state handling would the developer need to do.

docs/src/examples/components/Toolbar/Usage/ToolbarExamplePopupInMenu.shorthand.tsx
```diff
@@ -0,0 +1,62 @@
...
+            key: 'menu2',
+            icon: 'more',
+            active: menu2Open,
+            menu: [
+              {
+                key: 'popup',
+                content: 'Open Popup',
+                popup: {
+                  content: PopupContent,
+                },
```
layershifter: Can we try to do this? I expect that it should work

```
popup: <Input icon="search" placeholder="Search..." />,
```
````

lucivpav:
It is broken again when you click on popup's background. Will be fixed in seperate PR: #1944

packages/react/src/lib/accessibility/FocusZone/FocusTrapZone.tsx
lucivpav: **This also prevents an issue, where you click on the Popup's background and it gets closed.**

docs/src/examples/components/Toolbar/Usage/ToolbarExamplePopupInMenu.shorthand.tsx
lucivpav: Yes, its available here: https://github.com/stardust-ui/react/pull/1927/files#diff-1e0d18e5b10bb6e298c5bdfa5516de67

layershifter (approved):

jurokapsiar (approved):
nice work!

PR #1929, opened 2019-09-13
fix(Chat): use more accurate positioning in scrollable containers

### LINKED ISSUE #1926: Chat: Message action clipped on the left side if message is short
### Steps
Open testcase, hover on the third message

### Expected Result
Actions menu is shifted so that it does not get clipped

### Actual Result
Menu is clipped

### Version
0.38.0

### Testcase
https://codesandbox.io/s/stardust-ui-example-zmqf5

layershifter:
Fixes #1926.

***

Previously we used `unstable_pinned` on `actions`, this prop will only disable `flip` modifier which is not correct.

This PR does these changes:
- `flip` modifier is enabled and will move `actions` only on `top`
- `preventOverflow` will work only for `left`/`right` positions

#### Before

image

#### After

image


packages/docs-components/src/knobs/useRangeKnob.ts
```diff
@@ -8,13 +8,19 @@ type UseRangeKnobSpecificOptions = {
   step?: number | string
   unit?: string
 }
-type UseRangeKnobOptions = UseKnobOptions<number | string> & UseRangeKnobSpecificOptions
+type UseRangeKnobOptions<T> = UseKnobOptions<T> & UseRangeKnobSpecificOptions
 
-const useRangeKnob = (options: UseRangeKnobOptions) => {
-  const { initialValue = 3, min = 0, max = parseValue(initialValue), step = 1, ...rest } = options
+const useRangeKnob = <T extends number | string>(options: UseRangeKnobOptions<T>) => {
```
layershifter, line 13: Fix to properly define `number`/`string` return type

jurokapsiar (approved):

PR #1930, opened 2019-09-13
fix(FocusTrapZone): Do not force focus inside focus trap zone on outside focus

sophieH29:
Popup in Menu scenario was broken due to a recent focus trap zone update to the Fabric latest version.

For now, switching it back to Stardust default behavior, we will investigate later how to align.

Repro: https://codesandbox.io/s/stardust-ui-example-yjnqv


jurokapsiar:
Keeping some context for the future here:

This behavior changed between different versions of (Fabric) FocusTrapZone. So reverting to the previous behavior as it breaks some of the scenarios in Stardust.

To use the default Fabric behavior, we can still use 
```
trapFocus={ {
  forceFocusInsideTrapOnOutsideFocus: true
} }
```

We should probably use false for popup by default and true for dialog (?) but it requires some testing.

layershifter (approved):

PR #1934, opened 2019-09-16
docs(Tree): remove renderTreeTitle prop from exclusive example

silviuaavram:
There is no need for `renderItemTitle` in the `exclusive` example and it will make the example that actually illustrates `renderItemTitle` useless.


lucivpav:
Which example becomes useless? If so, should it be removed?

silviuaavram:
> Which example becomes useless? If so, should it be removed?

the `renderTreeTitle` one would have become useless if we also used `renderTreeTitle` in `exclusive` example.

These are 2 distinct examples that should illustrate 2 different props in the Tree.

That is why I removed the renderTreeTitle in the `exclusive` example.

lucivpav (approved):

PR #1937, opened 2019-09-16
fix(menuButtonBehavior): Don't add 'aria-haspopup', when 'MenuButton' behaves as context menu

kolaps33:
When 'MenuButton' is use like contextMenu, then 'aria-haspopup' attribute is confusing for the user.  
If 'aria-haspopup' is defined, then user expect menu will be invoke by enter/space key on the button.
But in context menu approach menu appear on shift+F10. 
We consider not use 'aria-haspopup' for contex menu button. In this case button will be need to be labelled by custom string, informing user about the context menu availability.


packages/react/test/specs/behaviors/testDefinitions.ts
```diff
@@ -258,6 +258,36 @@ definitions.push({
   },
 })
 
+// Adds attribute 'aria-haspopup=true' to 'trigger' slot if 'contextMenu' property is not set.
```
silviuaavram, line 261: Shouldn't it be `is falsy` ? I assume it has the same behavior for `contextMenu: false` right?

packages/react/test/specs/behaviors/testDefinitions.ts
```diff
@@ -258,6 +258,36 @@ definitions.push({
...
+  regexp: /Adds attribute '([\w-]+)=([\w\d]+)' to '([\w-]+)' slot if '([\w-]+)' property is not set\./g,
+  testMethod: (parameters: TestMethod) => {
+    const [
+      attributeToBeAdded,
+      valueOfAttributeToBeAdded,
+      slot,
+      propertyDependsOn,
+    ] = parameters.props
+
+    const propertyDependsOnUndefined = {}
```
silviuaavram, line 272: what i mean is shouldn't we test it with this with `const propertyDependsOnUndefined = {[propertyDependsOn]: false}` as well?

kolaps33: added below the test with "false" value:

  const propertyDependsOnFalse = {
      [propertyDependsOn]: false,
    }

packages/react/test/specs/behaviors/testDefinitions.ts
kolaps33: I added test for the "false" value below. :)
 
The description I would rather let "if 'contextMenu' property is not set."
Because high probably I expect the usage will be not setting 'contextMenu' prop at all.

Or use as you suggesting:
"if 'contextMenu' property is falsy." 
- it would make sense, but for me is more natural to inform user(consumer) of 'stardust' that "aria-haspopup=true" will be set, when he will not set the "contextMenu" prop...

silviuaavram (approved):
don't forget the changelog

Issue #1648, opened 2019-07-16
Improve `Form.Field` API for `control` slot


bmdalex:
# Improve `Form.Field` API for `control` slot

## Problem description

Currently the `control` slot for `Form.Field` is a shorthand that defaults to `{ as: Input }` object:
https://github.com/stardust-ui/react/blob/424ba3c8b20835921b79cf1591115b98623c7fef/packages/react/src/components/Form/FormField.tsx#L74

where `Input` is a Stardust component. This means the API for using a different control, let's say our `Slider` component and the primitive `<input type="range" />` is going to be:

- for custom components:
```jsx
<Form.Field label="Age" control={{ as: Slider }} id="age" />
```
- for primitives:
```jsx
<Form.Field label="Age" control={{ as: 'input', type: 'range' }} id="age" />
```

However, this API is unintuitive (why is `as` prop needed?) and should be simplified.

## Proposed solution

Change the type of `control` slot from shorthand, aka `ShorthandValue<BoxProps>` (`ReactNode | Props<BoxProps>`) to:

#### 1. `React.ReactNode`

The example above would become:
- for custom components:
```jsx
<Form.Field label="Age" control={<Slider />} id="age" />
```
- for primitives:
```jsx
<Form.Field label="Age" control={<input type='range' />} id="age" />
```

#### 2. `React.ElementType | 'button' | 'input' | 'select' | 'textarea'`

For this approach we need to add `type` props as well to the `Form.Field`. The example above would become:

- for custom components:
```jsx
<Form.Field label="Age" control={Slider} id="age" />
```
- for primitives:
```jsx
<Form.Field label="Age" control='input' type='range' id="age" />
```

similar to SUIR: https://github.com/Semantic-Org/Semantic-UI-React/blob/048bd35ad310bcae6e6ebc8cd9c5cf52f5b927f2/src/collections/Form/FormField.js#L153

### MVP
https://codesandbox.io/s/stardust-ui-example-x3x54

levithomason:
Per offline meeting on this, let's start with proposal number 2 👍 Additional bonus, you can copy patterns and examples from SUIR.

levithomason:
Yes, we'd like to change this.  We aren't getting any complaints nor time to do so, so housekeeping for now.

Contributions welcome.

PR #1938, opened 2019-09-16
fix(Dropdown): properly accept id on searchInput from Form

silviuaavram:
pass `searchInput.id` to `Downshift's` prop `inputId`.
fix the Form + Dropdown example to make the searcInput focusable on label click.


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix `CreateShorthandOptions` should be typed @lucivpav ([#1886](https://github.com/stardust-ui/react/pull/1886))
 - Add `shadowLevel1Dark` in Teams themes @notandrew ([#1887](https://github.com/stardust-ui/react/pull/1887))
 - When merging themes use deep merge for site and component variables @miroslavstastny ([#1907](https://github.com/stardust-ui/react/pull/1907))
+- Fix `Dropdown` to properly accept `id` on `searchInpiut` @silviuavram ([#1938](https://github.com/stardust-ui/react/pull/1938))
```
```suggestion
- Fix `Dropdown` to properly accept `id` on `searchInput` @silviuavram ([#1938](https://github.com/stardust-ui/react/pull/1938))
```
Typo

docs/src/examples/components/Form/Usage/FormExampleDropdown.shorthand.tsx
```diff
@@ -7,12 +7,17 @@ const fields = [
...
     control: {
       as: Dropdown,
       items: ['John Doe', 'Dohn Joe', 'John Joe', 'Dohn Doe'],
       'aria-labelledby': labelId,
       search: true,
       placeholder: 'Choose a friend',
+      searchInput: {
+        id: 'choose-friend-shorthand', // id needs to end up on the search input.
+      },
+      id: undefined, // not on the main wrapper element.
```
layershifter, line 20: This thing looks like very strange. Specify `id`s in three places to make it working?

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -431,6 +431,7 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
           onStateChange={this.handleStateChange}
           labelId={this.props['aria-labelledby']}
           environment={this.context.target.defaultView}
+          inputId={this.props.searchInput ? this.props.searchInput['id'] : undefined}
```
```tsx
<Dropdown searchInput={render => render('boo')} />
<Dropdown searchInput='Oops' />
```

This thing is pretty fragile. Do we have other options?

docs/src/examples/components/Form/Usage/FormExampleDropdown.shorthand.tsx
silviuaavram: I don't see another choice, the most we can do is explain why.

That ID needs to end up on the `searchInput` in this case, since that is what we want to be focused on label click.

The `undefined` part is a bit sad since our `Form` assumes that the `id` needs to end up on the wrapper.

Any other ideas?

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: you are right, I need to add another check. this will only work if `searchInput` is an object.

docs/src/examples/components/Form/Usage/FormExampleDropdown.shorthand.tsx
jurokapsiar: I think the whole integration of our "input" components with Form does not work as expected - Dropdown searchable/non-searchable, checkbox, radio - all work differently and need some workarounds. @silviuavram can you spend some time thinking about a better proposal? @bmdalex (hi) had a proposal to use

```
<Form.Field control={Dropdown} />
```
(https://github.com/stardust-ui/react/issues/1648)

That way, Stardust can at least understand what is the underlying input (`if control===Dropdown`) and pass the id in a correct way. It still won't work for wrapped or custom components, but at least it would work out of the box for the most common use cases. What do you think @layershifter ?

silviuaavram: I will look into this. Maybe our input components should provide a label API. Can we close this ticket?

layershifter: It's how it's done in SUIR.

We discussed this on our of our syncs and agreed that it will a better option.

jurokapsiar (approved):

PR #1939, opened 2019-09-16
refactor(Chat): enable Children API

layershifter:


docs/src/examples/components/Chat/Types/ChatExample.shorthand.tsx
```diff
@@ -8,101 +8,106 @@ const janeAvatar = {
...
+      mine: true,
+    },
     contentPosition: 'end',
     attached: 'bottom',
     key: 'message-id-3',
   },
   {
     gutter: <Avatar {...janeAvatar} />,
-    message: <Chat.Message content="Hi" author="Jane Doe" timestamp="Yesterday, 10:15 PM" />,
+    message: {
```
levithomason, line 49: Consider that this is getting wrapped by some other element that the user now no longer has shorthand access to.  They should be able to add props and styles to this hidden area.  Our APIs shouldn't ever hide an area from the user.

packages/react/src/themes/teams/components/Chat/chatGutterStyles.ts
```diff
@@ -0,0 +1,13 @@
+import { ICSSInJSStyle, ComponentSlotStylesPrepared } from '../../../types'
+
+import { ChatGutterProps } from '../../../../components/Chat/ChatGutter'
+import { ChatGutterVariables } from './chatGutterVariables'
+
+const chatItemStyles: ComponentSlotStylesPrepared<ChatGutterProps, ChatGutterVariables> = {
```
levithomason: `const chatGutterStyles`

packages/react/src/themes/teams/components/Chat/chatItemStyles.ts
```diff
@@ -11,22 +13,17 @@ const chatItemStyles: ComponentSlotStylesPrepared<ChatItemProps, ChatItemVariabl
...
-  gutter: ({ props: p, variables: v }): ICSSInJSStyle => ({
-    position: 'absolute',
-    marginTop: v.gutterMargin,
-    [p.contentPosition === 'end' ? 'right' : 'left']: 0,
-    ...((p.attached === 'bottom' || p.attached === true) && {
-      display: 'none',
-    }),
-  }),
+    [`& .${ChatGutter.className}`]: {
+      [p.contentPosition === 'end' ? 'right' : 'left']: 0,
```
levithomason: Pass props through to the ChatGutter instead of using nested selectors.  It will improve debugging, has better CSS specificity, and is inline with our current architecture statement, "We handle everything for you with shorthand, you handle everything with children."

PR #1941, opened 2019-09-17
fix(teams-theme): add proper typings

mnajdova:
Add typings for teams theme.


levithomason (commented):
Probably want to add a changelog entry for this as a fix, no?

levithomason (approved):
Just link this PR to the original work with a comment.

PR #1942, opened 2019-09-17
fix(tooltip): Make tooltip visible for screen reader 

kolaps33:
We hit the scenario that tooltip was not narrated, when was not visible in UI. 
Exact usage is, when tooltip is referenced in "aria-labelledby"  of the other element.

Solution which could work(at least for screen reader) is hide the tooltip element with opacity: 0


layershifter (approved):
Please add an entry to changelog. Change is related to Teams theme.

jurokapsiar (approved):

PR #1949, opened 2019-09-18
docs(Dialog): remove cancel button in an example containing X

lucivpav:
Both X and cancel button cannot coexist per design guidelines.

Don't forget to close the related VSTS issue once this is merged.


layershifter (changes requested):
Stardust is not a guideline to MS Teams. 
I want to avoid merging this as it's very frequent thing in Web.

jurokapsiar (approved):

layershifter (approved):

PR #1950, opened 2019-09-19
chore(codesandbox): add prettier dependency

lucivpav:
Currently, if you export an example to CodeSandbox, you'll get an error that a `prettier` dependency is missing. This PR adds the dependency.

## Before
image

## After
image


layershifter (approved):

lucivpav:
@layershifter I have changed the import to `prettier/standalone` to avoid build warnings. Please let me know if you agree with this change

jurokapsiar (approved):

docs/src/components/Playground/renderConfig.ts
```diff
@@ -53,6 +54,10 @@ export const imports: Record<string, { version: string; module: any }> = {
     version: stardustReactPackageJson.dependencies['react-fela'],
     module: ReactFela,
   },
+  prettier: {
+    version: docsComponentsPackageJson.peerDependencies['prettier'],
+    module: _prettier,
```
layershifter: It can be even just `null` as we don't use it in our examples.

PR #1951, opened 2019-09-19
feat(factories): use `children` in shorthands as render props

layershifter:
## ⏩ Migration

We decided to go with soft migration, existing API with render callback will be deprecated and will create warnings.

### Case 0: slot replacement

Used when you need to replace a component in a slot, for example replace `Button` with `SplitButton` in `Dialog`.

#### Before

```tsx
<Dialog
  confirmButton={render =>
    render({ content: "Confirm" }, (Component, props) => {
      const { primary, ...rest } = props;

      return (
        <SplitButton
          menu={["New group message", "New channel message"]}
          button={rest}
          primary={primary}
        />
      );
    })
  }
  trigger={<Button content="Open a dialog" />}
/>
```

#### After

```tsx
<Dialog
  confirmButton={{
    content: "Confirm",
    children: (Component, props) => {
      const { primary, ...rest } = props;

      return (
        <SplitButton
          menu={["New group message", "New channel message"]}
          button={rest}
          primary={primary}
        />
      );
    }
  }}
  trigger={<Button content="Open a dialog" />}
/>
```

### Case 1: wrapping slot

Used when you need to wrap component slot with `Popup` or `Tooltip`, for example.

#### Before

```jsx
<Button
  content={render =>
    render({ content: "Foo", tabIndex: 1 }, (Component, props) => (
      <Popup trigger={<Component {...props} />} />
    ))
  }
/>
```

#### After

```tsx
<Button
  content={{
    content: "Foo",
    tabIndex: 1,
    children: (Component, props) => <Popup trigger={<Component {...props} />} />
  }}
/>
```

### Case 2: wrapping item

Used when you need to wrap the whole item with `Popup` or `Tooltip`, for example.

#### Before

```tsx
const items = [
  render =>
    render({ content: 'Foo' }, (Component, props) => <Popup trigger={<Component {...props} />} />),
  { content: 'Bar' },
]

<Menu items={items} />
```

#### After

```jsx
const items = [
  {
    content: "Foo",
    children: (Component, props) => <Popup trigger={<Component {...props} />} />
  },
  { content: "Bar" }
];

const MenuExample = () => <Menu items={items} />;
```

### Case 3: delayed execution

Is used in async scenarios. We deprecated this use case, see Motivation part. There is only one way to migrate is to lift up state, it works even now.

CodeSandbox with Before & After: https://codesandbox.io/s/stardust-ui-example-23icy

### Before

```tsx
<Avatar
  image="https://stardust-ui.github.io/react/public/images/avatar/small/matt.jpg"
  status={render => (
    <QueryMock>
      {status => {
        if (status) {
          return render({
            color: "green",
            icon: "stardust-checkmark",
            title: "Available"
          });
        }

        return render({
          color: "grey",
          icon: "sync",
          title: "Loading"
        });
      }}
    </QueryMock>
  )}
/>
```

### After

```tsx
<QueryMock>
  {status => (
    <Avatar
      image="https://stardust-ui.github.io/react/public/images/avatar/small/matt.jpg"
      status={
        status
          ? {
              color: "green",
              icon: "stardust-checkmark",
              title: "Available"
            }
          : {
              color: "grey",
              icon: "sync",
              title: "Loading"
            }
      }
    />
  )}
</QueryMock>
```

## ⌚️ Motivation

Existing API with `render` callback is very powerful and in the same time is over-complicated: it's hard to explain to people how it works and it's hard to use for frequent use cases.

```tsx
<Menu
  items={[
    (
      render // callback 1
    ) =>
      render(
        // callback 2
        {
          key: "editorials",
          content: "Editorials"
        },
        (Component, props) => (
          <Popup content="Foo" trigger={<Component {...props} />} />
        )
      )
  ]}
/>
```

### Other cons

- it allows developers to do really strange things with passing `render` down into React tree:

```tsx
<Menu
  items={[
    render => <ItemRenderer render={render} />} />
  ]}
/>
```

- the resolution of `kind` prop is broken:

```tsx
<Menu
  items={[
    render =>
      render(
        {
          key: "editorials",
          kind: "divider",
          content: "Editorials"
        },
        (Component, props) => <Component /> // renders MenuItem 😭
      )
  ]}
/>
```

## ⭐️ Solution

React has well known [`render props`](https://reactjs.org/docs/render-props.html) pattern and I suggest to use instead as more friendly option.

I propose to have only one type of usage, via `children` prop:

- it allows to deprecate render callback, otherwise there will be not obvious collision if we will decide to go with deprecation
- single option is better than multiple

```tsx
<Menu
  items={[
    {
      key: "events",
      content: "Upcoming Events",
      children: (Component, props) => (
        <Popup content="Popup" trigger={<Component {...props} />} />
      )
    }
  ]}
/>
```

_It still allows to pass shorthand props._


jurokapsiar:
this looks awesome! Migration would be a bit tricky for the consumers though, let'see how we can handle that.

jurokapsiar:
passing the render down the tree, even though strange, was used for a reason - if the shorthand props are resolved asynchronously. Do you see a good solution in this case, or should the recommendation only be to move the state up and resolve everything before rendering?

layershifter:
> Do you see a good solution in this case, or should the recommendation only be to move the state up and resolve everything before rendering?

Lift up state, with hooks it will be easier. It also will solve issues like re-positioning requirement.

packages/react/src/types.ts
```diff
@@ -71,7 +71,11 @@ type ReactNode =
   | null
   | undefined
 
-export type ShorthandValue<P> = ReactNode | Props<P>
+export type ShorthandRenderProp<P> = (Component: React.ElementType, props: P) => React.ReactNode
+
+export type ShorthandValue<P extends Props> =
+  | ReactNode
+  | Props<P> & { children?: P['children'] | ShorthandRenderProp<P> }
```
layershifter, line 79: @miroslavstastny I am not sure about this typing, any thoughts?

miroslavstastny: I like it 😃

miroslavstastny (approved):

PR #1955, opened 2019-09-20
docs(VirtualizedTree): left-arrow navigation to hidden parent item

silviuaavram:
Based on `onFocusParent` callback from item.

If parent node was not rendered inside the virtualized tree then focus by left arrow from children would not work.

Should not do anything if element is already visible. The default Tree implementation in `onFocusParent` should handle the job.


docs/src/prototypes/VirtualizedTree/VirtualizedTree.tsx
```diff
@@ -32,6 +45,13 @@ function TreeVirtualizer(props: TreeVirtualizerProps) {
       height={300}
       rowCount={props.renderedItems.length}
       width={600}
+      scrollToIndex={scrollToIndex}
+      onRowsRendered={() => {
+        if (scrollToIndex !== undefined) {
+          props.renderedItems[scrollToIndex].props.contentRef.current.focus()
+          setScrollToIndex(undefined)
```
silviuaavram, line 73: Unfortunate implementation but should not cause a DOM render. Any better ideas?

Need to un-set it after this operation because if left it in the state then at each onRowsRendered that item will always be scrolled to.

docs/src/prototypes/VirtualizedTree/VirtualizedTree.tsx
```diff
@@ -12,13 +12,35 @@ function TreeVirtualizer(props: TreeVirtualizerProps) {
...
 
   const rowRenderer = ({ index, isScrolling, key, parent, style }) => {
     const { renderedItems } = props
 
     return (
       <CellMeasurer cache={cache} columnIndex={0} key={key} parent={parent} rowIndex={index}>
-        {React.cloneElement(renderedItems[index], { style })}
+        {React.cloneElement(renderedItems[index], {
+          style,
+          onFocusParent: (e, treeItemProps: TreeItemProps) => {
```
lucivpav, line 50: Extract to function?

miroslavstastny (approved):

PR #1956, opened 2019-09-20
docs(mixed): fix knobs on CodeSandbox, display LogInspector always

layershifter:
### fix Prettier imports

Previously we read it from `window.prettierPlugins`, but they should be registered via `import`s to work properly on CodeSandbox.

### `LogInspector` was not visible

Fixes a small regression from #1940, `LogInspector` was not visible until `Try it` button was not pressed.

### Fixes knobs on CodeSandbox

We previously haven't `KnobProvider` in `SandboxApp` so when example contained `useKnob()` hooks it throws an error.

### Display knobs panel

image

As we started to use knobs more, we need to display them in exported examples too.


docs/src/components/ComponentDoc/ComponentExample/ComponentExample.tsx
```diff
@@ -553,12 +552,15 @@ class ComponentExample extends React.Component<ComponentExampleProps, ComponentE
...
-                {React.createElement(component)}
-              </Segment>
+              <>
+                <Segment
+                  className={`rendered-example ${this.getKebabExamplePath()}`}
+                  styles={exampleStyles}
+                >
+                  {React.createElement(component)}
+                </Segment>
+                <LogInspector silent />
```
layershifter, line 562: Previously was absent 😭

packages/code-sandbox/src/KnobsSnippet.tsx
```diff
@@ -36,8 +36,8 @@ const ComponentExampleKnobs = createComponent({
...
 })
 
-ComponentExampleKnobs.defaultProps = {
-  styles: componentExampleKnobsStyles,
+KnobsSnippet.defaultProps = {
+  styles: knobsSnippetStyles,
 }
 
-export default ComponentExampleKnobs
+export default KnobsSnippet
```
layershifter, line 43: Move to be reused

PR #1957, opened 2019-09-23
perf(renderComponent): avoid multiple style function calls

layershifter:
## 📖 Story around

We need to compute `classes` and `styles` to pass them to components and pass down to slots. However, we previously we did the same amount of work twice:

```tsx
// `mergedStyles` is an object with style functions
const resolvedStyles: ComponentSlotStylesPrepared = resolveStyles(mergedStyles, styleParam)
const classes: ComponentSlotClasses = getClasses(renderer, mergedStyles, styleParam)
```

`resolveStyles()` and `mergedStyles()` resolved styles twice and means that all style functions were called twice. With these changes `getClasses()` will use already resolved `styles` to avoid double calls.

## 🐎 Measures

```bash
cross-env PERF=true gulp perf --filter *Chat.perf.tsx --times=100
```

Performance diff is around **5.80%**.

### Before

| Example | Avg | Median |
| --- |  ---  |  ---  |
| Chat.perf.tsx | 482.15 | 480.24 |
| Chat.perf.tsx | 483.21 | 482.86 |
| Chat.perf.tsx | 484.41 | 483 |
| **Total** | **483.25** | |

### After

| Example | Avg | Median |
| --- |  ---  |  ---  |
| Chat.perf.tsx | 454.69 | 452.21 |
| Chat.perf.tsx | 455.29 | 453.38 |
| Chat.perf.tsx | 457.94 | 454.18 |
| **Total** |  **455.97**| |


CHANGELOG.md
```diff
@@ -38,11 +38,12 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
...
 - Add `notes` icon and update `link` icon in Teams theme @codepretty ([#1953](https://github.com/stardust-ui/react/pull/1953))
 
+### Performance
+- Resolve styles once in `renderComponent()` @layershifter ([#1957](https://github.com/stardust-ui/react/pull/1957))
+
 ### Documentation
 - Remove Usage tab @lucivpav ([#1948](https://github.com/stardust-ui/react/pull/1948))
 - Put props on a single page, fix props links @lucivpav ([#1892](https://github.com/stardust-ui/react/pull/1892))
-
-### Documentation
```
layershifter, line 46: Oops, duplicate

miroslavstastny (approved):
👍

Issue #1478, opened 2019-06-09
bodyBackground is incorrect in teams-dark theme


dslane:
### Steps
For teams-dark theme, bodyBackground is supposed to be grey[650], but the theme's siteVariables currently have it set to grey[800]. File with the issue is [here](packages/react/src/themes/teams-dark/siteVariables.ts)

### Expected Result
bodyBackground for teams-dark should be grey[650]

### Actual Result
bodyBackground for teams-dark should be grey[800], which results in the wrong color as the background throughout the teams-dark theme💣

### Version
0.32

### Testcase
[Fork, update, and replace this pen to show the bug]:
No testcase, as this is a simple issue of the wrong variable set in the theme.

codepretty:
This is still an issue and should be fixed

PR #1961, opened 2019-09-23
fix(theme): updating `bodyBackground` color for Teams dark theme

codepretty:
For reference, https://github.com/stardust-ui/react/issues/1478 and https://github.com/stardust-ui/react/issues/1870

**This will create a lot of screener regressions, so any help reviewing or thinking about unintended consequences would be helpful!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixes
+- Fix `bodyBackground` color for Teams dark theme to be the correct grey value @codepretty ([#1961](https://github.com/stardust-ui/react/pull/1961))
+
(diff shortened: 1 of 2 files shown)


jurokapsiar (approved):

levithomason:
@codepretty, I went through all 142 screener diffs ( 😅) and confirmed only the background color was affected.

This was almost always just the body color except for two items:
1. In the case of some menu examples the `ul` element also changed its background color.  I believe this is correct and intentional however as the Menu element takes on the body background color.
2. Some SVG elements seem to be triggering a diff, but the glyphs look identical to my eye.


Example diff showing these two issues can be seen [here](https://screener.io/v2/states/rkGqtccHX.stardust-ui-react/fix-body-background/1024x768/Chrome/menuexampleicononlyshorthandtsx-hovers-2nd-item-hover-state-styles-1).

I think it is safe to merge.

PR #1963, opened 2019-09-24
chore(eslint): disable browser globals

layershifter:
Also enable:
```
  "react-hooks/rules-of-hooks": "error",
    "react-hooks/exhaustive-deps": "warn"
```


layershifter:
Actually I am not sure that we need to disable globals for everything.
We have already disabled them for `packages/react` and I think that this is enough for now.

PR #1969, opened 2019-09-25
chore(downshift): update to 3.2.14

silviuaavram:
To pick the changes that fixed the a11y message announcements in custom window.


packages/react/package.json
```diff
@@ -11,7 +11,7 @@
     "@stardust-ui/react-component-ref": "^0.39.0",
     "@stardust-ui/react-proptypes": "^0.39.0",
     "classnames": "^2.2.5",
-    "downshift": "~3.2.10",
+    "downshift": "3.2.14",
```
layershifter, line 16:
```suggestion
    "downshift": "~3.2.14", // to match `3.2.*`
```
_or_
```suggestion
    "downshift": "^3.2.14", // to match `3.*.*`
```

We should not require exact version of libraries, until it is completely broken in other versions. If there will be other library that requires `downshift@^3.2.15` or `downshift@^3.3.0` it will end with bloated bundle for consumers.

layershifter (approved):

PR #1970, opened 2019-09-25
fix(mixed): avoid usages of global `document` and `window`

### LINKED ISSUE #1964: Remove usage of global document and window from components
## Bug Report

Following components still contain references to global document/window:
- Dropdown.tsx line 814
- Loader.tsx line 105
- Popup.tsx line 276
- Toolbar.tsx lines 182, 189, 191
- ToolbarRadioGroup.tsx line 103

Also, check focusUtilities, lines 430 and 440 which also use the globals, but just as a fallback value.

### Version
0.39.0

layershifter:
Fixes #1964.

This PR removes remaining usages of `document` and `window` globals in components. Also forces ESLint to catch such issues.

### DIFF
--- a/packages/react/.eslintrc.json
+++ b/packages/react/.eslintrc.json
@@ -6,6 +6,16 @@
       "rules": {
         "react/prop-types": "off"
       }
+    },
+    {
+      "files": "src/**/*.{ts,tsx}",
+      "globals": {
+        "document": "off",
+        "window": "off"
+      },
+      "rules": {
+        "no-undef": "error"
+      }
     }
   ],
   "root": true
--- a/packages/react/src/components/Popup/Popup.tsx
+++ b/packages/react/src/components/Popup/Popup.tsx
@@ -273,7 +273,9 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
     const lastContentRef = getRefs().pop()
     const isLastOpenedPopup: boolean =
       lastContentRef && lastContentRef.current === this.popupDomElement
-    const bodyHasFocus: boolean = document.activeElement === document.body
+
+    const activeDocument = this.props.mountDocument || this.context.target
+    const bodyHasFocus: boolean = activeDocument.activeElement === activeDocument.body
 
     if (keyCode === keyboardKey.Escape && bodyHasFocus && isLastOpenedPopup) {
       this.close(e, () => _.invoke(this.triggerFocusableDomElement, 'focus'))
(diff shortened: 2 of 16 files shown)


packages/react/.eslintrc.json
```diff
@@ -6,6 +6,16 @@
...
+    },
+    {
+      "files": "src/**/*.{ts,tsx}",
+      "globals": {
+        "document": "off",
+        "window": "off"
+      },
+      "rules": {
+        "no-undef": "error"
+      }
```
layershifter, line 18: Temporary solution until we will merge #1963.

packages/react/.eslintrc.json
```diff
@@ -6,6 +6,16 @@
       "rules": {
         "react/prop-types": "off"
       }
+    },
+    {
+      "files": "src/**/*.{ts,tsx}",
+      "globals": {
+        "document": "off",
+        "window": "off"
```
layershifter, line 14: It's reasonable to disable only `document` and `window` globals as they are context dependent.

packages/react/src/components/Popup/Popup.tsx
```diff
@@ -273,7 +273,9 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
     const lastContentRef = getRefs().pop()
     const isLastOpenedPopup: boolean =
       lastContentRef && lastContentRef.current === this.popupDomElement
-    const bodyHasFocus: boolean = document.activeElement === document.body
+
+    const activeDocument = this.props.mountDocument || this.context.target
```
layershifter, line 277: We need to consider remove `mountDocument` prop from `Popup` at all.

packages/react/src/components/Toolbar/Toolbar.tsx
```diff
@@ -179,16 +179,17 @@ class Toolbar extends UIComponent<WithAsProp<ToolbarProps>, ToolbarState> {
...
   }
 
   afterComponentRendered() {
     if (this.animationFrameId !== undefined) {
-      window.cancelAnimationFrame(this.animationFrameId)
+      cancelAnimationFrame(this.animationFrameId)
     }
-    this.animationFrameId = window.requestAnimationFrame(() => {
+
+    this.animationFrameId = requestAnimationFrame(() => {
```
layershifter, line 192: Discussed with @miroslavstastny, there are no reasons to use it from `window`

jurokapsiar (approved):

PR #1972, opened 2019-09-25
fix(whatInput): use proper `document` to set `data-whatinput` attr

### LINKED ISSUE #1971: :focus-visible is broken for child windows
### Steps
1. Open https://codesandbox.io/s/stardust-ui-example-kekk8
2. Click "Open window" button
3. Try to navigate between `Attachment` and `Button` via `Tab` key

### Expected Result
See focus indicator.

### Actual Result
Focus indicator is absent 💣

### Version
0.39.1

***

The root issue we don't have `data-*` attributes in child window:

image

The issue happens in `whatInput.ts`:
- in `addListeners` we use a proper `window` object: https://github.com/stardust-ui/react/blob/88d17b1578cf168a56e002da9c0d7708b48bef40/packages/react/src/lib/whatInput.ts#L102
- but in `doUpdate` we settings `data-whatinput` only on a parent: https://github.com/stardust-ui/react/blob/88d17b1578cf168a56e002da9c0d7708b48bef40/packages/react/src/lib/whatInput.ts#L155

layershifter:
Fixes #1971.

***

This PR fixes our fork of `what-input` to properly set `data-whatinput` attribute as our `:focus-visible` implementation dependents on it.

### DIFF
--- a/docs/src/examples/components/Provider/Usage/ProviderExampleTargetFrame.steps.ts
+++ b/docs/src/examples/components/Provider/Usage/ProviderExampleTargetFrame.steps.ts
@@ -1,0 +1,5 @@
+const config: ScreenerTestsConfig = {
+  steps: [builder => builder.focus('iframe').snapshot('Focuses button in iframe')],
+}
+
+export default config
--- a/docs/src/examples/components/Provider/Usage/ProviderExampleTargetFrame.steps.ts
+++ b/docs/src/examples/components/Provider/Usage/ProviderExampleTargetFrame.steps.ts
@@ -1,5 +1,0 @@
-const config: ScreenerTestsConfig = {
-  steps: [builder => builder.focus('iframe').snapshot('Focuses button in iframe')],
-}
-
-export default config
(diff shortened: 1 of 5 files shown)


packages/react/src/lib/whatInput.ts
```diff
@@ -145,37 +134,14 @@ const setInput = event => {
...
-    return
-  }
-
-  currentElement = event.target.nodeName.toLowerCase()
-  docElem.setAttribute('data-whatelement', currentElement)
-
-  if (event.target.classList && event.target.classList.length) {
-    docElem.setAttribute('data-whatclasses', event.target.classList.toString().replace(' ', ','))
-  }
-}
```
layershifter, line 172: We don't use `data-whatclasses` & `data-whatelement`, so it's safe to remove them

docs/src/examples/components/Provider/Usage/ProviderExampleTargetFrame.steps.ts
```diff
@@ -0,0 +1,7 @@
+const config: ScreenerTestsConfig = {
+  steps: [
+    (builder, keys) => builder.keys('body', keys.tab).snapshot('Focuses attachment in iframe'),
```
jurokapsiar: cool!

jurokapsiar (approved):

docs/src/examples/components/Provider/Usage/ProviderExampleTargetFrame.steps.ts
layershifter: Yep, it should be cool, but it's not working 😭 

As we need this fix I will merge this PR and then we can investigate how to move focus for Screener. At least, now we can be sure that styles are rendered.

PR #1978, opened 2019-09-27
docs(prototypes): remove MenuButton

layershifter:
We have a `MenuButton` component, there is no more need in this prototype.


mnajdova (approved):
👍

PR #1980, opened 2019-09-27
feat(bindings): add useAccessibility hook

layershifter:
## 🍴 `useAccessibility()`

This PR adds support for `useAccessibility()` hook that allows to use existing or created behavior in custom components.

###  👨‍💻 Example
```tsx
  const getA11Props = useAccessibility(imageBehavior, {
    mapPropsToBehavior: () => ({
      disabled: props.disabled
    }),
    actionHandlers: {
      click: (e: React.KeyboardEvent<HTMLDivElement>) => {
        if (onClick) onClick(e);
      }
    }
  });
```

- `mapPropsToBehavior()` - passes `props` to behavior, allows to pass props strictly, similar to `mapPropsToState()` in `useStateManager()`
- `actionHandlers` - works as existing `actionHandlers` in our components and allows to bind component actions to actions described in behaviors
- `getA11Props(slotName, slotProps)` - a function that applies accessibility props and performs proper merge with user's props

### 📡 POC

https://codesandbox.io/s/dropdown-prototype-r3ux9
(_prototype of `Dropdown` component that uses `useStateManager()` and `useAccessibility()`_)

***

# Questions

### 1️⃣ Refactor to `.root()` syntax

During our discussion @levithomason, he proposed more fancy signature for `getA11Props()`:

```tsx
getA11Props('root', someProps) // existing
getA11Props.root(someProps) // proposed
```

Key issue with this proposal that I don't understand there is еру replacement of behaviors. For example, I can replace `imageBehavior` with `emptyBehavior` without definitions for any slots. _How I will not that I need to create a `getA11Props.image()`?_

As solution, I can use `Proxy`, but it's not supported in IE11 💣 

### 2️⃣ `FocusZone`

There is no support in `useAccessibility()` for `FocusZone`, I propose to ship it later. Now, I am not fully sure around approaches and want to discuss them separately.


packages/react-bindings/src/hooks/useAccessibility.ts
```diff
@@ -0,0 +1,67 @@
+import { Accessibility, AccessibilityAttributesBySlot } from '@stardust-ui/accessibility'
+import * as React from 'react'
+
+import getAccessibility from '../accessibility/getAccessibility'
+import { AccessibilityBehavior, AccessibilityActionHandlers } from '../accessibility/types'
+
+type UseAccessibilityOptions<Props> = {
+  actionHandlers?: AccessibilityActionHandlers
+  debugName?: string
+  mapPropsToBehavior?: () => Props
```
jurokapsiar, line 10: can this be `Props | () => Props` for API simplification? looking at the unit tests I *think* passing an object there should be sufficient for most use cases.

packages/react-bindings/src/accessibility/shouldHandleOnKeys.ts
```diff
@@ -1,6 +1,6 @@
 import { KeyCombinations } from '@stardust-ui/accessibility'
+// @ts-ignore
```
layershifter, line 2: There are no typings for `keyboard-key` 😭

packages/react-bindings/src/accessibility/getKeyDownHandlers.ts
```diff
@@ -14,31 +14,32 @@ const rtlKeyMap = {
...
-    const componentPartKeyAction = behaviorKeyActions[componentPart]
-    const handledActions = _.intersection(
-      _.keys(componentPartKeyAction),
-      _.keys(componentActionHandlers),
+  for (const slotName in behaviorActions) {
+    const behaviorSlotAction = behaviorActions[slotName]
+    const handledActions = Object.keys(behaviorSlotAction).filter(
+      actionName => componentHandlerNames.indexOf(actionName) !== -1,
     )
+
```
layershifter, line 42: I refactored this place to avoid Lodash usage in our core package, tests are passing.

I also renamed variables to improve their meaning.

packages/react-bindings/src/hooks/useAccessibility.ts
layershifter: I see that you mean, but I want to keep it obvious for now and in sync with `useStateManager()`. We can improve this part once we will receive feedback for this thing.

packages/react-bindings/README.md
```diff
@@ -29,13 +31,70 @@ yarn add @stardust-ui/react-bindings
 
 # Hooks
 
+## `useAccesibility()`
+
+A React hook that provides bindings for accessibility behaviors.
+
+#### Usage
+
+The example below assumes a component called `<Input>` will be used this way:
```
mnajdova:
```suggestion
The example below assumes a component called `<Image>` will be used this way:
```

Isn't the example for `Image`? Also, not sure we are providing some example of how it is used, it's rather how it is defined

layershifter: > Also, not sure we are providing some example of how it is used, it's rather how it is defined

Sorry, don't understand fully, how to use this `Image` component?

ecraig12345:
Hi, just a heads up that I made a change #2153 renaming all `lib` folders to `utils`. So the changes to `lib` files in this PR will need to be moved to the new location.

packages/react-bindings/src/hooks/useAccessibility.ts
mnajdova: I agree that `Props | () => Props` maybe make sense, in most of the cases I see this as a simple props mapping.. Anyways, not a blocker :)

packages/react/src/utils/renderComponent.tsx
```diff
@@ -321,7 +217,32 @@ const renderComponent = <P extends {}>(
     }
   }
 
-  return result
+  if (accessibility.focusZone && accessibility.focusZone.mode === FocusZoneMode.Wrap) {
```
mnajdova, line 220: Can we extract this in a separate function

packages/react-bindings/src/hooks/useAccessibility.ts
```diff
@@ -0,0 +1,77 @@
...
+): MergedProps<SlotProps> => {
+  const finalProps: MergedProps<SlotProps> = {
+    ...definition.attributes[slotName],
+    ...slotProps,
+  }
+  const slotHandlers = definition.keyHandlers[slotName]
+
+  if (slotHandlers) {
+    const onKeyDown = (e: React.KeyboardEvent, ...args: any[]) => {
+      const keyHandlers = definition.keyHandlers[slotName]
```
mnajdova:
```suggestion
      const keyHandlers = slotHandlers
```

layershifter:
### Points of confusion / Parts that were blamed

- `getA11Props()` - it's not clear that this function is responsible for merging accessibility and client props
- `mapPropsToBehavior()` - should it be just an object? How to handle overrides case?
- `childBehaviors` - are not implemented as there is a strict contract from child components (accept `accessibility` prop)

mnajdova (approved):
We reviewed and agreed that there are parts that can be improved (see  https://github.com/microsoft/fluent-ui-react/pull/1980#issuecomment-572064465), but it can be merged as is now, and we'll iterate later on them. Good job @layershifter!

PR #1981, opened 2019-09-27
perf(tooltip): Simplify rendering when tooltip is not visible

jurokapsiar:
Based on the perf measurements, I found out that adding tooltips to the toolbar items has significant perf impact (~20% on my machine). This PR simplifies the rendering if the tooltip is not visible (it has to be in the DOM for accessibility purposes). The change seems to get back half of the degradation, see the perf measurements below.

Added CustomToolbar.perf.tsx to do the measurements.

Before:

| --- |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |
| CustomToolbar.perf.tsx | 119.07 | **119.07** | 63.41 | **63.41** | 108.14 | **108.14** | 60.48 | **60.48** |

After:

| --- |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |  ---  |
| CustomToolbar.perf.tsx | 97.46 | **97.46** | 52.73 | **52.73** | 88.64 | **88.64** | 50.2 | **50.2** |


docs/src/examples/components/Chat/Performance/CustomToolbar.perf.tsx
```diff
@@ -0,0 +1,567 @@
+import * as _ from 'lodash'
```
jurokapsiar: TODO: move this to Toolbar/Performance

packages/react/src/themes/teams/components/Tooltip/tooltipStyles.ts
```diff
@@ -5,14 +5,17 @@ import { TooltipVariables } from './tooltipVariables'
...
+    p.open
+      ? {
+          zIndex: v.zIndex,
+          position: 'absolute',
+          textAlign: 'left',
+        }
+      : {
+          opacity: 0,
+          position: 'absolute',
+        },
```
layershifter: I think that this change can be reverted.

layershifter (approved):

PR #1982, opened 2019-09-30
WIP: try caching component variables in theme

jurokapsiar:
Looks like a lot of time is spent just by evaluating variables in each render. Can we instead cache the resolved component variables in the theme, only evaluate props.variables during render and deep merge them?

|--|--|--|--|--|--|--|--|--|
| CustomToolbar.perf.tsx (before) | 98.09 | **98.09** | 41.45 | **41.45** | 98.02 | **98.02** | 41.31 | **41.31** |
| CustomToolbar.perf.tsx (after) | 71.35 | **71.35** | 29.92 | **29.92** | 69.98 | **69.98** | 30.13 | **30.13** |

### DIFF
--- a/packages/react/src/lib/mergeThemes.ts
+++ b/packages/react/src/lib/mergeThemes.ts
@@ -27,6 +27,7 @@ export const emptyTheme: ThemePrepared = {
     fontSizes: {},
   },
   componentVariables: {},
+  resolvedComponentVariables: {},
   componentStyles: {},
   fontFaces: [],
   staticStyles: [],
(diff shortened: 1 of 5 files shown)


packages/react/src/lib/mergeThemes.ts
```diff
@@ -53,8 +54,17 @@ export const mergeComponentStyles = (
       const originalTarget = partStylesPrepared[partName]
       const originalSource = partStyle
 
-      partStylesPrepared[partName] = styleParam => {
-        return _.merge(callable(originalTarget)(styleParam), callable(originalSource)(styleParam))
```
jurokapsiar, line 57: this is causing some merging of styles with null, creating unnecessary instances. the perf improvement on the custom toolbar was around 5%

miroslavstastny:
done in #2041

PR #1984, opened 2019-09-30
feat(ToolbarMenuItem): add menu prop

mnajdova:
Added menu prop on the `ToolbarMenuItem` for allowing creating submenus in the Toolbar. It behaves the same way as the `Menu`'s submenus - the implementation is a bit different, because we need to support popups inside the `ToolbarMenuItem` too.


packages/react/src/themes/teams/components/Toolbar/toolbarMenuItemStyles.ts
```diff
@@ -48,6 +48,12 @@ const toolbarMenuItemStyles = {
     right: pxToRem(7),
   }),
 
+  submenuIndicator: ({ variables: v }): ICSSInJSStyle => ({
```
mnajdova, line 51: Can this be defined at the same time with active? Should we handle if both of the icons are visible?

packages/react/src/lib/accessibility/Behaviors/Toolbar/toolbarMenuItemBehavior.ts
```diff
@@ -0,0 +1,79 @@
+import { Accessibility, AccessibilityAttributes } from '../../types'
+import { IS_FOCUSABLE_ATTRIBUTE } from '../../FocusZone/focusUtilities'
+import * as keyboardKey from 'keyboard-key'
+// TODO: extend menuItemBehavior and ignore vertical
```
mnajdova: Do we have a pattern of how we are extending other behaviors? This is basically the same as the menuItemBehavior, it's different only in a way that it does not have vertical prop (but behave as if it has as menus in toolbar are always vertical) @jurokapsiar @kolaps33

packages/react/src/themes/teams/components/Toolbar/toolbarMenuItemStyles.ts
jurokapsiar: that would mean a checkbox/radio menu item would have a submenu - which is not valid.

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -255,6 +255,7 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>> {
       return Popup.create(popup, {
         defaultProps: {
           trapFocus: true,
+          inline: true,
```
jurokapsiar: As the best practices say, "Don't use trapFocus for inline popup, as it leads to broken behavior for screen reader users."

mnajdova: Fixed, using not inlined Popup, and the behavor is as it was before.

jurokapsiar (approved):

packages/react/src/components/Toolbar/ToolbarMenu.tsx
```diff
@@ -50,18 +70,30 @@ class ToolbarMenu extends UIComponent<ToolbarMenuProps> {
   static propTypes = {
     ...commonPropTypes.createCommon(),
     items: customPropTypes.collectionShorthandWithKindProp(['divider', 'item']),
+    // onClick: PropTypes.func,
```
kolaps33: maybe leftover? :)

kolaps33 (approved):
Tested with readers JAWS/NVDA and works great as expected.

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
```diff
@@ -186,21 +378,40 @@ class ToolbarMenuItem extends UIComponent<WithAsProp<ToolbarMenuItemProps>> {
...
+            {maybeSubmenu}
+          </>
+        ),
       }),
     })
   }
 
-  handleClick = (e: React.SyntheticEvent) => {
-    const { disabled } = this.props
+  handleClick = (e: MouseEvent) => {
```
layershifter:
```suggestion
  handleClick = (e: React.MouseEvent) => {
```

It's an React event

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
```diff
@@ -176,7 +338,37 @@ class ToolbarMenuItem extends UIComponent<WithAsProp<ToolbarMenuItemProps>> {
...
-    const menuItemInner = hasChildren ? children : elementType
+    const menuItemInner = hasChildren ? children : <Ref innerRef={this.itemRef}>{elementType}</Ref>
+
+    const maybeSubmenu =
+      menu && menuOpen ? (
+        <Unstable_NestingAuto>
+          {(getRefs, nestingRef) => (
+            <>
+              <Ref innerRef={nestingRef}>
+                <Ref innerRef={this.menuRef}>
```
layershifter: Oops, you don't need two `Ref` components:

```tsx
<Ref
  innerRef={(node: HTMLElement) => {
    nestingRef.current = node
    this.menuRef.current = node
  }}
/>
```

To make it assignable on `this.menuRef`:

```tsx
menuRef = React.createRef<HTMLElement>() as React.MutableRefObject<HTMLElement>
```

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
```diff
@@ -111,29 +158,134 @@ class ToolbarMenuItem extends UIComponent<WithAsProp<ToolbarMenuItemProps>> {
...
+    const { menu } = this.props
+    const { menuOpen } = this.state
+    if (menu && !menuOpen) {
+      this.trySetMenuOpen(true, e)
+      e.stopPropagation()
+      e.preventDefault()
+    }
+  }
+
+  closeMenu = (e: Event) => {
```
layershifter:
```suggestion
  closeMenu = (e: React.KeyboardEvent) => {
```

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
```diff
@@ -111,29 +158,134 @@ class ToolbarMenuItem extends UIComponent<WithAsProp<ToolbarMenuItemProps>> {
...
+    closeMenu: event => this.closeMenu(event),
+    closeMenuAndFocusTrigger: event => this.closeMenu(event),
+    doNotNavigateNextParentItem: event => {
+      event.stopPropagation()
+    },
+    closeAllMenus: event => this.closeAllMenus(event),
   }
 
-  renderComponent({ ElementType, classes, accessibility, unhandledProps, styles }) {
+  openMenu = (e: Event) => {
```
layershifter: openMenu = (e: React.KeyboardEvent) => {

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
```diff
@@ -111,29 +158,134 @@ class ToolbarMenuItem extends UIComponent<WithAsProp<ToolbarMenuItemProps>> {
...
+    const { menu } = this.props
+    const { menuOpen } = this.state
+    if (menu && !menuOpen) {
```
layershifter, line 190:
```suggestion
    if (!this.isSubmenuOpen()) {
```

?

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -310,6 +321,18 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>> {
     _.invoke(this.props, 'onClick', e, this.props)
   }
 
+  handleWrapperClick = e => {
+    const { menu } = this.props
+    if (menu) {
+      if (doesNodeContainClick(this.menuRef.current, e, this.context.target)) {
+        this.trySetMenuOpen(false, e)
+        if (this.itemRef) {
```
layershifter:
```suggestion
        if (this.itemRef.current) {
```

This condition is wrong as `itemRef` will be always object.
***
We can use use also `_.invoke()`: `_.invoke(this.itemRef.current, 'focus')`

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -221,20 +226,22 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>> {
...
-                    escapeWithReference: false, // escapeWithReference breaks positioning of ToolbarMenu in overflow mode because Popper components sets modifiers on scrollable container
-                  },
-                }}
-                targetRef={this.itemRef}
-              >
-                {ToolbarMenu.create(menu, {
-                  overrideProps: this.handleMenuOverrides(getRefs, variables),
-                })}
-              </Popper>
+              <Ref innerRef={this.menuRef}>
```
layershifter: https://github.com/microsoft/fluent-ui-react/pull/1984/files#r336946706

Same thing.

packages/accessibility/src/behaviors/Toolbar/toolbarMenuItemBehavior.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+const toolbarMenuItemBehavior: Accessibility<ToolbarMenuItemBehaviorProps> = props => {
+  return menuItemBehavior({ ...props, vertical: true })
+}
+
+export default toolbarMenuItemBehavior
+
+export type ToolbarMenuItemBehaviorProps = {
+  /** Indicated if menu item has submenu. */
+  menu?: boolean | object
```
layershifter, line 33: 😢 

Can we replace with `hasMenu`?

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -184,14 +188,15 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>> {
...
       _.invoke(predefinedProps, 'onItemClick', e, itemProps)
-      if (itemProps.popup) {
+      if (popup) {
         return
       }
       // TODO: should we pass toolbarMenuItem to the user callback so he can decide if he wants to close the menu?
-      this.trySetMenuOpen(false, e)
-      if (this.itemRef) {
+      this.trySetMenuOpen(menuOpen, e)
+      if (!menuOpen && this.itemRef) {
```
layershifter: https://github.com/microsoft/fluent-ui-react/pull/1984/files#r336950273

packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -161,6 +161,9 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>> {
       event.preventDefault()
       this.handleClick(event)
     },
+    performWrapperClick: event => {
+      this.handleWrapperClick(event)
+    },
     closeMenuAndFocusTrigger: event => {
       this.trySetMenuOpen(false, event)
       if (this.itemRef) {
```
layershifter: https://github.com/microsoft/fluent-ui-react/pull/1984/files#r336950273

docs/src/examples/components/Toolbar/Content/ToolbarExampleMenuWithSubmenu.shorthand.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+  return (
+    <Toolbar
+      items={[
+        {
+          key: 'more',
+          icon: 'more',
+          active: menuOpen,
+          menu: {
+            items: [
```
layershifter: Can be just `menu`:

```tsx
menu: [ /* ... */ ],
```

docs/src/examples/components/Toolbar/Content/ToolbarExampleMenuWithSubmenu.shorthand.tsx
```diff
@@ -0,0 +1,48 @@
+import { createCallbackLogFormatter } from '@stardust-ui/code-sandbox'
+import { useLogKnob } from '@stardust-ui/docs-components'
+import { Toolbar } from '@stardust-ui/react'
+import * as React from 'react'
+
+const ToolbarExampleMenuShorthand = () => {
```
layershifter: Please update example's name.

layershifter:
We also need some E2E to be sure that this feature will not break.

packages/react/src/components/Toolbar/ToolbarMenuItem.tsx
mnajdova: `!(menu && menuOpen)` is not the same with  `menu && !menuOpen`

packages/accessibility/src/behaviors/Toolbar/toolbarMenuItemBehavior.ts
mnajdova: This is a prop from the component propagated. It is aligned as the Menu's menuItemBehavior is created. If we decide to change these would rather change them consistently everywhere.

PR #1992, opened 2019-10-01
chore(docs): clean up miscellaneous component documentation

dvdzkwsk:
As I've been working on the new Stardust docs I've encountered various typos, grammatical errors, and tense/wording inconsistencies. Since the new docs are using the same exact schemas the current ones do, I figured I'd share the changes sooner than later :).

For wording changes, I tried to keep as true to the original intent as possible while cleaning up the language a bit. If I've gone astray with any of the changes, please let me know!

### DIFF
--- a/packages/react/src/components/Accordion/Accordion.tsx
+++ b/packages/react/src/components/Accordion/Accordion.tsx
@@ -37,7 +37,7 @@ export interface AccordionProps extends UIComponentProps, ChildrenComponentProps
   /** Initial activeIndex value. */
   defaultActiveIndex?: number[] | number
 
-  /** Only allow one panel open at a time. */
+  /** Only allow one panel to be expanded at a time. */
   exclusive?: boolean
 
   /** At least one panel should be expanded at any time. */
--- a/packages/react/src/components/Alert/Alert.tsx
+++ b/packages/react/src/components/Alert/Alert.tsx
@@ -45,7 +45,7 @@ export interface AlertProps
    */
   accessibility?: Accessibility
 
-  /** An Alert can contain action buttons. */
+  /** An alert can contain action buttons. */
   actions?: ShorthandValue<ButtonGroupProps> | ShorthandCollection<ButtonProps>
 
   /** An alert may contain an icon. */
@@ -54,7 +54,7 @@ export interface AlertProps
   /** An alert may contain a header. */
   header?: ShorthandValue<TextProps>
 
-  /** Controls Alert's relation to neighboring items. */
+  /** An alert's position relative to neighboring items. */
   attached?: boolean | 'top' | 'bottom'
 
   /** An alert can only take up the width of its content. */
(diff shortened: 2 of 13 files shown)


packages/react/src/components/Alert/Alert.tsx
```diff
@@ -55,7 +55,7 @@ export interface AlertProps
   /** An alert may contain a header. */
   header?: ShorthandValue<TextProps>
 
-  /** Controls Alert's relation to neighboring items. */
+  /** An alert's position relative to neighboring items. */
```
davezuko: We should standardize on capitalization conventions across all components, but for now I think it's sufficient to make them at least consistent within a single component.

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -87,22 +87,22 @@ export interface AlertProps
...
   /**
-   * Called after user's focus.
+   * Called after the alert is focused.
    * @param {SyntheticEvent} event - React's original SyntheticEvent.
    * @param {object} data - All props.
    */
   onFocus?: ComponentEventHandler<AlertProps>
 
-  /** An alert may be formatted to display a successful message. */
+  /** An alert can be formatted to display a successful message. */
```
dvdzkwsk, line 95: Minor consistency change: "may" and "can" were used interchangeably, so I just picked one.

packages/react/src/components/Alert/Alert.tsx
```diff
@@ -46,7 +46,7 @@ export interface AlertProps
    */
   accessibility?: Accessibility
 
-  /** An Alert can contain action buttons. */
+  /** An alert can contain action buttons. */
```
dvdzkwsk, line 48: We should standardize on capitalization conventions across all components, but for now I think it's sufficient to ensure they are consistent within a single component. In this case, most were already lowercase.

packages/react/src/components/Animation/Animation.tsx
```diff
@@ -24,42 +24,42 @@ export interface AnimationProps
   /** The name for the animation that should be applied, defined in the theme. */
   name?: string
 
-  /** The delay property specifies a delay for the start of an animation. Negative values are
+  /** Specifies a delay for the start of an animation. Negative values are
    * also allowed. If using negative values, the animation will start as if it had already been
-   * playing for N seconds.
+   * playing for that amount of time.
```
dvdzkwsk, line 29: Since this is a string, it's not necessarily in seconds.

packages/react/src/components/Accordion/Accordion.tsx
```diff
@@ -38,7 +38,7 @@ export interface AccordionProps extends UIComponentProps, ChildrenComponentProps
   /** Initial activeIndex value. */
   defaultActiveIndex?: number[] | number
 
-  /** Only allow one panel open at a time. */
+  /** Only allow one panel to be expanded at a time. */
```
dvdzkwsk, line 40: Consistency: Accordion uses the word `expanded` to describe its panels' open states, and that term is also already used in describing other props.

packages/react/src/components/Animation/Animation.tsx
```diff
@@ -24,42 +24,42 @@ export interface AnimationProps
   /** The name for the animation that should be applied, defined in the theme. */
   name?: string
 
-  /** The delay property specifies a delay for the start of an animation. Negative values are
```
dvdzkwsk, line 27: Most components don't reiterate the name of the property in the doc string, which I think is a good thing. I've removed it from Animation to make it consistent with the rest of the components.

packages/react/src/components/Checkbox/Checkbox.tsx
```diff
@@ -28,39 +28,39 @@ export interface CheckboxProps extends UIComponentProps, ChildrenComponentProps
   /** Accessibility behavior if overridden by the user. */
   accessibility?: Accessibility
 
-  /** Initial checked value. */
+  /** A checkbox can be checked by default. */
   defaultChecked?: SupportedIntrinsicInputProps['defaultChecked']
 
-  /** Whether or not item is checked. */
```
dvdzkwsk, line 33: I believe `checkbox` is meant by `item` here.

jurokapsiar (approved):

dvdzkwsk:
DangerJS improperly flagging a whitespace fix in the Changelog. Ignoring.

PR #2001, opened 2019-10-02
fix(Chat): adding border for reaction popup in Teams high contrast theme

codepretty:
**Before**
image

**After**
image


dvdzkwsk (approved):

packages/react/src/themes/teams/components/Chat/chatMessageStyles.ts
```diff
@@ -84,6 +84,8 @@ const chatMessageStyles: ComponentSlotStylesPrepared<
 
   actionMenu: ({ props: p, variables: v }): ICSSInJSStyle => ({
     backgroundColor: v.backgroundColor,
+    border: '1px solid',
+    borderColor: v.reactionGroupBorder,
```
layershifter: It should be:

```suggestion
    borderColor: v.reactionGroupBorderColor,
```

Or

```suggestion
    border: v.reactionGroupBorder,
```

layershifter (commented):
LGTM, added one small comment

PR #2003, opened 2019-10-03
chore: simplify getUnhandledProps() & getElementType()

layershifter:
Based on findings from #1355.

This PR simplifies our `getUnhandledProps()` and `getElementType()` helper functions that makes them more reusable.

I also moved them to `@stardust-ui/react-bindings` and improved types on them. Adds a missing UT for `getElementType()`.


packages/react/src/lib/getElementType.tsx
```diff
@@ -1,26 +0,0 @@
...
- * @param {object} props A ReactElement props object
- * @returns {string|function} A ReactElement type
- */
-function getElementType(Component: { defaultProps?: Props }, props: Props): React.ElementType {
-  const { defaultProps = {} } = Component
-
-  // ----------------------------------------
-  // user defined "as" element type
-
-  if (props.as && props.as !== defaultProps.as) return props.as
```
layershifter, line 18: After #1396 there is no sense in this condition.
We want to use `props.as` or `div`. In the case when `props.as` is not defined it will come from `defaultProps.as`, it's handled by React.

packages/react/src/lib/renderComponent.tsx
```diff
@@ -205,7 +202,7 @@ const renderComponent = <P extends {}>(
     rtl,
   )
 
-  const unhandledProps = getUnhandledProps({ handledProps }, props)
+  const unhandledProps = getUnhandledProps(handledProps, props)
```
layershifter, line 205: It was strange to pass `handledProps` in object when we can just pass them.

packages/react/src/lib/accessibility/FocusZone/AutoFocusZone.tsx
```diff
@@ -19,19 +18,16 @@ export default class AutoFocusZone extends React.Component<AutoFocusZoneProps> {
     firstFocusableSelector: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),
   }
 
-  static handledProps = _.keys(AutoFocusZone.propTypes)
+  static handledProps = _.keys(AutoFocusZone.propTypes) as any
```
layershifter, line 21: 💣 

`_.keys()` and `Object.keys()` returns `string[]` in TS:

image

So type cast is required is there as typings for `getUnhandledProps()` are more strict.

jurokapsiar (approved):

PR #2004, opened 2019-10-03
chore: move FocusZone components to react-bindings

layershifter:
### ↩️ Move of `FocusZone`

This PR moves `FocusZone`, `AutoFocusZone`, `FocusTrapZone`, related types and utils to `@stardust-ui/react-bindings` to make them reusable.

The main motivation for this is too avoid circular dependency: `@stardust-ui/react-bindings` => `@stardust-ui/react` => `@stardust-ui/react-bindings`


packages/react-bindings/package.json
```diff
@@ -5,7 +5,9 @@
   "author": "Oleksandr Fediashov <a@fedyashov.com>",
   "bugs": "https://github.com/stardust-ui/react/issues",
   "dependencies": {
-    "@babel/runtime": "^7.1.2"
+    "@babel/runtime": "^7.1.2",
+    "@stardust-ui/accessibility": "^0.39.0",
+    "@stardust-ui/state": "^0.39.0"
```
layershifter, line 10: These packages can't be in optional dependencies because TS will throw on typings...

packages/react-bindings/src/hooks/useStateManager.ts
```diff
@@ -13,9 +7,7 @@ type UseStateManagerOptions<State> = {
   sideEffects?: SideEffect<State>[]
 }
 
-export const getDefinedProps = <Props extends Record<string, any>>(
-  props: Props,
-): Partial<Props> => {
+const getDefinedProps = <Props extends Record<string, any>>(props: Props): Partial<Props> => {
```
layershifter, line 10: Should not be exported, nit

packages/react-bindings/src/utils/callable.ts
```diff
@@ -1,4 +1,3 @@
-// https://jsperf.com/startdust-callable
```
layershifter, line 1: Link is dead, 404

packages/react/test/specs/lib/factories-test.tsx
```diff
@@ -1,10 +1,10 @@
+import { callable } from '@stardust-ui/react-bindings'
```
layershifter, line 1: I also have to move `callable` as it used in FZ

jurokapsiar (approved):

PR #2006, opened 2019-10-03
fix(Dropdown): colliding toggle button and scroll

jurokapsiar:
Before:
image

After: 
image

### DIFF
--- a/packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
+++ b/packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
@@ -106,7 +106,7 @@ const dropdownStyles: ComponentSlotStylesPrepared<DropdownPropsAndState, Dropdow
     overflowY: 'auto',
     maxHeight: v.selectedItemsMaxHeight,
     width: '100%',
-    ...(p.toggleIndicator && { paddingRight: v.toggleIndicatorSize }),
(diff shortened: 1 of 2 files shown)


packages/react/src/themes/teams/components/Dropdown/dropdownStyles.ts
```diff
@@ -106,7 +106,7 @@ const dropdownStyles: ComponentSlotStylesPrepared<DropdownPropsAndState, Dropdow
     overflowY: 'auto',
     maxHeight: v.selectedItemsMaxHeight,
     width: '100%',
-    ...(p.toggleIndicator && { paddingRight: v.toggleIndicatorSize }),
+    ...(p.toggleIndicator && { marginRight: v.toggleIndicatorSize }),
```
jurokapsiar, line 109: this actually breaks IE in a strange way. Follwoing code added to the toggle indicator slot works better:
```
 root: {
   [`& .${Dropdown.slotClassNames.toggleIndicator}`]: {
     marginRight: "12px"
   }
  }
```

codepretty: @jurokapsiar what is the strange behavior?

codepretty: It looks like the new styling moves the scrollbar for all browsers and not just IE11. I don't think that is what we will want, but i can't load in ie11 right nowso i don't understand the issue...

PR #2010, opened 2019-10-04
feat(Toolbar): overflow menu via DOM

### LINKED ISSUE #1885: Toolbar: overflow is not working in child window
### Steps
1. Open attached CodeSandbox
2. Click on `Open Window` button
3. Notice that `Toolbar` looks properly
4. Resize child window

### Expected Result
Resize occurred.

### Actual Result
Nothing happened.

image

### Version
0.37.0

### Testcase
https://codesandbox.io/s/stardust-ui-example-sqh81

### LINKED ISSUE #1893: Toolbar overflow performance
## Bug Report
Toolbar overflow menu calculations seem to be non-performant. There is a noticeable lag in the UI when Toolbar is resized or first rendered

### Steps
Use this [codesandbox](https://codesandbox.io/s/compose-overflow-toolbar-xjnv1) for repro.

1. Click Show Toolbar button repeatedly
Notice the delay rendering the toolbar

1.a Resize window
Notice the delay in re-rendering the buttons

### Expected Result
Toolbar rendering should be smooth

### Actual Result
delay rendering the toolbar

### Version
0.37.0

### Testcase
https://codesandbox.io/s/compose-overflow-toolbar-xjnv1

### LINKED ISSUE #1935: Toolbar overflow prop
### Problem description
Currently, when one wants to use overflow toolbar, they have to write [a lot of code](https://stardust-ui.github.io/react/components/toolbar/definition#content-overflow) in `onReduceItems`. 

### Proposed solution
Making the toolbar support overflow functionality should be as simple as adding an `overflow` prop. Or even making `Toolbar` support `overflow` by default

miroslavstastny:
## BREAKING CHANGE

Legacy support for Toolbar overflow using `onReduceItems` has been removed.

Refactor to the new `overflow` implementation for supported scenarios or reimplement the logic on client side for use cases which are not covered.

## Motivation (old approach problems)
`onReduceItems` implementation was generic allowing number of edge-case scenarios:
- it was possible to render custom overflow item anywhere in the list, the overflow item could change its size
- it was possible to move items to overflow menu in any order, or even multiple items at a time
The problem with the approach was that to show Toolbar it was necessary to render it multiple times which took significant amount of time (hundreds of ms).

## New approach
The new implementation takes vanilla DOM/CSS approach.
1. All toolbar items including the overflow menu are rendered inside an additional overflow container with `overflow: hidden`.
2. After render (or window resize) all direct DOM children of the overflow container are measured. If some children do not fit:
    - these items are hidden.
    - overflow menu is absolutely positioned and displayed.
    - (this is done using CSS manipulation by applying inline styles, no React rerender)
3. When Toolbar is rerendered with overflow menu open, it calls a callback back to to its parent (with index of the first overflow item). Parent returns list of overflow items in a format suitable for a menu.

### New approach limitations
- Consumer cannot define own reduce logic.
- Each toolbar item must render exactly one root DOM element (10 toolbar items must render 10 DOM elements in the overflow container).
- The root DOM element must have the dimensions representing the size of the item.
- Any toolbar item CANNOT resize itself - To resize a toolbar item, rerender the whole `Toolbar`
- Overflow item cannot hold its own state to open/close the overflow menu. The technical reason for that is that to open the overflow menu, the whole Toolbar must be rerendered to remove overflow items from the main area and render them inside the overflow. Symmetrically, when closing the menu, the Toolbar must be rerendered to render all items directly in the main area (root DOM element). 
- There is no support for overflow item to be always displayed (can be added later)
- There is no support for items being rendered always in the overflow menu (can be added later)

### Known issues
- `ToolbarItem` with popup set to inline breaks the contract as it renders two sibling elements inside the root DOM element.
- Overflow item can be displayed out of its container during resize.
- Overflow item's margins are ignored in the computation (to be fixed later).
- Overflow menu can be incorrectly positioned in Portal window (general Stardust issue with Popper).

Fixes #1935, fixes #1893, fixes #1885.


packages/react/src/components/Toolbar/ToolbarItem.tsx
```diff
@@ -218,7 +218,16 @@ class ToolbarItem extends UIComponent<WithAsProp<ToolbarItemProps>> {
...
             <Ref innerRef={nestingRef}>
-              <Popper align="start" position="above" targetRef={this.itemRef}>
+              <Popper
+                align="start"
+                position="above"
+                modifiers={{
+                  preventOverflow: {
+                    escapeWithReference: false,
+                  },
+                }}
```
layershifter, line 231: Let's leave these a comment why it's need, something like: `escapeWithReference breaks positioning of ToolbarMenu in overflow mode because Popper components sets modifiers on scrollable container`

packages/react/src/components/Toolbar/ToolbarOverflowMenu.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import { UIComponent } from '../../lib'
+import { withSafeTypeForAs } from '../../types'
+import ToolbarItem from './ToolbarItem'
+
+export interface ToolbarOverflowMenuProps {
+  getOverflowItems: any
+}
+
+class ToolbarOverflowMenu extends UIComponent<ToolbarOverflowMenuProps, any> {
```
layershifter: Is it `ToolbarOverflowMenuItem`?

packages/react/src/themes/teams/components/Toolbar/toolbarStyles.ts
```diff
@@ -1,18 +1,24 @@
 import { ComponentSlotStylesPrepared, ICSSInJSStyle } from '../../../types'
+import { ToolbarVariables } from './toolbarVariables'
+import { ToolbarProps } from '../../../../components/Toolbar/Toolbar'
 
-const toolbarStyles: ComponentSlotStylesPrepared = {
-  root: (): ICSSInJSStyle => ({
+const toolbarStyles: ComponentSlotStylesPrepared<ToolbarProps, ToolbarVariables> = {
+  root: ({ props: { overflow }, variables: v }): ICSSInJSStyle => ({
```
layershifter:
```suggestion
  root: ({ variables: v }): ICSSInJSStyle => ({
```

nit, I don't see that `overflow` is used anywhere

packages/react/src/themes/teams/components/Toolbar/toolbarMenuStyles.ts
```diff
@@ -6,6 +6,7 @@ const toolbarMenuStyles = {
     const colors = getColorScheme(v.colorScheme)
 
     return {
+      position: 'absolute',
```
layershifter: nit, but I think that `Popper` sets it, or no?

docs/src/examples/components/Toolbar/Content/ToolbarExampleOverflow.shorthand.tsx
```diff
@@ -1,62 +1,123 @@
...
+        <Flex>
+          <Toolbar
+            overflow
+            styles={{ minWidth: 0, flexGrow: 1 }}
+            items={bold ? toolbarItems : toolbarItems.filter(i => i['icon'] !== 'bold')}
+          />
+          <Toolbar items={[{ key: 'trash', icon: 'trash-can' }]} />
+        </Flex>
+      )}
+    </>
```
layershifter: This example should be somehow simplified and reworked

packages/react/src/components/Toolbar/Toolbar.tsx
```diff
@@ -118,189 +97,348 @@ class Toolbar extends UIComponent<WithAsProp<ToolbarProps>, ToolbarState> {
...
 
-    return {
-      initialItems: props.items,
-      currentItems: props.items,
-      stableItems: state.stableItems,
-      stable: false,
+    el.style.visibility = 'hidden'
+    const wasFocusable = el.getAttribute('data-is-focusable')
+    if (wasFocusable) {
+      el.setAttribute('data-sd-was-focusable', wasFocusable)
```
layershifter:
```suggestion
      el.setAttribute('data-was-focusable', wasFocusable)
```
??

packages/react/src/components/Toolbar/Toolbar.tsx
```diff
@@ -118,189 +97,348 @@ class Toolbar extends UIComponent<WithAsProp<ToolbarProps>, ToolbarState> {
...
+      return
     }
 
-    return {
-      initialItems: props.items,
-      currentItems: props.items,
-      stableItems: state.stableItems,
-      stable: false,
+    el.style.visibility = 'hidden'
+    const wasFocusable = el.getAttribute('data-is-focusable')
```
layershifter: We have this as constant:

```suggestion
    const wasFocusable = el.getAttribute(IS_FOCUSABLE_ATTRIBUTE)
```

packages/react/src/components/Toolbar/Toolbar.tsx
```diff
@@ -118,189 +97,348 @@ class Toolbar extends UIComponent<WithAsProp<ToolbarProps>, ToolbarState> {
...
+  /**
+   * Checks if `item` overflows a `container`.
+   * TODO: check and fix all margin combination
+   */
+  isItemOverflowing(itemBoundingRect: ClientRect, containerBoundingRect: ClientRect) {
+    return (
+      itemBoundingRect.right > containerBoundingRect.right ||
+      itemBoundingRect.left < containerBoundingRect.left
+    )
+  }
```
layershifter, line 221: Move out this to function?

packages/react/src/components/Toolbar/Toolbar.tsx
```diff
@@ -118,189 +97,348 @@ class Toolbar extends UIComponent<WithAsProp<ToolbarProps>, ToolbarState> {
...
+   * Checks if `item` would collide with eventual position of `overflowItem`.
+   */
+  wouldItemCollide(
+    $item: Element,
+    itemBoundingRect: ClientRect,
+    overflowItemBoundingRect: ClientRect,
+    containerBoundingRect: ClientRect,
+  ) {
+    let wouldCollide
+    if (this.rtl) {
```
layershifter, line 235:
```suggestion
    if (this.props.rtl) {
```

Should it come from props?

packages/react/src/components/Toolbar/Toolbar.tsx
```diff
@@ -118,189 +97,348 @@ class Toolbar extends UIComponent<WithAsProp<ToolbarProps>, ToolbarState> {
...
+              defaultProps: {
+                styles: styles.offsetMeasure,
+              },
+            })}
+          </Ref>
+        </ElementType>
+        <EventListener
+          listener={_.debounce(this.hideOverflowItems, 16)}
+          targetRef={windowRef}
+          type="resize"
```
layershifter: Is it enough to listen for `resize`? What about `scroll`?

packages/react/src/components/Toolbar/ToolbarOverflowMenu.tsx
miroslavstastny: `ToolbarOverflowItem`, I guess.

packages/react/src/components/Toolbar/Toolbar.tsx
miroslavstastny: This already is a separate function

packages/react/src/components/Toolbar/Toolbar.tsx
miroslavstastny: No, `rtl` is received as part of config in `renderComponent`, there I store it to `this`.

docs/src/examples/components/Toolbar/Content/ToolbarExamplePortal.shorthand.tsx
```diff
@@ -0,0 +1,273 @@
...
+          )}
+        </PortalWindow>
+      )}
+
+      <ToolbarExampleShorthand />
+    </>
+  )
+}
+
+export default ProviderExampleTarget
```
layershifter: Suggested options:
- hide this example and use it in screener tests only
  - use `iframe` to render it
- remove it at all

packages/react/test/specs/components/Toolbar/Toolbar-test.tsx
```diff
@@ -7,7 +7,9 @@ import { ReactWrapper } from 'enzyme'
 
 describe('Toolbar', () => {
   isConformant(Toolbar)
-
+  isConformant(Toolbar, {
+    requiredProps: { overflow: true },
```
layershifter, line 11: Is it really required?

packages/react/src/components/Toolbar/Toolbar.tsx
layershifter: Ouch, you're right 😭

packages/react/src/components/Toolbar/Toolbar.tsx
```diff
@@ -98,15 +99,15 @@ class Toolbar extends UIComponent<WithAsProp<ToolbarProps>, ToolbarState> {
...
+    overflow: PropTypes.bool,
+    overflowItem: customPropTypes.shorthandAllowingChildren,
   }
 
   static defaultProps = {
     accessibility: toolbarBehavior,
-    measurement: {},
-    wrapper: {},
+    overflowItem: {
+      icon: { name: 'more', outline: true },
```
layershifter: We don't have merging of shorthands, it means that if I will pass:

```tsx
<Toobar overflowItem={{ onClick }} />
```

It will render `ToolbarItem` without any icon 😕 
I propose to move `icon` to `defaultProps` in `.create()` call

packages/react/src/components/Toolbar/ToolbarOverflowMenu.tsx
miroslavstastny: removed completely

docs/src/examples/components/Toolbar/Content/ToolbarExamplePortal.shorthand.tsx
miroslavstastny: removed for now

docs/src/components/Sidebar/Sidebar.tsx
```diff
@@ -284,6 +284,11 @@ class Sidebar extends React.Component<any, any> {
         title: { content: 'Custom Styled Toolbar', as: NavLink, to: '/prototype-custom-toolbar' },
         public: true,
       },
+      {
+        key: 'editor',
+        title: { content: 'Editor Toolbar', as: NavLink, to: '/prototype-editor-toolbar' },
+        public: true,
+      },
```
layershifter, line 291: Double item, should be removed

layershifter (approved):

packages/react/src/components/Toolbar/Toolbar.tsx
miroslavstastny: Should be

PR #2013, opened 2019-10-07
test(projects): add behaviors usage

layershifter:
This PR adds behavior imports to our project tests to be sure that everything is exported properly.


miroslavstastny (approved):

packages/accessibility/src/behaviors/Image/imageBehavior.ts
```diff
@@ -21,5 +21,4 @@ export default imageBehavior
 type ImageBehaviorProps = {
   /** Alternative text. */
   alt?: string
-  'aria-label'?: string
```
layershifter, line 24: This type is picked on next line

PR #2015, opened 2019-10-08
chore(Dropdown): use debounce for clearing functions

silviuaavram:
Refactored clearing methods for characters so far and selection a11y message.


silviuaavram:
closing as I had to mock debounce in tests and proved too time consuming

PR #2018, opened 2019-10-09
fix(Alert): correctly handle RTL if shorthand API is used

### LINKED ISSUE #2017: Alert component doesn't render RTL order correctly
## Bug Report
When using <Alert> element with icon, text and action buttons and switching to RTL, only text is rendered on the other side, but other elements stay in the same order

### Steps
Open Stardust docs for Alert with actions and switch to RTL 
With editor - https://stardust-ui.github.io/react/components/alert/definition?showCode=true&showRtl=false&showTransparent=false&showVariables=false#slots-actions
With popout - https://stardust-ui.github.io/react/maximize/alert-example-actions-shorthand/true
 
### Expected Result
The order of the elements is reversed for RTL, i.e. if we have 
icon - text - action buttons - close for LTR, then it would be
close - action buttons (reversed order) - text - icon for RTL

### Actual Result
Order of elements is the same as for LTR
icon - text - action buttons - close 

### Version
0.39.0

### Testcase
https://stardust-ui.github.io/react/components/alert/definition?showCode=true&showRtl=false&showTransparent=false&showVariables=false#slots-actions

miroslavstastny:
Fixes #2017
Set dir="auto" on `Alert` only if Children API is used.
If shorthand API is used, content is wrapped by its own `Text` component, no need to set `dir` on parent container.

### DIFF
--- a/docs/src/examples/components/Alert/Rtl/AlertExampleDismissAction.rtl.tsx
+++ b/docs/src/examples/components/Alert/Rtl/AlertExampleDismissAction.rtl.tsx
@@ -12,6 +12,7 @@ const AlertExampleDismissActionRtl = () => (
         content: 'عالم',
       },
     ]}
+    icon="exclamation-triangle"
     content="مرحبا العالم"
   />
 )
--- a/docs/src/examples/components/Alert/Rtl/AlertExampleDismissAction.rtl.tsx
+++ b/docs/src/examples/components/Alert/Rtl/AlertExampleDismissAction.rtl.tsx
@@ -1,0 +1,19 @@
+import * as React from 'react'
+import { Alert } from '@stardust-ui/react'
+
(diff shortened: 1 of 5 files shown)


docs/src/examples/components/Alert/Rtl/AlertExampleDismissAction.rtl.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react'
+import { Alert } from '@stardust-ui/react'
+
+const AlertExampleDismissActionRtl = () => (
+  <Alert
```
pompomon, line 5: Please could you add an icon prop to this example as well?

pompomon (approved):

layershifter (approved):

PR #2020, opened 2019-10-09
fix(docs): RTL and Theme it buttons

### LINKED ISSUE #1997: Theme it and RTL buttons not work in the components documentation
### Steps
Open docs site page for a component
Click on Theme it
Click on RTL

### Expected Result
Theme panel is displayed
Example is displayed in RTL

### Actual Result
Theme panel is not displayed
Example is displayed in LTR

### Version
0.39.0

### Testcase
N/A

miroslavstastny:
Fixes #1997 

**Hide unused variables** still does not work as it would require to pass **unresolved** styles and component variables to Fela and we removed that in #1957.

Performance wins over docs => **Hide unused variables** (temporarily) removed from docs.

### DIFF
--- a/docs/src/components/ComponentDoc/ComponentExample/ComponentExampleVariables.tsx
+++ b/docs/src/components/ComponentDoc/ComponentExample/ComponentExampleVariables.tsx
@@ -1,5 +1,4 @@
 import {
-  Checkbox,
   Grid,
   Header,
   Segment,
@@ -37,7 +36,7 @@ const ComponentExampleVariables: React.FunctionComponent<
   const { onChange, overriddenVariables, usedVariables } = props
 
   const { theme } = React.useContext<ProviderContextPrepared>(ThemeContext)
-  const [hideUnused, setHideUnused] = React.useState(true)
+  const [hideUnused] = React.useState(true)
 
   const componentVariables: ThemeComponentVariablesPrepared = _.pickBy(
     mergeThemeVariables(theme.componentVariables, overriddenVariables),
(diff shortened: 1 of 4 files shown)


docs/src/components/ComponentDoc/ComponentExample/ComponentExampleVariables.tsx
```diff
@@ -37,7 +36,7 @@ const ComponentExampleVariables: React.FunctionComponent<
   const { onChange, overriddenVariables, usedVariables } = props
 
   const { theme } = React.useContext<ProviderContextPrepared>(ThemeContext)
-  const [hideUnused, setHideUnused] = React.useState(true)
+  const [hideUnused] = React.useState(true)
```
miroslavstastny, line 39: Is this just temporary and we want to invest into finding a way hot to do it or should we remove this completely? 🤔

layershifter: Let's hide this temporary and see what we can do later

packages/react/src/lib/renderComponent.tsx
```diff
@@ -223,6 +223,7 @@ const renderComponent = <P extends {}>(
   const direction = rtl ? 'rtl' : 'ltr'
   const felaParam = {
     theme: { direction },
+    displayName, // does not affect styles, only used by useEnhancedRenderer
```
```suggestion
    displayName, // does not affect styles, only used by useEnhancedRenderer in doc site
```

layershifter (approved):

PR #2025, opened 2019-10-14
perf(List): remove Flex usage

layershifter:
### ⚠️ Do not use `Flex` in components in internals

Similar to #2023.

Any Stardust component will add additional style computations (merging, resolving) and will create performance degrade. One more reason to do not use `Flex`es: styles of `Flex` are not part of component so there is no way to generate static styles from if it used.

### 🚀 Measures

The test example below contains the worse scenario when all previously used `Flex`es will be rendered.

```tsx
import * as _ from 'lodash'
import { List } from '@stardust-ui/react'

const items = _.times(100, i => ({
  key: i,
  media: 'Media',
  header: 'Foo',
  headerMedia: 'Bar',
  content: 'Baz',
  contentMedia: 'Qux',
}))

const ListPerf = () => <List items={items} />

export default ListPerf
```

Diff is around **39%**. As `ListItem` is used in `Dropdown`, it should get performance benefits.

### Before

| Example | Avg | Median |
| --- |  ---  |  ---  |
| List.perf.tsx | 194.48 | 193.65 |

### After 

| Example | Avg | Median |
| --- |  ---  |  ---  |
| List.perf.tsx | 131.42 | 130.44 |


jurokapsiar (approved):

miroslavstastny (approved):

brambow:
This change broke the display of my ListItems I think? How do we now make sure that the list item media is inline with the content if it isn't using flex?

image
The bookmark icons were in-line prior to this update.

PR #2030, opened 2019-10-16
fix(Accessibility): Button titles

mshoho:
Adding the missing button attributes to the schema, adding button titles to the docsite.

### DIFF
--- a/packages/accessibility/src/behaviors/Button/buttonBehavior.ts
+++ b/packages/accessibility/src/behaviors/Button/buttonBehavior.ts
@@ -14,6 +14,8 @@ const buttonBehavior: Accessibility<ButtonBehaviorProps> = props => ({
     root: {
       role: props.as === 'button' ? undefined : 'button',
       tabIndex: props.as === 'button' ? undefined : 0,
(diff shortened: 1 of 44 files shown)


packages/accessibility/src/behaviors/Button/buttonBehavior.ts
```diff
@@ -14,6 +14,8 @@ const buttonBehavior: Accessibility<ButtonBehaviorProps> = props => ({
     root: {
       role: props.as === 'button' ? undefined : 'button',
       tabIndex: props.as === 'button' ? undefined : 0,
+      disabled:
+        props.disabled || props.loading ? (props.as === 'button' ? true : undefined) : undefined,
```
jurokapsiar, line 18: not sure about this, pls check with @kolaps33

mshoho: Look at the next line, it sets aria-disabled when `props.disabled || props.loading`.
I have just added a bit of consistency to have a native disabled attribute when it's a native button.

PR #2031, opened 2019-10-16
fix(Accessibility): Checking if the slot attributes are defined

mshoho:
An additional safety check to make sure the slot is defined in AccessibilityDefinition.


jurokapsiar (approved):
pls add changelog

miroslavstastny (approved):

packages/react/src/lib/renderComponent.tsx
```diff
@@ -86,6 +86,10 @@ const getAccessibility = (
     if (definition.attributes) {
       const slotNames = Object.keys(definition.attributes)
       slotNames.forEach(slotName => {
```
layershifter, line 88:
```suggestion

      slotNames.forEach(slotName => {
```

PR #2040, opened 2019-10-18
fix(accessibility): checking if the slot attributes are defined

mshoho:
An additional safety check to make sure the slot is defined in AccessibilityDefinition.


layershifter (approved):
Please add changelog before merging

PR #2041, opened 2019-10-18
perf: cache component variables

miroslavstastny:
Do not resolve component variables in each render.

Resolved variables are now cached in `ProviderContext`. Nested `Provider` starts with an empty cache.

Performance:

| Example | Avg | Median |
| :--- |  ---:  |  ---:  |
| Chat.perf BEFORE | 491.65 | 439.58 |
| Chat.perf AFTER | 436.13 | 397.77 |

### DIFF
--- a/packages/react/src/lib/renderComponent.tsx
+++ b/packages/react/src/lib/renderComponent.tsx
@@ -29,7 +29,7 @@ import {
 import { Props, ProviderContextPrepared } from '../types'
 import { ReactAccessibilityBehavior, AccessibilityActionHandlers } from './accessibility/reactTypes'
 import getKeyDownHandlers from './getKeyDownHandlers'
-import { emptyTheme, mergeComponentStyles } from './mergeThemes'
+import { emptyTheme, mergeComponentStyles, mergeComponentVariables } from './mergeThemes'
 import createAnimationStyles from './createAnimationStyles'
 import { isEnabled as isDebugEnabled } from './debug/debugEnabled'
 import { DebugData } from './debug/debugData'
@@ -151,7 +151,6 @@ const renderComponent = <P extends {}>(
   config: RenderConfig<P>,
   context?: ProviderContextPrepared,
 ): React.ReactElement<P> => {
-  performance.now()
   const {
     className,
     displayName,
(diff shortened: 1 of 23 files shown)


packages/react/src/types.ts
```diff
@@ -168,4 +168,5 @@ export interface ProviderContextPrepared {
   target: Document
   theme: ThemePrepared
   originalThemes: (ThemeInput | undefined)[]
+  resolvedComponentVariables: Record<string, object>
```
layershifter:
```suggestion
  DO_NOT_USE_OR_YOU_WILL_BE_FIRED_resolvedComponentVariables: Record<string, object>
```

miroslavstastny: would `_internal_resolvedComponentVariables` be enough?

packages/react/src/lib/renderComponent.tsx
```diff
@@ -166,17 +166,30 @@ const renderComponent = <P extends {}>(
...
   const ElementType = getElementType(props) as React.ReactType<P>
   const stateAndProps = { ...state, ...props }
 
-  // Resolve variables for this component, allow props.variables to override
-  const resolvedVariables: ComponentVariablesObject = mergeComponentVariables(
-    theme.componentVariables[displayName],
-    props.variables && withDebugId(props.variables, 'props.variables'),
-  )(theme.siteVariables)
+  // Resolve variables for this component, cache the result in provider
+  if (!(displayName in resolvedComponentVariables)) {
```
```suggestion
  if (!resolvedComponentVariables[displayName]) {
```

If there difference between this?

packages/react/src/lib/renderComponent.tsx
```diff
@@ -166,17 +166,30 @@ const renderComponent = <P extends {}>(
...
+      callable(theme.componentVariables[displayName])(theme.siteVariables) || {} // component variables must not be undefined/null (see mergeComponentVariables contract)
+  }
+
+  // Merge inline variables on top of cached variables
+  const resolvedVariables = props.variables
+    ? mergeComponentVariables(
+        resolvedComponentVariables[displayName],
+        withDebugId(props.variables, 'props.variables'),
+      )(theme.siteVariables)
+    : resolvedComponentVariables[displayName]
```
layershifter, line 192: nit, it's hard to read this, I suggest

```tsx
let resolvedVariables = resolvedComponentVariables[displayName]

if (props.variables) {
  resolvedVariables = mergeComponentVariables(
    resolvedComponentVariables[displayName],
    withDebugId(props.variables, 'props.variables'),
  )(theme.siteVariables)
}
```

packages/react/src/lib/renderComponent.tsx
miroslavstastny: yes, they are completely opposite :-P

other than that, yes:
image

but because of the `|| {}` on line 183, this is no longer necessary

packages/react/src/lib/renderComponent.tsx
miroslavstastny: I don't like `let`.

packages/react/src/types.ts
miroslavstastny: renamed

packages/react/src/lib/renderComponent.tsx
miroslavstastny: removed

layershifter (approved):

jurokapsiar (approved):

PR #2044, opened 2019-10-18
perf: use memoized renderRule()

layershifter:


ecraig12345:
Heads up if you reopen this that I made a change #2153 renaming all `lib` folders to `utils`. So the changes to `lib` files will need to be moved to the new location.

PR #2046, opened 2019-10-21
docs(readme): update references to fluent-ui-react

jurokapsiar:


layershifter (approved):
Not sure that this is enough, but we can go :)

PR #2047, opened 2019-10-21
docs(Toolbar): Label toolbars examples, adding best practices

kolaps33:
JAWS doesn't narrate some of our toolbars. 
The same issue appear when using toolbars on other pages. 

Issue is that if there is more than one toolbars on the page, then each toolbar need be labelled.

### DIFF
--- a/docs/src/examples/components/Toolbar/Content/ToolbarExampleCustomContent.shorthand.tsx
+++ b/docs/src/examples/components/Toolbar/Content/ToolbarExampleCustomContent.shorthand.tsx
@@ -3,6 +3,7 @@ import { Button, Text, Toolbar } from '@stardust-ui/react'
 
 const ToolbarExampleCustomContentShorthand = () => (
   <Toolbar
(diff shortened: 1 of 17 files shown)


docs/src/examples/components/Toolbar/Content/ToolbarExampleCustomContent.shorthand.tsx
```diff
@@ -3,6 +3,7 @@ import { Button, Text, Toolbar } from '@stardust-ui/react'
 
 const ToolbarExampleCustomContentShorthand = () => (
   <Toolbar
+    aria-labelledby="toolbar-can-contain-custom-content"
```
jurokapsiar: use `aria-label` as the examples should not reference anything outside of it

kolaps33: done :)

jurokapsiar (approved):

PR #2052, opened 2019-10-23
Extending the schema, fixing the Docsite accordingly

mshoho:
Checkbox, RadioGroupItem, ToolbarItem, PopupButton, various Docsite fixes.


packages/accessibility/src/behaviors/Button/buttonBehavior.ts
```diff
@@ -14,6 +14,8 @@ const buttonBehavior: Accessibility<ButtonBehaviorProps> = props => ({
     root: {
       role: props.as === 'button' ? undefined : 'button',
       tabIndex: props.as === 'button' ? undefined : 0,
+      disabled:
+        props.disabled || props.loading ? (props.as === 'button' ? true : undefined) : undefined,
```
jurokapsiar: pls revert

mshoho: I've changed it differently. The button will have `disabled` if it's not a loading button. For the loading button the class is redefined as LoadingButton and `disabled` is not being set (and the validator validates that it's not set).

docs/src/examples/components/Popup/Types/PopupExamplePointing.shorthand.tsx
```diff
@@ -18,6 +18,7 @@ const PopupWithButton = props => (
       <Button
         icon={props.icon}
         styles={{ padding: props.padding, height: '64px', minWidth: '64px' }}
+        title={'Show popup'}
```
miroslavstastny:
```suggestion
        title="Show popup"
```

miroslavstastny: Several other places as well

packages/accessibility/src/behaviors/Button/buttonBehavior.ts
```diff
@@ -9,26 +9,37 @@ import { Accessibility } from '../../types'
...
-      tabIndex: props.as === 'button' ? undefined : 0,
-      'aria-disabled': props.disabled || props.loading,
+const buttonBehavior: Accessibility<ButtonBehaviorProps> = props => {
+  const definition: AccessibilityDefinition = {
+    attributes: {
+      root: {
+        role: props.as === 'button' ? undefined : 'button',
+        tabIndex: props.as === 'button' ? undefined : 0,
+        disabled:
+          props.disabled && !props.loading ? (props.as === 'button' ? true : undefined) : undefined,
```
miroslavstastny, line 19: Does this mean that `<Button disabled loading />` is not disabled?

miroslavstastny (approved):

packages/accessibility/src/behaviors/Button/buttonBehavior.ts
mshoho: It will be disabled, but with aria-disabled only, not with the native `disabled` attribute. Using native attributes is more preferable in order to have the default native logic of setting an attribute. But because of that native logic we need this condition for the loading button case for the sake of backwards compatibility. If you disable a currently focused button, it will lose the focus. At the same time, we have this loading button (I would propose to make a separate component named LoadingButton, by the way, because Button is probably the most used component and the loading button overhead is something which is almost never used) and it shouldn't lose the focus (because it usually gets enabled back quickly enough and we don't want to have a dedicated logic for moving the focus back). So we have this condition, and there are two classes in the schema. One is for the standard button and it requires `<button>` to be disabled with the `disabled` attribute, and there is another one for the loading button.

PR #2060, opened 2019-10-24
fix(Accessibility): filter out key actions with no key combinations

### LINKED ISSUE #2058: Keyboard navigation key handlers exception in RTL
## Bug Report
In RTL, text input (Input, TextArea) crashes with uncaught exception:

TypeError
keyCombinations.map is not a function

This causes bad rendering behavior.

### Steps
Use this example:
https://codesandbox.io/s/stardust-ui-example-p3n6j

See that text input both in RTL and LTR work normally.

Download Chrome V69
https://www.slimjet.com/chrome/google-chrome-old-version.php

See that when RTL is false, text input works normally. Then set it to true. See that any text input now causes an uncaught exception to be thrown.

### Expected Result
Text input to work normally.

### Actual Result
Uncaught exception causes crash.

### Version
Chrome: Version 69.0.3497.92 (Official Build) (64-bit)
Stardust: Stardust UI @ 0.40.1

### Testcase
N/A

miroslavstastny:
Fixes #2058.
Ignore keyActions with empty keyCombinations. Fixes exception in RTL.

### DIFF
--- a/packages/react/test/specs/lib/getKeyDownHandlers-test.ts
+++ b/packages/react/test/specs/lib/getKeyDownHandlers-test.ts
@@ -34,6 +34,7 @@ describe('getKeyDownHandlers', () => {
       expect(keyHandlers.hasOwnProperty(partElementName)).toBeTruthy()
       expect(keyHandlers[partElementName].hasOwnProperty('onKeyDown')).toBeTruthy()
     })
+
     test('for few component elements', () => {
       const actions = {
         testAction: () => {},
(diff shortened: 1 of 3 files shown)


jurokapsiar (approved):

packages/react/test/specs/lib/getKeyDownHandlers-test.ts
```diff
@@ -142,17 +185,35 @@ describe('getKeyDownHandlers', () => {
...
+
+    test('when action definition has no keyCombinations', () => {
+      const actions = {
+        testAction: () => {},
+        otherAction: () => {},
+      }
+
+      actionsDefinition.anotherPart = {
+        otherAction: {
+          keyCombinations: false,
```
mnajdova, line 233: Can we add tests for other falsy values as well, like undefined, null, [] etc.

mnajdova (approved):

packages/react/test/specs/lib/getKeyDownHandlers-test.ts
miroslavstastny: added test, fixed code for []. could you please recheck?

mnajdova: Looks good!

PR #2061, opened 2019-10-24
feat(Tree): control activeItemIds through expanded prop of items

silviuaavram:
User passing `expanded` prop to items will allow him to control the expanded state of these items. Renamed `open` to `expanded`.

Controlling the Tree's `activeItemIds` will have precedence over this individual control method.

Added also missing tests for Tree, subcomponents and behaviors. Added / edited test definitions for behavior tests.


packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -64,8 +64,8 @@ export interface TreeItemProps extends UIComponentProps, ChildrenComponentProps
   /** Called when the item's parent is about to be focused. */
   onFocusParent?: ComponentEventHandler<TreeItemProps>
 
-  /** Whether or not the item is in the open state. Only makes sense if item has children items. */
-  open?: boolean
+  /** Whether or not the item is in the expanded state. Only makes sense if item has children items. */
+  expanded?: boolean
```
mnajdova, line 68: This is a breaking change, please add entry in the changelog under this section.

packages/accessibility/test/behaviors/testDefinitions.ts
```diff
@@ -597,7 +597,13 @@ definitions.push({
   regexp: /Triggers '(\w+)' action with '(\w+)' on '([\w-]+)', when has an opened subtree\./g,
   testMethod: (parameters: TestMethod) => {
     const [action, key, elementToPerformAction] = [...parameters.props]
-    const propertyOpenedSubtree = { open: true, items: [{ a: 1 }], siblings: [], hasSubtree: true }
+    const propertyOpenedSubtree = {
+      open: true,
```
mnajdova, line 670: Shouldn't this open be removed? Also would propose renaming of `propertyOpenedSubtree` to `propertyExpendedSubtree` for consistency.

mnajdova:
Let's add an example of how the user can control the expended prop

packages/react/src/components/Tree/TreeItem.tsx
```diff
@@ -117,7 +117,7 @@ class TreeItem extends UIComponent<WithAsProp<TreeItemProps>, TreeItemState> {
     onFocusFirstChild: PropTypes.func,
     onFocusParent: PropTypes.func,
     onSiblingsExpand: PropTypes.func,
-    open: PropTypes.bool,
+    expanded: PropTypes.bool,
```
mnajdova, line 120: I think this should be treated as auto controlled prop, by defining `defaultExpanded` and `onExpandedChange`. With this we will allow the user to define only the initial state, if they want that, or listen to the `onExpandedChange` and change the `expanded` prop by themselves.

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -146,9 +146,35 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
+    let { activeItemIds } = nextProps
+
+    if (!activeItemIds && items) {
+      activeItemIds = prevState.activeItemIds
+
+      const expandedItemsGenerator = items =>
+        _.reduce(
+          items,
+          (acc, item) => {
+            if (item['expanded'] && acc.indexOf(item['id']) === -1) {
```
mnajdova, line 163: The items could be defined as render functions too, we should handle that case, or at least check whether it is an object.

packages/react/src/components/Tree/Tree.tsx
```diff
@@ -146,9 +146,35 @@ class Tree extends AutoControlledComponent<WithAsProp<TreeProps>, TreeState> {
...
   static getAutoControlledStateFromProps(nextProps: TreeProps, prevState: TreeState) {
-    const itemsForRender = Tree.getItemsForRender(nextProps.items)
+    const { items } = nextProps
+    const itemsForRender = Tree.getItemsForRender(items)
+    let { activeItemIds } = nextProps
+
+    if (!activeItemIds && items) {
+      activeItemIds = prevState.activeItemIds
+
+      const expandedItemsGenerator = items =>
```
mnajdova: This currently doesn't work if the second item inside the Tree has `expanded`. May need to be refactored to something like:
```
  static expandedItemsGenerator = (items, activeItemIds) =>
  _.reduce(
    items,
    (acc, item) => {
      if (item['expanded'] && acc.indexOf(item['id']) === -1) {
        acc.push(item['id'])
      }

      if (item['items']) {
        return [...acc, Tree.expandedItemsGenerator(item['items'], acc)]
      }

      return acc
    },
    activeItemIds,
  )

  static getAutoControlledStateFromProps(nextProps: TreeProps, prevState: TreeState) {
    const { items } = nextProps
    const itemsForRender = Tree.getItemsForRender(items)
    let { activeItemIds } = nextProps

    if (!activeItemIds && items) {
      activeItemIds = prevState.activeItemIds
      activeItemIds = Tree.expandedItemsGenerator(items, activeItemIds)
    }

    return {
      activeItemIds,
      itemsForRender,
    }
  }
```

packages/react/test/specs/components/Tree/Tree-test.tsx
```diff
@@ -156,5 +156,23 @@ describe('Tree', () => {
         .simulate('keydown', { keyCode: keyboardKey['*'] })
       checkOpenTitles(wrapper, ['1', '11', '12', '2', '21', '22', '3'])
     })
+
+    it('should have items expanded based on their expanded prop', () => {
```
mnajdova, line 160: Maybe it would be worther to add tests for expand on second/third item. This doesn't work currently:
```
import React from 'react'
import { Tree } from '@stardust-ui/react'

const items = [
  {
    id: '1',
    title: 'House Lannister',
    items: [
      {
        id: '11',
        title: 'Tywin',
        items: [
          {
            id: '111',
            title: 'Jaime',
          },
          {
            id: '112',
            title: 'Cersei',
          },
          {
            id: '113',
            title: 'Tyrion',
          },
        ],
      },
      {
        id: '12',
        title: 'Kevan',
        items: [
          {
            id: '121',
            title: 'Lancel',
          },
          {
            id: '122',
            title: 'Willem',
          },
          {
            id: '123',
            title: 'Martyn',
          },
        ],
      },
    ],
  },
  {
    id: '2',
    title: 'House Targaryen',
    expanded: true,
    items: [
      {
        id: '21',
        title: 'Aerys',
        items: [
          {
            id: '211',
            title: 'Rhaegar',
          },
          {
            id: '212',
            title: 'Viserys',
          },
          {
            id: '213',
            title: 'Daenerys',
          },
        ],
      },
    ],
  },
]

const TreeInitiallyOpenExampleShorthand = () => (
  <Tree items={items} aria-label="Initially open" />
)

export default TreeInitiallyOpenExampleShorthand

```

See https://github.com/microsoft/fluent-ui-react/pull/2061/files#r344072359 for possible fix

packages/accessibility/test/behaviors/testDefinitions.ts
silviuaavram: We are not changing the name in the HierarchicalTree too. This was a hack to make this definition work with both behaviors.

packages/react/src/components/Tree/TreeItem.tsx
silviuaavram: we are not having any state in TreeItem. All expanded stateful logic is in Tree, since TreeItems can be unmounted on virtualization. This is just a prop. The user can control it as it is.

packages/react/test/specs/components/Tree/Tree-test.tsx
silviuaavram: are you sure this does not work? If I have `expanded` on  `Tywin` and open `House Lannister` manually, we find `Tywin` already expanded.

mnajdova: I would expect `House Targaryen` to be opened, but it is not, so I guess it doesn't work

packages/react/src/components/Tree/Tree.tsx
silviuaavram: Currently it won't work well with render functions. We are calculating some props that we pass to TreeItem from Tree in `getItemsForRender`. However we probably can support this scenario with render prop as well. We just call the children function with the props from `itemsForRender` so the child component has access to them as well.

I would consider working on this in a separate PR. And just perform here the object check as per suggestion.

What do you think @mnajdova @jurokapsiar @miroslavstastny ?

packages/accessibility/test/behaviors/testDefinitions.ts
```diff
@@ -167,11 +167,13 @@ function testMethodConditionallyAddAttribute(
   component,
   propertyDependsOn,
   valueOfProperty,
+  valueOfPropertyOtherwise,
   attributeToBeAdded,
   valueOfAttributeToBeAddedIfTrue,
   valueOfAttributeToBeAddedOtherwise,
 ) {
   const propertyWithAriaSelected = {}
+  propertyWithAriaSelected[propertyDependsOn] = valueOfPropertyOtherwise
```
silviuaavram, line 176: to reuse this method I had to prop the `otherwise` condition, which was by default `undefined`.

packages/accessibility/test/behaviors/testDefinitions.ts
```diff
@@ -337,16 +404,18 @@ definitions.push({
...
-    propertyFirstPropNegate[
-      propertyDependingOnFirst
-    ] = !testHelper.convertToMatchingTypeIfApplicable(attributeExpectedValue)
-    propertyFirstPropNegate[propertyDependingOnSecond] = true
-    const actualResultFirstPropertyNegate = parameters.behavior(propertyFirstPropNegate).attributes
-      .root[attributeToBeAdded]
-    expect(testHelper.convertToMatchingTypeIfApplicable(actualResultFirstPropertyNegate)).toEqual(
-      !testHelper.convertToMatchingTypeIfApplicable(attributeExpectedValue),
-    )
+    if (typeof testHelper.convertToMatchingTypeIfApplicable(attributeExpectedValue) === 'boolean') {
```
silviuaavram, line 407: it worked with other params apart from boolean, but this part does not make sense unless the value is a boolean.

packages/react/src/components/Tree/TreeItem.tsx
mnajdova: I am not sure about this one. I understand that it is just a prop, but all other similar props are treated as autocontrolled. @miroslavstastny thoughts on this?

miroslavstastny (approved):

PR #2062, opened 2019-10-24
fix(Popper): remove reference to popper element after destroy

jurokapsiar:
Popper keeps a reference to the DOM node, which needs to be cleaned up
Temporarily fix it in the popper wrapper component until fixed properly in popper.


packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -69,6 +69,11 @@ const Popper: React.FunctionComponent<PopperProps> = props => {
   const destroyInstance = React.useCallback(() => {
     if (popperRef.current) {
       popperRef.current.destroy()
+      if (popperRef.current.popper) {
+        // Popper keeps a reference to the DOM node, which needs to be cleaned up
+        // temporarily fix it here until fixed properly in popper
+        popperRef.current.popper = null
```
miroslavstastny, line 75: I don't understand this. Setting the whole `popperRef.current` to `null` on the next line should be enough.

miroslavstastny (approved):

PR #2073, opened 2019-10-30
fix(Popup): prevent unintentional content wrapping

### LINKED ISSUE #2072: Incorrect popup placement after jump-resize
### Steps
1. Open a popup
2. Resize browser window using a hotkey to a half

### Expected Result
Popup correctly positioned

### Actual Result
Position is incorrect, next repositioning fixes that 💣
out

### Version
0.40.2

Introduced by #2067, this does not happen with GPU acceleration enabled.

miroslavstastny:
Prevent unintentional Popup content wrapping to fix re-positioning on window resize.
Fixes #2072


packages/react/src/themes/teams/components/Popup/popupContentStyles.ts
```diff
@@ -9,6 +9,7 @@ const popupContentStyles: ComponentSlotStylesPrepared<PopupContentProps, PopupCo
     borderRadius: v.borderRadius,
     boxShadow: v.boxShadow,
 
+    width: 'fit-content',
```
layershifter: It seems that it will not work for IE:

S91030-15034276
https://developer.mozilla.org/en-US/docs/Web/CSS/width

miroslavstastny: And Edge 😢

miroslavstastny: But this is just a workaround for #2067 which is a workaround for bug in Chrome 69.
@jurokapsiar - what if we do #2067 conditionally only in Chrome?

jurokapsiar: agree - in chrome and presumably in all chromium browsers (new edge)

miroslavstastny: Both @levithomason and @dzearing are strongly against browser-specific code

miroslavstastny:
`width: fit-content` replaced by `display: table` which should do the same and be multiplatform.
Tested in Chrome, Edge, IE11

jurokapsiar (approved):

miroslavstastny:
Fixed better in #2187

PR #2076, opened 2019-10-30
feat(button): Adding inverted button for Teams theme

notandrew:
Adding button style that inherits its background.

image

high contrast:
image

high contrast hover state (different than default button)
image


packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -149,6 +149,35 @@ const buttonStyles: ComponentSlotStylesPrepared<ButtonProps, ButtonVariables> =
           },
         }),
 
+      ...(p.inverted && {
+        outline: 0,
```
codepretty: Do we ever want an outline on button? Should we move this to all buttons - ie outside of `p.inverted`?

packages/react/src/themes/teams/components/Button/buttonVariables.ts
```diff
@@ -95,6 +106,17 @@ export default (siteVars: any): ButtonVariables => ({
   primaryBackgroundColorDisabled: siteVars.colorScheme.default.backgroundDisabled,
   primaryBorderColor: 'transparent',
 
+  invertedColor: siteVars.colors.white,
```
codepretty: Is the reason we need all these same colors to override hover/focus/etc from other places in button styles?

packages/react/src/components/Button/Button.tsx
```diff
@@ -77,6 +77,9 @@ export interface ButtonProps
   /** A button can emphasize that it represents an alternative action. */
   secondary?: boolean
 
+  /** Inverted buttons are meant to be shown on dark backgrounds */
+  inverted?: boolean
```
codepretty, line 81: I would call this something else, because `inverted` is being used for `Input` and `Dropdown` with a very specific set of background colors.

Example) For `inverted` Input and Dropdown it's swapping the background colors
image

And this button is a different kind of variant than that.

packages/react/src/themes/teams/components/Button/buttonStyles.ts
notandrew: that's an artifact - removed. It's also on the 'rectangle'.

packages/react/src/components/Button/Button.tsx
```diff
@@ -77,6 +77,9 @@ export interface ButtonProps
   /** A button can emphasize that it represents an alternative action. */
   secondary?: boolean
 
+  /** A button that represents an alternative action and also has contrast against dark backgrounds */
+  secondaryAlt?: boolean
```
mnajdova: Why not just reusing the `secondary` prop? In my opinion it would be a good use-case. You can see different implementations here: https://open-ui.org/components/button Also, please update the changelog to reflect the real implementation.

notandrew: If we have like a 'transparent' property (or 'ghost', but I really dislike that term), it would only apply to a secondary button and that doesn't seem as intuitive to me. I think it's more straight-forward to use a new button variant here in the same way we don't reuse the primary button for secondary. What do you think?

notandrew:
@mnajdova - can you take another look?

packages/react/src/themes/teams-high-contrast/components/Button/buttonVariables.ts
```diff
@@ -15,6 +16,8 @@ export default (siteVars: any): Partial<ButtonVariables> & ButtonHighContrastVar
     primaryBackgroundColorActive: siteVars.colors.white,
     primaryBackgroundColorFocus: siteVars.accessibleCyan,
 
+    secondaryAltPressedColor: siteVars.colors.black,
```
codepretty: Should this be renamed to `secondaryAltForegroundColor` or something since it's used for more than just `pressed` state?

packages/react/src/themes/teams/components/Button/buttonStyles.ts
```diff
@@ -152,6 +152,36 @@ const buttonStyles: ComponentSlotStylesPrepared<ButtonProps, ButtonVariables> =
           },
         }),
 
+      ...(p.secondaryAlt && {
+        color: siteVariables.colorScheme.silver.foreground,
+        borderColor: siteVariables.colorScheme.silver.borderColor,
```
codepretty: I don't think `siteVariables.colorScheme.silver.borderColor` value exists... did you mean `siteVariables.colorScheme.silver.border`?

packages/react/src/themes/teams-high-contrast/components/Button/buttonVariables.ts
codepretty: Actually, is this needed or does the color schemes use the right values here? If not, should those be updated or new values added?

notandrew: The color scheme doesn't work with this button for some reason. This is the hover state if I remove all the overrides:
image

packages/react/src/themes/teams/components/Button/buttonStyles.ts
notandrew: That explains why it wasn't working. 🙄

codepretty (approved):

docs/src/examples/components/Button/Types/index.tsx
```diff
@@ -14,6 +14,11 @@ const Types = () => (
       description="A button can be formatted to show different levels of emphasis."
       examplePath="components/Button/Types/ButtonExampleEmphasis"
     />
+    <ComponentExample
+      title="Secondary Alt"
+      description="Secondary Alt button."
```
codepretty: I would update this to say what the secondary alt button is for... like "button that inherits its background". I know that is very Teams theme specific, but I think this button is also very Teams specific..

packages/react/src/components/Button/Button.tsx
mnajdova: We don't want to just invent new concept and add it as a prop on Fluent, if it is not a common thing on the web. This is why I shared the link with you, so you can take a look on the common Button props that exists. Semantic has the prop basic: https://react.semantic-ui.com/elements/button/ for describing something similar. You may consider using this.

packages/react/src/components/Button/Button.tsx
```diff
@@ -77,6 +77,9 @@ export interface ButtonProps
   /** A button can emphasize that it represents an alternative action. */
   secondary?: boolean
 
+  /** A button that represents an alternative action and also has contrast against dark backgrounds */
```
mnajdova: We don't want to suggest a styling concepts in the properties description. You can see how this is describe in Semantic for the basic button: `The basic button has a subtle appearance.` - it is not describing the styling, but the purpose of the new term (prop) on the Button

packages/react/src/themes/teams-high-contrast/components/Button/buttonStyles.ts
```diff
@@ -1,43 +0,0 @@
-import { ComponentSlotStylesPrepared, ICSSInJSStyle } from '../../../types'
-import { ButtonProps } from '../../../../components/Button/Button'
-import { ButtonVariables } from '../../../teams/components/Button/buttonVariables'
-import { ButtonHighContrastVariables } from './buttonVariables'
-
-const buttonStyles: ComponentSlotStylesPrepared<
-  ButtonProps,
-  ButtonVariables & ButtonHighContrastVariables
-> = {
-  root: ({ props: p, variables: v, theme: { siteVariables } }): ICSSInJSStyle => {
```
mnajdova: Are these files remove by intention?

mnajdova (approved):

packages/react/src/themes/teams-high-contrast/components/Button/buttonStyles.ts
notandrew: They were at one point but it looks like we should just revert, so I'm reverting them.

PR #2079, opened 2019-10-31
feat(Provider): Rendering performance telemetry

miroslavstastny:
Add rendering performance telemetry to understand how much time the application spends rendering Stardust components.

### How to use
Add `telemetryRef` prop to the `Provider`:
```jsx
const telemetryRef = React.useRef<Telemetry>()

return (
  <Provider telemetryRef={telemetryRef}>
  ...
  </Provider>
)
```

Stardust immediately starts measuring performance of all components inside that `Provider` (including nested `Provider`s).

To enable/disable the telemetry, set enabled property:
```jsx
telemetryRef.current.enabled = false
```

To reset the data gathered so far, call `reset()`:
```jsx
telemetryRef.current.reset()
```

Gathered data are available in performance:
```jsx
const totals = _.reduce(telemetry.current.performance, (acc, next) => {
    acc.count += next.count
    acc.msTotal += next.msTotal
    return acc
  },
  { count: 0, msTotal: 0 },
)

console.log(`Rendered ${totals.count} Stardust components in ${totals.msTotal} ms`)
console.table(telemetry.performance)
```

### Example output:
image

### Implementation
Most of the rendering time is spent inside `renderComponent` function.
This PR adds markers to that function to measure the time spent inside that per-component name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Features
 - Add bounce animation to button clicks in Teams theme @notandrew ([#1724](https://github.com/stardust-ui/react/pull/1724))
 - Update Silver color scheme, adding `foregroundHover`, `foregroundPressed` and `background` definitions @pompomon ([#2078](https://github.com/microsoft/fluent-ui-react/pull/2078))
+
+### Performance
 - Add rendering performance telemetry @miroslavstastny ([#2079](https://github.com/microsoft/fluent-ui-react/pull/2079))
 
 <!--------------------------------[ v0.40.2 ]------------------------------- -->
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Features
 - Add bounce animation to button clicks in Teams theme @notandrew ([#1724](https://github.com/stardust-ui/react/pull/1724))
 - Update Silver color scheme, adding `foregroundHover`, `foregroundPressed` and `background` definitions @pompomon ([#2078](https://github.com/microsoft/fluent-ui-react/pull/2078))
+- Add rendering performance telemetry @miroslavstastny ([#2079](https://github.com/microsoft/fluent-ui-react/pull/2079))
 
 <!--------------------------------[ v0.40.2 ]------------------------------- -->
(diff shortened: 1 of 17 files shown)


packages/react/src/lib/mergeProviderContexts.ts
```diff
@@ -84,11 +84,7 @@ const mergeProviderContexts = (
         acc.disableAnimations = mergedDisableAnimations
       }
 
-      const contextOriginalThemes = (next as ProviderContextPrepared).originalThemes
```
miroslavstastny, line 87: Removed cruft from old styles debug.

docs/src/prototypes/customToolbar/index.tsx
```diff
@@ -58,21 +58,53 @@ const CustomToolbarPrototype: React.FunctionComponent = () => {
...
+  const telemetryRef = React.useRef<Telemetry>()
+
   let theme = {}
   if (themeName === 'teamsDark') {
     theme = mergeThemes(themes.teamsDark, darkThemeOverrides)
   } else if (themeName === 'teamsHighContrast') {
     theme = mergeThemes(themes.teamsHighContrast, darkThemeOverrides, highContrastThemeOverrides)
   }
 
+  React.useEffect(() => {
```
miroslavstastny, line 71: Just an example, but let's leave it here for now, until we will have a dedicated doc.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 ### Features
 - Add bounce animation to button clicks in Teams theme @notandrew ([#1724](https://github.com/stardust-ui/react/pull/1724))
 - Update Silver color scheme, adding `foregroundHover`, `foregroundPressed` and `background` definitions @pompomon ([#2078](https://github.com/microsoft/fluent-ui-react/pull/2078))
+- Add rendering performance telemetry @miroslavstastny ([#2079](https://github.com/microsoft/fluent-ui-react/pull/2079))
```
levithomason, line 28: Move to Performance heading.  See line 8.

levithomason (approved):
Let move the telemetry counting to the Telemetry class per discussion.  Also, add an `msAvg` getter for ease of use.

miroslavstastny:
> Let move the telemetry counting to the Telemetry class per discussion. Also, add an `msAvg` getter for ease of use.

Discussed offline, we'll do this later. Let's get the first version rollin'.

PR #2087, opened 2019-11-04
fix(Provider): Honor disableAnimations

### LINKED ISSUE #2086: Unable to disable animations
### Steps
```jsx
<Provider disableAnimations>
  <AnimatedComponent>
</Provider>
```

### Expected Result
`AnimatedComponent` is NOT animated

### Actual Result
It still animates 💣

### Version
0.40.2

### Testcase
https://codesandbox.io/s/xbkuv?module=/example.js

miroslavstastny:
Fixes #2086


packages/react/src/lib/felaDisableAnimationsPlugin.ts
```diff
@@ -24,16 +24,17 @@ export default () => {
       return Object.keys(styles).reduce((acc, cssPropertyName) => {
         const cssPropertyValue = styles[cssPropertyName]
 
+        if (animationProps.indexOf(cssPropertyName) !== -1) {
```
miroslavstastny, line 27: We want to remove `animationProps` even if they are objects.

packages/react/src/lib/renderComponent.tsx
```diff
@@ -231,6 +231,7 @@ const renderComponent = <P extends {}>(
   const direction = rtl ? 'rtl' : 'ltr'
   const felaParam = {
     theme: { direction },
+    disableAnimations,
```
mnajdova, line 242: We have check in the plugin for this:
```
if (props && props.disableAnimations && type === 'RULE')
```

Do we really need this one here too?

miroslavstastny: But the `felaParam` here is the `props` in the plugin. That was the cause of the issue.
We originally passed or params to fela but that changed in #1957 and it caused this bug.

mnajdova: Got it! Thanks for the clarification

mnajdova (approved):
Please add changelog entry before merging

PR #2089, opened 2019-11-05
Fix incorrect color assignment for steel light variant

natashamayurshah:


packages/react/src/themes/teams/categoryColors.ts
```diff
@@ -466,7 +466,7 @@ export const categoryColors: TeamsCategoryColors = {
     350: '#C1CACC',
     400: '#ACABBB',
     450: '#A9B5B9',
-    500: '#A2A0B2',
+    500: '#A0AEB2',
```
mnajdova, line 469: Please add a changelog entry, so it would be easy to track when this was changed:

> - Updated steel light 500 variant in Teams theme @natashamayurshah ([#2089](https://github.com/microsoft/fluent-ui-react/pull/2089))

mnajdova (approved):
Good to be merge after updating the changelog

packages/react/src/themes/teams/categoryColors.ts
natashamayurshah: Done.

Issue #2080, opened 2019-11-04
SplitButton- Accessibility issues 


kolaps33:
## Bug Report

Bug no 1:  **Menu is not close, when user TAB away**
### Steps
Navigate to SplitButton example
open menu by Alt+down arrow key
press TAB

### Expected Result
user navigate away from menu 
menu is closed

### Actual Result
user navigate away from menu 
menu is not closed, stays open

Bug no 2: **Menu instruction narration** 
1.	[Screen readers]: On “New conversation” split button, JAWS reads as “New conversation split button to open menu, press Alt + Arrow Down Press space to activate the menu, then navigate with arrow keys”. Here reading “Press space to activate the menu, then navigate with arrow keys” is incorrect also pressing “Space” key will selects “New conversation”.

Comment: I think we should remove aria-haspopup this is causing narrationg for JAWS.

PR #2090, opened 2019-11-05
fix(Splitbutton) accessibility - screen reader fixes

### LINKED ISSUE #2083: SplitButton - screen readers - cursor/browse mode and disabled splitButton
Go through the points and fix it or report it as JAWS bugs. Please test it as well with NVDA and VoiceOver.
14.	[Screen readers]: In cursor mode, when navigating to “New conversation” disabled button JAWS reads as “New conversation button menu  Unavailable  menu”. (Expected: Screen reader should read as “New conversation split button menu Unavailable)
15.	[Screen readers]: On “More options” disabled button Screen reader reads as “Unlabeled 1 button menu Unavailable menu” (this will not repro when point 4 is fixed)

kolaps33:
fixes
#2083
#2080

 removing role='none' from menuButtonBehavior as this was not really needed.


packages/accessibility/src/behaviors/SplitButton/splitButtonBehavior.ts
```diff
@@ -25,6 +26,8 @@ const splitButtonBehavior: Accessibility = props => {
             keyCombinations: [
               { keyCode: keyboardKey.Escape },
               { keyCode: keyboardKey.ArrowUp, altKey: true },
+              { keyCode: keyboardKey.Tab, shiftKey: false },
```
mnajdova, line 39: Are we absolutely sure that the items inside the menu cannot have elements that can be focused?

packages/accessibility/src/behaviors/SplitButton/splitButtonBehavior.ts
```diff
@@ -12,6 +12,7 @@ import menuButtonBehavior from '../MenuButton/menuButtonBehavior'
 const splitButtonBehavior: Accessibility = props => {
   const splitButtonMenuButtonBehavior = props => {
     const menuButtonBehaviorData = menuButtonBehavior(props)
+    menuButtonBehaviorData.attributes.trigger['aria-haspopup'] = undefined
```
mnajdova, line 25: Shouldn't we add more info in the @specification?

packages/accessibility/src/behaviors/SplitButton/splitButtonBehavior.ts
jurokapsiar: yes, menu can only contain menu items, nothing else.

packages/accessibility/src/behaviors/SplitButton/splitButtonBehavior.ts
kolaps33: as aria-popup shouldn't be add I am not writing to the description, that we are not adding it, but add the test to verify it :)

jurokapsiar (approved):

PR #2091, opened 2019-11-05
docs(Tooltip): add usage example for Tooltip on disabled elements

mnajdova:
Added usage example for tooltip on disabled button, for clarifying the issue: https://github.com/microsoft/fluent-ui-react/issues/2021


miroslavstastny (commented):
It will not work with keyboard navigation as there is nothing focusable. What it the expectation there?
@jurokapsiar can you please review?

jurokapsiar:
> It will not work with keyboard navigation as there is nothing focusable. What it the expectation there?
> @jurokapsiar can you please review?

it's ok - by default, we don't focus disabled elements. there's inconsistency between Windows (does not allow focusing) and Office (allows focusing), we say that we should not focus disabled by default, but if justified, disabled buttons can be focusable in specific scenarios - partners can achieve that on their side (by just using `aria-disabled` instead of `disabled` and potentially some style overrides

jurokapsiar (approved):

PR #2092, opened 2019-11-06
feat(List+Dropdown): wrapper for custom scrollbar in list

jurokapsiar:
Optional prop/slot for container in list which allows the Dropdown to integrate with custom scrollbars.
Dropdown focuses its list after open. As some custom scrollbar libraries render itself twice, the focusing was broken (first rendered element was focused, then removed)

Instead, a wrapper between the list and list item is added, which can be used to inject the custom scrollbars.


silviuaavram (approved):

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -697,16 +701,18 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
-            {...accessibilityMenuProps}
-            styles={styles.list}
-            tabIndex={search ? undefined : -1} // needs to be focused when trigger button is activated.
-            aria-hidden={!open}
-            onFocus={this.handleTriggerButtonOrListFocus}
-            onBlur={this.handleListBlur}
-            items={items}
-          />
+          {List.create(list || {}, {
+            overrideProps: {
```
mnajdova: Why are all these `overrideProps`? I would expect the callbacks to be in override and to invoke the props[callbacks], but not all of them.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -697,16 +701,18 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
-            className={Dropdown.slotClassNames.itemsList}
-            {...accessibilityMenuProps}
-            styles={styles.list}
-            tabIndex={search ? undefined : -1} // needs to be focused when trigger button is activated.
-            aria-hidden={!open}
-            onFocus={this.handleTriggerButtonOrListFocus}
-            onBlur={this.handleListBlur}
-            items={items}
-          />
+          {List.create(list || {}, {
```
mnajdova: Can we add in the `defaultProps` `{ list: {} }` instead of adding it here?

packages/react/src/components/List/List.tsx
```diff
@@ -135,7 +145,15 @@ class List extends AutoControlledComponent<WithAsProp<ListProps>, ListState> {
   }
 
   renderComponent({ ElementType, classes, accessibility, unhandledProps }) {
-    const { children } = this.props
+    const { children, container, items } = this.props
+
+    const hasContent = childrenExist(children) || (items && items.length > 0)
+    const containerContent = childrenExist(children) ? children : this.renderItems()
+
+    const maybeWrappedContent =
```
mnajdova: 😥😥😥  I don't like introducing a new idea for a slot that is not a wrapper, but it is wrapping the content, but I don't have any better idea of how this can be achieved with the current implementation. Can we implement this as a real wrapper slot (around the list), so that we can at least reuse the logic we have in other components (MenuItem, Input etc...), or it has to be between the list and the items. It looks strange that the generated structure would be
```
<ul>
  <divl>
    <li />
    <li />
...
```

jurokapsiar: I wanted to avoid having wrapper around the whole list. This causes the problem of not being sure if the unhandled props are passed to the first or second element. This way, the list is always top-most element.

Yes, DOM structure is wrong in this case. But I think we will need to switch to `<div>s` everywhere (even in menu and list) with a propper aria role just because of the custom scrollbar.

layershifter: What about `wrap` prop instead? It will be not so flexible, but without `Box`es and additional DOM elements :)

```tsx
  renderComponent({ ElementType, classes, accessibility, unhandledProps }) {
    const { children, items, wrap } = this.props
    const hasContent = childrenExist(children) || (items && items.length > 0)

    return (
      <ElementType
        {...rtlTextContainer.getAttributes({ forElements: [children] })}
        {...unhandledProps}
        className={classes.root}
      >
        {hasContent && wrap(childrenExist(children) ? children : this.renderItems())}
      </ElementType>
    )
  }
```

And then it can be used as:

```jsx
      <Dropdown
        search
        items={items}
        list={{ wrap: (children) => <Scrollbars style={{ height: '20rem' }}>{children}</Scrollbars> }}
      />
```

layershifter (approved):
👍

PR #2093, opened 2019-11-06
docs(prototypes): Added prototype for hexagonal avatars

mnajdova:
Added prototype for hexagonal avatar:
image


jurokapsiar (approved):

docs/src/prototypes/hexagonalAvatar/CustomAvatar.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react'
+import { Avatar as StardustAvatar, AvatarProps, callable } from '@stardust-ui/react'
+
+const Avatar = (props: AvatarProps & { hexagonal?: boolean }) => {
+  const { hexagonal, variables: propVariables, ...rest } = props
+  const variables = siteVars => {
+    return {
+      ...callable(propVariables)(siteVars),
+      ...(hexagonal && {
```
miroslavstastny: Are you sure with the order of `propsVariables` and the rest? I would say props variables should win.

docs/src/prototypes/hexagonalAvatar/CustomAvatar.tsx
```diff
@@ -0,0 +1,25 @@
...
+        width: '36px',
+        height: '32px',
+        clipPath: siteVars.hexClipPath,
+        borderRadius: '0px',
+        avatarBorderColor: '',
+        avatarBorderWidth: 0,
+      }),
+    }
+  }
+  return <StardustAvatar {...rest} variables={variables} />
```
miroslavstastny: You should use `mergeComponentVariables` here instead of the callable above.

mnajdova: But @miroslavstastny  we are not exporting `mergeComponentVariables` currently, that's why I needed to do this... I can update it once we export this.

packages/react/src/index.ts
```diff
@@ -139,6 +139,9 @@ export { default as PopupContent } from './components/Popup/PopupContent'
 export * from './components/Portal/Portal'
 export { default as Portal } from './components/Portal/Portal'
 
+export * from './components/Portal/PortalInner'
+export { default as PortalInner } from './components/Portal/PortalInner'
```
jurokapsiar: I think it would be easier to just use `createPortal` in the prototype. We would need to add more documentation if we start exposing `PortalInner`

docs/src/prototypes/hexagonalAvatar/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+          <defs role="presentation">
+            <clipPath id="bot-hexagon-clip-path">
+              <path d="M23.93,2H10.07C9,2,8.01,2.57,7.47,3.5l-6.92,12c-0.54,0.93-0.54,2.07,0,3l6.93,12c0.54,0.93,1.53,1.5,2.6,1.5h13.85c1.07,0,2.06-0.57,2.6-1.5l6.93-12c0.54-0.93,0.54-2.07,0-3l-6.93-12C25.99,2.57,25,2,23.93,2z" />
+            </clipPath>
+          </defs>
+        </svg>
+      </PortalInner>
+      <Flex column padding="padding.medium" gap="gap.medium">
+        <div>
+          <Text content="Hexagonial Avatar" />
```
miroslavstastny: nit
```suggestion
          <Text content="Hexagonal Avatar" />
```

packages/react/src/index.ts
mnajdova: Replaced

PR #2094, opened 2019-11-06
[Abandoned] - fix(Splitbutton) accessibility - closing menu when opening with virtual cursor

kolaps33:
the PR contains fix for following bugs:

https://github.com/microsoft/fluent-ui-react/issues/2082


packages/react/src/components/Popup/Popup.tsx
```diff
@@ -571,9 +571,9 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
   trySetOpen(newValue: boolean, eventArgs: any) {
     const isOpenedByRightClick = newValue && isRightClick(eventArgs)
 
-    // when new state 'open' === 'true', save the last focused element
+    this.updateTriggerFocusableDomElement()
```
silviuaavram, line 574: why would this fix the issue?

jurokapsiar (approved):

packages/react/src/components/Popup/Popup.tsx
kolaps33: This should fix the issue as it will updated DOM element when popup will be closed. 
But nevertheless I figure if we move it here then following scenario is broken:
1.  in popup example "Focus Trap"
2.  open popup from button "Trap focus on appearance" click somewhere to popup 
3. press ESC to close popup
Expected: in production popup is closed and focus goes back to trigger button
Actual: with our change the focus is not going back to trigger button. 
I need to investigate more what is going on.

kolaps33: I spoke with @mnajdova and she is recommending to change the way how we opening the splitButton 
similarly like MenuButton.tsx has:

>       open: this.state.open,
>       onOpenChange: this.handleOpenChange,

then the same splitButton should have onOpenChange in overrideProps

PR #2097, opened 2019-11-06
perf: styles lazy evaluation

miroslavstastny:
Previously `renderComponent` resolved styles and add CSS classes for all slots in a rendered component.

There are components which have a lot of slots (`ChatMessage` currently consists of 7 slots) but not all are always used. There are even components which have mutually exclusive slots - `checkbox` vs `toggle` slots in `Checkbox` component.
In this PR, styles and CSS classes are computed only when used (that also means only IF used).

This can bring up to 10% rendering perf improvements.

Before (master):

| Example                  | Min    | Avg    | Median | Max    |
| ------------------------ | ------ | ------ | ------ | ------ |
| Chat.perf.tsx            | 378.88 | 459.88 | 511.41 | 566.8  |
| ChatWithPopover.perf.tsx | 386.31 | 457.55 | 417.21 | 623.64 |

After (this PR):

| Example                  | Min    | Avg    | Median | Max    |
| ------------------------ | ------ | ------ | ------ | ------ |
| Chat.perf.tsx            | 312.24 | 396.36 | 438.55 | 548.93 |
| ChatWithPopover.perf.tsx | 354.09 | 413.74 | 374.05 | 497.5  |


packages/react/src/components/Chat/ChatMessage.tsx
```diff
@@ -209,14 +209,16 @@ class ChatMessage extends UIComponent<WithAsProp<ChatMessageProps>, ChatMessageS
...
-      defaultProps: {
-        [IS_FOCUSABLE_ATTRIBUTE]: true,
-        accessibility: menuAsToolbarBehavior,
-        className: ChatMessage.slotClassNames.actionMenu,
-        styles: styles.actionMenu,
-      },
-    })
+    const actionMenuElement =
+      actionMenu &&
+      Menu.create(actionMenu, {
```
miroslavstastny: Rather this or `defaultProps` as function or?

packages/react/src/lib/renderComponent.tsx
```diff
@@ -238,22 +322,12 @@ const renderComponent = <P extends {}>(
...
-    if (process.env.NODE_ENV !== 'production' && isDebugEnabled) {
-      resolvedStylesDebug[slotName] = resolvedStyles[slotName]['_debug']
-      delete resolvedStyles[slotName]['_debug']
-    }
-
-    if (renderer) {
-      classes[slotName] = renderer.renderRule(callable(resolvedStyles[slotName]), felaParam)
-    }
-  })
+  const { resolvedStyles, resolvedStylesDebug, classes } = resolveStylesAndClasses(
```
miroslavstastny, line 245: Is it better to have this inline or in a separate function (with strange in/out API)?

levithomason: Yea, these args are odd and leaky:

- `renderer` is only used for `renderer.renderRule`.  We should just take in a single function here and pass `renderer.renderRule` in as something like `renderStyles`
- `felaParam` should probably be something like `renderStylesParam`

Ideally, this would be a function which you could use with any style function and/or renderer.  Crafting it in a way where it isn't leaking fela implementation details would be ideal.

packages/react/src/components/Chat/ChatMessage.tsx
miroslavstastny: function

packages/react/src/lib/renderComponent.tsx
miroslavstastny: `renderer.renderRule` -> `renderStyles`
`felaParam` removed - it is `renderStyle`'s responsibility

levithomason (approved):
👍

PR #2104, opened 2019-11-12
fix(Dialog): Content should have 100% width

### LINKED ISSUE #2038: <Dialog /> content does not stretch to the full width, as it should
## Bug Report

Please, check the codepen to see the issue in practice https://codesandbox.io/s/stardust-ui-example-q9lfw

### Expected Result
I have passed in some content to the dialog property that should stretch to the full width of the dialog automatically.

### Actual Result
Content does not stretch and the only way to make it stretch is to add some overrides 💣

Screenshot 2019-10-17 at 17 18 46

Please, notice that almost every override for the content in dialog-styles has "width: 100%"

### Version
x.y.z

### Testcase

jurokapsiar:
Fixes #2038


packages/react/src/themes/teams/components/Dialog/dialogStyles.ts
```diff
@@ -32,6 +32,7 @@ export default {
     gridColumn: '1 / span 2',
     gridRow: 2,
     justifySelf: 'left',
+    width: '100%',
```
mnajdova, line 35: Did we have some screener diff after this changes? If not, let's add some tests for catching regressions for this in future.

mnajdova (approved):

miroslavstastny (approved):

PR #2108, opened 2019-11-12
fix(SplitButton): Updating splitButton component to match Teams redlines

notandrew:
Updating the split button layout to match the Teams redlines:

before:
image

after:
image

before:
image

after:

image

new:
image


packages/react/src/components/SplitButton/SplitButton.tsx
```diff
@@ -202,13 +205,13 @@ class SplitButton extends AutoControlledComponent<WithAsProp<SplitButtonProps>,
...
           defaultProps: () => ({
+            styles: styles.toggleButton,
             className: SplitButton.slotClassNames.toggleButton,
             disabled,
-            icon: 'stardust-arrow-down',
+            icon: 'chevron-down-medium',
             iconOnly: true,
             primary,
             secondary,
-            ...accessibility.attributes.toggleButton,
```
notandrew: Removing this fixes some existing accessibility issues where the toggle button isn't focusable or clickable, but I'm not sure what this was doing...

packages/react/src/themes/teams/components/SplitButton/splitButtonStyles.ts
```diff
@@ -1,54 +1,52 @@
...
-      '::after': { border: 0 },
+    ...(p.small && {
+      height: v.smallDimension,
+      padding: v.smallPadding,
+      minWidth: v.smallMinWidth,
+      boxShadow: v.smallBoxShadow,
+    }),
+
+    ':active': {
+      animationName: '0',
```
notandrew: This probably isn't ideal but makes it so the button doesn't animate when clicked.

packages/react/src/components/SplitButton/SplitButton.tsx
silviuaavram: Easter egg :)

First of all the PR should fix styles, not accessibility. Secondly, the specs we had when developing the button was that the toggle button should not be tabbable by itself, only clickable. Just the main button should be tabbable, and the keyboard focus from that main button should apply a style on the whole split button (main + toggle).

@mnajdova Don't we have this covered in our isConformant tests? Or can we do it?

packages/react/src/components/SplitButton/SplitButton.tsx
```diff
@@ -84,6 +84,9 @@ export interface SplitButtonProps
   /** A split button can be formatted to show different levels of emphasis. */
   secondary?: boolean
 
+  /** A split button can be small */
+  small?: boolean
```
silviuaavram: add this to propTypes as well.

docs/src/examples/components/SplitButton/Types/SplitButtonExampleSmall.shorthand.tsx
```diff
@@ -0,0 +1,26 @@
...
+        { key: 'group', content: 'New group message' },
+        { key: 'channel', content: 'New channel message' },
+      ]}
+      button={{
+        content: 'small',
+        'aria-roledescription': 'splitbutton',
+        'aria-describedby': 'instruction-message',
+      }}
+      small
+      toggleButton={{ 'aria-label': 'options' }}
```
silviuaavram: label should be a bit more descriptive, like `more options` or `expand options menu` ...

silviuaavram (changes requested):
Thank you for fixing the styles! Please take a look at the other comments as well, especially the a11y one.

packages/react/src/themes/teams/components/SplitButton/splitButtonStyles.ts
mnajdova: `animationName: 'unset'` should be used here, and please unser the other animation* related props

packages/react/src/components/SplitButton/SplitButton.tsx
mnajdova: What would we cover in the isConformant test regarding this? It seems to me that this should be covered in the unit tests for the component (if the behavior is changed).

packages/react/src/components/SplitButton/SplitButton.tsx
notandrew: done

packages/react/src/components/SplitButton/SplitButton.tsx
```diff
@@ -202,9 +206,10 @@ class SplitButton extends AutoControlledComponent<WithAsProp<SplitButtonProps>,
         )}
         {Button.create(toggleButton, {
           defaultProps: () => ({
+            styles: styles.toggleButton,
             className: SplitButton.slotClassNames.toggleButton,
             disabled,
-            icon: 'stardust-arrow-down',
+            icon: 'chevron-down-medium',
```
codepretty: I think you should use `stardust-menu-arrow-down`

image

packages/react/src/components/SplitButton/SplitButton.tsx
codepretty: Seems like this might be better as an optional `size` prop. Right now we only need to allow for small and default sizes (and default without specifying), but it would allow flexibility to have more sizes for splitbutton in the future w/o adding more props

packages/react/src/themes/teams/components/SplitButton/splitButtonStyles.ts
```diff
@@ -1,51 +1,90 @@
...
   }),
+
+  toggleButton: ({ props: p, variables: v }): ICSSInJSStyle => ({
+    borderTopLeftRadius: 0,
+    borderBottomLeftRadius: 0,
+    borderColor: v.borderColor,
+    ...getIconFillOrOutlineStyles({ outline: true }),
+
+    ...(p.primary && {
+      borderWidth: v.borderWidthLeftOnly,
```
codepretty: Is `borderWidth` only needed for primary and disabled buttons?

packages/react/src/themes/teams/components/SplitButton/splitButtonVariables.ts
```diff
@@ -1,29 +1,30 @@
...
   borderRadius: string
-  backgroundColorFocus: string
-  boxShadow: string
+  borderColorPrimary: string
   borderColor: string
-  colorFocus: string
-  primaryBackgroundColorFocus: string
-  primaryColorFocus: string
+  borderColorDisabled: string
+  borderWidthLeftOnly: string
```
codepretty: This variable name is specific to this implementation. If there is a need to have only a left border, then I'd put this directly in the styles file. If this needs to be a variable, it should be named something that only allows changing the left border, eg

`borderWidthLeftOnly: .1rem,`

and in styles files..
`borderWidth: 0 0 0 ${borderWidthLeftOnly},`

also, should we be consistent with using `pxToRem` or not in these situations?

packages/react/src/themes/teams/components/SplitButton/splitButtonVariables.ts
```diff
@@ -1,29 +1,30 @@
...
   return {
     borderRadius: siteVars.borderRadius,
-    backgroundColorFocus: siteVars.colors.grey[200],
-    boxShadow: siteVars.shadowLevel1,
-    borderColor: siteVars.colors.grey[200],
-    colorFocus: siteVars.colors.grey[750],
-    primaryBackgroundColorFocus: siteVars.colors.brand[800],
-    primaryColorFocus: siteVars.colors.white,
+    borderColor: siteVars.colorScheme.default.border,
+    borderColorPrimary: siteVars.colors.brand[500],
```
codepretty, line 20: Should `borderColorPrimary` be a scheme value to change in different themes?

packages/react/src/themes/teams/components/SplitButton/splitButtonVariables.ts
```diff
@@ -1,29 +1,30 @@
...
-  backgroundColorFocus: string
-  boxShadow: string
+  borderColorPrimary: string
   borderColor: string
-  colorFocus: string
-  primaryBackgroundColorFocus: string
-  primaryColorFocus: string
+  borderColorDisabled: string
+  borderWidthLeftOnly: string
+  smallDimension: string
```
codepretty, line 9: I don't see a dimension variable for the default button size. If we need one for `small` size should there be one for the default size?

notandrew: It's inherited by Button's size, so I think we just keep it as-is unless we remove the dependency on Button...

packages/react/src/themes/teams/components/SplitButton/splitButtonStyles.ts
notandrew: Kind of the opposite, only primary & disabled buttons only have the left side border, secondary buttons have borders on all sides. 

borders on all sides of the toggle:
image

one border on the left of the toggle:
image

packages/react/src/themes/teams/components/SplitButton/splitButtonVariables.ts
notandrew: I switched to using borderWidth in SiteVariables and that's px instead of rem, so I'm not sure what we should be using

packages/react/src/themes/teams/components/SplitButton/splitButtonVariables.ts
notandrew: It's ok as-is but also brand[500] doesn't match any colorscheme...

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### Fixes
+- Updating `splitButton` component to match Teams redlines @notandrew ([#2108](https://github.com/microsoft/fluent-ui-react/pull/2108))
```
mnajdova:
```suggestion
- Update `SplitButton` styles in Teams theme [redlines] @notandrew ([#2108](https://github.com/microsoft/fluent-ui-react/pull/2108))
```

docs/src/examples/components/SplitButton/Types/index.tsx
```diff
@@ -14,6 +14,11 @@ const SplitButton = () => (
       description="A SplitButton can be formatted to show primary level of emphasis."
       examplePath="components/SplitButton/Types/SplitButtonExamplePrimary"
     />
+    <ComponentExample
+      title="small Split Button"
```
mnajdova:
```suggestion
      title="Small"
```

docs/src/examples/components/SplitButton/Types/index.tsx
```diff
@@ -14,6 +14,11 @@ const SplitButton = () => (
...
+      description="small splitbutton"
```
mnajdova, line 19: Actually, can we follow the size prop that is already in the Button component?

packages/react/src/components/SplitButton/SplitButton.tsx
```diff
@@ -84,6 +85,9 @@ export interface SplitButtonProps
   /** A split button can be formatted to show different levels of emphasis. */
   secondary?: boolean
 
+  /** A split button can be small */
+  size?: string
```
mnajdova:
```
  /** A split button can be sized. */
  size?: SizeValue
```

packages/react/src/components/SplitButton/SplitButton.tsx
```diff
@@ -118,14 +122,15 @@ class SplitButton extends AutoControlledComponent<WithAsProp<SplitButtonProps>,
     onMenuItemClick: PropTypes.func,
     onOpenChange: PropTypes.func,
     open: PropTypes.bool,
+    size: PropTypes.string,
```
mnajdova: Please follow the size implementation in the Button component

docs/src/examples/components/SplitButton/Types/index.tsx
notandrew: Updated it

codepretty:
Why did this example not get the updated caret?
image

codepretty (approved):

notandrew:
> Why did this example not get the updated caret?
> image

We don't have an up-facing caret of the same type as the down-caret. Not sure if we use a transform to fix this or check in a new icon...

notandrew:
> Why did this example not get the updated caret?
> image

fixed it by rotating the icon

silviuaavram (approved):

PR #2112, opened 2019-11-13
Preparing for API Extractor integration

ecraig12345:
Fabric uses [API Extractor](https://api-extractor.com) to help review API changes and generate docs. @kenotron and I would like to introduce it for FluentUI too, so we can use the API review functionality for converged components (and optionally existing components). It will also be a faster replacement for `react-docgen-typescript`.

**This PR doesn't actually enable API Extractor or change anything about doc generation**, but it introduces the config files and some required formatting and export changes. (I found these issues by running `npx @microsoft/api-extractor run <config>` in each project.)

### Formatting

API Extractor requires that `@param` and `@returns` **do not** contain type annotations, and requires the parameter name in `@param` to be followed by a `-` (example: `@param foo - description here`). So I added dashes, got rid of all the remaining type annotations, and moved the type info to the function signature itself in cases where it wasn't already there.

### Exports (and Babel upgrade)

(These more significant changes are grouped in the [first commit](https://github.com/microsoft/fluent-ui-react/commit/10c3228201f2670d74d17b8797731daec2e76924) for easy viewing.)

There were two re-exports which caused problems, because API Extractor doesn't currently support this type of re-export (from `packages/react/src/lib/index.ts`):
```ts
import * as commonPropTypes from './commonPropTypes'
export { commonPropTypes }
```

A workaround is to export a named module instead:
```ts
import {
  CreateCommonConfig as CreateCommonConfigLocal,
  createCommon as createCommonLocal,
} from './commonPropTypes'

export module commonPropTypes {
  export type CreateCommonConfig = CreateCommonConfigLocal
  export const createCommon = createCommonLocal
}
```

However, namespaces weren't supported by Babel until very recent versions. So I also had to upgrade Babel (and enable namespace support: `['@babel/preset-typescript', { allowNamespaces: true }]`), which then required upgrading `react-hot-loader` to eliminate a very spammy warning introduced by the new Babel version.

I'd appreciate help/suggestions for verifying that the Babel upgrade didn't introduce any issues.

(The other problematic export was in `packages/react/src/index.ts` and was much simpler to deal with since it involved exporting values not types.)


kenotron (approved):
This looks good to me... do you have some good numbers to drop here about how much faster this is?

ecraig12345:
> This looks good to me... do you have some good numbers to drop here about how much faster this is?

Nothing is faster yet, since it's not actually enabled (because I haven't written all the integration parts to write the output in the *.info.json format). Hopefully I'll have numbers sometime next week.

PR #2115, opened 2019-11-15
feat(Animation): add logic for mount/remove children

mnajdova:
This PR adds the option on the `Animation` component, to wait with the mounting of the component until it is visible, as well as remove the dom element if the element is hidden.

For this, the `react-transition-group` dependency was added, and the `Transition` component is used behind the scenes from the `Animation` component. The API exposed is partially the Transition's API.

There are example usages in the changes.

**BREAKING CHANGE**
Other change is that, the Animation component is **not** rendering the additional div anymore, it is rendering only the children.


packages/react/src/components/Animation/Animation.tsx
```diff
@@ -96,30 +158,184 @@ class Animation extends UIComponent<WithAsProp<AnimationProps>, any> {
...
+
+    if (mountOnShow || unmountOnHide) return { initial: Animation.UNMOUNTED }
+    return { initial: Animation.EXITED }
+  }
+
+  componentDidMount() {
+    this.updateStatus()
+  }
+
+  UNSAFE_componentWillReceiveProps(nextProps) {
```
layershifter: Do not use that code from SUIR, it's buggy and incompatible with React's Strict and Concurrent modes 😞

mnajdova: Will refactor it (I noticed one bug so far) or will try to use https://reactjs.org/docs/animation.html

packages/react/src/components/Animation/Animation.tsx
```diff
@@ -102,23 +127,65 @@ class Animation extends UIComponent<WithAsProp<AnimationProps>, any> {
...
+        unmountOnExit={unmountOnExit}
+        timeout={timeout}
+        onEnter={onEnter}
+        onEntering={onEntering}
+        onEntered={onEntered}
+        onExit={onExit}
+        onExiting={onExiting}
+        onExited={onExited}
+      >
+        {state => result}
```
mnajdova: I am currently not using the state, as in the usage I am changing the animation. We don't have the children as function concept anywhere else so far, so I didn't add it for the Animation component. (it would be easy to migrate if we decide in the future!)

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -251,9 +252,18 @@ class MenuItem extends AutoControlledComponent<WithAsProp<MenuItemProps>, MenuIt
         </ElementType>
       </Ref>
     )
-    const maybeSubmenu =
-      menu && active && menuOpen ? (
-        <>
+    const submenuVisible = menu && active && menuOpen
+
+    // TODO: allow the users to define these via props
```
mnajdova: These changes won't be part of the final PR, they are just showcasing the usage in a real example

packages/react/test/specs/commonTests/isConformant.tsx
```diff
@@ -27,6 +27,7 @@ export interface Conformant {
   exportedAtTopLevel?: boolean
   rendersPortal?: boolean
   wrapperComponent?: React.ReactType
+  handlesAsProp?: boolean
```
mnajdova, line 36: As the Animation is not rendering root element, but only copies the children, we don't have the "as" prop there.

packages/react/package.json
```diff
@@ -26,7 +26,8 @@
     "popper.js": "^1.15.0",
     "prop-types": "^15.7.2",
     "react-fela": "^10.6.1",
-    "react-is": "^16.6.3"
+    "react-is": "^16.6.3",
+    "react-transition-group": "^4.3.0"
```
mnajdova, line 30: New dependency added

packages/react/src/components/Animation/Animation.tsx
```diff
@@ -77,6 +75,39 @@ export interface AnimationProps
...
+  appear?: boolean
+
+  /** Wait until the first "enter" transition to mount the component (add it to the DOM). */
+  mountOnEnter?: boolean
+
+  /** Unmount the component (remove it from the DOM) when it is not shown. */
+  unmountOnExit?: boolean
+
+  /** The duration of the transition, in milliseconds. */
+  timeout?: number | { enter?: number; exit?: number; appear?: number }
```
codepretty, line 94: based on your comment, should `timeout` actually be named `duration`?

codepretty: ah, nevermind. i see that it is what it's called in `react-transition-group`... https://reactcommunity.org/react-transition-group/transition#Transition-prop-timeout

packages/react/src/components/Animation/Animation.tsx
codepretty: Does this work for adding/removing table items within the Table component or tree items within Tree?

packages/react/src/components/Animation/Animation.tsx
mnajdova: Yeah, this two are different concepts (duration is the css animationDuraion), and this is timeout for removing the elements. Maybe I should come up with a better name...

packages/react/src/components/Animation/Animation.tsx
mnajdova: It should work, if not, we can Implement the AnimationGroup component in a similar way of how the TransitionGroup (https://reactcommunity.org/react-transition-group/transition-group) component is implemented from the same library). But, to be honest, I would try to avoid adding different APIs and try to implement it with this API

packages/react/src/lib/positioner/Popper.tsx
```diff
@@ -174,6 +175,8 @@ const Popper: React.FunctionComponent<PopperProps> = props => {
           placement: computedPlacement,
           scheduleUpdate,
         })
+      : children && rest && Object.keys(rest).length > 0
+      ? React.cloneElement(React.Children.only(children as React.ReactElement<any>), rest)
```
layershifter: Looks complicated, is a there any to simplify it?

packages/react/src/components/Menu/MenuItem.tsx
```diff
@@ -251,9 +252,18 @@ class MenuItem extends AutoControlledComponent<WithAsProp<MenuItemProps>, MenuIt
...
+    const maybeSubmenu = menu ? (
+      <>
+        <Animation
+          visible={submenuVisible}
+          timeout={2000}
```
layershifter: Should be this timeout defined in the theme?

packages/react/src/components/Animation/Animation.tsx
```diff
@@ -102,23 +134,65 @@ class Animation extends UIComponent<WithAsProp<AnimationProps>, any> {
...
+      PropTypes.shape({
+        appear: PropTypes.number,
+        enter: PropTypes.number,
+        exit: PropTypes.number,
+      }),
+    ]),
+  }
+
+  static defaultProps = {
+    timeout: 0,
```
layershifter, line 187: Is it a proper default?

packages/react/src/components/Animation/Animation.tsx
```diff
@@ -102,23 +134,65 @@ class Animation extends UIComponent<WithAsProp<AnimationProps>, any> {
...
+        unmountOnExit={unmountOnExit}
+        timeout={timeout}
+        onEnter={onEnter}
+        onEntering={onEntering}
+        onEntered={onEntered}
+        onExit={onExit}
+        onExiting={onExiting}
+        onExited={onExited}
+      >
+        {state => result}
```
layershifter: https://github.com/reactjs/react-transition-group/blob/318db104c7ba4f279392111c11397d2f053594e7/src/Transition.js#L365-L367

nit: As I see we can do something like this:

```jsx
return (
  <Transition
    {...unhandledProps}
    className={cx(classes.root, child.props.className)}
 >
    {children}
 </Transition>
)
```

packages/code-sandbox/src/createCallbackLogFormatter.ts
```diff
@@ -16,7 +16,7 @@ const createCallbackLogFormatter = (props: string[] = []) => (
 
   return [
     `/*[${new Date().toLocaleTimeString()}]*/`,
-    `${name} (e: { "type": "${e.type}" }, data: ${formatCode(
+    `${name} (e: { "type": "${e && e.type}" }, data: ${formatCode(
```
layershifter:
```suggestion
    `${name} (e: { "type": "${e ? e.type : 'NA'}" }, data: ${formatCode(
```
 nit

docs/src/examples/components/Animation/Usage/AnimationExampleVisible.shorthand.tsx
```diff
@@ -0,0 +1,65 @@
+import * as React from 'react'
+import { Provider, Animation, Icon, Button } from '@stardust-ui/react'
+import { useLogKnob } from '@stardust-ui/docs-components'
+import { easeEasy } from 'src/themes/teams/animations/timingFunctions'
+
+const AnimationExampleVisible = () => {
+  const [visible, setVisible] = React.useState(false)
+
+  const onEnter = useLogKnob('onEnter', () => {})
```
layershifter:
```suggestion
  const onEnter = useLogKnob('onEnter')
```

I missed that thing, let's update `useLogKnob()`:

```diff
  const proxy = React.useCallback<any>(
    (...a) => {
      appendLog(formatter(name, ...a))
-      return typeof callback === 'function' ? (callback as any)(...a) : null   
+      if (typeof callback === 'function') {
+        return (callback as any)(...a)
+      }
+
+      if (process.env.NODE_ENV !== 'production') {
+       console.log('Please provide a function to "useLogKnob(name, callback)"')
+      }
+      
+      return null
    },
    [appendLog, callback, name, formatter],
  )
```

docs/src/examples/components/Animation/Types/AnimationExampleDirection.tsx
```diff
@@ -21,18 +21,26 @@ const AnimationExampleDirection = () => (
...
-      <Animation name="spinner" direction="reverse">
-        <Icon name="mention" circular bordered />
-      </Animation>
-      <Animation name="spinner" direction="alternate">
-        <Icon name="mention" circular bordered />
-      </Animation>
-      <Animation name="spinner" direction="alternate-reverse">
-        <Icon name="mention" circular bordered />
-      </Animation>
+      <div>
```
layershifter, line 24: This change looks strange...

layershifter (commented):
There are no better options now than `react-transition-group` 🤔 However, it can be an issue that child components will rerender during state transitions too frequently...

ecraig12345:
Hi, just a heads up that I made a change #2153 renaming all `lib` folders to `utils`. So the changes to `lib` files in this PR will need to be moved to the new location.

packages/react/src/lib/positioner/Popper.tsx
mnajdova: I can extract the condition to a separate const or function, but I guess is more clear to see it inlined, instead of scrolling back.. As for simplifying the condition, I don't see how. Do you have some suggestion? Btw the same condition is in the Ref component already.

packages/react/src/components/Menu/MenuItem.tsx
mnajdova: I believe so, we should have a way of defining timeouts in the theme in some structured way and reuse those (maybe in a similar way as we have some required icons... Haven't thought much about this, but generally I agree) But would not tackle this for now, as the `MenuItem` changes will not go as part of this PR, it is just a showcase of how it can be used and a prove of concept that it works. Good call though!

packages/react/src/components/Animation/Animation.tsx
mnajdova: Yep, this is the default in Transition as well (`0` for both entering and exiting) and it basically means no removing/mounting, the element will always be in the dom.

docs/src/examples/components/Animation/Types/AnimationExampleDirection.tsx
mnajdova: Previously the Animation component was creating a div, so the styles from the Grid were applied on it. Now those are applied on the icons and it is breaking the animation. If the Icon is used outside of Grid, there is no need for additional div.

CHANGELOG.md
```diff
@@ -23,6 +23,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - `mode` property from `focusZone` configuration in accessibility behaviors is no longer supported - the focus zone will always be in embed mode @layershifter ([#2265](https://github.com/microsoft/fluent-ui-react/pull/2265))
 - `FocusZoneMode` and `FOCUSZONE_WRAP_ATTRIBUTE` are no longer exported @layershifter ([#2265](https://github.com/microsoft/fluent-ui-react/pull/2265))
 - Changed `avatarBorderWidth` and `statusBorderWidth` avatar variables types from number to string and updated styles in Teams theme @mnajdova ([#2238](https://github.com/microsoft/fluent-ui-react/pull/2238))
+- Add logic for mounting/removing elements when they are shown/hidden using the `Animation` component @mnajdova ([#2115](https://github.com/microsoft/fluent-ui-react/pull/2115))
```
layershifter: Is it a breaking change?

packages/react/src/components/Animation/Animation.tsx
```diff
@@ -77,9 +75,45 @@ export interface AnimationProps
...
+  onEnter?: (node: HTMLElement, isAppearing: boolean) => void
+
+  /** Callback fired after the "entering" status is applied. An extra parameter isAppearing is supplied to indicate if the enter stage is occurring on the initial mount. */
+  onEntering?: (node: HTMLElement, isAppearing: boolean) => void
+
+  /** Callback fired after the "entered" status is applied. An extra parameter isAppearing is supplied to indicate if the enter stage is occurring on the initial mount. */
+  onEntered?: (node: HTMLElement, isAppearing: boolean) => void
+
+  /** Callback fired before the "exiting" status is applied. */
+  onExit?: (node: HTMLElement) => void
```
layershifter:
```suggestion
  onExit?: () => void
```

Do we want to expose `node` there? 🤔

packages/react/src/components/Animation/Animation.tsx
```diff
@@ -89,9 +123,11 @@ class Animation extends UIComponent<WithAsProp<AnimationProps>, any> {
   static propTypes = {
     ...commonPropTypes.createCommon({
       accessibility: false,
+      as: false,
       content: false,
-      children: 'element',
+      children: false,
     }),
+    children: PropTypes.oneOfType([PropTypes.func, PropTypes.element]),
```
layershifter, line 156: 👍

packages/react/src/components/Animation/Animation.tsx
```diff
@@ -101,28 +137,69 @@ class Animation extends UIComponent<WithAsProp<AnimationProps>, any> {
...
+    appear: PropTypes.bool,
+    unmountOnExit: PropTypes.bool,
+    timeout: PropTypes.oneOfType([
+      PropTypes.number,
+      PropTypes.shape({
+        appear: PropTypes.number,
+        enter: PropTypes.number,
+        exit: PropTypes.number,
+      }),
+    ]),
```
layershifter, line 177: Callbacks should be also listed there

packages/react/src/components/Animation/Animation.tsx
mnajdova: This is the API from the Transition component itself, I would rather not change anything for now...

CHANGELOG.md
mnajdova: It is breaking, but mostly because the Animation component is not rendering div anymore, so maybe that should be mention here too.. Will add it

packages/react/src/components/Animation/Animation.tsx
```diff
@@ -101,28 +137,69 @@ class Animation extends UIComponent<WithAsProp<AnimationProps>, any> {
...
+        onExiting={onExiting}
+        onExited={onExited}
+        {...unhandledProps}
+        className={
+          typeof children !== 'function' ? cx(classes.root, (child as any).props.className) : ''
+        }
+      >
+        {typeof children !== 'function'
+          ? child
+          : () => (children as AnimationChildrenProp)({ classes: classes.root })}
```
layershifter: nit: I would prefer an opposition condition:

```
        {typeof children === 'function'
          ? () => (children as AnimationChildrenProp)({ classes: classes.root })} : child
```

packages/react/src/utils/positioner/Popper.tsx
```diff
@@ -186,6 +187,8 @@ const Popper: React.FunctionComponent<PopperProps> = props => {
           placement: computedPlacement,
           scheduleUpdate,
         })
+      : children && rest && Object.keys(rest).length > 0
+      ? React.cloneElement(React.Children.only(children as React.ReactElement<any>), rest)
```
layershifter: Do we still need this change?

mnajdova: I will separate this to different PR, but it will be required in the future

packages/react/src/components/Animation/Animation.tsx
mnajdova: But maybe we should return here null as first argument for the event, and the node afterwards, to be consistent with the other callbacks. What do you think?

Maybe I am over-complicating, but let me know what you think @layershifter

layershifter (approved):

PR #2118, opened 2019-11-18
WIP - fix(Button): Fix text icon only button styles

codepretty:
1. Icons in buttons should not be filled on focus, only on hover

**Before**
image

**After**
image

2. Icon only text button should have a square background and black text color on hover

**Before**
image

**After**
image


packages/react/src/themes/teams/components/Button/buttonVariables.ts
```diff
@@ -98,7 +98,7 @@ export default (siteVars: any): ButtonVariables => ({
   circularBorderRadius: pxToRem(999),
 
   textColor: siteVars.colorScheme.default.foreground1,
-  textColorHover: siteVars.colorScheme.brand.foreground1,
+  textColorHover: siteVars.colorScheme.brand.foreground1, // todo: semantically this should be a hover token, should match same hover color token as hyperlinks
```
codepretty, line 101: Need to check with Daisy on what the correct value should be

packages/react/src/themes/teams/components/Attachment/attachmentStyles.ts
```diff
@@ -68,19 +67,12 @@ const attachmentStyles: ComponentSlotStylesPrepared<AttachmentProps, AttachmentV
     })
 
     return {
-      [`& .${Icon.className}`]: {
-        color: v.textColor, // this breaks the color change on hover
```
codepretty: I can't see why this was needed. Would love to know if I'm missing something. Here is the original PR where it was added... https://github.com/microsoft/fluent-ui-react/commit/eb47a325c0a410060ec99a66ae3702c4ba426107

packages/react/src/themes/teams-high-contrast/components/Button/buttonVariables.ts
```diff
@@ -3,6 +3,8 @@ import { ButtonVariables } from '../../../teams/components/Button/buttonVariable
 export interface ButtonHighContrastVariables {
   backgroundColorFocus: string
   primaryBackgroundColorFocus: string
+  textColorIconOnly: string
```
codepretty, line 6: Is it a good idea to have variables that are only used in high contrast theme? Seems like we should actually remove this whole section...?

notandrew (approved):

PR #2121, opened 2019-11-20
Preparing for API Extractor integration

ecraig12345:
(now from a branch instead of a fork) 

Fabric uses [API Extractor](https://api-extractor.com) to help review API changes and generate docs. @kenotron and I would like to introduce it for FluentUI too, so we can use the API review functionality for converged components (and optionally existing components). It will also be a faster replacement for `react-docgen-typescript`.

**This PR doesn't actually enable API Extractor or change anything about doc generation**, but it introduces the config files and some required formatting and export changes. (I found these issues by running `npx @microsoft/api-extractor run <config>` in each project.)

### Formatting

API Extractor requires that `@param` and `@returns` **do not** contain type annotations, and requires the parameter name in `@param` to be followed by a `-` (example: `@param foo - description here`). So I added dashes, got rid of all the remaining type annotations, and moved the type info to the function signature itself in cases where it wasn't already there.

### Exports (and Babel upgrade)

(These more significant changes are grouped in the [first commit](https://github.com/microsoft/fluent-ui-react/commit/10c3228201f2670d74d17b8797731daec2e76924) for easy viewing.)

There were two re-exports which caused problems, because API Extractor doesn't currently support this type of re-export (from `packages/react/src/lib/index.ts`):
```ts
import * as commonPropTypes from './commonPropTypes'
export { commonPropTypes }
```

A workaround is to export a named module instead:
```ts
import {
  CreateCommonConfig as CreateCommonConfigLocal,
  createCommon as createCommonLocal,
} from './commonPropTypes'

export module commonPropTypes {
  export type CreateCommonConfig = CreateCommonConfigLocal
  export const createCommon = createCommonLocal
}
```

However, namespaces weren't supported by Babel until very recent versions. So I also had to upgrade Babel (and enable namespace support: `['@babel/preset-typescript', { allowNamespaces: true }]`), which then required upgrading `react-hot-loader` to eliminate a very spammy warning introduced by the new Babel version.

I'd appreciate help/suggestions for verifying that the Babel upgrade didn't introduce any issues.

(The other problematic export was in `packages/react/src/index.ts` and was much simpler to deal with since it involved exporting values not types.)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2121)


docs/src/utils/getComponentPathname.ts
```diff
@@ -3,9 +3,9 @@ import * as _ from 'lodash'
 /**
  * Returns a pathname for a given component.
  *
- * @param {object} info - A Component's *.info.json
- * @returns {string}
+ * @param info - A Component's *.info.json
```
levithomason, line 6: Out of curiosity, is there any reason some of the JSDoc types were removed and others updated?

ecraig12345: All of them were removed I thought. And if they weren't already in the actual TS signatures, I added them there.

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -70,22 +70,22 @@ export interface DialogProps
 
   /**
    * Called after a user clicks the cancel button.
-   * @param {SyntheticEvent} event - React's original SyntheticEvent.
-   * @param {object} data - All props.
+   * @param event - React's original SyntheticEvent.
+   * @param data - All props.
```
levithomason, line 74: The changes here in the JSDoc will affect the TODO in `docs/src/components/ComponentDoc/ComponentProp/ComponentPropFunctionSignature.tsx`:

```jsx
        {/* TODO change these according to the react-docgen-typescript generated json */}
        {/* <ComponentPropFunctionSignature name={name} tags={tags} /> */}
```

This doc component shows the function signature for a callback in the prop table, like so:

image
image

It is currently disabled because the types (`SyntheticEvent`, `object`) are not currently getting parsed.  They are null.

***

I think this means moving forward we'll need to rely on actual typing resolution to figure out what the signature of `ComponentEventHandler<>` is.  Correct?

levithomason (approved):
👍

PR #2122, opened 2019-11-20
adding just-scripts

kenotron:
In the final convergent repo, we'll be using `just-scripts` to help with:

1. task orchestrator (this is an overlap of gulp, we will pick `just-task` eventually)
2. scripting (there's a growing set of shareable task functions that we can use that is shared internally with MSFT teams and externally via github)
3. just-stack provides templating via its integration with plop.js

These prepares for an ecosystem of many smaller component repos to be created that can take advantage of a set of scripts created by our team.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2122)


scripts/CHANGELOG.json
```diff
@@ -0,0 +1,139 @@
+{
```
jdhuntington: Do we want this since it's a new package?

scripts/CHANGELOG.md
```diff
@@ -0,0 +1,54 @@
+# Change Log - @fluentui/scripts
```
jdhuntington: Same, do we want this for a net-new package?

jdhuntington (approved):

scripts/CHANGELOG.md
kenotron: i can get rid of this for now :)

PR #2123, opened 2019-11-20
updating horizontal spacing between icons + buttons

notandrew:
###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2123)


notandrew:
This is a larger issue than at first glance. (The spacing between icons + text is different depending on component but we don't have that logic in place yet, so modifying this will modify all spacing between icon + text)

PR #2125, opened 2019-11-21
Add Azure pipeline definition

ecraig12345:
~~Add a build pipeline definition for moving the build from CircleCI to Azure Devops. (Migration plan is probably to run both for a few days, then delete CircleCI if all goes well.)~~ EDIT: We discovered the ADO machines are slower than the CircleCI ones and take about 1.5-2x as long per task, so we're sticking with CircleCI for now. The Azure build will only run for component governance checks on master.

Also switch Screener and CircleCI to use Node 10.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2125)

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,7 +6,7 @@ pool:
   vmImage: 'ubuntu-latest'
 
 variables:
-  - group: fluent-secrets
+  - group: fluent-variables-public
 
 steps:
   - task: NodeTool@0
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,6 +7,8 @@ pool:
 
 variables:
   - group: fluent-variables-public
+  - name: DANGER_GITHUB_API_TOKEN
+    value: '$(DANGER_GITHUB_API_TOKEN)'
 
 steps:
   - task: NodeTool@0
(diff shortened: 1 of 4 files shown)


ecraig12345:
Main build tasks were successful, but I'm marking this as do not merge until all the API keys and such have been added to the variable group (and I have to check the spelling of the branch name variables).

kenotron (approved):

azure-pipelines.yml
```diff
@@ -0,0 +1,78 @@
...
+      fi
+    displayName: Bundle statistics (master only)
+
+  - script: |
+      if [ $BUILD_SOURCEBRANCHNAME == 'master' ]; then yarn perf; fi
+    displayName: Performance tests (master only)
+
+  - script: |
+      if [ $BUILD_SOURCEBRANCHNAME == 'master' ] && [ -n "${STATS_URI}" ]; then
+        yarn stats:save --tag=`git tag --points-at HEAD`
```
miroslavstastny: `stats:save` reads several env variables, not sure all are available under the same name in azure.
See https://github.com/microsoft/fluent-ui-react/blob/master/build/gulp/tasks/stats.ts#L158-L167

kenotron: Great call out... we'll have to map those to these: https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#build-variables

ecraig12345: Thanks for catching that! I'll update it to check both sets of variables for now.

ecraig12345: Should be fixed now

kenotron (changes requested):
offline convo: will only use ADO for publishes and compliance.

kenotron (approved):
GO!

PR #2126, opened 2019-11-22
feat(Chat): Allow to disable actions menu positioning

jurokapsiar:
Adding `positionActionMenu` prop that disables popper and some size calculations. Can be set to false in not-focused and not-hovered state, to prevent unnecessary style calculations

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2126)


miroslavstastny (approved):
changelog

PR #2127, opened 2019-11-25
fix(whatInput): cleanup previously added event listeners

### LINKED ISSUE #2109: Memory: Window object leaks in what-input.ts
## Bug Report
We uncovered the following issues during memory investigations:
- The code in this file is referencing the global window object instead of working off of the supplied target window
- We are adding event listeners to the child window and not removing them. Refer to the addEventListener function in what-input.ts

### Version
0.40.3

silviuaavram:
Fixes https://github.com/microsoft/fluent-ui-react/issues/2109

Counting on `target[whatInputInitialized]` the number of Providers requesting to setup the custom-target whatInput.

To cleanup, in Provider `componentWillUnmount` we will attempt to remove listeners. If there are still Providers running (`target[whatInputInitialized] > 1`) then we will decrement that value. If it's the last one, then we perform cleanup.

Extra:
- in `whatInput.ts` I changed `useCapture` with `capture` (line 92). According to the [docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener) that is the correct property if we pass the information inside the `options` object params.

To Do:
- add test to check cleanup is performed.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2127)

### DIFF
--- a/packages/react/src/lib/whatInput.ts
+++ b/packages/react/src/lib/whatInput.ts
@@ -214,8 +214,9 @@ export const setUpWhatInput = (target: Document) => {
     'addEventListener' in targetWindow &&
     Array.prototype.indexOf
   ) {
-    if (target[whatInputInitialized]) {
-      target[whatInputInitialized] = target[whatInputInitialized] + 1
+    const initializedTimes = target[whatInputInitialized]
+    if (typeof initializedTimes === 'number' && initializedTimes > 0) {
+      target[whatInputInitialized] = initializedTimes + 1
       return
     }
     target[whatInputInitialized] = 1
--- a/packages/react/src/lib/whatInput.ts
+++ b/packages/react/src/lib/whatInput.ts
@@ -96,7 +96,7 @@ const addListeners = (eventTarget: Window) => {
   if (eventTarget.PointerEvent) {
     eventTarget.addEventListener('pointerdown', setInput)
     // @ts-ignore
-  } else if (window.MSPointerEvent) {
+  } else if (eventTarget.MSPointerEvent) {
     eventTarget.addEventListener('MSPointerDown', setInput)
   } else {
     // mouse events
@@ -256,7 +256,6 @@ export const tryCleanupWhatInput = (target: Document) => {
   if (isBrowser() && targetWindow && 'removeEventListener' in targetWindow) {
     if (target[whatInputInitialized] === 1) {
       delete target[whatInputInitialized]
-      console.log('clean--')
       cleanupWhatInput(targetWindow)
     } else {
       target[whatInputInitialized] = target[whatInputInitialized] - 1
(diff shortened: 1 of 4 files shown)


packages/react/src/lib/whatInput.ts
```diff
@@ -212,17 +214,55 @@ export const setUpWhatInput = (target: Document) => {
...
 }
 
+function cleanupWhatInput(eventTarget: Window) {
+  const options = supportsPassive ? { capture: true } : true
+
+  // @ts-ignore
+  if (eventTarget.PointerEvent) {
+    eventTarget.removeEventListener('pointerdown', setInput)
+    // @ts-ignore
+  } else if (window.MSPointerEvent) {
```
silviuaavram, line 236: still thinking why here there is `window` and not `eventTarget`.

silviuaavram: I am replacing this. @jurokapsiar @layershifter any comments?

packages/react/test/specs/components/Provider/Provider-test.tsx
```diff
@@ -232,4 +232,58 @@ describe('Provider', () => {
...
+
+      mount(
+        <Provider id="first-provider" target={externalDocument}>
+          <Provider id="second-provider" target={externalDocument}>
+            <div />
+          </Provider>
+        </Provider>,
+      )
+
+      expect(addEventListener).toHaveBeenCalledTimes(5)
```
miroslavstastny, line 260: black magic

packages/react/src/lib/whatInput.ts
```diff
@@ -87,14 +89,14 @@ const addListeners = (eventTarget: Window) => {
   // `pointermove`, `MSPointerMove`, `mousemove` and mouse wheel event binding
   // can only demonstrate potential, but not actual, interaction
   // and are treated separately
-  const options = supportsPassive ? { passive: true, useCapture: true } : true
+  const options = supportsPassive ? { passive: true, capture: true } : true
```
miroslavstastny, line 92: 👍

miroslavstastny (approved):
Approved, it would be great if @jurokapsiar could check it as well as he originally added multi-window support

packages/react/test/specs/components/Provider/Provider-test.tsx
jurokapsiar: https://www.numerology.com/numerology-numbers/5

silviuaavram: ok ok will add a comment why :))))

packages/react/src/lib/whatInput.ts
jurokapsiar: this is basically a platform check - saying "does this browser have MSPointerEvent?"

silviuaavram: will revert it then, but shouldn't the check above mean similarly? `eventTarget.PointerEvent`

PR #2131, opened 2019-11-26
fix(Dropdown): prevent unnecessary re-renders in DropdowItem

silviuaavram:
Breaking Changes🔥

We will not call `getItemProps` in `Dropdown`, we are passing the callback and its required params to `DropdownItem`. Unnecessary re-renders were triggered because that function returns an object, which is new every time.

Also doing a deep comparison between props in `DropdownItem` to make it act as a pure component, as we don't want it re-rendering when `Downshift` (parent component re-renders). The pure comparison is killing a bit the performance, but for now it is way faster than it was before while maintaining the same behavior. Maybe we can improve here as well, as without that deep compare the performance is fast.

Had to change prop types in `DropdownItem` to match the current implementation.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2131)

### DIFF
--- a/packages/react/src/components/Dropdown/DropdownItem.tsx
+++ b/packages/react/src/components/Dropdown/DropdownItem.tsx
@@ -88,6 +88,10 @@ class DropdownItem extends UIComponent<WithAsProp<DropdownItemProps>> {
     selected: PropTypes.bool,
   }
 
+  shouldComponentUpdate(nextProps: DropdownItemProps) {
+    return !_.isEqual(this.props, nextProps)
+  }
+
   handleClick = e => {
     _.invoke(this.props, 'onClick', e, this.props)
   }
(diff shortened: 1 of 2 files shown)


packages/react/src/components/Dropdown/DropdownItem.tsx
```diff
@@ -76,18 +83,24 @@ class DropdownItem extends UIComponent<WithAsProp<DropdownItemProps>> {
...
     image: customPropTypes.itemShorthandWithoutJSX,
+    index: PropTypes.number,
+    item: customPropTypes.itemShorthand,
     onClick: PropTypes.func,
     isFromKeyboard: PropTypes.bool,
     selected: PropTypes.bool,
   }
 
+  shouldComponentUpdate(nextProps: DropdownItemProps) {
+    return !_.isEqual(this.props, nextProps)
```
silviuaavram, line 101: to prevent it being re-rendered when a parent re-renders (in this case `Downshift` as it acts as a render prop component).

silviuaavram:
Will close this in favour of Downshift PR https://github.com/downshift-js/downshift/pull/841 . Once we get that into master and update the library we can just keep the 'pure component check' in `DropdownItem` and it should have the same effect.

PR #2141, opened 2019-12-03
feat(Divider): add vertical prop

silviuaavram:
Divider should now support the `vertical` prop which will position the Divider vertically.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2141)

### DIFF
--- a/docs/src/examples/components/Divider/Variations/DividerExampleVertical.shorthand.tsx
+++ b/docs/src/examples/components/Divider/Variations/DividerExampleVertical.shorthand.tsx
@@ -16,8 +16,8 @@ const DividerVerticalExampleShorthand = () => (
     />
     <Divider vertical />
     <Button icon="chevron-down" text iconOnly title="Close" />
-    <Button icon="stardust-circle" text iconOnly title="Full Screen" />
-    <Button icon="stardust-close" text iconOnly title="Minimize" />
+    <Button icon="icon-circle" text iconOnly title="Full Screen" />
+    <Button icon="icon-close" text iconOnly title="Minimize" />
     <Divider vertical />
   </div>
 )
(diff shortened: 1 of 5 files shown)


docs/src/examples/components/Divider/Variations/DividerExampleVertical.shorthand.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react'
+import { Divider, Button, Avatar, Text } from '@fluentui/react'
+
+const DividerVerticalExampleShorthand = () => (
+  <div style={{ display: 'flex', justifyContent: 'center', height: '32px', alignItems: 'center' }}>
+    <Divider vertical />
```
mnajdova, line 6: Should we support vertical Divider with content? Just curious how would it look like

mnajdova: If this is the case I believe it would complicate the styles in some way, as we will need to "leave space" between the two parts of the divider for the text..

silviuaavram: It wasn't a requirement so far. If it's with content then it changes significantly, as it will be `| content |` instead of just `|`. While the horizontal one is still a one liner. Do you think we need to handle this case now or can it be left for when we really need it?

mnajdova: If Teams needs doesn't need this, then it's okay.

mnajdova (approved):

PR #2142, opened 2019-12-04
Update to TS 3.7

ecraig12345:
Update TS version to `~3.7.2` (`~` to prevent accidental TS minor version upgrades, which can be disruptive). Due to stricter type checking, especially on index signatures, there are some new casts to `any`.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2142)


kenotron (approved):
yes!

packages/react-bindings/src/hooks/useStateManager.ts
```diff
@@ -12,7 +12,7 @@ const getDefinedProps = <Props extends Record<string, any>>(props: Props): Parti
 
   Object.keys(props).forEach(propName => {
     if (props[propName] !== undefined) {
-      definedProps[propName] = props[propName]
+      ;(<Record<string, any>>definedProps)[propName] = props[propName]
```
JasonGore, line 15: Was `;` applied when you saved in your editor? Seems inconsistently enforced to only changed lines.

miroslavstastny: That's because of the added `(cast)`.
https://prettier.io/docs/en/rationale.html#semicolons

PR #2145, opened 2019-12-04
fix: refactor some focus logic in MenuButton and SplitButton

silviuaavram:
MenuButton: single point of changing the `open` state prop, and calling `onOpenChange` from that place.

SplitButton: refactored the behavior. (Shift)Tab handlers are not needed, handled by menuButtonBehavior. Replaced the `open` call from Popup with `openAndFocusFrist` from MenuButton, in the behavior, on Alt+Arrow Down.

Popup: on update/mount will check if prop `open` is passed (controlled mode) and then update the trigger reference. Fixes on MenuButton the trigger focus on close when it was opened only by Arrow Down.


kolaps33 (approved):
tested with JAWS virtual cursor and NVDA browse mode and can confirm original issue is fixed :)

PR #2147, opened 2019-12-04
feat(Table): Adding table navigation 

kolaps33:
Adding table navigation like:
- row by row navigation 
- navigate inside the row use right arrow key
- in the row, then right and left arrow keys navigate between the cells 

What should works:
- when cell has more actionable elements, then:
    - "Enter" key goes inside the cell 
    - "Escape" key goes outside the cell
    - navigation between actionable elements inside cell is up/down and left/right arrows
- when cell has one actionable element, then focus goes directly to the actionable element
- space/enter invoke onclick on row
- space/enter invoke onclick on actionable element(more option button)
- click invoke onclick on whole row
- click invoke onclick on actionable element(more option button)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2147)


docs/src/examples/components/Table/Usage/TableExampleNavigable.shorthand.tsx
```diff
@@ -0,0 +1,125 @@
...
+    items: [
+      { content: '1', key: '1-1' },
+      { content: 'Roman van von der Longername', key: '1-2' },
+      { content: 'None', key: '1-3' },
+      { content: '30 years', key: '1-4' },
+      moreActionableElements,
+      moreOptionButton,
+    ],
+    onClick: () => handleRowClick(1),
+    'aria-label': 'custom text',
```
jurokapsiar: what does custom text mean? I think it would be better to use meaningful labels in the examples

packages/react/src/components/Table/TableRow.tsx
```diff
@@ -75,7 +77,25 @@ class TableRow extends UIComponent<WithAsProp<TableRowProps>, any> {
     accessibility: tableRowBehavior as Accessibility,
   }
 
-  renderCells(variables: ComponentVariablesObject) {
+  rowRef = React.createRef<HTMLElement>()
+
+  actionHandlers = {
+    unsetRowTabbable: e => {
+      this.rowRef.current.setAttribute('tabindex', '-1')
```
jurokapsiar, line 85: we should find a better way to do this, let's talk

jurokapsiar:
can you please add `handlesAccessibility` tests to the unit tests of the table, row and cell?

packages/accessibility/src/behaviors/Table/gridHeaderCellBehavior.ts
```diff
@@ -0,0 +1,22 @@
+import { Accessibility } from '../../types'
+import { IS_FOCUSABLE_ATTRIBUTE } from '../../attributes'
+// import * as keyboardKey from 'keyboard-key'
```
silviuaavram: remove these comments

packages/accessibility/src/behaviors/Table/gridHeaderRowBehavior.ts
```diff
@@ -0,0 +1,47 @@
+import { Accessibility } from '../../types'
+import { IS_FOCUSABLE_ATTRIBUTE } from '../../attributes'
+import { FocusZoneMode, FocusZoneDirection } from '../../focusZone/types'
+import * as keyboardKey from 'keyboard-key'
+import gridHeaderCellBehavior from './gridHeaderCellBehavior'
+
+// add key actions unit tests
```
silviuaavram: and these

packages/accessibility/src/behaviors/Table/gridHeaderRowBehavior.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * @description
```
silviuaavram: if you don't add anything to `description` just remove them, otherwise add description about the behavior.

packages/accessibility/src/behaviors/Table/gridNestedBehavior.ts
```diff
@@ -0,0 +1,44 @@
...
+    props: {
+      shouldEnterInnerZone: event => keyboardKey.getCode(event) === keyboardKey.ArrowRight,
+      direction: FocusZoneDirection.vertical,
+      shouldResetActiveElementWhenTabFromZone: true,
+    },
+  },
+  keyActions: {
+    root: {
+      focus: {
+        keyCombinations: [{ keyCode: keyboardKey.Escape }],
```
silviuaavram, line 34: `@specification` test?

packages/accessibility/test/behaviors/testDefinitions.ts
```diff
@@ -609,6 +609,21 @@ definitions.push({
   },
 })
 
+// Triggers 'unsetRowTabbable' action using SHIFT + TAB key on 'root'.
```
silviuaavram: Maybe we should abstract this, so we can include stuff like `Triggers 'openAndFocusFirst' action with 'altKey'+'ArrowDown' on 'root' slot.`

Maybe we can also add the part with `based on the prop $propname` so we ca include stuff like:
```js
          ...(!props.open && {
            openAndFocusFirst: {
              keyCombinations: [{ keyCode: keyboardKey.ArrowDown, altKey: true }],
            },
          }),
```

packages/react-bindings/src/FocusZone/FocusZone.tsx
```diff
@@ -513,16 +513,16 @@ export default class FocusZone extends React.Component<FocusZoneProps> implement
 
     const doc = getDocument(this._root.current)
 
+    if (this.props.onKeyDown) {
```
silviuaavram: you can probably use lodash invoke here: `_.invoke(this.props, 'onKeyDown', ev)`

packages/accessibility/src/behaviors/Table/gridNestedBehavior.ts
kolaps33: good catch :) added spec line above

packages/react/src/components/Table/TableRow.tsx
kolaps33: bug reported please see comment above

docs/src/examples/components/Table/BestPractices/TableBestPractices.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { Text } from '@fluentui/react'
+import ComponentBestPractices from 'docs/src/components/ComponentBestPractices'
+import { link } from '../../../../utils/helpers'
+
+const doList = [
+  <Text>
+    Choose desired accessibility behavior for grid cell depending on the use case.(Check the{' '}
+    {link('Behaviors', '/components/table/accessibility')} section).
+  </Text>,
+  'Provide label to the Table component using `aria-label` or `aria-labelledby` prop.',
```
jurokapsiar: pls add: `If no label is provided, the whole header will be read by the screen reader when the user navigates to the table`

docs/src/examples/components/Table/BestPractices/TableBestPractices.tsx
```diff
@@ -0,0 +1,26 @@
...
+  'Provide label to the Row component using `aria-label` or `aria-labelledby` prop. If not, then each cell of the row is narrated by screen reader.',
```
jurokapsiar: pls change to `Consider providing label to the TableRow component using aria-label or aria-labelledby prop. When not provided, the whole row (all cells) will ne narrated by the screen reader when the user navigates to the row.`

docs/src/examples/components/Table/BestPractices/TableBestPractices.tsx
```diff
@@ -0,0 +1,26 @@
...
+  'Provide label to the table header column, if cell has no content.',
+  'Stop event propagation, when you will add actionable element into the grid cell.',
```
jurokapsiar: If you use actionable element or component in the cell, stop propagation of the click event if you do not want the interacion with the element to perform row click action.

packages/react-bindings/src/FocusZone/FocusZone.tsx
kolaps33: as this is FocusZone code I rather don't want to change the code... 
Even this change we are asking if focusZone people agree with the change.

docs/src/examples/components/Table/BestPractices/TableBestPractices.tsx
```diff
@@ -0,0 +1,26 @@
...
+]
+
+const dontList = [
+  "Don't set onClick action on row, which will be not available in the cell of grid, as well. With screen reader navigation user can navigate only cell by cell.",
```
jurokapsiar: If you use onClick on the row, make also one (preferably the first) cell actionable and add onClick handler with the same action to it. Screen readers do not navigate to rows, so actions need to be available on cells as well.

docs/src/examples/components/Table/Usage/TableExampleNavigable.shorthand.tsx
```diff
@@ -0,0 +1,125 @@
...
+import { Table, Button, Provider, Flex } from '@fluentui/react'
+import {
+  gridNestedBehavior,
+  gridCellWithFocusableElementBehavior,
+  gridCellMultipleFocusableBehavior,
+} from '@fluentui/accessibility'
+
+function tagButton(tagName: string) {
+  return (
+    <Provider
```
jurokapsiar: make sure to have only one provider in the example, not for each button

jurokapsiar (approved):

PR #2152, opened 2019-12-05
Moving staging react package into playground package

kenotron:
This is a 1:1 move from staging/packages/react to playground - @fluentui/playground

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2152)


kenotron:
Will iterate here and add test back in for the non-stardust packages

PR #1999, opened 2019-10-02
[DO NOT MERGE] Docs Version 2

dvdzkwsk:
DO NOT MERGE! This PR exists so that I can track integration issues between the new doc site and the current Stardust build pipeline. The size of this PR will drop considerably once it is ready, since right now a lot of documentation is duplicated between /docs and /docs-v2. Also, a large portion of the additions are exclusively within yarn.lock, since Gatsby has been added as a dependency.

## New Features

* Global site search
* Component previews
* Server-side rendering
* Mobile support
* Convenient global light/dark theme toggle
* New and updated "getting started" pages
* Playground can toggle theme independently of global theme
* Full screen component playground
* Sitemap
* Eliminates all pre-build steps for docs
* Improved SEO (all pages have proper titles and descriptions)
* All site content is .mdx, not Javascript.

## Tradeoffs
* While the new site is still built with React components, it does not rely on Stardust components outside of the playground and some other assorted widgets. This has mostly been a limitation of Stardust's support for server-side rendering. As that aspect of the library improves, we can replace the existing components with their Stardust variants.

## Todo
* [ ] Developer documentation
* [ ] Port all existing content
* [ ] Verify screener tests still work
* [ ] Verify perf tests still work
* [ ] Accessibility signoff

## Screenshots

Screen Shot 2019-10-02 at 11 22 24 AM

Screen Shot 2019-10-02 at 11 24 12 AM

Screen Shot 2019-10-02 at 11 28 36 AM

PR #2153, opened 2019-12-06
Rename lib to utils

ecraig12345:
Rename all `lib` folders to `utils` so we can gitignore `lib` and use it for build output. (I could use `utilities` or some other name instead if people prefer.) 

Another option is to move the things formerly under `packages/react/src/lib` to their own package, but that would require significantly more work/breaks.

I did NOT actually add `lib` to `.gitignore` yet because I'm concerned that would cause any changes to `lib` in open PRs to just disappear when they merge with master (instead of being merged into the moved files).

The following open PRs touch files in `lib` folders: #2115, #2044, #2006, probably #1999, #1991, #1980, #1867, #1301. Other PRs may also add/modify references to `lib` imports, but this will be caught by a merge conflict and/or build error.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2153)


.github/CONTRIBUTING.md
```diff
@@ -1,10 +1,8 @@
-CONTRIBUTING
-============
+# CONTRIBUTING
```
ecraig12345, line 1: The formatting changes here are because I have my editor set up to run prettier on save (could revert if desired). Same in a couple other markdown files.

kenotron: I think we should revert this - danger actually checks for diffs. So this will blow up danger.

build/dangerjs/checkChangelog.ts
```diff
@@ -50,7 +50,9 @@ export default async ({ danger, fail, warn }: DangerJS) => {
   } else {
     const malformedChangelogEntries = await getMalformedChangelogEntries(danger)
     malformedChangelogEntries.forEach(entry => {
-      fail(`Invalid entry format in ${CHANGELOG_FILE}: >${entry}<`)
+      fail(`Invalid entry format in ${CHANGELOG_FILE}: >${entry}<
+
+The correct format is: \`- description @githubname ([#DDDD](https://github.com/microsoft/fluent-ui-react/pull/DDDD)\``)
```
kenotron, line 55: 👍

kenotron (approved):
I think the only thing I want to mention is to preserve the CHANGELOG.md formatting for now - maybe we should have added CHANGELOG.md in .prettierignore

.github/CONTRIBUTING.md
ecraig12345: Not quite sure what you mean about danger?

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ## [Unreleased]
 
+### BREAKING CHANGES
+- Rename `lib` directories to `utils` @ecraig12345 ([#2153](https://github.com/microsoft/fluent-ui-react/pull/2153))
```
layershifter: Not sure that this is a really breaking change as I don't see anything that can affect consumer that uses public API.

ecraig12345: That's good to know, thank you! Some of my experience with Fabric has led me to be very cautious about assuming internal path changes are non-breaking, but if you think it's safe here that's great.

PR #2156, opened 2019-12-07
feat(Icons): Add new icons to Teams theme

codepretty:
image
- `volume-down`
- `volume`
- `volume-up`

image
- `presenter`
- `no-presenter`

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2156)


miroslavstastny (commented):
Technically ok.
Visually the `demote` is ~1px up compared to `promote` which can cause an icon to jump in scenarios where you switch from one to the other on click (toolbar). Is that ok?

codepretty:
> Technically ok.
> Visually the `demote` is ~1px up compared to `promote` which can cause an icon to jump in scenarios where you switch from one to the other on click (toolbar). Is that ok?

Good catch - will see about getting that fixed

TanelVari:
Updated shapes and renamed presenter SVG-s.

image

miroslavstastny (approved):

PR #2158, opened 2019-12-09
Docs: Add Perf Test documentation

JasonGore:
Add documentation for adding and running perf tests.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2158)

### DIFF
--- a/.github/test-a-feature.md
+++ b/.github/test-a-feature.md
@@ -250,12 +250,13 @@ Performance tests will measure performance, set a baseline for performance and h
 - Formatting follows [Storybook CSF convention](https://storybook.js.org/docs/formats/component-story-format/) with special support for `iterations` metadata which tells the performance testing package how many iterations of your component to render:
 
 ```tsx
+// If this file is named ButtonBasic.perf.tsx, it will be picked up as kind of 'ButtonBasic' with story names of 'Blank' and 'WithText'.
 export default {
   iterations: 5000,
 }
 
-export const Fabric = () => <ButtonFabric />
-export const Fluent = () => <ButtonFluent />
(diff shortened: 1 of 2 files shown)


.github/test-a-feature.md
```diff
@@ -236,3 +239,42 @@ Add description under the `@description` tag. Like:
...
 #### I want to create unit tests in separate file not through the regex
 Add your spec file into the array of files `skipSpecChecksForFiles` in `testHelper.tsx`. And put description in behavior file under `@description` tag.
+
+## Performance Tests
+
+Performance tests will measure performance, set a baseline for performance and help guard against regressions. 
+
+### Adding a Perf Test
+
+- To add a perf test, simply add a file with `.perf.` in its naming. (As of writing, `.perf.` files in `docs/src` and `packages/perf-test` are automatically consumed.)
```
miroslavstastny, line 249: Do we want to describe the difference between the two? (docs/src perf example are visible in docsite)
But that might change soon.

miroslavstastny (approved):
👍

.github/test-a-feature.md
JasonGore: Yeah this will have to change. We'll have to update this as we go along, but for now, I wanted to document it primarily for internal use.

JasonGore: We should talk about strategy here but I think ideally these tests would live alongside component implementations similar to unit tests. I view `packages/perf-tests` as a temporary location for comparing components.

PR #2160, opened 2019-12-10
fix(renderer): properly expand colors in "border-color" shorthand

### LINKED ISSUE #1995: Plugins: expand is not working for `rgba()` values with space

layershifter:
Fixes #1995.

We use compiled version of `css-expand-shorthand` (#1542) and there is already `normalize()` function that processes colors before parse. This PR simply enabled it for `border-color` property.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2160)


miroslavstastny (approved):
👍

PR #2166, opened 2019-12-11
chore: capture heap snapshots with Fluent IDs

layershifter:
## Prototype

The general idea behind is try to make objects in memory discoverable so we will be able to automate their capture and measurements.

I don't see any other options than take heap snapshots:
- [`Performance.memory`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/memory) is vendor API and there is no useful information 😭 
- Puppeteer provides [queryObjects()](https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#executioncontextqueryobjectsprototypehandle) API, but there is nothing that allows to count memory 😭 _x 2_

#### Mark places that create heavy objects

Used [`Object.setPrototypeOf()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/setPrototypeOf) so they can be easy discovered in snapshots.

image

#### Automate recording of heap snaphots

A newly created workflow (see `perf.yml`) runs performance tests, calls GC and captures a snapshot. Also it stores created snapshot as an artifact.

image

#### Possible next steps ⏭ 

`.heapsnapshot` is a JSON file and can be parsed, so we can use it in automation. I checked [`chrome-heap-snapshot-parser`](https://github.com/julianburr/chrome-heap-snapshot-parser), it needs some polishes but can be used as a source of ideas.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2166)

### DIFF
--- a/build/gulp/tasks/perf.ts
+++ b/build/gulp/tasks/perf.ts
@@ -118,7 +118,13 @@ task('perf:build', cb => {
 })
 
 task('perf:run-memory', async () => {
-  const browser = await puppeteer.launch(safeLaunchOptions())
+  const browser = await puppeteer.launch(
+    safeLaunchOptions(
+      process.env.CI && {
+        executablePath: 'google-chrome-unstable',
+      },
+    ),
+  )
   const context = await browser.createIncognitoBrowserContext()
   const page = await context.newPage()
 
--- a/build/gulp/tasks/perf.ts
+++ b/build/gulp/tasks/perf.ts
@@ -117,6 +117,37 @@ task('perf:build', cb => {
   webpackPlugin(require('../../../build/webpack.config.perf').default, cb)
 })
 
+task('perf:run-memory', async () => {
+  const browser = await puppeteer.launch(safeLaunchOptions())
+  const context = await browser.createIncognitoBrowserContext()
+  const page = await context.newPage()
+
+  const client = await page.target().createCDPSession()
+  await client.send('HeapProfiler.enable')
+
+  const chunks: string[] = []
+
+  client.on('HeapProfiler.addHeapSnapshotChunk', ({ chunk }) => {
+    chunks.push(chunk)
+  })
+
+  await page.goto(`http://${config.server_host}:${config.perf_port}`)
+
+  await page.evaluate(() => window.runMeasures())
+  await page.evaluate(() => gc()) // collect garbage manually
+
+  await client.send('HeapProfiler.takeHeapSnapshot', { reportProgress: false })
+
+  await context.close()
+  await browser.close()
+
+  const heapSnapshot = chunks.join('')
+  const resultsFile = paths.perfDist('result.heapsnapshot')
+
+  fs.writeFileSync(resultsFile, heapSnapshot)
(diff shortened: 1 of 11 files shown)


yarn.lock
```diff
@@ -6899,6 +6899,11 @@ chownr@^1.0.1, chownr@^1.1.1:
   resolved "https://registry.yarnpkg.com/chownr/-/chownr-1.1.1.tgz#54726b8b8fff4df053c42187e801fb4412df1494"
   integrity sha512-j38EvO5+LHX84jlo6h4UzmOwi0UgW61WRyPtJz4qaadK5eY3BTS5TY/S1Stc3Uk2lIM6TPevAlULiEJwie860g==
 
+chrome-heap-snapshot-parser@^1.0.0-rc.2:
+  version "1.0.0-rc.2"
+  resolved "https://registry.yarnpkg.com/chrome-heap-snapshot-parser/-/chrome-heap-snapshot-parser-1.0.0-rc.2.tgz#0d61936875055f64eaa18bb1fcb1fc75274396d2"
+  integrity sha512-sir6w6qsqCgJk6Y8ZmwhZ/9eoRntwbFxH7FYi5AcrM60hh2fDKSymqOuZ+J7OQI68v/Z9XYmG1dKjPBxx3i2VQ==
+
```
layershifter: Cruft

packages/react/src/utils/withMemoryId.ts
```diff
@@ -0,0 +1,11 @@
+/* tslint:disable */
+
+function withMemoryId<T>(target: T): T {
+  let Prototype: Function // eslint-disable-line
+  eval(`Prototype = function FluentMemoryRecord () {}`) // eslint-disable-line
```
layershifter, line 5: Try to switch from `eval()` to `new Function()`

build/webpack.config.perf.ts
```diff
@@ -68,6 +68,8 @@ const webpackConfig: any = {
   },
   optimization: {
     nodeEnv: !!argv.debug ? 'development' : 'production',
+    // We don't want to run Terser on perf tests
+    minimize: false,
```
layershifter, line 72: Also disable TS checks for memory build

build/gulp/tasks/perf.ts
```diff
@@ -117,6 +117,43 @@ task('perf:build', cb => {
...
+  const chunks: string[] = []
+
+  client.on('HeapProfiler.addHeapSnapshotChunk', ({ chunk }) => {
+    chunks.push(chunk)
+  })
+
+  await page.goto(`http://${config.server_host}:${config.perf_port}`)
+
+  await page.evaluate(() => window.runMeasures())
+  await page.evaluate(() => gc()) // collect garbage manually
```
layershifter, line 143: Triggering garbage collection manually, need to be checked again

https://media-codings.com/articles/automatically-detect-memory-leaks-with-puppeteer

layershifter:
This PR was initially done as a prototype. Closing for now, maybe we will use parts of it in future.

PR #2170, opened 2019-12-12
chore: Run tests per-package

ecraig12345:
Add an individual jest configuration for each package instead of one configuration at the root. This is one step towards making the packages' builds more modular, which is a prereq for converging to one repo.

Also add per-package `test`, `test:watch`, and `clean` commands, and make `gulp bundle:package:*` automatically run against the current directory if in a sub-package.

### New workflow

While in the root directory:

* `yarn test` runs each project's `test` script (basically `lerna run test --stream`)
* Same as before, `yarn test:watch` runs/watches all tests in non-just projects
* (new) `yarn clean` cleans cache, docs, and each project

While in a package directory:

* `yarn test` and `yarn test:watch` runs that project's tests
* `yarn clean` cleans that project's output
* `gulp bundle:package:*` (where `*` is the rest of a task name) runs the specified task against only the current package

### Details

Getting the tests working per-project also required adding per-project `gulpfile.ts`, `.gulp.js` (formerly `.gulp.json`), and `babel.config.js`. For now these mostly just reuse the base configs.

Exception to the above: I removed the `build:docs:component-info` and `build:docs:component-behavior-menu` steps from the default `test` command, and moved them to overrides of the `test` command in the gulpfiles of the relevant projects (`react` and `accessibility` respectively).

I also made assorted typing updates to build files--turned out to not be needed, but I'm checking them in anyway since I already did the work and they help a bit with maintainability/understandability. (Background is I thought they were needed before Ken figured out that I needed to copy the `.gulp.js` file into each project to make it register babel properly. Without that file, it was trying to use ts-node to interpret TS scripts, and apparently ts-node does type checking based on the project's config, which caused type errors in newer projects with stricter configs.)

#### Other stuff

- Update `bundle:package:types` to only `cd` if it's being run at the root
- Update root jest config to automatically exclude just-based projects
- Add `isRoot` and `package` to `config.ts` for convenient reference in tasks
- Combine clean tasks and use `del` instead of `rimraf` so we can easily delete multiple things in one task
- Move jest `maxWorkers` CI setting from `config.yml` to `build/gulp/tasks/test-unit.ts` to avoid issues with the syntax of passing it through on the command line
- Split `watch:docs` task into separate parts so that the `react` and `accessibility` packages can run only the parts they need
- Make `perf-test/just.config.ts` wait to load anything from `@fluentui/digest` until it's actually needed. This makes the `clean` and `test` scripts work even if digest hasn't been built yet. (issue will go away with project references!)
- Babel config:
  - Switch Node compatibility target in Babel config to 10 (matches CI configs)
  - Add sourcemap options to help facilitate debugging build scripts in an editor...because I needed to do that (these options are only used when `process.env.DEBUG` is set--could switch to a different detection method if preferred)
- Add `moduleNameMapper` config using `lerna-alias` in jest config for just-based projects

If anyone is interested, this is the VS Code launch.json configuration I used for debugging scripts...
```
{
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Debug test",
      "type": "node",
      "request": "launch",
      "program": "${workspaceFolder}/node_modules/gulp/bin/gulp.js",
      "cwd": "${workspaceFolder}/packages/react",
      "runtimeExecutable": null,
      "runtimeArgs": ["--nolazy", "--debug"],
      "args": ["test"],
      "env": {
        "NODE_ENV": "development",
        "DEBUG": "1"
      },
      "sourceMaps": true,
      "outputCapture": "std",
      "console": "integratedTerminal"
    }
  ]
}
```


.circleci/config.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           command: yarn prettier
       - run:
           name: Unit Tests
-          command: yarn test --maxWorkers=2
+          command: yarn test
```
kenotron, line 48: I think you're hitting an issue with running unit tests before these packages are built (so, the @fluentui/digest/lib/index.js has been generated yet). Either we'll have to swap things around here or teach jest about how to resolve the modules (other packages use lerna-alias for this purpose I think?)

jest.config.js
```diff
@@ -1,16 +1,19 @@
...
-
-const projects = Object.keys(projectPackages)
-  .filter(p => !excluded.includes(p))
-  .map(packageName => projectPackages[packageName])
+// Exclude packages which build with just, and other special packages
+const excludedPackages = ['@fluentui/internal-tooling']
+const projects = Object.keys(packagePaths).filter(
+  packageName =>
+    !excludedPackages.includes(packageName) &&
+    !fs.existsSync(path.join(packagePaths[packageName], 'just.config.ts')),
```
kenotron, line 15: nice catch!

packages/perf-test/just.config.ts
```diff
@@ -11,6 +9,8 @@ import runPerfTest from './tasks/perf-test'
 // - Existing perf story format diverges from CSF format, requiring special loader.
 function bundleStories() {
   return async function() {
+    // delay require in case digest isn't built yet
```
kenotron, line 12: This will likely get fixed up when we split up the build steps a bit more per offline discussion

kenotron (approved):
Thanks so much for this, @ecraig12345 - this is a good step forward

config.ts
```diff
@@ -1,5 +1,6 @@
 import * as path from 'path'
 import * as _ from 'lodash'
+import { argv } from 'yargs'
```
kenotron: this file gets used in a webpack bundling - and it is causing build issues. we probably can't use yargs in config.ts

package.json
```diff
@@ -84,6 +92,7 @@
     "copy-webpack-plugin": "^4.5.2",
     "cross-env": "^5.1.4",
     "danger": "^6.0.5",
+    "del": "^5.1.0",
```
miroslavstastny, line 94: There is only couple of places left where `rimraf` is used. Shouldn't we replace them with `del` and get rid of `rimraf` completely?

miroslavstastny (approved):
Since Stardust became a monorepo it was not possible to run UTs direcly in WebStorm.
After this change I can run them again 🎉

ecraig12345:
> Since Stardust became a monorepo it was not possible to run UTs direcly in WebStorm.
> After this change I can run them again 🎉

@miroslavstastny Cool! I'm curious which part of the change enabled that?

miroslavstastny:
> Cool! I'm curious which part of the change enabled that?

I think the pure fact that it is now possible to run `jest` directly from the package directory.

packages/internal-tooling/babel/index.js
```diff
@@ -18,7 +18,7 @@ module.exports = api => {
       '@babel/preset-env',
       {
         modules: useESModules ? false : 'cjs',
-        targets: isNode ? { node: '8' } : undefined,
+        targets: isNode ? { node: '10' } : undefined,
```
layershifter, line 21: :heart:

layershifter (approved):
:+1:

kenotron:
So much awesome!!!

PR #2172, opened 2019-12-12
fix(dropdown):  update dropdown styles in Teams theme to match redlines

notandrew:
Updating the Dropdown styles to match redlines


CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 
 ### Fixes
 - Fix event lister leak in `FocusZone` @miroslavstastny ([#2227](https://github.com/microsoft/fluent-ui-react/pull/2227))
+- Fix `dropdown` styles to match Teams theme @notandrew ([#2172](https://github.com/microsoft/fluent-ui-react/pull/2172))
```
silviuaavram: `Dropdown` as in component.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -53,6 +53,7 @@ export interface DropdownSlotClassNames {
   selectedItem: string
   selectedItems: string
   triggerButton: string
+  inverted: string
```
silviuaavram: Don't need to add it as a slot.

Slots are used on complex components. For instance, a checkbox can have a label and a ticking box, so we will add those as slots to the checkbox component. Useful to apply styles.

Here `inverted` is just a prop, unless I am missing something.

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1374,6 +1381,7 @@ Dropdown.slotClassNames = {
   selectedItem: `${Dropdown.className}__selecteditem`,
   selectedItems: `${Dropdown.className}__selected-items`,
   triggerButton: `${Dropdown.className}__trigger-button`,
+  inverted: '',
```
silviuaavram: remove this as well

silviuaavram (approved):
I also saw https://screener.io/v2/states/BJRCORrtB.microsoft-fluent-ui-react/2172/1024x768/Chrome/popupexampleclosebuttonshorthandtsx-prepares-to-select-item-out-of-popup-0 (the one with the list of items visible).

The highlighted item has no changes, while the others have their header changed. Maybe it's nothing, but can you please double check?

Also left some comments, overall looks good! @mnajdova can you also take a look?

docs/src/examples/components/Dropdown/Types/index.tsx
```diff
@@ -24,6 +24,11 @@ const Types = () => (
       description="A dropdown can be searchable and allow a multiple selection."
       examplePath="components/Dropdown/Types/DropdownExampleSearchMultiple"
     />
+    <ComponentExample
+      title="Inverted"
+      description="A dropdown can be on an lighter or darker background."
```
mnajdova: Please make all descriptions of the inverted examples consistent.

packages/react/src/components/Dropdown/Dropdown.tsx
mnajdova: @silviuavram is right, this is just a regular prop, that should appear in the `DropdownProps` and `propTypes`, not in the `DropdownSlotClassNames`

packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -122,6 +123,9 @@ export interface DropdownProps
   /** A dropdown can be formatted to appear inline next to other elements. */
   inline?: boolean
 
+  /** It can be on a dark background */
+  inverted?: boolean
```
mnajdova, line 127: Consistent description with other inverted props. Also, please see how the other props descrption are defined in the form of `A dropdown can ....` 

Try to follow that format

packages/react/src/themes/teams-dark/componentVariables.ts
```diff
@@ -15,5 +15,4 @@ export { default as Icon } from './components/Icon/iconVariables'
 export { default as Reaction } from './components/Reaction/reactionVariables'
 export { default as Alert } from './components/Alert/alertVariables'
 export { default as ProviderBox } from './components/Provider/providerBoxVariables'
```
mnajdova: Great that we manage to remove all dark theme overrides, just double check the screener differences.

mnajdova: There are lots of changes in dark theme that changes the background of the Dropdown. Make sure that we are not breaking something

notandrew: Looks good to me

PR #2177, opened 2019-12-12
chore(Perf): Skip undef values in perf charts

miroslavstastny:
Skip undefined values in perf charts. Previously those were charted as 0 which might be confusing.

Also now there is only one `onNearestX` handler registered as expected.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluent-ui-react/pull/2177)

### DIFF
--- a/docs/src/components/ComponentDoc/PerfChart/PerfChart.tsx
+++ b/docs/src/components/ComponentDoc/PerfChart/PerfChart.tsx
@@ -51,15 +51,18 @@ const PerfChart: React.FC<PerfChartProps> = ({ perfData, withExtremes }) => {
       <LineSeries
         {...props}
         key={chartName + key}
-        data={perfData.map(sample => ({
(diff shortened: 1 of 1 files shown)


docs/src/components/ComponentDoc/PerfChart/PerfChart.tsx
```diff
@@ -149,6 +152,18 @@ const PerfChart: React.FC<PerfChartProps> = ({ perfData, withExtremes }) => {
         strokeWidth: '2px',
       })}
 
+      <LineSeries
+        opacity={0}
```
levithomason, line 156: Ah, I see.  Suggest a comment here as to "why" we have a transparent line.

levithomason (approved):

PR #2179, opened 2019-12-12
chore: rename screener project

levithomason:
Well, I wrote a description but `@msftbot` decided to destroy it with `null`, for the second time today...

### DIFF
--- a/build/screener/screener.config.js
+++ b/build/screener/screener.config.js
@@ -10,7 +10,7 @@ require('tsconfig-paths').register({
 
 // https://github.com/screener-io/screener-runner
 module.exports = {
-  projectRepo: 'stardust-ui/react',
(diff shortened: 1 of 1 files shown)


build/screener/screener.config.js
```diff
@@ -10,7 +10,7 @@ require('tsconfig-paths').register({
 
 // https://github.com/screener-io/screener-runner
 module.exports = {
-  projectRepo: 'stardust-ui/react',
+  projectRepo: 'microsoft/fluent-ui-react',
```
levithomason, line 13: This PR only creates our new screener project.  We need this commit separate so we can ensure our new set of baseline images are correct.  No other changes are made here to guarantee our first `microsoft/fluent-ui-react` images are exact matches to the latest approved master `@stardust-ui/react` images.

PR #2184, opened 2019-12-14
perf: add more perf-test comparisons

levithomason:
This PR adds just a few more comparisons to the `perf-test` package.   The `perf-test` package is not relevant to end users so I'm not adding a CHANGELOG entry here.


JasonGore:
I think Dropdown and Slider could both be reduced to 1,000-2,000 iterations based on the results. (They are currently taking approximately 16 seconds and 9 seconds to run, respectively.)

Makes me wonder if we could tweak scenario validation to target iterations as a proportion of TPI... (or just total Ticks)

levithomason:
Perhaps we add a dangerjs PR check similar to what we discussed.  We could fail PRs with perf scenarios that take more than 10k ticks?  Thoughts?

levithomason:
Also, in order to test "first render", couldn't we just use 1 iteration?  I get that the flamegraph wouldn't be as helpful, but the ticks would be very useful.

Right now, what we're seeing in the tick numbers are the result of cache hits.  I would like to see also 1:1 comparison of cost to mount, render, and style these components which only happens on first render.

levithomason:
Another option would be to chart the iterations, instead of showing the grand total and the TPI.  What I'd expect to see here is very similar first iterations for both components followed by very low ticks for cache hit iterations (Fabric).  Something like:



If we stored the total ticks for each iteration, instead of TPI, then we could do this.  We could also easily calculate TPI and other values from this.

JasonGore (approved):
I still think there's a lot of value "real world" scenarios like the Chat perf examples in docs. If a perf story can have 1 iteration (where TPI and total ticks are the same) yet communicate useful information I think that's ideal.

levithomason:
I also agree.  Perhaps we forego the single component perf tests in favor of multiple components with real-world variations then?

In the real-world, applications aren't rendering 10 or 100 of the same component with the exact same props.  There might be 10 buttons on the page, but not likely that any of them are exactly identical.  Same with the Chat per example, yes it has 100 messages, but they are exactly identical.   That doesn't happen in the real-world either.  Perhaps you'd have 100 messages, but they would all be different.  Something more like https://microsoft.github.io/fluent-ui-react/prototype-chat-messages.

We could make it a requirement to have something like 10 components all with real-world varied usage.  Then, we can do as you say and run a single iteration with a little more confidence of first-render cost.  This would also help the multi-iteration scenarios be a little more believable.  Right now, they are hyper fitted to show cache hits because after the first component renders there is never another style required to be processed.

What say ye?

P.S.
I still think a graphic representation, over a table, would be ideal for quickly and easily understanding the data.  We could generate this and post it to the PR still.

JasonGore:
Flamegrill (at least as it currently exists) isn't able to show every individual component's iteration time. It's a black box tester that renders a given test in full. The TPI calculated is **average** and simply calculated as `total ticks / iterations`. At least from Flamegrill's perspective, generating a graph like this isn't currently possible.

PR #2187, opened 2019-12-16
fix(Popper): improve positioning in multiple cases

### LINKED ISSUE #2064: Popup clipping in RTL
## Bug Report
Popup clips in RTL, even if the frame is large enough to support it.
### Steps
Opening Popup in RTL from toolbar item can cause clipping on right side. While, not present in this example you can see strange/jumpy rendering behavior:

https://codesandbox.io/s/stardust-ui-example-rgm6n

### Expected Result
Popup opens normally, with no rendering shift or clipping in both RTL and LTR

### Actual Result
Popup clips and has strange rendering and can clip in RTL mode

### Testcase

### LINKED ISSUE #2072: Incorrect popup placement after jump-resize

layershifter:
Fixes #2064.
Fixes #2072.

## Case 1: jump on focus

The same as https://github.com/mui-org/material-ui/issues/16740, when opening a `Popup` with an autofocus input inside it, should not cause the page to jump (scroll).

## Case 2: scrollbar sizing issues

> The situation arises when an element on the page has as height the entire viewport (100vh). Imagine the following DOM:
>
>What happens is that right before the computation of the position of the popper, the popper gets `position: absolute`. However, since the popper appears after the page in the DOM, it will be positioned after it as well. Since the height of the page then becomes 100vh + height(popper), the scrollbar appears.

***

See code comments for descriptions and linked issues.


miroslavstastny (approved):

packages/react/src/utils/positioner/Popper.tsx
```diff
@@ -11,16 +11,28 @@ import getScrollParent from './getScrollParent'
...
+  popper: HTMLElement,
   options?: PopperJS.PopperOptions,
 ): PopperJS => {
   const instance = new ((_PopperJS as any).default || _PopperJS)(reference, popper, {
     ...options,
     eventsEnabled: false,
   })
+
+  const originalUpdate = instance.update
+  instance.update = () => {
```
miroslavstastny, line 23: Why the PR title says 'first render'? This fixes resizing after the first render.

layershifter: Updated 👍

oliviertassinari:
😆

PR #2190, opened 2019-12-17
fix(factories): enforce `generateKey` property

layershifter:
This PR enforces `generateKey` option to force the proper configuration of factory usages:
- `generateKey=true` inside collections, i.e. `ListItem.create({}, { generateKey: true })` in `List`
- `generateKey=false` in all other cases, i.e. `Icon.create({}, { generateKey: false })` in `MenuItem`

_As other possible option we can set it to `false` by default instead of `true`._

### Avoid collision in generated keys

See https://github.com/Semantic-Org/Semantic-UI-React/pull/2424 for more details.

### Avoid useless remounts

For example `content` is a `string` that will be passed to `Tooltip` component:

```tsx
<Tooltip content={content} />
```

The `content` slot there is represented by `TooltipContent` component. As our current default value is `true`, it forces to generate keys always. And this will cause the component's remount on every `string` change 👎


levithomason:
### This might be removed?

This option is only needed due to "primitive shorthand collection" values.

- If we remove collection shorthand (#2207), we won't need this option.
- If we remove primitive shorthand values (#2206), we don't need this option.

If we go with either of these, we can just delete the option and tell consumers to use `key` when they need to.

### Breaking Change?

This could this be a breaking change for consumers, if merged.  Their tests could rely on having keys or things like animations which also need keys sometimes even when not in a collection.

### Repeating history?

Lastly, I cannot recall why but it seems we used to once use a default of false and it didn't work out in Semantic UI React:

https://github.com/Semantic-Org/Semantic-UI-React/issues/2418

>>Each of those .create() methods can accept an options object as a second argument that gets passed through to createShorthand() - a flag could be added that disables the autogenerated key
>
>We used to enable generateKey as a createShorthandFactory() option which is not possible as we don't know if/when keys will be needed for a given factory. However, disabling keys at the time of creation seems like a good solution.
>
>I'd gladly merge a PR for this!

layershifter:
> This might be removed?

Yep, if any of this proposals will be merged this will be not needed anymore. I see this PR as a temporary resolution.

> Their tests could rely on having keys or things like animations which also need keys sometimes even when not in a collection.

Can you please provide a case? Enzyme allows only to [get keys](https://airbnb.io/enzyme/docs/api/ReactWrapper/key.html), but do not select as `key` is not a real prop, https://github.com/airbnb/enzyme/issues/645. I also don't see there a case for animations as keys will not generated only for slots inside component. And it's valid because React generates keys automatically for JSX markup.


> Repeating history?

Yep, that's why Earth is flat 😁 That's why I don't consider `false` as default value and decided to force its setup.

PR #2192, opened 2019-12-17
fix(Provider): render portals inside styled elements

### LINKED ISSUE #2138: Scrollbar inside a Dialog doesn't get Provider styles
## Bug Report
A scroll bar inside a `Dialog` component doesn't get `Provider`'s theme. The issue arises as the dialog is rendered as sibling to root element, it ends up without a provider wrapper.

sibling

### Steps
Render a `List` of items inside a `Dialog` component with styles override of `overflow: scroll`, you will see that the scroll bar styles is not in par with Provider theme. See the codesandbox link below to replicate the issue.

### Expected Result
scroll-bar-with-provider

### Actual Result
scroll-bar-without-provider

### Version
0.40.4

### Testcase
<a href="https://codesandbox.io/s/stardust-ui-example-tmvgh" target="_blank">
  Edit stardust-ui-example
</a>

layershifter:
Fixes #2138.

# BREAKING CHANGES

- `ProviderBox` component was removed
- styles from `ProviderBox` are moved to `Provider`

---

`ProviderBox` was previously used to set styles for DOM tree. It was impossible to do it within `Provider` previously due technical limitations of `UIComponent`.

This PR also solves #2138 as creates elements where components that use `createPortal()` will be rendered.

### Before

```html
<body>
  <div class="ui popup" />
</body>
```

### After

```html
<body>
  <div class="ui provider">
    <div class="ui popup" />
  </div>
</body>
```


jurokapsiar (approved):

packages/react/src/components/Dialog/Dialog.tsx
```diff
@@ -285,7 +286,8 @@ class Dialog extends AutoControlledComponent<WithAsProp<DialogProps>, DialogStat
 
     const dialogContent = (
       <Ref innerRef={this.contentRef}>
-        <ElementType
+        <ProviderBox
```
mnajdova: My proposal would be to give guidance on the consumers to use the `ProviderBox` on their side (that was our suggestion with the `PopupContent` component. If we decide to change this, we should address it consistently in all portal components.

CHANGELOG.md
```diff
@@ -32,6 +34,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Fix `input` descenders being cropped in the Teams theme @bcalvery ([#2335](https://github.com/microsoft/fluent-ui-react/pull/2335))
 - Use referentially equal objects for `actions` in `useStateManager` @layershifter ([#2347](https://github.com/microsoft/fluent-ui-react/pull/2347))
 - Fix `Animation` component not to throw when `children` is not provided @mnajdova ([#2345](https://github.com/microsoft/fluent-ui-react/pull/2345))
+- Styles from `Provider` and applied to components that rendered out of it in DOM @layershifter ([#2192](https://github.com/microsoft/fluent-ui-react/pull/2192))
```
miroslavstastny: ...are applied...?

miroslavstastny (approved):

CHANGELOG.md
```diff
@@ -33,6 +35,7 @@ This project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.htm
 - Use referentially equal objects for `actions` in `useStateManager` @layershifter ([#2347](https://github.com/microsoft/fluent-ui-react/pull/2347))
 - Fix `Animation` component not to throw when `children` is not provided @mnajdova ([#2345](https://github.com/microsoft/fluent-ui-react/pull/2345))
 - Fix `loader` - adding labeling when loader get focus @kolaps33 ([#2352](https://github.com/microsoft/fluent-ui-react/pull/2352))
+- Styles from `Provider` and applied to components that rendered out of it in DOM @layershifter ([#2192](https://github.com/microsoft/fluent-ui-react/pull/2192))
```
```suggestion
- Styles from `Provider` are applied to components that rendered out of it in DOM @layershifter ([#2192](https://github.com/microsoft/fluent-ui-react/pull/2192))
```

CHANGELOG.md
layershifter: Yep, good catch

PR #2193, opened 2019-12-17
fix(Dialog): make width responsive

### LINKED ISSUE #2139: <Dialog /> not responsive on smaller viewports
## Bug Report
Dialog is not responsive on smaller viewports

### Expected Result
Screen Shot 2019-12-03 at 11 44 03 AM

### Actual Result
Screen Shot 2019-12-03 at 11 43 58 AM

### Version
0.40.4

layershifter:
Fixes #2139.

This PR adds `media` selectors for `Dialog` to make it responsible:

image

image

image


packages/react/src/themes/teams/components/Dialog/dialogStyles.ts
```diff
@@ -11,11 +11,18 @@ export default {
...
     gridTemplateColumns: '1fr auto',
     boxShadow: v.boxShadow,
     color: v.foregroundColor,
+
+    '@media (min-width: 768px)': {
+      width: v.rootWidthMedium,
+    },
+    '@media (min-width: 1280px)': {
+      width: v.rootWidth,
+    },
```
layershifter, line 25: @codepretty can you please provide proper breakpoints? 🙏

packages/react/src/themes/teams/components/Dialog/dialogVariables.ts
```diff
@@ -25,7 +30,13 @@ export interface DialogVariables {
 export default (siteVariables): Partial<DialogVariables> => ({
   rootBackground: siteVariables.colors.white,
   rootBorderRadius: pxToRem(3),
-  rootWidth: '50vw',
+
+  rootWidth: '90vw',
+  rootWidthLarge: '50vw',
+  rootWidthLargeBreakpoint: '@media (min-width: 1280px)',
+  rootWidthMedium: '75vw',
+  rootWidthBreakpointMedium: '@media (min-width: 768px)',
```
mnajdova, line 38:
```suggestion
  rootWidthMediumBreakpoint: '@media (min-width: 768px)',
```

packages/react/src/themes/teams/components/Dialog/dialogVariables.ts
```diff
@@ -25,7 +30,13 @@ export interface DialogVariables {
 export default (siteVariables): Partial<DialogVariables> => ({
   rootBackground: siteVariables.colors.white,
   rootBorderRadius: pxToRem(3),
-  rootWidth: '50vw',
+
+  rootWidth: '90vw',
+  rootWidthLarge: '50vw',
+  rootWidthLargeBreakpoint: '@media (min-width: 1280px)',
```
mnajdova, line 36: I am kind a leaning more towards allowing users to specify only the px value here, not the whole string, but I am not sure. We should make that decision now, as this is the first PR that is adding breakpoints

packages/react/src/themes/teams/components/Dialog/dialogStyles.ts
```diff
@@ -16,6 +16,13 @@ export default {
     gridTemplateColumns: '1fr auto',
     boxShadow: v.boxShadow,
     color: v.foregroundColor,
+
+    [v.rootWidthBreakpointMedium]: {
```
mnajdova, line 20: Shouldn't we check whether it is defined first?

packages/react/src/themes/teams/components/Dialog/dialogVariables.ts
layershifter: I am not sure that we should limit user there as media queries support a lot of attributes:
https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries

packages/react/src/themes/teams/components/Dialog/dialogStyles.ts
layershifter: It's required by typings. Why we need to have additional checks there?

packages/react/src/themes/teams/components/Dialog/dialogStyles.ts
notandrew: Trying to nail down the breakpoint logic with Design... Almost have it figured out.

PR #2199, opened 2020-01-02
docs: accessibility fix for component toolbar

kolaps33:
I recognized that in the production code
https://microsoft.github.io/fluent-ui-react/components/accordion/definition

In current production, when user navigate in component toolbar:
- screen reader doesn't narrate each button name in toolbar
- "CodeSandbox" and "Permalink" buttons doesn't work with space/enter key

This PR fix:
- each button will be read 
- on each button space/enter will execute action


jurokapsiar (approved):

docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
```diff
@@ -15,82 +24,86 @@ type ComponentControlsCodeSandboxProps = {
...
+          name={exampleName}
+          providedFiles={{
+            [main]: { content: appTemplate },
+            'package.json': createPackageJson(main, exampleLanguage),
+          }}
+          skipRedirect
+          ref={this.codeSandboxerRef}
+          template={template}
+        >
+          {() => null}
```
layershifter, line 104:
```suggestion
          // CodeSandboxer captures ref from children and binds an event handler directly to DOM, this hack allows to avoid this behavior
          {() => null}
```

Let's add a comment 💬

docs/src/components/ComponentDoc/ComponentControls/ComponentControlsCodeSandbox/ComponentControlsCodeSandbox.tsx
```diff
@@ -4,9 +4,18 @@ import CodeSandboxer from 'react-codesandboxer'
 import { ComponentSourceManagerLanguage } from 'docs/src/components/ComponentDoc/ComponentSourceManager'
 import { appTemplate } from './indexTemplates'
 import createPackageJson from './createPackageJson'
-import { Icon, Tooltip } from 'src/index'
+
+export enum CodeSandboxState {
+  Default = 'DEFAULT',
```
layershifter, line 9:
```suggestion
  Idle = 'IDLE',
```

_nit:_ I think `idle` will be better

layershifter (approved):

layershifter (approved):

PR #2202, opened 2020-01-02
fix(Popup): fix click outside when trigger is not defined

layershifter:
This PR fixes a cases when clicks outside of `Popup` were not properly handled in controlled mode without `trigger` prop.

### DIFF
--- a/packages/react/src/components/Popup/Popup.tsx
+++ b/packages/react/src/components/Popup/Popup.tsx
@@ -296,11 +296,11 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
 
   isOutsidePopupElementAndOutsideTriggerElement(refs: NodeRef[], e) {
     const isOutsidePopupElement = this.isOutsidePopupElement(refs, e)
-    const isOutsideTriggerElement =
+    const isInsideTriggerElement =
       this.triggerRef.current &&
-      !doesNodeContainClick(this.triggerRef.current, e, this.context.target)
+      doesNodeContainClick(this.triggerRef.current, e, this.context.target)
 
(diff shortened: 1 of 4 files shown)


packages/react/src/components/Popup/Popup.tsx
```diff
@@ -296,11 +296,11 @@ export default class Popup extends AutoControlledComponent<PopupProps, PopupStat
 
   isOutsidePopupElementAndOutsideTriggerElement(refs: NodeRef[], e) {
     const isOutsidePopupElement = this.isOutsidePopupElement(refs, e)
-    const isOutsideTriggerElement =
+    const isInsideTriggerElement =
       this.triggerRef.current &&
-      !doesNodeContainClick(this.triggerRef.current, e, this.context.target)
+      doesNodeContainClick(this.triggerRef.current, e, this.context.target)
```
mnajdova, line 301: Just an observation, this is not exactlly the negated condition. Is that expected? The negated would look like this:
```
  const isInsideTriggerElement =
      !this.triggerRef.current || doesNodeContainClick(this.triggerRef.current, e, this.context.target)
```
Just mentioning in case if this was done only for readability

mnajdova (approved):

packages/react/src/components/Popup/Popup.tsx
layershifter: Yep, it's expected as to check that click was inside we need to have a `trigger` element and click should inside of it. And if click was `isOutsidePopupElement` and `!isInsideTriggerElement` we can close `Popup`. But agree, this conditions are a bit crazy.

silviuaavram (approved):

PR #2203, opened 2020-01-03
fix: use and mock lodash debounce in Carousel and Dropdown

silviuaavram:
Removed in-house implementation of `debounce` and replaced its uses with `_.debounce` in `Dropdown` and `Carousel`.

Properly mocked the function in `Dropdown-test` as it wasn't working with jest timers. Details in https://github.com/facebook/jest/issues/3465 and other similar Github issues.


packages/react/src/components/Dropdown/Dropdown.tsx
```diff
@@ -361,15 +361,8 @@ class Dropdown extends AutoControlledComponent<WithAsProp<DropdownProps>, Dropdo
...
   }
 
-  a11yStatusTimeout: any
-  charKeysPressedTimeout: any
   defaultTriggerButtonId = _.uniqueId('dropdown-trigger-button-')
 
-  componentWillUnmount() {
-    clearTimeout(this.a11yStatusTimeout)
-    clearTimeout(this.charKeysPressedTimeout)
-  }
```
layershifter, line 371: Please add an `cancel()` call for `_.debounce`:

https://lodash.com/docs/4.17.15#debounce

silviuaavram: you mean `this.clearStartingString.cancel()` ?

packages/react/src/components/Carousel/Carousel.tsx
```diff
@@ -213,8 +212,7 @@ class Carousel extends AutoControlledComponent<WithAsProp<CarouselProps>, Carous
   itemRefs = [] as React.RefObject<HTMLElement>[]
   paddleNextRef = React.createRef<HTMLElement>()
   paddlePreviousRef = React.createRef<HTMLElement>()
-
-  focusItemAtIndex = debounce((index: number) => {
+  focusItemAtIndex = _.debounce((index: number) => {
     this.itemRefs[index].current.focus()
```
dzearing: If you're debouncing, you'll need to check `current` for null. There is a slight race condition where the debounced callback can occur asynchronously after unmounting.

packages/react/src/components/Dropdown/Dropdown.tsx
dzearing: I believe `this.focusItemAtIndex.cancel()` as well

packages/react/src/components/Carousel/Carousel.tsx
```diff
@@ -213,8 +212,7 @@ class Carousel extends AutoControlledComponent<WithAsProp<CarouselProps>, Carous
   itemRefs = [] as React.RefObject<HTMLElement>[]
   paddleNextRef = React.createRef<HTMLElement>()
   paddlePreviousRef = React.createRef<HTMLElement>()
-
-  focusItemAtIndex = debounce((index: number) => {
+  focusItemAtIndex = _.debounce((index: number) => {
```
dzearing, line 219: Minor nullref edge case if you don't cancel this.

packages/react/src/components/Dropdown/Dropdown.tsx
silviuaavram: Solved. Thanks!

layershifter (approved):

PR #2204, opened 2020-01-03
prototype(chat pane): accessibility fix attachment in chat

kolaps33:
There was the bug in attachment for chat prototype, that we had "button" trigger element for popup in another "button" element.

Issue was that "action" slot was not replaced by trigger button but was inserted into slot. And it cause "button" in another "button". 

Issue coming from the change done in the PR:
https://github.com/microsoft/fluent-ui-react/pull/1513

### DIFF
--- a/docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
+++ b/docs/src/prototypes/chatPane/services/messageFactoryMock.tsx
@@ -1,7 +1,6 @@
 import {
   Attachment,
   Popup,
-  Button,
   Menu,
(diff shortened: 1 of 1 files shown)


silviuaavram:
Code looks fine, but I can't understand what is fixed, what is broken etc. A bug link would be useful, or at least a description of what the bug was and why this PR fixes it.

jurokapsiar (approved):

kolaps33:
@silviuavram improved description of the PR. Thanks for point it out, I forgot did it.

PR #2207, opened 2020-01-03
feat(List): add proper support for children API

layershifter:
This PR uses `context-selectors` to add full support for Children API in `List` component.

# Performance 🏎 

There is no real difference between shorthand/Context usage.

### With Context usage

| Example             | min    | avg    | median | max    | renderComponent.min | renderComponent.avg | renderComponent.median | renderComponent.max | components |
| ------------------- | ------ | ------ | ------ | ------ | ------------------- | ------------------- | ---------------------- | ------------------- | ---------- |
| ListNested.perf.tsx | 793.39 | 818.49 | 817.41 | 857.03 | 641.66              | 668.76              | 667.75                 | 700.61              | 5426       |
| ListNested.perf.tsx | 797.84 | 817.34 | 815.36 | 858.54 | 650.95              | 667.88              | 665.44                 | 702.12              | 5426       |
| ListNested.perf.tsx | 800.74 | 821.73 | 820.21 | 858.53 | 649.26              | 671.6               | 671.02                 | 701.88              | 5426       |

---

### Previous implementation

| Example             | min    | avg    | median | max     | renderComponent.min | renderComponent.avg | renderComponent.median | renderComponent.max | components |
| ------------------- | ------ | ------ | ------ | ------- | ------------------- | ------------------- | ---------------------- | ------------------- | ---------- |
| ListNested.perf.tsx | 806.89 | 836.46 | 836.58 | 868.65  | 643.83              | 666.22              | 665.62                 | 691.24              | 5426       |
| ListNested.perf.tsx | 812.05 | 855.68 | 840.79 | 1024.46 | 643.09              | 681.92              | 671.1                  | 806.44              | 5426       |


packages/react-context-selector/src/utils.ts
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react'
+
+export const CONTEXT_SUBSCRIBE_PROPERTY = 's'
```
joschect: I might suggest looking into `Symbol.for('ContextSubscribeProperty`) it will help to ensure that the Symbol is unique to that text without the cost of passing that string around.

joschect: As a note, that's how react identifies it's components.

joschect: Although, IE doesn't support it.

layershifter: I will remove these constants until we will have a thing that measures bundle size

packages/react-context-selector/src/createContext.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  /* istanbul ignore else */
+  if (process.env.NODE_ENV !== 'production') {
+    Provider.displayName = 'ContextSelector.Provider'
+  }
+
+  return Provider
+}
+
+type CreateContextOptions = {
```
miroslavstastny: This type should be exported as it is used in public API

packages/react-context-selector/src/createContext.ts
```diff
@@ -0,0 +1,71 @@
...
+  strict?: boolean
+}
+
+export const createContext = <Value>(
```
miroslavstastny: Up until now we preferred default exports. Do you think we should move to named exports?

packages/react/src/components/List/List.tsx
```diff
@@ -136,38 +129,28 @@ const List: React.FC<WithAsProp<ListProps>> &
...
-        _.invoke(props, 'onSelectedIndexChange', e, {
-          ...props,
-          ...{ selectedIndex: itemProps.index },
-        })
+        actions.select(itemIndex)
+        _.invoke(props, 'onSelectedIndexChange', e, { ...props, selectedIndex: itemIndex })
       }
     },
-  })
+    [actions],
```
miroslavstastny: Having `actions` as the only dependency means you can get stale `props` in `onSelectedIndexChange` cb.

packages/react/src/components/List/context.ts
```diff
@@ -0,0 +1,31 @@
+import { createContext } from '@fluentui/react-context-selector'
```
miroslavstastny: I would name this file `ListContext` for easier developer navigation.

packages/react/src/components/List/context.ts
```diff
@@ -0,0 +1,31 @@
...
+    truncateContent: false,
+    truncateHeader: false,
+    variables: {},
+
+    onItemClick: () => {},
+    selectedIndex: -1,
+  },
+  { strict: false },
+)
+export const Provider = ListContext.Provider
```
miroslavstastny: I would be explicit here and name it `ListContextProvider`

packages/react-context-selector/src/useContextSelectors.ts
```diff
@@ -0,0 +1,80 @@
+import * as React from 'react'
+
+import { Context, ContextSelector, ContextValue } from './types'
+import { useIsomorphicLayoutEffect } from './utils'
+
+type UseSelectorRef<
```
miroslavstastny:
```suggestion
type UseSelectorsRef<
```

packages/react-context-selector/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@fluentui/react-context-selector",
+  "description": "TODO",
```
miroslavstastny: TODO

packages/react-context-selector/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "author": "Oleksandr Fediashov <a@fedyashov.com>",
+  "bugs": "https://github.com/microsoft/fluent-ui-react/issues",
+  "dependencies": {
+    "@babel/runtime": "^7.1.2"
+  },
+  "devDependencies": {
+    "@fluentui/internal-tooling": "^0.43.0",
+    "@types/react-is": "^16.7.1",
+    "lerna-alias": "^3.0.3-0",
+    "react": "^16.8.0",
```
miroslavstastny: is it necessary to have `react` as devDependency?

packages/react/src/components/List/ListItem.tsx
```diff
@@ -98,13 +100,26 @@ const ListItem: React.FC<WithAsProp<ListItemProps> & { index: number }> &
...
-    debug,
-    navigable,
-    selectable,
-    selected,
-    truncateContent,
-    truncateHeader,
-    variables,
+  } = props
+
+  const parentProps = useContextSelectors(ListContext, {
```
miroslavstastny: The typings does not seem to be strict enough to allow only the valid keys and values.
Let's verify it together offline.

packages/react/src/components/List/List.tsx
layershifter: Yep, you're totally right. I will tackle this separately

packages/react-context-selector/src/useContextSelectors.ts
layershifter: 👍

packages/react-context-selector/src/createContext.ts
layershifter: Good catch, done

packages/react-context-selector/src/createContext.ts
layershifter: Nope, previously all things were in one file. Moved to `defaults`

packages/react-context-selector/package.json
layershifter: Done in https://github.com/microsoft/fluent-ui-react/pull/2292/files#diff-e0c42cd190baafa1f2be63eef8c72bc4R3

packages/react-context-selector/package.json
layershifter: Yep, as we need it for unit tests

packages/react/src/components/List/context.ts
layershifter: Renamed 👍

packages/react/src/components/List/context.ts
layershifter: Renamed to `listContext.ts` to avoid issues with our docgen

miroslavstastny (approved):

PR #2208, opened 2020-01-03
fix(AlertBehavior) add missing aria-describedby to AlertBehavior when…

### LINKED ISSUE #2197: AlertBehavior does not add aria-describedby to the dismiss action
### Steps
open https://microsoft.github.io/fluent-ui-react/components/alert/definition#types-dismissible
check if dismiss action has aria-describedby

### Expected Result
aria-describedby points to the body of the alert

### Actual Result
aria-describedby is missing

### Version
0.42.0

### Testcase
https://microsoft.github.io/fluent-ui-react/components/alert/definition#types-dismissible

delprzemo:
This PR fixes issue with lack of `aria-describedby `in `AlertBehavior `when dismiss action is defined and type is different than warning/danger

Fixes #2197


delprzemo:
Need a tip how can I fix screener tests in that case. 
`child "apiKey" fails because ["apiKey" is not allowed to be empty]`
Shouldn't that be handled by environment variables on the agent machines?

Edit. After checking screener.yml I guess it may be related to branch and PR via fork

jurokapsiar (approved):
@delprzemo sorry for the delay, this looks good. please update the branch to master and check if the changelog entry gets to the unreleased section. thanks for your contribution!

delprzemo:
> @delprzemo sorry for the delay, this looks good. please update the branch to master and check if the changelog entry gets to the unreleased section. thanks for your contribution!

@jurokapsiar No problem. Branch updated to master - changelog entry in the unreleased section. I don't have write access so could you merge it?

jurokapsiar:
foreced push despite failing screener - currently the integration does not work for forked repos. This PR does not introduce any visual changes.

